Issue #7934, opened 2023-02-15
Sqlsmith: Sql generation


kwannoel:
As we add more features, these should be tested via sqlsmith as well.

SQL Features:
- [ ] Generate other table functions: `generate_series`, `unnest`. See: https://github.com/singularity-data/risingwave/issues/4540
- [ ] Generate array_agg
- [ ] Generate select from System Tables
- [ ] Generate `NATURAL JOIN`
- [ ] Generate `JOIN` with `USING`
- [ ] Generate query w/o `GROUP BY`
- [ ] Randomly toggle session variables
- [ ] Generate `pow`.
- [ ] Generate TopN by group: https://www.risingwave.dev/docs/current/sql-pattern-topn/
- [ ] Generate jsonb exprs, see: https://github.com/risingwavelabs/risingwave/issues/7714

PR #19333, opened 2024-11-11
refactor(agg): remove `state = "ref"` from `bool_and` + refactor `aggregate` macro

stdrc:

Previously `aggregate` macro didn't correctly handle aggregates that require the first (non-NULL) input as its "first state". For example, for `bool_and`, the behaviors of with and without `state = "ref"` are different, of which the latter one produce wrong results. This PR refactors the state maintanence code generated by the `aggregate` macro, as a result, we can remove `state = "ref"` from append-only `bool_and` to align with `bool_or`, avoiding further confusion and maintanence error.

## Checklist

## Documentation

## Release note

</details>

### DIFF
--- a/src/expr/impl/src/aggregate/bool_and.rs
+++ b/src/expr/impl/src/aggregate/bool_and.rs
@@ -48,7 +48,7 @@ use risingwave_expr::aggregate;
 /// statement ok
 /// drop table t;
 /// ```
-#[aggregate("bool_and(boolean) -> boolean", state = "ref")]
+#[aggregate("bool_and(boolean) -> boolean")]
 fn bool_and_append_only(state: bool, input: bool) -> bool {
     state && input
 }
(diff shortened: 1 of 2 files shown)


stdrc:
Tests failed, because `count` depends on the old behavior that creates a state with default value of the state data type and apply the function with the default state + first input.

To be more concrete, we need to fit all the following cases with one macro:

```
sum:
1. NULL if no non-NULL input values
2. On first input, `state = first input`, or, `state = 0` and apply `agg(state, value)`

sum0:
1. 0 if no non-NULL input values

count:
1. NULL if no non-NULL input values
2. On first input, `state = 0` and apply `agg(state, value)`

bool_or:
1. NULL if no non-NULL input values
2. On first input, `state = first input`, or, `state = false` and apply `agg(state, value)`

bool_and:
1. NULL if no non-NULL input values
2. On first input, `state = first input`, or, `state = true` and apply `agg(state, value)`
```

This requires us to revisit the macro arguments set when defining these aggregates. Now I want to leave it as it is for now.

Issue #12539, opened 2023-09-26
Discussion: Do we expect stream keys the same for continuous `UpdateDelete` and `UpdateInsert` in StreamChunk?


stdrc:
I expected continuous `UpdateDelete` and `UpdateInsert` rows in StreamChunk should have the same stream key, especially when we iterate over `chunk.records()` this can be quite intuitive. But in reality we seems never reached a consensus on this, and some executors produce stream chunks with different stream keys in continuous `UpdateDelete` and `UpdateInsert`, and hence https://github.com/risingwavelabs/risingwave/pull/12536.

I think the current situation forces executors to `match` on `Op` and view `UpdateDelete` and `Delete`, `UpdateInsert` and `Insert` the equivalent. That makes the notion of `Update` almost meaningless. The only semantics of `UpdateDelete` seems to be there's guaranteed to be an `UpdateInsert` after it, nothing more.

Any ideas?

---

An example:

Commit: 471bfa2c91 (which is the commit before #12536 is merged)

```sql
create table foo (
    a int,
    b int,
    c int,
    primary key (a, b, c)
);

create table bar (
    b int,
    d int,
    primary key (b)
);

create materialized view mv as
select
    foo.a a,
    foo.b b,
    bar.d d,
    row_number() over(partition by foo.a, foo.b order by foo.c desc) n
from
    foo left join bar on foo.b = bar.b;

insert into bar values (100, null), (200, null);
insert into foo values (1, 100, 2);

update bar set d = 10 where b = 100;
```

The last query will cause the `OverWindowExecutor` panic, of which the reason is because `HashJoin` produces `Update` with different stream keys and meanwhile `OverWindow` expects the `old_row` and `new_row` have the same stream key.

The explain:

```
                                                                             QUERY PLAN
---------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Fragment 0
 StreamMaterialize { columns: [a, b, d, n, foo.c(hidden), bar.b(hidden)], stream_key: [a, b, foo.c, bar.b], pk_columns: [a, b, foo.c, bar.b], pk_conflict: NoCheck }
 ├── materialized table: 4294967294
 ├── output: [ foo.a, foo.b, bar.d, row_number, foo.c, bar.b ]
 ├── stream key: [ foo.a, foo.b, foo.c, bar.b ]
 └── StreamProject { exprs: [foo.a, foo.b, bar.d, row_number, foo.c, bar.b] }
[...]

PR #19347, opened 2024-11-12
refactor(stream): add more assertion on update op handling in hash dispatcher

BugenZhao:

The logic for rewriting update `Op`s in the hash dispatcher relies on the invariant held in `update_check`, which is not enforced in production.

Some users encounter an assertion failure when executing this line,

https://github.com/risingwavelabs/risingwave/blob/9a01f67f7d6b84d57382e5b47df1fdea14249897/src/stream/src/executor/dispatch.rs#L835

where `col.len()` is 256 while `ops.len()` is 255, so I highly suspect there might be something wrong during the `Op` rewrite.

This PR adds some mroe assertions to help narrow down the scope of the issue.

We may also consider enabling `update_check` in production, or thoroughly abandon this convention as per discussion in https://github.com/risingwavelabs/risingwave/issues/12539.

## Checklist

## Documentation

## Release note

</details>

### DIFF
--- a/src/stream/src/executor/dispatch.rs
+++ b/src/stream/src/executor/dispatch.rs
@@ -787,6 +787,10 @@ impl Dispatcher for HashDataDispatcher {
             }
 
             if !visible {
+                assert!(
+                    last_vnode_when_update_delete.is_none(),
+                    "invisible row between U- and U+, op = {op:?}",
+                );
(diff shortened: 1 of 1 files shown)


stdrc (approved):
LGTM

st1page (approved):

PR #17156, opened 2024-06-06
fix(expr): don't panic on table function errors

wangrunji0408:
Signed-off-by: Runji Wang <wangrunji0408@163.com>I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://gist.github.com/TennyZhuang/f00be7f16996ea48effb049aa7be4d66#file-rw_cla).

## What's changed and what's your intention?

partially fix_ #11915 

Currently, when an error occurs in a table function call, the returned stream stops and the streaming actor will exit. This PR makes streaming tolerant to table function errors. When an error occurs, an error column will be appended to the returned chunk. (see the example in [arrow-udf](https://github.com/risingwavelabs/arrow-udf/tree/main/arrow-udf#fallible-functions)) The batch side will raise error immediately when seeing this column, while the streaming side will report errors and keep going. If some rows have been output when the error occurs, these rows will not be reverted.

## Checklist

- [x] I have written necessary rustdoc comments
- [x] I have added necessary unit tests and integration tests
- [ ] I have added test labels as necessary. See [details](https://github.com/risingwavelabs/risingwave/blob/main/docs/developer-guide.md#ci-labels-guide).
- [ ] I have added fuzzing tests or opened an issue to track them. (Optional, recommended for new SQL features #7934).
- [ ] My PR contains breaking changes. (If it deprecates some features, please create a tracking issue to remove them in the future).
- [x] All checks passed in `./risedev check` (or alias, `./risedev c`)
- [ ] My PR changes performance-critical code. (Please run macro/micro-benchmarks and show the results.)

- [x] My PR contains critical fixes that are necessary to be merged into the latest release. (Please check out the [details](https://github.com/risingwavelabs/risingwave/blob/main/CONTRIBUTING.md))

## Documentation

- [ ] My PR needs documentation updates. (Please use the **Release note** section below to summarize the impact on users)

## Release note

If this PR includes changes that directly affect users or other significant modifications relevant to the community, kindly draft a release note to provide a concise summary of these changes. Please prioritize highlighting the impact these changes will have on users.

</details>

PR #19250, opened 2024-11-04
test: vnode count chaos test in e2e madsim

BugenZhao:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://gist.github.com/TennyZhuang/f00be7f16996ea48effb049aa7be4d66#file-rw_cla).

## What's changed and what's your intention?

Introduce a new CLI arg `--random-vnode-count` for madsim e2e tests, which sets the `STREAMING_MAX_PARALLELISM` to a random value before running each DDL statement.

This covers the behavior of multiple jobs with varying vnode counts within a single cluster, especially when they have dependencies in the streaming graph.

Added as a new step in main-cron.

----

An example of passed run in PR workflow: https://buildkite.com/risingwavelabs/pull-request/builds/61753#01930563-f129-4363-9599-7b25af97510a

## Checklist

- [x] I have written necessary rustdoc comments
- [x] I have added necessary unit tests and integration tests
- [x] All checks passed in `./risedev check` (or alias, `./risedev c`)

## Documentation

- [ ] My PR needs documentation updates. (Please use the **Release note** section below to summarize the impact on users)

## Release note

If this PR includes changes that directly affect users or other significant modifications relevant to the community, kindly draft a release note to provide a concise summary of these changes. Please prioritize highlighting the impact these changes will have on users.

</details>

PR #19353, opened 2024-11-12
fix(streaming): correctly skip problematic table-function input

### LINKED ISSUE #19352: bug: streaming table function result missing if other input rows lead to error
Since https://github.com/risingwavelabs/risingwave/pull/17156, we should be able to tolerate table function error in streaming mode. However, the implementation seems incorrect: a problematic input can cause the results of other input rows to be missing, if they are in the same chunk.

To reproduce:

```sql
-- ensure the same chunk
set streaming_parallelism to 1;

create table t (x int);
create materialized view mv as select x, generate_series(1, 2, x) from t;

insert into t values (0), (1);
flush;

select * from mv;
-- got empty result, expected (x = 1) to be successfully executed
```

Revealed by https://github.com/risingwavelabs/risingwave/pull/19250

BugenZhao:

See #19352. Close #19352.

----

The behavior for table function returning an error in streaming mode will become:

https://github.com/risingwavelabs/risingwave/blob/c6d2ba881c59cd04b3940f163f4ec5f2339d23b0/src/stream/src/executor/project_set.rs#L205-L213

As described in #17156, there's currently a limitation that

> If some rows have been output when the error occurs, these rows will not be reverted.

which behaves kind of strangely. Since it might be uncommon (?) for a table function to fail halfway, I prefer not to document this behavior.

## Checklist

## Documentation

## Release note

</details>


BugenZhao:
* **#19250** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/19250?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#19353** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/19353?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @BugenZhao and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

src/stream/src/executor/project_set.rs
```diff
@@ -174,22 +184,33 @@ impl Inner {
                                 unsafe { std::mem::transmute(row.as_mut_slice()) };
 
                             row[0] = Some(projected_row_id.into());
-                            // if any of the set columns has a value
-                            let mut valid = false;
+
+                            // Whether all table functions has exhausted or has failed for current input row.
+                            let mut fully_consumed = true;
```
stdrc, line 189: In which case it's considered as "exhausted"?

BugenZhao: When this condition doesn't meet:

```rust
if let Some((i, result)) = state.peek()
  && is_current_input(i)
```

i.e., the cursor of the `state` has moved to the next input.

stdrc (approved):

Issue #19245, opened 2024-11-04
Bug: Resources are not properly isolated in Standalone mode


Li0k:
When RW starts in default mode, each component is deployed in a different pod and utilises separate memory.

For example 
- CN calculates the memory resources that hummock can use with the function `storage_memory_config
- Compactor uses all the memory provided by the system by default and divides it between worker and cache.

However, this can lead to OOM in `Standalone` mode. In `Standalone` mode we don't isolate CPU and Memory resources, and competition for CPU is natural, but not for Memory.

Li0k:
cc @kwannoel @hzxa21

lmatz:
I wonder:
1. If there is a minimum memory requirement for the compactor
2. if there is any relationship between the live CPU usage and the memory usage of the compactor, e.g. if the compactor is using 2CPU at the moment, is there a cap on the memory usage?


I am wondering if we also limit the CPU usage of compactor a bit to ensure stability at the cost of potentially leaving idle resources on the table, e.g. if 8CPU in total, then the compactor can take half = 4CPU at most.

Li0k:
> I wonder:
> 
> 1. If there is a minimum memory requirement for the compactor
> 2. if there is any relationship between the live CPU usage and the memory usage of the compactor, e.g. if the compactor is using 2CPU at the moment, is there a cap on the memory usage?
> 
> I am wondering if we also limit the CPU usage of compactor a bit to ensure stability at the cost of potentially leaving idle resources on the table, e.g. if 8CPU in total, then the compactor can take half = 4CPU at most.

1. It relay on the config of `sstable_size` and `block_size`
```rust
let min_compactor_memory_limit_bytes = (storage_opts.sstable_size_mb * (1 << 20)
            + storage_opts.block_size_kb * (1 << 10))
            as u64;
```

2. The number of tasks is limited by the number of available cpu cores, the more tasks the more memory is used. The amount of memory consumed by each task is related to the content of the task and there is no general formula for calculating this. (Do not isolation the cpu core).
3. I have no bias for CPUs, relying on Tokio scheduling, cpu competition is fair and doesn't cause fatal problems. 

Also, if we plan to deploy standalone on a high-spec machine (e.g. 64c 256GB), as you said, not having CPU isolation may cause stability issues, as it did in the affnity test.

hzxa21:
[...]

PR #19372, opened 2024-11-13
feat(memory): Separate total memory configurations for FE and Compactor

Li0k:

part of https://github.com/risingwavelabs/risingwave/issues/19245

This PR introduces a separate total memory configuration for FE and Compactor.

In standalone deployment mode, the total_memory of different nodes should be set to achieve memory isolation.  (Not included in this PR)

## Checklist

## Documentation

## Release note

</details>


hzxa21 (approved):

chenzl25 (approved):
LGTM

src/frontend/src/lib.rs
```diff
@@ -159,6 +160,10 @@ pub struct FrontendOpts {
         default_value = "./secrets"
     )]
     pub temp_secret_file_dir: String,
+
+    /// Total available memory for the frontend node in bytes. Used by both computing and storage.
+    #[clap(long, env = "RW_FE_TOTAL_MEMORY_BYTES", default_value_t = default_fe_total_memory_bytes())]
```
chenzl25, line 165: change `RW_FE_TOTAL_MEMORY_BYTES` to `RW_FRONTEND_TOTAL_MEMORY_BYTES`?

src/frontend/src/lib.rs
```diff
@@ -159,6 +160,10 @@ pub struct FrontendOpts {
         default_value = "./secrets"
     )]
     pub temp_secret_file_dir: String,
+
+    /// Total available memory for the frontend node in bytes. Used by both computing and storage.
```
kwannoel, line 164:
```suggestion
    /// Total available memory for the frontend node in bytes. Used by frontend.
```

src/storage/src/hummock/store/hummock_storage.rs
```diff
@@ -724,7 +724,6 @@ impl HummockStorage {
     }
 
     /// Used in the compaction test tool
-    #[cfg(any(test, feature = "test"))]
     pub async fn update_version_and_wait(&self, version: HummockVersion) {
         use tokio::task::yield_now;
         let version_id = version.id;
```
kwannoel, line 729: Are changes in this file expected?

src/storage/compactor/src/server.rs
```diff
@@ -81,7 +82,7 @@ pub async fn prepare_start_parameters(
         &system_params_reader,
         &storage_memory_config,
     )));
-    let non_reserved_memory_bytes = (system_memory_available_bytes() as f64
+    let non_reserved_memory_bytes = (compactor_opts.compactor_total_memory_bytes as f64
```
kwannoel, line 85: Compactor opts provide an instance of `compactor_total_memory_bytes`, but inside `RwConfig`, there's a separate value:
```
    pub compactor_memory_limit_mb: Option<usize>,
```

What's the difference between these configurations? Will the one in RwConfig always be ignored, if so can we remove it?

kwannoel (approved):
Generally LGTM, thanks for your work :)

src/storage/src/hummock/store/hummock_storage.rs
Li0k: right

src/storage/compactor/src/server.rs
Li0k: Good question:

I found that compactor_memory_limit_mb is only used in embedding compactor mode and occupies part of the CN's memory (you can refer to the storage_memory_config function).

I think this is different from standalone deployment mode, I didn't change it in this PR, maybe you have a better idea?

kwannoel: > Good question:
> I found that compactor_memory_limit_mb is only used in embedding compactor mode and occupies part of the CN's memory (you can refer to the storage_memory_config function).
> I think this is different from standalone deployment mode, I didn't change it in this PR, maybe you have a better idea?

Got your point. It's out of scope for this PR then. We can just rename it.
IIRC embedded compactor is not used for standalone deployment.

hzxa21:
Should we cherry pick this to 2.1?

src/frontend/src/lib.rs
fuyufjh, line 164: Does storage run in frontend? IIRC, even in local execution mode, the `TableScan` is pushed to compute nodes.

hzxa21: I think this comment is not resolved.

PR #19384, opened 2024-11-13
fix(source): `REFRESH SCHEMA` shall keep `INCLUDE` pk for `UPSERT`

xiangjinwu:

It is required that an `UPSERT` source (actually table) to `INCLUDE key` and use it as primary key. However, when `REFRESH SCHEMA`, the `INCLUDE`s are always lost, effectively resulting in `UPSERT` tables cannot evolve their schemas.

## Checklist

## Documentation

## Release note

</details>

### DIFF
--- a/e2e_test/source_inline/kafka/avro/alter_table.slt
+++ b/e2e_test/source_inline/kafka/avro/alter_table.slt
@@ -78,3 +78,20 @@ ABC
 
 statement ok
 drop table t;
+
+statement ok
+create table t (primary key (kafka_key))
+INCLUDE key as kafka_key
+WITH (
+    ${RISEDEV_KAFKA_WITH_OPTIONS_COMMON},
+    topic = 'avro_alter_table_test'
+)
+FORMAT UPSERT ENCODE AVRO (
+    schema.registry = '${RISEDEV_SCHEMA_REGISTRY_URL}'
+);
+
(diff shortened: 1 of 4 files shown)


e2e_test/source_inline/kafka/avro/alter_table.slt
```diff
@@ -78,3 +78,20 @@ ABC
...
+)
+FORMAT UPSERT ENCODE AVRO (
+    schema.registry = '${RISEDEV_SCHEMA_REGISTRY_URL}'
+);
+
+statement ok
+ALTER TABLE t REFRESH SCHEMA;
+
+statement ok
+drop table t;
```
tabVersion, line 97: no need to check the new schema meets expectation?

tabVersion (approved):
thanks for the fix

e2e_test/source_inline/kafka/avro/alter_table.slt
xiangjinwu: This is actually a no-op refresh. We refresh immediately after creation without any schema change. But yes it is always good to test more:
* schema meets expectation after refresh
* actual data meets schema after refresh (rather than panic due to mismatch)
* works for protobuf in addition to avro
* works for debezium in addition to plain or upsert

PR #19399, opened 2024-11-15
change `Binder::included_relations()` to return reference instead of a clone

stdrc:

## Checklist

## Documentation

## Release note

</details>


stdrc:
* **#19399** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/19399?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/19399?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

src/meta/src/manager/streaming_job.rs
```diff
@@ -299,7 +299,7 @@ impl StreamingJob {
         }
     }
 
-    // TODO: record all objects instead.
+    // TODO: to be removed, pass all objects uniformly through `dependencies` field instead.
     pub fn dependent_relations(&self) -> Vec<u32> {
```
xxchan, line 303: It seems `table.dependent_relations` is not set any more. Will it break sth to call this function?

xxchan: Oh. The caller uses `extend`

xxchan: I would be in favor of removing this field together now to avoid mistakes. Can put in an upstack PR and merge them together.

stdrc: > It seems `table.dependent_relations` is not set any more. Will it break sth to call this function?

I merged the result of this method and the new `dependencies` field to avoid breaking. And now it basically return empty vector for `MaterializedView` and `Sink`, and remain the same for `Table` and `Index`.

Since this PR is mainly for `drop function` check, I only changed the behavior of `MaterializedView` and `Sink`. Added TODO comments for the remaining ones.

src/meta/src/controller/streaming_job.rs
```diff
@@ -308,17 +316,19 @@ impl CatalogController {
...
+        );
 
-        let dependent_objs = dependent_relations
-            .iter()
-            .chain(dependent_secret_ids.iter());
         // record object dependency.
-        if !dependent_secret_ids.is_empty() || !dependent_relations.is_empty() {
-            ObjectDependency::insert_many(dependent_objs.map(|id| {
+        if !dependencies.is_empty() {
+            ObjectDependency::insert_many(dependencies.into_iter().map(|oid| {
```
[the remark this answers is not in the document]
stdrc: I don't quite get it. Can you elaborate? I'm not very familiar with this part. I suppose I impled the same behavior as `Secret`?

e2e_test/udf/drop_function.slt
```diff
@@ -0,0 +1,55 @@
...
+create materialized view mv as select add(a, b) as c from t;
+
+statement error
+drop function add;
+----
+db error: ERROR: Failed to run the query
+
+Caused by:
+  Permission denied: PermissionDenied: function used by 1 other objects.
+
```
BugenZhao, line 22: Shall we use inline error matching?

```suggestion
statement ok function used by 1 other objects
create materialized view mv as select add(a, b) as c from t;
```

proto/ddl_service.proto
```diff
@@ -86,6 +86,8 @@ message CreateSinkRequest {
   stream_plan.StreamFragmentGraph fragment_graph = 2;
   // It is used to provide a replace plan for the downstream table in `create sink into table` requests.
   optional ReplaceTablePlan affected_table_change = 3;
+  // The list of object IDs that this materialized view depends on.
```
BugenZhao, line 89:
```suggestion
  // The list of object IDs that this sink depends on.
```

proto/ddl_service.proto
```diff
@@ -86,6 +86,8 @@ message CreateSinkRequest {
...
+  repeated uint32 dependencies = 4;
```
BugenZhao, line 90: Is it accurately _direct_ dependencies? I suppose the meta service can recursively resolve all dependencies.

src/connector/src/sink/catalog/desc.rs
```diff
@@ -104,7 +103,7 @@ impl SinkDesc {
             downstream_pk: self.downstream_pk,
             distribution_key: self.distribution_key,
             owner,
-            dependent_relations,
+            dependent_relations: vec![], // TODO(rc): to be deprecated
```
BugenZhao, line 106: Are we going to deprecate all `dependent_relations` fields in different catalogs? Are they referenced anywhere?

src/frontend/src/handler/create_mv.rs
```diff
@@ -232,19 +220,23 @@ It only indicates the physical clustering of the data, which may improve the per
...
-            dependent_relations,
-            name,
-            columns,
-            emit_mode,
-        )?;
+        let (plan, table) =
+            gen_create_mv_plan_bound(&session, context.into(), query, name, columns, emit_mode)?;
+
+        let dependencies =
+            RelationCollectorVisitor::collect_with(dependent_relations, plan.clone())
```
BugenZhao, line 227: So for UDF, we record as long as we encounter one in the binder. But for relations, we record by visiting the _optimized_ plan. Is it possible to make this behavior consistent? Personally I think we should handle it in binder (i.e., regardless of optimization), as it can be confusing if a dropping a dependent relation is accepted even if it may invalidate the definition of the downstream job.

src/connector/src/sink/catalog/desc.rs
stdrc: I don't think these `dependent_relations` fields are used elsewhere except when creating meta objects. cc @yezizp2012 can you confirm?

yezizp2012: IIUC that's the only usage when creating meta objects. Previouly the process of detecting dependency cycles for `sink into table` relies on it in v1, but that functionality has been transferred to the meta side. However, the code related to this in the FE remains and can be removed for clarity.

src/frontend/src/handler/create_mv.rs
stdrc: > Personally I think we should handle it in binder (i.e., regardless of optimization)

Yes, I agree. I didn't change it because I want to mainly focus on UDF in this PR.

src/meta/src/controller/streaming_job.rs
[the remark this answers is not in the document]
yezizp2012: IIUC it won't break the functionality of `list_object_dependencies`. It only returns dependency info about streaming job, view and subscription.

src/connector/src/sink/catalog/desc.rs
yezizp2012: btw, @shanicky will rise a PR to clean that part.

src/frontend/src/handler/create_mv.rs
stdrc: added a todo comment

src/meta/service/src/ddl_service.rs
```diff
@@ -241,6 +242,7 @@ impl DdlService for DdlServiceImpl {
                         fragment_graph,
                         CreateType::Foreground,
                         None,
+                        HashSet::new(), // TODO(rc): pass dependencies through this field instead of `PbSource`
```
yezizp2012, line 245: 👍 connection and secret are handled in separated fields now and can be unified into `dependencies`. This also applies to the connection for the sink.

yezizp2012 (approved):
LGTM! Thx for the PR. Please help to unify the behaviors for other relations like `index`, `source` as well. 
Additionally, following this PR, we can provide syntax support for granting and revoking function privileges to enhance user experience.

PR #19402, opened 2024-11-15
refactor(frontend): rework `UPDATE` to support subqueries

### LINKED ISSUE #2672: Support subquery and `FROM` in `UPDATE`
- Unnesting.
  ```sql
  UPDATE employees SET sales_count = sales_count + 1 WHERE id =
    (SELECT sales_person FROM accounts WHERE name = 'Acme Corporation');
  ```

- Subquery on the right side of multi-assignment.
  ```sql
  -- Update contact names in an accounts table to match the currently assigned salesmen:
  UPDATE accounts SET (contact_first_name, contact_last_name) =
      (SELECT first_name, last_name FROM salesmen
       WHERE salesmen.id = accounts.sales_id);
  ```

- `UPDATE ... SET ... FROM`.
  ```sql
  -- A similar result could be accomplished with a join:
  UPDATE accounts SET contact_first_name = first_name,
                      contact_last_name = last_name
    FROM salesmen WHERE salesmen.id = accounts.sales_id;
  ```

Related PR: #2602

BugenZhao:

Rework `UPDATE` to support subqueries. Close #2672.

----

Previously, `Update` has a list of `index -> expr` to calculate the updated value for each column. This means that its capability cannot exceed a single `Project` node, so that there's no way to support subquery assignments like `UPDATE (v1, v2) = (select 114, 514);`.

After this PR, the input of an `Update` will calculate the old and new values for all columns, so that `Update` only needs to pick some columns with `InputRef` to generate the `U-` and `U+` chunks. Complex assignments like subqueries will be handled by the optimizer, just like within a normal `SELECT` statement.

To support this, a new subquery kind named `UpdateSet` is added to allow returning multiple columns (still single row) by compositing them into a `struct` type.

An example demonstrating the plan:

```sql
dev=> explain update t set (v1, v2) = (select v, 666 from generate_series(1, 1) as g(v));
                                          QUERY PLAN                                          
----------------------------------------------------------------------------------------------
 └─BatchUpdate { table: t, exprs: [Field($4, 0:Int32), Field($4, 1:Int32), $2] }
   └─BatchNestedLoopJoin { type: LeftOuter, predicate: true }
     ├─BatchExchange { order: [], dist: Single }
     │ └─BatchScan { table: t, columns: [v1, v2, _row_id, _rw_timestamp] }
     └─BatchMaxOneRow
       └─BatchProject { exprs: [Row(GenerateSeries(1:Int32, 1:Int32), 666:Int32) as $expr1] }
         └─BatchProjectSet { select_list: [GenerateSeries(1:Int32, 1:Int32)] }
           └─BatchValues { rows: [[]] }
(9 rows)
```

## Checklist

## Documentation

## Release note

</details>


chenzl25 (commented):
A corner case
```
create table t (a int, b int);
create table y (a int, b int);
insert into y values(11,11),(22,22);
insert into t values(1,1),(2,2);

update t set (a,b) = (select y.a, y.b from y);
-- PG will throw an error ERROR:  more than one row returned by a subquery used as an expression
-- but RW will succeed to execute and return UPDATE 4 rows
```

BugenZhao:
> A corner case

In this case, the query `select 1, (select y.a from y)` should also be rejected, but we accepted it. So I presume there's something wrong with `BatchMaxOneRow`. 🤔

chenzl25:
> I think it's because we don't enforce input of `BatchMaxOneRow` to be singleton

Oh, that cloud be the problem

BugenZhao:
> > I think it's because we don't enforce input of `BatchMaxOneRow` to be singleton
> 
> Oh, that cloud be the problem

Will fix in https://github.com/risingwavelabs/risingwave/pull/19452

chenzl25 (approved):
LGTM! The idea to use a new kind (UpdateSet) of subquery is brilliant. Please make sure the previous corner case is fixed and add it to a testcase.

PR #19425, opened 2024-11-18
feat(frontend): Add struct scan and simply optimise struct filter.

xxhZs:

(v1,v2) > (1,2) means v1 > 1 or (v1 =1  and v2 >2), so it eq all data from (1,2) 

in this pr, we add optimise for struct filter 
for example :
(v1,v2,v3) > (1,2,3) if v1 ,v2, v3 all pk, we can push it down table scan((v1,v2,v3) > (1,2,3));
if v1, v2 is pk , v3 is not pk , we can push v1 v2 down table scan and add filter , the plan like : 
filter(v1,v2,v3) > (1,2,3)
 scan : (v1,v2) > (1,2);
Although the filter remains unchanged, the number of scans is reduced

## Checklist

## Documentation

## Release note

</details>


src/batch/src/executor/row_seq_scan.rs
```diff
@@ -60,12 +60,15 @@ pub struct RowSeqScanExecutor<S: StateStore> {
...
+#[derive(Debug)]
 pub struct ScanRange {
     /// The prefix of the primary key.
     pub pk_prefix: OwnedRow,
 
     /// The range bounds of the next column.
     pub next_col_bounds: (Bound<Datum>, Bound<Datum>),
+
+    pub is_real_unbounded: bool,
 }
```
chenzl25, line 72: Let's use an Enum to separate these two different scan range.

src/frontend/src/utils/condition.rs
```diff
@@ -382,11 +391,108 @@ impl Condition {
...
+                .unwrap()
+                .inputs()
+                .get(1)
+                .unwrap()
+                .as_literal()
+                .unwrap();
+            let row_right_literal_data = row_right_literal.get_data().clone().unwrap();
+            let row_right_literal_type = row_right_literal.get_data_type().clone().unwrap();
+            let right_iter = row_right_literal_data
+                .as_struct()
```
chenzl25, line 436: I think we should use `is_struct` to ensure the literal data type is struct as well.

src/frontend/src/utils/condition.rs
```diff
@@ -382,11 +391,108 @@ impl Condition {
...
+                    })
+                    .collect::<HashMap<_, _>>();
+                let mut pk_struct = vec![];
+                for i in table_desc.order_column_indices() {
+                    if let Some((_, (data, _))) = row_inner_element_map.remove(&i) {
+                        pk_struct.push(data.clone());
+                    } else {
+                        break;
+                    }
+                }
```
chenzl25, line 462: Considering the primary key is (a, b, c), if users write a predicate like `(c, b, a) > (1, 2, 3)`, it seems we will treat it as `(a, b, c) > (3, 2, 1)` here. However, it is incorrect.

src/frontend/src/optimizer/plan_node/utils.rs
```diff
@@ -462,3 +465,124 @@ pub fn to_pb_time_travel_as_of(a: &Option<AsOf>) -> Result<Option<PbAsOf>> {
...
+            }
+        }
+    }
+    if scan_ranges.len() > explain_max_range {
+        range_strs.push("...".to_string());
+    }
+    range_strs
+}
+
+pub fn ub_struct_to_string(
```
chenzl25, line 513: I think we should consider change `struct` to `row`, since we use `RowScanRange` instead of `StructScanRange`

chenzl25 (commented):
Could you please add some planner tests?

src/frontend/src/utils/condition.rs
```diff
@@ -357,6 +356,104 @@ impl Condition {
         }
     }
 
+    fn split_struct_to_scan_ranges(
```
chenzl25, line 359: `split_row_cmp_to_scan_ranges` seems better

src/frontend/src/utils/condition.rs
```diff
@@ -432,7 +534,14 @@ impl Condition {
                     scan_range.eq_conds.extend(eq_conds.into_iter());
                 }
                 0 => {
-                    scan_range.range = (lower_bound, upper_bound);
+                    // scan_range.range.push((lower_bound, upper_bound));
+                    // scan_range.set_and_scan_range((lower_bound, upper_bound));
```
chenzl25, line 538: Remove the commemts.

chenzl25 (approved):
Rest LGTM. Please add a planner test.

PR #16626, opened 2024-05-08
feat(source): let shared SourceExecutor start from latest instead of specified offset

xxchan:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://gist.github.com/TennyZhuang/f00be7f16996ea48effb049aa7be4d66#file-rw_cla).

## What's changed and what's your intention?
https://github.com/risingwavelabs/risingwave/issues/16576#issuecomment-2095571890

mentioned by Eric, this is also his original idea in his RFC

## Checklist

- [ ] I have written necessary rustdoc comments
- [ ] I have added necessary unit tests and integration tests
- [ ] I have added test labels as necessary. See [details](https://github.com/risingwavelabs/risingwave/blob/main/docs/developer-guide.md#ci-labels-guide).
- [ ] I have added fuzzing tests or opened an issue to track them. (Optional, recommended for new SQL features #7934).
- [ ] My PR contains breaking changes. (If it deprecates some features, please create a tracking issue to remove them in the future).
- [ ] All checks passed in `./risedev check` (or alias, `./risedev c`)
- [ ] My PR changes performance-critical code. (Please run macro/micro-benchmarks and show the results.)

- [ ] My PR contains critical fixes that are necessary to be merged into the latest release. (Please check out the [details](https://github.com/risingwavelabs/risingwave/blob/main/CONTRIBUTING.md))

## Documentation

- [ ] My PR needs documentation updates. (Please use the **Release note** section below to summarize the impact on users)

## Release note

If this PR includes changes that directly affect users or other significant modifications relevant to the community, kindly draft a release note to provide a concise summary of these changes. Please prioritize highlighting the impact these changes will have on users.

</details>

PR #16348, opened 2024-04-16
feat: pause shared SourceExecutor until a downstream actor is created

xxchan:
Signed-off-by: xxchan <xxchan22f@gmail.com>I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://gist.github.com/TennyZhuang/f00be7f16996ea48effb049aa7be4d66#file-rw_cla).

## What's changed and what's your intention?

As title. Without this optimization, shared source will waste a lot work at the beginning. 
This applies to shared Kafka source, and also shared CDC source.

- The implementation of testing whether there are downstreams is similar to MView's optimization for conflict check
  - When a SourceExecutor is created, check `actor_context.dispatch_num` to pause.
  - Wait for barrier to resume the stream.
- A behavior change: now `rw_enable_shared_source` only applies to `CREATE SOURCE`, but not `CREATE MV`
- Also added some additional logic tests for shared source.

## Checklist

- [ ] I have written necessary rustdoc comments
- [ ] I have added necessary unit tests and integration tests
- [ ] I have added test labels as necessary. See [details](https://github.com/risingwavelabs/risingwave/blob/main/docs/developer-guide.md#ci-labels-guide).
- [ ] I have added fuzzing tests or opened an issue to track them. (Optional, recommended for new SQL features #7934).
- [ ] My PR contains breaking changes. (If it deprecates some features, please create a tracking issue to remove them in the future).
- [ ] All checks passed in `./risedev check` (or alias, `./risedev c`)
- [ ] My PR changes performance-critical code. (Please run macro/micro-benchmarks and show the results.)

- [ ] My PR contains critical fixes that are necessary to be merged into the latest release. (Please check out the [details](https://github.com/risingwavelabs/risingwave/blob/main/CONTRIBUTING.md))

## Documentation

- [ ] My PR needs documentation updates. (Please use the **Release note** section below to summarize the impact on users)

## Release note

If this PR includes changes that directly affect users or other significant modifications relevant to the community, kindly draft a release note to provide a concise summary of these changes. Please prioritize highlighting the impact these changes will have on users.

</details>

PR #19443, opened 2024-11-19
fix: fix potential data loss for shared source

xxchan:

Set up: Create a shared kafka source, and 1 MV on the source.

Data loss happens when:
1. alter source set rate limit to 0
2. push some new data to the kafka topic
3. resume the source. Then we can find the MV don't get the data. If we push more data to the topic, only new data comes.

Reason:
In #16626 we introduced an optimization to let shared SourceExecutor start from latest, but the implementation is problematic. Specifically, `hack_seek_to_latest` will not only take effect at the beginning, but will also when rebuilding the source reader (which happens when rate limit is applied).

The new implementation in this PR:
- Remove implicit `hack_seek_to_latest` flag, which is error prone. 
- Replaced with an explicit `seek_to_latest` call. At the same time, we also get the latest offsets. To make sure `SplitImpl` and `SourceReader` is consistent.
- Make sure the splits are written to state store even if no messages come.
- Do not pause the SourceExecutor. (#16348) It seems not very useful, but just confusion. Seeking to latest should be good enough.

## Checklist

## Documentation

## Release note

</details>

### DIFF
--- a/src/connector/src/source/reader/reader.rs
+++ b/src/connector/src/source/reader/reader.rs
@@ -37,8 +37,8 @@ use crate::source::filesystem::opendal_source::{
 use crate::source::filesystem::{FsPageItem, OpendalFsSplit};
 use crate::source::{
     create_split_reader, BackfillInfo, BoxChunkSourceStream, BoxTryStream, Column,
-    ConnectorProperties, ConnectorState, SourceColumnDesc, SourceContext, SplitId, SplitReader,
-    WaitCheckpointTask,
+    ConnectorProperties, ConnectorState, SourceColumnDesc, SourceContext, SplitId, SplitImpl,
+    SplitReader, WaitCheckpointTask,
 };
 use crate::{dispatch_source_prop, WithOptionsSecResolved};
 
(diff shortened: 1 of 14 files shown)


xxchan:
* **#19443** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/19443?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/19443?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

tabVersion:
> Specifically, hack_seek_to_latest will not only take effect at the beginning, but will also when rebuilding the source reader (which happens when rate limit is applied).

So when receiving a new mutation on rate_limit, the source exec refreshes the high watermark to `hw_1` but the source backfill exec keeps the original high watermark `hw_0` as the end position of backfill. Is my understanding correct?

xxchan:
It's not related with backfill's position. We may assume backfill already finished, and it's just forwarding messages now.

Rebuilding will make source exec jump from `hw_0` to `hw_1`.

tabVersion (approved):

xxchan:
Want to wait a while for more reviews. Just in case.

src/stream/src/from_proto/source/trad_source.rs
```diff
@@ -232,7 +232,7 @@ impl ExecutorBuilder for SourceExecutorBuilder {
                         barrier_receiver,
                         system_params,
                         source.rate_limit,
-                        is_shared,
+                        is_shared && !source.with_properties.is_cdc_connector(),
```
BugenZhao, line 235: Why change this?

src/connector/src/source/reader/reader.rs
```diff
@@ -211,14 +211,17 @@ impl SourceReader {
     }
 
     /// Build `SplitReader`s and then `BoxChunkSourceStream` from the given `ConnectorState` (`SplitImpl`s).
+    ///
+    /// If `seek_to_latest` is true, will also return the latest splits after seek.
```
BugenZhao, line 215: What if we always return the splits?

BugenZhao (approved):

xxchan:
### Merge activity

* **Nov 20, 3:57 PM GMT+8**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/risingwavelabs/risingwave/19443).

xxchan:
will cherry pick the whole stack together

PR #19449, opened 2024-11-19
refactor(optimizer): record error contexts when casting composite types

BugenZhao:

Record contexts for error when recursively calling `cast` on composite types, to provide better error messages.

Preview:

https://github.com/risingwavelabs/risingwave/blob/fc4223348295f029a4d88d0d9f64b4de8a760ac8/src/frontend/planner_test/tests/testdata/output/cast.yaml#L92-L115

## Checklist

## Documentation

## Release note

</details>

### DIFF
--- a/src/frontend/planner_test/tests/testdata/output/cast.yaml
+++ b/src/frontend/planner_test/tests/testdata/output/cast.yaml
@@ -87,7 +87,7 @@
     Failed to bind expression: CAST(CAST('[]' AS INT[]) AS BYTEA[])
 
     Caused by these errors (recent errors listed first):
-      1: cannot cast type "integer[]" to "bytea[]" in Explicit context
+      1: cannot cast type "integer[]" to "bytea[]"
       2: cannot cast type "integer" to "bytea" in Explicit context
 - name: composite type cast error message (struct)
   sql: |
@@ -97,9 +97,11 @@
     Failed to bind expression: CAST(v AS STRUCT<d STRUCT<e BYTEA>, f BOOLEAN>)
 
     Caused by these errors (recent errors listed first):
-      1: cannot cast type "struct<a struct<b integer>, c boolean>" to "struct<d struct<e bytea>, f boolean>" in Explicit context
-      2: cannot cast type "struct<b integer>" to "struct<e bytea>" in Explicit context
-      3: cannot cast type "integer" to "bytea" in Explicit context
+      1: cannot cast type "struct<a struct<b integer>, c boolean>" to "struct<d struct<e bytea>, f boolean>"
+      2: cannot cast struct field "a" to struct field "d"
+      3: cannot cast type "struct<b integer>" to "struct<e bytea>"
+      4: cannot cast struct field "b" to struct field "e"
+      5: cannot cast type "integer" to "bytea" in Explicit context
 - name: composite type cast error message (map)
   sql: |
     create table t (v map(int, int));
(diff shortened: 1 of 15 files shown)


src/frontend/planner_test/tests/testdata/output/cast.yaml
```diff
@@ -80,3 +80,35 @@
...
+    create table t (v map(int, int));
+    select v::map(int, bytea) from t;
+  binder_error: |
+    Failed to bind expression: CAST(v AS MAP(INT,BYTEA))
+
+    Caused by these errors (recent errors listed first):
+      1: cannot cast type "map(integer,integer)" to "map(integer,bytea)" in Explicit context
+      2: cannot cast type "struct<key integer, value integer>[]" to "struct<key integer, value bytea>[]" in Explicit context
+      3: cannot cast type "struct<key integer, value integer>" to "struct<key integer, value bytea>" in Explicit context
+      4: cannot cast type "integer" to "bytea" in Explicit context
```
xxchan, line 114: This might be a little confusing to users 🤪

BugenZhao: Updated. Check check again.

src/frontend/src/expr/type_inference/cast.rs
```diff
@@ -204,10 +275,10 @@ pub enum CastContext {
     Explicit,
 }
 
-pub type CastMap = BTreeMap<(DataTypeName, DataTypeName), CastContext>;
+pub type CastTable = BTreeMap<(DataTypeName, DataTypeName), CastContext>;
```
BugenZhao, line 278: Not to be confused with `cast_map`.

src/frontend/src/expr/type_inference/cast.rs
```diff
@@ -114,76 +116,145 @@ pub fn align_array_and_element(
     Ok(array_type)
 }
 
+/// Returns `Ok` if `ok` is true, otherwise returns a placeholder [`CastError`] to be further
+/// wrapped with a more informative context in [`cast`].
+fn canmeh(ok: bool) -> CastResult {
+    if ok {
+        Ok(())
+    } else {
+        bail_cast_error!()
```
xxchan, line 125: A little confused how it works. So this is an empty leaf CastError? Why it's not shown in the err msg stack? (Why we need it?)

src/frontend/src/expr/type_inference/cast.rs
```diff
@@ -114,76 +116,145 @@ pub fn align_array_and_element(
     Ok(array_type)
 }
 
+/// Returns `Ok` if `ok` is true, otherwise returns a placeholder [`CastError`] to be further
+/// wrapped with a more informative context in [`cast`].
+fn canmeh(ok: bool) -> CastResult {
```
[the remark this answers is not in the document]
BugenZhao: Sure, renamed.

xxchan (approved):
very good 👍

src/frontend/src/expr/type_inference/cast.rs
BugenZhao: Yes. In error report we'll disregard all empty and repeated entries, so it won't appear.

To preserve the reason for a cast failure, we must return a `Result` in functions like `cast_struct`. However, the default error message "cannot cast" is uniformly applied in `cast`, so we simply return an empty leaf error and let the caller to fill the context.

xxchan: It sounds like `Result<(), Option<CastError>>` 🤪

PR #19458, opened 2024-11-19
feat(connector): add session token for s3 connector

wcy-fdu:

Previously, when using the S3 connector, users were required to configure ak sk, which was fixed. This pr adds the session token field to support temporary ak sk created by users. Both s3 source and s3 sink is added.

## Checklist

## Documentation

## Release note

</details>

### DIFF
--- a/src/connector/src/parser/parquet_parser.rs
+++ b/src/connector/src/parser/parquet_parser.rs
@@ -223,7 +223,7 @@ pub async fn get_total_row_nums_for_parquet_file(
         }
         ConnectorProperties::OpendalS3(prop) => {
             let connector: OpendalEnumerator<OpendalS3> =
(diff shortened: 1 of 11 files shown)


wcy-fdu:
Will test locally and then request review.

wcy-fdu:
Think twice, we should allow setting session_token only for batch query, as it's a bit dangerous if streaming file source using a secret key that may have expired, causing some unpredictable errors.

However, we cannot know when creating a source whether it will be used for batch or streaming:
- if creating materialized view from source, it's long running streaming source
- if only run select from such source, it's batch query

In other words, if the source needs to support session tokens, it cannot be disabled for the streaming source when creating it.

Thus, I suggest when creating a source, only fixed keys that do not expire are allowed. 

We can add session_token for tvf `file_scan()` if needed.

PR #19467, opened 2024-11-20
feat(stream): create cdc table reader and source data stream with retry

### LINKED ISSUE #17807: bug(cdc source): should retry on upstream failure instead of blocking recovery
### Describe the bug

```
{"timestamp":"2024-07-25T01:39:26.827182302Z","level":"WARN","fields":{"message":"build_actors failed","error":"merged RPC Error: worker node 140004, gRPC request to stream service failed: Internal error: Executor error: Connector error: Postgres error: db error: ERROR: relation \"public.source_priorities\" does not exist;"},"target":"risingwave_meta::barrier::recovery","spans":[{"prev_epoch":6857419982438400,"name":"bootstrap_recovery"},{"name":"recovery_attempt"}]}
```
Currently, if the upstream of an PG CDC table doesn't work, the recovery will fail, causing the entire cluster unavailable.

The expected behavior should be keep retrying.

### Error message/log

_No response_

### To Reproduce

_No response_

### Expected behavior

_No response_

_No response_

### The version of RisingWave

_No response_

### Additional context

_No response_

StrikeW:

Previously we try to create the table reader in `src/stream/src/from_proto/stream_cdc_scan.rs`, if upstream database is down, the executor cannot be built so that the cluster will enter recovery loop. This pr try to create the table reader after we forward the initial barrier, so that it won't block recovery if cannot connect to upstream db.

We also add retry logic to the source executor when we create the data stream at first time.

close #17807 

## Checklist

## Documentation

## Release note

</details>


src/stream/src/executor/backfill/cdc/cdc_backfill.rs
```diff
@@ -168,29 +175,107 @@ impl<S: StateStore> CdcBackfillExecutor<S> {
...
+                }
+                Err(e) => {
+                    tracing::warn!(
+                        table_id,
+                        upstream_table_name,
+                        error = %e.as_report(),
+                        "failed to create external table reader",
+                    );
+                    if let Some(delay) = backoff.next() {
+                        sleep(delay).await;
```
hzxa21, line 239: This backoff will also cause delay in polling the upstream buffer so there is still side effect here when the upstream DB is unavailable.

src/stream/src/executor/backfill/cdc/cdc_backfill.rs
```diff
@@ -168,29 +175,107 @@ impl<S: StateStore> CdcBackfillExecutor<S> {
...
+        loop {
+            let create_result: StreamExecutorResult<_> = try {
+                let reader = upstream_buffer
+                    .run_future(create_table_reader(&self.external_table))
+                    .await?;
+                reader
+            };
+
+            // emit upstream message if any
+            if let Some(msg) = upstream_buffer.next_pending_message().await? {
```
hzxa21, line 205: Polling one message at a time may hurt barrier latency combining with the backoff above because barrier may not be polled in time.

src/stream/src/executor/backfill/cdc/cdc_backfill.rs
```diff
@@ -702,6 +787,103 @@ impl<S: StateStore> CdcBackfillExecutor<S> {
     }
 }
 
+struct UpstreamMessageBuffer {
```
hzxa21, line 790: The implementation seems to be inspired by snapshot backfill, which is an overkill and more complicated than I thought.  

Is there any ordering requirement between table reader creation and polling upstream? In other words, can we make sure the offset we consume from upstream is accurate if we blindly poll upstream when the table reader is still under creation. If we can blindly poll upstream, how about just spawning a task to create the table reader with retry? Pseudo code example:
```rust
async fn create_table_reader_with_retry(...) -> TableReader  {
  ...
}

-- execute_inner
let handle = tokio::spawn(create_table_reader_with_retry(...));
loop {
  select! {
      biased;
      reader = handle => {
        ...
        break;
      }
      msg = upstream.next() => {
          // Emit barrier and skip chunk
     }
  }
```

src/stream/src/executor/backfill/cdc/cdc_backfill.rs
```diff
@@ -168,29 +172,75 @@ impl<S: StateStore> CdcBackfillExecutor<S> {
...
         // Keep track of rows from the snapshot.
         let mut total_snapshot_row_count = state.row_count as u64;
 
+        // After init the state table and forward the initial barrier to downstream,
+        // we now try to create the table reader with retry.
+        // If backfill hasn't finished, we can ignore upstream cdc events before we create the table reader;
+        // If backfill is finished, we should forward the upstream cdc events to downstream.
+        let mut table_reader: Option<ExternalTableReaderImpl> = None;
+        let external_table = self.external_table.clone();
+        let mut handle = Box::pin(tokio::spawn(async move {
```
BugenZhao, line 192: Seems there's no need to `spawn` a task for creating the table reader? We just need to `select` between upstream and this.

src/stream/src/executor/backfill/cdc/cdc_backfill.rs
```diff
@@ -168,29 +172,75 @@ impl<S: StateStore> CdcBackfillExecutor<S> {
...
+            let backoff = get_backoff_strategy();
+            tokio_retry::Retry::spawn(backoff, || async {
+                external_table.create_table_reader().await
+            })
+            .instrument(tracing::info_span!("create_cdc_table_reader_with_retry"))
+            .await
+            .expect("Retry create cdc table reader until success.")
```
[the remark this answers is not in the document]
BugenZhao: This is actually unreachable.

src/stream/src/executor/backfill/cdc/cdc_backfill.rs
StrikeW: fixed

src/stream/src/executor/backfill/cdc/upstream_table/external.rs
```diff
@@ -101,4 +118,13 @@ impl ExternalStorageTable {
     pub fn database_name(&self) -> &str {
         self.database_name.as_str()
     }
+
+    pub async fn current_cdc_offset(
+        &self,
+        table_reader: &ExternalTableReaderImpl,
+    ) -> ConnectorResult<Option<CdcOffset>> {
+        let binlog = table_reader.current_cdc_offset();
+        let binlog = binlog.await?;
```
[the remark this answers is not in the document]
StrikeW: fixed

src/stream/src/executor/backfill/cdc/upstream_table/external.rs
```diff
@@ -28,8 +33,11 @@ pub struct ExternalStorageTable {
 
     database_name: String,
 
-    table_reader: ExternalTableReaderImpl,
+    config: ExternalTableConfig,
 
+    table_type: CdcTableType,
+
+    // table_reader: ExternalTableReaderImpl,
```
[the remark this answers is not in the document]
StrikeW: fixed

BugenZhao:
As per discussion in https://github.com/risingwavelabs/risingwave/issues/17807#issuecomment-2249576983, may I know if this still holds?

> When creating the physical executor of a new source, we recommend throwing errors to users immediately when creating actors and calling from_proto.

StrikeW:
> As per discussion in [#17807 (comment)](https://github.com/risingwavelabs/risingwave/issues/17807#issuecomment-2249576983), may I know if this still holds?
> 
> > When creating the physical executor of a new source, we recommend throwing errors to users immediately when creating actors and calling from_proto.

This pr doesn't modify the code of `Source`, so the behavior is same as before. For the cdc backfill executor, previously if we fail to create the table reader, the stream job will fail to create.

StrikeW:
> Can we add tests to verify this PR is working and prevent future regression?
> 
> 
> 
> I tried the following two cases with this PR:
> 
> 1. Start RW/MySQL -> Create cdc table -> shutdown MySQL
> 
> 2. Start RW/MySQL -> Create cdc table -> shutdown RW -> shutdown MySQL -> Start RW
> 
> 
> 
> 
> 
> 1 works but 2 doesn't. The cluster is in recovery loop due to
> 
> ```
> 
> error=get error from control stream, in worker node 1: gRPC request to stream service failed: Internal error: recv actor failure: Actor 1 exited unexpectedly: Executor error: Connector error: Connector configuration is not valid. Unable to connect: Communications link failure
> 
> ```
> 
> 
> 
> I guess it is related to source validation before creating the source executor.


I have tested again and cannot reproduce this issue. I created a source and table, then shutdown rw and postgres. Restart rw and then create a mv based on the table, everything is ok. Finally I restart postgres, the cdc source can be resumed.

StrikeW:
> > Can we add tests to verify this PR is working and prevent future regression?
> > I tried the following two cases with this PR:
> > 
> > 1. Start RW/MySQL -> Create cdc table -> shutdown MySQL
> > 2. Start RW/MySQL -> Create cdc table -> shutdown RW -> shutdown MySQL -> Start RW
> > 
> > 1 works but 2 doesn't. The cluster is in recovery loop due to
> > ```
> > 
> > error=get error from control stream, in worker node 1: gRPC request to stream service failed: Internal error: recv actor failure: Actor 1 exited unexpectedly: Executor error: Connector error: Connector configuration is not valid. Unable to connect: Communications link failure
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I guess it is related to source validation before creating the source executor.
> 
> I have tested again and cannot reproduce this issue. I created a source and table, then shutdown rw and postgres. Restart rw and then create a mv based on the table, everything is ok. Finally I restart postgres, the cdc source can be resumed.

The mysql source can reproduce the issue: `Executor error: Connector error: Connector configuration is not valid. Unable to connect: Communications link failure`, I am investigating.

src/stream/src/executor/backfill/cdc/cdc_backfill.rs
```diff
@@ -168,29 +171,75 @@ impl<S: StateStore> CdcBackfillExecutor<S> {
...
+        // After init the state table and forward the initial barrier to downstream,
+        // we now try to create the table reader with retry.
+        // If backfill hasn't finished, we can ignore upstream cdc events before we create the table reader;
+        // If backfill is finished, we should forward the upstream cdc events to downstream.
+        let mut table_reader: Option<ExternalTableReaderImpl> = None;
+        let external_table = self.external_table.clone();
+        let mut future = Box::pin(async move {
+            let backoff = get_backoff_strategy();
+            tokio_retry::Retry::spawn(backoff, || async {
+                external_table.create_table_reader().await
```
fuyufjh, line 194: Shall we print some error logs when `create_table_reader()` fails?

StrikeW: I added a warn log

src/stream/src/executor/source/source_executor.rs
```diff
@@ -490,19 +505,61 @@ impl<S: StateStore> SourceExecutor<S> {
...
+                }
+            })
+            .instrument(tracing::info_span!("build_source_stream_with_retry"))
+            .await
+            .expect("Retry build source stream until success.")
+        });
+
+        // loop to create source stream until success
+        loop {
+            if let Some(barrier) = build_source_stream_and_poll_barrier(
```
StrikeW, line 540: @xxchan @tabVersion PTAL this part for source executor

hzxa21 (approved):
Tested locally and it works. LGTM!

src/stream/src/executor/source/source_executor.rs
hzxa21: The changes here apply to all sources. Does it mean that prior to this PR when upstream system is not available, the cluster will fail to start up?

xxchan: I guess it's `Barrier::Resume` not handled correctly for replace table.

src/stream/src/executor/source/source_executor.rs
```diff
@@ -490,19 +505,72 @@ impl<S: StateStore> SourceExecutor<S> {
...
+            })
+            .instrument(tracing::info_span!("build_source_stream_with_retry"))
+            .await
+            .expect("Retry build source stream until success.")
+        });
+
+        let mut need_resume_after_build = false;
+        // loop to create source stream until success
+        loop {
+            if let Some(barrier) = build_source_stream_and_poll_barrier(
```
fuyufjh, line 541: FYI. Link a [Slack conversation](https://risingwave-labs.slack.com/archives/C04S2HVU99V/p1732778516091719?thread_ts=1732640723.960199&cid=C04S2HVU99V) here.


> @tabVersion: It also affects the process of creating table, auto retry here means the create mv / create table command will always succeed, even if meeting some runtime error, such as Kafka group auth issue. It prevents rw from consuming data and requires the user setting up the ACL. We can only find this issue when building an actor. 

So does this issue get noticed here? What happen if I create a source with some bad settings (such the permission issue mentioned above)?

fuyufjh (approved):
Overall LGTM!

src/stream/src/executor/source/source_executor.rs
StrikeW: > It prevents rw from consuming data and requires the user setting up the ACL. We can only find this issue when building an actor.

I think MQ source enumerator should employ a validation logic to check these prerequisites, so that prerequisite for connectivity should be no problem when we are creating executors. But I am not sure whether we have the check for MQ sources. cc @tabVersion

fuyufjh: +1 for relying on the validation logic. 

(Just saying, if someone is too lazy to write a full set of validation, they may just build a real `Consumer` and see if anything goes well.)

src/stream/src/executor/source/source_executor.rs
```diff
@@ -490,20 +505,98 @@ impl<S: StateStore> SourceExecutor<S> {
...
+            })
+                .instrument(tracing::info_span!("build_source_stream_with_retry"))
+                .await
+                .expect("Retry build source stream until success.")
+        });
+
+        let mut need_resume_after_build = false;
+        // loop to create source stream until success
+        loop {
+            if let Some(barrier) = build_source_stream_and_poll_barrier(
```
xxchan, line 541: nit: `while let` might be better than `loop { if let`

src/stream/src/executor/source/source_executor.rs
```diff
@@ -736,6 +828,29 @@ impl<S: StateStore> SourceExecutor<S> {
     }
 }
 
+async fn build_source_stream_and_poll_barrier(
```
xxchan, line 831: nit: maybe this should be called
```suggestion
async fn build_source_stream_or_receive_barrier(
```

xxchan (approved):
SourceExecutor generally LGTM. It would be better to have some e2e tests though.

PR #19474, opened 2024-11-20
refactor(meta): proactively GC more stale objects.

zwang28:

Previouly all SST objects are retained for at least `min_sst_retention_time_sec` even they're indeed useless.
This PR proactively GC these objects to reduce storage overhead.

## Checklist

## Documentation

## Release note

</details>


Li0k (approved):

src/meta/src/hummock/manager/gc.rs
```diff
@@ -501,6 +526,30 @@ impl HummockManager {
...
+            .list_pinned_ssts()
+            .into_iter()
+            .chain(self.all_object_ids_in_time_travel().await?)
+            .collect();
+        let object_ids = object_ids
+            .into_iter()
+            .filter(|s| !pinned_objects.contains(s))
+            .collect_vec();
+        // Retry is not necessary. Full GC will handle these objects eventually.
+        self.delete_objects(object_ids).await?;
```
hzxa21, line 550: Wil lit be safer to call `complete_gc_batch` directly here?

src/meta/src/hummock/manager/gc.rs
```diff
@@ -48,6 +48,8 @@ pub(crate) struct GcManager {
     store: ObjectStoreRef,
     path_prefix: String,
     use_new_object_prefix_strategy: bool,
+    /// These objects may still be used by backup or time travel.
+    may_delete_object_ids: parking_lot::Mutex<HashSet<HummockSstableObjectId>>,
```
hzxa21, line 52: Given that the minor gc is triggered directly after `create_version_checkpoint` and we can always get the full stale objects from `create_version_checkpoint` directly. Will it be better if we instead let `create_version_checkpoint` returns the stale objects without introducing the mutext protected hash set?

zwang28: may_delete_object_ids may also be modified when proactively GC time travel objects. See latest commits.

zwang28: Will refactor.

Issue #19470, opened 2024-11-20
Bug(Sink): Iceberg Sink picks up hidden columns 


Shreyas220:
### Describe the bug

When i try to create an iceberg sink i get a schema mismatch issue 
can see this 

```
[ColumnCatalog { column_desc: Some(ColumnDesc { column_type: INT16, column_id: 3, name: "kv_log_store_row_op", version: 1 }), is_hidden: false },
```

### Error message/log

```text
CREATE SINK iceberg_sink_queries AS 
SELECT 
    alias,
    query_id,
    query_hash,
    start_time,
    cluster_name,
    query,
    workspace,
    cluster_uuid,
    end_time FROM consumer
WITH (
    connector = 'iceberg',
    type = 'append-only',
    force_append_only = 'true',
    s3.endpoint = 'http://minio:9000',
    s3.access.key = 'admin',
    s3.secret.key = 'password',
    s3.region = 'us-east-1',
    catalog.type = 'rest',
    catalog.name = 'demo',
    catalog.uri = 'http://rest:8181',
    warehouse.path = 's3://warehouse/',
    database.name = 'shreyas',      -- Iceberg database name
    table.name = 'queries'          -- Iceberg table name
);

ERROR:  Failed to run the query

Caused by these errors (recent errors listed first):
  1: gRPC request to meta service failed: Internal error
  2: get error from control stream, in worker node 6
  3: gRPC request to stream service failed: Internal error
  4: recv actor failure
  5: Actor 157 exited unexpectedly
  6: Executor error
  7: exchange channel to downstream actor 145 closed unexpectedly


```

In compute Node 

```
[...]

PR #19485, opened 2024-11-20
fix(iceberg): fix create partition iceberg table sink with log store

chenzl25:

- Resolve https://github.com/risingwavelabs/risingwave/issues/19470
- Since the log store table doesn't contain the extra partition column of the sink, we need to remove this column from the schema before checking.

## Checklist

## Documentation

## Release note

</details>

### DIFF
--- a/src/stream/src/from_proto/sink.rs
+++ b/src/stream/src/from_proto/sink.rs
@@ -263,8 +263,29 @@ impl ExecutorBuilder for SinkExecutorBuilder {
                 );
 
                 let table = node.table.as_ref().unwrap().clone();
-                let input_schema = input_executor.schema();
-                let pk_info = resolve_pk_info(input_schema, &table)?;
+                let schema = if let Some(extra_partition_col_idx) =
+                    sink_write_param.extra_partition_col_idx
+                {
+                    // remove extra partition column from input schema, since log store table won't have it.
+                    let fields = input_executor
+                        .schema()
+                        .fields()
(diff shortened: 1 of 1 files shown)


src/stream/src/from_proto/sink.rs
```diff
@@ -263,8 +263,29 @@ impl ExecutorBuilder for SinkExecutorBuilder {
                 );
 
                 let table = node.table.as_ref().unwrap().clone();
-                let input_schema = input_executor.schema();
-                let pk_info = resolve_pk_info(input_schema, &table)?;
+                let schema = if let Some(extra_partition_col_idx) =
```
wenym1, line 266: Changing the schema may bypass the log store schema validation, but if the input is still in its original schema, we will get error, or write incorrect content to log store. How does this PR fix this? Is the extra partition column already removed from the input?

chenzl25: No, the extra column is still kept in the input. Interesting. There is no error and the log store works normally, we can select the log store internal table as well.

wenym1: Just checked the code. When the schema of row does not match the schema of the serializer, there won't be error when encoding the row and writing to storage. Error only happens when trying to decode the encoded byte slice to row. When the amount of data is less than the log store in memory buffer, rows are only written to storage, but won't be read, and we will read the original stream chunk. So I guess this is why there is no error currently.

PR #19495, opened 2024-11-20
feat(secret): alter secret in catalog

yuhao-su:

This PR introduced alter secret in the catalog. But for secrets used in running jobs, they will not be changed until the executor has been rebuilt (by restarting the cluster)

## Checklist

## Documentation

You can alter a secret value by using `ALTER SECRET your_secret_name WITH (backend = 'meta') AS 'your_new_secret'`.

Note that altering a secret will not result in a change of secret being used in a running job until the job is restarted (e.g. by restarting the cluster). 

## Release note

</details>


src/meta/src/rpc/ddl_controller.rs
```diff
@@ -551,6 +558,16 @@ impl DdlController {
             .await
     }
 
+    async fn alter_secret(&self, mut secret: Secret) -> MetaResult<NotificationVersion> {
+        let secret_plain_payload = secret.value.clone();
+        let encrypted_payload = self.get_encrypted_payload(&secret)?;
+        secret.value = encrypted_payload;
+        self.metadata_manager
+            .catalog_controller
+            .alter_secret(secret, secret_plain_payload)
```
tabVersion, line 567: why we need a plain secret here?

src/sqlparser/src/ast/ddl.rs
```diff
@@ -214,6 +214,12 @@ pub enum AlterConnectionOperation {
     SetSchema { new_schema_name: ObjectName },
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub enum AlterSecretOperation {
+    ChangeCredential { new_credential: Value },
+}
```
tabVersion, line 221: what else operation can alter secret do?

tabVersion (commented):
The impl LGTM, please add an e2e check if we can apply a secret change to running actor by calling `recover`. 

my thoughts about the test

* two kafka topic with 1 partition: `topic_0` and `topic_1`, 
  * fill some data, eg. message in `topic_0` starts with 0 and message in `topic_1` starts with 1. 
* create a table with kafka connector, using secret to spec the topic to `topic_0`
  * check the data from `topic_0`
* pause the cluster, alter the secret pointing to `topic_1`, pushing new data to both topics, then resume the cluster
* check the new data is from `topic_1`

e2e_test/ddl/secret.slt
```diff
@@ -40,6 +40,23 @@ create secret secret_1 with (
   backend = 'meta'
 ) as 'demo_secret';
 
+statement ok
+alter secret secret_1 with (
+  backend = 'meta'
```
fuyufjh, line 45: The `backend` option should not appear in the `alter secret` statement. To follow the convention of other `alter` commands, only the changed part should be provided. Here the `backend` option is obviously untouched, so users should not write it here.

But, furthermore, I think `backend` shouldn't be a per-secret option. In my mind it should be a global-wise config and can't be changed after cluster initialized.

src/common/secret/src/secret_manager.rs
```diff
@@ -74,7 +74,23 @@ impl LocalSecretManager {
...
+                secret_id = secret_id,
+                "adding a secret but it already exists, overwriting it"
+            );
+        };
+    }
+
+    pub fn update_secret(&self, secret_id: SecretId, secret: Vec<u8>) {
+        let mut secret_guard = self.secrets.write();
+        if secret_guard.insert(secret_id, secret).is_none() {
+            tracing::error!(
```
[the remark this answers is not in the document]
fuyufjh: Does it trigger the existing actors to upgrade to the new secret?

fuyufjh (commented):
Generally LGTM

e2e_test/ddl/secret.slt
fuyufjh: > But, furthermore, I think backend shouldn't be a per-secret option. In my mind it should be a global-wise config and can't be changed after cluster initialized.

It's a bit off-topic. We can create a new issue

src/common/secret/src/secret_manager.rs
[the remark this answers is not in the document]
yuhao-su: > Does it trigger the existing actors to upgrade to the new secret?

No. Already doced this limitation in PR.

src/sqlparser/src/ast/ddl.rs
yuhao-su: SET SCHEMA, change WITH option etc..

fuyufjh (approved):

yuhao-su:
> we can apply a secret change to running actor by calling recover

tabVersion:
> > we can apply a secret change to running actor by calling recover
> 
> Added the test in CDC test. A potential problem with your proposal is those 2 topics won't share the offset and it's not a real use case.

Yes, a little hack here. Anyway, it is good to have a test here.

src/common/secret/src/secret_manager.rs
```diff
@@ -74,7 +74,23 @@ impl LocalSecretManager {
 
     pub fn add_secret(&self, secret_id: SecretId, secret: Vec<u8>) {
         let mut secret_guard = self.secrets.write();
-        secret_guard.insert(secret_id, secret);
+        if secret_guard.insert(secret_id, secret).is_some() {
+            tracing::error!(
```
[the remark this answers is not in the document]
tabVersion: It is still a little weird to me. I understand the `LocalSecretManager` cannot reject adding a new secret but we can always do such check on the meta node.

src/meta/src/controller/catalog.rs
```diff
@@ -1397,6 +1397,42 @@ impl CatalogController {
...
+    ) -> MetaResult<NotificationVersion> {
+        let inner = self.inner.write().await;
+        let owner_id = pb_secret.owner as _;
+        let txn = inner.db.begin().await?;
+        ensure_user_id(owner_id, &txn).await?;
+        ensure_object_id(ObjectType::Database, pb_secret.database_id as _, &txn).await?;
+        ensure_object_id(ObjectType::Schema, pb_secret.schema_id as _, &txn).await?;
+
+        let secret: secret::ActiveModel = pb_secret.clone().into();
+        Secret::update(secret).exec(&txn).await?;
```
tabVersion, line 1413: Maybe do a secret existance check here?

st1page (approved):

src/meta/src/controller/catalog.rs
yuhao-su: Already checked in frontend. But I guess it won't here to check again on meta.

PR #19173, opened 2024-10-29
feat(meta): support database checkpoint isolation

wenym1:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://gist.github.com/TennyZhuang/f00be7f16996ea48effb049aa7be4d66#file-rw_cla).

## What's changed and what's your intention?

In this PR, we support checkpoint isolation between different databases.

Previously, in global barrier worker in meta node, we have a struct `CheckpointControl` to manage the barrier state of the whole global streaming graph. To support checkpoint isolation between different databases, the global streaming graph will be divided into per-database streaming graph. The previous `CheckpointControl` is renamed to `DatabaseCheckpointControl`, which tracks the barrier state of the streaming graph of the database. In recovery, we will divide the runtime information for different databases.

Each database will have its own streaming graph that independently injects and collects barrier. The `partial_graph_id` in `StreamingControlRequest` is changed from u32 to u64. Previously, the partial graph id is the u32 job id for creating snapshot backfill jobs, and u32::MAX for the global streaming graph. In this PR, the u64 partial graph id will be composed from database id and job id. The high 32 bits will store the database id, and the low 32 bits will store the job id (for the partial graph of created jobs, it's u32::MAX).

## Checklist

- [ ] I have written necessary rustdoc comments
- [ ] I have added necessary unit tests and integration tests
- [ ] I have added test labels as necessary. See [details](https://github.com/risingwavelabs/risingwave/blob/main/docs/developer-guide.md#ci-labels-guide).
- [ ] I have added fuzzing tests or opened an issue to track them. (Optional, recommended for new SQL features #7934).
- [ ] My PR contains breaking changes. (If it deprecates some features, please create a tracking issue to remove them in the future).
- [x] All checks passed in `./risedev check` (or alias, `./risedev c`)
- [ ] My PR changes performance-critical code. (Please run macro/micro-benchmarks and show the results.)

- [ ] My PR contains critical fixes that are necessary to be merged into the latest release. (Please check out the [details](https://github.com/risingwavelabs/risingwave/blob/main/CONTRIBUTING.md))

## Documentation

- [ ] My PR needs documentation updates. (Please use the **Release note** section below to summarize the impact on users)

## Release note

If this PR includes changes that directly affect users or other significant modifications relevant to the community, kindly draft a release note to provide a concise summary of these changes. Please prioritize highlighting the impact these changes will have on users.

</details>

PR #19556, opened 2024-11-25
refactor(barrier): explicitly maintain database barrier state separately in local barrier manager

wenym1:

To simplify the implementation to support database failure isolation, we should isolate different databases for fields in local barrier worker. Previously, we isolate the barrier injection and collection of different partial graphs. Ideally, it will be great to isolate the fields of local barrier worker by partial graph, so that the local barrier worker won't be aware of the concept of database. 

However, after some efforts, I found it complicated to directly support isolation in different partial graphs. In snapshot backfill, during the stage of backfill, the streaming job has its own partial graph, but after the backfill finishes, its own partial graph will be merged to the global partial graph, along with its actors migrating from the own partial graph to the global partial graph. In this case, partial graphs are not completely isolated between each other. If we change to not merging the partial graph and assume that actors always belong to a fixed partial graph, we can easily support isolation in partial graphs, but the related code in global barrier manager will be unnecessarily complicated. In either way, it will be complicated to isolate different partial graphs.

Therefore, in this PR, we will let local barrier worker be aware of the concept of database. Fields in local barrier manager will be maintained mostly in the way of `HashMap<DatabaseId, T>`. In this way, the logic of database isolation can be easily implemented. Changes are mostly as followed:
* the original `ManagedBarrierState` is renamed to `DatabaseManagedBarrierState`, to store the barrier state per database. In the streaming control bidi-stream, we will specify the `database_id` in each request to specify the database_id to operate on.
* The `LocalBarrierManager`, which holds two channel tx for actors to report barrier event of actor error, will become per-database, so that the barrier event or actor error can be reported independently. It is previously stored in the `SharedContext`. In this PR, we change to store it separately out of `SharedContext`.
* The `SharedContext`, which is used to store pending exchange channels, will also become per-database. This is based on the assumption that there won't be exchange channel between two different databases. In exchange service, to get the exchange channel, we need to specify the current database_id to locate the correct `SharedContext`.

## Checklist

## Documentation

## Release note

</details>

### DIFF
--- a/proto/stream_service.proto
+++ b/proto/stream_service.proto
@@ -12,9 +12,10 @@ option optimize_for = SPEED;
 message InjectBarrierRequest {
   string request_id = 1;
   stream_plan.Barrier barrier = 2;
+  uint32 database_id = 3;
   repeated uint32 actor_ids_to_collect = 4;
   repeated uint32 table_ids_to_sync = 5;
-  uint64 partial_graph_id = 6;
+  uint32 partial_graph_id = 6;
 
   repeated common.ActorInfo broadcast_info = 8;
   repeated stream_plan.StreamActor actors_to_build = 9;
@@ -48,9 +49,10 @@ message BarrierCompleteResponse {
   uint32 worker_id = 5;
   map<uint32, hummock.TableWatermarks> table_watermarks = 6;
   repeated hummock.SstableInfo old_value_sstables = 7;
-  uint64 partial_graph_id = 8;
+  uint32 partial_graph_id = 8;
   // prev_epoch of barrier
   uint64 epoch = 9;
+  uint32 database_id = 10;
 }
 
 message WaitEpochCommitRequest {
@@ -64,20 +66,27 @@ message WaitEpochCommitResponse {
 
 message StreamingControlStreamRequest {
   message InitialPartialGraph {
-    uint64 partial_graph_id = 1;
+    uint32 partial_graph_id = 1;
     repeated stream_plan.SubscriptionUpstreamInfo subscriptions = 2;
   }
 
+  message DatabaseInitialPartialGraph {
+    uint32 database_id = 1;
+    repeated InitialPartialGraph graphs = 2;
+  }
+
   message InitRequest {
-    repeated InitialPartialGraph graphs = 1;
+    repeated DatabaseInitialPartialGraph databases = 1;
   }
 
   message CreatePartialGraphRequest {
-    uint64 partial_graph_id = 1;
+    uint32 partial_graph_id = 1;
+    uint32 database_id = 2;
   }
 
   message RemovePartialGraphRequest {
-    repeated uint64 partial_graph_ids = 1;
+    repeated uint32 partial_graph_ids = 1;
+    uint32 database_id = 2;
   }
 
   oneof request {
--- a/src/stream/src/task/barrier_manager.rs
+++ b/src/stream/src/task/barrier_manager.rs
@@ -12,11 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::collections::{BTreeSet, HashSet};
+use std::collections::{HashMap, HashSet};
 use std::fmt::Display;
-use std::future::pending;
+use std::future::{pending, poll_fn};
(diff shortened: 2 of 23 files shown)


wenym1:
* **#19556** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/19556?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/19556?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

src/stream/src/task/mod.rs
```diff
@@ -40,16 +41,23 @@ pub type UpDownActorIds = (ActorId, ActorId);
 pub type UpDownFragmentIds = (FragmentId, FragmentId);
 
 #[derive(Hash, Eq, PartialEq, Copy, Clone, Debug)]
-struct PartialGraphId(u64);
+pub(crate) struct PartialGraphId(u32);
```
kwannoel, line 44: I recall we used to have database id part of the partial graph id https://github.com/risingwavelabs/risingwave/pull/19173#discussion_r1830932103.

I suppose now that we always have a separate graph per db, this is no longer necessary.

wenym1: Yes. partial graph id should be unique only under a same database, and does not require to be unique globally.

proto/stream_service.proto
```diff
@@ -12,9 +12,10 @@ option optimize_for = SPEED;
 message InjectBarrierRequest {
   string request_id = 1;
   stream_plan.Barrier barrier = 2;
+  uint32 database_id = 3;
   repeated uint32 actor_ids_to_collect = 4;
   repeated uint32 table_ids_to_sync = 5;
-  uint64 partial_graph_id = 6;
+  uint32 partial_graph_id = 6;
```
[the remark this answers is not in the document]
hzxa21: nvm. It is answered in your reply to noel's comment: "partial graph id should be unique only under a same database"

hzxa21 (approved):
Generally LGTM

kwannoel (approved):

BugenZhao:
> This is based on the assumption that there won't be exchange channel between two different databases.

Does it mean that we won't support cross-database references in streaming queries (at least for the initial release)?

wenym1:
> > This is based on the assumption that there won't be exchange channel between two different databases.
> 
> Does it mean that we won't support cross-database references in streaming queries (at least for the initial release)?

As discussed in some offline discussion with @hzxa21 , for the initial version of cross-database streaming queries, we will always read the L0 log store of the upstream table if the table is in other database, so we don't need a streaming dispatcher in this case.

src/stream/src/task/barrier_manager.rs
```diff
@@ -268,97 +276,111 @@ pub(super) struct LocalBarrierWorker {
...
+    ) {
+        poll_fn(|cx| {
+            for (database_id, futures) in &mut *futures {
+                if let Poll::Ready(Some((partial_graph_id, barrier, result))) =
+                    futures.poll_next_unpin(cx)
+                {
+                    return Poll::Ready((*database_id, partial_graph_id, barrier, result));
+                }
+            }
+            Poll::Pending
```
BugenZhao, line 338: If there are a considerable amount of databases, we may consider constructing them into an outer `FuturesUnordered`.

BugenZhao (approved):
LGTM

src/stream/src/task/barrier_manager.rs
wenym1: Might be difficult to do so. 

If we create a temporary `FuturesUnordered` in every select, it may hurt performance somehow since we need to allocate new memory to create new `FuturesUnordered`. If we store the `FuturesUnordered` as a field of local barrier worker, the future has to be static and otherwise we will need to self-reference the `DatabaseManagedBarrierState` stored in local barrier worker. If we make the future to be static, we need to move the ownership of `DatabaseManagedBarrierState` to the future, and we will not be able to access the `DatabaseManagedBarrierState` before the future returns.

PR #19561, opened 2024-11-25
feat(connector): support more parquet data types

wcy-fdu:

Since RisingWave does not have Uint data type, previously we do not support `UIntx`(`UInt8, UInt16, UInt32, UInt64`) in parquet encode, and `Float16` is also not supported.

However, user have a bad experience when they encounter an error of unsupported data type when using file_scan() or parquet file source. 

This pr add some types that can be workarounded:
```
Int8  -> Int16
UInt8  -> Int16
UInt16 -> Int32
UInt32 -> Int64
UInt64 -> Decimal
Float16 -> Float32
```
For example, if the Parquet file that the user needs to read contains a column of Uint64 type, this column should be defined as Decimal type when creating the RisingWave table.

## Checklist

## Documentation

- [x] My PR needs documentation updates. (Please use the **Release note** section below to summarize the impact on users)

## Release note

We need the documentation team to note these special conversions and inform us of the list of data types that we do not support.

Perhaps in https://docs.risingwave.com/integrations/sources/apache-iceberg#data-type-mapping

</details>


chenzl25 (approved):
LGTM!

src/connector/src/source/iceberg/parquet_file_handler.rs
```diff
@@ -318,3 +320,67 @@ pub async fn get_parquet_fields(
...
+/// It handles the following special cases:
+/// - Arrow's `timestamp(_, None)` types (all four time units) match with RisingWave's `TimeStamp` type.
+/// - Arrow's `timestamp(_, Some)` matches with RisingWave's `TimeStamptz` type.
+/// - Since RisingWave does not have an `UInt` type:
+///   - Arrow's `UInt8` matches with RisingWave's `Int16`.
+///   - Arrow's `UInt16` matches with RisingWave's `Int32`.
+///   - Arrow's `UInt32` matches with RisingWave's `Int64`.
+///   - Arrow's `UInt64` matches with RisingWave's `Decimal`.
+/// - Arrow's `Float16` matches with RisingWave's `Float32`.
+fn is_data_type_match(arrow_data_type: &ArrowDateType, rw_data_type: &RwDataType) -> bool {
```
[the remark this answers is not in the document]
wcy-fdu: Sounds reasonable, `arrow -> risingwave` means it's only used in source, let's call it `is_parquet_schema_match_source_schema` or something else.

xiangjinwu (commented):
Seems `int8` is missing from some parts of the test as well as implementation.

wcy-fdu:
> Seems `int8` is missing from some parts of the test as well as implementation.

see 
```
#[test]
    fn int8() {
        let array: PrimitiveArray<i16> = I16Array::from_iter([None, Some(-128), Some(127)]);
        let arr = arrow_array::Int8Array::from(vec![None, Some(-128), Some(127)]);
        let converted: PrimitiveArray<i16> = (&arr).into();
        assert_eq!(converted, array);
    }
```

fuyufjh (approved):
LGTM!

fuyufjh:
Perhaps a bit off-topic. @chenzl25 I think the iceberg test need to be enhanced to cover all data types, for both source and sink.

- Iceberg source

https://github.com/risingwavelabs/risingwave/blob/3607db5e1f5998905fe93497cfe6b0d32da43606/integration_tests/iceberg-source/python/main.py#L48-L53

- Iceberg sink

https://github.com/risingwavelabs/risingwave/blob/5065bb4d1e4ed81aa38f784ae4acb4b644c3e6a8/e2e_test/sink/iceberg_sink.slt#L5


Search for "all_types" and you will see most connectors are already covered, such as https://github.com/risingwavelabs/risingwave/blob/1eb4730cdf1602cf80ac7bb472ec7a65f044a4b8/integration_tests/nats/create_mv.sql#L1-L35

src/common/src/array/arrow/arrow_impl.rs
```diff
@@ -505,6 +505,11 @@ pub trait FromArrow {
             Int16 => DataType::Int16,
             Int32 => DataType::Int32,
             Int64 => DataType::Int64,
+            UInt8 => DataType::Int16,
```
xiangjinwu, line 508: Do we need `Int8` here? If yes, why is it not caught by tests?

e2e_test/s3/file_sink.py
```diff
@@ -62,6 +69,11 @@ def _table():
         sex bigint,
         mark bigint,
         test_int int,
+        test_uint8 smallint,
```
xiangjinwu, line 72: Do we need `test_int8` column here?

src/common/src/array/arrow/arrow_impl.rs
wcy-fdu: Uncover by test.

chenzl25:
> Perhaps a bit off-topic. @chenzl25 I think the iceberg test need to be enhanced to cover all data types, for both source and sink.

Indeed. But actually, the iceberg all types test is here. Agree to add a dedicated test.
https://github.com/risingwavelabs/risingwave/blob/34a2ad42da6a6dcb45b1260b6cecfce89d1c360c/e2e_test/iceberg/test_case/iceberg_sink_no_partition_append_only_table.slt#L12

PR #19562, opened 2024-11-25
feat(batch): add `batch_expr_strict_mode` to ignore expression error in batch query

### LINKED ISSUE #17844: feat(batch): expression's error handling
Some users complains that they do not want some rows' expression error prevent the whole query's processing and result.
### `TRY()` function
presto and trino do this
https://trino.io/docs/current/functions/conditional.html#try
https://prestodb.io/docs/current/functions/conditional.html#try

NOTE: after https://github.com/risingwavelabs/risingwave/pull/12461, the rw's implementation of `eval_infallible` is not equal to the `try()` function in these systems

### implement your own try version of each function, such as `TRY_TO_DATE()`, `TRY_CAST()`.
duckdb, snowflake and databricks do this
https://duckdb.org/docs/sql/expressions/cast.html#try_cast (though it does not have other functions)
https://docs.snowflake.com/en/sql-reference/functions/try_to_date
https://docs.databricks.com/en/sql/language-manual/functions/try_cast.html

Personally, prefer the first one.

fuyufjh:

Resolves #17844. See https://github.com/risingwavelabs/risingwave/issues/17844#issuecomment-2330694033.

Currently, the error is printed to logs only (with `LogSuppressor` to avoid flooding). It would be better to notice to users via psql's `NOTICE`, but it requires more work to pass these messages from compute nodes to frontend nodes. Left a `TODO` in the code now.

## Checklist

- [x] I have added test labels as necessary. See [details](https://github.com/risingwavelabs/risingwave/blob/main/docs/developer-guide.md#ci-labels-guide).

## Documentation

## Release note

- Added a session variable `batch_expr_strict_mode` to control whether to let the entire query fail or fill `NULL` values for expression evaluation failures.

### DIFF
--- a/src/batch/src/executor/filter.rs
+++ b/src/batch/src/executor/filter.rs
@@ -17,10 +17,7 @@ use risingwave_common::array::ArrayImpl::Bool;
 use risingwave_common::array::DataChunk;
 use risingwave_common::catalog::Schema;
 use risingwave_common::util::chunk_coalesce::DataChunkBuilder;
-use risingwave_expr::expr::{
-    build_from_prost, build_non_strict_from_prost_log_report, BoxedExpression,
-};
-use risingwave_expr::expr_context::strict_mode;
+use risingwave_expr::expr::{build_batch_expr_from_prost, BoxedExpression};
 use risingwave_pb::batch_plan::plan_node::NodeBody;
 
 use crate::error::{BatchError, Result};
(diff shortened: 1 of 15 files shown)


chenzl25 (approved):
LGTM！

src/batch/src/executor/filter.rs
```diff
@@ -17,7 +17,7 @@ use risingwave_common::array::ArrayImpl::Bool;
 use risingwave_common::array::DataChunk;
 use risingwave_common::catalog::Schema;
 use risingwave_common::util::chunk_coalesce::DataChunkBuilder;
-use risingwave_expr::expr::{build_from_prost, BoxedExpression};
+use risingwave_expr::expr::{build_batch_expr_from_prost, BoxedExpression};
```
BugenZhao, line 20: Does it mean that `build_from_prost` will not be directly used anymore?

fuyufjh: It is still used quite a bit. Only `Filter` and `Project` have been changed to `build_batch_expr_from_prost`. Examples of the rest that still uses `build_from_prost` include

- `HopWindowExecutor`: because the expression is derived from +/- interval, so it must be non-fallable
- `SortAggExecutor`: same as above, but it's even simpler - there are only `ValueRef`s
- `UpdateExecutor`: because I can't image why users need to ignore expressions here, so I just keep it unchanged.

Issue #19418, opened 2024-11-18
Tracking: Iceberg engine table


chenzl25:
# Milestone 1
- [ ] create table (non-atomic ddl) including creating a hummock table, iceberg sink, external iceberg table, and iceberg source.
- [ ] drop table (non-atomic ddl) including dropping a hummock table, iceberg sink, external iceberg table, and iceberg source.
- [ ] direct DQL to the iceberg source.
- [ ] direct DML to the hummoc table.
- [ ] allows for creating a materialized view on top of the table with an iceberg engine.
- [ ] iceberg catalog: SQL/JDBC catalog
- [ ] iceberg warehouse: S3

# MileStone 2
- [ ] complete data type support
- [ ] atomic create table
- [ ] atomic drop table
- [ ] alter table add column
- [ ] alter table change engine
- [ ] EMR serverless compaction & expiration
- [ ] backup &  restore

# MileStone 3
- [ ] Rust based compaction and expiration
- [ ] local disk cache for iceberg table meta and data
- [ ] iceberg batch query performance (including predicate pushdown and merge on read)

PR #17530, opened 2024-07-02
feat: support config sql metastore using username and password by env

yezizp2012:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://gist.github.com/TennyZhuang/f00be7f16996ea48effb049aa7be4d66#file-rw_cla).

## What's changed and what's your intention?

Resolve #16998 .

1. support config `sqlite`, `postgres` and `mysql` in `--backend`.
2. add new configs correspondingly: `--sql-username`, `--sql-password` and `--sql-database`, which only required for  `postgres` and `mysql`.
3. `--sql-endpoint` should be configured in format `{host}:{port}` if the backend is set to  `postgres` and `mysql`. For `sqlite` it should be the file path.

## Checklist

- [ ] I have written necessary rustdoc comments
- [ ] I have added necessary unit tests and integration tests
- [ ] I have added test labels as necessary. See [details](https://github.com/risingwavelabs/risingwave/blob/main/docs/developer-guide.md#ci-labels-guide).
- [ ] I have added fuzzing tests or opened an issue to track them. (Optional, recommended for new SQL features #7934).
- [ ] My PR contains breaking changes. (If it deprecates some features, please create a tracking issue to remove them in the future).
- [ ] All checks passed in `./risedev check` (or alias, `./risedev c`)
- [ ] My PR changes performance-critical code. (Please run macro/micro-benchmarks and show the results.)

- [ ] My PR contains critical fixes that are necessary to be merged into the latest release. (Please check out the [details](https://github.com/risingwavelabs/risingwave/blob/main/CONTRIBUTING.md))

## Documentation

- [ ] My PR needs documentation updates. (Please use the **Release note** section below to summarize the impact on users)

## Release note

If this PR includes changes that directly affect users or other significant modifications relevant to the community, kindly draft a release note to provide a concise summary of these changes. Please prioritize highlighting the impact these changes will have on users.

Support configure sql metastore using username, password and database separately. The corresponding envs are also supported.

1. support config `sqlite`, `postgres` and `mysql` in `--backend`.
2. add new configs correspondingly: `--sql-username`, `--sql-password` and `--sql-database`, which only required for  `postgres` and `mysql`.
3. `--sql-endpoint` should be configured in format `{host}:{port}` if the backend is set to  `postgres` and `mysql`. For `sqlite` it should be the file path.

PR #19577, opened 2024-11-26
feat(iceberg): support iceberg engine table (in local env)

chenzl25:

- ~~Support `./risedev d iceberg-engine` to set up a local environment to run the iceberg engine table.~~ Use `./risedev d full` directly.
- Support create/drop/select iceberg engine table. The ddl in this PR is not atomic, but it should be fine in the first version.
- The iceberg catalog is stored in our sql meta backend and the data is stored in S3 compatible cloud storage.
- ~~This PR will retrieve meta backend connection info and S3 warehouse info from the environment variable for simplicity, but we will improve it by fetching this info from meta in a later PR.~~
- Make the iceberg sink s3 access key and secret key optional and add an `enable_config_load` field to load credentials from the default credential provided chain.
- No compaction in this PR.

## Example
```sql
create table t(id int primary key, name varchar) engine = iceberg;
select * from t;
```

## Checklist

## Documentation

## Release note

</details>


src/frontend/src/utils/iceberg_engine_opts.rs
```diff
@@ -0,0 +1,60 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::ffi::OsString;
+use std::iter;
+
+use clap::Parser;
+
+#[derive(Parser, Clone, Debug)]
+pub struct IcebergEngineOpts {
```
chenzl25, line 21: Use to capture the environment variables needed for the iceberg engine table, but it would be refactored to retrieve data from meta node

src/frontend/src/handler/drop_table.rs
```diff
@@ -57,10 +63,127 @@ pub async fn handle_drop_table(
...
+            } else {
+                None
+            };
+
+            // TODO(iceberg): handle drop table failures in the middle.
+            // Drop sink
+            // Drop iceberg table
+            //   - Purge table from warehouse
+            //   - Drop table from catalog
+            // Drop source
```
chenzl25, line 111: The drop statement is not atomic and should be improved in the future

src/frontend/src/handler/create_table.rs
```diff
@@ -1261,9 +1287,261 @@ pub async fn handle_create_table(
     );
 
     let catalog_writer = session.catalog_writer()?;
-    catalog_writer
-        .create_table(source, table, graph, job_type)
-        .await?;
+
+    // Handle engine
+    match engine {
```
chenzl25, line 1292: The create table statement is not atomic. It creates hummock table, iceberg sink and iceberg source one by one. It needs to be improved in the future.

ci/workflows/pull-request.yml
```diff
@@ -288,6 +288,21 @@ steps:
     timeout_in_minutes: 15
     retry: *auto-retry
 
+  - label: "end-to-end iceberg engine test"
+    if: build.pull_request.labels includes "ci/run-e2e-iceberg-sink-tests" || build.env("CI_STEPS") =~ /(^|,)e2e-iceberg-sink-tests?(,|$$)/
```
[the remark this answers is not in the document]
chenzl25: We can change the `iceberg sink` label to `iceberg` label.

risedev.yml
```diff
@@ -128,6 +128,64 @@ profile:
...
+    steps:
+      - use: minio
+      - use: postgres
+        port: 5432
+        address: db
+        database: metadata
+        user: postgres
+        password: postgres
+        user-managed: true
+        application: metastore
```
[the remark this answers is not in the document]
chenzl25: Both PG and MySQL are ok.

proto/catalog.proto
```diff
@@ -340,6 +340,12 @@ message Table {
     int32 next_column_id = 2;
   }
 
+  enum Engine {
+    ENGINE_UNSPECIFIED = 0;
+    HUMMOCK = 1;
+    ICEBERG = 2;
```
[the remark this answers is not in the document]
chenzl25: It seems every enum needed to have a UNSPECIFIED field and their id is 0🤔

src/meta/model/src/table.rs
```diff
@@ -102,6 +102,34 @@ impl From<PbHandleConflictBehavior> for HandleConflictBehavior {
...
+        }
+    }
+}
+
+impl From<PbEngine> for Engine {
+    fn from(engine: PbEngine) -> Self {
+        match engine {
+            PbEngine::Hummock => Self::Hummock,
+            PbEngine::Iceberg => Self::Iceberg,
+            PbEngine::Unspecified => unreachable!("Unspecified engine"),
```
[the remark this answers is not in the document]
chenzl25: Oh, it should be Hummock if Unspecified proto id is 0

proto/catalog.proto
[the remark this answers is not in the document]
kwannoel: I think we can deserialize missing Engine to `Hummock` when decode the proto. So it shouldn't caue backwards compat issues.

src/frontend/src/handler/mod.rs
```diff
@@ -1108,3 +1114,115 @@ pub async fn handle(
         _ => bail_not_implemented!("Unhandled statement: {}", stmt),
     }
 }
+
+fn check_ban_ddl_for_iceberg_engine_table(
```
kwannoel, line 1118: Did we ban sink into table for iceberg table as well?

src/frontend/src/utils/iceberg_engine_opts.rs
chenzl25: Sure, I can create another PR first and then refactor this PR.

src/connector/src/connector_common/iceberg/mod.rs
```diff
@@ -85,6 +85,9 @@ pub struct IcebergCommon {
         deserialize_with = "deserialize_optional_bool_from_string"
     )]
     pub path_style_access: Option<bool>,
+    /// enable config load currently is used by iceberg engine, so it only support jdbc catalog.
+    #[serde(default, deserialize_with = "deserialize_optional_bool_from_string")]
+    pub enable_config_load: Option<bool>,
```
kwannoel, line 90: Why not just a boolean flag instead of `option<bool>`

src/meta/model/migration/src/m20241121_101830_table_engine.rs
```diff
@@ -0,0 +1,38 @@
...
+                    .drop_column(Alias::new(Table::Engine.to_string()))
+                    .to_owned(),
+            )
+            .await?;
+        Ok(())
+    }
+}
+
+#[derive(DeriveIden)]
+
```
kwannoel, line 34:
```suggestion
```

kwannoel (approved):
Rest LGTM

src/frontend/src/utils/iceberg_engine_opts.rs
chenzl25: https://github.com/risingwavelabs/risingwave/pull/19594

src/frontend/src/handler/mod.rs
chenzl25: I think we shouldn't ban sink into table, because it should work based on our design.

src/meta/model/migration/src/m20241121_101830_table_engine.rs
```diff
@@ -0,0 +1,38 @@
...
+#[async_trait::async_trait]
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        // Replace the sample below with your own migration scripts
+        manager
+            .alter_table(
+                MigrationTable::alter()
+                    .table(Table::Table)
+                    .add_column(ColumnDef::new(Table::Engine).string().not_null())
```
yezizp2012, line 14: To avoid backward compatible issue we need to specify a default value for this column. Or we can make it nullable instead.

src/connector/src/connector_common/iceberg/mod.rs
chenzl25: option means this is not a required field, especially for user facing iceberg sink and source

src/meta/model/src/table.rs
```diff
@@ -136,6 +164,7 @@ pub struct Model {
     pub cdc_table_id: Option<String>,
     pub vnode_count: i32,
     pub webhook_info: Option<WebhookSourceInfo>,
+    pub engine: Engine,
```
xxchan, line 167: I was thinking whether this would break compatibility (adding a non-null field).

Then I found that 
1. `risedev d` on main, and create a table
2. checkout this branch, `risedev d`.

Then I cannot start cluster 
```
thread 'main' panicked at /Users/xxchan/Projects/risingwave/src/common/common_service/src/observer_manager.rs:76:62:
called `Result::unwrap()` on an `Err` value: Rpc(GrpcStatus(TonicStatusWrapper { inner: Status { code: Internal, message: "invalid hummock context 1", metadata: MetadataMap { headers: {"content-type": "application/grpc", "risingwave-error-bin": "GQAAAAAAAABpbnZhbGlkIGh1bW1vY2sgY29udGV4dCAxAAEHAAAAAAAAAGh1bW1vY2sA", "date": "Fri, 29 Nov 2024 10:06:25 GMT", "content-length": "0"} }, source: Some(ServerError { error: invalid hummock context 1, service_name: Some("hummock"), extra: Extra { score: None } }) }, client_side_service_name: Some("meta") }))
```
The error is a bit strange.. Not sure what's the cause.

src/frontend/src/handler/mod.rs
xxchan: Then what about adding a test case for sink into table?

src/meta/model/src/table.rs
chenzl25: Let me take a look at whether I can reproduce it.

chenzl25: Making the newly added field `engine` optional can resolve this issue. I think it is indeed a backward compatibility issue.

xxchan: @yezizp2012 Do you have any idea why the error is so strange. 🤪

xxchan: BTW, another way to fix it is to add column with default value in migration script. Then we can make this field required. Both OK to me.

xxchan: BTW, just nitpicking: if it's `Option<Engine>`, why bother adding `Hummock` engine. 🤪

chenzl25: To me optional is used for backward compatibility, but the engine without hummock seems a bit weird.

ci/scripts/e2e-iceberg-engine-test.sh
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env bash
+
+# Exits as soon as any line fails.
+set -euo pipefail
+
+source ci/scripts/common.sh
+
+# prepare environment
+export CONNECTOR_LIBS_PATH="./connector-node/libs"
+export AWS_REGION="us-east-1"
```
fuyufjh, line 10: Why `AWS_REGION` is required? IIUC the CI scripts are running on MinIO

src/frontend/src/handler/create_table.rs
```diff
@@ -1320,12 +1349,334 @@ pub async fn handle_create_table(
...
+    let system_params = meta_client.get_system_params().await?;
+    let state_store_endpoint = system_params.state_store().to_string();
+    let data_directory = system_params.data_directory().to_string();
+    let meta_store_endpoint = meta_client.get_meta_store_endpoint().await?;
+
+    let s3_region = if let Ok(s3_region) = std::env::var("AWS_REGION") {
+        s3_region
+    } else {
+        bail!("To create an iceberg engine table, AWS_REGION needed to be set");
+    };
```
fuyufjh, line 1406: Why?

src/frontend/src/handler/create_table.rs
```diff
@@ -1320,12 +1349,334 @@ pub async fn handle_create_table(
...
+    with.insert("table.name".to_string(), iceberg_table_name.to_string());
+    with.insert("enable_config_load".to_string(), "true".to_string());
+    source_handler_args.with_options = WithOptions::new_with_options(with);
+
+    // before we create the table, ensure the JVM is initialized as we use jdbc catalog right now.
+    // If JVM isn't initialized successfully, current not atomic ddl will result in a partially created iceberg engine table.
+    let _ = JVM.get_or_init()?;
+
     let catalog_writer = session.catalog_writer()?;
+    // TODO(iceberg): make iceberg engine table creation ddl atomic
```
fuyufjh, line 1672: Yeah, this is a critical issue... especially if we create tables first, before source and sink. This is because table is self-contained, while `create source` employs a validation stage to check whether the upstream system really work, so it has a high chance to fail. Shall we create the source/sink before table?

src/frontend/src/handler/create_table.rs
```diff
@@ -1320,12 +1349,334 @@ pub async fn handle_create_table(
...
     catalog_writer
         .create_table(source, table, graph, job_type)
```
fuyufjh, line 1674: Here the `source` is passed as well, can it work? 🤔

What I am thinking is that, for a common table with connector, the corresponding `source` is supposed to generate changes that will be applied to the `table`. But here the Iceberg source is just for batch read, which I think is actually irrelevant/unconnected to the iceberg table internally.

src/frontend/src/handler/drop_table.rs
```diff
@@ -57,10 +63,127 @@ pub async fn handle_drop_table(
...
+                .map(|(sink, _)| sink.clone())
+            {
+                // If iceberg source does not exist, use iceberg sink to load iceberg table
+                let iceberg_config = IcebergConfig::from_btreemap(sink.properties.clone())?;
+                Some(Either::Right(iceberg_config))
+            } else {
+                None
+            };
+
+            // TODO(iceberg): make iceberg engine table drop ddl atomic
```
fuyufjh, line 106: I think the current approach is good and no need for the `TODO`. It can ensure the table, source, and sink to be dropped as long as one of them exists.

src/meta/model/src/table.rs
yezizp2012: 1. define the column `engine` without `not_null()` and set it as optional in model.
2. define the column `engine` with `not_null()` and `default(xxx)`, set it as non-optional in model. i.e: https://github.com/risingwavelabs/risingwave/blob/e9e3b24442cc55755c826ec9264fff909e553ef7/src/meta/model/migration/src/m20240911_083152_variable_vnode_count.rs#L3-L7

IIUC These two ways will not cause any backward compatibility issues. Besides please make sure that the definitions in the migration and model are consistent.

ci/scripts/e2e-iceberg-engine-test.sh
chenzl25: Because we use jdbc catalog, the underlying sdk requires aws_region even for minio.

src/frontend/src/handler/create_table.rs
chenzl25: There are some dependencies here. To create an iceberg source, we first need to have an iceberg table. To create an iceberg table we need to create an iceberg sink (with `create_table_if_not_exists`). To create an iceberg sink we need to create a hummock table first. Finally, we have this order `hummock table -> iceberg sink -> iceberg source`

src/frontend/src/handler/create_table.rs
chenzl25: The source here is something like `kafka` and `postgres cdc` instead of the iceberg source.  For example
```
create table t (a int) with (connector = 'kafka' ...) engine = iceberg.
```

src/frontend/src/handler/create_table.rs
chenzl25: https://github.com/risingwavelabs/risingwave/pull/19577#discussion_r1867268505

chenzl25: I think we can check this one for s3 only. For minio, we can always pass a default s3 region.

src/meta/model/src/table.rs
chenzl25: Got it. Right now engine is an optional field, so it is nullable.

xiangjinwu (approved):
for Cargo.lock

src/meta/model/src/table.rs
BugenZhao: For me a non-nullable column with default value can be much clearer. There are redundant semantics between `None` and `Some(Hummock)`.

As mentioned by @yezizp2012, variable vnode count work introduces a required field but not cause any backwards compat issues.

src/frontend/src/handler/create_table.rs
```diff
@@ -1320,12 +1349,334 @@ pub async fn handle_create_table(
...
+        }
+        MetaBackend::Mysql => {
+            format!(
+                "jdbc:mysql://{}:{}/{}",
+                meta_store_host.clone(),
+                meta_store_port.clone(),
+                meta_store_database.clone()
+            )
+        }
+        MetaBackend::Sqlite | MetaBackend::Sql | MetaBackend::Mem => {
```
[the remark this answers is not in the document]
chenzl25: Oh, I previously thought it was deprecated. For iceberg jdbc right now, we need to know the underlying database implementation to choose the right driver.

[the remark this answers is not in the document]
BugenZhao: I think we can simply add a `jdbc:` prefix to the database URL? 🤣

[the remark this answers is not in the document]
yezizp2012: When getting the endpoint from meta, it has already been converted into a form with `postgres:`, `mysql:` as prefixes, including configure as `MetaBackend::Sql`. So `MetaBackend::Sql` should be unreachable.

Unrelated to this PR, 🤔 I think we'd better use `sql` config in `risedev` only for testing purpose. To support scenarios where user and password contain special characters, it's best to specify them separately through env in both production environment and cloud. That's the reason why a subdivided backend was introduced in #17530 .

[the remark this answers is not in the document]
chenzl25: If the underlying database is Oracle or SQL Server, I think that's acceptable. However, I still want to verify the underlying database type. For instance, SQLite is not a suitable catalog for Iceberg. Concurrent updates to SQLite by both the metadata service and Iceberg can easily cause SQLite to become unresponsive.

PR #19582, opened 2024-11-26
feat(storage): Provide table id filter for sstable_iterator

Li0k:

Hummock provides key_range checking at the UserIterator level. But the checking depends on the key_range passed in by the caller.

Due to the introduction of split, the key_range of sstable may not represent the whole object, so this PR further introduces `table_id` check for SstableIterator. 

I think it would make more sense to isolate the check at each level, but this would incur more overhead. We could consider the approach
- UserIterator key check is decentralized to SstableIterator to reduce the overhead, but this doesn't quite satisfy the hierarchical design.
- Introduce a lightweight `table_id` check for SstableIterator (this PR implementation).

This PR changes the behavior of SstableIterator based on key_range.
- Filter block_metas by sstable_info's hints, and read only the block_metas that are in the table_id_range, not all of them.
- Adapting forward_sstable_iterator
- Add UT

This PR does not include changes to the Backward Iterator, and will use a separate PR.

## Checklist

## Documentation

## Release note

</details>


src/storage/src/hummock/sstable/forward_sstable_iterator.rs
```diff
@@ -32,6 +32,7 @@ pub trait SstableIteratorType: HummockIterator + 'static {
         sstable: TableHolder,
         sstable_store: SstableStoreRef,
         read_options: Arc<SstableIteratorReadOptions>,
+        read_table_id_range: (StateTableId, StateTableId),
```
hzxa21, line 35: I think it is more extensible to pass `SstableInfoRef` here instead of the table id range. The implementation of this method can decide what information is needed.

src/storage/src/hummock/sstable/forward_sstable_iterator.rs
```diff
@@ -70,13 +91,15 @@ impl SstableIterator {
...
         }
     }
 
     fn init_block_prefetch_range(&mut self, start_idx: usize) {
         self.preload_end_block_idx = 0;
         if let Some(bound) = self.options.must_iterated_end_user_key.as_ref() {
-            let block_metas = &self.sst.meta.block_metas;
+            let block_metas = &self.sst.meta.block_metas
+                [self.read_block_meta_range.0..=self.read_block_meta_range.1];
             let next_to_start_idx = start_idx + 1;
```
hzxa21, line 103: hmm... I think the `idx` here can be completely wrong given that `block_metas` are sliced in L101.

Li0k: Thanks, I happened to find the bug and have fixed it in a new commit .

src/storage/src/hummock/sstable/forward_sstable_iterator.rs
Li0k: done

src/storage/src/hummock/sstable/backward_sstable_iterator.rs
```diff
@@ -162,6 +163,7 @@ impl SstableIteratorType for BackwardSstableIterator {
         sstable: TableHolder,
         sstable_store: SstableStoreRef,
         _: Arc<SstableIteratorReadOptions>,
+        _sstable_info_ref: &SstableInfo,
```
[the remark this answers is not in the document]
Li0k: separate PR is good to me

src/storage/src/hummock/sstable/forward_sstable_iterator.rs
```diff
@@ -52,14 +53,43 @@ pub struct SstableIterator {
...
+        {
+            start_idx += 1;
+        }
+
+        while end_idx > start_idx
+            && sstable.meta.block_metas[end_idx].table_id().table_id() > read_table_id_range.1
+        {
+            end_idx -= 1;
+        }
+
```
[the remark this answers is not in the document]
Li0k: We assume that the table id read must exist in the sstable, otherwise it is a fatal error.

let me add some assertions to provide the strongest possible assurance.

hzxa21 (approved):
LGTM

src/storage/src/hummock/sstable/forward_sstable_iterator.rs
```diff
@@ -52,14 +53,89 @@ pub struct SstableIterator {
...
 impl SstableIterator {
     pub fn new(
         sstable: TableHolder,
         sstable_store: SstableStoreRef,
         options: Arc<SstableIteratorReadOptions>,
+        sstable_info_ref: &SstableInfo,
     ) -> Self {
+        let mut start_idx = 0;
+        let mut end_idx = sstable.meta.block_metas.len() - 1;
+        let read_table_id_range = (
```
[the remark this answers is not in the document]
Li0k: Yes, it implies the semantics of sort. I remember that vec structs are better for index indexing, like split table_ids into two vec. Would it be easier to implement similar logic with a sorted struct?

zwang28 (approved):

Issue #17020, opened 2024-05-30
Support Avro ref type in source


xxchan:
We added the support for Debezium avro, but not otherwise, which makes no sense.

A little background: In debezium, there are `after` and `before` sharing the same schema.

https://github.com/risingwavelabs/risingwave/blob/8dfae832334ae1dc0585dcae4b3071e9ee6c9b1d/src/connector/src/parser/debezium/avro_parser.rs#L83

BTW, this line is introduced in this huge refactor: https://github.com/risingwavelabs/risingwave/pull/10096, which is too confusing and scary. How does it work before?

tabVersion:
resolve ref in schema is an internal API and has some bugs, we use our patched crate in Cargo.toml. Not sure @xiangjinwu includes the fix when migrating deps.

xiangjinwu:
> Not sure @xiangjinwu includes the fix when migrating deps.

Not yet.

> resolve ref in schema is an internal API and has some bugs, we use our patched crate in Cargo.toml. 

The bug is because we use the internal API the wrong way, or the API itself is poorly designed and hard to use. In short, we need to replaces almost all references to `Schema` into `ResolvedSchema` and use functions that take `schemata` (vec of schemas) rather than a single schema. I will work with @xxchan on resolving this.

xiangjinwu:
The cloning hack does not work in the following situation:
```
{
  "type": "record",
  "name": "Root",
  "fields": [
    {
      "name": "f1",
      "type": {
        "type": "record",
        "name": "Nested",
        "fields": [
          {
            "name": "f1",
            "type": {
              "type": "enum",
              "name": "Case",
              "symbols": ["A", "B", "C"]
            }
          },
          {
            "name": "f2",
            "type": "Case"
          }
        ]
      }
    },
    {
      "name": "f2",
      "type": "Nested"
    }
  ]
}
```

In `ResolvedAvroSchema::resolved_schema` the field `f2.f2` is not resolved and still a `ref`:
<details>

```
Record(
    RecordSchema {
        name: Name {
            name: "Root",
            namespace: None,
        },
        aliases: None,
        doc: None,
        fields: [
            RecordField {
                name: "f1",
                doc: None,
                aliases: None,
                default: None,
                schema: Record(
                    RecordSchema {
                        name: Name {
                            name: "Nested",
                            namespace: None,
                        },
                        aliases: None,
                        doc: None,
                        fields: [
                            RecordField {
                                name: "f1",
[...]

PR #10499, opened 2023-06-23
fix: detect cycle ref in proto

tabVersion:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://gist.github.com/TennyZhuang/f00be7f16996ea48effb049aa7be4d66#file-rw_cla).

## What's changed and what's your intention?

resolve #10475

## Checklist

- [ ] I have written necessary rustdoc comments
- [ ] I have added necessary unit tests and integration tests
- [ ] I have added fuzzing tests or opened an issue to track them. (Optional, recommended for new SQL features #7934).
- [ ] My PR contains breaking changes. (If it deprecates some features, please create a tracking issue to remove them in the future).
- [ ] All checks passed in `./risedev check` (or alias, `./risedev c`)
- [ ] My PR changes performance-critical code. (Please run macro/micro-benchmarks and show the results.)

## Documentation

- [ ] My PR contains user-facing changes.

<details><summary>Click here for Documentation</summary>

### Types of user-facing changes

Please keep the types that apply to your changes, and remove the others.

- Installation and deployment
- Connector (sources & sinks)
- SQL commands, functions, and operators
- RisingWave cluster configuration changes
- Other (please specify in the release note below)

### Release note

</details>

PR #19601, opened 2024-11-28
fix(source): resolve avro `Ref` during `avro_schema_to_column_descs` without hack

xiangjinwu:

Part of #17020. `Ref` in avro used to be supported by a hack we added in our apache_avro fork. This PR avoids the hack (replacing `Ref` with their referee in-place, resulting in an invalid tree containing duplicate definitions). The hack fails to work when there is `Ref` inside `Ref` - resulting in either unresolved `Ref` or infinite recursion.

This PR only corrects part of the usage - in `avro_schema_to_column_descs` to derive RisingWave column data types from avro schema. There will be a follow-up to correct the usage in `convert_to_datum`/`AvroAccess`. Without the latter part, simple data types like `int` (inside a `Ref` in another `Ref`) can already be supported.

Instead of building the expanded-yet-invalid tree as in the hack, this solution passes a `NamesRef` obtained from the root. Its complexity of having an associated lifetime is easy to deal with in this context. (To contrast, `prost_reflect` for protobuf builds the tree with `Arc` - no lifetime and no duplication. But `apache_avro::Schema` does not use `Arc`.)

The circular reference rejection logic is same as #10499 for protobuf. It can be DRY'ed with error message improved later.

This is intended to be part of v2.2 and NOT cherry-picked into earlier versions. There has been a further hack available for earlier versions that can be cherry-picked on demand - where the user is responsible for not using circular reference or an infinite recursion would happen.

## Checklist

## Documentation

## Release note

</details>

### DIFF
--- a/e2e_test/source_inline/kafka/avro/ref.slt
+++ b/e2e_test/source_inline/kafka/avro/ref.slt
@@ -1,0 +1,128 @@
+control substitution on
+
+
+system ok
+rpk topic create avro-ref
+
+
+system ok
+sr_register avro-ref-value AVRO <<EOF
+{
+  "type": "record",
+  "name": "Node",
+  "fields": [
+    {
+      "name": "value",
+      "type": "int"
+    },
+    {
+      "name": "next",
+      "type": ["null", "Node"]
+    }
+  ]
+}
+EOF
+
+
+statement error
+create source s WITH (${RISEDEV_KAFKA_WITH_OPTIONS_COMMON}, topic = 'avro-ref') FORMAT PLAIN ENCODE AVRO (schema.registry = '${RISEDEV_SCHEMA_REGISTRY_URL}');
+----
+db error: ERROR: Failed to run the query
+
+Caused by these errors (recent errors listed first):
+  1: connector error
+  2: circular reference detected: Node -> Node
+
+
+system ok
+curl -X DELETE "${RISEDEV_SCHEMA_REGISTRY_URL}/subjects/avro-ref-value"
+
+
+system ok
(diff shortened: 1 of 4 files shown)


e2e_test/source_inline/kafka/avro/ref.slt
```diff
@@ -0,0 +1,128 @@
...
+
+
+statement error
+create source s WITH (${RISEDEV_KAFKA_WITH_OPTIONS_COMMON}, topic = 'avro-ref') FORMAT PLAIN ENCODE AVRO (schema.registry = '${RISEDEV_SCHEMA_REGISTRY_URL}');
+----
+db error: ERROR: Failed to run the query
+
+Caused by these errors (recent errors listed first):
+  1: connector error
+  2: circular reference detected: Node -> Node
```
xxchan, line 34: We might need to add to err msg that it's from Avro

chenzl25 (approved):
rubber stamp

xxchan (approved):
LGTM

xiangjinwu:
For context, this is the error reported for the added test case if this PR was not applied:
```

  1: connector error
  2: Feature is not yet implemented: Avro type: Ref { name: Name { name: "Point", namespace: None } }
No tracking issue yet. Feel free to submit a feature request at https://github.com/risingwavelabs/risingwave/issues/new?labels=type%2Ffeature&template=feature_request.yml
```

> There will be a follow-up to correct the usage in `convert_to_datum`/`AvroAccess`. Without the latter part, simple data types like `int` (inside a `Ref` in another `Ref`) can already be supported.

Both parts will be included in v2.2. In this intermediate state, `decimal` inside a `Ref` in another `Ref` would result in a parsing error and filled with null, instead of creation rejected without this first part.

PR #12615, opened 2023-10-03
feat: Avoid inserting back operator cache for hash join if the cache entry is too large.

fuyufjh:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://gist.github.com/TennyZhuang/f00be7f16996ea48effb049aa7be4d66#file-rw_cla).

## What's changed and what's your intention?

Resolves #10979

## Checklist

- [ ] I have written necessary rustdoc comments
- [ ] I have added necessary unit tests and integration tests
- [ ] I have added fuzzing tests or opened an issue to track them. (Optional, recommended for new SQL features #7934).
- [ ] My PR contains breaking changes. (If it deprecates some features, please create a tracking issue to remove them in the future).
- [ ] All checks passed in `./risedev check` (or alias, `./risedev c`)
- [ ] My PR changes performance-critical code. (Please run macro/micro-benchmarks and show the results.)

- [ ] My PR contains critical fixes that are necessary to be merged into the latest release. (Please check out the [details](https://github.com/risingwavelabs/risingwave/blob/main/CONTRIBUTING.md))

## Documentation

- [ ] My PR needs documentation updates. (Please use the **Release note** section below to summarize the impact on users)

## Release note

If this PR includes changes that directly affect users or other significant modifications relevant to the community, kindly draft a release note to provide a concise summary of these changes. Please prioritize highlighting the impact these changes will have on users.

</details>

Issue #10979, opened 2023-07-17
feat: Avoid inserting back operator cache for hash join if the cache entry is too large.


chenzl25:
### Is your feature request related to a problem? Please describe.

The cache entry could potentially be quite large, especially when considering scenarios where a fact table is joined with a dimensional table. In such cases, any updates made to the dimensional table would require fetching rows from the fact table, resulting in a significant amplification of rows - possibly reaching into the thousands or even hundreds of thousands. To address this issue, my proposal suggests avoiding inserting these rows back into the LRU cache if the row number exceeds a threshold (e.g.  10000). Make it a configurable option `max_cache_entry_size_for_join`.
https://github.com/risingwavelabs/risingwave/blob/95876460e00e242e800de4aff1cc723c79687bce/src/stream/src/executor/hash_join.rs#L1137-L1138

BugenZhao:
This makes sense because the updates to the dimensional table seem too ad-hoc or occasional in this case and it's not worthwhile caching all these rows.

BTW, to further reduce the memory usage and also the latency, I guess we can even make `hash_eq_match` return a `Stream` instead of collecting the rows into an entry before we can yield them.

fuyufjh:
@yuhao-su Ping

fuyufjh:
The Won'https://github.com/risingwavelabs/risingwave/pull/12615#issuecomment-1744372874 is not significant. Let me close it.

PR #19629, opened 2024-12-01
perf(stream): concurrently fetch row from storage and refill cache

kwannoel:

Currently we read all rows matching a key, encode them, and store them in an in-memory data structure. We add this to the cache. Then we iterate each row from the in-memory data structure, decode them. This means we have to incur IO latency of reading all rows first. We also have to incur the memory cost of buffering all matching records.

Instead of doing this, we can iterate the rows, concurrently refill cache for them. Then we don't have to wait for IO to finish for ALL rows. We also avoid OOM, since while refilling cache, if we notice the number of rows exceeds some threshold, we can stop refilling.

Further, if we tolerate inconsistency, we will revert to the old strategy of reading all records into cache first, so we will preserve that logic.

This is done for all join types in the hash join executor.

## Handling degree table matches and updates

Before this PR we:
1. Read from the match side degree table when doing cache refill (hold an immutable ref)
2. Update the match side degree table when doing cache refill.

However, in this PR we concurrently refill cache and handle matches. This means that 1&2 happen concurrently, which means we hold an immutable and mutable reference to the underlying match side degree table. This will be rejected by the borrow checker.

To solve this, we read from the match side degree table, keeping all the degrees in-memory. Only after that is complete, we start concurrently doing cache refill of the match side, and also updating degrees of match side. In this way the lifetimes of the mutable and immutable reference to match side degree table won't overlap.

In the future we can optimize it further, holding a separate cache for degrees, from their join state, since the degree state is much smaller.

## Nexmark Benchmark results

Build: https://buildkite.com/risingwave-test/nexmark-benchmark/builds/4994
Dashboards:

- [q3](http://metabase.risingwave-cloud.xyz/question/38984-nexmark-q3-blackhole-8c32g-1cn-affinity-avg-sink-output-rows-per-second-rows-s-history-thtb-3222?start_date=2024-10-30)
- [q5](http://metabase.risingwave-cloud.xyz/question/46071-nexmark-q5-blackhole-8c32g-1cn-affinity-avg-sink-output-rows-per-second-rows-s-history-thtb-3228?start_date=2024-11-28)
- [q7](http://metabase.risingwave-cloud.xyz/question/43834-nexmark-q7-blackhole-8c32g-1cn-affinity-avg-sink-output-rows-per-second-rows-s-history-thtb-3240?start_date=2024-11-14)
- [q8](http://metabase.risingwave-cloud.xyz/question/43905-nexmark-q8-blackhole-8c32g-1cn-affinity-avg-sink-output-rows-per-second-rows-s-history-thtb-3246?start_date=2024-11-14)
- [q20](http://metabase.risingwave-cloud.xyz/question/44231-nexmark-q20-blackhole-8c32g-1cn-affinity-avg-sink-output-rows-per-second-rows-s-history-thtb-3213?start_date=2024-11-15)
- [q101](http://metabase.risingwave-cloud.xyz/question/40466-nexmark-q101-blackhole-8c32g-1cn-affinity-avg-sink-output-rows-per-second-rows-s-history-thtb-3170?start_date=2024-10-30)
- [q102](http://metabase.risingwave-cloud.xyz/question/44256-nexmark-q102-blackhole-8c32g-1cn-affinity-avg-sink-output-rows-per-second-rows-s-history-thtb-3173?start_date=2024-11-15)
- [q105](http://metabase.risingwave-cloud.xyz/question/40602-nexmark-q105-blackhole-8c32g-1cn-affinity-avg-sink-output-rows-per-second-rows-s-history-thtb-3182?start_date=2024-10-30)

|        | unoptimized | optimized | remarks |
| - | - | - | - |
| q3 | 7.3k | 7.2k | |
| q5 | 45.7k | 222.8k | |
| q7 | 915 | 3.7k | |
| q8 | 7.7K | 6.5k | |
| q20 | 51.3K | 50.9K | |
| q101 | 250.7K | 283.3K | |
| q102 | 21.8K | 42.6K | |
| q105 | 7.34 | 4.62 | Tps is naturally low, fluctuating between 4-11 |

### Cache Miss

With a 30,000 record limit for each key we measure the performance + memory consumption when a single record is matched against N records which are not in-cache. Here's the comparison (using in-memory statestore, inner join, sample size = 100):

Amp | Total Bytes | Total Bytes (opt) | Diff | Runtime (ns) | Runtime (ns) (opt) | Diff
-- | -- | -- | -- | -- | -- | --
20,000.00 | 3279096 | 3260364 | -0.57 | 17073601.28 | 16143637.7 | -5.45
40,000.00 | 6444552 | 4863221 | -24.54 | 35936541.66 | 29335407.29 | -18.37
200,000.00 | 31769736 | 4863221 | -84.69 | 191953280.4 | 100895451.7 | -47.44
400,000.00 | 63427368 | 4863221 | -92.33 | 387335817.9 | 185162442.9 | -52.20

You can observe how after 30,000, we don't do cache refill anymore, and so the memory usage becomes capped, and we can avoid OOM.

Do also note that because this is an in-memory statestore, we can see significant runtime improvements. I expect that with block cache enabled, we can also see similar improvements, although may not be as high.

### Cache Hit

We join ingest 64 records on the update (probe) side, to match all amplified records on the build side. Sample size is N, using in-memory state store. For memory utilization, we measure mem table usage, it's unavoidable because we have to use a barrier to synchronize the sequence of ingestion, so we can make the sequence of join deterministic, rather than interleaved from lhs and rhs of join. But it's fine because we don't expect better memory utilization, since everything is in cache.

Cache hit results show some regression for runtime (~5-6%):

Amp | Total Bytes | Total Bytes (opt) | Diff | Runtime (ns) | Runtime (ns) (opt) | Diff
-- | -- | -- | -- | -- | -- | --
20,000.00 | 8586702 | 8586702 | 0.00 | 225010070 | 239153682.1 | 6.29
40,000.00 | 17024222 | 17024222 | 0.00 | 457963188.8 | 482480081.3 | 5.35
200,000.00 | 84527806 | 84527806 | 0.00 | 2318848400 | 2468701041 | 6.46
400,000.00 | 168909854 | 168909854 | 0.00 | 4621595553 | 4900965761 | 6.04

However:
1. nexmark e2e performance tests show that the performance is good in most cases.
2. The microbenchmark is a highly specific case. It's an inner join with records in-cache, under high join amplification. This is not exactly the common case. Hence I think it's acceptable to further optimize later where needed.
3. The regression might be due to abstracting various functionality into standalone functions. Some of these are functions that return generators (streams), and are difficult for the compiler to inline. We can inline them by hand to improve performance. But I think we should optimize in other areas first (e.g. append_n (~50%), decode (15%)), which are major cost centers in the flamegraph, rather than optimize areas which lead to poorer readability and maintainability of the code.
4. There are also other more "obvious" optimizations, such as inlining cond expr evaluation if there's no cond expr, eliminating the `is_append_only` code path if join is non-append-only.

## Further improvements

We can also check the cache size when inserting the update records into cache. Currently we only do so for the match side, since that has N matches for a single update record. We differ this optimization to later to avoid further complicating this PR.

We may also need to call try_flush more frequently. To ensure records in join state get flushed.

We also need to remove merge chunks triggered OOM.

## Checklist

## Documentation

## Release note

</details>


kwannoel:
* **#19629** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/19629?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/19629?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

fuyufjh:
The motivaion is similiar to #10979 / #12615, but this is much better. Love it!

IIRC, Nexmark has some queries with large join state. May use it as benchmark.

kwannoel:
Benchmarking this in RWC does not show decrease in mem utilization. There's probably something wrong with the implementation. Writing a memory benchmark first.

kwannoel:
Added benchmark: https://github.com/risingwavelabs/risingwave/pull/19712. Continuing investigation.

kwannoel:
Works well. See PR description. Memory utilization plateaus after the threshold set.

kwannoel:
We don't have any issues supporting this for `INNER JOIN`, however for joins which require a degree table, more consideration is needed. Carving out this optimization just for `INNER JOIN` is possible, but it will add a new code branch and more complexity.

---

Currently we:
1. Read all the degrees and matched rows into memory.
2. Iterate through them, updating the degree table for the build side.

So the upfront IO cost is: `max(read_degree_latency, read_matched_rows_latency)`. Typically `read_matched_rows_latency` should be higher, since `degree` will have less data, only key + degree.

Then the total latency for cache missed will be `max(read_degree_latency, read_matched_rows_latency) + cpu latency of processing matched rows`.

After this PR:
We concurrently read **matched rows** and handle them. However, for **degrees** we can't do so because
we need to iterate over the degrees, and update them at the same time. This means holding an immutable and mutable reference to `StateTable` at the same time.

Currently I workaround this by reading all the degrees into memory first. The size of the degrees shouldn't be too bad, since we just need to store the degrees.
This means the upfront IO cost will be `read_degree_latency`. Then we will need to pay IO costs of `read_matched_rows_latency` in an amortized way, since we concurrently stream and handle each matched row.

Then the total latency for cache missed will be `read_degree_latency + cpu AND io latency of processing matched rows`.
So it *might* take longer to do cache refill.

However, given that cache miss is expected to seldom occur, I think this is a reasonable trade-off to make.

Will benchmark this approach against a nexmark query with `LEFT OUTER JOIN`.

Another approach is to concurrently do point get, and write to the degree state table, since the point get reference to the degree table is short-lived. However, point get does not have prefetch interface yet.

---

The second issue is tolerating inconsistency. Previously we buffered all matched rows into memory. If there's a mismatch in the number of rows in the degree table rows and the matched table rows, we will compare their pk.

I think it's possible to support this with the point get approach. Alternatively, if tolerate inconsistency is set to true, we can always follow the old approach of refilling cache first.

kwannoel:
Will add a bench for `left join`, `cache hit` before proceeding.

kwannoel:
Benchmarking https://buildkite.com/risingwave-test/nexmark-benchmark/builds/4974

src/stream/src/executor/hash_join.rs
```diff
@@ -725,27 +742,27 @@ impl<K: HashKey, S: StateStore, const T: JoinTypePrimitive> HashJoinExecutor<K,
...
         ht: &mut JoinHashMap<K, S>,
-    ) -> StreamExecutorResult<Option<HashValueType>> {
-        if !key.null_bitmap().is_subset(ht.null_matched()) {
-            Ok(None)
+    ) -> StreamExecutorResult<CacheResult> {
+        if enable_strict_consistency() {
+            Ok(ht.take_state_opt(key))
         } else {
-            ht.take_state(key).await.map(Some)
+            ht.take_state(key).await.map(CacheResult::Hit)
```
kwannoel, line 749: If we enable strict consistency, we can avoid having to fetch all state into cache.

However if we tolerate inconsistency, we must fetch all state into cache before handling the matching. Otherwise we can't normalize the inconsistent state between degree table and corresponding state table.

src/stream/src/executor/hash_join.rs
```diff
@@ -821,246 +839,389 @@ impl<K: HashKey, S: StateStore, const T: JoinTypePrimitive> HashJoinExecutor<K,
...
-                        }
-                        // Insert back the state taken from ht.
-                        side_match.ht.update_state(key, matched_rows);
-                        for matched_row in matched_rows_to_clean {
-                            if side_match.need_degree_table {
-                                side_match.ht.delete(key, matched_row)?;
-                            } else {
-                                side_match.ht.delete_row(key, matched_row.row)?;
-                            }
-                        }
```
kwannoel, line 942: This old section of code is entirely removed, because it contains a lot of duplicated logic between the `Insert / Delete` arms. That made it hard to modify.

Hence we refactor it into some common functions and macros.

kwannoel:
Tested for q20, q101, q102, q105. Results are in the PR description.

src/stream/src/executor/join/hash_join.rs
```diff
@@ -224,12 +231,160 @@ pub struct JoinHashMap<K: HashKey, S: StateStore> {
...
+        .row
+        .as_ref()
+        .project(order_key_indices)
+        .chain(once(Some(ScalarImpl::Int64(matched_row.degree as i64))));
+    degree_state.table.update(old_degree_row, new_degree_row);
+}
+
 pub struct TableInner<S: StateStore> {
     /// Indices of the (cache) pk in a state row
     pk_indices: Vec<usize>,
```
kwannoel, line 381: Should this be `cache_key` instead? Or upstream pk indices?

Because `pk_indices` is too vague. It could refer to either:
1. `pk_indices` of the upstream table.
2. `pk_indices` of the join side table (includes join key prefix + upstream table pk).

kwannoel: cc @yuhao-su

src/stream/src/executor/join/hash_join.rs
```diff
@@ -676,53 +877,6 @@ impl<K: HashKey, S: StateStore> JoinHashMap<K, S> {
...
-        join_row: &mut JoinRow<OwnedRow>,
-    ) {
-        self.manipulate_degree(join_row_ref, join_row, |d| {
-            *d = d.checked_sub(1).unwrap_or_else(|| {
-                consistency_panic!("Tried to decrement zero join row degree");
-                0
-            });
-        })
-    }
-
```
kwannoel, line 725: We can handle all of these with a single `update_degree` function above.

src/stream/src/executor/join/hash_join.rs
yuhao-su: yes `cache_key` should be a more accurate name. Thanks

src/stream/src/executor/hash_join.rs
st1page: I am not sure if the normalization is nessarry 🤔 I think it is not HashJoin's duty to "fix" that inconsistency, in other word, it can give a inconsistent output cc @yuhao-su

kwannoel: Actually this just preserves existing logic written by rc. I didn't specially tolerate it.

kwannoel: Inside `take_state` we have some existing logic to handle the case where the state of degree table and join state are inconsistent.

kwannoel: `take_state_opt` does not preserve this logic however, since when cache miss happens, it will not read the entire state into memory. So we can't compare the degree and join state.

st1page: Not implementation-wise, it just feels like there are some performance-related behaviors that change when inconsistent mode is turned on. This would be counter-intuitive for users and ops, so it might be better to adopt the same behavior

kwannoel: I see, so you're suggesting to no longer tolerate inconsistency as was done here previously: https://github.com/risingwavelabs/risingwave/blob/2601c237a7dd677c1314482f91720f4ea78841b5/src/stream/src/executor/join/hash_join.rs#L419-L446.

Trade-offs:
Cons: 
1. Unable to tolerate certain cases of inconsistency in join, e.g. when degree table and join state are inconsistent with each other.
Pros: 
1. More consistent performance for join, even when non-strict mode is enabled. Do note that currently there can already be performance decrease when non-strict mode enabled.
2. Less code branches to maintain.

cc @fuyufjh wdyt?

fuyufjh: +1 for @st1page's viewpoint. This would be counter-intuitive for users and ops.

> take_state_opt does not preserve this logic however, since when cache miss happens, it will not read the entire state into memory. So we can't compare the degree and join state.

"Tolerating inconsistency" always follow a best-effort way - we don't have any guarantee of what the system should behave when inconsistency happens, but just ensure it doesn't panic. As a result, the implementation of tolerating can be different for different code paths. 

Here I think you can just look on your `take_state_opt()` code and avoid all panic, explicit or implicit assertions to keep it run.

src/stream/src/executor/hash_join.rs
```diff
@@ -220,6 +230,61 @@ impl<K: HashKey, S: StateStore, const T: JoinTypePrimitive> HashJoinExecutor<K,
...
+            is_append_only,
+            metrics,
+            chunk_size,
+            high_join_amplification_threshold,
+            None,
+        )
+    }
+
+    #[allow(clippy::too_many_arguments)]
+    pub fn new_with_cache_size(
```
chenzl25, line 258: It seems we could use `new_with_cache_size` implementation to replace the `new` one.

src/stream/src/executor/hash_join.rs
```diff
@@ -821,246 +890,401 @@ impl<K: HashKey, S: StateStore, const T: JoinTypePrimitive> HashJoinExecutor<K,
...
+        )
+        .await;
+
+        if join_condition_satisfied {
+            // update degree
+            *update_row_degree += 1;
+            // send matched row downstream
+
+            // Inserts must happen before degrees are updated,
+            // FIXME(kwannoel): We should let deletes and inserts happen BEFORE degree updates.
```
chenzl25, line 1189: What if we don't let deletes and inserts happen BEFORE degree updates?

src/stream/src/executor/hash_join.rs
kwannoel: We only need it for benchmarking purposes. `new` can implicitly decide the cache size.
It will cause more refactoring to happen, and we have to pass in redundant `None` as an argument in all instantiation sites.

src/stream/src/executor/hash_join.rs
kwannoel: This is more like a TODO to improve code quality. Currently handling inserts happens before updating degree table, whereas handling deletes happens after.

The reason for that is the `with_match_on_<OP>` functions make some implicit assumptions.
For `with_match_on_insert`, it assumes the matched row does not have its degree updated yet.
For `with_match_on_delete`, it assumes the matched row already has its degree updated.

This is existing logic. If I refactor it, I will have to add more changes, which are IMO out-of-scope. It will also touch ASOF join code.

So I leave a FIXME here. I will fix it in subsequent PRs.

src/stream/src/executor/hash_join.rs
kwannoel: Patched.

src/stream/src/executor/hash_join.rs
chenzl25: Thanks for the explanation!

chenzl25 (approved):
Overall LGTM!

st1page:
> We concurrently read matched rows and handle them. However, for degrees we can't do so because
we need to iterate over the degrees, and update them at the same time. This means holding an immutable and mutable reference to StateTable at the same time.
> Currently I workaround this by reading all the degrees into memory first. The size of the degrees shouldn't be too bad, since we just need to store the degrees.

I haven't read the code closely enough. But there are still some contention issues here after just reading all the degree into memory? For a chunk of input, if there are multiple rows corresponding to the same join key, they should access the join key's degree in order. 🤔

kwannoel:
> > We concurrently read matched rows and handle them. However, for degrees we can't do so because
> > we need to iterate over the degrees, and update them at the same time. This means holding an immutable and mutable reference to StateTable at the same time.
> > Currently I workaround this by reading all the degrees into memory first. The size of the degrees shouldn't be too bad, since we just need to store the degrees.
> 
> ~ I haven't read the code closely enough. But there are still some contention issues here after just reading all the degree into memory? For a chunk of input, if there are multiple rows corresponding to the same join key, they should access the join key's degree in order. 🤔 ~
> 
> Nevermind, I think it wrong(

Good question, I also thought about it before. Let me explain here, in case others have the same question. We process the matches for each record in an update chunk one by one. For each record, we will fetch the matched rows and their degrees from cache or object store. Even if multiple records in an update chunk have the same join key, each match for these records will fetch consistent state. Degrees of the cached join rows are also updated accordingly, to make sure they are consistent after matching.

kwannoel:
Any further comments @st1page @yuhao-su if not I'll merge it in tmr.

yuhao-su:
Thanks for the work! LGTM.

yuhao-su (approved):

st1page (approved):

fuyufjh (approved):
LGTM for the Cargo.lock changes.

PR #19657, opened 2024-12-03
refactor(meta): insert time travel metadata in batch

zwang28:

This PR decreases the number of calls to the meta store. The default batch size is set to a small value, because a row containing `SstableInfo` can sometimes be quite large due to the `KeyRange` field.

## Checklist

## Documentation

## Release note

</details>

### DIFF
--- a/src/meta/src/hummock/manager/time_travel.rs
+++ b/src/meta/src/hummock/manager/time_travel.rs
@@ -395,21 +395,35 @@ impl HummockManager {
             })
             .collect::<HashSet<_>>();
         async fn write_sstable_infos(
-            sst_infos: impl Iterator<Item = &SstableInfo>,
+            mut sst_infos: impl Iterator<Item = &SstableInfo>,
             txn: &DatabaseTransaction,
+            batch_size: usize,
         ) -> Result<usize> {
             let mut count = 0;
-            for sst_info in sst_infos {
-                let m = hummock_sstable_info::ActiveModel {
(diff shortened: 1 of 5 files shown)


src/meta/src/hummock/manager/time_travel.rs
```diff
@@ -395,21 +395,35 @@ impl HummockManager {
...
-            for sst_info in sst_infos {
-                let m = hummock_sstable_info::ActiveModel {
-                    sst_id: Set(sst_info.sst_id.try_into().unwrap()),
-                    object_id: Set(sst_info.object_id.try_into().unwrap()),
-                    sstable_info: Set(SstableInfoV2Backend::from(&sst_info.to_protobuf())),
-                };
-                hummock_sstable_info::Entity::insert(m)
+            let mut is_finished = false;
+            while !is_finished {
+                let mut remain = batch_size;
```
Li0k, line 405: Can we use the batch size limit to determine the timing of commits instead of count? 

If I understand correctly, splitting the commit is due to the size limit of the pg insert, so we only need to configure the size, not a count.

zwang28: > the size limit of the pg insert

There's also inherent size limit which is not a fixed value. It depends both the DB setting and the workload (i.e. a row containing SstableInfo can sometimes be quite large due to the KeyRange field.)

Li0k (approved):
LGTM

PR #19677, opened 2024-12-04
feat(storage): support online cache resize via risectl

MrCroxx:

Usage:

```
Usage: risingwave ctl hummock resize-cache [OPTIONS]

Options:
      --meta-cache-capacity-mb <META_CACHE_CAPACITY_MB>  
      --data-cache-capacity-mb <DATA_CACHE_CAPACITY_MB>  
  -h, --help                                             Print help
  -V, --version                                          Print version
```

Related foyer side PRs: https://github.com/foyer-rs/foyer/pull/794 (Already merged 2 weeks ago, not related to the version bump in this PR.)

## Checklist

## Documentation

## Release note

</details>


src/compute/src/rpc/service/config_service.rs
```diff
@@ -42,13 +50,53 @@ impl ConfigService for ConfigServiceImpl {
...
+    async fn resize_cache(
+        &self,
+        request: Request<ResizeCacheRequest>,
+    ) -> Result<Response<ResizeCacheResponse>, Status> {
+        let req = request.into_inner();
+
+        if let Some(meta_cache) = &self.meta_cache
+            && req.meta_cache_capacity > 0
+        {
+            match meta_cache.memory().resize(req.meta_cache_capacity as _) {
```
hzxa21, line 63: IIUC, the resize operation is not persistent, which means if the CN restarts, the cache capacity will be back to the configured values from the toml config. Will this be a problem? For example, if we have 2 CNs in the cluster and after resize_cache, 1 CN restarts for some reason while the other one doesn't. I think it depends on when we are going to use resize_cache. If it is mainly for testing or perf tunning and after the tuning we will update the configs, this is fine. But if we rely on resize_cache in production, that will not be ideal.

src/ctl/src/cmd_impl/hummock/resize_cache.rs
```diff
@@ -0,0 +1,64 @@
...
+    let meta_client = context.meta_client().await?;
+
+    let GetClusterInfoResponse { worker_nodes, .. } = match meta_client.get_cluster_info().await {
+        Ok(resp) => resp,
+        Err(e) => {
+            fail!("Failed to get cluster info: {}", e.as_report());
+        }
+    };
+
+    let futures = worker_nodes.iter().map(|worker| async {
```
Li0k, line 46: just a question:

I'm considering a partial success. Do we need to provide a retry capability to perform an rpc retry for the wrong worker node, to avoid as much as possible inconsistencies in the configs of multiple CNs? (I believe this is an idempotent operation.)

Li0k (approved):
Rest LGTM, thanks for the efforts

src/compute/src/rpc/service/config_service.rs
Li0k: I think it depends on how we define resize.
Is this a permanent operation or a temporary one?

Li0k: In fact, I also think that multi-CN cache config inconsistencies are a concern, and it's not easy to detect. If we allow online resize cache size, then we need to make sure that the operation succeeds on all machines, and is persistent, otherwise this is a risk.

MrCroxx: > IIUC, the resize operation is not persistent, which means if the CN restarts, the cache capacity will be back to the configured values from the toml config. Will this be a problem?

This feature just aims to resize the in-memory meta/data block cache without downtime. If there are restarts, it is fine to modify the persistent configuration directly. Modifying the per-node configuration can be (and maybe better be) achieved in the cloud control panel.

src/ctl/src/cmd_impl/hummock/resize_cache.rs
MrCroxx: IMO, it is okay to make the user responsible to retry it.

hzxa21 (approved):
LGTM

Issue #13801, opened 2023-12-05
Add rate limit for DML executor


kwannoel:

kwannoel:
Usecase:
1. User updates dimension table. The resulting amplification from the fact table is too high, leading to OOM.
2. They can apply dml rate limit.

PR #16407, opened 2024-04-19
ensure `chunk_size` under `rate_limit`

stdrc:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://gist.github.com/TennyZhuang/f00be7f16996ea48effb049aa7be4d66#file-rw_cla).

## Checklist

- [ ] I have written necessary rustdoc comments
- [ ] I have added necessary unit tests and integration tests
- [ ] I have added test labels as necessary. See [details](https://github.com/risingwavelabs/risingwave/blob/main/docs/developer-guide.md#ci-labels-guide).
- [ ] I have added fuzzing tests or opened an issue to track them. (Optional, recommended for new SQL features #7934).
- [ ] My PR contains breaking changes. (If it deprecates some features, please create a tracking issue to remove them in the future).
- [ ] All checks passed in `./risedev check` (or alias, `./risedev c`)
- [ ] My PR changes performance-critical code. (Please run macro/micro-benchmarks and show the results.)

- [ ] My PR contains critical fixes that are necessary to be merged into the latest release. (Please check out the [details](https://github.com/risingwavelabs/risingwave/blob/main/CONTRIBUTING.md))

## Documentation

- [ ] My PR needs documentation updates. (Please use the **Release note** section below to summarize the impact on users)

## Release note

If this PR includes changes that directly affect users or other significant modifications relevant to the community, kindly draft a release note to provide a concise summary of these changes. Please prioritize highlighting the impact these changes will have on users.

</details>

PR #19679, opened 2024-12-04
feat: add support for dml_rate_limit

zwang28:

close https://github.com/risingwavelabs/risingwave/issues/13801

## Checklist

## Documentation

## Release note

</details>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12142,6 +12142,7 @@ name = "risingwave_stream"
 version = "2.2.0-alpha"
 dependencies = [
  "anyhow",
+ "arc-swap",
  "assert_matches",
  "async-recursion",
  "async-stream",
--- a/proto/meta.proto
+++ b/proto/meta.proto
@@ -309,6 +309,7 @@ enum ThrottleTarget {
   MV = 2;
   TABLE_WITH_SOURCE = 3;
   CDC_TABLE = 4;
+  TABLE_DML = 5;
 }
 
 message ApplyThrottleRequest {
--- a/proto/stream_plan.proto
+++ b/proto/stream_plan.proto
@@ -768,6 +768,7 @@ message DmlNode {
   uint64 table_version_id = 3;
   // Column descriptions of the table.
   repeated plan_common.ColumnDesc column_descs = 2;
+  optional uint32 rate_limit = 4;
 }
 
 message RowIdGenNode {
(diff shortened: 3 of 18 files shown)


kwannoel (commented):
Can we add a test for DML rate limit?

Something like:
1. Set dml rate limit on table + parallelism=1.
2. Trigger DML statement in background
    ```
    system ok
    risedev psql -c 'insert into ... values ...' &
    ```

3. Sleep 10s
4. number of rows in table < 20.

Run the test only in non-madsim context.

src/stream/src/executor/utils.rs
```diff
@@ -22,3 +22,21 @@ impl Execute for DummyExecutor {
...
+pub fn compute_rate_limit_chunk_permits(chunk: &StreamChunk, limit: usize) -> usize {
+    let chunk_size = chunk.capacity();
+    let ends_with_update = if chunk_size >= 2 {
+        // Note we have to check if the 2nd last is `U-` to be consistenct with `StreamChunkBuilder`.
+        // If something inconsistent happens in the stream, we may not have `U+` after this `U-`.
+        chunk.ops()[chunk_size - 2].is_update_delete()
+    } else {
+        false
+    };
+    if chunk_size == limit + 1 && ends_with_update {
```
chenzl25, line 35: What if the limit is `1` and the chunk is [-U, +U]

chenzl25 (approved):
Rest LGTM!

src/stream/src/executor/utils.rs
zwang28: The chunk of size 2 can successfully obtain 1 permit from the rate limiter.
Actually this is the code snippet from https://github.com/risingwavelabs/risingwave/pull/16407. It's been moved to utils.rs for reusability.

kwannoel (approved):
LGTM for now. But IMO dml rate limit should be able to break the atomicity. In some cases, the user may insert a large number of records in a single insert statement.

After all persistence is only guaranteed by `flush`.

MrCroxx (approved):

zwang28:
> LGTM for now. But IMO dml rate limit should be able to break the atomicity. 

Agree.

> In some cases, the user may insert a large number of records in a single insert statement.

The atomicity will be broken after processing MAX_CHUNK_FOR_ATOMICITY chunks. Since MAX_CHUNK_FOR_ATOMICITY is hard-coded to 32, I think it's good for now.

PR #19690, opened 2024-12-05
feat(iceberg): refine iceberg table column datatype and name.

chenzl25:

- Previously, the iceberg engine table replaced the scan with the iceberg source in the binder, but it will make the batch query datatype different from the datatype defined in the table. This PR moves this replacement from the binder to the optimizer and casts iceberg datatype to risingwave datatype. 
- Support those types that aren't supported in Iceberg (like `jsonb`, `smallint`, `serial`, and so on).
- Iceberg V3 spec use `_row_id` as a reserved column name for row lineage, so if the table without primary key, we can't use `_row_id` directly for iceberg, this PR change it to `_risingwave_iceberg_row_id`

## Checklist

## Documentation

## Release note

</details>


src/frontend/src/planner/relation.rs
```diff
@@ -93,10 +97,92 @@ impl Planner {
...
+                                    FunctionCall::cast_mut(
+                                        &mut input_ref,
+                                        field.data_type().clone(),
+                                        CastContext::Explicit,
+                                    )
+                                    .unwrap();
+                                    input_ref
+                                }
+                            } else {
+                                // fields like `_rw_timestamp`, would not be found in source.
```
fuyufjh, line 164: Is there any other example? 

I am think whether we should `bail!` if a column is not found in the Iceberg source and it's not `_rw_timestamp`

fuyufjh (approved):
LGTM

src/frontend/src/planner/relation.rs
chenzl25: I think `_rw_timestamp` is the only one that would hit this branch. 

> I am think whether we should bail! if a column is not found in the Iceberg source and it's not _rw_timestamp

This is the planner's beginning, so we must select all columns from a table (including the `_rw_timestamp`) unless we move this conversion to the end of the optimizer (after column pruning).

PR #19701, opened 2024-12-06
refactor(source): extract avro inner schema precisely

xiangjinwu:

Previously, when recursively handling an avro value, its associated schema is extracted with `extract_inner_schema` (wrapper over `avro_extract_field_schema`). It handles `Record`/`Array`/`Union`/`Map` together and allows `None` as output - meaning we may have a value with unknown schema.

This PR revisits all places extracting child schema and ensures the corresponding schema is always present:
* `AvroParseOptions::schema` is no longer an `Option`. It cannot be `None`.
* Extraction of `Record`/`Array`/`Union`/`Map` are handled in their corresponding context. Removes `avro_extract_field_schema` and `extract_inner_schema`.
* `avro_schema_skip_nullable_union` which builds `anyhow::Error` (with backtrace) for `get_nullable_union_inner` is removed. Callers use `get_nullable_union_inner` directly and may build an error without backtrace.

## Checklist

## Documentation

## Release note

</details>

### DIFF
--- a/e2e_test/source_inline/kafka/avro/ref.slt
+++ b/e2e_test/source_inline/kafka/avro/ref.slt
@@ -109,7 +109,11 @@ select
   (bar).b.y
 from s;
 ----
-3 4 5 6 6 5 4 3
+3 4 5 6 NULL NULL NULL NULL
+
+# Parsing of column `bar` fails even with ints because now `schema` is required.
+# This will be fully supported in the next PR
+# 3 4 5 6 6 5 4 3
 
 
 statement ok
--- a/src/connector/codec/src/decoder/avro/mod.rs
+++ b/src/connector/codec/src/decoder/avro/mod.rs
@@ -450,7 +450,7 @@ pub(crate) fn avro_decimal_to_rust_decimal(
 }
 
 /// If the union schema is `[null, T]` or `[T, null]`, returns `Some(T)`; otherwise returns `None`.
-fn get_nullable_union_inner(union_schema: &UnionSchema) -> Option<&'_ Schema> {
+pub fn get_nullable_union_inner(union_schema: &UnionSchema) -> Option<&'_ Schema> {
     let variants = union_schema.variants();
     // Note: `[null, null] is invalid`, we don't need to worry about that.
     if variants.len() == 2 && variants.contains(&Schema::Null) {
@@ -461,19 +461,6 @@ fn get_nullable_union_inner(union_schema: &UnionSchema) -> Option<&'_ Schema> {
         Some(inner_schema)
     } else {
         None
-    }
-}
-
-pub fn avro_schema_skip_nullable_union(schema: &Schema) -> anyhow::Result<&Schema> {
-    match schema {
-        Schema::Union(union_schema) => match get_nullable_union_inner(union_schema) {
-            Some(s) => Ok(s),
-            None => Err(anyhow::format_err!(
-                "illegal avro union schema, expected [null, T], got {:?}",
-                union_schema
-            )),
-        },
-        other => Ok(other),
     }
 }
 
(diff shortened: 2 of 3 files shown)


src/connector/codec/src/decoder/avro/mod.rs
```diff
@@ -266,14 +253,17 @@ impl<'a> AvroParseOptions<'a> {
                 ScalarImpl::Interval(Interval::from_month_day_usec(months, days, usecs))
             }
             // ---- Struct -----
-            (DataType::Struct(struct_type_info), Value::Record(descs)) => StructValue::new(
+            (DataType::Struct(struct_type_info), Value::Record(descs)) => StructValue::new({
+                let Schema::Record(record_schema) = &self.schema else {
+                    return Err(create_error());
+                };
```
xxchan, line 259: random thought: maybe ideally, there should be some wrapper like `ValueWithSchema::Record(descs, schema)`.

xxchan: check in each branch looks verbose and error-prone

src/connector/codec/src/decoder/avro/mod.rs
```diff
@@ -405,13 +401,23 @@ impl Access for AvroAccess<'_> {
...
                 }
                 Value::Record(fields) => {
-                    if let Some((_, v)) = fields.iter().find(|(k, _)| k == key) {
-                        value = v;
-                        options.schema = options.extract_inner_schema(Some(key));
+                    let Schema::Record(record_schema) = &options.schema else {
+                        return Err(create_error());
+                    };
+                    if let Some(idx) = record_schema.lookup.get(key) {
+                        // if let Some((_, v)) = fields.iter().find(|(k, _)| k == key) {
```
xxchan, line 418: why to leave this comment? and what if lookup failed? 🤔

e2e_test/source_inline/kafka/avro/ref.slt
```diff
@@ -109,7 +109,11 @@ select
   (bar).b.y
 from s;
 ----
-3 4 5 6 6 5 4 3
+3 4 5 6 NULL NULL NULL NULL
+
+# Parsing of column `bar` fails even with ints because now `schema` is required.
+# This will be fully supported in the next PR
+# 3 4 5 6 6 5 4 3
```
xxchan, line 116: 👀

xxchan (approved):
👍 Generally LGTM. The idea is great

src/connector/codec/src/decoder/avro/mod.rs
xiangjinwu: This annoying pattern has been common, where we match both `DataType::Struct` and `Scalar::Struct`. Are you suggesting there is a better way to handle such cases?

My own practice has been to match on `DataType` and *assert* on `Scalar` with `as_struct`. An alternative form of `as_struct` is the `try_match_expand` macro. But in this PR I tried to avoid such large structural refactor and focus on correcting the inner schema extraction within the existing gigantic match.

Btw I do not think matching them as a tuple (`(DataType, Value)` or `(DataType, Value, Schema)`) is good either. It means we inevitably need a wildcard `_` and compiler would not remind us of unhandled data types.

xxchan: It's like this:
Basically wrap your practice into a separate step, and later we can just exhaustively match.
```rust
enum ScalarWithType {
    Struct(struct_value, struct_type),
    ....
}

impl Scalar {
    fn combine_type(self, type: DataType) -> ScalarWithType {
        match type {
            DataType::Struct(struct_type) => { 
                ScalarWithType(self.as_struct(), struct_type)
            }
            ...
        }
    }
}
```

src/connector/codec/src/decoder/avro/mod.rs
xiangjinwu: * When I cleanup I only checked `convert_to_datum` but missed `access`
* When lookup fails, the behavior is as before: `continue` is not executed so `Err(create_error())?;` at the end of `while`-body is executed. This function returns an error to the caller and the caller would set this whole column as `NULL`.

PR #19746, opened 2024-12-10
fix(source): resolve avro `Ref` during `access` without hack

### LINKED ISSUE #17020: Support Avro ref type in source

xiangjinwu:

Sequel to #19601 and resolves #17020.

The recursive call is changed from `Self {schema: inner, relax_numeric: self.relax_numeric}.convert_to_datum(v, field_type)` into `self.convert_to_datum(inner, v, field_type)` so that
* The changing part `schema: inner` is passed as argument rather than keep reconstructing new `AvroParseOptions`.
* `self: AvroParseOptionsInner` captures the immutable "global" context: `relax_numeric: bool` and `refs: NamesRef<'a>`

On the use site, `self.schema` is changed to `self.lookup_ref(unresolved_schema)`. Note that introducing a new type to distinguish `resolved` from `unresolved` would not make it less error-prone for a careless developer - nothing prevents them from matching against unresolved `apache_avro::Schema` directly and ignoring `Schema::Ref`.

~~Full removal of `ResolvedAvroSchema` will be done in an immediate follow-up, to minimize the signature changes in this PR.~~ `ResolvedAvroSchema` will either be fully removed or repurposed to hold `HashMap<Name, Schema>` (an owned `NamesRef`) in a follow-up.

Same as #19601, this is intended to be part of v2.2 and NOT cherry-picked into earlier versions.

## Checklist

## Documentation

## Release note

</details>

### DIFF
--- a/src/connector/codec/src/decoder/avro/mod.rs
+++ b/src/connector/codec/src/decoder/avro/mod.rs
@@ -14,7 +14,7 @@
 
 mod schema;
 
-use apache_avro::schema::{DecimalSchema, UnionSchema};
+use apache_avro::schema::{DecimalSchema, NamesRef, UnionSchema};
 use apache_avro::types::{Value, ValueKind};
 use apache_avro::{Decimal as AvroDecimal, Schema};
 use chrono::Datelike;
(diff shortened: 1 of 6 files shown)


xxchan (approved):
:GTM

src/connector/codec/src/decoder/avro/mod.rs
```diff
@@ -33,28 +33,46 @@ use super::{bail_uncategorized, uncategorized, Access, AccessError, AccessResult
 use crate::decoder::avro::schema::avro_schema_to_struct_field_name;
 
 #[derive(Clone)]
-/// Options for parsing an `AvroValue` into Datum, with an optional avro schema.
+/// Options for parsing an `AvroValue` into Datum, with a root avro schema.
 pub struct AvroParseOptions<'a> {
-    /// Currently, this schema is only used for decimal.
-    ///
-    /// FIXME: In theory we should use resolved schema.
-    /// e.g., it's possible that a field is a reference to a decimal or a record containing a decimal field.
```
xiangjinwu, line 41: Now we can support _reference to a decimal or a record containing a decimal field_.

PR #19771, opened 2024-12-12
feat(sink): introduce exactly once iceberg sink

wcy-fdu:

This pr introduce exactly once consistency semantics for iceberg sink. 

Related RFC: [Exactly Once File Sink](https://www.notion.so/risingwave-labs/RFC-Exactly-Once-File-Sink-126f0d69cb768068a751e62e80544c60?pvs=4)

## implementation detail
The implementation of this feature can be mainly divided into four parts:
1. introduced a table that exists in the meta store.
2. normal write process(when no recovery occurs):
-    Pre-commit the relevant metadata to the meta store before the Iceberg commit occurs. (Since the actual commit of Iceberg is completed by the coordinator after collecting all parallel tasks in the meta store, this is directly implemented within the `coordinator.commit()` interface.)
-   Immediately delete the written metadata after the Iceberg commit returns successfully. (Since those data is successfully written into iceberg, deleting metadata is safe).
3. Recovery logic
All logic is completed in the `coordinator.init()` interface when recovery occurs. This is because all memory states are cleared during recovery, and then the sink writer is built first, and then the sink writer builds the coordinator.
- check whether the pre-commit metadata recorded in the meta store < the current epoch contains any uncommitted data, and do `re_commit`. 
    - If yes, this means that the previous commit failed, and iceberg's repeated submission is idempotent, so re-commit.
    - If no, this means that the previous commit success, just skip.
- Modify the log store rewind so that it can start rewinding from a certain epoch. 
- After re-commit or skip is completed, and pass the `max_epoch` called `last_recommit_epoch` in the code, is passed to CN through the existing `StartCoordinationResponse` for log store rewind. The rewind here is to avoid repeated consumption of the log store, because this batch of data must have been in the downstream iceberg at this time.

## test logic
The testing part is still being improved and will be continuously updated in PR.
After some offline discussions, we decided to test it this way: simulate error injection through madsim, and then sink the same batch of source data into iceberg twice, showing two iceberg tables. One of them has errors injected, and the other does not. After comparing the two tables, they are exactly the same.

## Notes:
1. Since iceberg is currently hosted in RisingWave, the GC cycle is 2 weeks. This PR currently assumes that the original s3 file that failed to be committed to iceberg `will not `be GC before the next successful commit. In later pr, the `re-commit` step after recovery will be changed to rewind to the start epoch.
2. All metadata deletions happen immediately after iceberg commit successfully. correct me if I am wrong.

## Checklist

## Documentation

<details>

</details>


src/connector/src/sink/iceberg/mod.rs
```diff
@@ -1166,21 +1233,98 @@ impl<'a> TryFrom<&'a IcebergCommitResult> for SinkMetadata {
...
+            let mut last_recommit_epoch = 0;
+            for (end_epoch, sealized_bytes) in metadata_map {
+                let write_results_bytes = deserialize_metadata(sealized_bytes);
+                let mut write_results = vec![];
+                let mut write_results_to_recommit = vec![];
+
+                for each in write_results_bytes {
+                    let write_result = IcebergCommitResult::try_from_sealized_bytes(&each)?;
+                    write_results.push(write_result);
+                    let write_result1 = IcebergCommitResult::try_from_sealized_bytes(&each)?;
```
wcy-fdu, line 1301: Should fix the double allocate.

wcy-fdu: @ZENOTME is working on add `#[derive(Clone)]` for `DataFile` in iceberg-rust.

hzxa21 (commented):
Left some early comments

src/stream/src/executor/sink.rs
```diff
@@ -541,11 +541,24 @@ impl<F: LogStoreFactory> SinkExecutor<F> {
...
-            .new_log_sinker(sink_writer_param.clone())
-            .and_then(|log_sinker| log_sinker.consume_log_and_sink(&mut log_reader))
+        while let Err(e) = {
+            let res = sink.new_log_sinker(sink_writer_param.clone()).await;
+
+            async {
+                match res {
+                    Ok((log_sinker, log_store_rewind_start_offset)) => {
+                        log_reader
+                            .rewind(log_store_rewind_start_offset)
```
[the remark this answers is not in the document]
wcy-fdu: I think truncate will only work after checkpoint finish, which means it is asynchronous, while the rewind operation  will effective immediately.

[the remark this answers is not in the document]
hzxa21: In terms of persistence, truncate is async. But for rewind, I think it is less relevant to persistence because it uses the in memory truncate offset to do the seek. You can check the implementation of rewind.

src/meta/src/manager/sink_coordination/coordinator_worker.rs
```diff
@@ -252,10 +257,12 @@ impl CoordinatorWorker {
...
     ) -> anyhow::Result<()> {
-        coordinator.init().await?;
+        let log_store_rewind_start_epoch = coordinator.init().await?;
         loop {
-            let (handle_id, vnode_bitmap, epoch, metadata) =
-                self.handle_manager.next_commit_request().await?;
+            let (handle_id, vnode_bitmap, epoch, metadata) = self
+                .handle_manager
+                .next_commit_request(log_store_rewind_start_epoch)
+                .await?;
```
chenzl25, line 265: The `log_store_rewind_start_epoch` here is a const, while in `next_commit_request` we can't rewind to `log_store_rewind_start_epoch` for each new worker (considering online scaling, a new worker will join in the middle).

src/connector/src/sink/iceberg/mod.rs
```diff
@@ -1366,8 +1559,189 @@ impl SinkCommitCoordinator for IcebergSinkCommitter {
...
+        &self,
+        iceberg_config: &IcebergConfig,
+        write_results: Vec<IcebergCommitResult>,
+    ) -> Result<bool> {
+        let iceberg_common = iceberg_config.common.clone();
+        let table = iceberg_common
+            .load_table(&iceberg_config.java_catalog_props)
+            .await?;
+
+        if let Some(snapshot) = table.metadata().current_snapshot() {
```
chenzl25, line 1655: Currently, we only search files in the current snapshot, however, asynchronous compaction might create a new snapshot which might replace all files in the current snapshot. I think we need to find files in all snapshots (a little bit heavy).

src/meta/model/src/exactly_once_iceberg_sink.rs
```diff
@@ -0,0 +1,32 @@
...
+#[derive(Clone, Debug, PartialEq, DeriveEntityModel, Eq, Default)]
+#[sea_orm(table_name = "exactly_once_iceberg_sink_metadata")]
+pub struct Model {
+    #[sea_orm(primary_key, auto_increment = false)]
+    pub sink_id: u32,
+    #[sea_orm(primary_key, auto_increment = false)]
+    pub end_epoch: u64,
+    pub start_epoch: u64,
+    pub metadata: ::prost::alloc::vec::Vec<u8>,
+}
```
chenzl25, line 27: I suggested to add one more enum field `status` (`prepared`, `committed`) for better auditing and debugging.

src/connector/src/sink/iceberg/mod.rs
```diff
@@ -1366,8 +1559,189 @@ impl SinkCommitCoordinator for IcebergSinkCommitter {
...
+        let models: Vec<Model> = Entity::find()
+            .filter(Column::SinkId.eq(sink_id))
+            .all(db)
+            .await?;
+
+        let mut result = BTreeMap::new();
+
+        for model in models {
+            result.insert(model.end_epoch, model.metadata);
+        }
```
chenzl25, line 1638: We can add an index on this column and use `Order by` to sort the result.

src/connector/src/sink/iceberg/mod.rs
```diff
@@ -1366,8 +1559,189 @@ impl SinkCommitCoordinator for IcebergSinkCommitter {
         }
 
         tracing::info!("Succeeded to commit to iceberg table in epoch {epoch}.");
+
+        if self.is_exactly_once {
+            self.delete_row_by_sink_id_and_end_epoch(&self.db, self.sink_id, epoch)
+                .await?;
+            tracing::info!("Succeeded delete pre commit metadata in epoch {epoch}.");
+        }
```
chenzl25, line 1567: For better observability, I suggest using the additional field `status` and mark it as `committed` instead of deleting it immediately. We can delete the log asynchronously.

chenzl25: BTW, for dropping sink, we need to delete the commit log as well.

src/meta/model/migration/src/m20250303_081439_exactly_once_iceberg_sink_metadata.rs
```diff
@@ -0,0 +1,55 @@
...
+
+#[derive(DeriveMigrationName)]
+pub struct Migration;
+#[async_trait::async_trait]
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        manager
+            .create_table(
+                Table::create()
+                    .table(ExactlyOnceIcebergSinkMetadata::Table)
```
chenzl25, line 13: Shall we rename the table to a more general name instead of  `iceberg` specific? I think we can reuse this table for other exactly once sink as well

src/connector/src/sink/iceberg/mod.rs
chenzl25: I think searching the snapshot files one by one is too heavy. It would be better if Iceberg-rs can let us to specify the snapshot id for a (Iceberg fast append) transaction. cc @ZENOTME  We can generate the snapshot id in advance and record it to the metastore. Later we can check whether it has been committed more easily.  https://github.com/risingwavelabs/iceberg-rust/blob/683fb89edeaf8d1baae69e1f376d68b92be1d496/crates/iceberg/src/transaction.rs#L139

ZENOTME: For this scene, instead of specifying the snapshot ID, can we just read back the snapshot ID? The snapshot ID will be generated before the transaction commit.

src/meta/model/src/exactly_once_iceberg_sink.rs
wcy-fdu: Sounds good.

src/meta/model/migration/src/m20250303_081439_exactly_once_iceberg_sink_metadata.rs
wcy-fdu: -1. Previously we had some discussion that make per connector per meta store table, which means for different sink types(file sink, kafka, pg sink), they using different metadata instead of sharing one. Reason:
- Although most sinks follow the general 2pc framework, it may not be a good idea to assume that pre-commit metadata is consistent before implementing them. eg. for unimplemented file sinks, perhaps its meta data needs to include vnode information
- Multiple tables are easier to manage than one table? Although we use sink_id as pk in the table, I appreciate that there is still management cost for general.

src/meta/src/manager/sink_coordination/coordinator_worker.rs
chenzl25: cc @wenym1 It doesn't appear to be a correctness issue, as the coordinator will be terminated after scaling.

src/meta/model/migration/src/m20250303_081439_exactly_once_iceberg_sink_metadata.rs
chenzl25: Yes. I think using an individual table for each connector is easier to manage.

src/meta/model/src/exactly_once_iceberg_sink.rs
wcy-fdu: After some offline discussion, this is not needed at the moment.

src/meta/src/rpc/ddl_controller.rs
```diff
@@ -1551,7 +1553,13 @@ impl DdlController {
         let version = self
             .drop_object(object_type, object_id, drop_mode, target_replace_info)
             .await?;
+        if let StreamingJobId::Sink(sink_id) = job_id {
+            // delete system table for exactly once iceberg sink
+            // todo(wcy-fdu): optimize the logic to be Iceberg unique.
 
+            let db = self.env.meta_store_ref().conn.clone();
+            clean_all_rows_by_sink_id(&db, sink_id).await?;
+        }
```
chenzl25, line 1562: I think we should put this logic into `drop_object` of `CatalogController`, so that all these tables modifications could be handled in one transaction,

wcy-fdu:
For changes in log store part, please review https://github.com/risingwavelabs/risingwave/pull/20772.

src/meta/src/rpc/ddl_controller.rs
wcy-fdu: -1 for put this logic inside the transaction of `CatalogControlle::drop_object`. Had some offline discussion with @chenzl25 and share the consensus we reached:
1. When dropping the sink, we need to perform a guaranteed deletion to handle a corner case: if `iceberg.commit()` keeps failing, the deletion of the system table that follows the commit will not be executed. Since the sink is decoupled, there won't be any recovery, and the user might choose to directly drop the sink, which means that the record from the failed commit won't be deleted from the system table. Therefore, I agree that a fallback deletion should be performed when dropping the sink.
2. Regarding the aforementioned deletion operation, should it be placed inside the transaction or outside? 
The motivation for placing it inside the transaction is to consider that during a cascade delete, the provided object_id might be the table_id instead of the sink_id, and only within the transaction can we obtain the sink_id. However, if we delete within the transaction, there is a problem: any inserts that occur after the transaction starts will not be deleted, which will definitely leave a dirty record, eg:

```
txn.begin();
delete (sink_id=1, epoch=1);
//during this txn, new commit occurs, it will perform
// (insert(sink_id = 1, epoch=2); 
// iceberg.commit(epoch=2))
// as sink is not dropped yet, and the new insert is blocked by txn
txn.commit()

// after txn committed, the blocked insert will performed
insert(sink_id = 1, epoch=2);
// since sink is dropped, iceberg.commit(2) will not perform, so no more deletion will be performed

then (sink_id = 1, epoch=2) will not be deleted.
```

Therefore, we have decided to perform the deletion outside the transaction. This way, for non-cascade drop sink operations, we can ensure that the deletion occurs. 

Of course, if the downstream Iceberg sink keeps reporting errors, but the user executes a cascade drop (dropping its upstream table or materialized view), this sink will be cascade dropped, resulting in dirty data. This case is even more of a corner case than the previous one, so we will maintain the current implementation for now. In the future, we will clean up this dirty data through periodic garbage collection. The reason is that there can be at most one dirty record for each Iceberg sink, and it does not affect correctness.

src/bench/sink_bench/main.rs
```diff
@@ -377,13 +381,13 @@ where
     <S as risingwave_connector::sink::Sink>::Coordinator: std::marker::Send,
     <S as risingwave_connector::sink::Sink>::Coordinator: 'static,
 {
-    if let Ok(coordinator) = sink.new_coordinator().await {
+    if let Ok(coordinator) = sink.new_coordinator(DatabaseConnection::Disconnected).await {
```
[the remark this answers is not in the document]
wcy-fdu: The iceberg sink will only write to the system table in one case, that is, `is_exactly_once = true` is set in the create sink statement. Otherwise, this DatabaseConnection will not be used.

wcy-fdu:
Apply changes in #20772

wcy-fdu:
Testing, and I just found a corner case is not handled correctly, will fix it and explain it later.

wcy-fdu:
# Everything about testing(still updating)

The "exactly once" semantic guarantee ensures that the system maintains data consistency even in the event of a crash, meaning that data is neither lost nor duplicated. The prevention of data loss is primarily ensured by the persistence of the log store, while the prevention of duplication is guaranteed by various operations during the recovery phase in the processing layer.

Based on the current implementation, let's examine the complete timeline of an Iceberg commit.

Take the `commit in epoch 2` as an example

### Iceberg Commit Timeline

| time | Action                                                                                   |
|------|------------------------------------------------------------------------------------------|
|  0   | Collect metadata from various parallelism.                                              |
|  1   | Prepare 2pc: write pre-commit into system table.                                       |
|  2   | Do iceberg commit.                                                                       |
|  3   | Mark the record with `epoch = 2` in the system table as committed.                     |
|  4   | Delete the record with `epoch < 2` in the system table.      

### Recovery Action
The error can happen at any of the above points, let's see how to handle each case when recovery: 


| **Crash Scenario**                                   | **Actions**                                                                                     | **Log Store Operations**                      |
|-----------------------------------------------------|------------------------------------------------------------------------------------------------|-----------------------------------------------|
| **Crash before 0**                                  | Just rewind log store is fine.                                                                 | Rewind to truncate offset.                    |
| **Crash during (0, 1)**                             | Just rewind log store is fine.                                                                 | Rewind to truncate offset.                    |
| **Crash during (1, 2)**                             | Find record with `epoch = 2`, not in Iceberg (snapshot does not exist), committed is false. Re-commit this record. | Rewind to truncate offset.                    |
| **Crash during (2, 3)**                             | Find record with `epoch = 2` in system table and Iceberg, committed is false. Set `committed` to true. | Rewind from epoch 2 (consume data after epoch = 2). |
| **Crash during (3, 4)**                             | Find record with `epoch = 2` in system table and Iceberg, committed is false. Skip; this record will be deleted when committing a larger epoch later. | Rewind from epoch 2 (consume data after epoch = 2). |
| **Crash after 4**                                   | Do nothing since everything is fine.                                                           | Rewind from epoch 2 (consume data after epoch = 2). |

src/meta/model/src/exactly_once_iceberg_sink.rs
```diff
@@ -0,0 +1,35 @@
...
+use sea_orm::entity::prelude::*;
+use sea_orm::{DeriveEntityModel, DeriveRelation, EnumIter};
+
+use crate::{Epoch, SinkId};
+#[derive(Clone, Debug, PartialEq, DeriveEntityModel, Eq, Default)]
+#[sea_orm(table_name = "exactly_once_iceberg_sink_metadata")]
+pub struct Model {
+    #[sea_orm(primary_key, auto_increment = false)]
+    pub sink_id: SinkId,
+    #[sea_orm(primary_key, auto_increment = false)]
```
chenzl25, line 24: I think we can't define 2 primary keys here. cc @yezizp2012 Do you know how to define an index on a column with sea_orm?

src/connector/src/sink/iceberg/mod.rs
```diff
@@ -1315,14 +1416,75 @@ impl IcebergSinkCommitter {
 
 #[async_trait::async_trait]
 impl SinkCommitCoordinator for IcebergSinkCommitter {
-    async fn init(&mut self) -> Result<()> {
+    async fn init(&mut self) -> Result<Option<u64>> {
+        if self.is_exactly_once {
+            tracing::info!("Icberg sink coordinator initing");
```
chenzl25, line 1421: typo `Icberg`

src/connector/src/sink/iceberg/mod.rs
```diff
@@ -1315,14 +1416,75 @@ impl IcebergSinkCommitter {
...
+            if self
+                .iceberg_sink_has_pre_commit_metadata(&self.db, self.param.sink_id.sink_id())
+                .await?
+            {
+                let ordered_metadata_list_by_end_epoch = self
+                    .get_pre_commit_info_by_sink_id(&self.db, self.param.sink_id.sink_id())
+                    .await?;
+
+                let mut last_recommit_epoch = 0;
+                for (end_epoch, sealized_bytes, snapshot_id, committed) in
```
chenzl25, line 1431: typo `sealized_bytes`

src/connector/src/sink/iceberg/mod.rs
```diff
@@ -1315,14 +1416,75 @@ impl IcebergSinkCommitter {
...
+                    let mut write_results = vec![];
+
+                    for each in write_results_bytes {
+                        let write_result = IcebergCommitResult::try_from_sealized_bytes(each)?;
+                        write_results.push(write_result);
+                    }
+
+                    match (
+                        committed,
+                        self.is_snapshot_id_valid(&self.config, snapshot_id).await?,
```
chenzl25, line 1444: snapshot_id is always valid, maybe we should give it a better name.

src/connector/src/sink/iceberg/mod.rs
```diff
@@ -1315,14 +1416,75 @@ impl IcebergSinkCommitter {
...
+                                self.sink_id,
+                                end_epoch,
+                            )
+                            .await?;
+                        }
+                        (false, false) => {
+                            tracing::info!(
+                                "There are files that were not successfully committed; re-commit these files."
+                            );
+                            self.re_commit(end_epoch, write_results).await?;
```
chenzl25, line 1467: If the snapshot_id is not contained in iceberg table, we should use this snapshot id to commit, however, we didn't use it here.

src/connector/src/sink/iceberg/mod.rs
```diff
@@ -1378,6 +1573,36 @@ impl SinkCommitCoordinator for IcebergSinkCommitter {
...
+                snapshot_id,
+            )
+            .await?;
+        }
+        if self.is_exactly_once && is_normal_commit {
+            // tracing::info!(
+            //     "Finish write pre_commit data into system table, sleep 10min before commit into iceberg to meet crash."
+            // );
+            // tokio::time::sleep(Duration::from_secs(10 * 60)).await;
+        }
```
chenzl25, line 1604: Useless code

src/connector/src/sink/iceberg/mod.rs
```diff
@@ -1378,6 +1573,36 @@ impl SinkCommitCoordinator for IcebergSinkCommitter {
             .partition_type(schema)
             .map_err(|err| SinkError::Iceberg(anyhow!(err)))?;
 
+        let txn = Transaction::new(&self.table);
+        let snapshot_id = txn.generate_unique_snapshot_id();
```
chenzl25, line 1577: Only generate snapshot id when `is_normal_commit == true`, otherwise, we should use `snapshot_id` recorded in the system table.

src/meta/model/src/exactly_once_iceberg_sink.rs
yezizp2012: Well, I think you can create the table via migration locally and then run the following command to generate the model files:

`sea-orm-cli generate entity`

Check more details here: src/meta/model/src/README.md

yezizp2012: It appears that defining two primary keys for a model is feasible. An example was given here: https://www.sea-ql.org/SeaORM/docs/generate-entity/entity-structure/#composite-key

chenzl25: Yep. Thanks. We have checked the table definition from the meta store and it works.

src/connector/src/sink/iceberg/mod.rs
```diff
@@ -1315,14 +1415,77 @@ impl IcebergSinkCommitter {
...
+                    }
+
+                    match (
+                        committed,
+                        self.is_snapshot_id_in_iceberg(&self.config, snapshot_id)
+                            .await?,
+                    ) {
+                        (true, _) => {
+                            tracing::info!(
+                                "All data in log store has been written into external sink, skip"
```
chenzl25, line 1448: For all these logs, shall we add a sink_id to distinguish different sinks? Or should we change it to debug mode.

src/connector/src/sink/iceberg/mod.rs
```diff
@@ -1315,14 +1415,77 @@ impl IcebergSinkCommitter {
...
+                            );
+                        }
+                        (false, true) => {
+                            // skip
+                            tracing::info!(
```
chenzl25, line 1453: ditto

chenzl25 (approved):
Rest LGTM, thanks!

src/connector/src/sink/iceberg/mod.rs
wcy-fdu: Let's add `sink_id` first, and change it to debug mode after fully tested under the new framework.

cyliu0 (approved):

PR #19793, opened 2024-12-13
fix(connector): file scan use correct path

wcy-fdu:

https://github.com/risingwavelabs/risingwave/pull/19398 bring a bug that file_scan use incorrect path(With s3://bucket_name/ as prefix), so this pr fix this bug.

## Checklist

## Documentation

<details>

</details>


Cargo.lock
```diff
@@ -10510,56 +10510,88 @@ name = "risingwave_batch"
...
  "either",
- "foyer",
  "futures",
  "futures-async-stream",
  "futures-util",
- "hashbrown 0.14.5",
- "hytra",
  "iceberg",
  "itertools 0.13.0",
+ "linkme",
```
wcy-fdu, line 10522: Should not change Cargo.lock, will remove.

wcy-fdu:
todo: add e2e test

wcy-fdu:
Tess passed.

zwang28 (approved):

src/connector/src/source/iceberg/parquet_file_handler.rs
```diff
@@ -197,44 +198,39 @@ pub async fn list_s3_directory(
...
-        None => Ok(vec![]),
-    }
+                if is_parquet_schema_match_source_schema(arrow_data_type, rw_data_type) {
+                    Some(pos)
+                } else {
+                    None
+                }
+            })
+    })
+    .collect();
```
hzxa21, line 232: fmt?

hzxa21 (approved):
Rest LGTM. Should we add test cases for TVF file scan?

src/connector/src/source/iceberg/parquet_file_handler.rs
wcy-fdu: Strange, how did these lines escape `cargo fmt` escape?

wcy-fdu:
> Rest LGTM. Should we add test cases for TVF file scan?

I tried to add some test in CI, but found that currently tvf only support read file in s3, but not support read file in minio, see [definition of endpoint, it hard-coded into s3 instead of letting users pass it].(https://github.com/risingwavelabs/risingwave/blob/6f14e7937f3524a29ede0f7da4a3e71d8b65a3b9/src/connector/src/source/iceberg/parquet_file_handler.rs#L118-L121) .
Due to environmental issues, similar tests (parquet file source and file sink) are tested on minio.


I think we can add more backend such as 
```
file_scan(connector = minio)
file_scan(connector = gcs)
```
and then enhance ci test. This will take some time but this fix is a bit urgent, so I just test it locally and leave these in subsequent PRs.

PR #19797, opened 2024-12-13
fix(frontend): fix row filter with pk diff order type

xxhZs:

## Checklist

## Documentation

<details>

</details>


src/frontend/planner_test/tests/testdata/output/row_filter.yaml
```diff
@@ -10,14 +10,14 @@
     select * from t where (v1,v2) > (1,2);
   batch_plan: |-
     BatchExchange { order: [], dist: Single }
-    └─BatchScan { table: t, columns: [t.v1, t.v2, t.v3], scan_ranges: [(t.v1, t.v2) > (Int32(1), Int32(2))], distribution: UpstreamHashShard(t.v1, t.v2, t.v3) }
+    └─BatchScan { table: t, columns: [t.v1, t.v2, t.v3], scan_ranges: [(t.v1, t.v2) >= (Int32(1), Int32(2))], distribution: UpstreamHashShard(t.v1, t.v2, t.v3) }
```
hzxa21, line 13: In addition to planner test, let's add e2e tests for the issues fixed by this PR as well to verify the result is correct.

src/frontend/src/utils/condition.rs
```diff
@@ -415,35 +415,61 @@ impl Condition {
...
                         }
+                        match order_type {
+                            Some(o) => {
+                                if o != column_order.order_type {
+                                    all_added = false;
+                                    break;
+                                }
+                            }
+                            None => order_type = Some(column_order.order_type),
+                        }
```
hzxa21, line 435: I think the check here will pass if the PKs are all ordered desc:
```
create materialized view mv1 as select * from t1 order by v1 desc, v2 desc, v3 desc;
```

However, in this case, I think the optimization is not applicable.

xxhZs: We just need to make sure that the order type of all the pk's is the same, in the row_seq_scan_exec will generate the appropriate scan according to the order type of the first pk to scan the data.
https://github.com/risingwavelabs/risingwave/blob/fe84f69e1611dad37f0ea97ff9dcdcbbaa972b63/src/batch/executors/src/executor/row_seq_scan.rs#L438

e2e_test/batch/basic/row_filter.slt.part
```diff
@@ -0,0 +1,65 @@
+statement ok
+create table t1(v1 int, v2 int, v3 int);
+
+statement ok
+insert into t1 values( 1,1,1),( 1,2,1),( 1,2,2),(1,3,1),(1,3,2),(1,3,3);
```
hzxa21, line 5: nits: fmt

src/frontend/src/utils/condition.rs
hzxa21: Got it. That means when all pks are ordered desc, with the following scan range:
```
ExprType::GreaterThan => {
    (Bound::Included(pk_struct), Bound::Unbounded)
}
```

the storage scan will be done by scanning from `Bound::Unbounded` to `Bound::Included(pk_struct)`

hzxa21 (approved):
LGTM

chenzl25 (approved):
LGTM

Issue #19681, opened 2024-12-04
Some tables doesn't sync for PG CDC v2.1.0-rt.8


cyliu0:
### Describe the bug

https://buildkite.com/risingwave-test/ch-benchmark-pg-cdc-shared-source/builds/124#019391f2-6d26-4d75-bcc4-ab39abc277cb/8586-8587

https://grafana.test.risingwave-cloud.xyz/d/EpkBw5W4k/risingwave-dev-dashboard?orgId=1&from=1733320258820&to=1733323410627&var-datasource=cdtasocg64074c&var-namespace=ch-benchmark-pg-cdc-shared-source-6ffck&var-instance=benchmark-risingwave&var-pod=All&var-component=All&var-table=All

```
{
  "consistent": false,
  "table-compare-results": [
    {
      "consistent": false,
      "table-checksums": [
        {
          "url": "[REDACTED]://[REDACTED]:[REDACTED]@benchmark-[REDACTED]ql:5432/[REDACTED]",
          "table-name": "customer",
          "table-checksum": -5419689675690500630,
          "table-rows": 4200000
        },
        {
          "url": "[REDACTED]://root:@benchmark-risingwave-frontend:4567/dev",
          "table-name": "customer",
          "table-checksum": 0,
          "table-rows": 0
        }
      ]
    },
    {
      "consistent": false,
      "table-checksums": [
        {
          "url": "[REDACTED]://[REDACTED]:[REDACTED]@benchmark-[REDACTED]ql:5432/[REDACTED]",
          "table-name": "district",
          "table-checksum": -1426567335868077560,
          "table-rows": 1400
        },
        {
          "url": "[REDACTED]://root:@benchmark-risingwave-frontend:4567/dev",
          "table-name": "district",
          "table-checksum": 0,
          "table-rows": 0
        }
      ]
    },
    {
      "consistent": false,
      "table-checksums": [
        {
          "url": "[REDACTED]://[REDACTED]:[REDACTED]@benchmark-[REDACTED]ql:5432/[REDACTED]",
          "table-name": "history",
          "table-checksum": 7137233099402833883,
          "table-rows": 4234095
        },
        {
          "url": "[REDACTED]://root:@benchmark-risingwave-frontend:4567/dev",
          "table-name": "history",
          "table-checksum": 0,
          "table-rows": 0
        }
      ]
    },
    {
      "consistent": true,
      "table-checksums": [
        {
[...]

PR #19805, opened 2024-12-16
feat(source): create source stream with retry during recovery

StrikeW:

Previously we reverted the retry for source executor #19754. This PR we re-introduce the retry mechanism but only during recovery process. Because when source job create in the first time, it should fail if cannot reach the upstream system.

And we merge the patch https://github.com/risingwavelabs/risingwave/pull/19714 into this PR

passed main-cron https://buildkite.com/risingwavelabs/main-cron/builds/4092
close https://github.com/risingwavelabs/risingwave/issues/19681

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/PostgresValidator.java
+++ b/java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/PostgresValidator.java
@@ -363,6 +363,13 @@ public class PostgresValidator extends DatabaseValidator implements AutoCloseabl
         // If the source properties is created by share source, skip the following
         // check of publication
         if (isCdcSourceJob) {
+            if (!isPublicationExists) {
+                LOG.info(
+                        "creating cdc source job: publication '{}' doesn't exist, creating...",
+                        pubName);
+                DbzSourceUtils.createPostgresPublicationInValidate(userProps);
+                LOG.info("creating cdc source job: publication '{}' created successfully", pubName);
+            }
             return;
         }
 
--- a/java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/PostgresValidator.java
+++ b/java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/PostgresValidator.java
@@ -354,17 +354,10 @@ public class PostgresValidator extends DatabaseValidator implements AutoCloseabl
             }
         }
 
-        if (!isPublicationExists) {
+        if (!isPublicationExists && !pubAutoCreate) {
             // We require a publication on upstream to publish table cdc events
-            if (!pubAutoCreate) {
-                throw ValidatorUtils.invalidArgument(
-                        "Publication '" + pubName + "' doesn't exist and auto create is disabled");
-            } else {
-                // createPublicationIfNeeded(Optional.empty());
-                LOG.info(
-                        "Publication '{}' doesn't exist, will be created in the process of streaming job.",
-                        this.pubName);
-            }
+            throw ValidatorUtils.invalidArgument(
+                    "Publication '" + pubName + "' doesn't exist and auto create is disabled");
         }
 
         // If the source properties is created by share source, skip the following
--- a/src/stream/src/executor/source/source_executor.rs
+++ b/src/stream/src/executor/source/source_executor.rs
@@ -532,7 +532,7 @@ impl<S: StateStore> SourceExecutor<S> {
                     // so don't waste work on it.
                     // For more details, see https://github.com/risingwavelabs/risingwave/issues/16576#issuecomment-2095413297
                     // Note that shared CDC source is special. It already starts from latest.
-                    self.is_shared_non_cdc && is_uninitialized,
+                    self.is_shared_non_cdc,
                 )
                 .instrument_await("source_build_reader")
                 .await?;
@@ -559,7 +559,7 @@ impl<S: StateStore> SourceExecutor<S> {
                             recover_state.clone(),
                             column_ids.clone(),
                             source_ctx.clone(),
-                            false,  // not need to seek to latest since source is initialized
+                            false,  // not need to seek to latest since source state is initialized
                         )
                         .await {
                         Ok((stream, latest_splits)) => Ok((stream, latest_splits)),
--- a/src/stream/src/executor/source/source_executor.rs
+++ b/src/stream/src/executor/source/source_executor.rs
@@ -629,7 +629,6 @@ impl<S: StateStore> SourceExecutor<S> {
                 apply_rate_limit(source_chunk_reader, self.rate_limit_rps)
                     .boxed()
                     .map_err(StreamExecutorError::connector_error),
-                // source_chunk_reader.map_err(StreamExecutorError::connector_error),
                 latest_splits,
             )
         };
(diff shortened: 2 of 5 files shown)


xxchan (commented):
We may remove the workaround here to get it tested.

https://github.com/risingwavelabs/risingwave/blob/1e24ca482c29c08ba3aeb31ee44f672e3eaaef05/e2e_test/source_inline/kafka/issue_19563.slt.serial#L49-L51

BTW, is there any test case covered the CDC sync issue #19681?

xxchan (commented):
Code LGTM

StrikeW:
> We may remove the workaround here to get it tested.
> 
> https://github.com/risingwavelabs/risingwave/blob/1e24ca482c29c08ba3aeb31ee44f672e3eaaef05/e2e_test/source_inline/kafka/issue_19563.slt.serial#L49-L51
> 
> BTW, is there any test case covered the CDC sync issue #19681?

I am testing #19681 with the test pipeline https://buildkite.com/risingwave-test/ch-benchmark-pg-cdc-shared-source

src/stream/src/executor/source/source_executor.rs
```diff
@@ -500,20 +516,133 @@ impl<S: StateStore> SourceExecutor<S> {
...
+            let (source_chunk_reader, latest_splits) = self
+                .build_stream_source_reader(
+                    &source_desc,
+                    recover_state,
+                    // For shared source, we start from latest and let the downstream SourceBackfillExecutors to read historical data.
+                    // It's highly probable that the work of scanning historical data cannot be shared,
+                    // so don't waste work on it.
+                    // For more details, see https://github.com/risingwavelabs/risingwave/issues/16576#issuecomment-2095413297
+                    // Note that shared CDC source is special. It already starts from latest.
+                    self.is_shared_non_cdc && is_uninitialized,
```
hzxa21, line 535: nits: `is_uninitialized` must be true here so we don't need to check it.

src/stream/src/executor/source/source_executor.rs
```diff
@@ -500,20 +516,133 @@ impl<S: StateStore> SourceExecutor<S> {
...
-                self.is_shared_non_cdc && is_uninitialized,
+
+        let mut received_resume_during_build = false;
+        let mut barrier_stream = barrier_to_message_stream(barrier_receiver).boxed();
+
+        // Build the source stream reader.
+        let (source_chunk_reader, latest_splits) = if is_uninitialized {
+            tracing::info!("source uninitialized, build source stream reader w/o retry.");
+            let (source_chunk_reader, latest_splits) = self
+                .build_stream_source_reader(
```
hzxa21, line 527: Inside `build_stream_source_reader` we already called `apply_rate_limit` while in L639 we called `apply_rate_limit` again. Will this be a problem?

src/stream/src/executor/source/source_executor.rs
```diff
@@ -500,20 +516,133 @@ impl<S: StateStore> SourceExecutor<S> {
...
             )
-            .instrument_await("source_build_reader")
-            .await?;
-        let source_chunk_reader = source_chunk_reader.map_err(StreamExecutorError::connector_error);
+        } else {
+            tracing::info!("source initialized, build source stream reader with retry.");
+            // Build the source stream reader with retry during recovery.
+            // We only build source stream reader with retry during recovery,
+            // because we can rely on the persisted source states to recover the source stream
+            // and can avoid the potential race with "seek to latest"
```
hzxa21, line 548: Better include the issue link so that people know what the issue is with "seek to latest".

src/stream/src/executor/source/source_executor.rs
```diff
@@ -492,6 +499,15 @@ impl<S: StateStore> SourceExecutor<S> {
...
         }
 
+        // Since DDL always accompany with a checkpoint barrier, we can assume that
+        // if initial_dispatch_num != 0, the source state must be initialized.
+        if self.actor_ctx.initial_dispatch_num != 0 {
+            assert!(
+                !is_uninitialized,
+                "is_uninitialized must be false if initial_dispatch_num != 0"
+            );
+        }
```
BugenZhao, line 509: I believe the assertion here is no-op.

StrikeW: agree. `is_uninitialized` is inited with `self.actor_ctx.initial_dispatch_num == 0`

src/stream/src/executor/source/source_executor.rs
StrikeW: These are two different code paths, L639 is to apply rate limit if in recovery process.

StrikeW:
> > We may remove the workaround here to get it tested.
> > https://github.com/risingwavelabs/risingwave/blob/1e24ca482c29c08ba3aeb31ee44f672e3eaaef05/e2e_test/source_inline/kafka/issue_19563.slt.serial#L49-L51
> > 
> > BTW, is there any test case covered the CDC sync issue #19681?
> 
> I am testing #19681 with the test pipeline https://buildkite.com/risingwave-test/ch-benchmark-pg-cdc-shared-source

Test passed https://buildkite.com/risingwave-test/ch-benchmark-pg-cdc-shared-source/builds/133

java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/PostgresValidator.java
```diff
@@ -354,22 +354,22 @@ private void validatePublicationConfig(boolean isSuperUser) throws SQLException
...
         }
 
         // If the source properties is created by share source, skip the following
         // check of publication
         if (isCdcSourceJob) {
+            if (!isPublicationExists) {
+                LOG.info(
+                        "creating cdc source job: publication '{}' doesn't exist, creating...",
+                        pubName);
+                DbzSourceUtils.createPostgresPublicationInValidate(userProps);
```
hzxa21, line 370: Why do we need to create the publication in meta instead of doing it in the source reader? I thought the CREATE SOURCE logic is unchanged (block the 1st barrier until source reader creation succeeds) before and after this PR.

hzxa21:
> > > We may remove the workaround here to get it tested.
> > > https://github.com/risingwavelabs/risingwave/blob/1e24ca482c29c08ba3aeb31ee44f672e3eaaef05/e2e_test/source_inline/kafka/issue_19563.slt.serial#L49-L51
> > > 
> > > BTW, is there any test case covered the CDC sync issue #19681?
> > 
> > 
> > I am testing #19681 with the test pipeline https://buildkite.com/risingwave-test/ch-benchmark-pg-cdc-shared-source
> 
> Test passed https://buildkite.com/risingwave-test/ch-benchmark-pg-cdc-shared-source/builds/133

Will the ch-benchmark test the case when upstream is unavailable? If not, I think we still in lack of a test case for that.

java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/PostgresValidator.java
StrikeW: You can see in the `source_executor.rs`, **we forward the initial barrier first before we create the source reader**. IIRC, after Meta collected the initial barrier the `CREATE SOURCE` will return success, and then following `CREATE TABLE` command can be issued to Meta. So during validation of `CREATE TABLE`, the pg publication may not have been created.

StrikeW: I think this race exist for a long time ago, not the new problem introduced by this PR.

hzxa21 (approved):
LGTM

PR #19822, opened 2024-12-17
feat: jdbc sink user password

### LINKED ISSUE #19766: bug: error reporting accidentally leaked the secret
### Describe the bug

As title.

Perhaps we should use `getConnection(String url, String user, String password)` rather than putting username and password in url.

### Error message/log

```text
1: gRPC request to meta service failed: Internal error
2: failed to validate sink
3: sink cannot pass validation: INVALID_ARGUMENT: failed
to connect to target database: 08001: No suitable driver
found for postgresql://xxxx:xxxx@xxxx:xxxx/xxxx?user=xxxx&password=xxxx
```

### To Reproduce

Create a JDBC Postgres Sink.

_No response_

### The version of RisingWave

_No response_

### Additional context

_No response_

fuyufjh:

Fix #19766.

The root cause of #19766 is that JDBC URL is included in many JDBC Driver's error message, which is okay. But we put username and password in that URL, which is not a good practice.

This PR uses `getConnection(jdbcUrl, user, password)` instead of `getConnection(jdbcUrl)`, so that the error message won't show the password.

No breaking change - the old syntax can work as well.

## Checklist

## Documentation

Need to update the docs of JDBC sink to use the new format.

No need to add release note - it's just a minor change.

<details>

</details>


java/connector-node/risingwave-sink-jdbc/src/main/java/com/risingwave/connector/JDBCSinkFactory.java
```diff
@@ -56,7 +56,15 @@ public void validate(
...
-        try (Connection conn = DriverManager.getConnection(jdbcUrl);
+        java.util.Properties info = new java.util.Properties();
+        if (config.getUser() != null) {
+            info.put("user", config.getUser());
+        }
+        if (config.getPassword() != null) {
+            info.put("password", config.getPassword());
+        }
+
+        try (Connection conn = DriverManager.getConnection(jdbcUrl, info);
```
StrikeW, line 67: I think you can directly use `Connection getConnection(String url, String user, String password)` and it has null check inside.

fuyufjh: Yeah, that's my original code (as written in the PR description). Here I use `getConnection(jdbcUrl, info)` is to unify the function call in this `try (...)`,

StrikeW: Don’t get it.😂 It seems `Connection conn = DriverManager.getConnection(jdbcUrl, config.getUser(), config.getPassword())`. correct me if I am wrong.

StrikeW (approved):
Rest lgtm

java/connector-node/risingwave-sink-jdbc/src/main/java/com/risingwave/connector/JDBCSinkFactory.java
fuyufjh: Make sense. Fixed.

PR #19827, opened 2024-12-17
feat(types): support ns timestamp

xxhZs:

our timestamp only support ms, But our internal implementation is capable of ns, it's just that we automatically convert it to ms when we read it in, and this pr can read ns to support ns.

## Checklist

## Documentation

<details>

</details>


src/common/src/types/datetime.rs
```diff
@@ -495,18 +491,24 @@ impl Timestamp {
...
-        output
-            .write(&(self.0.and_utc().timestamp_micros()).to_be_bytes())
-            .map_err(Into::into)
+        let timestamp_size = output
+            .write(&(self.0.and_utc().timestamp()).to_be_bytes())
+            .map_err(Into::<ArrayError>::into)?;
+        let timestamp_subsec_nanos_size = output
+            .write(&(self.0.and_utc().timestamp_subsec_nanos()).to_be_bytes())
+            .map_err(Into::<ArrayError>::into)?;
+        Ok(timestamp_subsec_nanos_size + timestamp_size)
```
fuyufjh, line 511: This will break the compatibility of Timestamp stored in protobuf. Not sure whether it's used somewhere in our catalog.

fuyufjh: One of the ways to keep **backward** compatibility is: using a special value such as `-1` as a placeholder indicating the new format. That is

- If read `-1`, parse the subsequent 12 bytes in new format
- If read otherwise, consider it as a timestamp_micros (old format)

wcy-fdu:
I tested whether I could read nanosecond in the parquet file on this branch, and it seemed to work.
image

wcy-fdu:
sorry, mis clicked update branch

wcy-fdu:
cc @xiangjinwu  PTAL❤️

src/common/src/types/datetime.rs
```diff
@@ -168,29 +168,18 @@ impl FromStr for Timestamp {
...
-        } else {
-            let res =
-                speedate::Date::parse_str_rfc3339(s).map_err(|_| ErrorKind::ParseTimestamp)?;
-            Ok(
-                Date::from_ymd_uncheck(res.year as i32, res.month as u32, res.day as u32)
-                    .and_hms_micro_uncheck(0, 0, 0, 0),
-            )
-        }
+        let dt = s
+            .parse::<jiff::civil::DateTime>()
```
xiangjinwu, line 172: Why do we need to introduce yet another dependency in addition to `chrono` and `speedate`?

src/common/src/types/datetime.rs
```diff
@@ -495,18 +491,47 @@ impl Timestamp {
     }
 
     pub fn from_protobuf(cur: &mut Cursor<&[u8]>) -> ArrayResult<Timestamp> {
-        let micros = cur
+        let secs = cur
```
xiangjinwu, line 494: nit: `secs_or_usecs`

src/common/src/types/datetime.rs
```diff
@@ -495,18 +491,47 @@ impl Timestamp {
...
+        value ^ (0b01 << 62)
+    }
+
+    fn has_timestamp_namo_format_state(value: i64) -> bool {
+        let state = (value >> 62) & 0b11;
+        state == 0b10 || state == 0b01
+    }
+
+    fn remove_timestamp_namo_format_state(value: i64) -> i64 {
+        value ^ (0b01 << 62)
```
xiangjinwu, line 534: Encapsulate as a new type:
```
/// document about old and new format, including the meaning of highest 2 bits, and the corresponding accepted ranges.
enum FirstI64 {
  V0{ usecs: i64 }
  V1{ secs: i64 }
}
impl From<i64> for FirstI64 {
  // check and remove the tagging bits
}
impl From<FirstI64> for i64 {
  // add the tagging bits
}
```

src/storage/hummock_sdk/src/compaction_group/hummock_version_ext.rs
```diff
@@ -495,7 +495,7 @@ impl HummockVersion {
                             .clone_from(&group_construct.table_ids);
                         self.levels.insert(*compaction_group_id, new_levels);
                         let member_table_ids = if group_construct.version
-                            >= CompatibilityVersion::NoMemberTableIds as _
+                            >= CompatibilityVersion::NoMemberTableIds as i32
```
xiangjinwu, line 498: avoid unrelated change

xiangjinwu (commented):
* Could you also include the testing SQLs you have tried?
* Is any caller of `to_protobuf`/`from_protobuf` assuming that `Timestamp` would be fixed 64-bit? This assumption is no longer true.

xxhZs:
> * Could you also include the testing SQLs you have tried?

add ci

src/common/src/types/datetime.rs
xxhZs: Because chrono can't be parsed, and speedate is only parsed at the us level

src/storage/hummock_sdk/src/compaction_group/hummock_version_ext.rs
xiangjinwu: It seems we do need this as long as we use `jiff` in `src/common/src/types/datetime.rs`. Not sure why though ... maybe a rustc bug.

Btw it can be `group_construct.version() >= CompatibilityVersion::NoMemberTableIds` rather than `if group_construct.version >= CompatibilityVersion::NoMemberTableIds as i32`

e2e_test/batch/types/timestamp_ns.slt.part
```diff
@@ -0,0 +1,112 @@
...
+1  2013-01-01 01:01:01.123456+00:00
+4  2213-01-01 01:01:01.123456+00:00
+2  2012-01-01 01:01:01.123456+00:00
+
+query T
+select v1, to_char(v2, 'YYYY-MM-DD HH24:MI:SS.NS') as formatted_v2 from t1;
+----
+3  0001-01-01 01:01:01.123456789 BC
+5  null
+1  2013-01-01 01:01:01.123456000
```
xiangjinwu, line 101: Did this case pass? Is it supposed to be `2013-01-01 01:01:01.123456789`?

src/storage/hummock_sdk/src/compaction_group/hummock_version_ext.rs
xxhZs: It doesn't seem to work because the type of CompatibilityVersion::NoMemberTableIds is not i32

xiangjinwu: `group_construct.version()` returns `CompatibilityVersion`, while `group_construct.version` is `i32`

xiangjinwu:
Btw you can fix the e2e test with `rowsort`.

https://www.sqlite.org/sqllogictest/doc/trunk/about.wiki
> The default is "nosort". In nosort mode, the results appear in exactly the order in which they were received from the database engine. The nosort mode should only be used on queries that have an ORDER BY clause or which only have a single row of result, since otherwise the order of results is undefined and might vary from one database engine to another. The "rowsort" mode gathers all output from the database engine then sorts it by rows on the client side.

#6776
> If there's `rowsort`, slt file will contain sorted results.

xiangjinwu:
FYI regarding your latest e2e failure, it is actually a bug in our existing `to_char` implementation:

`1 BC` in PostgreSQL is year `0` in chrono. Although chrono also provides `DateLike::year_ce` method, its `strftime` module does not provide a specifier to use `year_ce` instead of `year`.

Given it is out of scope of this PR, let's just replace the case with one without `BC`.

src/common/src/types/datetime.rs
```diff
@@ -168,29 +168,18 @@ impl FromStr for Timestamp {
...
-            let res =
-                speedate::Date::parse_str_rfc3339(s).map_err(|_| ErrorKind::ParseTimestamp)?;
-            Ok(
-                Date::from_ymd_uncheck(res.year as i32, res.month as u32, res.day as u32)
-                    .and_hms_micro_uncheck(0, 0, 0, 0),
-            )
-        }
+        let dt = s
+            .parse::<jiff::civil::DateTime>()
+            .map_err(|_| ErrorKind::ParseTimestamp)?;
```
xiangjinwu, line 173: Error message of `ParseTimestamp` shall be updated from _up to **6** digits_ to _up to **9** digits_.

xiangjinwu (approved):
Mostly looks good to me. Just more to test:
* `distinct` / `group by` / `order by` / `max`
* `extract` with `epoch` / `second` / `milliseconds` / `microseconds `
* `make_timestamp`

wcy-fdu:
Thanks for the efforts. I will doing some testing on sink side later.

xiangjinwu:
> * Is any caller of `to_protobuf`/`from_protobuf` assuming that `Timestamp` would be fixed 64-bit? This assumption is no longer true.

Answering my own question: there is but it is only used as a hint - so okay to be inaccurate.

`to_protobuf` is used here. The `size_of::<T>` may result in an inaccurate estimate for buffer capacity.

https://github.com/risingwavelabs/risingwave/blob/05397f91957f9ca9ed7c893ef6b667faf1bae4a0/src/common/src/array/primitive_array.rs#L232-L236

`from_protobuf` is used here. Movement of `Cursor` makes no assumption on item sizes.

https://github.com/risingwavelabs/risingwave/blob/05397f91957f9ca9ed7c893ef6b667faf1bae4a0/src/common/src/array/proto_reader.rs#L69-L76

src/expr/impl/src/scalar/extract.rs
```diff
@@ -42,17 +42,18 @@ fn extract_from_datelike(date: impl Datelike, unit: Unit) -> Decimal {
 
 /// Extract field from `Timelike`.
 fn extract_from_timelike(time: impl Timelike, unit: Unit) -> Decimal {
-    let usecs = || time.second() as u64 * 1_000_000 + (time.nanosecond() / 1000) as u64;
+    let usecs = || time.second() as u64 * 1_000_000_000 + time.nanosecond() as u64;
```
xiangjinwu, line 45:
```suggestion
    let nanos = || time.second() as u64 * 1_000_000_000 + time.nanosecond() as u64;
```

pjpringle:
Nice useful feature for trading use cases

Issue #19736, opened 2024-12-10
bug: panic when adding new field in schema registry proto definition


fuyufjh:
### Describe the bug

After updating the schema in schema registry and run `REFRESH` from RisingWave, the compute nodes start crash loop

Possible reason: Currently, RisingWave doesn’t support alter a struct column i.e. adding a new field inside the struct column. If doing this with SQL, it will return an “unsupported” error. But this schema change caused exactly this kind of column schema change, and RW fails to handle it correctly.

### Error message/log

```text
22:     0x5ad9a35c8cd2 - risingwave_common::util::value_encoding::inner_deserialize_datum::h5d12937549c43ae1
                               at /risingwave/src/common/src/util/value_encoding/mod.rs:195:20
  23:     0x5ad9a35c8cd2 - risingwave_common::util::value_encoding::deserialize_struct::h9d502dba455bb3c1
                               at /risingwave/src/common/src/util/value_encoding/mod.rs:371:27
  24:     0x5ad9a35c8cd2 - risingwave_common::util::value_encoding::deserialize_value::hf4b8c836e0f09fb4
                               at /risingwave/src/common/src/util/value_encoding/mod.rs:357:41
  25:     0x5ad9a35a1306 - <risingwave_common::util::value_encoding::column_aware_row_encoding::Deserializer as risingwave_common::util::value_encoding::ValueRowDeserializer>::deserialize::hc70588ccd8708ede
                               at /risingwave/src/common/src/util/value_encoding/column_aware_row_encoding.rs:232:30
  26:     0x5ad9abd7c00d - <risingwave_common::util::value_encoding::column_aware_row_encoding::ColumnAwareSerde as risingwave_common::util::value_encoding::ValueRowDeserializer>::deserialize::ha0cde7b4168459e3
                               at /risingwave/src/common/src/util/value_encoding/column_aware_row_encoding.rs:278:9
  27:     0x5ad9abd7c00d - risingwave_stream::common::table::state_table::StateTableInner<S,SD,_,_>::get_row::{{closure}}::hfc439b1e9f62e24f
                               at /risingwave/src/stream/src/common/table/state_table.rs:818:27
  28:     0x5ad9abd7c00d - risingwave_stream::common::table::state_table::StateTableInner<S,SD,_,_>::get_compacted_row::{{closure}}::hc7a971c6f3feee86
                               at /risingwave/src/stream/src/common/table/state_table.rs:877:18
[...]

Issue #17128, opened 2024-06-05
when using external schema, `struct`'s fields are not shown in `describe <table>`


xxchan:
when using external schema, `struct`'s fields are not shown in `describe <table>`

https://github.com/risingwavelabs/risingwave/pull/17110/files#r1627299710

example:
```
query
describe kafka_json_schema_plain;
----
dimensions (empty) false NULL
map jsonb false NULL
notMap (empty) false NULL
price double precision false NULL
productId bigint false NULL
productName character varying false NULL
tags character varying[] false NULL
_row_id serial true NULL
primary key _row_id NULL NULL
distribution key _row_id NULL NULL
table description kafka_json_schema_plain NULL NULL
```
There should be `dimenstions.xx` and `notMap.xx`. BTW, maybe we can also show `struct` instead of `null` for them.

xxchan:
It seems this is because `ColumnDesc` have too many constructors. And only some of them has `type_name`

xxchan:
related #13104

xxchan:
More specifically, `ColumnDesc.field_descs` is not set.

PR #19828, opened 2024-12-17
fix(frontend): check data type in column id generator

BugenZhao:

When generating column ID for existing columns (by name), also check and reject if the data type changes. Otherwise, the persisted data will be incompatible and lead to undefined behavior.

Note that for struct types, the nested fields are encoded with value-encoding (instead of column-aware encoding) under current implementation. As a result, adding new nested fields cannot be supported and should be rejected as well. Close https://github.com/risingwavelabs/risingwave/issues/19736.

This won't be hit for a regular table as we don't provide syntax support for altering the data type for an existing column. However, schema registries may consider this as a compatible change so it can be hit by `ALTER TABLE REFRESH SCHEMA`.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/handler/create_table.rs
+++ b/src/frontend/src/handler/create_table.rs
@@ -23,13 +23,14 @@ use fixedbitset::FixedBitSet;
 use itertools::Itertools;
 use pgwire::pg_response::{PgResponse, StatementType};
 use risingwave_common::catalog::{
-    CdcTableDesc, ColumnCatalog, ColumnDesc, Engine, TableId, TableVersionId, DEFAULT_SCHEMA_NAME,
-    INITIAL_TABLE_VERSION_ID, RISINGWAVE_ICEBERG_ROW_ID, ROWID_PREFIX,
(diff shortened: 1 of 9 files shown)


src/frontend/src/handler/create_table.rs
```diff
@@ -1985,16 +2025,41 @@ mod tests {
                     ),
                     is_hidden: false,
                 },
+                ColumnCatalog {
+                    column_desc: ColumnDesc::from_field_with_column_id(
+                        &Field::with_name(
+                            StructType::new([("f1", DataType::Int32)]).into(),
+                            "nested",
+                        ),
```
xiangjinwu, line 2033: Yet another example of why `Field::sub_fields` (and `ColumnDesc::field_descs`) shall be [consolidated](https://github.com/risingwavelabs/risingwave/compare/main...refactor-common-fields-and-struct-WIP-r1) into `StructType`. Currently `Field::with_name` on `struct` produces incomplete results - and there is a special `Field::with_struct`.

No need to fix the usage in this PR as the incomplete results are NEVER used except `describe` #17128

PR #19834, opened 2024-12-17
feat: support expression in limit clause

### LINKED ISSUE #19830: Need to support expression evaluation after LIMIT

Risingwave now cannot deal with expression after `LIMIT`.
For example, for query below:
```
create table test_table (id int);
select * from test_table LIMIT 0 + 1;
```
There is sql parse error.

Supporting expression evaluation after keyword `LIMIT` would be helpful.

_No response_

### Additional context

_No response_

lmatz:

fix #19830 

required by SQLMesh's CI test

The fix is to evaluate the expression during binding stage so that we either accept the result (it is a number >=0 after casting to bigint) or reject it (all the other cases)

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/e2e_test/batch/order/test_limit.slt.part
+++ b/e2e_test/batch/order/test_limit.slt.part
@@ -182,6 +182,45 @@ INSERT INTO integers VALUES (1), (2), (3), (4), (5);
 # 4
 # 5
 
+# expression in the limit clause
+query I
+SELECT * FROM generate_series(0,10,1) as t(v) order by v limit 1+2;
+----
+0
+1
+2
+
+query I
+SELECT * FROM generate_series(0,10,1) as t(v) order by v limit 1.5*2;
+----
+0
+1
+2
+
+query I
+SELECT * FROM generate_series(0,10,1) as t(v) order by v limit -3+4;
+----
+0
+
+query I
+SELECT * FROM generate_series(0,10,1) as t(v) order by v limit 1.5*2 OFFSET 2;
+----
+2
+3
+4
+
+query I
+SELECT * FROM generate_series(0,3,1) as t(v) order by v limit NULL + 1;
+----
+0
+1
+2
+
+query I
+SELECT * FROM generate_series(0,3,1) as t(v) order by v limit 1.1::Decimal;
+----
+0
+
 
 # Subqueries that return negative values
 statement error
--- a/e2e_test/batch/order/test_limit.slt.part
+++ b/e2e_test/batch/order/test_limit.slt.part
@@ -215,12 +215,43 @@ SELECT * FROM generate_series(0,3,1) as t(v) order by v limit NULL + 1;
 0
 1
 2
+3
 
 query I
-SELECT * FROM generate_series(0,3,1) as t(v) order by v limit 1.1::Decimal;
+SELECT * FROM generate_series(0,3,1) as t(v) order by v limit NULL + '1'::bigint;
 ----
 0
+1
+2
+3
 
+statement error
+SELECT * FROM generate_series(0,3,1) as t(v) order by v limit NULL + '1';
+
+statement error
+SELECT * FROM generate_series(0,3,1) as t(v) order by v limit NULL + '1'::jsonb;
+
+query I
+SELECT * FROM generate_series(0,3,1) as t(v) order by v limit 2.1::Decimal;
+----
+0
+1
+
+statement error
+SELECT * FROM generate_series(0,3,1) as t(v) order by v limit '1'::jsonb;
+
+
+query I
+SELECT * FROM generate_series(0,3,1) as t(v) order by v limit '2';
+----
+0
+1
+
+statement error
+SELECT * FROM generate_series(0,3,1) as t(v) order by v limit '-2';
+
+statement error
+SELECT * FROM generate_series(0,3,1) as t(v) order by v limit '2.2';
 
 # Subqueries that return negative values
 statement error
--- a/e2e_test/batch/order/test_limit.slt.part
+++ b/e2e_test/batch/order/test_limit.slt.part
@@ -291,7 +291,7 @@ db error: ERROR: Failed to run the query
 Caused by these errors (recent errors listed first):
   1: Expr error
   2: expects an integer or expression that can be evaluated to an integer after LIMIT,
-but the evaluation of the expression returns error:Expr error: error while evaluating expression `str_parse('2.2')`
+but the evaluation of the expression returns error:Expr error: error while evaluating expression `str_parse('2.2')`: Parse error: bigint invalid digit found in string
 
 
 statement error
--- a/e2e_test/batch/order/test_limit.slt.part
+++ b/e2e_test/batch/order/test_limit.slt.part
@@ -273,7 +273,7 @@ SELECT * FROM generate_series(0,3,1) as t(v) order by v limit '2';
 1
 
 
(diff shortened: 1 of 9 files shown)


src/sqlparser/src/ast/query.rs
```diff
@@ -655,7 +655,7 @@ impl fmt::Display for OrderByExpr {
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub struct Fetch {
     pub with_ties: bool,
-    pub quantity: Option<String>,
+    pub quantity: Option<Expr>,
```
xiangjinwu, line 658: Not as simple as it looks like. A parsing conflict can happen:

https://github.com/postgres/postgres/blob/REL_17_2/src/backend/parser/gram.y#L13233
> ```
> * Allowing full expressions without parentheses causes various parsing
> * problems with the trailing ROW/ROWS key words.  SQL spec only calls for
> * <simple value specification>, which is either a literal or a parameter (but
> * an <SQL parameter reference> could be an identifier, bringing up conflicts
> * with ROW/ROWS). We solve this by leveraging the presence of ONLY (see above)
> * to determine whether the expression is missing rather than trying to make it
> * optional in this rule.
>
> * c_expr covers almost all the spec-required cases (and more), but it doesn't
> * cover signed numeric literals, which are allowed by the spec. So we include
> * those here explicitly. We need FCONST as well as ICONST because values that
> * don't fit in the platform's "long", but do fit in bigint, should still be
> * accepted here. (This is possible in 64-bit Windows as well as all 32-bit
> * builds.)
> ```

For example,
```
select 1 offset 1 row fetch first row only;
                                  ^
```
This is valid in PostgreSQL. When we see this `row` keyword, it should be part of the `fetch first [n] row` syntax where `n` is omitted, but it may be mistakenly treated as part of the `row(field0, field1)` expression because we will attempt to parse an expression (instead of a numerical constant) after `first`.

What is the actual parsing behavior when this commit sees this SQL?

lmatz: how about let me revert the changes for "fetch" but keep the changes for "limit", 
the CI test required by SQLMesh integration needs `limit simple expression` only.

let `fetch` still stick to a literal number

lmatz: or we force parentheses around the expression in Fetch just like PG
```
SELECT id
FROM generate_series(1, 10) AS id
ORDER BY id
OFFSET 5 ROWS
FETCH NEXT (1+4) ROWS ONLY;

id 
----
  6
  7
  8
  9
 10
(5 rows)

SELECT id
FROM generate_series(1, 10) AS id
ORDER BY id
OFFSET 5 ROWS
FETCH NEXT 1+4 ROWS ONLY;

psql:commands.sql:11: ERROR:  syntax error at or near "+"
LINE 5: FETCH NEXT 1+4 ROWS ONLY;
```

src/frontend/src/planner/query.rs
```diff
@@ -50,7 +52,37 @@ impl Planner {
...
+                .into());
+            }
+            let limit_cast_to_bigint = limit.cast_explicit(DataType::Int64).map_err(|_| {
+                RwError::from(ErrorCode::ExprError(
+                        "expects an integer or expression that can be evaluated to an integer after LIMIT"
+                    .into(),
+                ))
+            })?;
+            let limit = match limit_cast_to_bigint.fold_const() {
+                Ok(Some(datum)) => {
```
xiangjinwu, line 70: * This _validation_ logic is better placed in binder, and `BoundQuery::limit` remains an `Option<u64>`. Planner is in charge of _transforming_ SQL into a plan graph.
* We can do cast first and then `try_fold_const`, rather than using `is_const` and `fold_const` separately.
* The allowed casts in PostgreSQL is `assign` rather than `explicit`. (`int` requires `implicit`, `decimal` requires `assign`, `jsonb` requires `explicit`. I tested in PostgreSQL and `LIMIT '1'::jsonb` is not allowed.)
* When the evaluated output is `None`, which means SQL NULL, PostgreSQL treats it as no limit. Let's add a test case for this.
* Test error cases: wrong type, not const, eval error, negative

src/sqlparser/src/ast/query.rs
xiangjinwu: > or we force parentheses around the expression in Fetch just like PG

PostgreSQL is more complex than this, but yes parenthesized expr is a subset of what PostgreSQL allows (`c_expr` as defined in the linked `gram.y` above, while `limit` allows `a_expr`).

src/frontend/src/planner/query.rs
lmatz: > Test error cases: wrong type, not const, eval error, negative

wrong type, negative can be constructed, I added them

but "eval error and not const" seems to be avoided by the parsing stage already

src/frontend/src/binder/query.rs
```diff
@@ -174,13 +174,47 @@ impl Binder {
...
+                }
+                // If evaluated to NULL, we follow PG to treat NULL as no limit
+                Some(Ok(None)) => {
+                    u64::MAX
+                }
+                // wrong type, not const, eval error all belongs to this branch
+                _ => return Err(ErrorCode::ExprError(
+                    "expects an integer or expression that can be evaluated to an integer after LIMIT"
+                        .into(),
+                ).into()),
```
xiangjinwu, line 211: * Wrong type already handled as `cast_assign` error above.
* Not const corresponds to `None`. The test case can be `select 1 limit generate_series(1, 2);`
* Eval error corresponds to `Some(Err(_))`. The test case can be `select 1 limit 2/0;`

Preferably `_` shall be used with extra care.

src/frontend/src/binder/query.rs
```diff
@@ -174,13 +174,47 @@ impl Binder {
...
+        let limit = if let Some(limit_expr) = limit_expr {
+            let limit_cast_to_bigint = limit_expr.cast_assign(DataType::Int64).map_err(|_| {
+                RwError::from(ErrorCode::ExprError(
+                    "expects an integer or expression that can be evaluated to an integer after LIMIT"
+                        .into(),
+                ))
+            })?;
+            let limit = match limit_cast_to_bigint.try_fold_const() {
+                Some(Ok(Some(datum))) => {
+                    let value = datum.as_integral();
```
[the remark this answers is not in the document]
lmatz: I overlooked the methods implemented in the macro

e2e_test/batch/order/test_limit.slt.part
```diff
@@ -182,6 +182,76 @@ INSERT INTO integers VALUES (1), (2), (3), (4), (5);
...
+query I
+SELECT * FROM generate_series(0,3,1) as t(v) order by v limit NULL + '1'::bigint;
+----
+0
+1
+2
+3
+
+statement error
+SELECT * FROM generate_series(0,3,1) as t(v) order by v limit NULL + '1';
```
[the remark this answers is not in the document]
lmatz: updated

src/frontend/src/binder/query.rs
lmatz: separated

src/frontend/src/planner/query.rs
lmatz: used "cast_assign" and "try_fold_const" instead

xiangjinwu (approved):

PR #19689, opened 2024-12-05
feat: add connector props and format options in rw_catalog

tabVersion:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://gist.github.com/TennyZhuang/f00be7f16996ea48effb049aa7be4d66#file-rw_cla).

## What's changed and what's your intention?

resolve #19072

## Checklist

- [ ] I have written necessary rustdoc comments
- [ ] I have added necessary unit tests and integration tests
- [ ] I have added test labels as necessary. See [details](https://github.com/risingwavelabs/risingwave/blob/main/docs/developer-guide.md#ci-labels-guide).
- [ ] I have added fuzzing tests or opened an issue to track them. (Optional, recommended for new SQL features #7934).
- [ ] My PR contains breaking changes. (If it deprecates some features, please create a tracking issue to remove them in the future).
- [ ] All checks passed in `./risedev check` (or alias, `./risedev c`)
- [ ] My PR changes performance-critical code. (Please run macro/micro-benchmarks and show the results.)

- [ ] My PR contains critical fixes that are necessary to be merged into the latest release. (Please check out the [details](https://github.com/risingwavelabs/risingwave/blob/main/CONTRIBUTING.md))

## Documentation

- [ ] My PR needs documentation updates. (Please use the **Release note** section below to summarize the impact on users)

## Release note

If this PR includes changes that directly affect users or other significant modifications relevant to the community, kindly draft a release note to provide a concise summary of these changes. Please prioritize highlighting the impact these changes will have on users.

</details>

PR #19845, opened 2024-12-18
fix: reduct credentials in rw_catalog

tabVersion:

as title 

related https://github.com/risingwavelabs/risingwave/pull/19689 & https://github.com/risingwavelabs/risingwave/issues/19766

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/catalog/system_catalog/rw_catalog/mod.rs
+++ b/src/frontend/src/catalog/system_catalog/rw_catalog/mod.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+mod reduct_props;
 mod rw_actor_infos;
 mod rw_actors;
 mod rw_columns;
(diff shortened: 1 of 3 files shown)


yezizp2012 (approved):

fuyufjh:
Weakly -1. IMO, This is not something that a system table should do.

fuyufjh:
> Weakly -1. IMO, This is not something that a system table should do.

Just saying, for example, if we develop a `rw_dump` tool like [`pg_dump`](https://www.postgresql.org/docs/current/app-pgdump.html), it will fetch all information from system catalog. The system table is supposed to be a source of truth.

----


I think we are able to mitigate the issue via proper privilge check https://github.com/risingwavelabs/risingwave/issues/19722.

tabVersion:
> I think we are able to mitigate the issue via proper privilge check https://github.com/risingwavelabs/risingwave/issues/19722.

+1 for this. Limiting the privilege to rw_catalog can prevent leaking. 

Another motivation for the PR is the cloud team also don't want to display the credentials on the portal, there has to be a mosaic step.

PR #19595, opened 2024-11-27
feat(ci): more meta backend agnostic e2e tests in `main-cron`

BugenZhao:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://gist.github.com/TennyZhuang/f00be7f16996ea48effb049aa7be4d66#file-rw_cla).

## What's changed and what's your intention?

- Use `meta-backend: env` for all CI RiseDev profiles.
  * use a persisted SQLite database as a fallback when env var is not set, behaving as if `meta-backend: sqlite` is specified, ensuring compatibility with the behavior prior to this PR.

- Extend coverage for PG/MySQL meta backends on `main-cron` to
  * slow e2e test
  * parallel e2e test
  * e2e source test
  * e2e sink test

Note that although there are many other steps for specific connectors, I suppose it may not be necessary to cover them all, as the execution path on the meta node should not differ significantly.

## Checklist

- [x] I have written necessary rustdoc comments
- [x] I have added necessary unit tests and integration tests
- [x] All checks passed in `./risedev check` (or alias, `./risedev c`)

## Documentation

- [ ] My PR needs documentation updates. (Please use the **Release note** section below to summarize the impact on users)

## Release note

If this PR includes changes that directly affect users or other significant modifications relevant to the community, kindly draft a release note to provide a concise summary of these changes. Please prioritize highlighting the impact these changes will have on users.

</details>

PR #19846, opened 2024-12-18
fix(meta): fix & enable e2e tests with MySQL meta backend

BugenZhao:

Follow up of #19595. Address bugs for MySQL meta backend and re-enable it.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/meta/model/migration/clippy.toml
+++ b/src/meta/model/migration/clippy.toml
@@ -1,0 +1,4 @@
+disallowed-methods = [
+    { path = "sea_query::table::column::ColumnDef::binary", reason = "Please use `.rw_binary(..)` instead." },
+    { path = "sea_query::table::column::ColumnDef::blob", reason = "Please use `.rw_binary(..)` instead." },
+]
--- a/src/meta/model/migration/src/m20241125_043732_connection_params.rs
+++ b/src/meta/model/migration/src/m20241125_043732_connection_params.rs
@@ -1,5 +1,7 @@
 use sea_orm_migration::prelude::*;
 
+use crate::utils::ColumnDefExt;
+
 #[derive(DeriveMigrationName)]
 pub struct Migration;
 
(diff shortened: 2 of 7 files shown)


yezizp2012 (approved):
LGTM!

src/meta/model/migration/clippy.toml
```diff
@@ -0,0 +1,4 @@
+disallowed-methods = [
+    { path = "sea_query::table::column::ColumnDef::binary", reason = "Please use `.rw_binary(..)` instead." },
```
fuyufjh, line 2: Can you explain it by some comments?

fuyufjh (approved):

src/meta/model/migration/clippy.toml
BugenZhao: Yes, it's explained in docs:

```rust
    /// Set column type as `longblob` for MySQL, `bytea` for Postgres, and `blob` for Sqlite.
    ///
    /// Should be preferred over [`binary`](ColumnDef::binary) or [`blob`](ColumnDef::blob) for large binary fields,
    /// typically the fields wrapping protobuf or other serialized data. Otherwise, MySQL will return an error
    /// when the length exceeds 65535 bytes.
```

src/meta/model/migration/src/m20241125_043732_connection_params.rs
```diff
@@ -10,7 +12,11 @@ impl MigrationTrait for Migration {
             .alter_table(
                 Table::alter()
                     .table(Connection::Table)
-                    .add_column(ColumnDef::new(Connection::Params).binary().not_null())
+                    .add_column(
+                        ColumnDef::new(Connection::Params)
+                            .rw_binary(manager)
+                            .not_null(),
+                    )
```
stdrc, line 19: Is it correct to directly change an existing migration script? Not sure, just want to confirm.

yezizp2012: Only for MySQL, I think it's okay.

PR #19849, opened 2024-12-18
fix: Fix the panic issue with the parallelism() call

shanicky:

In theory, only compute nodes should have the concept of parallelism, so invoking `parallelism()` for other types of workers is meaningless and should not be done. 

This pull request renames the original `parallelism` to `compute_parallelism` and adds a new `parallelism` return option to prevent potential misuse.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/prost/src/lib.rs
+++ b/src/prost/src/lib.rs
@@ -221,7 +221,7 @@ impl stream_plan::MaterializeNode {
 
 // Encapsulating the use of parallelism.
 impl common::WorkerNode {
-    pub fn compute_parallelism(&self) -> usize {
+    pub fn compute_node_parallelism(&self) -> usize {
         assert_eq!(self.r#type(), WorkerType::ComputeNode);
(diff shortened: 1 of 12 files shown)


yezizp2012 (approved):

xxchan (approved):
Thanks for the fix. BTW maybe we can add a test for diagnose?

src/prost/src/lib.rs
```diff
@@ -221,13 +221,21 @@ impl stream_plan::MaterializeNode {
 
 // Encapsulating the use of parallelism.
 impl common::WorkerNode {
-    pub fn parallelism(&self) -> usize {
+    pub fn compute_parallelism(&self) -> usize {
```
stdrc, line 224: Does this new name look like "computing" the parallelism when it's called? What about `cn_parallelism` or `cnode_parallelism`?

PR #19869, opened 2024-12-20
fix(sink): fix mogodb write error handling

xxhZs:

in doc https://www.mongodb.com/zh-cn/docs/manual/reference/method/db.collection.bulkWrite/#mongodb-method-db.collection.bulkWrite 
The correct error is `writeErrors` and `writeConcernErrors`
And you shouldn't use n to check if the write was successful, because there are times when n will be 0.
Let's say we delete a piece of mongodb non-existent data

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/src/sink/mongodb.rs
+++ b/src/connector/src/sink/mongodb.rs
@@ -81,11 +81,17 @@ mod send_bulk_write_command_future {
             )))
         })?;
 
-        if result.get_array("writeErrors").is_ok() || result.get_array("writeConcernError").is_ok()
-        {
+        if let Ok(write_errors) = result.get_array("writeErrors") {
             return Err(SinkError::Mongodb(anyhow!(
                 "bulk write respond with write errors: {:?}",
-                result,
+                write_errors,
+            )));
+        }
+
+        if let Ok(write_concern_error) = result.get_array("writeConcernError") {
+            return Err(SinkError::Mongodb(anyhow!(
+                "bulk write respond with write errors: {:?}",
+                write_concern_error,
             )));
         }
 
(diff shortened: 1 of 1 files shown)


ly9chee:
> And you shouldn't use n to check if the write was successful, because there are times when n will be 0.
Let's say we delete a piece of mongodb non-existent data

The initial implementation checks n > 0 becase we assumes ownership of the sinking collection. There should not be a case where n == 0, unless the user manually modified the collection.

Should we also check the ok status? Then we can expect the command is succeeded and without any write errors.
https://www.mongodb.com/docs/manual/reference/command/delete/#output

xxhZs:
> > And you shouldn't use n to check if the write was successful, because there are times when n will be 0.
> > Let's say we delete a piece of mongodb non-existent data
> 
> The initial implementation checks n > 0 becase we assumes ownership of the sinking collection. There should not be a case where n == 0, unless the user manually modified the collection.
> 
> Should we also check the ok status? Then we can expect the command is succeeded and without any write errors. https://www.mongodb.com/docs/manual/reference/command/delete/#output

add ok, Please review if you have time, thanks!

ly9chee:
> > > And you shouldn't use n to check if the write was successful, because there are times when n will be 0.
> > > Let's say we delete a piece of mongodb non-existent data
> > 
> > 
> > The initial implementation checks n > 0 becase we assumes ownership of the sinking collection. There should not be a case where n == 0, unless the user manually modified the collection.
> > Should we also check the ok status? Then we can expect the command is succeeded and without any write errors. https://www.mongodb.com/docs/manual/reference/command/delete/#output
> 
> add ok, Please review if you have time, thanks!

LGTM!

chenzl25 (approved):

PR #19136, opened 2024-10-25
fix(runtime): set default crypto provider

zwang28:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://gist.github.com/TennyZhuang/f00be7f16996ea48effb049aa7be4d66#file-rw_cla).

## What's changed and what's your intention?

The bug was identified when upgrading a cluster from v1.10 to v2.0. The meta node encountered a panic during startup:
```
no process-level CryptoProvider available -- call CryptoProvider::install_default() before this point
...
2024-10-25 14:06:15.627	  33:     0x574ae1e287c4 - risingwave_connector::source::google_pubsub::PubsubProperties::subscription_client::{{closure}}::h16ec7b21592f473b
2024-10-25 14:06:15.627	                               at /risingwave/src/connector/src/source/google_pubsub/mod.rs:126:14
```

I've confirmed that google-cloud-pubsub version 0.25 is free from this bug (used in RW v1.10), while version 0.28 (used in RW v2.0) does contain the bug.

The fix has been manually tested.

## Checklist

- [ ] I have written necessary rustdoc comments
- [ ] I have added necessary unit tests and integration tests
- [ ] I have added test labels as necessary. See [details](https://github.com/risingwavelabs/risingwave/blob/main/docs/developer-guide.md#ci-labels-guide).
- [ ] I have added fuzzing tests or opened an issue to track them. (Optional, recommended for new SQL features #7934).
- [ ] My PR contains breaking changes. (If it deprecates some features, please create a tracking issue to remove them in the future).
- [ ] All checks passed in `./risedev check` (or alias, `./risedev c`)
- [ ] My PR changes performance-critical code. (Please run macro/micro-benchmarks and show the results.)

- [ ] My PR contains critical fixes that are necessary to be merged into the latest release. (Please check out the [details](https://github.com/risingwavelabs/risingwave/blob/main/CONTRIBUTING.md))

## Documentation

- [ ] My PR needs documentation updates. (Please use the **Release note** section below to summarize the impact on users)

## Release note

If this PR includes changes that directly affect users or other significant modifications relevant to the community, kindly draft a release note to provide a concise summary of these changes. Please prioritize highlighting the impact these changes will have on users.

</details>

PR #19874, opened 2024-12-20
fix: use ring as default provider

zwang28:

The original aws_lc_rs provider may default to FIPS mode in certain situations, where FIPS may not be available, leading to panic due to `error:12800067:DSO support routines:dlfcn_load:could not load the shared library:../crypto/dso/dso_dlfcn.c:118:filename(/usr/lib/x86_64-linux-gnu/ossl-modules/fips.so): /usr/lib/x86_64-linux-gnu/ossl-modules/fips.so: cannot open shared object file: No such file or directory, error:12800067:DSO support routines:DSO_load:could not load the shared library:../crypto/dso/dso_lib.c:152:, error:07880025:common libcrypto routines:provider_init:reason(524325):../crypto/provider_core.c:912:name=fips`

This PR changes the provider to `ring`.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/utils/runtime/src/lib.rs
+++ b/src/utils/runtime/src/lib.rs
@@ -74,7 +74,7 @@ where
 {
     set_panic_hook();
 
-    rustls::crypto::aws_lc_rs::default_provider()
+    rustls::crypto::ring::default_provider()
         .install_default()
         .inspect_err(|e| {
             tracing::error!(?e, "Failed to install default crypto provider.");


arkbriar:
Several questions:
- What's the difference between `aws_lc_rs` and `ring`?
- Shall we also bundle the fips share lib in our image?

arkbriar:
+ @lmatz

zwang28:
> What's the difference between aws_lc_rs and ring?

[aws_lc does not appear to be actively tested on BSD systems.](https://github.com/aws/aws-lc?tab=readme-ov-file#other-platforms)

hzxa21 (approved):
LGTM

Have we tested that using `ring` can also prevent the google-pub-sub issue mentioned in #19136?

PR #19883, opened 2024-12-20
refactor(optimizer): use `Cell` instead of `RefCell` for interior mutability

stdrc:

In `OptimizerContext`, we have some fields representing the last assigned optimizer element IDs, e.g. plan node ID. Previously we use `RefCell` for these fields, but what we actually want is just interior mutability, not runtime borrow checking. This PR changes these fields to use `Cell` instead.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/optimizer/optimizer_context.rs
+++ b/src/frontend/src/optimizer/optimizer_context.rs
@@ -59,11 +59,11 @@ pub struct OptimizerContext {
     rcte_cache: RefCell<HashMap<ShareId, PlanRef>>,
 
     /// Last assigned plan node ID.
-    next_plan_node_id: RefCell<i32>,
+    last_plan_node_id: RefCell<i32>,
     /// Last assigned correlated ID.
-    next_correlated_id: RefCell<u32>,
+    last_correlated_id: RefCell<u32>,
     /// Last assigned expr display ID.
-    next_expr_display_id: RefCell<usize>,
+    last_expr_display_id: RefCell<usize>,
 
     _phantom: PhantomUnsend,
 }
@@ -96,9 +96,9 @@ impl OptimizerContext {
             overwrite_options,
             rcte_cache: RefCell::new(HashMap::new()),
 
-            next_plan_node_id: RefCell::new(RESERVED_ID_NUM.into()),
-            next_correlated_id: RefCell::new(0),
-            next_expr_display_id: RefCell::new(RESERVED_ID_NUM.into()),
+            last_plan_node_id: RefCell::new(RESERVED_ID_NUM.into()),
+            last_correlated_id: RefCell::new(0),
+            last_expr_display_id: RefCell::new(RESERVED_ID_NUM.into()),
 
             _phantom: Default::default(),
         }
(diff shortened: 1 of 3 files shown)


src/frontend/src/optimizer/optimizer_context.rs
```diff
@@ -107,53 +116,57 @@ impl OptimizerContext {
...
+        self.last_expr_display_id.set(0);
     }
 
-    pub fn next_correlated_id(&self) -> CorrelatedId {
-        *self.next_correlated_id.borrow_mut() += 1;
-        *self.next_correlated_id.borrow()
+    pub(in crate::optimizer) fn restore_elem_ids(&self, backup: LastAssignedIds) {
+        self.last_plan_node_id.set(backup.last_plan_node_id);
+        self.last_correlated_id.set(backup.last_correlated_id);
+        self.last_expr_display_id.set(backup.last_expr_display_id);
```
xxchan, line 169: What about just inlining these functions info `reorganize_elements_id`?

xxchan (approved):

src/frontend/src/optimizer/optimizer_context.rs
stdrc: When adding a new `id` later, we can hardly remember that there's a `reorganize_elements_id` far away need to be updated. So I decided to hide all the ids inside `LastAssignedIds`.

xxchan: I don't get you. We can still add an `id` in `OptimizerContext` but not in `LastAssignedIds`

xxchan: Maybe you should replace the fields to `LastAssignedIds` in `OptimizerContext`

src/frontend/src/optimizer/optimizer_context.rs
```diff
@@ -255,12 +268,12 @@ impl std::fmt::Debug for OptimizerContext {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         write!(
```
BugenZhao, line 269: Use `.debug_struct`?

src/frontend/src/optimizer/optimizer_context.rs
stdrc: The fact is that we missed `last_correlated_id` in `reorganize_elements_id` before. I want to reduce the possibility of future error, so I moved them to `LastAssignedIds` and put it right next to `OptimizerContext`. I also definitely considered directly using `LastAssignedIds` as a field in `OptimizerContext`, but IMO it's too annoying to access these ID fields via a long `self.last_assigned_ids.last_plan_node_id`.

This PR doesn't aim to completely refactor the optimizer element IDs, otherwise I would extract an `IdAllocator` struct to allocate IDs instead of `self.last_plan_node_id.update` in `OptimizerContext`. Also I don't think use interior mutability here in `OptimizerContext` is a good idea. But I don't want to do that in this PR. It's just a quick fix to unblock my future work on other things.

Issue #18802, opened 2024-10-08
Tracking (storage): support state cleaning on non-pk prefix watermark column


hzxa21:
We decided to adopt option 3 mentioned [here](https://github.com/risingwavelabs/risingwave/issues/18728#issuecomment-2376059930)

Li0k:
Disscuss wtih @hzxa21 and @st1page, i found some assumptions exist for `non_pk_watermark`
1. the length of the watermark col is 1.
2. the state table that needs to be cleaned, the watermrak_col exists in pk.
3. the operator guarantees not to access data below the watermark, storage does not need to give this guarantee.

Therefore, the design on the storage side is as follows:

1. state table holds watermark_col
2. serialise watermark based on watermark_col instead of prefix(1)
3. when writing to TableWaterMark, record that it is a `non_pk_watermark`
4. modify the `skip_watermark_iter` implementation so that `non_pk_watermark` needs to solve for watermark_col before comparing.

PR #19889, opened 2024-12-23
feat(storage): support non_pk_prefix_watermark state cleaning

Li0k:

related to https://github.com/risingwavelabs/risingwave/issues/18802

This PR supports `non_pk_prefix_watermark` state cleaning for Hummock. 

Since `non_pk_prefix_watermark` relies on catalogs, this introduces additional overhead.  Therefore, this PR does not guarantee read filtering for `non_pk_prefix_watermark` and only handles expired data.

The changes are as follows:
1. watermarks of type `non_pk_prefix_watermark` are not added to `ReadWatermarkIndex`.

2. state table support to write `non_pk_prefix_watermark` and serialize.

3. compaction catalog agent support to get watermark serde 

4. ~skip watermark iterator supports filtering `non_pk_prefix_watermark`.~ support `NonPkPrefixSkipWatermarkIterator` to filter table data of `NonPkPrefix` Watermark.

## Checklist

## Documentation

<details>

</details>


src/storage/src/hummock/store/version.rs
```diff
@@ -249,19 +249,24 @@ impl HummockReadVersion {
...
+                            table_watermarks.clone(),
+                            committed_version
+                                .state_table_info
+                                .info()
+                                .get(&table_id)
+                                .expect("should exist")
+                                .committed_epoch,
+                        )),
+
+                        WatermarkSerdeType::NonPkPrefix => None, /* do not fill the non-pk prefix watermark to index */
```
hzxa21, line 265: Given that non pk prefix watermark will never be used in CN, should we exclude it in meta-CN version delta notification? We can add sanity check in the CN side to make sure this assumption holds.

src/meta/src/hummock/manager/compaction/mod.rs
```diff
@@ -728,6 +729,46 @@ impl HummockManager {
...
+                        .map(|id| id.table_id() as _)
+                        .collect(),
+                )
+                .await
+                .map_err(|e| Error::Internal(e.into()))?
+                .into_iter()
+                .filter_map(|table| {
+                    // pk prefix watermark.
+                    if table.clean_watermark_index_in_pk.is_none()
+                        || table.clean_watermark_index_in_pk.unwrap() == 0
```
hzxa21, line 750: Does this assume the pk prefix watermark can only be a single column? Do we have a sanity check somewhere to make sure this assumption holds?

src/meta/src/hummock/manager/compaction/mod.rs
```diff
@@ -728,6 +729,46 @@ impl HummockManager {
...
+                    }
+                })
+                .collect();
+
+            let table_watermarks = version
+                .latest_version()
+                .table_watermarks
+                .iter()
+                .filter_map(|(table_id, table_watermarks)| {
+                    if table_id_with_pk_prefix_watermark.contains(table_id) {
```
hzxa21, line 764: We already have a `WaterMarkType` [define](https://github.com/risingwavelabs/risingwave/blob/3500061180b4eafb23dbeed1c41c0ee7d00ae97d/src/storage/hummock_sdk/src/table_watermark.rs#L412) in the version, why don't we just use that to filter out table with non pk prefix watermark?

hzxa21: Also, if we filter out non pk prefix watermark here, how can compactor retrieve the non pk prefix watermark? Based on the logic [here](https://github.com/risingwavelabs/risingwave/blob/3500061180b4eafb23dbeed1c41c0ee7d00ae97d/src/storage/src/hummock/compactor/compactor_runner.rs#L249), it seems that we rely on the fact that non pk prefix watermark is present in the compact task.

src/storage/hummock_sdk/src/compact_task.rs
```diff
@@ -114,6 +115,48 @@ impl CompactTask {
     }
 }
 
+impl CompactTask {
+    // The compact task may need to reclaim key with TTL
+    pub fn is_contains_ttl(&self) -> bool {
```
hzxa21, line 120: rename: `contains_ttl`

src/storage/hummock_sdk/src/compact_task.rs
```diff
@@ -114,6 +115,48 @@ impl CompactTask {
...
+        self.table_options
+            .iter()
+            .any(|(_, table_option)| table_option.retention_seconds.is_some_and(|ttl| ttl > 0))
+    }
+
+    // The compact task may need to reclaim key with range tombstone
+    pub fn is_contains_range_tombstone(&self) -> bool {
```
hzxa21, line 127: rename: `contains_range_tombstone`

src/storage/hummock_sdk/src/compact_task.rs
```diff
@@ -114,6 +115,48 @@ impl CompactTask {
...
+        self.input_ssts
+            .iter()
+            .flat_map(|level| level.table_infos.iter())
+            .any(|sst| sst.range_tombstone_count > 0)
+    }
+
+    // The compact task may need to reclaim key with split sst
+    pub fn is_contains_split_sst(&self) -> bool {
```
hzxa21, line 135: rename: `contains_split_sst`

src/storage/hummock_sdk/src/table_watermark.rs
```diff
@@ -292,7 +295,11 @@ impl Display for WatermarkDirection {
 }
 
 impl WatermarkDirection {
-    pub fn filter_by_watermark(&self, key: impl AsRef<[u8]>, watermark: impl AsRef<[u8]>) -> bool {
+    pub fn filter_by_watermark_key(
```
hzxa21, line 298: minor: `key_filter_by_watermark`

src/storage/hummock_sdk/src/table_watermark.rs
```diff
@@ -301,6 +308,26 @@ impl WatermarkDirection {
         }
     }
 
+    pub fn filter_by_watermark_datum(
```
hzxa21, line 311: minor: `datum_filter_by_watermark`

src/storage/hummock_sdk/src/table_watermark.rs
```diff
@@ -19,11 +19,14 @@ use std::mem::size_of;
 use std::ops::Bound::{Excluded, Included, Unbounded};
 use std::sync::Arc;
 
+use bincode::{Decode, Encode};
```
hzxa21, line 22: We need this because we define `Decode` `Encode` on `WatermarkSerdeType`. Is this neccessary?

src/storage/hummock_sdk/Cargo.toml
```diff
@@ -8,6 +8,7 @@ license = { workspace = true }
 repository = { workspace = true }
 
 [dependencies]
+bincode = { version = "=2.0.0-rc.3", features = ["serde"] }
```
hzxa21, line 11: ditto. See comments above.

src/storage/hummock_sdk/src/table_watermark.rs
hzxa21: I think the reason why we need it is because we need it for hummock trace. Like other trace related struct, let's not leak encode/decode outside of hummock trace and handle it [here](https://github.com/risingwavelabs/risingwave/blob/3500061180b4eafb23dbeed1c41c0ee7d00ae97d/src/storage/src/store.rs#L831) because we can just use a bool to do the encode and decode in hummock trace.

src/storage/src/hummock/iterator/skip_watermark.rs
```diff
@@ -42,10 +47,14 @@ pub struct SkipWatermarkIterator<I> {
 }
 
 impl<I: HummockIterator<Direction = Forward>> SkipWatermarkIterator<I> {
```
hzxa21, line 49: nits: since `SkipWatermarkIterator` is only used by compactor, how about moving `skip_watermark.rs` into src/hummock/compactor?

src/storage/src/hummock/iterator/skip_watermark.rs
```diff
@@ -252,9 +320,38 @@ impl SkipWatermarkState {
...
+                                        watermark_col_idx_in_pk,
+                                    ) = self.last_serde.as_ref().unwrap();
+                                    let row = pk_prefix_serde
+                                            .deserialize(inner_key)
+                                            .unwrap_or_else(|_| {
+                                                panic!("Failed to deserialize pk_prefix inner_key {:?} serde data_types {:?} order_types {:?}", inner_key, pk_prefix_serde.get_data_types(), pk_prefix_serde.get_order_types());
+                                            });
+                                    let watermark_col_in_pk =
+                                        row.datum_at(*watermark_col_idx_in_pk);
+                                    cmp_datum(
```
hzxa21, line 346: IIUC, if `cmp_datum` returns `Euqal | Greater`, based on the logic in L360, the watermark will be advanced. I think this is incorrect for non pk prefix watermark because the non pk prefix watermark and the pk doesn't have the same ordering.

src/meta/src/hummock/manager/compaction/mod.rs
Li0k: Good catch , we should filter the watermark by WaterMarkType directly.

And, the filtered results are only passed to the picker, while all relevant watermarks are passed to the compactor (pk or non-pk).

src/storage/src/hummock/store/version.rs
Li0k: https://github.com/risingwavelabs/risingwave/blob/3500061180b4eafb23dbeed1c41c0ee7d00ae97d/src/storage/src/hummock/store/local_hummock_storage.rs#L532

Yes, we have avoided sending the NonPkPrefix Watermark in notifications

src/storage/src/hummock/iterator/skip_watermark.rs
Li0k: Of course, I use a separate pr for it

src/storage/src/hummock/store/version.rs
hzxa21: I guess you pasted the wrong codes? Where do we exclude sending NonPkPrefix watermark in notifications? I cannot find the relevant codes. Also, if we do exlude that from notification, we should add sanity check for `WatermarkSerdeType::NonPkPrefix` the match arms in the CN side instead of filtering it out.

src/meta/src/hummock/manager/compaction/mod.rs
```diff
@@ -728,6 +729,22 @@ impl HummockManager {
...
+                    if matches!(
+                        table_watermarks.watermark_type,
+                        WatermarkSerdeType::PkPrefix,
+                    ) {
+                        Some((*table_id, table_watermarks.clone()))
+                    } else {
+                        None
+                    }
+                })
+                .collect();
```
hzxa21, line 746: Actually why don't we do the filtering inside the picker instead like in here if the watermark type is part of TableWatermarks: https://github.com/risingwavelabs/risingwave/blob/5ed4920f2c61bd4a1ebf25ec902b0c9521d01c15/src/meta/src/hummock/compaction/selector/vnode_watermark_selector.rs#L53

We can avoid cloning the table watermark, which can be large given that it stores bytes from user data, with no harm.

src/storage/src/hummock/iterator/skip_watermark.rs
```diff
@@ -252,9 +319,23 @@ impl SkipWatermarkState {
...
                     match direction {
                         WatermarkDirection::Ascending => {
-                            match inner_key.cmp(watermark.as_ref()) {
+                            let cmp = match watermark {
+                                WatermarkSerdeType::Bytes(watermark) => {
+                                    inner_key.cmp(watermark.as_ref())
+                                }
+                                WatermarkSerdeType::Serde(_watermark) => {
+                                    // do not skip the non-pk prefix watermark when vnode is the same
+                                    return false;
```
hzxa21, line 334: I am afraid this is still incorrect based on the semantic of `advance_watermark`:
```
/// Return a flag indicating whether the current key will be filtered by the current watermark.
```

If we always return false when the table, vnode are the same here, that means none of the keys can be filtered by the watermark. Please clearfully walk through the logics of `advance_watermark`, `should_delete` and `advance_key_and_watermark`. I am still concerned that the implementation of `SkipWatermarkState` and `SkipWatermarkIterator` rely on the assumption that the key ordering and watermark ordering is the same and we may still miss some changes.

src/storage/src/hummock/compactor/compactor_runner.rs
```diff
@@ -239,13 +243,26 @@ impl CompactorRunner {
...
         // The `SkipWatermarkIterator` is used to handle the table watermark state cleaning introduced
         // in https://github.com/risingwavelabs/risingwave/issues/13148
-        Ok(SkipWatermarkIterator::from_safe_epoch_watermarks(
-            MonitoredCompactorIterator::new(
-                MergeIterator::for_compactor(table_iters),
-                task_progress.clone(),
-            ),
-            &self.compact_task.table_watermarks,
-        ))
+        let combine_iter = {
```
hzxa21, line 246: Let's add comment and referencing issue for state cleaning based on NonPkPrefix watermark iterator as well.

src/storage/src/hummock/compactor/compaction_utils.rs
```diff
@@ -676,3 +663,23 @@ pub fn calculate_task_parallelism_impl(
...
+    BTreeMap<u32, TableWatermarks>,
+) {
+    let mut pk_watermarks = BTreeMap::default();
+    let mut non_pk_prefix_watermarks = BTreeMap::default();
+
+    for (table_id, watermark) in &compact_task.table_watermarks {
+        if let WatermarkSerdeType::PkPrefix = watermark.watermark_type {
+            pk_watermarks.insert(*table_id, watermark.clone());
+        } else {
+            non_pk_prefix_watermarks.insert(*table_id, watermark.clone());
```
hzxa21, line 680: Can we avoid the watermark clone here by pre-spliting the table watermark into pk prefix watermark and non-pk prefix watermark in the compact task struct here? https://github.com/risingwavelabs/risingwave/blob/7e8f6cddc2ffd0f0fb0c65e110471e5812c07309/src/storage/hummock_sdk/src/compact_task.rs#L200

src/storage/src/hummock/iterator/skip_watermark.rs
```diff
@@ -303,24 +307,341 @@ impl SkipWatermarkState {
...
+                    let row = pk_prefix_serde
+                        .deserialize(inner_key)
+                        .unwrap_or_else(|_| {
+                            panic!("Failed to deserialize pk_prefix inner_key {:?} serde data_types {:?} order_types {:?}", inner_key, pk_prefix_serde.get_data_types(), pk_prefix_serde.get_order_types());
+                        });
+                    let watermark_col_in_pk = row.datum_at(*watermark_col_idx_in_pk);
+                    return direction.datum_filter_by_watermark(
+                        watermark_col_in_pk,
+                        watermark,
+                        watermark_col_serde.get_order_types()[0],
```
hzxa21, line 372: Given that we use `[0]` here, does it mean that either the watermark column is always a single column or the order type for all watermark columns must be the same? Is this assumption checked and guaranteed somewhere?

src/storage/src/hummock/iterator/skip_watermark.rs
```diff
@@ -303,24 +307,341 @@ impl SkipWatermarkState {
...
+                        .unwrap_or_else(|_| {
+                            panic!("Failed to deserialize pk_prefix inner_key {:?} serde data_types {:?} order_types {:?}", inner_key, pk_prefix_serde.get_data_types(), pk_prefix_serde.get_order_types());
+                        });
+                    let watermark_col_in_pk = row.datum_at(*watermark_col_idx_in_pk);
+
+                    return direction.datum_filter_by_watermark(
+                        watermark_col_in_pk,
+                        watermark,
+                        watermark_col_serde.get_order_types()[0],
+                    );
```
hzxa21, line 449: The logics here is identical to the logics in `should_delete`. How about extract it into a common function?

src/storage/src/hummock/iterator/skip_watermark.rs
```diff
@@ -303,24 +307,341 @@ impl SkipWatermarkState {
...
+                Ordering::Greater => {
+                    return false;
+                }
+            }
+        }
+        false
+    }
+}
+
+pub struct NonPkPrefixSkipWatermarkIterator<I> {
```
hzxa21, line 460: The logic for `SkipWatermarkIterator` and `NonPkPrefixSkipWatermarkIterator` are almost identical. Is is possible to unify them? We can still wrap non-pk skip watermark iter on top of skip watermark iterator even if we unify the implementation.

src/common/src/util/row_serde.rs
```diff
@@ -52,6 +52,18 @@ impl OrderedRowSerde {
         }
     }
 
+    #[must_use]
+    pub fn index(&self, idx: usize) -> Cow<'_, Self> {
+        if 1 == self.order_types.len() {
```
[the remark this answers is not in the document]
Li0k: No, row can be any length and index can be a generic function.

src/storage/src/hummock/iterator/skip_watermark.rs
Li0k: This assumes that the watermark is a single column. I've found no such guarantee on the state table side either. 

From my point of view, this should be the point that the optimizer needs to guarantee, what is the reasonable way to do it? @hzxa21 @st1page

Li0k: As said above, I think this guarantee should come from `clean_watermark_index_in_pk`. If the assumption is broken then using [0] should crash, which is fine with me.

hzxa21: If there are multiple watermark column of different ordering type, I think this code won't crash but generate incorrect result.

hzxa21 (approved):
Rest LGTM

src/storage/src/hummock/iterator/skip_watermark.rs
st1page: > the watermark column is always a single column or the order type

Yes and I think it should be assert when `watermark_col_serde` is generated?

PR #19894, opened 2024-12-23
feat(optimizer): introduce watermark group

stdrc:

### Background

Previously we don't record the derivation relation among different watermark columns. For instance, `window_start` and `window_end` generated by `tumble`/`hop` time window functions are always related to each other and the original event time column, however when we plan for an EOWC query, we lost this information and hence couldn't succeed if more than 1 of these exist (not pruned) at the same time.

### Changes

This PR introduces a new way to represent watermark columns in plan nodes -- watermark groups. In this PR we organize watermark columns into several groups, in each of which, all watermark columns are related. With this, when planning EOWC queries, we can handle multiple watermark columns as long as they are in the same one group. For example in the following query, we won't warn our user that there're more than one watermark column any more.

```sql
create materialized view mv as
select
  foo, ts, window_start, window_end
from tumble(t, ts, interval '5 mins')
emit on window close;
```

Among the changes, ~400 LoC are real, others are planner tests. Basically the core changes all happen in `StreamXxx::new`, with watermark derivation logic unchanged. The main contribution is to determine which columns among all output watermark columns are related for every stream node.

### Future work

1. We still use `FixedBitSet` to record a flattened `watermark_columns` field in `TableCatalog` and `TableDesc`, and hence in `Table` protobuf. As a result, When we create MV on MV, all the watermark columns in the base MV will be considered belonging to different watermark groups no matter whether they are actually related. I will change the table catalog field later to fix this.
2. We still can't do `create mv as select window_start, window_end, count(*) as cnt from tumble(t, ts, ...) group by window_start, window_end emit on window close`, because I don't want to have too many changes in this single PR. Will modify `StreamHashAgg` later to support it. (Resolved in https://github.com/risingwavelabs/risingwave/pull/19998)
3. Since we don't support duplicated plan node reusing in streaming plan, when we have two side of a join both scanning the same table with join condition `left.window = right.window`, we cannot know that the two window columns are derived from the same source watermark column. This is because the watermark group ID is allocated and assigned during optimization phase and not persisted.

## Checklist

## Documentation

<details>

</details>


stdrc:
* **#19894** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/19894?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/19894?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

chenzl25 (commented):
> Since we don't support shared node in streaming plan (IIRC),

Do you want this self-join plan with `StreamShare`?
https://github.com/risingwavelabs/risingwave/blob/c7a4a3943691b2f9b48eddd63741f561a6dcb145/src/frontend/planner_test/tests/testdata/output/share.yaml#L19-L55

src/frontend/planner_test/tests/testdata/output/watermark_group.yaml
```diff
@@ -0,0 +1,97 @@
...
+    └─StreamEowcSort { sort_column: $expr1 }
+      └─StreamProject { exprs: [(t.foo + $expr2) as $expr4, $expr1, $expr3, t._row_id, t._row_id], output_watermarks: [[$expr1], [$expr3]] }
+        └─StreamHashJoin [window, append_only] { type: Inner, predicate: $expr1 = $expr3, output_watermarks: [[$expr1], [$expr3]], output: [t.foo, $expr1, $expr2, $expr3, t._row_id, t._row_id] }
+          ├─StreamExchange { dist: HashShard($expr1) }
+          │ └─StreamProject { exprs: [t.foo, TumbleStart(t.ts, '00:05:00':Interval) as $expr1, t._row_id], output_watermarks: [[$expr1]] }
+          │   └─StreamTableScan { table: t, columns: [t.foo, t.ts, t._row_id], stream_scan_type: ArrangementBackfill, stream_key: [t._row_id], pk: [_row_id], dist: UpstreamHashShard(t._row_id) }
+          └─StreamExchange { dist: HashShard($expr3) }
+            └─StreamProject { exprs: [(t.foo + 1:Int32) as $expr2, TumbleStart(t.ts, '00:10:00':Interval) as $expr3, t._row_id], output_watermarks: [[$expr3]] }
+              └─StreamTableScan { table: t, columns: [t.foo, t.ts, t._row_id], stream_scan_type: ArrangementBackfill, stream_key: [t._row_id], pk: [_row_id], dist: UpstreamHashShard(t._row_id) }
```
stdrc, line 97: @chenzl25 In this plan, is it possible to use a shared `StreamTableScan` on the two sides?

chenzl25: You can use a cte to construct a share operator. 
```
create materialized view v as  
with cte as (select foo, window_start as win from tumble(t, ts, interval '5 mins'))
select l.foo + r.foo as foo, l.win as win1, r.win as win2
from cte as l join cte as r on l.win = r.win;
```

stdrc: Thanks!

BugenZhao (approved):
LGTM!

src/frontend/src/optimizer/plan_node/stream_changelog.rs
```diff
@@ -34,21 +34,14 @@ impl StreamChangeLog {
     pub fn new(core: generic::ChangeLog<PlanRef>) -> Self {
         let input = core.input.clone();
         let dist = input.distribution().clone();
-        let input_len = input.schema().len();
         // Filter executor won't change the append-only behavior of the stream.
```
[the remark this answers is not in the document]
stdrc: Seems to be a blind copy-paste

PR #19895, opened 2024-12-23
feat(single-node): add memory allocation

fuyufjh:

Follow up of #19477. This makes it work in `single-node` mode.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/cmd_all/src/single_node.rs
+++ b/src/cmd_all/src/single_node.rs
@@ -87,7 +87,7 @@ impl SingleNodeOpts {
 #[derive(Eq, PartialOrd, PartialEq, Debug, Clone, Parser)]
 pub struct NodeSpecificOpts {
     // ------- Compute Node Options -------
-    /// Total available memory for the compute node in bytes. Used by both computing and storage.
+    /// Total available memory for all the nodes
     #[clap(long)]
     pub total_memory_bytes: Option<usize>,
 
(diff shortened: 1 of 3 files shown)


src/cmd_all/src/single_node.rs
```diff
@@ -207,6 +209,16 @@ pub fn map_single_node_opts_to_standalone_opts(opts: SingleNodeOpts) -> ParsedSt
     frontend_opts.meta_addr = meta_addr.parse().unwrap();
     compactor_opts.meta_address = meta_addr.parse().unwrap();
 
+    // Allocate memory for each node
+    let system_total_mem = system_memory_available_bytes();
+    frontend_opts.frontend_total_memory_bytes = memory_for_frontend(system_total_mem);
+    compactor_opts.compactor_total_memory_bytes = memory_for_compactor(system_total_mem);
+    compute_opts.total_memory_bytes = system_total_mem
+        - memory_for_frontend(system_total_mem)
+        - memory_for_compactor(system_total_mem);
```
hzxa21, line 218: I had a discussion of @Li0k recently on the memory allocation between CN and compator in standalone. 
Let's say `CN_MEM = x * Total_MEM`, `Compactor_MEM = y * Total_MEM`, `FE_MEM = z * Total_MEM`, yhe choices are
1. `x + y + z = 1`. 
    - Pros: more likely to avoid OOM.
    - Cons: less efficient on memory usage because when loads on compaction/FE is not large, the memory will be wasted and cannot used by CN.

2. `x + y + z > 1`
    - Pros: More efficient on memory usage because CN operator cache can use the idle memory when compactor/FE is not fully loaded.
    - Cons: More vulnerable to OOM because CN operator cache eviction is lazy.


I am more leaning towards 2 because it fits more to the CN dynamic operator cache design. For example, we can have `x = y = z = 0.8` as well as the gradient allocation. WDYT?

fuyufjh: CN operator cache is special - keep in mind that it's controlled according to the **process-level** jemalloc statistics.

Here the 3 options

- `frontend_opts.frontend_total_memory_bytes`
- `compactor_opts.compactor_total_memory_bytes`
- `compute_opts.total_memory_bytes`

Mostly decides decides the storage memory, including uploading buffer and meta & block cache:

- `frontend_opts.frontend_total_memory_bytes` ([code here](https://github.com/risingwavelabs/risingwave/blob/2c0d286ddb2ef25e722dc330550eec938676d2fd/src/frontend/src/session.rs#L453-L458))
   - `batch_memory_limit`
- `compactor_opts.compactor_total_memory_bytes` ([code here](https://github.com/risingwavelabs/risingwave/blob/f00385991226042ace31eff9bf7bc3ba6d3a0366/src/storage/compactor/src/server.rs#L88-L92))
   - `meta_cache_capacity_bytes`
   - `compactor_memory_limit_bytes`
- `compute_opts.total_memory_bytes` decides ([code here](https://github.com/risingwavelabs/risingwave/blob/7005c0591e7f2f58c28d8615c87d7aab3d604f8f/src/compute/src/server.rs#L142-L164))
   - `block_cache_capacity_mb`
   - `meta_cache_capacity_mb`
   - `shared_buffer_capacity_mb`
   - (embedded) `compactor_memory_limit_mb`
 
As you are more familiar with storage, I'll follow your decisions.

Besides, it doesn't count the Meta's memory usage (because there is no such as option), so actually it's already `> 1`.

hzxa21 (approved):

src/cmd_all/src/single_node.rs
Li0k: According to the example above 

```
x = y = z = 0.8
```

Total memory is well over 1.0 (0.8 * 3 = 2.4)
My concern is that when we use “excessive” configurations, it will cause the CN's memory to be evicted faster (according to jemalloc statistics), which will severely weaken the CN's capabilities.

Li0k (approved):
LGTM

src/cmd_all/src/single_node.rs
Li0k: I'm also in favor of option 2, but it seems that the limit should not be exceeded. For example, it is only allowed to exceed 0.5 or less, in order to guarantee the capacity of the CN.

- The memory footprint of the compactor is not a long term “cache”, so it is freed when it is used up, and it can be more “over-occupied”.
- The FE footprint seems to come from batch query? I'm not sure it's released in time cc @fuyufjh @hzxa21

PR #19695, opened 2024-12-06
fix: double RUST_MIN_STACK to resolve stack overflow

tabVersion:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://gist.github.com/TennyZhuang/f00be7f16996ea48effb049aa7be4d66#file-rw_cla).

## What's changed and what's your intention?

when running SQL planner test, we meet a stack overflow panic https://buildkite.com/risingwavelabs/pull-request/builds/63818#0193962f-47f4-484f-bd3e-21aa7aa25ae5
and same observation by @Li0k 

## Checklist

- [ ] I have written necessary rustdoc comments
- [ ] I have added necessary unit tests and integration tests
- [ ] I have added test labels as necessary. See [details](https://github.com/risingwavelabs/risingwave/blob/main/docs/developer-guide.md#ci-labels-guide).
- [ ] I have added fuzzing tests or opened an issue to track them. (Optional, recommended for new SQL features #7934).
- [ ] My PR contains breaking changes. (If it deprecates some features, please create a tracking issue to remove them in the future).
- [ ] All checks passed in `./risedev check` (or alias, `./risedev c`)
- [ ] My PR changes performance-critical code. (Please run macro/micro-benchmarks and show the results.)

- [ ] My PR contains critical fixes that are necessary to be merged into the latest release. (Please check out the [details](https://github.com/risingwavelabs/risingwave/blob/main/CONTRIBUTING.md))

## Documentation

- [ ] My PR needs documentation updates. (Please use the **Release note** section below to summarize the impact on users)

## Release note

If this PR includes changes that directly affect users or other significant modifications relevant to the community, kindly draft a release note to provide a concise summary of these changes. Please prioritize highlighting the impact these changes will have on users.

</details>

PR #19911, opened 2024-12-24
feat(pb): box stream NodeBody to reduce stack memory usage

### LINKED ISSUE #19910: frontend stack overflow at stream fragmenter
To reproduce: TPCH q5

It's not stack depth too deep, but memory usage too large. So #16279 doesn't help.
(On mac, default stack size is 8MB.)

Analysis:
- stream `NodeBody`: 4120 bytes. (BTW, batch node body is 680 bytes)
    - largest is `HashJoin` 4120 bytes, 2nd largest is DyanamicFilter` 2144 bytes. 
    - small ones: `Project` 112 bytes, `Filter` 336 bytes
    - It's an unbalanced `enum`, so we will use 4120 bytes for all nodes.
- The main reason of the large plan node is `Table`: 904 bytes (`HashJoin` has 4 tables)
    - Largest problem: recently introduced `webhook_info` is 344 bytes. (That's why we hit the stackoverflow problem recently but not before)
        - This is because `ExprNode` is 328 bytes, where `DataType` is 72 bytes and `RexNode` is 248 bytes.
        - For `RexNode`, `Udf` is 248 bytes (!), and other branch is 32 bytes.
    - Some other minor issues: `Option<u32>` wasted 4 bytes, since the zero value is not used, or we don't need this large range. We may use `i32` or `u32` instead.
    - 41 fields in total! @st1page said that this was a mistake: we should pub sth like `TableDesc` in plan, instead of `TableCatalog`.

Mitigate:
- make `expr.ExprNode.rex_node.udf` boxed: `ExprNode` becomes 112 bytes, `Table` 688 bytes, `NodeBody` 3040 bytes. So this seems not enough.
- Currently the easiest way seems to be just boxing all variants of `NodeBody`. I'm not sure whether this is optimal.
- Alternatively we can box all occurences of `Table`

xxchan:

fix #19910 See analysis in the issue

also revert #19695

## Checklist

## Documentation

<details>

</details>


src/prost/src/lib.rs
```diff
@@ -551,4 +560,11 @@ mod tests {
         };
         assert!(!new_data_type.is_nullable);
     }
+
+    #[test]
+    fn test_size() {
+        use static_assertions::const_assert_eq;
+
+        const_assert_eq!(std::mem::size_of::<NodeBody>(), 16);
```
xxchan, line 568: 4120 bytes -> 16 bytes

src/prost/build.rs
```diff
@@ -112,6 +112,57 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
             "#[derive(::enum_as_inner::EnumAsInner)]",
         )
         .btree_map(btree_map_paths)
+        // node body is a very large enum, so we box it to avoid stack overflow.
+        // TODO: ideally we should box all enum variants automatically https://github.com/tokio-rs/prost/issues/1209
+        .boxed(".stream_plan.StreamNode.node_body.source")
```
fuyufjh, line 117: Why not `.boxed(".stream_plan.StreamNode")`?

fuyufjh (approved):

stdrc (approved):
LGTM

MrCroxx (approved):
LGTM

st1page (approved):

src/prost/build.rs
```diff
@@ -112,6 +112,57 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+        .boxed(".stream_plan.StreamNode.node_body.project")
+        .boxed(".stream_plan.StreamNode.node_body.filter")
+        .boxed(".stream_plan.StreamNode.node_body.materialize")
```
chenzl25, line 120: Should we box the batch plan node as well?

src/prost/build.rs
xxchan: Sounds can also fix the stackoverflow. 🤔 But the enum itself still wastes memory.

src/prost/build.rs
xxchan: batch node body was 680 bytes. After boxing Udf in Expr, it's 464 bytes. Sounds acceptable

xxchan: LocalLookupJoinNode is the largest

src/prost/build.rs
BugenZhao: Listing all possible variants here seems verbose. Is this the only approach to achieve the goal?

src/prost/src/lib.rs
```diff
@@ -551,4 +562,11 @@ mod tests {
...
```
BugenZhao, line 570: May add some comments to instruct future developers to box the newly introduced variant if this assertion fails.

Also, can use `static_assertions` from the standard library and there's no need to put it inside a `#[test]` function.

BugenZhao (approved):

src/prost/build.rs
xxchan: can implement this upstream https://github.com/tokio-rs/prost/issues/1209

chenzl25 (approved):

Li0k (approved):

src/prost/src/lib.rs
xxchan: Where is the static_assertions in the standard library ?

BugenZhao: Oh, there's not. I confused it with `assert_matches`. 😄

Issue #18312, opened 2024-08-29
Support configuration change in a single barrier


wenym1:
## Motivation

Currently, when doing configuration change, such as scaling, replace table (including sink into table), we follow the following steps:
1. inject a pause barrier, let's say epoch1, and after the barrier is collected and committed, call `try_wait_epoch(epoch1)` explicitly for each CN
2. inject a configuration change barrier, let's say epoch2, to apply the configuration. For newly created stateful executors, they call `StateTable::init` on epoch2, for existing stateful executors, they call `update_vnode_bitmap`.
3. inject a resume barrier to resume the stream

The reason for this design is mainly for data synchronization, which means, each newly created actors, or actor that is assigned with new vnode, can see the latest data of the vnodes it owns. The mechanism works, because we call `try_wait_epoch(epoch1)` in each CN, so that all created actors and any actor created later can at least read all data at epoch1, and since we have paused the stream after epoch1, data is the same in epoch1 and epoch2, and when configuration change happens in epoch2, all actors can read all latest data.

Though the current mechanism works, it has many drawbacks:
1.  We pause the stream explicitly, and during which concurrent barriers become useless
2. The correctness depends on a too large scope. First, it depends on all executors to handle the pause barrier correctly, to ensure that there is actually no data between epoch1 and epoch2. Second, it depends on correctly calling `try_wait_epoch`. Sometimes, for example https://github.com/risingwavelabs/risingwave/issues/17602, the correctness implicitly depends on calling `try_wait_epoch`, but we didn't do it, and the error is either hard to reproduced, or happens in somewhere far away in downstream, which cost many efforts to debug.

Since the main purpose of this mechanism is for data synchronization, we'd better narrow the scope that the correctness depends on. In solely the implementation of `StateTable`, we can ensure that data are always ready when `StateTable::Init` and `StateTable::update_vnode_bitmap`, without any implicit dependency on the correct behavior of higher level streaming executor or the streaming runtime. 

## Proposal

As discussed offline with @BugenZhao and @st1page , we have the following rough proposal.

When we call `StateTable::Init` and `StateTable::update_vnode_bitmap`, the state table itself will wait the committed epoch of the table id to bump up before return, and after the methods return, the state table is ready to serve the read request.

[...]

PR #19720, opened 2024-12-09
feat(snapshot-backfill): introduce state to snapshot backfill

wenym1:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://gist.github.com/TennyZhuang/f00be7f16996ea48effb049aa7be4d66#file-rw_cla).

## What's changed and what's your intention?

In this PR, we will introduce state to track the progress of snapshot backfill. The progress state table will have the following schema
```
| vnode | `epoch` | `row_count` | `is_epoch_finished` | pk ...
```

* The pk of the state is `vnode`, which means we track the snapshot backfill progress in per-vnode way. 
* `epoch` is the epoch that the `vnode` is currently consuming on. We can assume that `epoch >= snapshot_epoch`. When `epoch == snapshot_epoch`, the `vnode` will be consuming the snapshot, and when `epoch > snapshot_epoch`, the `vnode` will be consuming the change log on `epoch`.
* `row_count` is the number of rows that we have consumed in this epoch. The `row_count` is measured in per-vnode manner.
* `is_epoch_finished` means whether the vnode has finished consuming the `epoch`.
* `pk...` is the current progress of consuming the epoch. When the fields only make sense when `is_epoch_finished` is `false`. Note that the `pk...` is the `pk` of the first unconsumed row, rather than the `pk` of the last consumed row, which is different to the one in arrangement backfill.

With the `VnodeStream` implemented in https://github.com/risingwavelabs/risingwave/pull/19936, we can get the current progress of each vnode before the stream is finished. When barrier is received, we can first drain the buffer, and then use `Peekable::peek` to get the first unconsumed row, and then save the `pk` to the backfill progress state.

In this PR, we only write the progress state, and only read the state for sanity check. In future PR, we will read the progress state to support recoverable snapshot backfill.

## Checklist

- [ ] I have written necessary rustdoc comments
- [ ] I have added necessary unit tests and integration tests
- [ ] I have added test labels as necessary. See [details](https://github.com/risingwavelabs/risingwave/blob/main/docs/developer-guide.md#ci-labels-guide).
- [ ] I have added fuzzing tests or opened an issue to track them. (Optional, recommended for new SQL features #7934).
- [ ] My PR contains breaking changes. (If it deprecates some features, please create a tracking issue to remove them in the future).
- [x] All checks passed in `./risedev check` (or alias, `./risedev c`)
- [ ] My PR changes performance-critical code. (Please run macro/micro-benchmarks and show the results.)

[...]

PR #19931, opened 2024-12-25
feat(storage): lazily remove instance data in uploader to avoid data loss

wenym1:

Currently in uploader, when an local instance is destroyed, we will remove the instance from the uploader immediately, even if there is some unsync data in the instance. The logic works in current code, because an instance can be destroyed in 3 scenarios: drop streaming job, configuration change, and actor exit on error or recovery. When dropping streaming job, the table will be removed soon, and therefore the loss of data is invisible. In configuration change, before the barrier to drop actors, we will have a pause barrier to ensure that there is no data in between, so it works for configuration. 

However, when we support https://github.com/risingwavelabs/risingwave/issues/18312, there will not be a pause barrier, and then we may lose data if the sync is called later than the actor exits. So we need to fix it before implementing https://github.com/risingwavelabs/risingwave/issues/18312.

In this PR, we change to lazily remove the instance data. When the instance is destroyed, in uploader we only mark the instance as `destroyed`. The instance will later be destroyed when all the sealed epochs have been started syncing.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/storage/src/hummock/event_handler/uploader/mod.rs
+++ b/src/storage/src/hummock/event_handler/uploader/mod.rs
@@ -466,6 +466,7 @@ struct LocalInstanceUnsyncData {
     sealed_data: VecDeque<LocalInstanceEpochData>,
     // newer data comes first
     flushing_imms: VecDeque<SharedBufferBatchId>,
+    is_destroyed: bool,
 }
 
 impl LocalInstanceUnsyncData {
@@ -476,6 +477,7 @@ impl LocalInstanceUnsyncData {
             current_epoch_data: Some(LocalInstanceEpochData::new(init_epoch)),
             sealed_data: VecDeque::new(),
             flushing_imms: Default::default(),
+            is_destroyed: false,
         }
     }
 
@@ -610,6 +612,10 @@ impl LocalInstanceUnsyncData {
                 assert!(current_data.imms.is_empty() && !current_data.has_spilled);
             }
         }
+    }
+
+    fn is_finished(&self) -> bool {
+        self.is_destroyed && self.sealed_data.is_empty()
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/storage/src/hummock/event_handler/uploader/mod.rs
```diff
@@ -1414,28 +1435,7 @@ impl HummockUploader {
         let UploaderState::Working(data) = &mut self.state else {
             return;
         };
-        if let Some(removed_table_data) = data.unsync_data.may_destroy_instance(instance_id) {
-            data.task_manager.remove_table_spill_tasks(
```
hzxa21, line 1418: For "drop streaming job", I think it is still preferable to cancel the task instead of waiting the task to finsih because the data will be deleted anyway. Can you elabrate in which cases we should preserve the upload task instead of cancelling it on instance destruction?

wenym1:
* **#19720** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/19720?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#19931** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/19931?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/19931?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/storage/src/hummock/event_handler/uploader/mod.rs
wenym1: Just updated the PR description. 

In short, this PR is not for `drop streaming job`, but for the single barrier configuration change in https://github.com/risingwavelabs/risingwave/issues/18312. In configuration change that may concurrently stop some existing actors and create new actors, for actors to be dropped, let's say in epoch1, they should write epoch1 data to hummock, and their data should be included in the SSTs of epoch1. But in the current main branch, for these actor to be dropped, their data is not likely to be included in the SSTs of epoch1, because their data might be dropped immediately when the instance is destroyed but before `sync(epoch1)` is called. For the same reason, we should also preserve the spilling task.

The current main branch works because in configuration change, we have a pause barrier, let's say epoch0, and we can ensure that no data is written between epoch0 and epoch1, and therefore even the instances are dropped, we won't lose data.

For normal `drop streaming job`, it's fine to drop instance and cancel related spilling task and ignore data loss. But given that we cannot distinguish whether dropping an instance is caused by drop streaming job or configuration change, we may let it follow the same handling logic as configuration change. I think the benefit of uploading less data may not be worthy compared to the increased complexity. Besides, a spilling task usually mix data from different tables, which may include normal table and the table to be dropped. Therefore, the spilling task may not be really cancelled when we call the previous `remove_table_spill_tasks`.

hzxa21: >  a spilling task usually mix data from different tables, which may include normal table and the table to be dropped.

This is a good point. Thanks for the explanation.

hzxa21 (approved):
LGTM

PR #19932, opened 2024-12-25
refactor(stream): send barriers in batch when possible

zwang28:

Excessive barrier messages between dispatcher and merger have been shown to hinder performance, by consuming CPU resources on message polling. This issue is significant when parallelism is high, because a barrier would result in NxM barrier messages between dispatcher and merger. Please refer to  the benchmark attached below for further details.

This PR aims to reduce the number of barrier messages between dispatcher and merger, by combining consecutive non-mutation barriers into one message. 

This PR primarily reduces the total number of incoming message for both RemoteInput and LocalInput. However from the perspective of MergeExecutor, the total number of message to poll doesn't change since BarrierBatch has been reverted to multiple Barriers for the output RemoteInput and LocalInput. So regarding the source of performance gain in this PR, I see two:
- A batch of barriers only requires 1 barrier permit, consequently there're less permits to add back.
- Although the total number of messages remains unchanged for the MergeExecutor, the cost of polling a message can be reduced due to [this very simple new state](https://github.com/risingwavelabs/risingwave/blob/366a139fe42de6bf2e7d239d7ac7ea14c054a97c/src/stream/src/executor/exchange/input.rs#L301) in the auto-generated future, i.e. less useless poll::pending.

[More discussion here](https://risingwave-labs.slack.com/archives/C056MK4GU2X/p1731483977171209).

## Checklist

## Documentation

<details>

</details>


zwang28:
The cluster is running a release build, deployed via risedev in a 16c32GB EC2: https://github.com/risingwavelabs/risingwave/commit/d7b5cf5af0230e3370932932acb45717aa7ff4b3

Screenshot 2024-12-25 at 15 22 52

src/stream/src/executor/dispatch.rs
```diff
@@ -110,56 +112,63 @@ struct DispatchExecutorInner {
 }
 
 impl DispatchExecutorInner {
-    async fn dispatch(&mut self, msg: Message) -> StreamResult<()> {
+    async fn dispatch(&mut self, msg: MessageBatch) -> StreamResult<()> {
         let limit = (self.context.config.developer).exchange_concurrent_dispatchers;
-
+        // Only barrier can be batched for now.
```
hzxa21, line 117: Not strongly related to this PR: I wonder whether peaking the message including barrier/chunk and batch them by size into one gRPC message can be a general optimization to amortize network cost.

Theoretically it is http2 and tcp under the hook so the network package should be batched already but given that the barrier batch optimization works, I wonder whether there are other overheads that we are not aware of in gRPC.

I also saw a few asks for gRPC streaming rpc message batching elsewhere:
- https://github.com/grpc/grpc-go/issues/524
- https://github.com/grpc/grpc-dotnet/issues/2074

zwang28: > but given that the barrier batch optimization works

I think this PR primarily reduces the total number of poll from `RemoteInput` and `LocalInput` in `MergeExecutor`.

zwang28: I believe the bottleneck lies in the [SelectReceivers](https://github.com/risingwavelabs/risingwave/blob/366a139fe42de6bf2e7d239d7ac7ea14c054a97c/src/stream/src/executor/merge.rs#L388), which aligns multiple inputs.

src/stream/src/executor/exchange/input.rs
```diff
@@ -358,3 +362,16 @@ pub(crate) fn new_input(
 
     Ok(input)
 }
+
+impl DispatcherMessageBatch {
```
wenym1, line 366: Can change to return with an iterator with [auto_enums](https://docs.rs/auto_enums/latest/auto_enums/) to avoid creating a vec for each message.

```
impl DispatcherMessageBatch {
    fn into_messages(self) -> impl Iterator<Item = DispatcherMessage> {
        #[auto_enum(Iterator)]
        match self {
            DispatcherMessageBatch::BarrierBatch(barriers) => barriers
                .into_iter()
            DispatcherMessageBatch::Chunk(c) => once(DispatcherMessage::Chunk(c)),
            DispatcherMessageBatch::Watermark(w) => once(DispatcherMessage::Watermark(w)),
        }
    }
}
```

src/common/src/config.rs
```diff
@@ -1150,6 +1150,9 @@ pub struct StreamingDeveloperConfig {
     /// When true, all jdbc sinks with connector='jdbc' and jdbc.url="jdbc:postgresql://..."
     /// will be switched from jdbc postgresql sinks to rust native (connector='postgres') sinks.
     pub switch_jdbc_pg_to_native: bool,
+
+    #[serde(default = "default::developer::stream_max_barrier_batch_size")]
```
BugenZhao, line 1154: Would you add some docs?

src/stream/src/executor/exchange/input.rs
```diff
@@ -358,3 +362,18 @@ pub(crate) fn new_input(
...
+    fn into_messages(self) -> impl Iterator<Item = DispatcherMessage> {
+        #[auto_enums::auto_enum(Iterator)]
```
BugenZhao, line 368: I think we can simply adopt `Either` instead of creating a new TAIT which may confuses IDEs.

src/stream/src/executor/mod.rs
```diff
@@ -1046,6 +1046,29 @@ impl<M> MessageInner<M> {
...
+    BarrierBatch(Vec<BarrierInner<BarrierMutationType>>),
+    Watermark(Watermark),
+}
+pub type DispatcherBarriers = Vec<DispatcherBarrier>;
+#[derive(Debug, EnumAsInner, PartialEq, Clone)]
+pub enum DispatcherMessageBatch {
+    Chunk(StreamChunk),
+    BarrierBatch(Vec<BarrierInner<()>>),
+    Watermark(Watermark),
+}
```
BugenZhao, line 1061: Can we unify these types with type parameter?

Also, would you add some docs?

src/stream/src/executor/mod.rs
```diff
@@ -1046,6 +1046,29 @@ impl<M> MessageInner<M> {
 
 pub type Message = MessageInner<BarrierMutationType>;
 pub type DispatcherMessage = MessageInner<()>;
+#[derive(Debug, EnumAsInner, PartialEq, Clone)]
+pub enum MessageBatch {
+    Chunk(StreamChunk),
+    BarrierBatch(Vec<BarrierInner<BarrierMutationType>>),
```
BugenZhao, line 1052: Do we have some empirical idea of the length of barrier batch? If it's small, shall we consider `SmallVec` here?

src/stream/src/executor/dispatch.rs
```diff
@@ -389,37 +398,85 @@ impl StreamConsumer for DispatchExecutor {
...
-            #[for_await]
-            for msg in input {
+            let mut input = self.input.execute().peekable();
+            let mut end_of_stream = false;
+            while !end_of_stream {
+                let Some(msg) = input.next().await else {
+                    end_of_stream = true;
+                    continue;
+                };
+                let mut barrier_batch = vec![];
```
BugenZhao, line 417: Is it possible to decouple the logic for batching messages from dispatching them?

src/stream/src/executor/dispatch.rs
```diff
@@ -389,37 +398,85 @@ impl StreamConsumer for DispatchExecutor {
     type BarrierStream = impl Stream<Item = StreamResult<Barrier>> + Send;
 
     fn execute(mut self: Box<Self>) -> Self::BarrierStream {
+        let barrier_batch_size = self
+            .inner
+            .context
+            .config
+            .developer
+            .max_barrier_batch_size
+            .saturating_sub(1);
```
BugenZhao, line 407: Why do we `sub` here?

src/stream/src/executor/mod.rs
zwang28: During my tests, it remains in single digits for most of the time.

src/stream/src/executor/dispatch.rs
zwang28: The resulting `barrier_batch_size` indicates that at most `barrier_batch_size` additional barriers will be peeked at after the first one.

src/stream/src/executor/mod.rs
BugenZhao: It turns out the stack size of `Barrier` is considerable, so `SmallVec<[Barrier; 8]>` can be significantly large.

src/stream/src/executor/dispatch.rs
zwang28: Currently it's done in following orders:
1. Poll one msg.
2. Peek N consecutive barriers, only if the msg from step 1 is non-mutation. Stop on any mutation barrier as well.
3. Send the batch msg.

Could you please elaborate "decouple"?

fuyufjh (approved):
Rubber stamp. Any concerns?

src/stream/src/executor/dispatch.rs
BugenZhao: For example, encapsulate step 1 and step 2 into a function, so that the "send" part can be cleaner.

zwang28: Refactored.

zwang28:
I'm verifying nexmark performance with this PR.

src/stream/src/executor/dispatch.rs
```diff
@@ -389,44 +399,112 @@ impl StreamConsumer for DispatchExecutor {
...
+            // end_of_stream
+            break;
+        };
+        let Ok(Message::Barrier(barrier)) = msg else {
+            break;
+        };
+        if barrier.mutation.is_some() {
+            break;
+        }
+        let msg: Message = input.next().await.unwrap()?;
```
BugenZhao, line 499: I think we can also `now_or_never` on the `next()` future?

BugenZhao (approved):
LGTM!

PR #19936, opened 2024-12-25
refactor(snapshot-backfill): extract common logic of consuming snapshot and log store

wenym1:

Part of https://github.com/risingwavelabs/risingwave/pull/19720.

For snapshot backfill, in the main branch and the future PR https://github.com/risingwavelabs/risingwave/pull/19720, the stage of consuming snapshot and consuming log store have the similar logic of consuming a stream of row with op and convert the stream into stream chunk. In this PR, we will extra the common logic and let them share the same code.

Previously, the row streams of multiple vnodes are combined in the stream returned from `StorageTable` method. In this PR we will change to combine the stream of multiple vnodes outside `StorageTable` with `FuturesUnordered`, so that we can access each vnode stream and get the progress of each vnode, which will be useful in https://github.com/risingwavelabs/risingwave/pull/19720.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/executor/backfill/snapshot_backfill/executor.rs
+++ b/src/stream/src/executor/backfill/snapshot_backfill/executor.rs
@@ -15,31 +15,33 @@
 use std::cmp::min;
 use std::collections::VecDeque;
 use std::future::{pending, ready, Future};
-use std::mem::{replace, take};
+use std::mem::take;
 use std::sync::Arc;
 
 use anyhow::anyhow;
-use futures::future::Either;
-use futures::{pin_mut, Stream, TryStreamExt};
+use futures::future::{try_join_all, Either};
+use futures::{pin_mut, Stream, TryFutureExt, TryStreamExt};
 use risingwave_common::array::{Op, StreamChunk};
+use risingwave_common::hash::VnodeBitmapExt;
 use risingwave_common::metrics::LabelGuardedIntCounter;
 use risingwave_common::row::OwnedRow;
-use risingwave_common::util::chunk_coalesce::DataChunkBuilder;
+use risingwave_common::util::epoch::EpochPair;
 use risingwave_hummock_sdk::HummockReadEpoch;
-use risingwave_storage::error::StorageResult;
+use risingwave_storage::store::PrefetchOptions;
 use risingwave_storage::table::batch_table::storage_table::StorageTable;
 use risingwave_storage::table::ChangeLogRow;
 use risingwave_storage::StateStore;
 use tokio::select;
 use tokio::sync::mpsc::UnboundedReceiver;
 
+use crate::executor::backfill::snapshot_backfill::vnode_stream::VnodeStream;
 use crate::executor::backfill::utils::{create_builder, mapping_message};
 use crate::executor::monitor::StreamingMetrics;
 use crate::executor::prelude::{try_stream, StreamExt};
 use crate::executor::{
-    expect_first_barrier, ActorContextRef, BackfillExecutor, Barrier, BoxedMessageStream,
-    DispatcherBarrier, DispatcherMessage, Execute, MergeExecutorInput, Message,
-    StreamExecutorError, StreamExecutorResult,
+    expect_first_barrier, ActorContextRef, Barrier, BoxedMessageStream, DispatcherBarrier,
+    DispatcherMessage, Execute, MergeExecutorInput, Message, StreamExecutorError,
+    StreamExecutorResult,
 };
 use crate::task::CreateMviewProgressReporter;
 
@@ -128,6 +130,8 @@ impl<S: StateStore> SnapshotBackfillExecutor<S> {
                 false
             }
         };
+        let first_recv_barrier_epoch = first_recv_barrier.epoch;
+        yield Message::Barrier(first_recv_barrier);
 
         let (mut barrier_epoch, mut need_report_finish) = {
             if should_backfill {
@@ -162,7 +166,7 @@ impl<S: StateStore> SnapshotBackfillExecutor<S> {
                             self.rate_limit,
                             &mut self.barrier_rx,
                             &mut self.progress,
-                            first_recv_barrier,
+                            first_recv_barrier_epoch,
                         );
 
                         pin_mut!(snapshot_stream);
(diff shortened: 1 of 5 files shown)


wenym1:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/19936?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#19720** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/19720?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#19936** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/19936?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/19936?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#19931** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/19931?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/stream/src/executor/backfill/snapshot_backfill/executor.rs
```diff
@@ -515,22 +482,69 @@ async fn receive_next_barrier(
...
-    }
+async fn make_log_stream(
+    upstream_table: &StorageTable<impl StateStore>,
+    prev_epoch: u64,
+    start_pk: Option<OwnedRow>,
+    chunk_size: usize,
+) -> StreamExecutorResult<VnodeStream<impl super::vnode_stream::BackfillRowStream>> {
+    let data_types = upstream_table.schema().data_types();
+    let start_pk = start_pk.as_ref();
+    let vnode_streams = try_join_all(upstream_table.vnodes().iter_vnodes().map(move |vnode| {
```
kwannoel, line 493: I suppose we don't iterate by vnode. Which means there can be poor locality when backfilling, since we interleave results from different streams.

Can we make a note of it here, in case we need to optimize this case down the road?

src/stream/src/executor/backfill/snapshot_backfill/vnode_stream.rs
```diff
@@ -0,0 +1,198 @@
...
+}
+
+impl<St: Stream + Unpin> Stream for StreamWithVnode<St> {
+    type Item = St::Item;
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        self.stream.poll_next_unpin(cx)
+    }
+}
+
```
kwannoel, line 45: Can we adapt the docs from the PR and add them here? To my understanding the reason we need a custom stream, rather than rely on a `try_stream` macro, is because we need to encapsulate some state inside the stream as well.

It is possible to do it with `try_stream`, but the internal state of the sink will be exposed.

src/stream/src/executor/backfill/snapshot_backfill/vnode_stream.rs
```diff
@@ -0,0 +1,198 @@
...
+                        assert_eq!(pk, extra_row.project(pk_indices).to_owned_row());
+                    }
+                    on_vnode_progress(vnode_stream.vnode, Some(pk));
+                }
+                Some(Err(_)) => {
+                    return Err(vnode_stream
+                        .stream
+                        .try_next()
+                        .await
+                        .expect_err("checked Err"));
```
kwannoel, line 139: Add a more descriptive error message?

src/stream/src/executor/backfill/snapshot_backfill/vnode_stream.rs
```diff
@@ -0,0 +1,198 @@
...
+
+impl<St: BackfillRowStream> Stream for VnodeStream<St> {
+    type Item = StreamExecutorResult<StreamChunk>;
+
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        let this = self.get_mut();
+        let capacity = this.data_chunk_builder.batch_size();
+        loop {
+            match ready!(this.poll_next_row(cx)) {
+                Ok(Some(((op, row), extra))) => {
```
kwannoel, line 158: What's "extra" referring to? Perhaps you can add some docs to clarify the semantics.

src/stream/src/executor/backfill/snapshot_backfill/executor.rs
wenym1: The current implementation basically have the same behavior as the stream returned from `StorageTable::iter_with_encoded_key_range`, which merges the streams of multiple vnodes with  `streams.flatten_unordered(1024)`, and also interleave results from different streams as what we do in this PR.

In the future I think we can try limit the number of vnodes that we concurrently poll on.

src/stream/src/executor/backfill/snapshot_backfill/vnode_stream.rs
wenym1: It's hard to use the `try_stream` macro, because the ownership of the vnode streams will be owned by the stream generated by `try_stream`, and then we can't get the state of the vnode streams.

src/stream/src/executor/backfill/snapshot_backfill/vnode_stream.rs
wenym1: This is similar to `result.expect("...")`, but just expect that the result is `Err`.

src/stream/src/executor/backfill/snapshot_backfill/vnode_stream.rs
wenym1: Just changed the term `extra_xxx` to `second_xxx`, and added following comment to `BackfillRowItem`
```
/// Can be either one row or two rows. When having two rows, the two rows should be included in the same stream chunk.
/// The case of two rows is for combining the `UpdateDelete` and `UpdateInsert` of `ChangeLogValue::Update`
```

src/stream/src/executor/backfill/snapshot_backfill/vnode_stream.rs
kwannoel: Okay let's add some docs to provide this justification.

src/stream/src/executor/backfill/snapshot_backfill/vnode_stream.rs
```diff
@@ -0,0 +1,200 @@
...
+    type Item = StreamExecutorResult<StreamChunk>;
+
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        let this = self.get_mut();
+        let capacity = this.data_chunk_builder.batch_size();
+        loop {
+            match ready!(this.poll_next_row(cx)) {
+                Ok(Some(((op, row), second))) => {
+                    let may_chunk = if let Some((second_op, second_row)) = second {
+                        if this.data_chunk_builder.can_append(2) {
```
kwannoel, line 162:
```suggestion
                        if this.data_chunk_builder.can_append_update
```

Two rows will always correspond to update. This will make it more readable.

src/stream/src/executor/backfill/snapshot_backfill/vnode_stream.rs
```diff
@@ -0,0 +1,200 @@
...
+                                .expect("should be Some when not can_append");
+                            let ops = replace(&mut this.ops, Vec::with_capacity(capacity));
+                            this.ops.extend([op, second_op]);
+                            assert!(this.data_chunk_builder.append_one_row(row).is_none());
+                            assert!(this.data_chunk_builder.append_one_row(second_row).is_none());
+                            break Poll::Ready(Some(Ok(StreamChunk::from_parts(ops, chunk))));
+                        }
+                    } else {
+                        this.ops.push(op);
+                        this.data_chunk_builder.append_one_row(row)
```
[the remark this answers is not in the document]
kwannoel: Oh because when we hit `Ok(None)` below we always drain it.

Will be helpful to add this as a comment.

kwannoel (approved):
LGTM.

PR #19937, opened 2024-12-26
fix: support string to number type in JSONB_POPULATE_RECORD

### LINKED ISSUE #19918: JSONB_POPULATE_RECORD cannot convert jsonb strings to number type
### Describe the bug

`JSONB_POPULATE_RECORD` & `JSONB_POPULATE_RECORDSET` cannot convert jsonb strings to any SQL number type

Besides, parsing orders of magnitudes slower when schema set at source level versus reading in the raw jsonb and using a function like `JSONB_POPULATE_RECORD` in an MV (slack archive) 

### Error message/log

_No response_

### To Reproduce

_No response_

### Expected behavior

Better to `try_parse` the string instead of reporting error and leaving a `null`.

PostgresQL:

```
test=# CREATE TYPE test_table_type AS (id BIGINT);
CREATE TYPE
test=# select JSONB_POPULATE_RECORD(NULL::test_table_type, '{"id": "12345"}'::jsonb);
 jsonb_populate_record
-----------------------
 (12345)
(1 row)
```

_No response_

### The version of RisingWave

_No response_

### Additional context

_No response_

zwang28:

close #19918

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/common/src/types/jsonb.rs
+++ b/src/common/src/types/jsonb.rs
@@ -14,7 +14,6 @@
 
 use std::fmt::{self, Write};
 use std::hash::Hash;
-use std::str::FromStr;
 
 use bytes::{Buf, BufMut, BytesMut};
 use jsonbb::{Value, ValueRef};
@@ -325,16 +324,12 @@ impl<'a> JsonbRef<'a> {
     /// According to RFC 8259, only number within IEEE 754 binary64 (double precision) has good
     /// interoperability. We do not support arbitrary precision like PostgreSQL `numeric` right now.
     pub fn as_number(&self) -> Result<F64, String> {
-        if let Some(s) = self.0.as_str() {
-            F64::from_str(s).map_err(|e| format!("{}", e.as_report()))
-        } else {
-            self.0
-                .as_number()
-                .ok_or_else(|| format!("cannot cast jsonb {} to type number", self.type_name()))?
-                .as_f64()
-                .map(|f| f.into_ordered())
-                .ok_or_else(|| "jsonb number out of range".into())
-        }
+        self.0
+            .as_number()
+            .ok_or_else(|| format!("cannot cast jsonb {} to type number", self.type_name()))?
+            .as_f64()
+            .map(|f| f.into_ordered())
+            .ok_or_else(|| "jsonb number out of range".into())
     }
 
     /// This is part of the `->>` or `#>>` syntax to access a child as string.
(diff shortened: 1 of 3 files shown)


zwang28:
[Is decimal excluded here](https://github.com/risingwavelabs/risingwave/blob/3430b7df6befc3164bbf14b85f791e9a2334e89c/src/common/src/types/jsonb.rs#L429) by design?

@xiangjinwu

fuyufjh (approved):

src/common/src/types/jsonb.rs
```diff
@@ -323,12 +324,16 @@ impl<'a> JsonbRef<'a> {
...
     pub fn as_number(&self) -> Result<F64, String> {
-        self.0
-            .as_number()
-            .ok_or_else(|| format!("cannot cast jsonb {} to type number", self.type_name()))?
-            .as_f64()
-            .map(|f| f.into_ordered())
-            .ok_or_else(|| "jsonb number out of range".into())
+        if let Some(s) = self.0.as_str() {
+            F64::from_str(s).map_err(|e| format!("{e}"))
+        } else {
```
xiangjinwu: Keep `as_number` at its current behavior and let its caller (`JSONB_POPULATE_RECORD`) to do more work in addition to calling `as_number`.

Besides `JSONB_POPULATE_RECORD`, or actually before `JSONB_POPULATE_RECORD` is supported, this `as_number` is intended to support `CAST`. In PostgreSQL:
```
test=# select ('{"f1":4}'::jsonb -> 'f1')::int;
 int4 
------
    4
(1 row)

test=# select ('{"f1":"4"}'::jsonb -> 'f1')::int;
ERROR:  cannot cast jsonb string to type integer
```

That is, a JSON string shall NOT be `CAST`ed into integer. But it can be `POPULATE`ed into integer:
```
test=# select jsonb_populate_record(row(null::int), '{"f1":4}');
 jsonb_populate_record 
-----------------------
 (4)
(1 row)

test=# select jsonb_populate_record(row(null::int), '{"f1":"4"}');
 jsonb_populate_record 
-----------------------
 (4)
(1 row)
```

This is the PostgreSQL doc on `JSONB_POPULATE_RECORD`:
https://www.postgresql.org/docs/17/functions-json.html#id-1.5.8.22.8.13.2.2.9.1.2.1
> To convert a JSON value to the SQL type of an output column, the following rules are applied in sequence:
> * A JSON null value is converted to an SQL null in all cases.
> * If the output column is of type `json` or `jsonb`, the JSON value is just reproduced exactly.
> * If the output column is a composite (row) type, and the JSON value is a JSON object, the fields of the object are converted to columns of the output row type by recursive application of these rules.
> * Likewise, if the output column is an array type and the JSON value is a JSON array, the elements of the JSON array are converted to elements of the output array by recursive application of these rules.
> * Otherwise, if the JSON value is a string, the contents of the string are fed to the input conversion function for the column's data type.
> * Otherwise, the ordinary text representation of the JSON value is fed to the input conversion function for the column's data type.

That is, only `null`/`jsonb`/`object`/`array` get special treatment. Other cases (`bool`/`number`) are always `force_str` (function defined below) first and then `from_text` (defined [here](https://github.com/risingwavelabs/risingwave/blob/62818d32bb7a0ab59a9187c34545ec29b391520a/src/common/src/types/mod.rs#L903) as the input conversion function).

xiangjinwu:
> [Is decimal excluded here](https://github.com/risingwavelabs/risingwave/blob/3430b7df6befc3164bbf14b85f791e9a2334e89c/src/common/src/types/jsonb.rs#L429) by design?
> 
> @xiangjinwu

No. According to the PostgreSQL doc I just shared above, all data types having an _input conversion function_ is supported.

xiangjinwu:
The current implementation is still different from PostgreSQL but we may prefer simplicity over pg-compatibility here:
> * If the output column is a composite (row) type, **and the JSON value is a JSON object**, the fields of the object are converted to columns of the output row type by recursive application of these rules.
> * Likewise, if the output column is an array type **and the JSON value is a JSON array**, the elements of the JSON array are converted to elements of the output array by recursive application of these rules.

That is, when the output column is a struct or array, but the JSON value is a JSON string, it shall invoke the _input conversion function_. For example in PostgreSQL:
```
test=# select jsonb_populate_record(row(array[8]), '{"f1": "{2, 3}"}');
 jsonb_populate_record 
-----------------------
 ("{2,3}")
(1 row)
```

However, the _input conversion function_ for PostgreSQL struct and array is wired and [discouraged](https://www.postgresql.org/docs/17/rowtypes.html#:~:text=The%20ROW%20constructor%20syntax%20is%20usually%20easier%20to%20work%20with%20than%20the%20composite%2Dliteral%20syntax%20when%20writing%20composite%20values%20in%20SQL%20commands.) to [use](https://www.postgresql.org/docs/17/arrays.html#:~:text=The%20ARRAY%20constructor%20syntax%20(see%20Section%C2%A04.2.12)%20is%20often%20easier%20to%20work%20with%20than%20the%20array%2Dliteral%20syntax%20when%20writing%20array%20values%20in%20SQL%20commands.). Practically we do not need to support it here:
* `"{2, 3}"` as PostgreSQL array. Note the usage of `{}` instead of `[]` as in JSON.
* `"(2, true)"` as PostgreSQL struct. Note the usage of `()` and lack of field names, compared to `{}` in JSON.

Issue #19884, opened 2024-12-20
The brand new worker does not have any existing `ActorInfo` for the actors.


shanicky:
### Describe the bug

During the development of the `alter resource group` feature, I found that when an existing materialized view is altered to a completely new worker, the newly created actor cannot obtain the actor info of the upstream actor, resulting in a failure to start. 

After checking the code, it seems that only the newly created actor is broadcasted.

```
             broadcast_info: new_actors_location_to_broadcast.clone(),
```

This issue may not have existed before refactoring broadcast.

Issue #19536, opened 2024-11-22
chaos-mesh `podchaos` failure: `actor 64 not found in info table`


lmatz:
### Describe the bug

https://buildkite.com/risingwave-test/chaos-mesh/builds/1166#01934f59-c721-4cf0-8600-52124c801f6d

### Error message/log

```text
Opening a connection
Running command show create materialized view nexmark_q0_1;
Opening a connection
Running command CREATE MATERIALIZED VIEW nexmark_q0_1_chaos_mesh AS SELECT auction, bidder, price, date_time FROM bid
Failed to run the query
Caused by these errors (recent errors listed first):
  1: gRPC request to meta service failed: Internal error
  2: get error from control stream, in worker node 6
  3: gRPC request to stream service failed: Internal error
  4: failed to handle barrier event
  5: Actor 1136 exited unexpectedly
  6: actor 64 not found in info table
```
```


### To Reproduce

_No response_

### Expected behavior

_No response_

### How did you deploy RisingWave?

_No response_

### The version of RisingWave

_No response_

### Additional context

https://buildkite.com/risingwave-test/chaos-mesh/builds?branch=main

The first time appeared on Nov 13th

And repeatedly triggered in recent executions

**Comment by lmatz** [comment]
![SCR-20241122-hqg](https://github.com/user-attachments/assets/6d9f3842-c56e-445d-9db2-76b11a785d05)

**Comment by wenym1** [comment]
Chaos mech test has been successful since https://github.com/risingwavelabs/risingwave/pull/19579 is merged.

**Comment by wenym1** [comment]
Found a way to reproduce the bug locally thanks to @shanicky 

Launch a cluster and create a table
```
./risedev d ci-3cn-1fe-with-recovery

create table t(v int);
```

Run the following command to add a new worker node
```
.risingwave/bin/risingwave/compute-node --config-path .risingwave/config/risingwave.toml --listen-addr 127.0.0.1:5686 --prometheus-listener-addr 127.0.0.1:1225 --advertise-addr 127.0.0.1:5686 --async-stack-trace verbose --parallelism 6 --total-memory-bytes 8589934592 --role both --meta-address http://127.0.0.1:5690 --data-file-cache-dir .risingwave/data/foyer/5686/data --meta-file-cache-dir .risingwave/data/foyer/5686/meta
```

[...]

PR #19939, opened 2024-12-26
fix(barrier): broadcast actor info to newly added node

wenym1:

Root cause of https://github.com/risingwavelabs/risingwave/issues/19536.
Resolve https://github.com/risingwavelabs/risingwave/issues/19884

Previously for newly added node, we won't broadcast the existing actor info to the node. When actors that are spawned later on this node try to get the actor info in other nodes, it will get the `actor id not found` error.

In this PR, we fix this by including all the existing actor infos in the initial request for the newly add node. 

By the way, we also change the way to broadcast inflight subscription infos in recovery. Previously in recovery, we include the existing subscription infos in the initial request to CN. It works in global recovery, but in per-database recovery, we will not send this initial request, and then the existing subscriptions will be missed in the recovered database. In this PR, we change to broadcast the existing subscription infos in the initial inject barrier request. In this way, we can also broadcast the existing subscriptions in per-database recovery.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/proto/stream_service.proto
+++ b/proto/stream_service.proto
@@ -68,6 +68,7 @@ message StreamingControlStreamRequest {
   message InitialPartialGraph {
     uint32 partial_graph_id = 1;
     repeated stream_plan.SubscriptionUpstreamInfo subscriptions = 2;
+    repeated common.ActorInfo actor_infos = 3;
   }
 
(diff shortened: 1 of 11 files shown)


yezizp2012 (approved):
lgtm

PR #19950, opened 2024-12-27
feat(connector): add minio file scan type and enhance test

wcy-fdu:

Previously, `file_scan` only supported reading Parquet files on S3 and had no test coverage. This PR adds support for `file_scan(parquet, minio, ...)` and includes file scan test in the file source tests of the main cron job. Since MinIO is S3-compatible, it requires nearly the same parameters as S3, with the only difference being the need for a special endpoint setting when building the operator. Therefore, this PR reuses the existing file scan node. 

Note: File scan from MinIO is currently for testing purposes only.

Later, we will also add file_scan(parquet, gcs), and there may be a new plan node for that; we'll address that in subsequent PRs.

## Checklist

## Documentation

<details>

</details>


wcy-fdu:
Out of topic, interesting finding:
When creating an opendal operator to access minio
```
let mut builder = S3::default().bucket(&bucket).region(&minio_region);
    builder = builder.secret_access_key(&minio_access_key);
    builder = builder.secret_access_key(&minio_secret_key);
    builder = builder.endpoint("http://127.0.0.1:9301/");
```
will raise error 
```
  1: Unexpected (persistent) at stat, context: { called: reqsign::LoadCredential, service: s3, path: test_0.parquet } => loading credential to sign http request, source
  2: request to AWS EC2 Metadata Services failed: 
```
while 
```
 builder = builder
        .region(&s3_region)
        .access_key_id(&s3_access_key)
        .secret_access_key(&s3_secret_key)
        .bucket(&bucket);
```
can work normally, but the above two can both access s3 normally.

chenzl25:
> can work normally, but the above two can both access s3 normally.
> 
> @hzxa21 We discussed this offline before, just to keep you informed.

Maybe you can try these configs in opendal `disable_ec2_metadata`, `disable_config_load` and `allow_anonymous`.

src/connector/src/source/iceberg/parquet_file_handler.rs
```diff
@@ -109,16 +109,23 @@ pub fn new_s3_operator(
...
-    ));
-
+    let mut builder = S3::default();
+    builder = builder
+        .region(&s3_region)
+        .access_key_id(&s3_access_key)
+        .secret_access_key(&s3_secret_key)
+        .bucket(&bucket);
+    builder = match is_minio {
+        true => builder.endpoint(&format!("http://{}.127.0.0.1:9301", bucket)),
```
chenzl25, line 122: Should we allow users to provide s3 endpoint instead of hardcoding the endpoint?

wcy-fdu:
> `disable_ec2_metadata`, `disable_config_load` and `allow_anonymous`.

`disable_ec2_metadata`, `disable_config_load` and `allow_anonymous`, all tried, doesn't work.

src/connector/src/source/iceberg/parquet_file_handler.rs
wcy-fdu: I think minio is only for test, so I hardcode it. IMO, we don't public `file_scan(minio)` to user, what do you think?

wcy-fdu: >Should we allow users to provide s3 endpoint instead of hardcoding the endpoint?

And this will require another parameter in file scan, which is a bit redudant.

chenzl25: Can we reuse the s3 region parameter position to allow users to set the endpoint as well, i.e. (region or endpoint)?

wcy-fdu: region is a required field. If you change region to endpoint, the user's endpoint must include region. Is this a bit strange?

wcy-fdu: When testing minio, the region is `custom`, the endpoint is `http://bucket/127.0.0.1:9301`, they have no intersection.

chenzl25: If an endpoint is specified, the region field is useless (Correct?).  I mean `file_scan('parquet', 's3', region_or_endpoint, ...)` the `region_or_endpoint` parameter could be region or endpoint, we can detect it simply by `http` prefix.

wcy-fdu: Let me take a try.

wcy-fdu: >If an endpoint is specified, the region field is useless (Correct?)

No, for minio, both endpoint and region is needed.
without config region
```
  1: connector error
  2: ConfigInvalid (permanent) at Builder::build, context: { service: s3 } => region is missing. Please find it by S3::detect_region() or set them in env.
```
without endpoint, it will use default endpoint of aws instead of minio.
```
  1: Unexpected (persistent) at stat, context: { url: https://s3.custom.amazonaws.com/hummock001/test_0.parquet, called: http_util::Client::send, service: s3, path: test_0.parquet } => send http request, source
  2: error sending request for url (https://s3.custom.amazonaws.com/hummock001/test_0.parquet)
  3: client error (Connect)
  4: connection closed via error
```

chenzl25: Can we use a hardcode region in this case? e.g. `us-east1`

wcy-fdu: Modified.

src/connector/src/source/iceberg/parquet_file_handler.rs
```diff
@@ -110,14 +110,30 @@ pub fn new_s3_operator(
...
 ) -> ConnectorResult<Operator> {
-    // Create s3 builder.
-    let mut builder = S3::default().bucket(&bucket).region(&s3_region);
-    builder = builder.secret_access_key(&s3_access_key);
-    builder = builder.secret_access_key(&s3_secret_key);
-    builder = builder.endpoint(&format!(
-        "https://{}.s3.{}.amazonaws.com",
-        bucket, s3_region
-    ));
+    let is_minio = s3_region.starts_with("http");
```
chenzl25, line 113: Here is the executor. I think it's better to put this logic at src/frontend/src/optimizer/rule/table_function_to_file_scan_rule.rs

wcy-fdu: It won't work to put it in the `table_function_to_file_scan_rule`. I understand that you want this logic to be completed in the frontend, but the FileScanExecutor is constructed [directly from the proto](https://github.com/risingwavelabs/risingwave/blob/2601c237a7dd677c1314482f91720f4ea78841b5/src/batch/executors/src/executor/s3_file_scan.rs#L120-L131), and modifying it would require changes to the proto. Therefore, I suggest we maintain the current implementation.

chenzl25: I think changing the proto is acceptable, because it is a batch proto, so no breaking change.

proto/batch_plan.proto
```diff
@@ -97,6 +97,7 @@ message FileScanNode {
   string s3_access_key = 5;
   string s3_secret_key = 6;
   repeated string file_location = 7;
+  bool is_minio = 8;
```
chenzl25, line 100: Shouldn't we add an optional `s3_endpoint` field?

src/frontend/src/expr/table_function.rs
```diff
@@ -150,13 +150,19 @@ impl TableFunction {
             {
                 let (bucket, _) = extract_bucket_and_file_name(&eval_args[5].clone())?;
 
-                let is_minio = eval_args[2].starts_with("http");
+                let (s3_region, s3_endpoint) = match eval_args[2].starts_with("http") {
+                    true => ("us-east-1".to_owned(), eval_args[2].clone()), /* for minio, hard code region as not used but needed. */
+                    false => (
+                        eval_args[2].clone(),
+                        format!("https://{}.s3.{}.amazonaws.com", bucket, eval_args[2],),
```
chenzl25, line 157: Should we make the endpoint optional?

chenzl25 (approved):
Rest LGTM

src/frontend/src/expr/table_function.rs
wcy-fdu: I suggest no, since it goes directly as a parameter to the new s3_operator.

PR #19955, opened 2024-12-27
feat: support schedule/reschedule resource group

shanicky:

~blocked by #19884~

This PR adds support for resource groups for database (although it is not enabled), and provides specific resource group support for the materialized views that will be used for serverless backfill.

At the same time, scaling also supports resource groups.

This PR also adds support for altering resource groups (only for materialized views). 

Since altering resource groups is also a form of rescheduling, I have unified the logic with the previous alter parallelism and changed most of the existing alter logic for parallelism to reschedule streaming jobs, making it more consistent.

This PR still needs some testing, including simulations with madsim. 
Currently, the manual invocation of the alter command is working fine.

```
dev=> create table t(v int);
CREATE_TABLE
dev=> create materialized view m1 as select count(*) from t;
CREATE_MATERIALIZED_VIEW
dev=> create materialized view m2 as select * from m1;
CREATE_MATERIALIZED_VIEW
dev=> alter materialized view m1 set resource_group to test;

  1: gRPC request to meta service failed: Internal error
  2: Cannot resize streaming_job 7 to empty worker set

# add new worker with `test` resource group

dev=> alter materialized view m1 set resource_group to test;
ALTER_MATERIALIZED_VIEW
dev=> alter materialized view m2 set resource_group to test;
ALTER_MATERIALIZED_VIEW
dev=> alter materialized view m1 reset resource_group;
ALTER_MATERIALIZED_VIEW
dev=> alter materialized view m2 reset resource_group;
ALTER_MATERIALIZED_VIEW
```

By the way, this PR also adds support for creating a materialized view with a resource group.

```
dev=> create table t(v int);
CREATE_TABLE
dev=> create materialized view m with ( resource_group = 'test' ) as select * from t;
CREATE_MATERIALIZED_VIEW
```

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -322,7 +322,7 @@ redundant_explicit_links = "allow"
 lto = "off"
 # use parallel frontend to speed up build
 # TODO: may consider applying to release/production profile as well
-#rustflags = ["-Z", "threads=8"]
+rustflags = ["-Z", "threads=8"]
 
 [profile.release]
 debug = "full"


yezizp2012 (approved):

qccash:
Hi @shanicky, I noticed this is currently only supported for materialized views, while similar syntax is not available for ALTER TABLE or ALTER SINK.
Is there any specific design reason or technical constraint behind limiting it to materialized views?
Would appreciate any context on the tradeoff and whether support for table/sink is planned.

shanicky:
> Hi @shanicky, I noticed this is currently only supported for materialized views, while similar syntax is not available for ALTER TABLE or ALTER SINK.
> 
> Is there any specific design reason or technical constraint behind limiting it to materialized views?
> 
> Would appreciate any context on the tradeoff and whether support for table/sink is planned.

Hello, this feature is currently still in the experimental stage. At present, directly modifying a job's resource group may inadvertently compromise the isolation of the database, as the job's default resource group inherits from its database.

May I ask what kind of issue you encountered that requires the use of this command?

qccash:
> > Hi @shanicky, I noticed this is currently only supported for materialized views, while similar syntax is not available for ALTER TABLE or ALTER SINK.
> > Is there any specific design reason or technical constraint behind limiting it to materialized views?
> > Would appreciate any context on the tradeoff and whether support for table/sink is planned.
> 
> Hello, this feature is currently still in the experimental stage. At present, directly modifying a job's resource group may inadvertently compromise the isolation of the database, as the job's default resource group inherits from its database.
> 
> May I ask what kind of issue you encountered that requires the use of this command?

Main reason for needing ALTER ... SET RESOURCE_GROUP:

In real workloads, resource usage can be skewed even within the same database. Different fragments may have very uneven compute cost, and some become hotspots.
Another important factor is that fragment parallelism is not uniform. Some fragments can scale out, but others are inherently constrained (for example, they must run with single parallelism).
Because of this, load cannot always be balanced just by tuning parallelism. We still need ALTER ... SET RESOURCE_GROUP as a finer-grained isolation mechanism to move hotspot jobs/fragments and prevent resource imbalance.

Issue #19857, opened 2024-12-18
refactor(streaming): unify and refine `rate_limit` related type


MrCroxx:
### Background

Currently, we are using two types to present "rate_limit".

In `.proto` files, all rate limit values are `optional uint32`:

- `None`: infinite
- `0`: pause
- other value `v`: rate limit at `v`

However, when we are parsing the rate limit with SQL, we are using `i32`:

- `-1` (and other negative value for compatibility): infinite
- `0`: pause
- other positive value `v`: rate limit at `v`

Although it looks wired, it works fine for now.

But things are getting worse when we want to introduce some new rate limit policy. It is too tricky to use more conventional values, especially with `u32` type.

### Design

Use a new `message Throttle` to unify and refine all rate limits.

e.g.

```proto
enum ThrottleKind {
	THROTTLE_KIND_ADAPTIVE = 1;
	THROTTLE_KIND_INFINiTE = 2
	THROTTLE_KIND_FIXED = 3;
}

message Throttle {
	ThrottleKind kind = 1;
	// Only valid when `kind` is `FIXED`.
	uint32 value = 2;
}
```

MrCroxx:
Any comments? cc @hzxa21 @StrikeW @kwannoel @tabVersion

tabVersion:
at least it should not be `THROTTLE_KIND_INFINiTE` (`THROTTLE_KIND_INFINITE`) 🤣

MrCroxx:
> at least it should not be `THROTTLE_KIND_INFINiTE` (`THROTTLE_KIND_INFINITE`) 🤣

Fixed. 🫠

tabVersion:
proposing

```
enum ThrottleKind {
	THROTTLE_KIND_ADAPTIVE = 1;
	THROTTLE_KIND_INFINITE = 2
	THROTTLE_KIND_FIXED = 3;
++  THROTTLE_KIND_PAUSE = 4; // more distinguishable than contained by FIXED
}

message Throttle {
	ThrottleKind kind = 1;
	// Only valid when `kind` is `FIXED`.
	uint32 value = 2;
}
```

kwannoel:
How about `THROTTLE_KIND_DISABLED` instead of infinite.

MrCroxx:
Proposings on naming accepted.

Can I assume this refactor works for you?

We need to keep both fields for a few versions. Because the streaming plan is stored as proto in meta

Issue #19623, opened 2024-11-29
build: split pb and common to speed up compilation


xxchan:
`pb` and `common` are the root of all other crates. Whenever they are modified, all crates need to re-compiled, which take a lot of time.

One example: We can try to make `expr` compile separately. It depends on very few stuff actually, not the whole `pb` and `common` crate.
- `risingwave_pb::expr`
- `risingwave_common::{array,types,row,...}` 

Besides, these parts are modified rarely. If we separate these part out into new crates (maybe `risingwave_pb_data` and `risingwave_data`):
- They can be compile in an earlier stage, and compiled in parallel with the remaining crates.
- More importantly, they are less likely to be re-compiled.

<img width="1119" alt="image" src="https://github.com/user-attachments/assets/950207d9-b381-49e1-8ad7-51e2a276d637">

PR #19967, opened 2024-12-30
refactor(streaming): unify rate limiter and related components design

MrCroxx:
Signed-off-by: MrCroxx <mrcroxx@outlook.com>I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

Changes:

- Introduce unified `RateLimit`, `RateLimiter`, and `MonitoredRateLimiter`.
- `RateLimiter` follows interior mutability design, and supports online rate limit policy updates.
- Compared to `governor::RateLimiter`, the new `RateLimiter` will not deny request that overflow a "burst quota". So, it will be unnecessary to assert the cardinality is smaller than the burst quota.

Remains:

- `*.proto` files are still using the old design, I'll update it in another PR to keep changes easy for reviews. 

Related:

#19857 

## Checklist

## Documentation

<details>

</details>


MrCroxx:
Sometimes rate limit basic slt failed because of timeout.

https://github.com/risingwavelabs/risingwave/blob/794afcb784405385e0c283c6f8ead90b559cd728/e2e_test/streaming/rate_limit/basic.slt#L40-L41

This line may backfill the tombstones?

src/common/src/lib.rs
```diff
@@ -82,6 +82,7 @@ pub use {
 pub mod lru;
 pub mod opts;
 pub mod range;
+pub mod rate_limit;
```
xxchan, line 85: Better not to put in `common`, but a new crate

MrCroxx: BTW. May I ask what crate `common` is supposed to do now?

The `RateLimiter` is used by source, sink, streaming executor, dml, etc. It is actually more widely used than other modules in the `common` crate.

xxchan: It's convenient to put things in `common`, but the problem is we want to avoid `common` being too large since it's the single bottleneck of compilation. And we do want to split some code from there. https://github.com/risingwavelabs/risingwave/issues/19623

IMO the most important aspect is dependency relationship between crates. e.g., `RateLimiter` is depended by only `stream` and `connector` crates. If something is depended by also `expr`, `meta`, then it's perfectly `common`. 🤪

src/common/rate_limit/src/lib.rs
```diff
@@ -0,0 +1,544 @@
...
+            DelayProj::Infinite => Poll::Pending,
+        }
+    }
+}
+
+/// Rate limit policy.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum RateLimit {
+    /// Rate limit disabled.
+    Disabled,
```
stdrc, line 92: What about `Unlimited`?

src/common/rate_limit/src/lib.rs
```diff
@@ -0,0 +1,544 @@
...
+
+    pub fn rate_limit(&self) -> RateLimit {
+        self.inner.load().rate_limit()
+    }
+
+    pub fn check(&self, quota: u64) -> Check {
+        self.inner.load().check(quota)
+    }
+
+    pub async fn wait(&self, quota: u64) {
```
stdrc, line 212: So for a fixed limit, what if `quota > limit`? Will this method finally return?

src/common/rate_limit/src/lib.rs
MrCroxx: There was a discussion for this. I'm okay with either one. https://github.com/risingwavelabs/risingwave/issues/19857#issuecomment-2550864990

src/common/rate_limit/src/lib.rs
MrCroxx: It equals to wait for at least  `quota as f64 / limit as f64`, rather than returning an `InsufficientCapacity` error.

Let me add an unit test for this case.

src/common/rate_limit/src/lib.rs
```diff
@@ -0,0 +1,600 @@
...
+        let tat = loop {
+            let tat = ltat + weight;
+
+            if tat > tnow {
+                self.total_waited_nanos
+                    .fetch_add((tat - tnow) as i64, Ordering::Relaxed);
+                return Err(Duration::from_nanos(tat - tnow));
+            }
+
+            let ltat_new = std::cmp::max(tat, tnow);
```
hzxa21, line 426: When we reach here, `tat <= tnow` must hold. In this case, should we set ltat to `tat` instead of `tnow`? Otherwise, we will consume more quota than the provided one.

MrCroxx: The `max` here is necessary. I forgot to remove it after adding the early return above.

However, the `ltat` is supposed to be set to `tnow` (the larger one) to prevent form quota accumulation with no limitation.

e.g.

```plain
rate limit: 1 row/s

t(0s) ~ t(100s): no requests
t(101s) ~ t(110s): 2 rows
``` 

If we set the `ltat` to `tat`, during 101s ~ 110s, the 2 rows requests will be allowed with no delay, which makes the avg rate during the 10 seconds larger than expected.

But if we set the `ltat` to `tnow` here,  the request has to be delayed after each 2 row request is allowed, to make the avg rate during the 10 seconds reasonable.

hzxa21: I see. I thought we intended to support quota accumulation. Good to know that is not the case. Thanks for the explanation.

hzxa21 (approved):

src/common/src/lib.rs
MrCroxx: Just found the following PR will need the `RateLimit` in meta. Let me move it back to common. 🥲

xxchan: 🤪 I think if it's already out of common, move it back won't provide benefits. We can just keep it

MrCroxx: Already commit it in my new branch. But the modification is trivial. It's okay for me to move it back into a dedicated crate later.

Issue #19960, opened 2024-12-30
Reduce hummock version overhead on L0 log store large retention


hzxa21:
Currently metadata of L0 log store is stored as [TableChangeLog](https://github.com/risingwavelabs/risingwave/blob/0b3f4fb8a54bcd804b67f5cca04b7f9fbcef6d26/proto/hummock.proto#L195) in hummock version, which will be propagated via the notification service.

Recently we have seen that then the metadata size of `TableChangeLog` is larger (e.g. due to 30-day subscription retention), the hummock version size can become several hundred MBs, which can cause inefficiency in hummock event handler (see the [clone](https://github.com/risingwavelabs/risingwave/blob/0b3f4fb8a54bcd804b67f5cca04b7f9fbcef6d26/src/storage/src/hummock/event_handler/hummock_event_handler.rs#L530) here) and potentially in other places.

Some ideas to improve it:
- Don't store `TableChangeLog` in hummock version and exclude it from the notification. `iter_log` retrieves the `TableChangeLog` from meta on demand.
- Keep `TableChangeLog` in hummock version and notification. But we can use a separate structure to store `TableChangeLog` in CN to avoid cloning the full `TableChangLog` on every new version delta.

Li0k:
https://github.com/risingwavelabs/risingwave/pull/20050

PR #19984, opened 2025-01-02
perf(storage): use VecDeque for TableChangeLog to avoid large clone on truncate

wenym1:

May alleviate the issue mentioned in https://github.com/risingwavelabs/risingwave/issues/19960.

Previously we use `Vec` to store all table change log. When applying version delta, we use `Vec::retain` to clean the stale change log. However, for subscription with large retention time, the `Vec` can be large, and the `Vec::retain` called in `truncate` will incur great cost. In this PR, we change to use `VecDeque` to store the change log, so that the cost of `truncate` will reduce from `O(N)` to `O(1)`.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/storage/hummock_sdk/src/change_log.rs
+++ b/src/storage/hummock_sdk/src/change_log.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::collections::HashMap;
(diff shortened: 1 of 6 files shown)


zwang28 (approved):
LGTM

Li0k (approved):
LGTM

Issue #15144, opened 2024-02-20
bug: inconsistency table info between catalog and hummock in meta node


zwang28:
### Describe the bug

We've encountered this bug several times:
1. Compactor nodes keep failing task because [inconsistency table info between catalog and hummock manager](https://github.com/risingwavelabs/risingwave/blob/be30f20e2c14b5d12efd5b33a4ac9e537602fc37/src/storage/src/hummock/compactor/compactor_runner.rs#L340).
2. After restarting meta node, the issue is gone. It works due to the [rectification process](https://github.com/risingwavelabs/risingwave/blob/be30f20e2c14b5d12efd5b33a4ac9e537602fc37/src/meta/src/hummock/manager/compaction_group_manager.rs#L152) during meta node startup.

In meta node, whenever a table is remove from catalog, it must [notify hummock manager](https://github.com/risingwavelabs/risingwave/blob/be30f20e2c14b5d12efd5b33a4ac9e537602fc37/src/meta/src/hummock/manager/compaction_group_manager.rs#L382) as well. Otherwise it could lead to the issue above. So I suspect there is a corner case that doesn't follow this rule, during dropping a stream job.

hzxa21:
I found several suspicious things:
1. In ddl_controller_v2.rs [drop_object](https://github.com/risingwavelabs/risingwave/blob/4197ad50c766c8d086d29948ef34e13582e72251/src/meta/src/rpc/ddl_controller_v2.rs#L260), catalog is dropped first ([here](https://github.com/risingwavelabs/risingwave/blob/4197ad50c766c8d086d29948ef34e13582e72251/src/meta/src/rpc/ddl_controller_v2.rs#L286)) before `unregister_table_ids_fail_fast` is called ([here](https://github.com/risingwavelabs/risingwave/blob/4197ad50c766c8d086d29948ef34e13582e72251/src/meta/src/rpc/ddl_controller_v2.rs#L423)).
2. In [drop_streaming_jobs_v2](https://github.com/risingwavelabs/risingwave/blob/4197ad50c766c8d086d29948ef34e13582e72251/src/meta/src/stream/stream_manager.rs#L557), `unregister_table_ids_fail_fast` will not be called if `removed_actors` is empty
[...]

PR #19989, opened 2025-01-02
fix(meta): resolve deadlock caused by Hummock write stop

zwang28:

Before this PR, tables are unregistered from Hummock either after the drop-stream-job barrier succeeds or during recovery. However there is a corner case that can cause a deadlock situation:

1. The cluster encounters backpressure originating from Hummock. So the earliest barrier becomes stuck. It is expected to be resolved via Hummock compaction.
2. User drops the table. Meta removes the table from catalog immediately on receiving the drop command. But Hummock manager won't remove the table until the barrier finishes, which is stuck already.
3. At the moment, compaction task related to this dropped table will always fail due to the inconsistency between catalog and Hummock. So the backpressure will never recover. It's a deadlock situation. Neither the barrier or the compaction can make any progress.

This PR fixes it by ensuring Hummock unregister tables immediately after catalog has done so, **only if any of the dropped tables are causing Hummock write stop**. Note that during the period between the immediate unregistration of tables and the later actor dropping due to the mutation barrier, 
1. The actors related to the dropped tables will read incorrect data from Hummock. This is fine since those tables and their dependent tables will be eventually dropped anyway.
2. Consequenly compute nodes may panic due to consistency check failure. This is expected and the cluster will recover successfully after this.

related https://github.com/risingwavelabs/risingwave/issues/15144

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/meta/src/controller/catalog/drop_op.rs
+++ b/src/meta/src/controller/catalog/drop_op.rs
@@ -238,7 +238,7 @@ impl CatalogController {
 
         txn.commit().await?;
 
-        // After catalog changes have been saved, synchronize them with Hummock.
+        // After catalog changes have been saved, may need to synchronize them with Hummock to resolve a corner case.
         tracing::debug!(?to_drop_state_table_ids, "May unregister from Hummock.");
         let tables_to_unregister_from_hummock = to_drop_state_table_ids
             .iter()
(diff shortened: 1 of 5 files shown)


zwang28:
This PR does not address the deadlock scenario where a write stall occurs between the issuance of a drop command and the application of a drop barrier.

PR #18621, opened 2024-09-20
feat: deprecate etcd meta store

yezizp2012:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://gist.github.com/TennyZhuang/f00be7f16996ea48effb049aa7be4d66#file-rw_cla).

## Checklist

- [ ] I have written necessary rustdoc comments
- [ ] I have added necessary unit tests and integration tests
- [ ] I have added test labels as necessary. See [details](https://github.com/risingwavelabs/risingwave/blob/main/docs/developer-guide.md#ci-labels-guide).
- [ ] I have added fuzzing tests or opened an issue to track them. (Optional, recommended for new SQL features #7934).
- [ ] My PR contains breaking changes. (If it deprecates some features, please create a tracking issue to remove them in the future).
- [ ] All checks passed in `./risedev check` (or alias, `./risedev c`)
- [ ] My PR changes performance-critical code. (Please run macro/micro-benchmarks and show the results.)

- [ ] My PR contains critical fixes that are necessary to be merged into the latest release. (Please check out the [details](https://github.com/risingwavelabs/risingwave/blob/main/CONTRIBUTING.md))

## Documentation

- [ ] My PR needs documentation updates. (Please use the **Release note** section below to summarize the impact on users)

## Release note

If this PR includes changes that directly affect users or other significant modifications relevant to the community, kindly draft a release note to provide a concise summary of these changes. Please prioritize highlighting the impact these changes will have on users.

</details>

PR #20000, opened 2025-01-02
feat: batching telemetry event request avoid too many requests

tabVersion:

## Checklist

## Documentation

<details>

</details>


tabVersion:
https://github.com/risingwavelabs/telemetry-backend/pull/45 related changes on the server side

proto/telemetry.proto
```diff
@@ -7,8 +7,7 @@ option go_package = "risingwavelabs.com/risingwave/proto/telemetry";
 enum MetaBackend {
   META_BACKEND_UNSPECIFIED = 0;
   META_BACKEND_MEMORY = 1;
-  reserved 2;
-  reserved "META_BACKEND_ETCD";
```
tabVersion, line 11: telemetry service still receive requests from legacy clusters and it is still possible getting reports from a ETCD backend cluster. 
I believe the field is changed by mistake in https://github.com/risingwavelabs/risingwave/pull/18621 
There is no breaking change in the proto file because there is no updates on the server side since then.

src/common/telemetry_event/src/lib.rs
```diff
@@ -95,15 +117,8 @@ pub fn request_to_telemetry_event(
...
-
-    tokio::spawn(async move {
-        const TELEMETRY_EVENT_REPORT_TYPE: &str = "event";
-        let url = (TELEMETRY_REPORT_URL.to_owned() + "/" + TELEMETRY_EVENT_REPORT_TYPE).to_owned();
-        post_telemetry_report_pb(&url, report_bytes)
-            .await
-            .unwrap_or_else(|e| tracing::info!("{}", e))
-    });
+
+    TELEMETRY_EVENT_REPORT_STASH.blocking_lock().push(event);
```
hzxa21, line 121: Given that we call `blocking_lock` here, why not just using a non-async mutex (`parking_lot::Mutex`) instead of the tokio async mutex?

src/common/telemetry_event/src/lib.rs
```diff
@@ -42,6 +43,27 @@ pub fn get_telemetry_risingwave_cloud_uuid() -> Option<String> {
...
+        batch_message.events.push(event);
+    }
+    drop(stash_guard);
+
+    post_telemetry_report_pb(&url, batch_message.encode_to_vec())
+        .await
+        .unwrap_or_else(|e| tracing::debug!("{}", e));
+}
+
+pub const TELEMETRY_EVENT_REPORT_INTERVAL: u64 = 10; // 10 seconds
```
hzxa21, line 65: Is the size of the events accumulated in 10s controllable? If not, I am concerned that doing only time-based batching can cause memory pressure / OOM on the node. We might need to consider adding size-based batching as well.

src/common/telemetry_event/src/lib.rs
```diff
@@ -42,6 +43,27 @@ pub fn get_telemetry_risingwave_cloud_uuid() -> Option<String> {
     env::var(TELEMETRY_RISINGWAVE_CLOUD_UUID).ok()
 }
 
+static TELEMETRY_EVENT_REPORT_STASH: LazyLock<Mutex<Vec<PbEventMessage>>> =
```
hzxa21, line 46: You can consider using tokio unbounded channel which uses a lock free list under the hook to implement batching better.
```rust
let (tx, rx) = unbounded_channel();

// In `fn request_to_telemetry_event`
tx.send(event)

// In the task loop of `start_telemetry_reporting`
const BATCH_SIZE: usize = 100;
let mut batch_message = PbBatchEventMessage { events: Vec::with_capacity(BATCH_SIZE) };
loop {
    tokio::select! {
        ...
        event = rx.recv() => {
          batch_message.events.push(event);
          if batch_message.events.len() >= BATCH_SIZE {
            // size-based batching
            post_telemetry_report_pb(...)
          }
          continue;
        }
        _ = event_interval.tick() => {
             // time-based batching
            post_telemetry_report_pb(...)
            continue;
        },
        ...
...
```

src/common/telemetry_event/src/lib.rs
tabVersion: It makes sense. Let me fix.

src/common/telemetry_event/src/lib.rs
tabVersion: after switching to unbounded channel, there seems no need for the sync Mutex.

src/common/src/telemetry/report.rs
```diff
@@ -91,9 +101,29 @@ where
...
+        });
+
         loop {
             tokio::select! {
                 _ = interval.tick() => {},
+                event = event_rx.recv() => {
+                    if let Some(event) = event {
+                        TELEMETRY_EVENT_REPORT_STASH.lock().await.push(event);
+                    }
+                    if TELEMETRY_EVENT_REPORT_STASH.lock().await.len() >= TELEMETRY_EVENT_REPORT_STASH_SIZE {
```
hzxa21, line 118: `TELEMETRY_EVENT_REPORT_STASH` with locking is unnecessary because `do_telemetry_event_report` is called in this tokio task only and there is no real contention on it. We can define a local var for the message batch and pass it to `do_telemetry_event_report` directly.
```rust
let mut batch = Vec::new();

loop {
  ....
  tokio::select! {
    ...
    ... => {
     ...
     do_telemetry_event_report(std::mem::take(&mut batch)).await;
    }
    ...
  }
}

Comment (tabVersion) [reply src/common/src/telemetry/report.rs:118]: My intention is to move all event report stuff inside the `telemetry_event` crate, and defining a local var here may make the two telemetry object mixed. 
But yes, a local var does better here.

File: src/common/telemetry_event/src/lib.rs [hunk 6]
```diff
@@ -42,6 +44,23 @@ pub fn get_telemetry_risingwave_cloud_uuid() -> Option<String> {
...
 }
 
+pub async fn do_telemetry_event_report(event_stash: &mut Vec<PbEventMessage>) {
+    const TELEMETRY_EVENT_REPORT_TYPE: &str = "events"; // the batch report url
+    let url = (TELEMETRY_REPORT_URL.to_owned() + "/" + TELEMETRY_EVENT_REPORT_TYPE).to_owned();
+    let mut batch_message = PbBatchEventMessage { events: Vec::new() };
+
+    for event in event_stash.drain(..) {
+        batch_message.events.push(event);
+    }
```
hzxa21, line 54: We can use std::mem::take to avoid vec iterator and per item memory allocation/copy here.

src/common/telemetry_event/src/lib.rs
```diff
@@ -95,15 +114,10 @@ pub fn request_to_telemetry_event(
...
-    tokio::spawn(async move {
-        const TELEMETRY_EVENT_REPORT_TYPE: &str = "event";
-        let url = (TELEMETRY_REPORT_URL.to_owned() + "/" + TELEMETRY_EVENT_REPORT_TYPE).to_owned();
-        post_telemetry_report_pb(&url, report_bytes)
-            .await
-            .unwrap_or_else(|e| tracing::info!("{}", e))
-    });
+
+    if let Some(tx) = TELEMETRY_EVENT_REPORT_TX.get() {
+        tx.send(event).unwrap();
```
hzxa21, line 119: To be safer, let's use `inspect` to print a warn log instead of `unwrap` here.

src/common/src/telemetry/report.rs
```diff
@@ -91,9 +100,30 @@ where
...
+                "Telemetry failed to set event reporting tx, event reporting will be disabled"
+            );
+        });
+        let mut event_stash = Vec::new();
+
         loop {
             tokio::select! {
                 _ = interval.tick() => {},
+                event = event_rx.recv() => {
+                    if let Some(event) = event {
```
hzxa21, line 115: Given that `tx` is held in a static var, which will never be dropped, `recv()` should never return `None`. To be safer, can we assert that in debug build (cfg!(debug_assertions) == true) and print a warn log in release build?

hzxa21 (approved):
Rest LGTM

tabVersion:
manually tested, merging now

src/common/src/telemetry/report.rs
```diff
@@ -91,9 +100,29 @@ where
...
+            );
+        });
+        let mut event_stash = Vec::new();
+
         loop {
             tokio::select! {
                 _ = interval.tick() => {},
+                event = event_rx.recv() => {
+                    debug_assert!(event.is_some());
+                    event_stash.push(event.unwrap());
```
lmatz, line 116:
```thread 'rw-standalone-frontend' panicked at /risingwave/src/common/src/telemetry/report.rs:116:44:
```

when using standalone, each of the components panics

tabVersion: I guess no, will validate tmr.

PR #20006, opened 2025-01-02
feat: Add offline scaling support for background ddl with arrangement backfill

shanicky:

~*WIP STATE*~

It seems that No Shuffle backfill is still difficult to support in offline scaling, and this PR is still in a WIP state.

I found it somewhat challenging to directly support background jobs in online scaling, so I proposed this PR to add support for background jobs in offline scaling. At the same time, I attempted to skip jobs related to background DDL jobs in online scaling instead of completely omitting them.

This PR makes several specific changes.

1. Due to changes in the stream job fragment after offline scaling, it needs to be retrieved again.
2. Since the scaling occurs before the created event, the FE will receive the scaling mapping first.

## Checklist

### DIFF
--- a/src/meta/src/controller/fragment.rs
+++ b/src/meta/src/controller/fragment.rs
@@ -20,7 +20,7 @@ use anyhow::Context;
 use itertools::Itertools;
 use risingwave_common::bail;
 use risingwave_common::hash::{VnodeCount, VnodeCountCompat, WorkerSlotId};
-use risingwave_common::util::stream_graph_visitor::visit_stream_node_mut;
+use risingwave_common::util::stream_graph_visitor::{visit_stream_node, visit_stream_node_mut};
 use risingwave_meta_model::actor::ActorStatus;
 use risingwave_meta_model::fragment::DistributionType;
 use risingwave_meta_model::object::ObjectType;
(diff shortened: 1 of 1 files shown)


src/meta/src/controller/fragment.rs
```diff
@@ -723,6 +724,40 @@ impl CatalogController {
...
+
+        let mut result = HashMap::new();
+
+        for fragment::Model {
+            fragment_id,
+            stream_node,
+            ..
+        } in fragments
+        {
+            let mut stream_node = stream_node.to_protobuf();
```
yezizp2012, line 745: I think we can use `fragment_type_mask` instead to reduce memory usage.

kwannoel: We can shortcut for snapshot backfill

yezizp2012 (approved):
lgtm

src/meta/src/stream/scale.rs
```diff
@@ -2452,45 +2502,44 @@ impl GlobalStreamManager {
...
 
         let job_ids: HashSet<_> = {
             let streaming_parallelisms = self
                 .metadata_manager
                 .catalog_controller
-                .get_all_created_streaming_parallelisms()
+                .get_all_streaming_parallelisms()
                 .await?;
 
-            // streaming_parallelisms
```
yezizp2012, line 2480: Can we remove it?

PR #20008, opened 2025-01-03
fix(stream): avoid committing to in-memory state store in join benchmark

kwannoel:

If we commit to state store, we will end up measuring the memory consumption of writing the amplified chunks to the in-memory state store.
In production, state store is on top of object storage / disk. The in-memory state store is only used in benchmark / tests. So we shouldn't benchmark the memory consumption.

Further, the amplified chunks are not written in a large batch typically. The scenario we care about is OOM from dimension table update. In such a scenario, the fact table side's memory consumption will be high due to reads from the update side, rather than writes.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/executor/test_utils.rs
+++ b/src/stream/src/executor/test_utils.rs
@@ -852,6 +852,12 @@ pub mod hash_join_executor {
             // Push a single record into tx_r, so 100K records to be matched are cached.
             let chunk = build_chunk(amp, 200_000);
             tx_r.push_chunk(chunk);
+
+            // Ensure that the chunk on the rhs is processed, before inserting a chunk
+            // into the lhs. This is to ensure that the rhs chunk is cached,
+            // and we don't get interleaving of chunks between lhs and rhs.
+            tx_l.push_barrier(test_epoch(2), false);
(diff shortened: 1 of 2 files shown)


kwannoel:
* **#20008** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20008?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20008?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/stream/src/executor/test_utils.rs
```diff
@@ -854,9 +854,6 @@ pub mod hash_join_executor {
             tx_r.push_chunk(chunk);
         }
 
-        tx_l.push_barrier(test_epoch(2), false);
-        tx_r.push_barrier(test_epoch(2), false);
```
kwannoel, line 858: This second barrier will trigger a state store flush. Because the barrier before this is the initial barrier, it will not commit data to state store.

chenzl25 (approved):

PR #20010, opened 2025-01-03
fix(meta): fix notification when dropping tables

zwang28:

When dropping a table from catalog, the corresponding state tables will be later unregistered from Hummock either after the drop-stream-job barrier succeeds or during recovery. However there is a corner case that can cause compaction being stuck:

1. The cluster encounters backpressure originating from Hummock. So the earliest barrier becomes stuck. It is expected to be resolved via Hummock compaction.
2. User drops the table. Meta removes the table from catalog immediately on receiving the drop command. But Hummock won't remove the table until the barrier finishes, which is blocked due to backpressure in step 1.
3. Afterwards, compaction task related to this dropped table will always fail due to the inconsistency between catalog and Hummock.
    - If no additional operation is taken, the cluster is in deadlock because the barrier is waiting for backpressure to be resolved, while the backpressure cannot be resolved as explained in step 3.
    - If RECOVER is executed, the stuck barrier is cancelled and the cluster is partially recovered. However the compaction involving the dropped tables will still fail due to the same reason explained in step 3.
    - If meta node is restarted, the cluster is completely recovered, because there's a purge process to sync tables between catalog and Hummock during meta node startup.

This PR addresses the issue by ensuring that HummockObserver and CompactorObserver don't recognize those tables as dropped until the barrier corresponding to the drop command has been completed.

This PR also fixes the issue that notifications of catalog deletion are never sent to HummockObserver and CompactorObserver, which is likely a bug introduced in SQL backend.
- For catalog additions and updates, those observers will maintain their current behavior, which is to fetch data on demand. The rationale behind this decision is to simplify meta node notifications, as there are many scenarios that can trigger catalog additions and updates. Not all of these notifications are necessary for the HummockObserver and CompactorObserver, as they are primarily concerned with one catalog type, Table.

related https://github.com/risingwavelabs/risingwave/issues/15144

## Checklist

## Documentation

<details>

</details>


src/meta/src/controller/catalog/drop_op.rs
```diff
@@ -239,11 +240,21 @@ impl CatalogController {
 
         // notify about them.
         self.notify_users_update(user_infos).await;
+        let dropped_tables = inner
+            .list_all_state_tables(Some(to_drop_state_table_ids.iter().copied().collect()))
```
hzxa21, line 244: I may be missing something. `list_all_state_tables` selects `Table` relation. Since this happens after `Object::delete_many`, I think the corresponding rows will be deleted already and we cannot get any results here?

zwang28: You're right. Need to obtain them earlier.

src/meta/src/controller/catalog/mod.rs
```diff
@@ -851,4 +862,21 @@ impl CatalogControllerInner {
             .await?;
         Ok(table_ids)
     }
+
+    /// Since the tables have been dropped from both meta store and streaming jobs, this method removes those table copies.
+    /// Returns the removed table copies.
+    pub(crate) fn complete_dropped_tables(
```
Li0k, line 868: how about replacing the mut Vec with filter_map, it looks like 

```rust
pub(crate) fn complete_dropped_tables(
    &mut self,
    table_ids: impl Iterator<Item = TableId>,
) -> Vec<PbTable> {
    table_ids
        .filter_map(|table_id| {
            self.dropped_tables.remove(&table_id).map_or_else(
                || {
                    tracing::warn!("table {table_id} not found");
                    None
                },
                Some,
            )
        })
        .collect()
}
```

Li0k (approved):
Rest LGTM, thanks for the efforts

PR #20012, opened 2025-01-03
fix(optimizer): can't apply pull_up_correlated_predicate_agg_rule with non-null-propagating expr

### LINKED ISSUE #19835: PullUpCorrelatedPredicateAggRule is wrong for non-null-propagating expression
### Describe the bug

I couldn't create a sample as I don't know how to how to create an empty array of rows. But a query like below return a `null` map when the subquery return no rows. I expected it to return an empty map instead.

Also, couldn't we simplify the syntax by removing the explicit `array()` ?

### Error message/log

_No response_

### To Reproduce

```sql
select map_from_entries(array(select row(foo.id, foo.name) from foo where foo.id=unrelated.id))
```

### Expected behavior

Return an empty map instead of null

_No response_

### The version of RisingWave

version | PostgreSQL 13.14.0-RisingWave-2.0.4 (2d757981a1f747c057f79bfb4a19fffbde563ece)

### Additional context

_No response_

xxchan:

fix #19835

## Checklist

## Documentation

<details>

</details>


src/frontend/src/optimizer/rule/pull_up_correlated_predicate_agg_rule.rs
```diff
@@ -153,55 +153,32 @@ impl Rule for PullUpCorrelatedPredicateAggRule {
...
-
-            // Only all expr in conjunctions are null, we can apply this rule, otherwise, bail out.
-            if top_filter
-                .predicate()
-                .conjunctions
-                .iter()
-                .any(|expr| !Strong::is_null(expr, top_proj_null_bitset.clone()))
-            {
-                return None;
-            }
```
chenzl25, line 204: Why this logic needed to be changed?

xxchan: I don't get why previously we check top_filter. I think we should check the output exprs instead

chenzl25: Checking top_filter to ensure the null agg row generated by left outer join finally could be eliminated by the filter. If it could be eliminated, then we can allow the transformation, even if we meet `count`, `array`, and other aggs will generate a not null output when input is empty.

src/frontend/src/optimizer/rule/pull_up_correlated_predicate_agg_rule.rs
```diff
@@ -153,55 +153,32 @@ impl Rule for PullUpCorrelatedPredicateAggRule {
...
-            // When group input is empty, not count agg would return null.
+        // If the `group by + proj` doesn't return null for empty input, we can't apply this rule.
+        {
+            // When group input is empty, if the agg is not `count`, it would return null.
             let null_agg_pos = agg_calls
                 .iter()
                 .positions(|agg_call| {
                     !matches!(agg_call.agg_type, AggType::Builtin(PbAggKind::Count))
                 })
                 .collect_vec();
```
chenzl25, line 165: Shall we add `array` aggregation here?

xxchan: array subquery is planned as `array_agg + coalesce`

xxchan: and `array_agg` is null-propagating

chenzl25: OK, I see.

src/frontend/src/optimizer/rule/pull_up_correlated_predicate_agg_rule.rs
xxchan: Oh, I think the condition can be:
`top_filter is null OR all of top_proj are null`

chenzl25: Yes, you can construct a planner test case I think.

src/frontend/src/optimizer/rule/pull_up_correlated_predicate_agg_rule.rs
```diff
@@ -185,21 +178,26 @@ impl Rule for PullUpCorrelatedPredicateAggRule {
...
 
-            // Only all expr in conjunctions are null, we can apply this rule, otherwise, bail out.
-            if top_filter
+            let top_filter_all_null = top_filter
                 .predicate()
                 .conjunctions
                 .iter()
-                .any(|expr| !Strong::is_null(expr, top_proj_null_bitset.clone()))
-            {
+                .all(|expr| Strong::is_null(expr, top_proj_null_bitset.clone()));
```
chenzl25, line 196: It should be any.
```suggestion
                .any(|expr| Strong::is_null(expr, top_proj_null_bitset.clone()));
```

src/frontend/src/optimizer/rule/pull_up_correlated_predicate_agg_rule.rs
```diff
@@ -185,21 +178,26 @@ impl Rule for PullUpCorrelatedPredicateAggRule {
...
             // Shift the top project expressions to the right by apply_left schema len, because it is used to check null-rejected by the top filter.
             let apply_left_schema = apply_left.schema().len();
+            let mut top_proj_all_null = true;
             let mut top_proj_null_bitset =
                 FixedBitSet::with_capacity(top_project.base.schema().len() + apply_left_schema);
             for (i, expr) in top_proj_exprs.iter().enumerate() {
                 if Strong::is_null(expr, agg_null_bitset.clone()) {
                     top_proj_null_bitset.insert(i + apply_left_schema);
+                } else {
+                    top_proj_all_null = false;
```
chenzl25, line 188: When `Strong::is_null` return false, it means we can't get any useful information, so we shouldn't allow apply rule based on it.

chenzl25 (approved):
Rest LGTM! Thanks

maingoh:
@xxchan I don't see this commit in the latest tags `2.0.6` or `2.1.2` or `v2.2.0-rc.1`. I am wondering if will be tagged or if it has been forgotten ? Thank you !

xxchan:
I think it can be cherry-picked into 2.2 which hasn't been released yet

PR #20045, opened 2025-01-07
fix(iceberg): only convert iceberg table to iceberg source for batch dql

chenzl25:

- We should only convert the iceberg table to iceberg source for batch dql, since other batch sql (especially dml update and delete) should read the hummock table and use indexes to improve the performance and freshness.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/handler/query.rs
+++ b/src/frontend/src/handler/query.rs
@@ -247,7 +247,11 @@ fn gen_batch_query_plan(
         ..
     } = bind_result;
 
-    let mut planner = Planner::new_for_batch(context);
+    let mut planner = if matches!(bound, BoundStatement::Query(_)) {
+        Planner::new_for_batch_dql(context)
+    } else {
+        Planner::new_for_batch(context)
+    };
 
     let mut logical = planner.plan(bound)?;
     let schema = logical.schema();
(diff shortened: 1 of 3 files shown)


kwannoel (commented):
What's the difference between using iceberg table for DQL and iceberg source for DQL?
When running DQL against iceberg table does it use the row-based replica rather than the actual iceberg table, and hence we convert it to iceberg source?

I suppose we may also want to let user configure this down the road, since their SQL could have a combination of DQL and DML.

chenzl25:
> What's the difference between using iceberg table for DQL and iceberg source for DQL? When running DQL against iceberg table does it use the row-based replica rather than the actual iceberg table, and hence we convert it to iceberg source?
> 
> I suppose we may also want to let user configure this down the road, since their SQL could have a combination of DQL and DML.

Let me change the ambiguous word iceberg table and iceberg source to row store and column store. The column store and the row store, which is the best one we want to use. As we don't have a cost-based optimizer to help us compare the cost among different plans, we should choose a default option for users. For DML, it seems choosing the row store is more beneficial since we can utilize indexes to accelerate point update/delete and more importantly there is a freshness lag between row store and column store right now, if we read column store to performance dml, which means we can't manipulate a row after it has been committed to iceberg, which might be a bad use experience for a demo. Last, it's fair to provide a session variable to allow users to control whether read row store or column store.

kwannoel (approved):

PR #19974, opened 2024-12-31
feat(connector): introduce gcs file scan

wcy-fdu:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## What's changed and what's your intention?
Support table function file_scan from gcs bucket, the syntax is `file_scan(parquet, gcs, credential, service_account, file_loaction)`. The same with s3 file_scan, it also supports reading a single parquet file or all parquet files in a directory.

Implementation detail:
- add GcsFileScan Node
- The public functions are all modified to pass in the opendal operator. Depending on the type of the passed in parameter, it is determined creating s3 operator or gcs operator.
- test on gcs and passed. For general file scan tests, see https://github.com/risingwavelabs/risingwave/pull/19950

## Checklist

- [ ] I have written necessary rustdoc comments.
- [ ]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  My PR contains critical fixes that are necessary to be merged into the latest release. 

## Documentation

- [ ]  My PR needs documentation updates. 

Pls add this feature in documentation, `file_scan(parquet, gcs, credential, service_account, file_loaction)`.

<details>
<summary><b>Release note</b></summary>

</details>

PR #20046, opened 2025-01-07
feat(connector): introduce azblob file scan

wcy-fdu:

Similar and follow up https://github.com/risingwavelabs/risingwave/pull/19974
Support table function file_scan from azblob container, the syntax is `file_scan(parquet, azblob, account_name, account_key, endpoint, file_loaction)`. The same with s3 file_scan, it also supports reading a single parquet file or all parquet files in a directory.

Implementation detail:
- add AzblobFileScan Node
- The public functions are all modified to pass in the opendal operator. Depending on the type of the passed in parameter, it is determined creating s3 operator or azblob operator.
- test on azblob container and work well.

## Checklist

## Documentation

<details>

</details>


src/frontend/src/optimizer/rule/table_function_to_file_scan_rule.rs
```diff
@@ -96,6 +97,25 @@ impl Rule for TableFunctionToFileScanRule {
                     )
                     .into(),
                 )
+            } else if "azblob".eq_ignore_ascii_case(&eval_args[1]) {
+                let account_name = eval_args[2].clone();
+                let account_key = eval_args[3].clone();
+                let endpoint = eval_args[4].clone();
+                // The rest of the arguments are file locations
+                let file_location = eval_args[5..].iter().cloned().collect_vec();
```
hzxa21, line 105: Is it possible to have "index out of bound" here? In other words, how do prevent panicking for bad user inputs (i.e. missing some args)

wcy-fdu: No, previously, there was logic [here](https://github.com/risingwavelabs/risingwave/blob/bb4dec288c5d93cc12c549e56ca7d11b0e827dc3/src/frontend/src/expr/table_function.rs#L86) to first check that the length of eval_args must be 6, otherwise an error would be reported. 
Due to the introduction of gcs file_scan, the number of parameters passed in when reading gcs is 4, so a stricter check is done, see `table_function.rs` in pr change. If the parameters do not match, an error will be reported directly on the front end.
```
 if (eval_args.len() != 4 && eval_args.len() != 6)
                || (eval_args.len() == 4 && !"gcs".eq_ignore_ascii_case(&eval_args[1]))
                || (eval_args.len() == 6
                    && !"s3".eq_ignore_ascii_case(&eval_args[1])
                    && !"azblob".eq_ignore_ascii_case(&eval_args[1]))
            {
                return Err(BindError(
                "file_scan function supports three backends: s3, gcs, and azblob. Their formats are as follows:: \n
                    file_scan('parquet', 's3', s3_region, s3_access_key, s3_secret_key, file_location) \n
                    file_scan('parquet', 'gcs', credential, service_account, file_location) \n
                    file_scan('parquet', 'azblob', account_name, account_key, endpoint, file_location)"
                        .to_owned(),
                )
                .into());
            }
```

wcy-fdu:
@hzxa21 @chenzl25 PTAL❤️

src/frontend/src/optimizer/rule/table_function_to_file_scan_rule.rs
```diff
@@ -106,6 +107,25 @@ impl Rule for TableFunctionToFileScanRule {
                     )
                     .into(),
                 )
+            } else if "azblob".eq_ignore_ascii_case(&eval_args[1]) {
+                let account_name = eval_args[2].clone();
+                let account_key = eval_args[3].clone();
+                let endpoint = eval_args[4].clone();
```
chenzl25, line 113: To make it more aligned with `s3`, I think we should put endpoint before account_name and account_key, i.e. file_scan('parquet', 'azblob', endpoint, account_name, account_key, file_location).

src/frontend/src/expr/table_function.rs
```diff
@@ -80,12 +80,10 @@ impl TableFunction {
...
-            // storage type e.g. s3, gcs
-            // For s3: file_scan(file_format, s3, s3_region, s3_access_key, s3_secret_key, file_location_or_directory)
-            // For gcs: file_scan(file_format, gcs, credential, file_location_or_directory)
-            if args.len() != 6 && args.len() != 4 {
-                return Err(BindError("file_scan function only accepts: file_scan('parquet', 's3', s3 region, s3 access key, s3 secret key, file location) or file_scan('parquet', 'gcs', credential, service_account, file location)".to_owned()).into());
-            }
+            // storage type e.g. s3, gcs, azblob
+            // For s3: file_scan('parquet', 's3', s3_region, s3_access_key, s3_secret_key, file_location_or_directory)
+            // For gcs: file_scan('parquet', 'gcs', credential, file_location_or_directory)
+            // For azblob: file_scan('parquet', 'azblob', account_name, account_key, endpoint, file_location)
```
chenzl25, line 86:
```suggestion
            // For azblob: file_scan('parquet', 'azblob', endpoint, account_name, account_key, file_location)
```

chenzl25 (approved):
Rest LGTM

PR #20048, opened 2025-01-07
feat(sink): support es sink partial update 

xxhZs:

We will send all the columns every time we update, which will cause the load of es to become high, after this pr, es sink will judge the columns of update, and only send the modified columns

## Checklist

## Documentation

<details>

</details>


src/connector/src/sink/elasticsearch_opensearch/elasticsearch_opensearch_formatter.rs
```diff
@@ -114,6 +117,7 @@ impl ElasticSearchOpenSearchFormatter {
     }
 
     pub fn convert_chunk(&self, chunk: StreamChunk) -> Result<Vec<BuildBulkPara>> {
+        let mut update_map: HashMap<String, RowRef<'_>> = HashMap::new();
```
hzxa21, line 120: I think this map is unnecessary because `UpdateDelete` and `UpdateInsert` must be adjacent to each other so we only need to cache one row for the check.

src/connector/src/sink/elasticsearch_opensearch/elasticsearch_opensearch_formatter.rs
```diff
@@ -147,7 +151,27 @@ impl ElasticSearchOpenSearchFormatter {
...
+                        delete_row
+                            .iter()
+                            .enumerate()
+                            .zip_eq_debug(rows.iter())
+                            .for_each(|((index, insert_column), delete_column)| {
+                                if let Some(insert_column) = insert_column
+                                    && let Some(delete_column) = delete_column
+                                    && insert_column == delete_column
+                                {
+                                    col_indices.retain(|&x| x != index);
```
hzxa21, line 168: It is more efficient to populate the new `col_indices` proactively instead of calling `retain` on the full list for each column. Example:
```
if let Some(insert_column) = insert_column
        && let Some(delete_column) = delete_column
        && insert_column == delete_column {
  // do nothing
} else {
  modified_col_indices.push_back(index);
}
```

src/connector/src/sink/elasticsearch_opensearch/elasticsearch_opensearch_formatter.rs
```diff
@@ -152,7 +154,30 @@ impl ElasticSearchOpenSearchFormatter {
...
-                    let value = self.value_encoder.encode(rows)?;
+                    let mut modified_col_indices = Vec::new();
+                    if let Some((delete_key, delete_row)) = update_delete_row.take()
+                        && delete_key == key
+                    {
+                        delete_row
+                            .iter()
+                            .enumerate()
+                            .zip_eq_debug(rows.iter())
+                            .for_each(|((index, delete_column), insert_column)| {
```
wenym1, line 165: Should we just compare the Option<delete_column> and Option<insert_column> directly? For a column that was None in delete_column and also None in insert_column, we also don't have to update the column.

src/connector/src/sink/elasticsearch_opensearch/elasticsearch_opensearch_formatter.rs
```diff
@@ -152,7 +154,30 @@ impl ElasticSearchOpenSearchFormatter {
             match op {
                 Op::Insert | Op::UpdateInsert => {
                     let key = self.key_encoder.encode(rows)?.into_string()?;
-                    let value = self.value_encoder.encode(rows)?;
+                    let mut modified_col_indices = Vec::new();
+                    if let Some((delete_key, delete_row)) = update_delete_row.take()
+                        && delete_key == key
```
wenym1, line 159: We may separate the handling of `Op::Insert` and `Op::UpdateInsert`, because we should only check the `update_delete_row` on `Op::UpdateInsert`, and it's expected to have  `Some(update_delete_row)`, and we should add warning log on `update_delete_row.is_none()` or `delete_key != key`, and on `Op::Insert` we can reuse the original logic and don't check the `update_delete_row` and don't do partial update.

integration_tests/elasticsearch-sink/create_sink.sql
```diff
@@ -14,6 +15,7 @@ CREATE SINK bhv_es8_sink
 FROM
     bhv_mv WITH (
     connector = 'elasticsearch',
+    type = 'upsert',
```
wenym1, line 18: Is the case of partial update already covered in the current tests? If not we should add some tests to cover it.

wenym1 (commented):
Rest LGTM

src/connector/src/sink/elasticsearch_opensearch/elasticsearch_opensearch_formatter.rs
```diff
@@ -152,7 +154,30 @@ impl ElasticSearchOpenSearchFormatter {
             match op {
                 Op::Insert | Op::UpdateInsert => {
                     let key = self.key_encoder.encode(rows)?.into_string()?;
-                    let value = self.value_encoder.encode(rows)?;
+                    let mut modified_col_indices = Vec::new();
```
wenym1, line 157: Use `Vec::with_capacity(rows.len())`

wenym1 (approved):
LGTM

PR #20050, opened 2025-01-07
feat(storage): Splitting table change log from HummockVersion on CN side

Li0k:

This PR optimize clone behavior on the CN side. Previously, the hummock event handler copied the hummock version every time it applied a delta, and the overhead of cloning could cause performance problems.issues.https://github.com/risingwavelabs/risingwave/blob/95abcc9493709e4547b75b4e01d19eed7e28d3ce/src/storage/src/hummock/event_handler/hummock_event_handler.rs#L529

This PR split the table change log from the hummock version to avoid copying all table change logs at each version delta.

Key changes include:

### Enhancements to `HummockVersion` and `HummockVersionDelta`:

* Introduced a second type parameter `L` to `HummockVersionCommon` and `HummockVersionDeltaCommon` to improve type safety and flexibility. Split table_change_log into separate fields and protect them with RwLock. (`src/storage/hummock_sdk/src/version.rs`, `src/storage/hummock_sdk/src/compaction_group/hummock_version_ext.rs`)

### New Methods and Type Changes:

* Added `change_log_into_iter` method to `TableChangeLogCommon` to allow iteration over change logs. (`src/storage/hummock_sdk/src/change_log.rs`)

### Type Aliases:

* Updated type aliases to use the new two-parameter versions of `HummockVersionCommon` and `HummockVersionDeltaCommon`. (`src/storage/hummock_sdk/src/time_travel.rs`, `src/storage/hummock_sdk/src/version.rs`)

### Import Adjustments:

* Adjusted imports to accommodate the new types and methods. (`src/storage/hummock_sdk/src/compaction_group/hummock_version_ext.rs`, `src/storage/hummock_sdk/src/version.rs`, `src/storage/src/hummock/event_handler/hummock_event_handler.rs`)

## Checklist

## Documentation

<details>

</details>


src/storage/src/hummock/local_version/pinned_version.rs
```diff
@@ -112,13 +118,41 @@ impl PinnedVersion {
...
+        let t = std::mem::take(&mut version.table_change_log);
 
         Some(PinnedVersion {
-            version: Arc::new(version),
             guard: Arc::new(PinnedVersionGuard::new(
                 version_id,
                 self.guard.pinned_version_manager_tx.clone(),
             )),
+            table_change_log: Arc::new(RwLock::new(t)),
+            version: Arc::new(LocalHummockVersion::from(version)),
```
zwang28, line 129: I just want to leave a note here.

This `LocalHummockVersion::from` is an addtional `HummockVersion` conversion introduced by this PR. However I don't think it will have significant performance implications, as it primarily involves move semantics.

src/storage/src/hummock/store/version.rs
```diff
@@ -992,11 +992,15 @@ impl HummockVersionReader {
...
         options: ReadLogOptions,
     ) -> HummockResult<ChangeLogIterator> {
-        let change_log = if let Some(change_log) = version.table_change_log.get(&options.table_id) {
-            change_log.filter_epoch(epoch_range).collect_vec()
-        } else {
-            Vec::new()
+        let change_log = {
+            let table_change_logs = version.table_change_log().read();
+            if let Some(change_log) = table_change_logs.get(&options.table_id) {
+                change_log.filter_epoch(epoch_range).cloned().collect_vec()
```
zwang28, line 998: This cloned() is an additional cost introduced in this PR.

If multiple iter_log processes are running simultaneously, will the memory usage be substantial?

Li0k: cc @wenym1 , suggests that `iter_log` is executed less frequently and that this clone is acceptable.

zwang28 (approved):
LGTM

Li0k:
Two interfaces are affected:

- iter_log
- next_epoch 

cc @wenym1

src/storage/src/hummock/event_handler/hummock_event_handler.rs
```diff
@@ -518,31 +520,59 @@ impl HummockEventHandler {
...
-                let mut version_to_apply = (*pinned_version).clone();
-                for version_delta in &version_deltas {
-                    assert_eq!(version_to_apply.id, version_delta.prev_id);
-                    if let Some(sst_delta_infos) = &mut sst_delta_infos {
-                        sst_delta_infos.extend(
-                            version_to_apply
-                                .build_sst_delta_infos(version_delta)
-                                .into_iter(),
+                let mut version_to_apply = pinned_version.version().clone();
+                let table_change_log_to_apply = pinned_version.take_change_log();
```
wenym1, line 530: Instead of `take_change_log`, we can have method `change_log_write_lock` to return the write lock on the `table_change_log`. And then in `new_pin_version_with_table_change_log`, we don't have to pass the `table_change_log_to_apply` again. We can use the original `table_change_log` wrapped by `Arc` and simply clone the `Arc`.

Besides, it seems that the parameter `pinned_version` of the current method is not necessary to take the ownership. We can change to pass `&PinnedVersion`, so that the caller can avoid doing clone on it.

wenym1 (commented):
Rest LGTM. Thanks for the PR.

src/storage/hummock_sdk/src/version.rs
```diff
@@ -476,10 +473,26 @@ impl HummockVersion {
             state_table_info_delta: Default::default(),
         }
     }
+
+    pub fn split_change_log(mut self) -> (LocalHummockVersion, HashMap<TableId, TableChangeLog>) {
+        let table_change_log = std::mem::take(&mut self.table_change_log);
```
wenym1, line 478: Here after `std::mem::take(&mut self.table_change_log)`, the `epochs` in `self.table_change_log` becomes empty. But we still need to use the `epochs` in `LocalHummockVersion`

src/storage/hummock_sdk/src/version.rs
```diff
@@ -217,17 +219,19 @@ impl HummockVersionStateTableInfo {
...
     #[deprecated]
     pub(crate) max_committed_epoch: u64,
     pub table_watermarks: HashMap<TableId, Arc<TableWatermarks>>,
-    pub table_change_log: HashMap<TableId, TableChangeLogCommon<T>>,
+    pub table_change_log: HashMap<TableId, TableChangeLogCommon<L>>,
     pub state_table_info: HummockVersionStateTableInfo,
 }
 
-pub type HummockVersion = HummockVersionCommon<SstableInfo>;
+pub type HummockVersion = HummockVersionCommon<SstableInfo, SstableInfo>;
```
wenym1, line 232:
```suggestion
pub type HummockVersion = HummockVersionCommon<SstableInfo>;
```

src/storage/hummock_sdk/src/version.rs
```diff
@@ -488,11 +501,13 @@ pub struct HummockVersionDeltaCommon<T> {
     pub trivial_move: bool,
     pub new_table_watermarks: HashMap<TableId, TableWatermarks>,
     pub removed_table_ids: HashSet<TableId>,
-    pub change_log_delta: HashMap<TableId, ChangeLogDeltaCommon<T>>,
+    pub change_log_delta: HashMap<TableId, ChangeLogDeltaCommon<L>>,
     pub state_table_info_delta: HashMap<TableId, StateTableInfoDelta>,
 }
 
-pub type HummockVersionDelta = HummockVersionDeltaCommon<SstableInfo>;
+pub type HummockVersionDelta = HummockVersionDeltaCommon<SstableInfo, SstableInfo>;
```
wenym1, line 508:
```suggestion
pub type HummockVersionDelta = HummockVersionDeltaCommon<SstableInfo>;
```

src/storage/src/hummock/local_version/pinned_version.rs
```diff
@@ -111,14 +117,42 @@ impl PinnedVersion {
...
+            guard: Arc::new(PinnedVersionGuard::new(
+                version_id,
+                self.guard.pinned_version_manager_tx.clone(),
+            )),
+            table_change_log: Arc::new(RwLock::new(table_id_to_change_logs)),
+            version: Arc::new(local_version),
+        })
+    }
+
+    pub fn new_with_change_log(
```
wenym1, line 132: We can use `self.table_change_log.clone()` instead of passing from outside.

src/storage/src/hummock/event_handler/hummock_event_handler.rs
```diff
@@ -518,31 +521,61 @@ impl HummockEventHandler {
...
-                        sst_delta_infos.extend(
-                            version_to_apply
-                                .build_sst_delta_infos(version_delta)
-                                .into_iter(),
+                let mut version_to_apply = pinned_version.deref().clone();
+                {
+                    let mut table_change_log_to_apply_guard =
+                        pinned_version.table_change_log_write_lock();
+                    // let mut table_change_log_to_apply_guard = table_change_log_to_apply.write();
+                    for version_delta in &version_deltas {
```
wenym1, line 535:
```suggestion
                    for version_delta in version_deltas {
```

And then below we use do 
```
let local_hummock_version_delta = LocalHummockVersionDelta::from(version_delta);
```

src/storage/src/hummock/event_handler/hummock_event_handler.rs
```diff
@@ -518,31 +521,61 @@ impl HummockEventHandler {
...
             HummockVersionUpdate::VersionDeltas(version_deltas) => {
-                let mut version_to_apply = (*pinned_version).clone();
-                for version_delta in &version_deltas {
-                    assert_eq!(version_to_apply.id, version_delta.prev_id);
-                    if let Some(sst_delta_infos) = &mut sst_delta_infos {
-                        sst_delta_infos.extend(
-                            version_to_apply
-                                .build_sst_delta_infos(version_delta)
-                                .into_iter(),
+                let mut version_to_apply = pinned_version.deref().clone();
```
wenym1, line 530:
```suggestion
                let mut version_to_apply = (**pinned_version).clone();
```

src/storage/hummock_sdk/src/version.rs
Li0k: Good catch  !!!

Li0k:
@wenym1 PTAL

wenym1 (approved):
Rest LGTM!

PR #20053, opened 2025-01-07
feat: add support for global max_adaptive_parallelism

zwang28:

Before this PR, for a streaming job of adpative parallelism (i.e. default_parallelism="Full"), it will always use all available parallelism units in the clusters. It's not suitable for scenarios that scale up or out the cluster, where the goal is to add resources for accommodating additional streaming jobs. Because all existing streaming jobs will also scale up or out, creaing more actors and consuming additional resources unnecessarily.

This PR extends config parameter default_parallelism by adding a new mode `Full(max_parallelism)`, alongside the existing modes `Full` and `Fixed(parallelism)`. With `Full(max_parallelism)`, a streaming job of adaptive parallelism is capped at `max_parallelism`. As a config parameter, it takes effect when:
1. Creating a new streaming job.
2. Recovering or bootstraping the cluster.

## Checklist

## Documentation

<details>

</details>


BugenZhao (commented):
I think we are going to deprecate `default_parallelism` in the config file as it can be superseded by setting the (session) variable `streaming_parallelism` globally via `ALTER SYSTEM SET`. Am I correct? cc @yezizp2012

fuyufjh:
The motivation seems to be similar with [RFC: clarify streaming job’s parallelism strategy](https://www.notion.so/risingwave-labs/RFC-clarify-streaming-job-s-parallelism-strategy-175f0d69cb7680d096e9cfc700a67ce0#175f0d69cb7680b68b40d0f5b4036f06). Have you 2 discussed offline?

st1page:
> I think we are going to deprecate `default_parallelism` in the config file as it can be superseded by setting the (session) variable `streaming_parallelism` globally via `ALTER SYSTEM SET`. Am I correct? cc @yezizp2012
> 
> Also mentioned in https://www.notion.so/risingwave-labs/RFC-clarify-streaming-job-s-parallelism-strategy-175f0d69cb7680d096e9cfc700a67ce0?pvs=4#175f0d69cb7680b68b40d0f5b4036f06

Yes, actually this PR inspired me to write the RFC, so let's discuss it later.

zwang28:
go for the RFC

Issue #18949, opened 2024-10-16
Meta node can only handle ~600 Kafka sources.


ka-weihe:
### Describe the bug

I attempted to create 700 Kafka Plain Avro sources. However, after successfully creating approximately 600 sources, the meta-node becomes unresponsive, and I encounter the following error:

```
ERROR librdkafka: librdkafka: THREAD [thrd:main]: Unable to create broker thread
```

Additionally, when I shell into the meta-node and try to execute any commands, I get:

```
bash: fork: retry: Resource temporarily unavailable
```

To investigate, I ran the command `(ps -eLf | wc -l)` to monitor the number of OS threads while creating the sources. The thread count reached 32,769 after I had created 629 sources, at which point I could no longer execute commands—likely due to the OS exhausting its available resources for creating threads.

This behavior seems like a bug. I would expect the number of threads to be significantly lower than the number of sources. Even if there was only one OS thread per source, while inefficient, it would still fit within our use case requirements, but we are seeing more than 50 OS threads per source. It’s also worth noting that none of the sources were being actively used at any time.

### Error message/log

```text
ERROR librdkafka: librdkafka: THREAD [thrd:main]: Unable to create broker thread
```

### To Reproduce

- Setup a Kafka cluster with 50 brokers in K8s
- Create 1000 Avro topics with 16 partitions
- Deploy a simple RisingWave with the RisingWave operator with Postgres metastore and MinIO statestore (in K8s)
- Connect and create 1 source for each topic.

### Expected behavior

I expect the sources to be created and the Meta node to not be unresponsive and use many resources.

### How did you deploy RisingWave?

More or less like this:
https://github.com/risingwavelabs/risingwave-operator/blob/main/docs/manifests/risingwave/risingwave-postgresql-s3.yaml

But higher limits

### The version of RisingWave

 PostgreSQL 13.14.0-RisingWave-2.0.1 (0d15632ac8d18fbd323982b53bc2d508dc7e148a)

xxchan:
According to https://github.com/confluentinc/librdkafka/issues/1600, it seems each kafka consumer will create ~`num_brokers` threads, and threads cannot be shared across client instances.

In Meta, currently each source has a `KafkaSplitEnumerator`, which has a Kafka consumer, so this looks like exactly your problem. (1 source ~ 50 threads) In theory, we could improve Meta by sharing consumer connecting to the same broker.

[...]

PR #20065, opened 2025-01-08
feat: delete Kafka consumer group on drop

### LINKED ISSUE #18416: delete Kafka consumer group when job is dropped
A user complaint that RisingWave created too many consumer groups and exceeded the limit in Kafka. They have to manually clean dead groups.

Although we only have 1 consumer group for 1 MV now (A long time ago, it was `consumer-{current_time}` , then it was `rw-{fragment_id}-{actor_id}`, now `{prefix:rw}-{fragment_id}`), it may still increase quickly:
They have ~50 sources, but multiple MVs on the same source. So when they recreate MVs, the number of consumer groups will increase a lot.

There are several solutions
1. (This is what I suggested them to do now.) Use `TABLE` instead of `SOURCE`. So recreating MVs will only have MV backfill, and won't consume from Kafka. Drawback: storage.
2. Allow specifying custom group id. (Now we supported custom group id prefix, and will still generate a suffix.) So when they drop and recreate a job, they can reuse the same group. 

    Flink supports this. But this may be error-prone. If we have multiple MVs on the same source, different MVs' Kafka consumers will report offset to the same consumer group, and will make it a mess. (Note: this only affect offset monitoring, but not correctness.) Since multiple MVs on the same source is a natural use case in RisingWave, I don't suggest to do it. (If really necessary, we might add a config like `unsafe_exact_group_id=xxx`)
3. Delete consumer group on drop. This is reasonable, and perhaps is good for most users. Some possible problems:

    Users may want to retain the consumer group after drop? I guess this is unlikely, since the groups are dedicated to RisingWave.

    More importantly, deleting consumer group requires higher authorization permission. Note that currently the creation is automatically done by Kafka consumers implicitly, instead of manual creating via admin API. 

    We might introduce a config option `delete_on_drop`, and make sure delete failure won't block DROP.

Note: Shared source (#16003) may look like 1 (i.e., materialized source). But it will do source backfill, which will also create consumer groups.

xxchan:

fix #18416 

Delete Kafka consumer group when deleting source fragments. (Backfill is handled in the next PR) i.e.,
- DROP (shared) SOURCE
- DROP MV on (non-shared) source

Impl details:
- The deletion is done in meta, when handling source changes.
- Source manager sends a command to the Worker (which owns the SplitEnumerator)
- When error happens, it will be ignored.

## Checklist

## Documentation

<details>

Now Kafka consumer groups created by RisingWave Kafka source will be deleted when the sources or related materialized views are dropped.

</details>


xxchan:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20065?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#20065** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20065?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20065?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#20071** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20071?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

stdrc (approved):
LGTM

xxchan:
```
2025-01-09T06:54:23.898047083Z DEBUG risingwave_connector::source::kafka::enumerator: delete groups result: [Ok("my_group_non_shared-543")] topic="test_consumer_group_non_shared" fragment_ids=[543]





failed to run `e2e_test/source_inline/kafka/consumer_group_non_shared.slt.serial`
  |  
  | Caused by:
  | system command stdout mismatch:
  | [command] ./e2e_test/source_inline/kafka/consumer_group.mjs count-groups \| grep my_group_non_shared
  | [Diff] (-expected\|+actual)
  | -   my_group_non_shared: 1
  | +   my_group_non_shared: 2
  | at e2e_test/source_inline/kafka/consumer_group_non_shared.slt.serial:115
```

strange

src/meta/src/stream/source_manager/worker.rs
```diff
@@ -317,4 +334,50 @@ impl ConnectorSourceWorkerHandle {
...
+
+    /// Force [`ConnectorSourceWorker::tick()`] to be called.
+    pub async fn force_tick(&self) -> MetaResult<()> {
+        let (tx, rx) = oneshot::channel();
+        self.send_command(SourceWorkerCommand::Tick(tx))?;
+        rx.await
+            .context("failed to receive tick command response from source worker")?
+            .context("source worker tick failed")?;
+        Ok(())
+    }
```
tabVersion, line 354: I don't fully agree with making `force_tick` rely on the command channel. 
In prev impl, each enumerator's `force_tick` is independent with each other. But in this impl, each request to `force_tick` must be handled sequentially.

Consider this case, a channel containing `(drop_fragment(x), force_tick(y))`, when handling `drop_fragment(x)`, the network to x's broker is unstable and the call may hang for a while, or forever. But it blocks call for `force_tick(y)`.
A real-world case, we call `force_tick` when creating a source, the behavior to the user is like everything is good but SQL never returns, though the root cause is irrelevant to the creating one. It can cause much trouble when doing POC.

tabVersion (commented):
One more thing to confirm, we do not remove source backfill's consumer group until the relevant streaming job is dropped.

src/connector/src/source/kafka/enumerator.rs
```diff
@@ -196,6 +168,70 @@ impl SplitEnumerator for KafkaSplitEnumerator {
 
         Ok(ret)
     }
+
+    async fn on_drop_fragments(&mut self, fragment_ids: Vec<u32>) -> ConnectorResult<()> {
+        let admin = build_kafka_admin(&self.config, &self.properties).await?;
```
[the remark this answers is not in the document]
xxchan: I'm worried that this increases unnecessary idle clients.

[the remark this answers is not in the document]
xxchan: User also once asked why not close the connection after fetching metadata.

https://github.com/risingwavelabs/risingwave/issues/18949#issuecomment-2426395579

> If that is the case, why not close the consumer after fetching the metadata?

src/meta/src/stream/source_manager/worker.rs
xxchan: I don't get your point. The new impl is exactly the same as before

xxchan: > A real-world case, we call `force_tick` when creating a source

No it will only be called in rare cases when `splits` is `None`. When `CREATE SOURCE`, we call `create_source_worker`, which will `tick` before creating the source worker

xxchan: I don't think it's a big deal to handle `force_tick` separately here. If `drop_fragment` hangs, `tick` will also unlikely succeed

tabVersion: > If drop_fragment hangs, tick will also unlikely succeed

I don't seem to understand your logic here. Drop_fragment and force_tick can happen to different sources, why they tend to fail together?

src/connector/src/source/kafka/enumerator.rs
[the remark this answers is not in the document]
tabVersion: > > This seems to be the Kafka library's limitation. Even just for fetching metadata, we need to create consumers. And the consumers will immediately connect to all brokers (instead of only the bootstrap server)
> 
> If that is the case, why not close the consumer after fetching the metadata?

I think the user mistakenly believes we do nothing after the first fetch. But, actually, we keep querying the metadata and we cannot take the price of recreate one for each tick. 

> I'm worried that this increases unnecessary idle clients.

It is just 1 additional client per cluster, per one source. I think it should be acceptable. 
If you still have concerns, we may do a bench with MSK to see how large creating multiple clients simultaneously can impact the broker and local server.

src/meta/src/stream/source_manager/worker.rs
xxchan: 1 ConnectorSourceWorker corresponds to 1 SplitEnumerator. 

> In prev impl, each enumerator's force_tick is independent with each other

enumerators are still independent with each other now

src/connector/src/source/kafka/enumerator.rs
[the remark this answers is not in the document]
xxchan: > we cannot take the price of recreate one for each tick.

For `tick` yes, but for `drop_fragment` here it's by nature short-lived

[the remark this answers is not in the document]
tabVersion: as I mentioned before, 

> "I am afraid if iter over all source and create an admin client for each one when doing migration will put too much pressure on kafka broker." 

the thing is about creating many clients in a short period of time rather than how long each client lives. 

And again, a benchmark can resolve our gap here, hope it can correct my misunderstanding of the overhead of creating Kafka clients.

src/meta/src/stream/source_manager/worker.rs
xxchan: 1 source - 1 split enumerator - 1 source worker - 1 channel for drop_fragment & force_tick

So they don't "happen to different sources" :eyes:

xxchan:
> ```
> 2025-01-09T06:54:23.898047083Z DEBUG risingwave_connector::source::kafka::enumerator: delete groups result: [Ok("my_group_non_shared-543")] topic="test_consumer_group_non_shared" fragment_ids=[543]
> 
> 
> 
> 
> 
> failed to run `e2e_test/source_inline/kafka/consumer_group_non_shared.slt.serial`
>   |  
>   | Caused by:
>   | system command stdout mismatch:
>   | [command] ./e2e_test/source_inline/kafka/consumer_group.mjs count-groups \| grep my_group_non_shared
>   | [Diff] (-expected\|+actual)
>   | -   my_group_non_shared: 1
>   | +   my_group_non_shared: 2
>   | at e2e_test/source_inline/kafka/consumer_group_non_shared.slt.serial:115
> ```
> 
> strange

I guess this is because the work in source manager happened before stream manager, fixed

src/connector/src/source/kafka/enumerator.rs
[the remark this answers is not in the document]
xxchan: I did some basic testing (MSK and EC2 on the same VPC, create thousands of clients sequentially) and found that simply "create clients" (and fetch metadata) is very quick and basically has neglectable overhead. (<1%CPU on both client and broker). 

It may help if you can tell where you think the overhead lies.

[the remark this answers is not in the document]
xxchan: I just realized we could use time based cache eviction

xxchan:
Merged first. Feel free to comment more and I can resolve later

src/meta/src/stream/source_manager/worker.rs
tabVersion: My mistake, the command channel is built inside each worker rather than the manager.

PR #20097, opened 2025-01-10
feat: derive monotonicity for date_truc with timezone when unit larger than day

st1page:
…r than day

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/optimizer/property/monotonicity.rs
+++ b/src/frontend/src/optimizer/property/monotonicity.rs
@@ -186,6 +186,23 @@ impl MonotonicityAnalyzer {
             tz_is_utc // conservative
         }
 
+        fn date_unit_larger_than_day(date_unit: Option<&str>) -> bool {
+            const DAY: &str = "day";
+            const WEEK: &str = "week";
+            const MONTH: &str = "month";
+            const QUARTER: &str = "quarter";
+            const YEAR: &str = "year";
+            const DECADE: &str = "decade";
+            const CENTURY: &str = "century";
+            const MILLENNIUM: &str = "millennium";
+            date_unit.map_or(false, |date_unit| {
+                matches!(
+                    date_unit.to_ascii_lowercase().as_str(),
+                    DAY | WEEK | MONTH | QUARTER | YEAR | DECADE | CENTURY | MILLENNIUM
+                )
+            })
+        }
+
         match func_call.func_type() {
             ExprType::Unspecified => unreachable!(),
             ExprType::Add => match self.visit_binary_op(func_call.inputs()) {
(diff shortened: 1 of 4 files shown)


src/frontend/src/optimizer/property/monotonicity.rs
```diff
@@ -257,7 +274,13 @@ impl MonotonicityAnalyzer {
...
                             .and_then(|literal| literal.get_data().as_ref())
                             .map(|tz| tz.as_utf8().as_ref());
-                        if time_zone_is_without_dst(time_zone) {
+                        let date_unit = func_call.inputs()[0]
+                            .as_literal()
+                            .and_then(|literal| literal.get_data().as_ref())
+                            .map(|tz| tz.as_utf8().as_ref());
+                        if time_zone_is_without_dst(time_zone)
+                            || date_unit_larger_than_day(date_unit)
+                        {
```
stdrc, line 283: Can you please add an example to demonstrate the correctness?

What I can think of is like in a timezone with DST, `MM-dd 23:50:00` + interval `12 minutes` may be `MM-dd 23:02:00`, while the `date_trunc('day', ...)` result be `MM-dd 00:12:00` versus `MM-dd (00:00:00)`. Seems this will break the correctness. Am I understanding it wrong?

xxchan: https://arc.net/l/quote/wfxyyukb

src/frontend/src/optimizer/property/monotonicity.rs
```diff
@@ -186,6 +186,23 @@ impl MonotonicityAnalyzer {
...
+            const YEAR: &str = "year";
+            const DECADE: &str = "decade";
+            const CENTURY: &str = "century";
+            const MILLENNIUM: &str = "millennium";
+            date_unit.map_or(false, |date_unit| {
+                matches!(
+                    date_unit.to_ascii_lowercase().as_str(),
+                    DAY | WEEK | MONTH | QUARTER | YEAR | DECADE | CENTURY | MILLENNIUM
+                )
+            })
```
xxchan, line 203: what's the point of the `const` here

BugenZhao: In case SQL is standardized to use a language other than English in the future?

st1page: just maintain it same as the expression's implementation https://github.com/risingwavelabs/risingwave/blob/c06b0c6ba7009855e2a43630a8e3ceb4a5ead6ba/src/expr/impl/src/scalar/date_trunc.rs#L20-L33

st1page: This modification is a bit troublesome; we have to expose these const from a very deep level within expr... Since we might never modify it, I would rather make a copy.

st1page: I am too lazy....

xxchan: Your brain will become rusty!

Issue #19982, opened 2025-01-02
Support defining schema in STRUCT type for mongodb cdc


hzxa21:
Currently data ingested from mongodb cdc is parsed as the `JSONB` type. See example [here](https://github.com/risingwavelabs/risingwave/blob/05f9e9e7cfdfe52d958aaa6c4a917a067de56c6f/e2e_test/source_legacy/cdc/mongodb/mongodb_basic.slt#L5).

Recently we saw user requesting supporting strongly typed schema for data ingested by mongodb cdc, which leads to supporting defining strongly typed `STRUCT` in `CREATE TABLE` statement for mongodb cdc. Exmaple:
```
CREATE TABLE mongdb_source_table (
    payload STRUCT<
        foo STRUCT<
            foo1 JSONB,
            foo2 BIGINT
        >,
        bar STRUCT<
            bar1 STRUCT<
                bar1_1 VARCHAR,
                bar1_2 VARCHAR,
                bar1_3 STRUCT<
                    bar1_3_array STRUCT<
                        item1 DOUBLE,
                        item2 BIGINT,
                        ...
                    >[]
                >,
         ...
) WITH (
  connector=mongodb-cdc
  ...
);
```

We already support parsing jsonb to struct via [josnb_populate_record](https://docs.risingwave.com/sql/functions/json#jsonb-populate-record) so we can borrow the implementation here as well. The expected behavior is:
- If the fields of a document ingested from mongodb is exactly the same as the defined struct, the document will be parsed and ingested successfully.
- If the fields of a document ingested from mongodb is a superset of the defined struct, the document will be parsed and ingested successfully by ignoring the extra fields.
- If the fields of a document ingested from mongodb is a subset of the defined struct, the document will be parsed and ingested successfully by filling in NULL for the missing fields.
- If the payload a given field in the document ingested from mongodb cannot be converted to the type defined in the struct (e.g. cannot cast text to int), the whole document will be skipped by filling in NULL for the struct.

xiangjinwu:
The conversion rule may be different from `jsonb_populate_record`. Or in other words, AVOID updating `jsonb_populate_record`  behavior to be closer to their usage in MongoDB. The function is supposed to be PostgreSQL compatible.

It can be a dialect here instead:
https://github.com/risingwavelabs/risingwave/blob/17c4cd2b6e8bd5988230c55f856c1c82fce2422c/src/connector/src/parser/unified/json.rs#L154-L155

[...]

Issue #17264, opened 2024-06-14
feat(source): protobuf field presence


xiangjinwu:
Just opening an issue to acknowledge this has not been supported yet.

Without field presence (default), missing fields from a message would be filled with their default value (e.g. `0` or `''`) rather than `null`. This used to be the common practice of protobuf.

However in some cases, it may be desired to use `null` for unset fields and only ingest `0`/`''` when the producer explicitly set them to this value.

This is related to `optional` and `oneof` in protobuf specification.

https://protobuf.dev/programming-guides/field_presence/

stefreak:
I'd really like to have this feature, or at least be able to have an automatically generated discriminator when using `oneof`.

PR #15304, opened 2024-02-27
feat: add new option `ignore_option` for debeizum format

tabVersion:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://gist.github.com/TennyZhuang/f00be7f16996ea48effb049aa7be4d66#file-rw_cla).

## What's changed and what's your intention?

add a format spec option `ignore_key` for `format DEBEZIUM`. (accepts `true` and `false`, default to `false`)

The option controls whether to ignore the key part for given messages. If set to true, debezium format will work with only the payload part.
Note that in this case, RW cannot handle tombstone message. So users should guarantee that payload MUST NOT be empty. 

## Checklist

- [ ] I have written necessary rustdoc comments
- [ ] I have added necessary unit tests and integration tests
- [ ] I have added test labels as necessary. See [details](https://github.com/risingwavelabs/risingwave/blob/main/docs/developer-guide.md#ci-labels-guide).
- [ ] I have added fuzzing tests or opened an issue to track them. (Optional, recommended for new SQL features #7934).
- [ ] My PR contains breaking changes. (If it deprecates some features, please create a tracking issue to remove them in the future).
- [x] All checks passed in `./risedev check` (or alias, `./risedev c`)
- [ ] My PR changes performance-critical code. (Please run macro/micro-benchmarks and show the results.)

- [ ] My PR contains critical fixes that are necessary to be merged into the latest release. (Please check out the [details](https://github.com/risingwavelabs/risingwave/blob/main/CONTRIBUTING.md))

## Documentation

- [x] My PR needs documentation updates. (Please use the **Release note** section below to summarize the impact on users)

## Release note

If this PR includes changes that directly affect users or other significant modifications relevant to the community, kindly draft a release note to provide a concise summary of these changes. Please prioritize highlighting the impact these changes will have on users.

as described above

</details>

PR #20100, opened 2025-01-10
feat: mongo parse strong schema

ClSlaid:

solves: #19982 

## Checklist

## Documentation

<details>

</details>


ClSlaid:
Oops...this PR seems biased with the original issue.

src/connector/src/parser/unified/debezium.rs
```diff
@@ -535,9 +540,376 @@ pub fn extract_bson_id(id_type: &DataType, bson_doc: &serde_json::Value) -> Acce
...
+                Some(ScalarImpl::Timestamp(dt))
+            }
+            DataType::Timestamptz => {
+                let dt = chrono_datetime.into();
+                Some(ScalarImpl::Timestamptz(dt))
+            }
+            _ => unreachable!("DebeziumMongoJsonParser::new must ensure column datatypes."),
+        }
+    } else if let serde_json::Value::Object(ref obj) = datum
+        && obj.contains_key("date")
```
hzxa21, line 812: Should be `$numberLong` here?

src/connector/src/parser/unified/debezium.rs
```diff
@@ -535,9 +540,376 @@ pub fn extract_bson_id(id_type: &DataType, bson_doc: &serde_json::Value) -> Acce
...
+            } else {
+                return Err(type_error(datum));
+            }
+        }
+        DataType::Decimal => {
+            if let serde_json::Value::Object(ref obj) = datum
+                && obj.contains_key("$numberDecimal")
+            {
+                let number = obj["$numberDecimal"].as_str().unwrap_or_default();
+                let dec = risingwave_common::types::Decimal::from_str(number).unwrap_or_default();
```
hzxa21, line 685: BSON seems to support the standard [decimal-128](https://github.com/mongodb/specifications/blob/master/source/bson-decimal128/decimal128.md) so there can be parsing error here as well.

src/connector/src/parser/unified/debezium.rs
```diff
@@ -535,9 +540,376 @@ pub fn extract_bson_id(id_type: &DataType, bson_doc: &serde_json::Value) -> Acce
...
+                Some(ScalarImpl::Decimal(dec))
+            } else {
+                return Err(type_error(datum));
+            }
+        }
+
+        DataType::Bytea => {
+            if let serde_json::Value::Object(ref obj) = datum
+                && obj.contains_key("$binary")
```
hzxa21, line 694: The [bson doc](https://www.mongodb.com/docs/manual/reference/mongodb-extended-json/#mongodb-bsontype-Binary) suggests that the binary type is in this format:
```
{ "$binary":
   {
      "base64": "<payload>",
      "subType": "<t>"
   }
}
```

The implementation seems using the wrong format. Am I missing something?

hzxa21 (commented):
Left some comments.

Can you separate the integration test related changes from this PR if the plan is not doing integration tests for now?

src/connector/src/source/base.rs
```diff
@@ -592,6 +592,18 @@ impl ConnectorProperties {
             || matches!(self, ConnectorProperties::Azblob(_))
     }
 
+    /// Whether to enable strong schema for the source.
+    pub fn enable_strong_schema(&self) -> bool {
+        let ConnectorProperties::MongodbCdc(mongo) = self else {
+            return false;
+        };
```
xiangjinwu, line 599: As a general method on `ConnectorProperties` this feels like _all other connectors do not support strong schema except for mongo_. This is definitely not the truth. The truth is that only mongo defaults to weak schema and requires explicit opt-in.

Consider inline such logic in mongo-specific code, and you can use `try_match_expand!(source_ctx.connector_props, ConnectorProperties::MongodbCdc)?` rather than defaulting to `false`.

src/connector/src/parser/debezium/mongo_json_parser.rs
```diff
@@ -112,6 +125,12 @@ impl DebeziumMongoJsonParser {
     }
 }
 
+impl DebeziumMongoJsonParser {
+    pub fn strong_schema(&self) -> bool {
```
xiangjinwu, line 129: Seems unused

src/connector/src/parser/unified/debezium.rs
```diff
@@ -535,9 +540,376 @@ pub fn extract_bson_id(id_type: &DataType, bson_doc: &serde_json::Value) -> Acce
...
+            } else {
+                return Err(type_error(datum));
+            }
+        }
+        DataType::Int32 => {
+            if let serde_json::Value::Object(ref obj) = datum
+                && obj.contains_key("$numberInt")
+            {
+                let int_str = obj["$numberInt"].as_str().unwrap_or_default();
+                Some(ScalarImpl::Int32(int_str.parse().unwrap_or_default()))
```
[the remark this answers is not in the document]
xiangjinwu: No we do not fill in rust `Default` trait. These just look like copy-pasta code from `extract_bson_id`. So the question goes to why they existed when extracting id.

However in the following cases SQL `NULL` is not used either:
* Protobuf default value on missing fields. [#17264](https://github.com/risingwavelabs/risingwave/issues/17264)
* Avro field can have a non-trivial user specified value (e.g. `99` in [this PR](https://github.com/risingwavelabs/risingwave/pull/20699/files)).
* ~~MySQL/PostgreSQL CDC upstream column can have a non-trivial user specified value~~ (reverted by [#20671](https://github.com/risingwavelabs/risingwave/pull/20671))

src/connector/src/parser/unified/debezium.rs
```diff
@@ -535,9 +540,376 @@ pub fn extract_bson_id(id_type: &DataType, bson_doc: &serde_json::Value) -> Acce
...
+            serde_json::Value::Array(_) => "array",
+            serde_json::Value::Object(_) => "object",
+        }
+        .to_owned(),
+        value: datum.to_string(),
+    };
+
+    let res = if let serde_json::Value::String(ref _iso8601_str) = datum {
+        // relaxed format
+        let Ok(parsed) = iso8601::deserialize(datum) else {
```
xiangjinwu, line 783: There is already `chrono` or `speedate`, why `time` crate here?

ClSlaid: This part is used for relaxed format, won't do in this PR, I'll remove it.

src/connector/src/parser/unified/debezium.rs
```diff
@@ -559,7 +1089,24 @@ where
...
                 }
             }
-            ["after" | "before", "payload"] => self.access(&[path[0]], &DataType::Jsonb),
+
+            ["after" | "before", "payload"] if !self.strong_schema => {
+                self.access(&[path[0]], &DataType::Jsonb)
+            }
+
+            ["after" | "before", field] if self.strong_schema => {
+                let payload = self.access_owned(&[path[0]], &DataType::Jsonb)?;
```
xiangjinwu, line 1098: We can:
* Access `after`/`before` as `DataType::Varchar` (to avoid parsing into `Jsonb` first)
* Use the official `bson` crate to parse this EJSONv2 string into `Bson`. ([doc](https://docs.rs/bson/latest/bson/extjson/index.html#deserializing-extended-json))

It looks liked your just implemented your own EJSONv2 parser 😂

xiangjinwu: Oh it seems debezium outputs EJSONv1 despite Mongo website says Java driver outputs EJSONv2. Furthermore, out of v2 canonical mode, v2 relaxed mode, and v1 strict mode, a bson `date` is either:
* `{"$date": {"$numberLong": "1734607421137"}}`
* `{"$date": "2024-12-19T11:23:41.137Z"}`

However debezium actually outputs:
* `{"$date": 1734607421137}`

And the `bson` crate rejects such invalid EJSON. So we do need to implement our own parse to tolerate such debezium issues.

> *Legacy Extended JSON* - A string format based on the JSON standard that describes a BSON document. The Legacy Extended JSON format does not describe a specific, standardized format, and many tools, drivers, and libraries implement Extended JSON in conflicting ways.

xiangjinwu: https://github.com/mongodb/specifications/blob/master/source/extended-json/extended-json.md
> Mongoexport 2.4 and the Java driver always transform a Datetime object into an Extended JSON string of the form `{"$date": <ms since epoch>}`. This form has the problem of a potential loss of precision or range on the Datetimes that can be represented. Mongoexport 2.6 transforms Datetime objects into an extended JSON string of the form `{"$date": <ISO-8601 date string in local time>}`for dates starting at or after the Unix epoch (UTC). Dates prior to the epoch take the form `{"$date": {"$numberLong": "<ms since epoch>"}}`. Starting in version 3.0, mongoexport always turns Datetime objects into strings of the form `{"$date": <ISO-8601 date string in UTC>}`. The NPM mongodb-extended-json module does the same. The Python driver can also transform Datetime objects into strings like `{"$date": {"$numberLong": "<ms since epoch>"}}`. This specification canonicalises this form, since this form is the most portable. In Relaxed Extended JSON format, this specification provides for ISO-8601 representation for better readability, but limits it to a portable subset, from the epoch to the end of the largest year that can be represented with four digits. This should encompass most typical use of dates in applications.

src/connector/src/parser/debezium/mongo_json_parser.rs
```diff
@@ -63,25 +67,43 @@ impl DebeziumMongoJsonParser {
...
 
-        // _rw_{connector}_file/partition & _rw_{connector}_offset are created automatically.
-        if rw_columns
-            .iter()
-            .filter(|desc| desc.is_visible() && desc.additional_column.column_type.is_none())
-            .count()
-            != 2
-        {
-            bail!("Debezium Mongo needs no more data columns except `_id` and `payload` in table");
+        let strong_schema = match &source_ctx.connector_props {
```
ClSlaid, line 71: This does not guarantee that only MongodbCdc properties would be received here, strange.

src/connector/src/parser/unified/debezium.rs
```diff
@@ -535,9 +540,554 @@ pub fn extract_bson_id(id_type: &DataType, bson_doc: &serde_json::Value) -> Acce
...
+            if *type_expected == DataType::Float64 {
+                return Ok(Some(ScalarImpl::Float64(parsed_num.into())));
+            } else {
+                let parsed_num = parsed_num as f32;
+                return Ok(Some(ScalarImpl::Float32(parsed_num.into())));
+            }
+        }
+        // parse to large int
+        if *type_expected == DataType::Int256 {
+            let parsed_num = match i256::from_string(num_str) {
```
xiangjinwu, line 811: Use `risingave_common::types::Int256::from_str`. Avoid coupling mongodb with deltalake

src/connector/src/parser/unified/debezium.rs
```diff
@@ -535,9 +540,554 @@ pub fn extract_bson_id(id_type: &DataType, bson_doc: &serde_json::Value) -> Acce
...
+/// BSON document is a JSON object with some special fields, such as:
+/// long integer: {"$numberLong": "1630454400000"}
+/// date time: {"$date": {"$numberLong": "1630454400000"}}
+///
+/// For now, we support only the Canonical format of the date and timestamp.
+// similar to extract the "_id" field from the message payload
+pub fn extract_bson_field(
+    type_expected: &DataType,
+    bson_doc: &serde_json::Value,
+    field: Option<&str>,
```
xiangjinwu, line 554: Add to doc comment above that `field` is `None` when accessing elements within an array.

xiangjinwu (approved):
Approved to unblock further testing and improvement. It is hard to polish everything in a single PR.

THIS FEATURE IS NOT COMPLETE. More to do before we declare this feature is done:
* e2e / integration tests (which will expose some issues below)
* handle `kafka + format debezium_mongo` in addition to `mongodbcdc`
* handle differences in extjson variants produced by different versions of mongodb (java drivers)
  * It is better to **contribute to debezium** to (add an option to) switch to `JsonMode.EXTENDED` from the deprecated `JsonMode.STRICT`
* avoid parsing the whole `before`/`after` repeatedly for each field access

<details>
<summary>how strong_schema is currently passed</summary>

* `DebeziumMongoJsonParser::new`
* `build_accessor_builder`
* `DebeziumMongoJsonAccessBuilder::new`
* `DebeziumMongoJsonAccessBuilder`
* `MongoJsonAccess::new`
* `MongoJsonAccess`
</details>

src/connector/src/parser/debezium/mongo_json_parser.rs
ClSlaid: > And there is one more check in the frontend that enforces it to have the only 2 columns [here](https://github.com/risingwavelabs/risingwave/blob/72bbe0aa8a4d3e0c55d39c7e5bc5c08b7472d1ee/src/frontend/src/handler/create_source.rs#L318), which only reads the FORMAT DEBEZIUM_MONGO ENCODE JSON (...); clause.

Then this PR is far from complete then. 🤔

I'll
+ Remove the check
+ Also allow 'strong_schema' placing in the tailing clause

Thanks for your patient and detailed reviews! 😊

xiangjinwu: The CI error is due to `strong_schema` not being part of `KafkaProperties::unknown_fields`. You can refer to #15304 for handling options after `format xxx encode xxx`. Again, it is better to handle different issues in their own PRs.

xiangjinwu (approved):
At least we have one working e2e test. Let's move on ...

PR #20154, opened 2025-01-14
feat: make `privatelink.endpoint` accept json array

tabVersion:

resolve cloud issue

## Checklist

## Documentation

<details>

```
        // new syntax: endpoint can either be a string or a json array of strings
        // if it is a string, rewrite all broker addresses to the same endpoint
        // eg. privatelink.endpoint='some_url'
        // ==> broker1:9092 -> some_url:9092, broker2:9093 -> some_url:9093
        //
        // if it is a json array, rewrite each broker address to the corresponding endpoint
        // eg. privatelink.endpoint = '[{"host": "aaaa"}, {"host": "bbbb"}, {"host": "cccc"}]'
        // ==> broker1:9092 -> aaaa:9092, broker2:9093 -> bbbb:9093, broker3:9094 -> cccc:9094
```

</details>


src/connector/src/source/kafka/private_link.rs
```diff
@@ -166,3 +175,126 @@ pub fn insert_privatelink_broker_rewrite_map(
...
+                )
+            })?;
+            // rewrite the broker address to endpoint:port
+            broker_rewrite_map.insert(
+                broker.to_string(),
+                format!("{}:{}", host.as_str().unwrap(), link.port),
+            );
+        }
+    } else {
+        unreachable!()
```
fuyufjh, line 217: Please return error, otherwise an arbitrary json e.g. `{}` will cause panic.

src/connector/src/source/kafka/private_link.rs
```diff
@@ -30,6 +30,7 @@ use crate::error::ConnectorResult;
 use crate::source::kafka::{KAFKA_PROPS_BROKER_KEY, KAFKA_PROPS_BROKER_KEY_ALIAS};
 
 pub const PRIVATELINK_ENDPOINT_KEY: &str = "privatelink.endpoint";
```
wjf3121, line 32: Posting the same comment I raised during offline chat:

The key `privatelink.endpoint` is in singular form, which feels confusing that we allow users to specify multiple endpoints through this setting.

What about passing multiple endpoints through a new `privatelink.endpoints` key? One can take precedence over the other between the two keys.

tabVersion: I think the semantic here is not diff enough to introduce a new key. 
We can also tell the users to paste whatever the wizard shows after fixing the UI, without telling them to distinguish the two keys.

wjf3121: > We can also tell the users to paste whatever the wizard shows after fixing the UI.

Isn't the key used by non-cloud users as well?

src/connector/src/source/kafka/private_link.rs
```diff
@@ -166,3 +175,141 @@ pub fn insert_privatelink_broker_rewrite_map(
...
     Ok(())
 }
+
+fn handle_privatelink_endpoint(
+    endpoint: &str,
+    broker_rewrite_map: &mut HashMap<String, String>,
+    link_targets: &[AwsPrivateLinkItem],
+    broker_addrs: &[&str],
+) -> ConnectorResult<()> {
+    let endpoint = if let Ok(json) = serde_json::from_str::<serde_json::Value>(endpoint) {
```
tabVersion, line 185: here we introduce a potential risk is if the user input an illegal json, which cannot be parsed, we will see it as a string, instead of returning error. The case seems acceptable to me because he cannot connect to the broker using the wrong endpoint addr and he will find out later.

src/connector/src/source/kafka/private_link.rs
wjf3121: Discussed offline, we want to avoid introducing new keys as the original syntax is not very well designed.

tabVersion:
convert to draft for now as the poc user does not need it.

fuyufjh (approved):

PR #20156, opened 2025-01-14
fix(parquet): handle nested data types correctly

wcy-fdu:

[Here](https://github.com/risingwavelabs/risingwave/blob/f41dab78ddfb74e5e08f9b63d91ae2152b9f7802/src/connector/src/source/iceberg/parquet_file_handler.rs#L257), there was a bug previously where the columns in the parquet file metadata returned the flattened columns, while `parquet_to_arrow_schema`.fields returned the un flattened schema. Therefore, when the parquet file contains nested columns, such as Structs, their lengths differ, which can lead to an index out of bound panic.

This PR fixes the issue by not performing column trimming for parquet files that contain nested columns, and replaces the previous method of finding data types by index with a method that find them up by name.

## Checklist

## Documentation

<details>

</details>


wcy-fdu:
@hzxa21 @chenzl25  PTAL❤️ 
Users are waiting for this fix.

hzxa21 (approved):
Generally LGTM. @chenzl25 PTAL

src/connector/src/source/iceberg/parquet_file_handler.rs
```diff
@@ -265,7 +275,13 @@ pub fn extract_valid_column_indices(
             })
     })
     .collect();
-    Ok(valid_column_indices)
+            Ok(ProjectionMask::leaves(
```
chenzl25, line 278: ProjectionMask has `ProjectionMask::roots` method. Is it what you want? In our forked iceberg-rust, we use this method to handle nested types. https://github.com/risingwavelabs/iceberg-rust/blob/683fb89edeaf8d1baae69e1f376d68b92be1d496/crates/iceberg/src/arrow/reader.rs#L387

wcy-fdu: Yes, this is what I wanted. Thanks for letting me know. I have modified it.

chenzl25 (approved):
LGTM

PR #20158, opened 2025-01-14
refactor: optimize delta log deletion

zwang28:

Since compatibility with the etcd meta store is no longer necessary, delta log deletion can now be performed with a single SQL query.

## Checklist

## Documentation

<details>

</details>


src/meta/src/hummock/manager/gc.rs
```diff
@@ -175,44 +173,36 @@ impl GcManager {
...
-        let deltas_to_delete_count = versioning
-            .hummock_version_deltas
-            .range(..=versioning.checkpoint.version.id)
-            .count();
         // If there is any safe point, skip this to ensure meta backup has required delta logs to
         // replay version.
         if !context_info.version_safe_points.is_empty() {
-            return Ok((0, deltas_to_delete_count));
+            return Ok(0);
         }
```
Li0k, line 187: An implicit problem may cause the request to time out if the number of bulk deletions is too large.

Optional Handling Means
- Stop the service
- Log in to pg and actively clean it up.

Li0k (approved):
Rest LGTM, thanks for the PR

src/meta/src/hummock/manager/gc.rs
Li0k: just for the record, as we think its probability is low.

Issue #19631, opened 2024-12-02
support crossdb query


st1page:
- support ad-hoc query  
   ```
       SELECT * FROM db1.schema.t join db2.schema.t; 
   ``` 
   after we will support the frontend part, which mainly is in binder and 
- support steaming
   ```
       CREATE materialized view mv as SELECT * FROM schema.t join db.schema.t; 
   ``` 
   The db.schema.t's changes need to be read from the log store and achieve isolation

wenym1:
> support ad-hoc query
    SELECT * FROM db1.schema.t join db2.schema.t; 
after we will support the frontend part, which mainly is in binder and

What will be the expected epoch to be used in cross database ad-hoc query🤔 

Since we won't ensure consistency between databases, I guess the intuitive one is that each database uses its own latest epoch. But currently we assume that all tables in a batch query shared a same batch query epoch. So this may require some refactors in the frontend batch query execution, to support different epoch for different tables.

wenym1:
> support steaming
    CREATE materialized view mv as SELECT * FROM schema.t join db.schema.t; 
The db.schema.t's changes need to be read from the log store and achieve isolation

Short note about the implementation of crossdb query.

Since we want to ensure failure isolation between databases, there won't be streaming dispatcher in the cross database stream table scan. In the initial version of cross database stream table scan, to read upstream data, we will only read the L0 log store of the upstream table. 

The behavior is somehow like treating the upstream table as a source. Such source has different vnode as splits. The split offset is our consume progress in different vnodes. During backfill, we will consume the snapshot of the upstream table in a fixed epoch. In this stage, the consume offset under each vnode will be the pk of upstream table. After backfill, we will read the L0 log store of upstream table. In this stage, the consume offset under each vnode will be epoch + pk of upstream table. In general, we will use a separate state table to track the store the progress. The state table is similar to the one we use in recoverable backfill, with an extra epoch column, The schema of the state table will be 
```
| vnode | epoch | is_finished | .. upstream pk|
```

Different to the current snapshot backfill, epoch in upstream table is only treated as an offset of consuming upstream table. We will receive barrier with independent epochs from the barrier manager. We may get the next upstream epoch to be consumed by an extra method call `next_epoch` in the `StateStoreRead` trait.

PR #20167, opened 2025-01-15
feat(snapshot-backfill): implement executor to consume upstream table

wenym1:

This PR is for supporting the cross db streaming query described in https://github.com/risingwavelabs/risingwave/issues/19631.

In this PR, we implement a `ConsumeUpstreamStream` to consume the upstream table epoch by epoch and yield `StreamChunk`. Inside `ConsumeUpstreamStream`, we use `VnodeStream` implemented in https://github.com/risingwavelabs/risingwave/pull/19936 to consume each vnode, and with it we are able to get the latest progress of each vnode at anytime. After it finishes one epoch, it will call `next_epoch` the next epoch and create a new `VnodeStream` to consume the new epoch. A `UpstreamTableExecutor` will poll the `ConsumeUpstreamStream` and the `barrier_rx` concurrently. We use the backfill progress state introduced in https://github.com/risingwavelabs/risingwave/pull/19720 to track the progress state.  When receiving a new barrier, the executor will inspect the latest progress of `ConsumeUpstreamStream` and write progress state, and then yield the barrier, and then continue consuming the `ConsumeUpstreamStream`. On update vnode bitmap, we will recreate the `ConsumeUpstreamStream` for the new vnode bitmap. The logic of the executor is like the following:

```
let mut stream = ConsumeUpstreamStream::new(
    ...
);
loop {
    select! {
        chunk = stream.next() => {
            yield Message::Chunk(chunk);
        }
        barrier = receive_next_barrier(&mut self.barrier_rx) => {
            for progress in stream.progress() {
                progress_state.write(progress);
            }
            progress.flush();
            yield Message::Barrier(barrier);
            if let Some(new_vnode_bitmap) {
                // recreate stream on new vnode bitmap
                stream = ConsumeUpstreamStream::new(
                    new_vnode_bitmap
                    ...
                )
            }
        }
    }
    
}
```

Ideally, the `ConsumeUpstreamStream` should be implemented with the `try_stream` macro, so that the state machine of async execution can be generated automatically by the rust compiler. However, we need to be able to access the progress of the stream at the time we receive barrier, but the stream generated by `try_stream` macro will take the ownership of the internal `VnodeStream`, and then we won't be able to get the latest progress of the ongoing `VnodeStream`. Therefore, in this PR, we will implement the state machine by ourselves. The state machine is like the following:
```
enum ConsumeUpstreamStreamState<'a, T> {
    CreatingSnapshotStream {
        future: UpstreamTableSnapshotStreamFuture<'a, T>,
    },
    ConsumingSnapshotStream {
        stream: UpstreamTableSnapshotStream<T>,
    },
    CreatingChangeLogStream {
        future: UpstreamTableChangeLogStreamFuture<'a, T>,
    },
    ConsumingChangeLogStream {
        stream: UpstreamTableChangeLogStream<T>,
    },
    ResolvingNextEpoch {
        future: NextEpochFuture<'a>,
    },
}
```

The states represent the await point in the following code
```
let snapshot_stream = create_snapshot_stream(snapshot_epoch).await; // CreatingSnapshotStream
while let Some(chunk) = snapshot_stream.next().await { // ConsumingSnapshotStream
    yield Message::Chunk(chunk);
}
let mut epoch = snapshot_epoch;
loop {
    epoch = next_epoch(epoch).await;  // ResolvingNextEpoch
    let change_log_stream = create_change_log_stream(epoch).await; // CreatingChangeLogStream
    while let Some(chunk) = change_log_stream.next().await { // ConsumingChangeLogStream
        yield Message::Chunk(chunk);
    }
}
```

Note that in each state, we need to store the progress of all vnodes owned by the executor even if the progress is made in the previous epoch, so that the previous progress won't be lost.

To support better future extension, the input of `ConsumeUpstreamStream` can be any type that implements the following `UpstreamTable` trait.
```
pub trait UpstreamTable: Send + Sync + 'static {
    type SnapshotStream: Stream<Item = StreamExecutorResult<OwnedRow>> + Send + 'static;
    type ChangeLogStream: Stream<Item = StreamExecutorResult<ChangeLogRow>> + Send + 'static;

    fn next_epoch(&self, epoch: u64)
        -> impl Future<Output = StreamExecutorResult<u64>> + Send + '_;
    fn snapshot_stream(
        &self,
        vnode: VirtualNode,
        epoch: u64,
        start_pk: Option<OwnedRow>,
    ) -> impl Future<Output = StreamExecutorResult<Self::SnapshotStream>> + Send + '_;
    fn change_log_stream(
        &self,
        vnode: VirtualNode,
        epoch: u64,
        start_pk: Option<OwnedRow>,
    ) -> impl Future<Output = StreamExecutorResult<Self::ChangeLogStream>> + Send + '_;

    ...
}
```
In this way, we can easily reuse the logic for different ways to consume upstream table, such as consuming the subscription of other RisingWave cluster, as long as we implement this trait for it.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/executor/backfill/snapshot_backfill/consume_upstream/executor.rs
+++ b/src/stream/src/executor/backfill/snapshot_backfill/consume_upstream/executor.rs
@@ -1,0 +1,220 @@
+// Copyright 2025 RisingWave Labs
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::future::ready;
+
+use anyhow::anyhow;
+use futures::future::{select, Either};
+use futures::{FutureExt, TryStreamExt};
+use futures_async_stream::try_stream;
+use risingwave_common::catalog::TableId;
+use risingwave_common_rate_limit::{
+    MonitoredRateLimiter, RateLimit, RateLimiter, RateLimiterTrait,
+};
+use risingwave_storage::StateStore;
+use rw_futures_util::drop_either_future;
+use tokio::sync::mpsc::UnboundedReceiver;
+
+use crate::executor::backfill::snapshot_backfill::consume_upstream::stream::ConsumeUpstreamStream;
+use crate::executor::backfill::snapshot_backfill::consume_upstream::upstream_table_trait::UpstreamTable;
+use crate::executor::backfill::snapshot_backfill::receive_next_barrier;
+use crate::executor::backfill::snapshot_backfill::state::{BackfillState, EpochBackfillProgress};
+use crate::executor::backfill::utils::mapping_message;
+use crate::executor::prelude::{StateTable, *};
+use crate::executor::{Barrier, Message, StreamExecutorError};
+use crate::task::CreateMviewProgressReporter;
+
+pub struct UpstreamTableExecutor<T: UpstreamTable, S: StateStore> {
+    upstream_table: T,
+    progress_state_table: StateTable<S>,
+    snapshot_epoch: u64,
+    output_indices: Vec<usize>,
+
+    chunk_size: usize,
+    rate_limiter: MonitoredRateLimiter,
+    actor_ctx: ActorContextRef,
+    barrier_rx: UnboundedReceiver<Barrier>,
+    progress: CreateMviewProgressReporter,
+}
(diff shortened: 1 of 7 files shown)


wenym1:
* **#20210** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20210?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#20167** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20167?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20167?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

kwannoel:
Any description?

wenym1:
> Any description?

Added the PR description. @kwannoel @hzxa21 @yezizp2012 @st1page  PTAL

hzxa21 (approved):
LGTM

st1page (approved):

PR #20153, opened 2025-01-14
feat(storage): refactor and support local state store flushed snapshot reader

wenym1:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## What's changed and what's your intention?

Previously in the sink log store, for log reader, we use methods in `StateStore` to access the data flushed by log writer. Methods in `StateStore` will read all shards (aka HummockReadVersion) of the `table_id` on the same CN. However, since the log reader is always in the same SinkExecutor as  log writer, it actually only needs to read the shard that belongs to the writer, and using `StateStore` will unnecessarily read more unrelated data.

In this PR, we will add a new method `new_flushed_snapshot_reader` to `LocalStateStore`, whose return type is a new associated type `FlushedSnapshotReader` that should implement the `StateStoreRead` trait. 

```
pub trait LocalStateStore: StaticSendSync {
    type FlushedSnapshotReader: StateStoreRead + Clone;
   ...

    fn new_flushed_snapshot_reader(&self) -> Self::FlushedSnapshotReader;
}
```

The semantic of this new `FlushedSnapshotReader` is, it can read the uncommitted data written by the parent `LocalStateStore` as long as the data has been `flushe`ed by the parent `LocalStateStore`. Such `FlushedSnapshotReader` will only read the same shard as the local state store.

The KvLogStoreReader will change to use this `FlushedSnapshotReader` to read to avoid reading unnecessary data in other shards in the CN.

## Checklist

- [ ] I have written necessary rustdoc comments.
- [ ]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  My PR contains critical fixes that are necessary to be merged into the latest release. 

## Documentation

- [ ]  My PR needs documentation updates. 

<details>
<summary><b>Release note</b></summary>

</details>

PR #20172, opened 2025-01-15
refactor(storage): create new read snapshot to read for StateStore trait

wenym1:

Previously we have two traits to read in storage, `LocalStateStore` and `StateStoreRead`. One difference in the two traits is that, for `StateStoreRead`, we have to specify the epoch to read on, while for `LocalStateStore`, since its read methods always read from the latest snapshot, we don't need to specify the epoch.

In https://github.com/risingwavelabs/risingwave/pull/20153, we introduce a new associated type `FlushedSnapshotReader`, which also reads data from the latest snapshot. However, since it implements the `StateStoreRead` trait, when we call its read methods, we still need to specify the epoch, which is actually unnecessary. Besides, it will be cleaner if we are able to make the read methods more unified in the trait definition.

Therefore, in this PR, we will remove the `epoch` parameter from the read methods of `StateStoreRead` trait. Instead, we add a new method `new_read_snapshot` to `StateStore`, whose return type is a new associated type `ReadSnapshot` that implements `StateStoreRead`. In the `new_read_snapshot` method, we specify the `epoch` and `table_id` to read. In the implementation of `HummockStorage` on `new_read_snapshot`, we also do `try_wait_epoch` on the epoch, so that the returned `ReadSnapshot` is ready to read.

Changes in the trait definition is as followed:
```
pub trait StateStoreRead: StaticSendSync {
    fn get(
        &self,
        key: TableKey<Bytes>,
        //  epoch: u64, // epoch is removed
        ...

    fn iter(
        &self,
        key_range: TableKeyRange,
        //  epoch: u64, // epoch is removed
        ...

    fn rev_iter(
        &self,
        key_range: TableKeyRange,
        //  epoch: u64, // epoch is removed
        ...
}

pub trait StateStore: StateStoreReadLog + StaticSendSync + Clone {
    type ReadSnapshot: StateStoreRead + Clone;

    fn new_read_snapshot(
        &self,
        epoch: HummockReadEpoch,
        options: NewReadSnapshotOptions,
    ) -> impl StorageFuture<'_, Self::ReadSnapshot>;
    ...
}
```

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/storage/src/hummock/store/local_hummock_storage.rs
+++ b/src/storage/src/hummock/store/local_hummock_storage.rs
@@ -22,7 +22,7 @@ use bytes::Bytes;
 use risingwave_common::bitmap::Bitmap;
 use risingwave_common::catalog::{TableId, TableOption};
 use risingwave_common::hash::VirtualNode;
-use risingwave_common::util::epoch::{EpochPair, MAX_SPILL_TIMES};
+use risingwave_common::util::epoch::{EpochPair, MAX_EPOCH, MAX_SPILL_TIMES};
 use risingwave_hummock_sdk::EpochWithGap;
 use risingwave_hummock_sdk::key::{TableKey, TableKeyRange, is_empty_key_range, vnode_range};
 use risingwave_hummock_sdk::sstable_info::SstableInfo;
@@ -106,7 +106,6 @@ impl LocalHummockFlushedSnapshotReader {
         hummock_version_reader: &HummockVersionReader,
         read_version: &HummockReadVersionRef,
         table_key: TableKey<Bytes>,
-        epoch: u64,
         read_options: ReadOptions,
     ) -> StorageResult<Option<StateStoreKeyedRow>> {
         let table_key_range = (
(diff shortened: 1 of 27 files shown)


wenym1:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20172?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#20172** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20172?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20172?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#20153** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20153?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/storage/src/hummock/store/local_hummock_storage.rs
```diff
@@ -106,34 +106,36 @@ impl LocalHummockFlushedSnapshotReader {
...
+            read_version,
+        )?;
 
         if is_empty_key_range(&table_key_range) {
             return Ok(None);
         }
 
         hummock_version_reader
-            .get(table_key, epoch, read_options, read_snapshot)
+            .get(table_key, MAX_EPOCH, read_options, read_snapshot)
```
MrCroxx, line 128: Why the `MAX_EPOCH` is used here and below instead of the `epoch`.

wenym1: The semantic of `LocalHummockFlushedSnapshotReader` is to read the latest uncommitted data, therefore we use `MAX_EPOCH` to specify reading the latest data.

Issue #19321, opened 2024-11-08
Discussion: Use ORDER BY as part of MV's distribution key


xxchan:
context: https://risingwave-labs.slack.com/archives/C034TRPKN1F/p1731041466238309?thread_ts=1731032250.800469&cid=C034TRPKN1F

- related https://github.com/risingwavelabs/risingwave/issues/12824
    + Also mentioned:  "would like to introduce a `distribued by <column_list>` syntax as well, allowing users to specify the distribution column manually"
- partly related with https://github.com/risingwavelabs/risingwave/issues/19298

Some nodes:
- `dist_key` just need to be subset of `pk`, not prefix.
- MV `pk` is `order_by | stream_key`
- before #13022, MV `dist_key` is `upstream_dist_key` (subset of `stream_key`). And we changed to whole `stream_key`.
- For Index, we support [`distributed by`](https://docs.risingwave.com/sql/commands/sql-create-index#parameters). Since index is essentially the same as MV, there seem to be no reason not to also support it for MV

hzxa21:
An example to show what the dist key looks like now on main:
```sql
create table test_dist (v1 int, v2 int, v3 int, primary key (v1, v2));
create table test_dist2 (v1 int, v2 int, v3 int, primary key (v1, v2));
create table test_dist3 (v1 int, v2 int, v3 int, primary key (v1, v2));


describe test_dist;
-- primary key v1, v2
-- distribution key v1, v2

create materialized view test_dist_mv as select * from test_dist order by v1, v3;

describe test_dist_mv;
-- primary key v1, v3, v2
-- distribution key v1, v2

drop MATERIALIZED view test_join_dist;

create MATERIALIZED view test_join_dist as
select t1.v3 as a, t2.v3 as b, t3.v3 as c
from test_dist as t1
join test_dist2 as t2
on t1.v1 = t2.v1
join test_dist3 as t3
on t2.v1 = t3.v1;

describe test_join_dist;
-- primary key test_dist.v1, test_dist.v2, test_dist2.v2, test_dist3.v2, test_dist2.v1
-- distribution test_dist.v1, test_dist.v2, test_dist2.v2, test_dist3.v2, test_dist2.v1


create MATERIALIZED view test_join_dist2 as
select t1.v3 as a, t2.v3 as b, t3.v3 as c
from test_dist as t1
join test_dist2 as t2
on t1.v1 = t2.v1
join test_dist3 as t3
on t2.v1 = t3.v1
[...]

PR #20176, opened 2025-01-15
feat: use order key as mv's dist key

st1page:

close https://github.com/risingwavelabs/risingwave/issues/19321

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/planner_test/tests/testdata/output/append_only.yaml
+++ b/src/frontend/planner_test/tests/testdata/output/append_only.yaml
@@ -25,9 +25,10 @@
     select v1 from t1 order by v1 limit 3 offset 3;
   stream_plan: |-
     StreamMaterialize { columns: [v1, t1._row_id(hidden)], stream_key: [t1._row_id], pk_columns: [v1, t1._row_id], pk_conflict: NoCheck }
-    └─StreamTopN [append_only] { order: [t1.v1 ASC], limit: 3, offset: 3 }
-      └─StreamExchange { dist: Single }
-        └─StreamTableScan { table: t1, columns: [t1.v1, t1._row_id], stream_scan_type: ArrangementBackfill, stream_key: [t1._row_id], pk: [_row_id], dist: UpstreamHashShard(t1._row_id) }
+    └─StreamExchange { dist: HashShard(t1.v1) }
+      └─StreamTopN [append_only] { order: [t1.v1 ASC], limit: 3, offset: 3 }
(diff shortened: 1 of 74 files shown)


src/frontend/planner_test/tests/testdata/output/append_only.yaml
```diff
@@ -25,9 +25,10 @@
     select v1 from t1 order by v1 limit 3 offset 3;
   stream_plan: |-
     StreamMaterialize { columns: [v1, t1._row_id(hidden)], stream_key: [t1._row_id], pk_columns: [v1, t1._row_id], pk_conflict: NoCheck }
-    └─StreamTopN [append_only] { order: [t1.v1 ASC], limit: 3, offset: 3 }
-      └─StreamExchange { dist: Single }
-        └─StreamTableScan { table: t1, columns: [t1.v1, t1._row_id], stream_scan_type: ArrangementBackfill, stream_key: [t1._row_id], pk: [_row_id], dist: UpstreamHashShard(t1._row_id) }
+    └─StreamExchange { dist: HashShard(t1.v1) }
+      └─StreamTopN [append_only] { order: [t1.v1 ASC], limit: 3, offset: 3 }
+        └─StreamExchange { dist: Single }
+          └─StreamTableScan { table: t1, columns: [t1.v1, t1._row_id], stream_scan_type: ArrangementBackfill, stream_key: [t1._row_id], pk: [_row_id], dist: UpstreamHashShard(t1._row_id) }
```
st1page, line 31: I do not know if we should make some special behavior for the `order by` clause with `limit`... Maybe it is ok because there are not many people write a ToN streaming job and has expectation on its performance

chenzl25 (approved):
LGTM, thanks~

src/frontend/planner_test/tests/testdata/output/append_only.yaml
chenzl25: +1

BugenZhao (commented):
Could you show that in which cases we will gain performance benefits from this change?

st1page:
> Could you show that in which cases we will gain performance benefits from this change?

I think the main discussion is in the issue and hard to give a overall summary.

BugenZhao:
> > Could you show that in which cases we will gain performance benefits from this change?
> 
> I think the main discussion is in the issue and hard to give a overall summary.

Is there at least any simple example? 🤣

PR #20180, opened 2025-01-16
fix(telemetry): not send report if cluster is in test env

tabVersion:

per request from @neverchanje , suspect many UNSPECIFIED deployment type are leaked from test env.

## Checklist

## Documentation

<details>

</details>


tabVersion:
> You can consider marking the cluster with ReportBase_IsTest = true. The approach in this PR can work too. You may choose whichever approach that requires least efforts.

Yes, it also make sense. I will check how I can broadcast the status from meta to other nodes.

src/meta/src/telemetry.rs
```diff
@@ -156,6 +156,10 @@ impl MetaTelemetryInfoFetcher {
 #[async_trait::async_trait]
 impl TelemetryInfoFetcher for MetaTelemetryInfoFetcher {
     async fn fetch_telemetry_info(&self) -> TelemetryResult<Option<String>> {
+        // the err here means building cluster on test env, so we don't need to report telemetry
```
tabVersion, line 159: other nodes get TrackingID from this func, via meta rpc interface. If we do not offer a TrackingID here, the report on other nodes will be disabled.

tabVersion:
tested locally with 

```
  set_env ENABLE_TELEMETRY "true"
  set_env RW_TELEMETRY_TYPE "test"
```

image

The telemetry service is disabled.

PR #17605, opened 2024-07-08
feat(source): Avro with AWS Glue Schema Registry

xiangjinwu:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://gist.github.com/TennyZhuang/f00be7f16996ea48effb049aa7be4d66#file-rw_cla).

## What's changed and what's your intention?

Before this PR, there is a boolean `use_schema_registry` that differentiates usage of `schema.location` vs (confluent) `schema.registry`.

In this PR, we introduce AWS Glue schema registry with `aws.glue.schema_arn` option, and the boolean is deprecated in favor of `enum`s.
* Hint
  * Review with `hide whitespace` because `if-else` are updated to `match`.
  * The 2 largest change `glue.slt` and majority of `glue_resolver.rs` are for tests.
* In `catalog.proto`, `StreamSourceInfo` still has `use_schema_registry` for backward compatibility.
* In `src/frontend/src/handler/create_source.rs` and `src/connector/src/parser/mod.rs`, the boolean has been replaced by `enum SchemaLocation`.
* In `src/connector/src/parser/avro/parser.rs`, the `Option<Arc<ConfluentSchemaCache>>` has been replaced by `enum WriterSchemaCache`.

## Checklist

- [x] I have written necessary rustdoc comments
- [x] I have added necessary unit tests and integration tests
- [ ] I have added test labels as necessary. See [details](https://github.com/risingwavelabs/risingwave/blob/main/docs/developer-guide.md#ci-labels-guide).
- [ ] I have added fuzzing tests or opened an issue to track them. (Optional, recommended for new SQL features #7934).
- [ ] My PR contains breaking changes. (If it deprecates some features, please create a tracking issue to remove them in the future).
- [x] All checks passed in `./risedev check` (or alias, `./risedev c`)
- [ ] My PR changes performance-critical code. (Please run macro/micro-benchmarks and show the results.)

- [ ] My PR contains critical fixes that are necessary to be merged into the latest release. (Please check out the [details](https://github.com/risingwavelabs/risingwave/blob/main/CONTRIBUTING.md))

## Documentation

- [x] My PR needs documentation updates. (Please use the **Release note** section below to summarize the impact on users)

## Release note

Added support for AWS Glue Schema Registry when using `encode avro` for sources:

```
encode avro (
  aws.glue.schema_arn = ''arn:aws:glue:ap-southeast-1:123456123456:schema/default-registry/MyEvent',
  // other aws auth options same as s3 / kinesis
[...]

PR #20181, opened 2025-01-16
feat(sink): Avro with AWS Glue Schema Registry

xiangjinwu:

Supports loading schema from and encoding metadata for AWS Glue Schema Registry, in addition to Confluent Schema Registry, when using `encode avro`.

Notes:
* Glue Schema Registry has no references (eg `import` in protobuf).
* Glue Schema Registry has no name strategy. So it would require an additional `aws.glue.key_schema_arn`. This is left unimplemented in favor of real use case `KEY ENCODE [TEXT | BYTES]`.

Related: #17605 for source

Refactor to come next: {Confluent, Glue} x {Source, Sink} x {Avro, Protobuf}

## Checklist

## Documentation

Added support for AWS Glue Schema Registry when using `encode avro` for sinks:

```
encode avro (
  aws.glue.schema_arn = 'arn:aws:glue:ap-southeast-1:123456123456:schema/default-registry/MyEvent',
  // other aws auth options same as s3 / kinesis
  aws.region = '...',
  aws.credentials.access_key_id = '...',
  ...
) [KEY ENCODE [TEXT | BYTES]] -- when specifying `primary_key` as kafka key
```

When using IAM permission policies, the following `action`s shall be allowed:
* `glue:GetSchemaVersion`

### DIFF
--- a/e2e_test/source_inline/kafka/avro/glue.slt.serial
+++ b/e2e_test/source_inline/kafka/avro/glue.slt.serial
@@ -145,6 +145,76 @@ foo 2006-01-02 22:04:05.123456+00:00 NULL
 AB  2022-04-08 00:00:00.123456+00:00 \xdeadbeef
 
 statement ok
+create sink sk as select
+  'bar' as f1,
+  to_timestamp(1735689600) as f2,
+  'a0A'::bytea as f3
+with (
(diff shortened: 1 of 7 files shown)


xxchan (approved):
Generally LGTM

PR #20194, opened 2025-01-17
refactor(storage): Reduce get_table_id_tables lock calls.

Li0k:

This pull request introduce a new field "table_id_to_table_options" to reduce get_table_id_tables lock calls. 

Since table_id_to_table_options are allowed to not update in real time, we don't have to refetch them from the sql backend every time we get_compact_task. Accessing the catalog_controller requires an async lock, which may cause performance degradation if accessed frequently. Introducing a temporary cache, and reducing its update frequency, is expected to alleviate these problems.

## Checklist

## Documentation

<details>

</details>


zwang28 (approved):
LGTM

PR #20201, opened 2025-01-17
fix: fix several issues about the public schema

yezizp2012:

1. ensure the commands "show objects" and "show create object" function correctly within `search_path`, rather than use `public` schema by default.
2. make the `public` schema also managed by grant and revoke commands, which means that when creating a new non-super user, it is necessary to explicitly grant them the privilege of the `public` schema, which is also the default behavior of pg.

## Checklist

## Documentation

<details>

</details>


chenzl25 (approved):
LGTM

PR #20131, opened 2025-01-13
feat(frontend): use purified definition for replacing table

BugenZhao:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## What's changed and what's your intention?

Adopt purified definition for table replacement like schema change and sink-into-table.

The main change is to introduce a new enum named `SqlColumnStrategy`, defining the behavior when the schema is defined in the SQL and derived from external system simultaneously. This was always rejected in the past. However, with the purified definition used for table replacement, we need to handle it with different strategies.

1. For `REFERSH`, use the newly resolved schema.
2. For `SINK INTO TABLE`, ensure schema unchanged by using the defined schema after purification.
3. For `[ADD | DROP] COLUMN`, use the altered defined schema.
4. For `CREATE` with schema registry, reject defined schema.

Case 2 actually fixes a long-standing issue caused by impurity. Previously, creating a sink into a table will accidentally refresh its schema.

Cases 3 and 4 adhere to the current behavior. However, we can now modify this behavior simply by passing a different `SqlColumnStrategy`, which will accomplish the objective in #12199.

## Checklist

- [x] I have written necessary rustdoc comments.
- [ ]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  My PR contains critical fixes that are necessary to be merged into the latest release. 

## Documentation

- [ ]  My PR needs documentation updates. 

<details>
<summary><b>Release note</b></summary>

</details>

PR #20174, opened 2025-01-15
feat(frontend): allow schema change & sink into a CTAS table

BugenZhao:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## What's changed and what's your intention?

Allow performing schema changes (`ALTER TABLE ... [ADD | DROP] COLUMN`) on or create a sink into a table created by `CREATE TABLE AS` statement.

----

After #19949, we now have two different methods for retrieving the definition of a table: `create_sql` and `create_sql_purified`. Prior to this PR, `create_sql` always directly return the persisted definition string, while `create_sql_purified` will re-assemble a definition with all columns qualified based on the `TableCatalog`.

However, the definitions for CTAS tables are missing, causing `create_sql_ast` to always return an error. This seems pointless. So this PR makes `create_sql_ast` also run purification if (and only if) the definition is missing (indicating that it's a CTAS table). This enables replacing functionalities for CTAS table, including schema change and sink-into-table.

## Checklist

- [x] I have written necessary rustdoc comments.
- [x]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  My PR contains critical fixes that are necessary to be merged into the latest release. 

## Documentation

- [ ]  My PR needs documentation updates. 

<details>
<summary><b>Release note</b></summary>

</details>

PR #20203, opened 2025-01-17
feat(frontend): allow specifying (partial) schema when creating table with derived schema

### LINKED ISSUE #12199: feat: specifying table schema when creating source/table from avro/protobuf encode whose full schema is externally defined
Currently, we do not allow user to define schema in column clause
```SQL
CREATE SOURCE(a int, b int) FORMAT PLAIN ENCODE PROTOBUF;
 ERROR: ExecuteError: Protocol error: User-defined schema is not allowed with FORMAT PLAIN ENCODE PROTOBUF
```
But some user needs it to prune columns, especially when they are creating a table with the connector, the selected columns determine how many columns will be materialized in storage. Also, casting the source data into the expected datatype is needed too.

<del>Another issue is that if the user want to define a generated column, he must specify columns in the create source/table statement. We might need to introduce another syntax https://github.com/risingwavelabs/risingwave/issues/12209 </del> fixed

- [ ] avro
- [ ] protobuf
- [ ] json with schema registry

BugenZhao:

This is a follow-up of #20131:

> we can now modify this behavior simply by passing a different `SqlColumnStrategy`, which will accomplish the objective in https://github.com/risingwavelabs/risingwave/issues/12199.

This PR removes the `Reject` strategy and adds a `FollowChecked` instead, which means that we will follow the columns specified in SQL with schema check against the external system.

As a result, users can now specify the (maybe partial) schema when creating a table with a derived schema (i.e., using a schema registry). Consequently, `REFRESH SCHEMA` and `[ADD | DROP] COLUMN` can be called on any table, regardless of whether its schema was initially derived or manually specified.

Also affect `CREATE SOURCE`. For `ALTER SOURCE`, as its code path now differs from `ALTER TABLE`, the behavior keeps unchanged. Will check it in future PRs.

Close #12199.

## Checklist

## Documentation

<details>

When creating a table with schema registry (Avro, Protobuf, JSON with schema registry), users are now allowed to specify the columns they need to ingest in the column definition list, which will be checked against the external schema. Previously, we only allow wildcard `*` here to ingest all columns.

Manually manipulating the column set by calling `ALTER TABLE [ADD | DROP] COLUMN` on a table with schema registry is also supported now. Previously we only support `REFRESH SCHEMA` to update the full schema.

</details>

### DIFF
--- a/e2e_test/source_inline/kafka/avro/partial_schema.slt
+++ b/e2e_test/source_inline/kafka/avro/partial_schema.slt
@@ -35,7 +35,19 @@ FORMAT PLAIN ENCODE AVRO (
     schema.registry = '${RISEDEV_SCHEMA_REGISTRY_URL}'
 );
 
-# Specify incorrect schema
+# Specify partial schema, with a generated column having the same name as a column in the source schema.
+# We should treat the generated column as a new column.
+statement ok
+create table t3 (bar int, foo varchar as 'generated')
+WITH (
+    ${RISEDEV_KAFKA_WITH_OPTIONS_COMMON},
+    topic = 'avro_partial_schema_test'
+)
+FORMAT PLAIN ENCODE AVRO (
+    schema.registry = '${RISEDEV_SCHEMA_REGISTRY_URL}'
+);
+
+# Specify incorrect schema: inexistent column
 statement error
 create table t (bar int, foo varchar, baz int)
 WITH (
@@ -52,6 +64,7 @@ Caused by:
   Protocol error: Column "baz" is defined in SQL but not found in the source
 
 
+# Specify incorrect schema: data type mismatch
 statement error
 create table t (bar double)
 WITH (
@@ -95,8 +108,9 @@ FORMAT PLAIN ENCODE AVRO (
 query TT rowsort
 SELECT name, SUBSTRING(definition, 1, POSITION(' WITH' IN definition) - 1) FROM rw_tables WHERE name LIKE 't%';
 ----
-t1	    CREATE TABLE t1 (foo CHARACTER VARYING, bar INT)
-t2	    CREATE TABLE t2 (bar INT, gen_col INT AS bar + 1)
+t1	CREATE TABLE t1 (foo CHARACTER VARYING, bar INT)
+t2	CREATE TABLE t2 (bar INT, gen_col INT AS bar + 1)
+t3	CREATE TABLE t3 (bar INT, foo CHARACTER VARYING AS 'generated')
 tstar	CREATE TABLE tstar (bar INT, foo CHARACTER VARYING, gen_col INT AS bar + 1)
 tstar2	CREATE TABLE tstar2 (bar INT, foo CHARACTER VARYING, gen_col INT AS bar + 1)
 
@@ -131,6 +145,18 @@ Caused by:
 
 
 statement ok
+alter table t3 add column baz double;
+
+statement error
+alter table t3 add column foo varchar;
+----
+db error: ERROR: Failed to run the query
+
+Caused by:
+  Invalid input syntax: column "foo" of table "t3" already exists
+
+
+statement ok
 alter table tstar drop column foo;
 
 statement error
@@ -158,8 +184,9 @@ Caused by:
 query TT rowsort
 SELECT name, SUBSTRING(definition, 1, POSITION(' WITH' IN definition) - 1) FROM rw_tables WHERE name LIKE 't%';
 ----
-t1	    CREATE TABLE t1 (bar INT)
-t2	    CREATE TABLE t2 (bar INT, gen_col INT AS bar + 1, baz DOUBLE)
+t1	CREATE TABLE t1 (bar INT)
+t2	CREATE TABLE t2 (bar INT, gen_col INT AS bar + 1, baz DOUBLE)
+t3	CREATE TABLE t3 (bar INT, foo CHARACTER VARYING AS 'generated', baz DOUBLE)
 tstar	CREATE TABLE tstar (bar INT, gen_col INT AS bar + 1, baz DOUBLE)
 tstar2	CREATE TABLE tstar2 (bar INT, foo CHARACTER VARYING, gen_col INT AS bar + 1)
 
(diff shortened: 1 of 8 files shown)


BugenZhao:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20203?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#20203** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20203?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20203?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#20182** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20182?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#20131** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20131?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#20174** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20174?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

xiangjinwu:
Actually regarding the motivation #12199 @st1page:

This week I was asked twice about the `User-defined schema from SQL is not allowed` error, and in both cases the user did not intend to select a subset of columns - the proper fix is to let it derive.

BugenZhao:
> This week I was asked twice about the `User-defined schema from SQL is not allowed` error, and in both cases the user did not intend to select a subset of columns - the proper fix is to let it derive.

I think this is more like a potential documentation improvement. 🥵

e2e_test/source_inline/kafka/avro/partial_schema.slt
```diff
@@ -0,0 +1,199 @@
...
+statement ok
+alter table t1 drop column foo;
+
+statement error
+alter table t2 add column baz int;
+----
+db error: ERROR: Failed to run the query
+
+Caused by:
+  Protocol error: Data type mismatch for column "baz". Defined in SQL as "integer", but found in the source as "double precision"
```
tabVersion, line 118: does it mean we always keep track of the latest version in the schema registry and prevent possible future name/type collide?

e2e_test/source_inline/kafka/avro/partial_schema.slt
```diff
@@ -0,0 +1,199 @@
...
+SELECT name, SUBSTRING(definition, 1, POSITION(' WITH' IN definition) - 1) FROM rw_tables WHERE name LIKE 't%';
+----
+t1	    CREATE TABLE t1 (foo CHARACTER VARYING, bar INT)
+t2	    CREATE TABLE t2 (bar INT, gen_col INT AS bar + 1)
+tstar	CREATE TABLE tstar (bar INT, foo CHARACTER VARYING, gen_col INT AS bar + 1)
+tstar2	CREATE TABLE tstar2 (bar INT, foo CHARACTER VARYING, gen_col INT AS bar + 1)
+
+# create a new schema
+system ok
+sr_register avro_partial_schema_test-value AVRO <<< '{"type":"record","name":"Root","fields":[{"name":"bar","type":"int","default":0},{"name":"foo","type":"string"}, {"name":"baz", "type":"double", "default":0}]}'
```
tabVersion, line 105: Can you add one more test

1. declare schema (a int, b varchar)
2. create table with generated column (a int, b varchar, c int as a + 1)
3. create a newer version in schema registry (a int, b varchar, c int)
4. what does the purified SQL tend to be at the moment?

e2e_test/source_inline/kafka/avro/partial_schema.slt
BugenZhao: Yes

e2e_test/source_inline/kafka/avro/partial_schema.slt
BugenZhao: Step 3 will not affect the purified SQL because we don't access (then test against) the external schema unless we are going to perform schema change.

However, this is still an interesting example, as I'm not sure about the behavior if we call `REFRESH` or `ADD COLUMN` afterwards. I'm afraid this has long been undefined behavior since we allow the definition of a generated column on a source. Will test it.

e2e_test/source_inline/kafka/avro/partial_schema.slt
```diff
@@ -0,0 +1,199 @@
...
+create table tstar (*, gen_col int as bar + 1)
+WITH (
+    ${RISEDEV_KAFKA_WITH_OPTIONS_COMMON},
+    topic = 'avro_partial_schema_test'
+)
+FORMAT PLAIN ENCODE AVRO (
+    schema.registry = '${RISEDEV_SCHEMA_REGISTRY_URL}'
+);
+
+# No wildcard will be interpreted as `*` for syntax backward compatibility
```
fuyufjh, line 82: The old syntax indeed looks a bit strange. Let's update the docs to reflect the new syntax then.

src/frontend/src/handler/create_source.rs
```diff
@@ -654,25 +699,24 @@ pub fn bind_connector_props(
...
-    /// This is the behavior when `SINK INTO` or `[ADD | DROP] COLUMN` atop the purified SQL,
-    /// ensuring that no accidental refresh will happen and the schema remains unchanged.
-    // TODO(purify): we may validate whether defined columns are valid against the resolved schema.
-    Follow,
+    /// This is the behavior when re-planning the target table of `SINK INTO`.
+    FollowUnchecked,
+
+    /// Follow all column defined in SQL, check the columns from external system with name & type.
+    /// This ensures that no accidental side effect will change the schema.
+    ///
```
fuyufjh, line 709:
```suggestion
    ///
    /// - When wildcard （`*`) is presented, all columns are used, and the user-provided columns will be checked against the actual schema.
    /// - When wildcard is not presented, only columns from user's definition will be used and checked.
    ///
```

fuyufjh (approved):
LGTM

e2e_test/source_inline/kafka/avro/partial_schema.slt
BugenZhao: Covered in 4cce72a149bccbea19aef2412d9b7e1cd2ffa33b

src/frontend/src/handler/create_source.rs
BugenZhao: This is not that true. When `*` is presented, regular (non-generated) columns specified by users will be rejected.

https://github.com/risingwavelabs/risingwave/blob/4cce72a149bccbea19aef2412d9b7e1cd2ffa33b/src/frontend/src/handler/create_source.rs#L239-L245

I will include this in the release notes part of this PR.

e2e_test/source_inline/kafka/avro/partial_schema.slt
BugenZhao: There were some discussions: https://risingwave-labs.slack.com/archives/C03A2PSS8KU/p1737537653835029

It seemed to be a breaking change.

PR #20204, opened 2025-01-17
feat(stream): add synced logstore

kwannoel:

Add synced logstore for unaligned join and decoupled sink into table.

We have 2 futures we repeatedly construct and poll:
1. One to read from upstream.
2. Another to read from logstore (state_store / buffer).

May look at https://github.com/risingwavelabs/risingwave/pull/20204#discussion_r1925037081 for the implementation design.

We can't use the desynced (existing) logstore because:
1. Truncation interface and barrier commit are decoupled. It's designed in such a way because sinks only require at-least-once semantics. So it is fine to read duplicated data after rewind happens. This doesn't apply to join, because downstream will need to ensure it is buffered.
2. The desynced logstore does not yield barrier downstream directly from the writer side. This is not a requirement for external sinks, because they don't need to receive barrier. However for sink into table or join downstream, we need to propagate barrier typically.

TODOs:
- [ ] Handle update vnode bitmap
- [ ] Handle watermark

## Checklist

## Documentation

<details>

</details>


kwannoel:
* **#20204** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20204?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20204?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/stream/src/executor/sync_kv_log_store.rs
```diff
@@ -0,0 +1,1022 @@
...
+//!
+//!    Finally we will pop the earliest item in the buffer.
+//!    - If it's a chunk yield it.
+//!    - If it's a watermark yield it.
+//!    - If it's a flushed chunk reference (`LogStoreBufferItem::Flushed`),
+//!      we will read the corresponding chunks in the log store.
+//!      This is done by constructing a `flushed_chunk_future` which will read the log store
+//!      using the `seq_id`.
+//!   - Barrier,
+//!     because they are directly propagated from the upstream when polling it.
```
kwannoel, line 54: See this docstring for the design.

src/stream/src/executor/sync_kv_log_store.rs
```diff
@@ -0,0 +1,1022 @@
...
+                assert_eq!(barrier.epoch.curr, test_epoch(2));
+            }
+            other => panic!("Expected a barrier message, got {:?}", other),
+        }
+    }
+
+    // When we hit buffer max_chunk, we only store placeholder `FlushedItem`.
+    // So we just let capacity = 0, and we will always flush incoming chunks to state store.
+    #[tokio::test]
+    async fn test_max_chunk_persisted_read() {
```
kwannoel, line 933: This test is still buggy... For some reason serde fails with:
```
thread 'executor::sync_kv_log_store::tests::test_max_chunk_persisted_read' panicked at /Users/noelkwan/.cargo/registry/src/index.crates.io-6f17d22bba15001f/bytes-1.9.0/src/lib.rs:138:5:
advance out of bounds: the len is 4 but advancing by 10
stack backtrace:
   0: rust_begin_unwind
             at /rustc/52fd9983996d9fcfb719749838336be66dee68f9/library/std/src/panicking.rs:665:5
   1: core::panicking::panic_fmt
             at /rustc/52fd9983996d9fcfb719749838336be66dee68f9/library/core/src/panicking.rs:74:14
   2: bytes::panic_advance
             at /Users/noelkwan/.cargo/registry/src/index.crates.io-6f17d22bba15001f/bytes-1.9.0/src/lib.rs:138:5
   3: <&[u8] as bytes::buf::buf_impl::Buf>::copy_to_slice
             at /Users/noelkwan/.cargo/registry/src/index.crates.io-6f17d22bba15001f/bytes-1.9.0/src/buf/buf_impl.rs:1431:13
   4: <&mut T as bytes::buf::buf_impl::Buf>::copy_to_slice
             at /Users/noelkwan/.cargo/registry/src/index.crates.io-6f17d22bba15001f/bytes-1.9.0/src/buf/buf_impl.rs:1260:13
   5: risingwave_common::util::value_encoding::deserialize_str
             at /Users/noelkwan/projects/risingwave/src/common/src/util/value_encoding/mod.rs:390:5
   6: risingwave_common::util::value_encoding::deserialize_value
             at /Users/noelkwan/projects/risingwave/src/common/src/util/value_encoding/mod.rs:343:47
   7: risingwave_common::util::value_encoding::inner_deserialize_datum
             at /Users/noelkwan/projects/risingwave/src/common/src/util/value_encoding/mod.rs:198:19
   8: risingwave_common::util::value_encoding::deserialize_datum
             at /Users/noelkwan/projects/risingwave/src/common/src/util/value_encoding/mod.rs:189:5
   9: risingwave_common::row::owned_row::RowDeserializer<D>::deserialize
             at /Users/noelkwan/projects/risingwave/src/common/src/row/owned_row.rs:169:25
  10: <risingwave_common::util::value_encoding::BasicSerde as risingwave_common::util::value_encoding::ValueRowDeserializer>::deserialize
             at /Users/noelkwan/projects/risingwave/src/common/src/util/value_encoding/mod.rs:117:12
  11: risingwave_stream::common::log_store_impl::kv_log_store::serde::LogStoreRowSerde::deserialize
             at ./src/common/log_store_impl/kv_log_store/serde.rs:350:24
  12: risingwave_stream::common::log_store_impl::kv_log_store::serde::stream_de::deserialize_stream::{{closure}}
             at ./src/common/log_store_impl/kv_log_store/serde.rs:661:39

```

src/stream/src/executor/sync_kv_log_store.rs
xxchan: So what does "synced" mean here? (and what is unsynced logstore?)

kwannoel: unsynced logstore is our current design for the sink logstore, where we have 2 separate streams for reading from and to the logstore.

Synced means that we have a single stream instead, rather than 2 concurrently running ones.

xxchan: I'm not very familiar with our current design for the sink logstore. So why can't we use the "unsynced" version here? What problems do the "synced" version want to solve?

xxchan: btw, why sink into table needs this new logstore, instead of the old one

kwannoel: > I'm not very familiar with our current design for the sink logstore. So why can't we use the "unsynced" version here? What problems do the "synced" version want to solve?

Updated in PR description. Briefly:
1. Truncate the logstore before barrier gets committed, rather than asynchronously. This ensures exactly-once semantics.
2. Yield barriers downstream from the writer side. External sink does not need barriers so the desynced logstore does not have barriers yielded downstream.

src/stream/src/executor/sync_kv_log_store.rs
```diff
@@ -0,0 +1,1027 @@
...
+        let init_epoch_pair = barrier.epoch;
+        local_state_store
+            .init(InitOptions::new(init_epoch_pair))
+            .await?;
+        let state_store_stream = read_persisted_log_store(
+            serde,
+            table_id,
+            metrics,
+            state_store,
+            barrier.epoch.curr, // FIXME(kwannoel): Should this be curr or prev?
```
chenzl25, line 230: It seems to be the prev barrier. It works like backill snapshot read.

chenzl25 (commented):
Generally LGTM. The implementation is similar to the sink logstore, but we need to take care of the watermark truncation which is related to compaction.

src/stream/src/common/log_store_impl/kv_log_store/buffer.rs
```diff
@@ -28,7 +28,7 @@ use crate::common::log_store_impl::kv_log_store::{
 };
 
 #[derive(Clone)]
-pub(crate) enum LogStoreBufferItem {
+pub enum LogStoreBufferItem {
```
wenym1, line 31: The `synced_kv_log_store` is also in the same crate, so not need to extend the visibility.

src/stream/src/common/log_store_impl/kv_log_store/mod.rs
```diff
@@ -32,11 +32,11 @@ use crate::common::log_store_impl::kv_log_store::serde::LogStoreRowSerde;
 use crate::common::log_store_impl::kv_log_store::writer::KvLogStoreWriter;
 use crate::executor::monitor::StreamingMetrics;
 
-mod buffer;
-mod reader;
+pub mod buffer;
```
wenym1, line 35: `pub(crate)` should be enough for the changes in this file.

src/stream/src/common/log_store_impl/kv_log_store/reader.rs
```diff
@@ -672,6 +598,107 @@ impl<S: StateStoreRead + Clone> LogReader for KvLogStoreReader<S> {
     }
 }
 
+#[allow(clippy::too_many_arguments)]
```
wenym1, line 601: Are the `read_flushed_chunk` and `read_persisted_log_store` just a simple copy-paste from the original code without any change in the logic?

src/stream/src/executor/sync_kv_log_store.rs
```diff
@@ -0,0 +1,1027 @@
...
+
+// Stream interface
+impl<S: StateStore, LS: LocalStateStore> SyncedKvLogStoreExecutor<S, LS> {
+    #[try_stream(ok = Message, error = StreamExecutorError)]
+    pub async fn execute_inner(mut self) {
+        let mut input = self.upstream.execute();
+        let Some(msg) = input.next().await else {
+            bail!("Expected a barrier message, got end of stream")
+        };
+        let barrier = match msg? {
```
wenym1, line 175: Use `expect_first_barrier`.

src/stream/src/executor/sync_kv_log_store.rs
```diff
@@ -0,0 +1,1027 @@
...
+            Message::Barrier(barrier) => barrier,
+            other => bail!("Expected a barrier message, got {:?}", other),
+        };
+        let mut state_store_stream = Some(
```
wenym1, line 179: We should yield the barrier first and then init the local state store. Otherwise, we will get stuck at `LocalStateStore::init`.

wenym1 (commented):
Leave some early comments. Will continue review the executor implementation. At first glance, the implementation seems to be too complicated. Besides, it seems that update vnode bitmap is not handled in this PR.

src/stream/src/common/log_store_impl/kv_log_store/reader.rs
kwannoel: Yes. These are refactored and reused in the synced_log_store executor.

src/stream/src/executor/sync_kv_log_store.rs
```diff
@@ -0,0 +1,1027 @@
...
+// Write methods
+impl<S: StateStore, LS: LocalStateStore> SyncedKvLogStoreExecutor<S, LS> {
+    async fn write_barrier(
+        state_store: &mut LS,
+        serde: &LogStoreRowSerde,
+        barrier: Barrier,
+        metrics: &mut KvLogStoreMetrics,
+        truncation_offset: Option<ReaderTruncationOffsetType>,
+        seq_id: &mut SeqIdType,
+        buffer: &mut Mutex<SyncedLogStoreBuffer>,
```
wenym1, line 462: The current code, the `Mutex<SyncedLogStoreBuffer>` always has the mutable reference. Therefore, it's likely that we don't have to wrap the buffer with the `Mutex`.

src/common/src/util/value_encoding/mod.rs
```diff
@@ -385,6 +385,7 @@ fn deserialize_list(item_type: &DataType, data: &mut impl Buf) -> Result<ScalarI
 
 fn deserialize_str(data: &mut impl Buf) -> Result<Box<str>> {
     let len = data.get_u32_le();
+    tracing::trace!("len: {len}");
```
kwannoel, line 388:
```suggestion
```

src/stream/src/executor/sync_kv_log_store.rs
```diff
@@ -0,0 +1,1045 @@
...
+type StateStoreStream<S> = Pin<Box<LogStoreItemMergeStream<TimeoutAutoRebuildIter<S>>>>;
+type ReadFlushedChunkFuture = BoxFuture<'static, LogStoreResult<(ChunkId, StreamChunk, u64)>>;
+
+struct SyncedKvLogStoreExecutor<S: StateStore> {
+    actor_context: ActorContextRef,
+    table_id: TableId,
+    read_metrics: KvLogStoreReadMetrics,
+    metrics: KvLogStoreMetrics,
+    serde: LogStoreRowSerde,
+    seq_id: SeqIdType,
```
wenym1, line 108: Fields like `seq_id`, `truncation_offset`, `flushed_chunk_future` and `buffer` can be removed from the executor struct, and instead declare as variables in `execute_inner`.

src/stream/src/executor/sync_kv_log_store.rs
```diff
@@ -0,0 +1,1045 @@
...
+                &barrier,
+                &mut self.local_state_store,
+                &self.serde,
+                self.table_id,
+                &self.metrics,
+                self.state_store.clone(),
+            )
+            .await?,
+        );
+        loop {
```
wenym1, line 183: We can use multiple levels of nested loops to make the control logic clearer. The code framework can be like the following:
```
'reacreate_consume_stream: loop {
    let persisted_stream = read_persisted();
    let flushed_chunk_future = None;
    let seq_id = 0;
    let truncation_offset = None;
    let buffer = SyncedLogStoreBuffer::new();
    loop {
        let checkpoint_barrier = loop {
            select!{
                ...
                Message::Barrier(barrier) if barrier.is_checkpoint() => {
                    break barrier;
                }
            }
        };
        flush(...);
        yield barrier;
        if barrier.has_upate_vnode_bitmap() {
            [state_store, serde].update_vnode_bitmap();
            continue `reacreate_consume_stream;
        }
    }
}
```

src/stream/src/executor/sync_kv_log_store.rs
```diff
@@ -0,0 +1,1045 @@
...
+    Ok(vnode_bitmap_builder.finish())
+}
+
+impl SyncedLogStoreBuffer {
+    fn add_or_flush_chunk(
+        &mut self,
+        start_seq_id: SeqIdType,
+        end_seq_id: SeqIdType,
+        chunk: StreamChunk,
+        state_store: &mut impl LocalStateStore,
```
wenym1, line 603: Seems that the `state_store` is passed only to call `state_store.epoch()`. We'd better pass the barrier epoch directly.

src/stream/src/executor/sync_kv_log_store.rs
```diff
@@ -0,0 +1,1045 @@
...
+        buffer.next_chunk_id = 0;
+        buffer.update_unconsumed_buffer_metrics();
+
+        // Apply Vnode Update
+        if let Some(vnode_bitmap) = barrier.as_update_vnode_bitmap(actor_id) {
+            state_store.update_vnode_bitmap(vnode_bitmap.clone());
+            serde.update_vnode_bitmap(vnode_bitmap.clone());
+            buffer
+                .buffer
+                .push_back((epoch, LogStoreBufferItem::UpdateVnodes(vnode_bitmap)));
```
wenym1, line 520: We don't need to push this `UpdateVnodes` item. And in other code appearance we shall mark it as unreachable.

src/stream/src/executor/sync_kv_log_store.rs
```diff
@@ -0,0 +1,1045 @@
...
+        let table = gen_test_log_store_table(pk_info);
+
+        let log_store_executor = SyncedKvLogStoreExecutor::new(
+            ActorContext::for_test(123),
+            table.id,
+            KvLogStoreReadMetrics::for_test(),
+            KvLogStoreMetrics::for_test(),
+            LogStoreRowSerde::new(&table, vnodes, pk_info),
+            0,
+            MemoryStateStore::new(),
```
wenym1, line 785: In the future we'd better add test for recovery and update vnode bitmap, and we may use hummock state store for the test then.

wenym1 (approved):
LGTM!

fuyufjh (approved):
On behalf of cargo.lock

PR #20218, opened 2025-01-20
refactor: ensure update vnode bitmap after yield barrier

wenym1:

Part of https://github.com/risingwavelabs/risingwave/issues/18312.

As described in https://github.com/risingwavelabs/risingwave/issues/18312, we should ensure a correct handling order as `state_table.commit()` -> `yield barrier` -> `state_table.update_vnode_bitmap()` in streaming executors while update vnode bitmap. 

In this PR, for `state_table.commit()`, we change to return a `StateTablePostCommit<'_>`, which captures the mutable reference to the original `state_table` and will be used as a callback after yielding the barrier. The `StateTablePostCommit` has a method `async fn post_yield_barrier(mut self, new_vnodes: Option<Arc<Bitmap>>)` to be called after the we yield the barrier. 

The `StateTablePostCommit` is marked with `#[must_use]`. The method name `post_yield_barrier` indicates that it should be called after we have yielded the barrier. In state table, we add a flag `on_post_commit`, to indicate that whether the `StateTablePostCommit` is handled properly. On `state_table.commit()`, we will mark the `on_post_commit` as true, and in `StateTablePostCommit::post_yield_barrier`, we will remark the flag as false, and on `state_table.commit()`, we will assert that the `on_post_commit` must be false. Note that, the `post_yield_barrier` should be called for all barriers rather than only for the barrier with update vnode bitmap. In this way, though we don't have scale test for all streaming executor, we can ensure that all executor covered by normal e2e test have properly handled the `StateTablePostCommit`.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/common/table/state_table.rs
+++ b/src/stream/src/common/table/state_table.rs
@@ -163,6 +163,8 @@ pub struct StateTableInner<
 
     clean_watermark_index_in_pk: Option<i32>,
 
+    /// Flag to indicate whether the state table has called `commit`, but has not called
+    /// `post_yield_barrier` on the `StateTablePostCommit` callback yet.
     on_post_commit: bool,
 }
 
@@ -687,6 +689,20 @@ where
     }
 }
 
+/// A callback struct returned from [`StateTableInner::commit`].
+///
+/// Introduced to support single barrier configuration change proposed in <https://github.com/risingwavelabs/risingwave/issues/18312>.
+/// In brief, to correctly handle the configuration change, when each stateful executor receives an upstream barrier, it should handle
+/// the barrier in the order of `state_table.commit()` -> `yield barrier` -> `update_vnode_bitmap`.
+///
+/// The `StateTablePostCommit` captures the mutable reference of `state_table` when calling `state_table.commit()`, and after the executor
+/// runs `yield barrier`, it should call `StateTablePostCommit::post_yield_barrier` to apply the vnode bitmap update if there is any.
+/// The `StateTablePostCommit` is marked with `must_use`. The method name `post_yield_barrier` indicates that it should be called after
+/// we have yielded the barrier. In `StateTable`, we add a flag `on_post_commit`, to indicate that whether the `StateTablePostCommit` is handled
+/// properly. On `state_table.commit()`, we will mark the `on_post_commit` as true, and in `StateTablePostCommit::post_yield_barrier`, we will
+/// remark the flag as false, and on `state_table.commit()`, we will assert that the `on_post_commit` must be false. Note that, the `post_yield_barrier`
+/// should be called for all barriers rather than only for the barrier with update vnode bitmap. In this way, though we don't have scale test for all
+/// streaming executor, we can ensure that all executor covered by normal e2e test have properly handled the `StateTablePostCommit`.
 #[must_use]
 pub struct StateTablePostCommit<
     'a,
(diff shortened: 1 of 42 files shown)


wenym1 (commented):
Added some comments for code with logic changed and need more careful review.

src/stream/src/common/table/state_table.rs
BugenZhao, line 1: Can you please add documentation for new public interfaces in this file?

src/stream/src/common/table/state_table.rs
```diff
@@ -682,41 +685,94 @@ where
             .await
             .map_err(Into::into)
     }
+}
+
+#[must_use]
+pub struct StateTablePostCommit<
```
BugenZhao, line 691: Can we add assertion for `self.inner.on_post_commit` when it gets `drop`ped?

src/stream/src/executor/backfill/snapshot_backfill/executor.rs
```diff
@@ -209,8 +209,9 @@ impl<S: StateStore> SnapshotBackfillExecutor<S> {
 
                     let recv_barrier = self.barrier_rx.recv().await.expect("should exist");
                     assert_eq!(first_upstream_barrier.epoch, recv_barrier.epoch);
-                    backfill_state.commit(recv_barrier.epoch).await?;
+                    let post_commit = backfill_state.commit(recv_barrier.epoch).await?;
                     yield Message::Barrier(recv_barrier);
+                    post_commit.post_yield_barrier(None).await?;
```
BugenZhao, line 214: Is it possible to introduce a sugar method for `post_yield_barrier(None)`?

BugenZhao (approved):
LGTM!

BugenZhao:
cc @stdrc to also take a look

src/stream/src/common/table/state_table.rs
```diff
@@ -162,6 +162,8 @@ pub struct StateTableInner<
     op_consistency_level: StateTableOpConsistencyLevel,
 
     clean_watermark_index_in_pk: Option<i32>,
+
+    on_post_commit: bool,
```
stdrc, line 166: Please leave some comment for this field, just like those in the PR description.

src/stream/src/common/table/state_table.rs
stdrc: +1

stdrc (approved):
Generally LGTM. Please copy some of the PR description to code comment for future reference.

src/stream/src/common/table/state_table.rs
wenym1: No. We cannot ensure that the `post_yield_barrier` can be called if the actor is aborted after yield barrier but before calling `post_yield_barrier`.

src/stream/src/executor/backfill/snapshot_backfill/executor.rs
wenym1: In most cases, if we are sure that there is no update vnode bitmap, we should call `state_table.commit_assert_no_update_vnode_bitmap`. This special call of `post_yield_barrier(None)` is only a temporary workaround for snapshot backfill executor, which hasn't, but will support update vnode bitmap in the future. The snapshot backfill executor has its own `ProgressState` struct that wraps the `StateTable` and has its own `commit` method. We don't have to provide a separate method sugar `commit_assert_no_update_vnode_bitmap` for this temporary workaround.

PR #20221, opened 2025-01-20
fix(frontend): cursor should not be schema-bound

BugenZhao:

According to Postgres' specification, `CURSOR`s are session-scoped objects and are not schema-bound. We should reject the following statement; however, we currently accept it.

```
postgres=# declare a.b cursor for select 1;
ERROR:  syntax error at or near "."
LINE 1: declare a.b cursor for select 1;
```

This PR fixes this by changing the AST type of `cursor_name` from `ObjectName` to `Ident`.

This should not be considered a breaking change, as cursors are temporary and only active within a session.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/handler/declare_cursor.rs
+++ b/src/frontend/src/handler/declare_cursor.rs
@@ -18,7 +18,9 @@ use pgwire::pg_response::{PgResponse, StatementType};
 use risingwave_common::catalog::Field;
 use risingwave_common::session_config::QueryMode;
 use risingwave_common::util::epoch::Epoch;
-use risingwave_sqlparser::ast::{DeclareCursorStatement, ObjectName, Query, Since, Statement};
+use risingwave_sqlparser::ast::{
+    DeclareCursorStatement, Ident, ObjectName, Query, Since, Statement,
+};
 
 use super::query::{
     gen_batch_plan_by_statement, gen_batch_plan_fragmenter, BatchPlanFragmenterResult,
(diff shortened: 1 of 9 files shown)


src/frontend/src/handler/declare_cursor.rs
```diff
@@ -54,17 +56,15 @@ pub async fn handle_declare_cursor(
...
-        Binder::resolve_schema_qualified_name(db_name, cursor_name.clone())?;
-
-    let cursor_from_subscription_name = sub_name.0.last().unwrap().real_value().clone();
-    let subscription =
-        session.get_subscription_by_name(schema_name, &cursor_from_subscription_name)?;
+    let subscription = {
+        let db_name = &session.database();
+        let (sub_schema_name, sub_name) = Binder::resolve_schema_qualified_name(db_name, sub_name)?;
+        session.get_subscription_by_name(sub_schema_name, &sub_name)?
+    };
```
BugenZhao, line 67: Previously, we used the schema name from `cursor: ObjectName` to resolve the subscription. This PR addresses and corrects that issue.

yezizp2012 (approved):

WanYixian:
Hi @BugenZhao, shall we list this change in document or release note, or both? Thanks in advance

BugenZhao:
> Hi @BugenZhao, shall we list this change in document or release note, or both? Thanks in advance

I think it's not that necessary. :)

pjpringlenom:
The risingwave-py library constructs cursor with schema prefix.

BugenZhao:
> The risingwave-py library constructs cursor with schema prefix.

Thanks for your feedback. Will resolve in https://github.com/risingwavelabs/risingwave-py/pull/14.

pjpringlenom:
Any test case for the **FULL** cursor subscription. This was the one erroring about the schema.

DECLARE curxxxx subscription CURSOR FOR mysub **full**

PR #20236, opened 2025-01-21
fix(source): modify the logic of gzip compressed file source

wcy-fdu:

Previously the offset variable serves two main purposes:
1. It is used to check whether a file has been completely read. In the fetch executor, when offset >= split.size, it indicates that the end of the file has been reached, prompting the removal of the corresponding entry from the state table.
2. After recovery occurs, offset prevents the reading process from starting over from the beginning of the file. Instead, it allows the reading to continue directly from the last successfully read offset.

Then, we realized that the current coupling between the lister and the fetcher/reader is a bit heavy: the fetcher relies on the size obtained from the lister to determine whether the file has been completely read and to perform deletions. This requires that the representations of offset and size must be consistent. We have encountered two bugs at this point:
- The offset for Parquet should be based on the number of rows rather than the size.
- The offset and size for gzip-compressed files are not equal.

In fact, the reader itself can fully sense whether it has finished reading, so the lister and reader can be completely independent. The lister is only responsible for informing the reader which files to read, while the reader is responsible for reading and informing the fetch executor whether it has finished reading.

Based on this, this PR makes the following changes:

1. After the reader has finished reading, ~~it yields one additional chunk, solely to indicate EOF~~ by setting the offset of the last chunk to `usize::MAX` to indicate eof.
2. The logic for the fetch executor to determine whether a file has been completely read and to delete it is changed to delete upon encountering the MAX offset.
3. For gzip files, to ensure exactly once:
  a. Its reader starts reading from the beginning each time.
  b. When yielding, it only yields chunks that are larger than the offset in the previous state table to ensure exactly once.

This change is effective for all formats (Parquet, CSV, JSON, gzip). After this, we can stop recording the file size in the state table, and we can modify this in a future PR.

## Checklist

## Documentation

<details>

</details>


wcy-fdu:
Tests will be added shortly.

src/connector/src/source/filesystem/opendal_source/opendal_reader.rs
```diff
@@ -182,20 +182,63 @@ impl<Src: OpendalSource> OpendalReader<Src> {
...
+            // 1. It is used to check whether a file has been completely read. In the fetch executor,
+            //    when `offset >= split.size`, it indicates that the end of the file has been reached,
+            //    prompting the removal of the corresponding entry from the state table.
+            //
+            // 2. After recovery occurs, `offset` prevents the reading process from starting
+            //    over from the beginning of the file. Instead, it allows the reading to continue
+            //    directly from the last successfully read `offset`.
+            //
+            // Special handling is required for files compressed with gzip:
+            // - Gzip files must always be read from the beginning; they cannot be read from
```
tabVersion, line 195: So we cannot achieve exactly once if gzip enabled?

src/stream/src/executor/source/fetch_executor.rs
```diff
@@ -374,10 +374,11 @@ impl<S: StateStore, Src: OpendalSource> FsFetchExecutor<S, Src> {
                                     .unwrap();
                             debug_assert_eq!(mapping.len(), 1);
                             if let Some((split_id, offset)) = mapping.into_iter().next() {
-                                let row = state_store_handler
-                                    .get(split_id.clone())
-                                    .await?
-                                    .expect("The fs_split should be in the state table.");
+                                let row = state_store_handler.get(split_id.clone()).await?
+                                    .unwrap_or_else(|| {
+                                        panic!("The fs_split (file_name) {:?} should be in the state table.",
```
tabVersion, line 379: we happen to meet this error, better not to panic here.

src/connector/src/source/filesystem/opendal_source/opendal_reader.rs
wcy-fdu: I think so.

kwannoel (commented):
anyway to test

src/connector/src/source/filesystem/opendal_source/opendal_reader.rs
hzxa21: Good point. I think exactly once is important for source. Some ideas to make reading gzip files exactly once:
1. Yield one large chunk per gzip file containing all contents after decompression. This ensures that no barrier can be placed in the middle of the file but will increase memory usage.
2. Pause fetcher upstream while reading a gzip file. This also ensures that no barrier can be placed in the middle.
3. Use offset and size after decompression to track gzip file fetch progress, like what we already did right now. Opendal reader always starts from the beginning and skip contents below the recorded offset.

Personally perfer 3.

wcy-fdu: I also prefer 3 because it has the least impact.

src/stream/src/executor/source/fetch_executor.rs
wcy-fdu: I think should panic, otherwise it lost data.

src/connector/src/source/filesystem/opendal_source/opendal_reader.rs
wcy-fdu: Resolved, had some offline discussion with @hzxa21 .
Please see the description of the PR for specific details. I modified some of the lister and fetcher logic. cc @tabVersion This part was designed by you, for awareness.

src/connector/src/source/filesystem/opendal_source/opendal_reader.rs
```diff
@@ -132,11 +132,19 @@ impl<Src: OpendalSource> OpendalReader<Src> {
...
         let object_name = split.name.clone();
         let start_offset = split.offset;
-        let reader = op
-            .read_with(&object_name)
-            .range(start_offset as u64..)
-            .into_future() // Unlike `rustc`, `try_stream` seems require manual `into_future`.
-            .await?;
+        // After a recovery occurs, for gzip-compressed files, it is necessary to read from the beginning each time,
+        // other files can continue reading from the last read `start_offset`.
+        let reader = match object_name.ends_with(".gz") || object_name.ends_with(".gzip") {
```
hzxa21, line 137: In addition to suffix check, should we also check `compression_format` (See L152)?

src/connector/src/source/filesystem/opendal_source/opendal_reader.rs
```diff
@@ -181,23 +189,28 @@ impl<Src: OpendalSource> OpendalReader<Src> {
...
             let msg_offset = (offset + n_read).to_string();
-            batch.push(SourceMessage {
-                key: None,
-                payload: Some(std::mem::take(&mut line_buf).into_bytes()),
-                offset: msg_offset,
-                split_id: split.id(),
-                meta: SourceMeta::Empty,
-            });
+            if (object_name.ends_with(".gz") || object_name.ends_with(".gzip"))
+                && offset + n_read <= start_offset
```
hzxa21, line 194: The logic here is incorrect because offset is assigned to `start_offset` in L168. A clearer approach is to return the reader as well as the read offset when constructing the reader. For gzip reader, the initial read offset will be `0` and for others the, the initial read offset will be `start_offset`

src/connector/src/source/filesystem/opendal_source/opendal_reader.rs
```diff
@@ -181,23 +189,28 @@ impl<Src: OpendalSource> OpendalReader<Src> {
...
+                    offset: msg_offset,
+                    split_id: split.id(),
+                    meta: SourceMeta::Empty,
+                });
+            }
+
             offset += n_read;
             partition_input_bytes_metrics.inc_by(n_read as _);
 
+            partition_input_bytes_metrics.inc_by(n_read as _);
```
hzxa21, line 212: Duplicate codes.

src/connector/src/parser/parquet_parser.rs
```diff
@@ -65,6 +65,12 @@ impl ParquetParser {
 
             yield chunk;
         }
+        // We no longer determine whether a file is finished by comparing `offset >= size`.
+        // The reader itself can sense whether it has reached the end.
+        // Therefore, after a file is read completely, we yield one more chunk marked as EOF, with its offset set to `usize::MAX` to indicate that it is finished.
+        // In fetch executor, if `offset = usize::MAX` is encountered, the corresponding file can be deleted from the state table.
+        let eof_chunk = self.generate_eof_chunk()?;
```
hzxa21, line 72: It is not a good idea to generate eof chunk in `ParquetParser` because it is not just used by file source, but also file scan TVF. How about generate the eof chunk [here](https://github.com/risingwavelabs/risingwave/blob/e760a896f565c475eea9c1df73d3972ca098b1a2/src/connector/src/source/filesystem/opendal_source/opendal_reader.rs#L118)? I don't think the eof chunk cares about the file format.

src/connector/src/source/filesystem/opendal_source/opendal_reader.rs
wcy-fdu: I think no need, but we can add a check: if compression_format is set, suffix must be `.gz` or `.gzip`.

src/connector/src/parser/parquet_parser.rs
wcy-fdu: I didn't make the change here because the hidden columns are inconsistent for parquet and other formats, so it's not appropriate to use a common chunk here.

- For parquet encode, the order of hidden column is (`_row_id` `_rw_s3_v2_file`, `_rw_s3_v2_offset`);
- for other encode, the order of hidden column is ( `_rw_s3_v2_file`, `_row_id`,`_rw_s3_v2_offset`).

Actually I haven't found out why the order of hidden columns is inconsistent for these two encodings.

But since ParquetParser is also used for file scan TVF, how about judging the encode [here](https://github.com/risingwavelabs/risingwave/blob/e760a896f565c475eea9c1df73d3972ca098b1a2/src/connector/src/source/filesystem/opendal_source/opendal_reader.rs#L118) and
- for parquet encode, construct the eof chunk here
- for other encode, construct the eof chunk inside reader.


The above is one solution. A better way is to unify the order of hidden columns, but I am not sure how much work this will take, as this fix is ​​quite urgent. What do you think? And if you know why the hidden columns are not in the same order, please let me know.

wcy-fdu: I changed it as mentioned above, correct me if I'm wrong.

tabVersion:
the motivation and solution LGTM

e2e_test/s3/file_source.py
```diff
@@ -265,19 +282,45 @@ def _table():
...
     )
     run_id = str(random.randint(1000, 9999))
+    
     _local = lambda idx: f'data_{idx}.{fmt}'
-    _s3 = lambda idx: f"{run_id}_data_{idx}.{fmt}"
+    if fmt == 'json':
+        _s3 = lambda idx: f"{run_id}_data_{idx}.{fmt}.gz"
+        for idx, file_str in enumerate(formatted_files):
+            with open(_local(idx), "w") as f:
+                with gzip.open(_local(idx) + '.gz', 'wb') as f_gz:
```
tabVersion, line 291: better add some comments to clarify we only do gzip compress for json encoding

src/connector/src/source/filesystem/opendal_source/opendal_reader.rs
```diff
@@ -116,6 +118,19 @@ impl<Src: OpendalSource> OpendalReader<Src> {
             for chunk in chunk_stream {
                 yield chunk?;
             }
+
+            if let EncodingProperties::Parquet = &self.parser_config.specific.encoding_config {
+                // We no longer determine whether a file is finished by comparing `offset >= size`, the reader itself can sense whether it has reached the end.
+                // Therefore, after a file is read completely, we yield one more chunk marked as EOF, with its offset set to `usize::MAX` to indicate that it is finished.
+                // In fetch executor, if `offset = usize::MAX` is encountered, the corresponding file can be deleted from the state table.
```
tabVersion, line 125: so for all encoding (parquet and json?), we have the usize::MAX as the EOF mark?

src/connector/src/source/filesystem/opendal_source/opendal_reader.rs
```diff
@@ -116,6 +118,19 @@ impl<Src: OpendalSource> OpendalReader<Src> {
...
+
+                // FIXME(wcy-fdu): The order of hidden columns in parquet encode and other encodes is inconsistent, maybe we can yeild a common eof chunk for both parquet encode and other encodes.
+                let eof_chunk = Self::generate_eof_chunk_for_parquet_encode(
```
tabVersion, line 128: TBH, I don't get why yielding an empty chunk is essential... shouldn't we carry the offset info along with data chunk? We can directly change the offset to usize::MAX when reaching a file's end.

tabVersion (commented):
rest LGTM

src/connector/src/source/filesystem/opendal_source/opendal_reader.rs
wcy-fdu: yes, I think usize::MAX is safe.

src/stream/src/executor/source/mod.rs
```diff
@@ -134,30 +136,67 @@ pub async fn apply_rate_limit(stream: BoxSourceChunkStream, rate_limit_rps: Opti
...
-        yield chunk;
+    if rate_limit_rps.is_none() || chunk_size == 0 {
+        // no limit, or empty chunk
+        return chunk;
     }
+
+    let limit = rate_limit_rps.unwrap() as u64;
+    let required_permits = chunk.compute_rate_limit_chunk_permits();
+    if required_permits > limit {
+        // This should not happen after the mentioned PR.
```
kwannoel, line 189: What mentioned PR

kwannoel (commented):
Changes for metrics LGTM. Need another review for other parts.

src/stream/src/executor/source/mod.rs
wcy-fdu: not related.

src/stream/src/executor/source/fs_fetch_executor.rs
```diff
@@ -312,100 +336,85 @@ impl<S: StateStore, Src: OpendalSource> FsFetchExecutor<S, Src> {
...
-                                                    0,
-                                                    total_row_num - 1, // -1 because offset start from 0.
-                                                ),
+                                    let file_assignment = chunk
+                                        .data_chunk()
+                                        .rows()
+                                        .map(|row| {
+                                            let filename = row.datum_at(0).unwrap().into_utf8();
+
+                                            let size = row.datum_at(2).unwrap().into_int64();
```
tabVersion, line 345: prefer avoid hard cod

src/stream/src/executor/source/fs_fetch_executor.rs
```diff
@@ -312,100 +336,85 @@ impl<S: StateStore, Src: OpendalSource> FsFetchExecutor<S, Src> {
...
+                                let mapping = get_split_offset_mapping_from_chunk(
+                                    &chunk, split_idx, offset_idx,
+                                )
+                                .unwrap();
+                                debug_assert_eq!(mapping.len(), 1);
+                                if let Some((split_id, _offset)) = mapping.into_iter().next() {
+                                    delete_file = split_id.clone();
+                                    let row = state_store_handler.get(split_id.clone()).await?
+                                        .unwrap_or_else(|| {
+                                            panic!("The fs_split (file_name) {:?} should be in the state table.",
```
tabVersion, line 389: report error and skip the file? Seems the case is not that critical to the system, WDYT

src/stream/src/executor/source/mod.rs
```diff
@@ -134,30 +136,67 @@ pub async fn apply_rate_limit(stream: BoxSourceChunkStream, rate_limit_rps: Opti
...
-        if rate_limit_rps.is_none() || chunk_size == 0 {
-            // no limit, or empty chunk
-            yield chunk;
-            continue;
-        }
+#[try_stream(ok = Option<StreamChunk>, error = ConnectorError)]
+pub async fn apply_rate_limit_with_for_streaming_file_source_reader(
+    stream: BoxStreamingFileSourceChunkStream,
+    rate_limit_rps: Option<u32>,
+) {
```
tabVersion, line 147: Do you mean to avoid nesting try_stream as much as we can? @kwannoel 
The upper layer `apply_rate_limit` can be removed?

src/stream/src/executor/source/fs_fetch_executor.rs
```diff
@@ -312,100 +336,85 @@ impl<S: StateStore, Src: OpendalSource> FsFetchExecutor<S, Src> {
...
+                        // file has been fully read. Upon encountering `None`, the file is deleted.
+                        Either::Right(optional_chunk) => match optional_chunk {
+                            Some(chunk) => {
+                                let mapping = get_split_offset_mapping_from_chunk(
+                                    &chunk, split_idx, offset_idx,
+                                )
+                                .unwrap();
+                                debug_assert_eq!(mapping.len(), 1);
+                                if let Some((split_id, _offset)) = mapping.into_iter().next() {
+                                    delete_file = split_id.clone();
```
tabVersion, line 386: maybe `reading_file`? delete seems the file has already dropped

tabVersion (commented):
the overall idea LGTM, may take another look later

src/stream/src/executor/source/fs_fetch_executor.rs
wcy-fdu: +1 for avoiding hard code. 
This is just following the previous approach. Since the fetch executor is not aware of the schema, this is hard coded after the fetch executor is introduced, which is also the reason for the parquet include hidden_column panic I mentioned above.

src/stream/src/executor/source/fs_fetch_executor.rs
wcy-fdu: prefer to panic because it's not that we missed some files, but that some files that should have been read were deleted.

src/stream/src/executor/source/mod.rs
wcy-fdu: Not related.

tabVersion (approved):

src/connector/src/source/filesystem/opendal_source/opendal_reader.rs
```diff
@@ -177,32 +202,36 @@ impl<Src: OpendalSource> OpendalReader<Src> {
...
-            // understand that this message offset is not semantically consistent with the offset of
-            // other source connectors.
-            let msg_offset = (offset + n_read).to_string();
-            batch.push(SourceMessage {
-                key: None,
-                payload: Some(std::mem::take(&mut line_buf).into_bytes()),
-                offset: msg_offset,
-                split_id: split.id(),
-                meta: SourceMeta::Empty,
-            });
```
stdrc, line 195: Sorry I just see this change a long time after this PR got merged. Want to confirm that, is the `msg_offset` for plain files (not compressed) still the "end offset" of a line? If so then I think it is still not intuitive, and still need a comment here.

PR #20241, opened 2025-01-21
fix(meta): fix receiving scheduled command on blocked database

wenym1:

Discovered in https://buildkite.com/risingwavelabs/main-cron/builds/4257#_

Get the following panic in madsim test.
```
thread '<unnamed>' panicked at src/meta/src/barrier/checkpoint/control.rs:453:17:
should be at running: should not have command when not running
```

The assertion is based on the assumption that when a database is in recovery, it should not receive command. We have marked the database as `Blocked` when it enters recovery, so normal command won't be able to add to the command queue. However, we still allow `DropXxx` command to be enqueued, and since we didn't prevent polling the queue of the recovering database, we will still receive this command and cause panic in global barrier manager.

In this PR, we will change to avoid polling command queue of any blocked database. When the database is marked ready, we will notify on any change subscriber if the database queue is not empty.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/ci/scripts/deterministic-e2e-test.sh
+++ b/ci/scripts/deterministic-e2e-test.sh
@@ -28,6 +28,26 @@ export LOGDIR=.risingwave/log
 
 mkdir -p $LOGDIR
 
-echo "--- deterministic simulation e2e, ci-3cn-2fe, parallel, streaming"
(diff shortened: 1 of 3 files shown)


yezizp2012 (approved):

kwannoel (commented):
Does this mean we cannot drop stream jobs until db recovered?

wenym1:
> Does this mean we cannot drop stream jobs until db recovered?

No. The stream job will have been removed from catalog before we enqueue the command. The command can still be enqueue, and will take affect either by `pre_apply_drop_cancel_scheduled` during recovery, or by the normal handling logic after db has recovered and resumed running.

kwannoel (approved):

PR #20258, opened 2025-01-22
feat(metrics): add counter for retry create stream reader

tabVersion:

follow up https://github.com/risingwavelabs/risingwave/pull/19805

manually test 

1. create table with kafka
2. stop kafka container
3. trigger recovery in psql
4. observe grafana

image

## Checklist

## Documentation

<details>

</details>


src/stream/src/executor/source/source_executor.rs
```diff
@@ -543,14 +543,20 @@ impl<S: StateStore> SourceExecutor<S> {
...
             let source_ctx = Arc::new(source_ctx);
+            let source_ctx_clone = source_ctx.clone();
             let mut build_source_stream_fut = Box::pin(async move {
                 let backoff = get_infinite_backoff_strategy();
                 tokio_retry::Retry::spawn(backoff, || async {
+                    let source_ctx_inner = source_ctx_clone.clone();
+                    source_ctx_inner.metrics.init_stream_reader_retry_count.with_guarded_label_values(&[
+                        &source_ctx_inner.actor_id.to_string(),
+                        &source_ctx_inner.source_name,
+                    ]).inc();
```
xxchan, line 554: Could we simply move this into `source_reader.build_stream`? And perhaps rename to `stream_reader_build_count`.

xxchan: Actually, I feel the figure is a little hard to understand. Why did it go up and down 🤔

image

src/connector/src/source/monitor/metrics.rs
```diff
@@ -89,6 +90,13 @@ impl SourceMetrics {
             registry
         )
         .unwrap();
+        let init_stream_reader_retry_count = register_guarded_int_counter_vec_with_registry!(
+            "source_init_stream_reader_retry_count",
+            "Total number of retries to initialize stream reader",
+            &["actor_id", "source_name",],
+            registry
+        )
+        .unwrap();
```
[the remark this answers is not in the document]
tabVersion: I think it overlaps with `source_status_is_up` (from source enumerator), so I don't add it to grafana. 
But here we focus more on the retry on the compute node, planning add it to cloud alert.

src/stream/src/executor/source/source_executor.rs
tabVersion: I am struggling with it too, suspect somewhere calls a rate.

src/connector/src/source/monitor/metrics.rs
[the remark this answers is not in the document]
xxchan: Oh, I don't know we have `source_status_is_up`! 

Then what about adding it to the same panel (`Source Upstream Status`), so that we can both know whether the source is up on meta and each CN. 🤔

lmatz (approved):

src/stream/src/executor/source/source_executor.rs
tabVersion: I believe the reason is RW is reporting the metric from multiple nodes and the count value from each node varies, leading to the fluctuation. 
But it is irrelevant to the alert part, we always alert when the value is non-zero.

tabVersion:
let's wait for https://github.com/risingwavelabs/risingwave/pull/20295, to minimize the changes.

fuyufjh (approved):

PR #20261, opened 2025-01-22
feat: support cross-db query in frontend and meta

yezizp2012:

1. enable support for cross-database binding in batch operations, `create_mv`, and `create_sink`.
2. check db `CONNECT` privilege for cross-db query.
3. Integrated cross-db snapshot backfill executor, and address issues in metadata validation.
4. reject drop mv/sink cascade with cross-db upstreams.
5. verify the existence of log stores for all cross-db upstreams in the DDL.

## Checklist

## Documentation

<details>

Support cross-db query for both batch and streaming. The current requirements include:

1. There is at least one subscription created on the upstream table or materialized view.

2. The user is super or has connect permissions for the upstream databases and must also have the corresponding relation's query permissions.

</details>


src/meta/src/stream/stream_graph/actor.rs
```diff
@@ -173,6 +173,11 @@ impl FragmentActorBuilder {
             // "Leaf" node `StreamScan`.
             NodeBody::StreamScan(stream_scan) => {
                 let input = stream_node.get_input();
+                if stream_scan.stream_scan_type() == StreamScanType::CrossDbSnapshotBackfill {
+                    // CrossDbSnapshotBackfill is a special case, which doesn't have any upstreams.
```
wenym1, line 177:
```suggestion
                    // CrossDbSnapshotBackfill is a special case, which doesn't have any upstream actor and always reads from log store.
```

src/meta/src/rpc/ddl_controller.rs
```diff
@@ -1601,9 +1601,29 @@ impl DdlController {
...
+            .iter()
+            .filter(|id| {
+                !cross_db_snapshot_backfill_info
+                    .upstream_mv_table_id_to_backfill_epoch
+                    .contains_key(id)
+            })
+            .cloned()
+            .collect();
+
         let (upstream_root_fragments, existing_actor_location) = self
```
wenym1, line 1624: It seems that the `upstream_root_actors` obtained from `upstream_root_fragments` will be used in create mview tracker to track the progress. Is it correct to filter out the cross db backfill actor here?

src/frontend/src/optimizer/mod.rs
```diff
@@ -1027,6 +1041,12 @@ impl PlanRoot {
             .streaming_use_snapshot_backfill()
     }
 
+    pub fn should_use_cross_db_snapshot_backfill(&self) -> bool {
+        // todo:
+        // self.plan.append_only()
+        self.plan.emit_on_window_close()
```
wenym1, line 1047: Any reason for this?

src/frontend/src/handler/privilege.rs
```diff
@@ -62,79 +66,41 @@ pub(crate) fn resolve_relation_privileges(
...
-        BoundStatement::CreateView(ref create_view) => {
-            objects.extend(resolve_query_privileges(&create_view.query))
-        }
-    };
-    objects
-}
-
 /// resolve privileges in `query`
-pub(crate) fn resolve_query_privileges(query: &BoundQuery) -> Vec<ObjectCheckItem> {
+fn resolve_query_privileges(query: &BoundQuery) -> (Vec<ObjectCheckItem>, HashSet<DatabaseId>) {
```
wenym1, line 90: It seems that all usages are first collect the two variables, and then extend to the outer list or sets. If so, we may pass the &mut of the outer list or sets here instead of returning a new one.

src/frontend/src/handler/privilege.rs
```diff
@@ -163,6 +129,100 @@ impl SessionImpl {
         Ok(())
     }
 
+    pub fn check_privileges_for_query(&self, query: &BoundQuery) -> Result<()> {
```
wenym1, line 132: We can avoid cloning the `query` if we still have a common method `check_privileges`, and then we can directly `resolve_query_privileges(query)` and then `check_privileges`.

e2e_test/streaming/cross_db_mv.slt
```diff
@@ -0,0 +1,88 @@
...
+set database to d1;
+
+statement ok
+drop table t1 cascade;
+
+statement ok
+drop database d2;
+
+system ok
+risedev psql -c 'drop database d1;'
```
wenym1, line 88: Seems not supported to run command in madsim.

yezizp2012: Yes, I also discovered that using a system command to drop the database is unsafe, as it may disrupt the ongoing SQL connection while it has been set to database `d1`.
Unfortunately, I found that it is impossible to directly run the entire cross db mv logic in slt. I will comment out this test in this PR first and add some shell scripts later to test it.

wenym1: Can we assume that the original database is always `dev`, and then switch back to dev and then drop database d1?

wenym1: Maybe we can use the `connection` statement to perform the operation on other database in a new connection. The statement can be used like the following example.
```
connection other
query T
show RW_STREAMING_ENABLE_DELTA_JOIN;
----
false
```

yezizp2012: 👍 Haven't found this usage before, let me give it a try!

src/frontend/src/optimizer/mod.rs
yezizp2012: I was going to add a helper function, but didn't implement or use it later. I will delete it.

src/meta/src/rpc/ddl_controller.rs
yezizp2012: Yes, you're right, we can't simply filter out them to avoid panic. `upstream_root_actors` will be used to track the progress along with `dispatchers`, which does not apply to cross db backfill. The progress info cannot be derived from these fields, so I changed to use an existing function that extract the info from field `fragments` instead. I also removed the field `upstream_root_actors` in the corresponding context.

wenym1 (approved):
LGTM!

Shall we add a release note to announce this new feature, and mention that currently user need to create a subscription on the table and should not drop the subscription?

yezizp2012:
Cc @WanYixian . An example of performing a cross-database query from table t1 in database d1 in database d2:
```sql
# in database d1
CREATE TABLE t1 (v1 INT);

# at lease one subscription required to enable log store for t1
CREATE subscription sub FROM t1 WITH (retention = '1D');

# switch to database d2
SET DATABASE TO d2;

CREATE TABLE t2 (v2 INT);

SELECT COUNT(*) FROM d1.public.t1;

CREATE MATERIALIZED VIEW mv1 AS SELECT * FROM d1.public.t1;

CREATE MATERIALIZED VIEW mv2 AS SELECT t1.v1, t2.v2 FROM d1.public.t1 JOIN t2 ON t1.v1 = t2.v2;
```

PR #20287, opened 2025-01-24
feat: add fips option for aws-lc

yuhao-su:

 add fips option for aws-lc

Manually tested in ci and locally.

Need to add a pipeline to build fips images

## Checklist

## Documentation

<details>

</details>


src/utils/workspace-config/Cargo.toml
```diff
@@ -51,6 +53,11 @@ vendored-openssl-sys = { package = "openssl-sys", version = "0.9.96", optional =
 dynamic-zstd-sys = { package = "zstd-sys", version = "2", optional = true, default-features = false, features = [
     "pkg-config",
 ] }
+
+# FIPS
+aws-lc-rs ={ version = "1.6", optional = true, default-features = false, features = [
+    "fips",
+] }
```
xxchan, line 60: Why not enable it by default?

yuhao-su: I don't have a strong reason against doing so. It's simply because I didn't see any other product enabling this option by default.

lmatz (approved):

cyliu0:
> Need to add a pipeline to build fips images

Which images need to enable this? nightly? stable version?

yuhao-su:
> > Need to add a pipeline to build fips images
> 
> Which images need to enable this? nightly? stable version?

For now I guess there is no need to release with this option. But can we just build the image with the option and run a e2e test in our release pipeline? cc @lmatz

src/utils/workspace-config/Cargo.toml
tabVersion: +1 for trying enable it by default.It does not bring us any additional cost I think.

lmatz:
> > > Need to add a pipeline to build fips images
> > 
> > 
> > Which images need to enable this? nightly? stable version?
> 
> For now I guess there is no need to release with this option. But can we just build the image with the option and run a e2e test in our release pipeline? cc @lmatz

One customer is requesting fips compliance and we may need to deliver the image, not 100% sure but could be soon

but we can add the test first

PR #20295, opened 2025-01-24
refactor: make rebuild reader inner of reader stream

tabVersion:

As title, encapsulate the rebuilding source reader process to the source stream inner and make this process independent of the source executor itself. 

Key changes:

* on source executor & source backfill executor first startup (the actor is first created)
  * try create reader stream and exit if meeting error when consuming
* on recovery
  * the executor keep trying to connect **inside** reader stream, rather than rebuilding the message stream

for a retry_reader_stream: it tries to start with the connector given in `into_retry_stream`, but if something happens, leading to rebuild the reader, it calls the closure of type `GetLatestSplitInfoFn` to sync the latest state from outside to keep exactly-once.

image

## Checklist

## Documentation

<details>

</details>


src/stream/src/executor/source/executor_core.rs
```diff
@@ -35,7 +37,7 @@ pub struct StreamSourceCore<S: StateStore> {
 
     /// Split info for stream source. A source executor might read data from several splits of
     /// external connector.
-    pub(crate) latest_split_info: HashMap<SplitId, SplitImpl>,
+    pub(crate) latest_split_info: Arc<RwLock<HashMap<SplitId, SplitImpl>>>,
```
wenym1, line 40: We may try using the non-async `parking_lot::RwLock` to ensure that any stream or future will not capture the `RwLockGuard` (hold the guard across await point). If a stream or future capture the lock guard by accidence, the stream or future won't be `Send` and we will fail to pass the compiler check. Holding the lock guard across await point can easily cause deadlock, and from my understanding, in our current code, there is no code that we must hold the lock guard to call async method.

wenym1 (commented):
Instead of letting the reader stream access and modify the `latest_split_info`, can we just own the `latest_split_info` in the source executor rather than being shared in the core, and let reader stream yield the updated split info of the chunk along with the chunk? And then when the source executor receives a new chunk from reader stream, it can also get the updated split info and then apply it to the its `latest_split_info`.

When reader stream returns error, the source executor can then recreate the reader stream with the `latest_split_info` it owns.

src/stream/src/executor/source/executor_core.rs
tabVersion: > We may try using the non-async parking_lot::RwLock to ensure that any stream or future will not capture the RwLockGuard (hold the guard across await point).

Tried `parking_lot::RwLock` and it fails to compile and I switched to tokio locks. 

> in our current code, there is no code that we must hold the lock guard to call async method.

make sense, let me open another pr for it.

PR #20297, opened 2025-01-24
fix(playground): listen to `127.0.0.1` instead of `0.0.0.0` for gRPC services

BugenZhao:

`risedev d` listens on `127.0.0.1` while playground (single-node) listens on `0.0.0.0`. As a result, when there's a `risedev d` running in the background, starting `risedev p` can be confusing:

- meta node of `p` listens on `0.0.0.0`, which does not conflict with `risedev d` and starts successfully
- then, compute node and frontend may connect to the meta node of `d`
- no error will occur as long as the wire protocol is compatible between the build of `d` and `p`, but the behavior can be quite weird

This PR changes the gRPC listen addresses of playground (single-node) to `127.0.0.1` as well to prevent this from happening. Please note that we continue to have the frontend pgwire service and the meta dashboard service listening on `0.0.0.0` to allow for remote connections.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/cmd_all/src/single_node.rs
+++ b/src/cmd_all/src/single_node.rs
@@ -194,14 +194,11 @@ pub fn map_single_node_opts_to_standalone_opts(opts: SingleNodeOpts) -> ParsedSt
     }
 
     // Set listen addresses (force to override)
-    meta_opts.listen_addr = "0.0.0.0:5690".to_owned();
+    meta_opts.listen_addr = "127.0.0.1:5690".to_owned();
     meta_opts.advertise_addr = "127.0.0.1:5690".to_owned();
     meta_opts.dashboard_host = Some("0.0.0.0:5691".to_owned());
-    compute_opts.listen_addr = "0.0.0.0:5688".to_owned();
-    compactor_opts.listen_addr = "0.0.0.0:6660".to_owned();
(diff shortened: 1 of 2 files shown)


kwannoel (approved):

src/cmd_all/src/single_node.rs
```diff
@@ -194,14 +194,11 @@ pub fn map_single_node_opts_to_standalone_opts(opts: SingleNodeOpts) -> ParsedSt
...
 
     // Set listen addresses (force to override)
-    meta_opts.listen_addr = "0.0.0.0:5690".to_owned();
+    meta_opts.listen_addr = "127.0.0.1:5690".to_owned();
     meta_opts.advertise_addr = "127.0.0.1:5690".to_owned();
     meta_opts.dashboard_host = Some("0.0.0.0:5691".to_owned());
-    compute_opts.listen_addr = "0.0.0.0:5688".to_owned();
-    compactor_opts.listen_addr = "0.0.0.0:6660".to_owned();
-    if let Some(frontend_addr) = &opts.node_opts.listen_addr {
-        frontend_opts.listen_addr.clone_from(frontend_addr);
```
BugenZhao, line 203: Already on L227

xxchan (approved):

PR #20298, opened 2025-01-24
feat(iceberg): support iceberg engine connection

chenzl25:

- Related: https://github.com/risingwavelabs/risingwave/issues/19418
- Introduce a session variable `iceberg_engine_connection` to allow users to provide their own bucket for the iceberg engine via iceberg connection. Currently, only warehouse information is allowed to be configured in the iceberg engine connection. Iceberg catalog is still handled by us in the meta sql backend. With this config, it can make us much easier to share iceberg tables with users, since the underlying warehouse is managed by users and they can have a better control of the warehouse credential.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/e2e_test/iceberg/test_case/iceberg_engine.slt
+++ b/e2e_test/iceberg/test_case/iceberg_engine.slt
@@ -113,3 +113,46 @@ select count(_row_id) from t_without_pk;
 
 statement ok
 DROP TABLE t_without_pk
+
+# test iceberg connection in iceberg engine
+statement ok
+create secret my_secret with (
+  backend = 'meta'
+) as 'hummockadmin';
+
+statement ok
+create connection my_conn
+with (
+    type = 'iceberg',
+    warehouse.path = 's3://hummock001/iceberg_connection',
+    s3.access.key = secret my_secret,
+    s3.secret.key = secret my_secret,
+    s3.endpoint = 'http://127.0.0.1:9301',
+    s3.region = 'us-west-2',
+);
+
+statement ok
+set iceberg_engine_connection = 'public.my_conn';
+
+statement ok
+create table t1(i int) with(commit_checkpoint_interval = 1) engine = iceberg;
+
+statement ok
+insert into t1 values(1);
+
+statement ok
+FLUSH;
+
+query ?
+select * from t1;
+----
+1
+
+statement ok
+DROP TABLE t1;
+
+statement ok
+DROP CONNECTION my_conn;
+
+statement ok
+DROP SECRET my_secret;
--- a/src/connector/src/connector_common/connection.rs
+++ b/src/connector/src/connector_common/connection.rs
@@ -210,7 +210,7 @@ impl Connection for IcebergConnection {
             }
         };
 
-        // test storage
+        // Test warehouse
         if let Some((scheme, bucket, root)) = info {
             match scheme.as_str() {
                 "s3" | "s3a" => {
@@ -246,7 +246,13 @@ impl Connection for IcebergConnection {
             }
         }
 
-        // test catalog
+        // If there is no catalog type, we won't test the catalog.
+        // The iceberg connection could be used to provide warehouse information only.
+        if self.catalog_type.is_none() {
+            return Ok(());
+        }
+
+        // Test catalog
         let iceberg_common = IcebergCommon {
             catalog_type: self.catalog_type.clone(),
             region: self.region.clone(),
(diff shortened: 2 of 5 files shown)


e2e_test/iceberg/test_case/iceberg_engine.slt
```diff
@@ -113,3 +113,46 @@ select count(_row_id) from t_without_pk;
...
+    type = 'iceberg',
+    warehouse.path = 's3://hummock001/iceberg_connection',
+    s3.access.key = secret my_secret,
+    s3.secret.key = secret my_secret,
+    s3.endpoint = 'http://127.0.0.1:9301',
+    s3.region = 'us-west-2',
+);
+
+statement ok
+set iceberg_engine_connection = 'public.my_conn';
```
xxchan, line 135: Why don't we use the `connection = conn` syntax, but use a session variable instead? :thinking

chenzl25: Finding a way to allow users to set a default behavior for a whole database. Adding a syntax is also ok, however, our iceberg table can used with a connector which also can have a connection. Maybe a little bit confused to the user.

xxchan: Oh, that's indeed a problem... So the `with` can be both for connector and for the table.

chenzl25: crunchybridge https://arc.net/l/quote/vdcxffua

xxchan: Oh, we will allow user to specify both volume and catalog (or any 1 of them) in the same connection, right?

chenzl25: yes. iceberg connection can specify catalog and bucket together. Although for iceberg engine, we only allow specifying bucket

src/frontend/src/handler/create_table.rs
```diff
@@ -1459,48 +1461,151 @@ pub async fn create_iceberg_engine_table(
...
+                    .ok_or_else(|| anyhow!("`s3.region` must be set in iceberg engine connection"))?
+                    .to_owned();
+                let _s3_endpoint = params.properties.get("s3.endpoint").map(|s| s.to_owned());
+                let _warehouse_path = params
+                    .properties
+                    .get("warehouse.path")
+                    .map(|s| s.to_owned())
+                    .ok_or_else(|| {
+                        anyhow!("`warehouse.path` must be set in iceberg engine connection")
+                    })?;
```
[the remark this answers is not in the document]
chenzl25: Iceberg connection could be specified either storage/volume information only or storage + catalog information. For the iceberg engine, we don't allow users to use their own catalog and storage other than s3, so we need to check the iceberg connection here.

chenzl25:
Now, the iceberg engine allows users to use the iceberg engine connection to provide catalog and storage information.

src/connector/src/connector_common/iceberg/mod.rs
```diff
@@ -96,9 +96,12 @@ pub struct IcebergCommon {
         deserialize_with = "deserialize_optional_bool_from_string"
     )]
     pub path_style_access: Option<bool>,
-    /// enable config load currently is used by iceberg engine, so it only support jdbc catalog.
+    /// enable config load currently is used by iceberg engine.
     #[serde(default, deserialize_with = "deserialize_optional_bool_from_string")]
     pub enable_config_load: Option<bool>,
+    /// enable compaction currently is used by iceberg engine.
+    #[serde(default, deserialize_with = "deserialize_optional_bool_from_string")]
+    pub enable_compaction: Option<bool>,
```
fuyufjh, line 104: According to the discussion today, shall we remove this now?

src/connector/src/connector_common/connection.rs
```diff
@@ -250,7 +250,13 @@ impl Connection for IcebergConnection {
             }
         }
 
-        // test catalog
+        // If there is no catalog type, we won't test the catalog.
+        // The iceberg connection could be used to provide warehouse information only.
```
fuyufjh, line 254: According to today's talk, we won't deliver a stage with self-hosted JDBC catalog; as a result, I think the iceberg connection must provide both warehouse information and catalog info.

src/connector/src/connector_common/iceberg/mod.rs
chenzl25: Yes.

src/connector/src/connector_common/connection.rs
chenzl25: Correct.

fuyufjh (approved):

PR #20316, opened 2025-01-26
fix: Fix the issue of actor migration panic caused by the in-place scale-down

shanicky:

In certain situations, a worker node may experience an in-place parallelism reduction, potentially leading to a scenario where some actors of the worker do not need to be migrated. As a result, the worker slot containing these actors is not included in the plan. The solution is straightforward: simply skip this worker slot.

## Checklist


yezizp2012 (approved):
LGTM

PR #20341, opened 2025-01-29
check chunk inconsistency

kwannoel:

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/executor/join/builder.rs
+++ b/src/stream/src/executor/join/builder.rs
@@ -158,7 +158,6 @@ impl<const T: JoinTypePrimitive, const SIDE: SideTypePrimitive> JoinChunkBuilder
     }
 
     pub fn post_process(c: StreamChunk) -> StreamChunk {
-        c.check_consistency();
         let mut c = StreamChunkMut::from(c);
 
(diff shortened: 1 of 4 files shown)


src/stream/src/executor/join/builder.rs
```diff
@@ -158,6 +158,7 @@ impl<const T: JoinTypePrimitive, const SIDE: SideTypePrimitive> JoinChunkBuilder
     }
 
     pub fn post_process(c: StreamChunk) -> StreamChunk {
+        c.check_consistency();
```
kwannoel, line 161: We check before doing post_process... but it does not panic here, meaning there's no trailing `UpdateDelete`.

kwannoel: It panics later on, when converting back from `StreamChunkMut` to `StreamChunk`: https://github.com/risingwavelabs/risingwave/pull/20341/files#diff-477cd761851ca4cd7ad5a34567afc8b2b4271f9c9096a0aa854f68847d5ad369R200

However I can't find anything in the logic below that can cause a hanging `UpdateDelete`.

kwannoel: Oh nevermind. There can be a hanging update delete because the visibility is marked to be blank...

Issue #20346, opened 2025-01-30
postgres_cdc: fails miserably when there are less columns in RW table than in PG source table


pkit:
### Describe the bug

See title
I really hope you don't panic when a new column is added upstream.

### Error message/log

```text
ERROR:  Panicked when handling the request: assertion `left == right` failed
  left: 1
 right: 2
This is a bug. We would appreciate a bug report at:
  https://github.com/risingwavelabs/risingwave/issues/new?labels=type%2Fbug&template=bug_report.yml
server closed the connection unexpectedly
        This probably means the server terminated abnormally
        before or while processing the request.
The connection to the server was lost. Attempting reset: Succeeded.
psql (16.6 (Ubuntu 16.6-0ubuntu0.24.04.1), server 13.14.0)
```

### To Reproduce

Postgres
```sql
CREATE TABLE IF NOT EXISTS test_52 (
                                       user_id UUID PRIMARY KEY,
                                       ar_col2 point[] NOT NULL
);
```
RW
```sql
CREATE SOURCE IF NOT EXISTS af_pg_src WITH (
                    connector='postgres-cdc',                                                                                          
                    hostname='host.docker.internal',
                    port=15432,
                    username='master',
                    password='password',
                    database.name='main'
                );
CREATE TABLE IF NOT EXISTS af_pg_test_57 (
                    user_id VARCHAR PRIMARY KEY

                )
                FROM af_pg_src TABLE 'public.test_52';
```

### Expected behavior

I expect the "unknown" columns to be ignored.
And I do not expect **PANIC** at all

### How did you deploy RisingWave?

docker compose

### The version of RisingWave

PostgreSQL 13.14.0-RisingWave-2.1.2 (13d928542866eb9e1638b3a75485c98c12b0d0c5)

PR #19354, opened 2024-11-12
feat(pg-cdc): init default value for cdc table columns

StrikeW:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://gist.github.com/TennyZhuang/f00be7f16996ea48effb049aa7be4d66#file-rw_cla).

## What's changed and what's your intention?

Read external db to init default value for cdc table columns

## Checklist

- [ ] I have written necessary rustdoc comments
- [ ] I have added necessary unit tests and integration tests
- [ ] I have added test labels as necessary. See [details](https://github.com/risingwavelabs/risingwave/blob/main/docs/developer-guide.md#ci-labels-guide).
- [ ] I have added fuzzing tests or opened an issue to track them. (Optional, recommended for new SQL features #7934).
- [ ] My PR contains breaking changes. (If it deprecates some features, please create a tracking issue to remove them in the future).
- [ ] All checks passed in `./risedev check` (or alias, `./risedev c`)
- [ ] My PR changes performance-critical code. (Please run macro/micro-benchmarks and show the results.)

- [ ] My PR contains critical fixes that are necessary to be merged into the latest release. (Please check out the [details](https://github.com/risingwavelabs/risingwave/blob/main/CONTRIBUTING.md))

## Documentation

- [ ] My PR needs documentation updates. (Please use the **Release note** section below to summarize the impact on users)

## Release note

If this PR includes changes that directly affect users or other significant modifications relevant to the community, kindly draft a release note to provide a concise summary of these changes. Please prioritize highlighting the impact these changes will have on users.

</details>

PR #20348, opened 2025-01-31
fix(cdc): fix wrong default column matching

KeXiangWang:

Fix the bug introduced in https://github.com/risingwavelabs/risingwave/pull/19354. Resolve https://github.com/risingwavelabs/risingwave/issues/20346
The bug will cause two problems:
1. Differences in the orders of rw schema and pg schema can lead to wrong default value assigned.
2. Due to `zip_eq_fast`, if the numbers of columns in rw schema and pg schema are different, panic happens.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/handler/create_table.rs
+++ b/src/frontend/src/handler/create_table.rs
@@ -1196,10 +1196,13 @@ pub(super) async fn handle_create_table_plan(
                         .collect();
 
                     for col in &mut columns {
-                        if let Some(external_col) = external_columns.get(col.name()) {
-                            col.column_desc.generated_or_default_column =
-                                external_col.column_desc.generated_or_default_column.clone();
-                        }
+                        let external_col = external_columns.get(col.name()).ok_or_else(|| {
+                            ErrorCode::ConnectorError(
+                                format!("Column {} not found in external table", col.name()).into(),
+                            )
+                        })?;
+                        col.column_desc.generated_or_default_column =
+                            external_col.column_desc.generated_or_default_column.clone();
                     }
                     (columns, pk_names)
                 }
--- a/src/frontend/src/handler/create_table.rs
+++ b/src/frontend/src/handler/create_table.rs
@@ -33,7 +33,6 @@ use risingwave_common::license::Feature;
 use risingwave_common::session_config::sink_decouple::SinkDecouple;
 use risingwave_common::system_param::reader::SystemParamsRead;
 use risingwave_common::types::DataType;
-use risingwave_common::util::iter_util::ZipEqFast;
 use risingwave_common::util::sort_util::{ColumnOrder, OrderType};
 use risingwave_common::util::value_encoding::DatumToProtoExt;
 use risingwave_common::{bail, bail_not_implemented};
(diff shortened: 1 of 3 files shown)


src/frontend/src/handler/create_table.rs
```diff
@@ -1182,20 +1181,25 @@ pub(super) async fn handle_create_table_plan(
...
-                        columns.iter_mut().zip_eq_fast(external_columns.into_iter())
-                    {
-                        col.column_desc.generated_or_default_column =
-                            external_col.column_desc.generated_or_default_column;
+
+                    for col in &mut columns {
+                        if let Some(external_col) = external_columns.get(col.name()) {
+                            col.column_desc.generated_or_default_column =
+                                external_col.column_desc.generated_or_default_column.clone();
+                        }
```
BugenZhao, line 1202: Is it possible that we don't find a column in `external_columns`? Should we return an error (just like in #20203)?

BugenZhao (commented):
Rest LGTM

src/frontend/src/handler/create_table.rs
KeXiangWang: I cannot find a case for now. But a potential case is https://github.com/risingwavelabs/risingwave/pull/19112/#issuecomment-2613712659. 
An example haven't been supported but to be supported:
```
CREATE TABLE test (
    id int,
    name varchar,
    my_id int as id + 1, // generated column
    PRIMARY KEY (id)
) FROM pg_source TABLE 'public.test';
```
I can make it throw an error for now.

KeXiangWang: Updated

KeXiangWang: The same logic can be found [here](https://github.com/risingwavelabs/risingwave/blob/08bdf2743469d7aab35c98403f53a3f4738b93f5/java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/PostgresValidator.java#L223), so it's OK not to throw an error, either.

KeXiangWang: Sorry, I just recheck the implementation. If we don't throw an error here. Generated columns can be then supported in tables on cdc source. Previously, only tables with connectors can support generated columns. So I reverted the codes back and now it won't throw errors in this case.

BugenZhao: > So I reverted the codes back and now it won't throw errors in this case.

Shall we instead assert it's a generated column if it's not found?

KeXiangWang: > Shall we instead assert it's a generated column if it's not found?

We can throw an error if a not-found column is not a generated column. 
It's also OK we don't do that, as it would also be checked in the [java code](https://github.com/risingwavelabs/risingwave/blob/08bdf2743469d7aab35c98403f53a3f4738b93f5/java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/PostgresValidator.java#L223).

KeXiangWang: > If we don't throw an error here. Generated columns can be then supported in tables on cdc source. 

And I found the implementation here was buggy. The newly inserted rows are not received in this type of table. I'll try fixing it soon. If the fix is non-trivial, I throw an error here and leave the fix for future.

KeXiangWang: > If the fix is non-trivial, I throw an error here and leave the fix for future.

I just reverted the codes. I'll open another PR to support generated columns in table on cdc source.

fuyufjh (approved):

st1page (approved):

Issue #20342, opened 2025-01-29
StreamChunk inconsistency bug: /risingwave/src/common/src/array/stream_chunk.rs:142:13 assertion left == right failed   left: 1   right: 0


kwannoel:
### Describe the bug

There's a bug inside StreamChunk which panics at:
```
/risingwave/src/common/src/array/stream_chunk.rs:142:13
assertion left == right failed
  left: 1
  right: 0
```

This is on RisingWave v2.1.2.

I have added a branch to attempt to debug it, by checking for inconsistency when constructing stream chunks: https://github.com/risingwavelabs/risingwave/pull/20341.

### Error message/log

```text

```

PR #20350, opened 2025-01-31
feat(stream): normalize unmatched updates

kwannoel:

Closes: https://github.com/risingwavelabs/risingwave/issues/20342

The bug is that stream_key and storage_key can mismatch for an MV, and can result in UpdateDelete + UpdateInsert pairs being split up.
```
dev=> explain create materialized view m1 as select * from t order by v2, v3;
                                                      QUERY PLAN                                                       
-----------------------------------------------------------------------------------------------------------------------
 StreamMaterialize { columns: [v1, v2, v3, v4, v5], stream_key: [v1], pk_columns: [v2, v3, v1], pk_conflict: NoCheck }
 └─StreamTableScan { table: t, columns: [v1, v2, v3, v4, v5] }
```

Details:
1. If upstream updates some record, it will be updated on some stream_key .
2. This propagates to the backfilling MV.
3. When receiving barrier, we will flush buffered upstream chunks. Here we have logic to remove rows which have storage_key > current_pos, where current_pos is the storage_key we have backfilled until.
4. This means that an UpdateDelete + UpdateInsert pair could be split up, because while their stream keys are the same, their storage keys might be different, leading to one being pruned, but not the other. An example is when upstream is an MV with ORDER BY on non-pk columns.
5. Subsequently, UpdateDelete gets propagated to dispatcher, and the dispatcher has some optimization to remove redundant UpdateDelete and UpdateInsert pairs. We may lose an op entry here in this code branch. Then op.len() != col.len(), and we will panic when reconstructing the chunk.

This bug is sort of worked around with https://github.com/risingwavelabs/risingwave/pull/20176/files. Since dispatcher will rewrite U+/U- going to different vnodes into +/-, and dist_key is set to order key. But this PR defensively handles it in backfill as well.

For unmatched update delete, i.e. storage_key <= current_pos, we can rewrite it to a normal delete. Because in the next snapshot read, we should read the new update insert.
For unmatched update insert, we have not backfilled the corresponding update delete key from storage yet. So there won’t be a conflict to just insert it. We can rewrite it to a normal insert.

We also reconstruct the ops lazily, since in most cases it should not be necessary, only when there' a hanging update.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/executor/backfill/utils.rs
+++ b/src/stream/src/executor/backfill/utils.rs
@@ -12,7 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::cmp::Ordering;
 use std::collections::HashMap;
 use std::ops::Bound;
 
@@ -30,7 +29,7 @@ use risingwave_common::types::{DataType, Datum};
 use risingwave_common::util::chunk_coalesce::DataChunkBuilder;
 use risingwave_common::util::epoch::EpochPair;
 use risingwave_common::util::iter_util::ZipEqDebug;
-use risingwave_common::util::sort_util::{cmp_datum_iter, OrderType};
+use risingwave_common::util::sort_util::{cmp_datum_iter_le, OrderType};
 use risingwave_common::util::value_encoding::BasicSerde;
 use risingwave_common_rate_limit::RateLimit;
 use risingwave_connector::error::ConnectorError;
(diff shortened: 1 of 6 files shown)


BugenZhao:
> 1. If upstream updates some record, it will be updated on some stream_key .

This reminds me (again) of the discussion in (https://github.com/risingwavelabs/risingwave/issues/12539. In my opinion, if maintaining the `U-` and `U+` invariants proves more challenging than beneficial, should we consider only retaining `-` and `+` in the system?

src/stream/src/executor/backfill/utils.rs
```diff
@@ -341,31 +340,41 @@ pub(crate) fn mark_chunk_ref_by_vnode<S: StateStore, SD: ValueRowSerde>(
...
             BackfillProgressPerVnode::Completed { .. } => true,
             // If not started, no need to forward.
             BackfillProgressPerVnode::NotStarted => false,
             // If in progress, we need to check row <= current_pos.
             BackfillProgressPerVnode::InProgress { current_pos, .. } => {
-                match cmp_datum_iter(pk.iter(), current_pos.iter(), pk_order.iter().copied()) {
-                    Ordering::Less | Ordering::Equal => true,
-                    Ordering::Greater => false,
-                }
+                cmp_datum_iter_le(pk.iter(), current_pos.iter(), pk_order.iter().copied())
```
BugenZhao, line 357:
```suggestion
                cmp_datum_iter(pk.iter(), current_pos.iter(), pk_order.iter().copied()).is_le()
```

src/stream/src/executor/backfill/utils.rs
```diff
@@ -379,20 +388,93 @@ fn mark_chunk_inner(
...
+    current_visibility: bool,
+    current_op_index: usize,
+    current_op: &Op,
+) {
+    if *unmatched_update_delete {
+        assert_eq!(*current_op, Op::UpdateInsert);
+        let visible_update_insert = current_visibility;
+        match (visible_update_delete, visible_update_insert) {
+            (true, false) => {
+                // Lazily clone the ops here.
```
BugenZhao, line 438: Can we simplify this with `Cow::to_mut`?

chenzl25 (approved):
The fix LGTM, thanks!

kwannoel:
> > 1. If upstream updates some record, it will be updated on some stream_key .
> 
> This reminds me (again) of the discussion in (https://github.com/risingwavelabs/risingwave/issues/12539. In my opinion, if maintaining the `U-` and `U+` invariants proves more challenging than beneficial, should we consider only retaining `-` and `+` in the system (instead of patching or rewriting everywhere)?

Generally agree. We may lose some traceability, but we will gain increased performance and code simplicity. We can now encode ops in a bitmap.

kwannoel:
* **#20350** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20350?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20350?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #20352, opened 2025-01-31
feat: allow pass-by source enumerator check

tabVersion:

introduce a debug-mode-only attr `debug_splits`, it will return the given splits instead of requesting the broker. \

an example here

```
# producing 101 messages to the topic
for i in {0..100}; do echo "{\"a\": $i}^{\"a\": $i, \"b\": \"$i\"}" | kcat -P -b 127.0.0.1:9092 -t test_1 -K ^ -H "header1=v1" -H "header2=v2"; done

--- psql ---

create table t2 ( a int, b varchar ) with ( connector = 'kafka', properties.bootstrap.server = '127.0.0.1:9092', topic = 'test_1' , debug_splits = '[{"split_info": {"partition": 0, "start_offset": -1, "stop_offset": null, "topic": "test_1"}, "split_type": "kafka"}]') format plain encode json ;

dev=> select count(*) from t2 ;
 count
-------
    50
(1 row)
```

## Checklist

## Documentation

it is for dev use only, no need for doc change

<details>

</details>


src/meta/src/stream/source_manager/worker.rs
```diff
@@ -12,13 +12,17 @@
...
+#[cfg(not(debug_assertions))]
+use risingwave_connector::error::ConnectorError;
 use risingwave_connector::source::AnySplitEnumerator;
 
 use super::*;
 
 const MAX_FAIL_CNT: u32 = 10;
 const DEFAULT_SOURCE_TICK_TIMEOUT: Duration = Duration::from_secs(10);
 
+const DEBUG_SPLITS_KEY: &str = "debug_splits";
```
fuyufjh, line 24: Recommend adding some comments in code

fuyufjh (approved):

Issue #19783, opened 2024-12-12
bug: segmentation fault on accessing a dangling channel when sending error from debezium


BugenZhao:
### Describe the bug

JVM thread running Debezium engine may access dangling channel sender in its `completionCallback`, leading to segmentation fault.

### Error message/log

```text
2024-12-12T16:49:21.062787+08:00 ERROR risingwave_connector_node: engine#16 terminated with error. message: Stopping connector after error in the application's handler method: null: java.lang.InterruptedException
	at java.base/java.util.concurrent.locks.AbstractQueuedSynchronizer.acquireInterruptibly(AbstractQueuedSynchronizer.java:1261)
	at java.base/java.util.concurrent.locks.ReentrantLock.lockInterruptibly(ReentrantLock.java:317)
	at java.base/java.util.concurrent.ArrayBlockingQueue.put(ArrayBlockingQueue.java:364)
	at com.risingwave.connector.source.core.DbzChangeEventConsumer.handleBatch(DbzChangeEventConsumer.java:293)
	at io.debezium.embedded.ConvertingEngineBuilder$ConvertingChangeConsumer.handleBatch(ConvertingEngineBuilder.java:75)
	at io.debezium.embedded.EmbeddedEngine.pollRecords(EmbeddedEngine.java:727)
	at io.debezium.embedded.EmbeddedEngine.run(EmbeddedEngine.java:466)
	at io.debezium.embedded.ConvertingEngineBuilder$1.run(ConvertingEngineBuilder.java:163)
	at com.risingwave.connector.source.core.DbzCdcEngine.run(DbzCdcEngine.java:67)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:829)
 thread="rw-dbz-engine-runner-16" class="com.risingwave.connector.source.core.DbzCdcEngineRunner"

[PROCESS EXIT WITH CODE 139]
```

### To Reproduce

I first encountered this bug in https://github.com/risingwavelabs/risingwave/pull/19595#issuecomment-2511159913. It appears that using the same MySQL instance for both CDC tests and the meta store consistently reproduces the issue. Although it's not likely the case in production, it clearly shows that there's unsafe or unsoundness in our JNI code. 

[...]

PR #20353, opened 2025-01-31
fix(cdc): pass ownership of source message sender to java thread

BugenZhao:

Pass the ownership of Debezium CDC source message sender to the Java thread by creating a wrapper of it. Use `Cleaner` in the Java code to release memory when it's no longer needed.

Previously there was memory unsafety: https://github.com/risingwavelabs/risingwave/issues/19783.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/jni_core/src/lib.rs
+++ b/src/jni_core/src/lib.rs
@@ -1068,8 +1068,8 @@ extern "system" fn Java_com_risingwave_java_binding_Binding_sendCdcSourceErrorTo
 }
 
 #[no_mangle]
-extern "system" fn Java_com_risingwave_java_binding_Binding_cdcSourceSenderClose<'a>(
-    _env: EnvParam<'a>,
+extern "system" fn Java_com_risingwave_java_binding_Binding_cdcSourceSenderClose(
+    _env: EnvParam<'_>,
     channel: OwnedPointer<JniSenderType<GetEventStreamResponse>>,
 ) {
     channel.release();
--- a/src/jni_core/src/lib.rs
+++ b/src/jni_core/src/lib.rs
@@ -151,6 +151,10 @@ impl<T> From<T> for Pointer<'static, T> {
 }
 
 impl<'a, T> Pointer<'a, T> {
+    pub fn pointer(&self) -> jlong {
+        self.pointer
+    }
+
     fn as_ref(&self) -> &'a T {
         debug_assert!(self.pointer != 0);
         unsafe { &*(self.pointer as *const T) }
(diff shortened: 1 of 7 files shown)


src/jni_core/src/lib.rs
```diff
@@ -1052,6 +1067,14 @@ extern "system" fn Java_com_risingwave_java_binding_Binding_sendCdcSourceErrorTo
     })
 }
 
+#[no_mangle]
+extern "system" fn Java_com_risingwave_java_binding_Binding_cdcSourceSenderClose<'a>(
+    _env: EnvParam<'a>,
```
BugenZhao, line 1072:
```suggestion
extern "system" fn Java_com_risingwave_java_binding_Binding_cdcSourceSenderClose(
    _env: EnvParam<'_>,
```

fuyufjh (approved):
LGTM, thanks for the fix!

src/jni_core/src/lib.rs
```diff
@@ -152,21 +152,36 @@ impl<T> From<T> for Pointer<'static, T> {
...
+        self.pointer
+    }
+
+    /// Release the object behind the pointer.
+    fn release(self) {
+        tracing::debug!(
+            type_name = std::any::type_name::<T>(),
+            address = self.pointer,
+            "release jni OwnedPointer"
+        );
```
BugenZhao, line 183: I'm not able to observe this line in the log. Not sure if I'm using `ref.Cleaner` correctly...

wenym1 (approved):
LGTM!

src/jni_core/src/lib.rs
wenym1: Not sure if it's because memory is enough and GC is not triggered. You can try running `jcmd <pid> GC.run` to trigger the gc.

chenzl25 (approved):
LGTM！

src/jni_core/src/lib.rs
BugenZhao: I realized that the channel is considered reachable by Java even if the Debezium runner is stopped, since we forget to call `JniDbzSourceRegistry.unregister` in the normal path. Add a commit to move it into the `finally` block and triggering GC with `jcmd` now successfully releases the memory.

```
2025-02-11T17:53:06.507871+08:00 DEBUG risingwave_jni_core: release jni OwnedPointer type_name="tokio::sync::mpsc::bounded::Sender<core::result::Result<risingwave_pb::connector_service::GetEventStreamResponse, anyhow::Error>>" address=5795941504
```

image

PR #19935, opened 2024-12-25
feat: `message_as_jsonb` to handle circle dep in protobuf

tabVersion:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://gist.github.com/TennyZhuang/f00be7f16996ea48effb049aa7be4d66#file-rw_cla).

## What's changed and what's your intention?

resolve https://github.com/risingwavelabs/risingwave/issues/19915
following https://github.com/risingwavelabs/risingwave/pull/18380

## Checklist

- [ ] I have written necessary rustdoc comments
- [ ] I have added necessary unit tests and integration tests
- [ ] I have added test labels as necessary. See [details](https://github.com/risingwavelabs/risingwave/blob/main/docs/developer-guide.md#ci-labels-guide).
- [ ] I have added fuzzing tests or opened an issue to track them. (Optional, recommended for new SQL features #7934).
- [ ] My PR contains breaking changes. (If it deprecates some features, please create a tracking issue to remove them in the future).
- [ ] All checks passed in `./risedev check` (or alias, `./risedev c`)
- [ ] My PR changes performance-critical code. (Please run macro/micro-benchmarks and show the results.)

- [ ] My PR contains critical fixes that are necessary to be merged into the latest release. (Please check out the [details](https://github.com/risingwavelabs/risingwave/blob/main/CONTRIBUTING.md))

## Documentation

- [x] My PR needs documentation updates. (Please use the **Release note** section below to summarize the impact on users)

## Release note

If this PR includes changes that directly affect users or other significant modifications relevant to the community, kindly draft a release note to provide a concise summary of these changes. Please prioritize highlighting the impact these changes will have on users.

add a new field `messages_as_jsonb` dedicated to `encode protobuf`, it lists fields should be seen as jsonb type and we can leverage this to handle recursive types

for example,

```
message A {
  int32 int_value,
  string string_value,
}
```

If not set in `messages_as_jsonb`, the type in RW should be `A struct<int_value int, string_value varchar>`.
If `messages_as_jsonb='some_pkg.A'`, the type in RW should be `A jsonb`. 

---

When finding a recursive def in proto, we will display the circular dep relation and prevent creating the source/table. 

Users can refer the error message and adding the types to `messages_as_jsonb`

```
[...]

PR #20401, opened 2025-02-06
feat(sink): resolve avro `Ref` to corresponding `Record` definition

xiangjinwu:

This is the sink analogy for #19601 and #19746. Comparing the source and sink support:
* The sink did not support any form of `Ref` previously, while the source supported a subset (single-layer) via a hack.
* The sink unifies schema and datum handling via `MaybeData`, while the source took 2 PRs to fix separately.
* The sink uses an owned newtype `NamesRef(HashMap<Name, AvroSchema>)` while the source uses `apache_avro::schema::NamesRef<'s> = HashMap<Name, &'s AvroSchema>`. The former requires more memory while the latter requires more time to resolve every time. We can unify to one approach once we learned which is better in practice.
* The sink does not need to reject circular references as source did, because the upstream SQL schema is never circular.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/src/sink/encoder/avro.rs
+++ b/src/connector/src/sink/encoder/avro.rs
@@ -12,9 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::collections::HashMap;
 use std::sync::Arc;
 
-use apache_avro::schema::{RecordSchema, Schema as AvroSchema};
+use apache_avro::schema::{Name, RecordSchema, Schema as AvroSchema};
 use apache_avro::types::{Record, Value};
 use risingwave_common::catalog::Schema;
 use risingwave_common::row::Row;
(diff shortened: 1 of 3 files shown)


src/connector/src/sink/encoder/avro.rs
```diff
@@ -395,6 +435,8 @@ fn on_field<D: MaybeData>(data_type: &DataType, maybe: D, expected: &AvroSchema)
         _ => (expected, OptIdx::NotUnion),
     };
 
+    let inner = refs.lookup(inner);
```
xiangjinwu, line 438: `Ref` never resolves to a `Union`, while a `Union` may contain a `Ref`. As a result we check for nullable `Union` first and then resolve `Ref`. General (instead of nullability) `Union` is not supported in sink yet.

ci/workflows/pull-request.yml
```diff
@@ -114,7 +114,7 @@ steps:
     plugins:
       - docker-compose#v5.5.0: *docker-compose
       - ./ci/plugins/upload-failure-logs
-    timeout_in_minutes: 30
+    timeout_in_minutes: 35
```
xiangjinwu, line 117: https://buildkite.com/risingwavelabs/pull-request/builds/67521

chenzl25 (approved):
LGTM. Shall we support circular reference in source in the future?

xiangjinwu:
> LGTM. Shall we support circular reference in source in the future?

There is no corresponding SQL data type for such circular reference (linked list, tree, etc). If JSON is good enough, there is `message_as_jsonb` for protobuf (but not avro due to its poor rust library). #19935

PR #20402, opened 2025-02-06
feat(sink): support geospatial for redis sink

xxhZs:

in this pr, we support geospatial type for redis sink
https://redis.io/glossary/geospatial-indexing/

## Checklist

## Documentation

We can use geospatial like 

```
create table t1(v1 float, v2 float,v3 varchar, v4 varchar);
CREATE SINK s1
FROM
    t1 WITH (
    primary_key = 'v3,v4',
    connector = 'redis',
    redis.url= 'redis://127.0.0.1:6379/',
)FORMAT UPSERT ENCODE TEMPLATE
(redis_value_type ='geospatial', 
longitude = 'v1', 
latitude = 'v2', 
member = 'v3',
key_format = 'abcd3:{v4}'
);
insert into t1 values(1.1, 1.1, 'test','test');
```
then we can get it in redis like 
```
GEOPOS abcd3:test test
------
1) 1) "1.10000044298171997"
   2) "1.10000000482478555"
```

`redis_value_type` must be set to `string` and `geospatial`, default is `string`
if `redis_value_type` is set to `string`, we must set `key_format` and `value_fomat`(Consistent with previous behavior)
if `redis_value_type` is set to `geospatial`, we must set `key_format`, `longitude`, `latitude` and `member` 
The meaning of `key_format` is consistent with `redis_value_type = string`
The `longitude` is the column name of rw, the datatype must be float, real or varchar
The `latitude` is the column name of rw, the datatype must be float, real or varchar
The `member` is the column name of rw, the datatype must be varchar, and must be `primary_key`

---- 

rewrite the release note (@tabVersion) 

* New Features
  * Added support for Redis geospatial data type in sink connectors
  * Introduced new configuration parameter redis_value_type with options:
    * string (default): Traditional key-value storage
    * geospatial: Store location data using Redis' geospatial features
* Configuration Requirements
  * For geospatial storage:
    * Required parameters: key_format, longitude, latitude, and member
    * longitude and latitude columns must be float, real, or varchar type
    * member column must be varchar type and part of the primary key
  * For string storage (unchanged):
    * Required parameters: key_format and value_format

### DIFF
--- a/src/connector/src/sink/encoder/template.rs
+++ b/src/connector/src/sink/encoder/template.rs
@@ -12,8 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use core::any::Any;
-use std::collections::{HashMap, HashSet};
+use std::collections::HashMap;
 
 use regex::Regex;
 use risingwave_common::catalog::Schema;
--- a/src/connector/src/sink/encoder/template.rs
+++ b/src/connector/src/sink/encoder/template.rs
@@ -32,6 +32,10 @@ impl TemplateEncoder {
     pub fn new_string(schema: Schema, col_indices: Option<Vec<usize>>, template: String) -> Self {
         TemplateEncoder::String(TemplateStringEncoder::new(schema, col_indices, template))
     }
+
+    pub fn new_geo(schema: Schema, col_indices: Option<Vec<usize>>, lat_name: &str, lon_name: &str, member_name: &str) -> Result<Self> {
+        Ok(TemplateEncoder::RedisGeo(TemplateRedisGeoEncoder::new(schema, col_indices, lat_name, lon_name, member_name)?))
+    }
 }
 impl RowEncoder for TemplateEncoder {
     type Output = TemplateEncoderOutput;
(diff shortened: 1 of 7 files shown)


src/connector/src/sink/encoder/template.rs
```diff
@@ -95,3 +150,181 @@ impl RowEncoder for TemplateEncoder {
...
+            col_indices,
+            lat_col,
+            lon_col,
+        })
+    }
+
+    pub fn encode_cols(
+        &self,
+        row: impl Row,
+        _col_indices: impl Iterator<Item = usize>,
```
fuyufjh, line 193: Perhaps off-the-topic but the trait function is confusing to me: why `col_indices` need to be passed from outside, while it's stored in `self`? 😂

fuyufjh (approved):
LGTM

src/connector/src/sink/encoder/template.rs
xxhZs: cc @xiangjinwu

tabVersion:
just curious, for `longitude` and `latitude`, supporting float and real types should be enough, why string type?

xxhZs:
> just curious, for `longitude` and `latitude`, supporting float and real types should be enough, why string type?

Just because redis supports string and float, it's consistent here

PR #20403, opened 2025-02-06
feat(udf): introduce async and batched JavaScript UDF

### LINKED ISSUE #20391: feat: Async JavaScript UDF and Fetch API
### Background

As part of [RFC: Promoting Embedded UDF](https://risingwave-labs.notion.site/RFC-Promoting-Embedded-UDF-11af0d69cb76808c88e0f8d4c64492c3), we are going to allow UDFs to call external resources. 

After some research, I found JavaScript to be a good fit for this use case. The reasons are:

1. JavaScript is inherently designed for the web and comes with a powerful `fetch()` function for API calls (especially JSON APIs), whereas other languages often require third-party libraries.
2. JavaScript's event-loop-based asynchronous execution allows the JS runtime to integrate well with our Rust async runtime (`tokio`), avoiding concerns such as I/O blocking.

### Design

Here, let's focus on the RisingWave part. For the internal implementation, please refer to https://github.com/arrow-udf/arrow-udf/issues/95.

First, we need to change the syntax: 

Currently, the JavaScript UDF is defined with *function body **only***, e.g.

```sql
create function decimal_add(a decimal, b decimal) returns decimal language javascript as $$
    return a + b;
$$;
```

which makes it impossible to write an `async function`. We need to change it to

```diff
 create function decimal_add(a decimal, b decimal) returns decimal language javascript as $$
+export function decimal_add(a, b) {
     return a + b;
+}
$$;
```

This is also consistent with [embedded Python UDF](https://docs.risingwave.com/sql/udfs/embedded-python-udfs) and [Rust (WASM) UDF](https://docs.risingwave.com/sql/udfs/use-udfs-in-rust).  

Then, for the async JS functions, we introduce a extension: `language javascript async`. Example:

```sql
create function my_udf(s varchar) returns decimal language javascript async as $$
export async function my_udf(s) {
    ... // able to call async functions such as `await fetch()` here
}
$$;
```

The reason why we don't detect async functions from function definition is explained [here](https://github.com/arrow-udf/arrow-udf/pull/97#discussion_r1924987785).

### Future Optimizations

_No response_

### Discussions

_No response_

### Q&A

_No response_

fuyufjh:

This PR adds support for async and batched JavaScript UDF as well as `fetch()` API. Resolves #20391.

Syntax:

```sql
create function my_udf(s varchar) returns decimal language javascript async as $$
export async function my_udf(s) {
    ... // able to call async functions such as `await fetch()` here
}
$$ WITH (
   async = [true|false],
   batch = [true|false],
)
```

Notice that UDAF is **not** supported in this PR. You can confirm this by checking the code of frontend `handle_create_aggregate()` doesn't support the new syntax. However, the internal implementation is mostly ready except [here](https://github.com/risingwavelabs/risingwave/blob/dfd20c611cbffd2f5b218d85a15fb9cec5d91a85/src/expr/core/src/aggregate/user_defined.rs#L45).

## Checklist

## Documentation

<details>

</details>


Cargo.toml
```diff
@@ -164,7 +164,7 @@ iceberg-catalog-glue = { git = "https://github.com/risingwavelabs/iceberg-rust.g
 opendal = "0.49"
 # used only by arrow-udf-flight
 arrow-flight = "53"
-arrow-udf-js = "0.5"
+arrow-udf-js = { git = "https://github.com/arrow-udf/arrow-udf.git", rev = "4562678944210da2b3a094a6383f7b9ebe6cc54c" }
```
fuyufjh, line 167: Wait for the release of `arrow-udf-js` before this PR merges.

src/expr/core/src/aggregate/user_defined.rs
```diff
@@ -42,8 +42,11 @@ impl AggregateFunction for UserDefinedAggregateFunction {
 
     /// Creates an initial state of the aggregate function.
     fn create_state(&self) -> Result<AggregateState> {
-        let state = self.runtime.call_agg_create_state()?;
-        Ok(AggregateState::Any(Box::new(State(state))))
+        // FIXME(eric): This is bad. Let's make `create_state` async later
+        futures::executor::block_on(async {
+            let state = self.runtime.call_agg_create_state().await?;
+            Ok(AggregateState::Any(Box::new(State(state))))
+        })
```
stdrc, line 49: Since we don't support `async function create_state` for now, what about using `now_or_never` instead of `block_on` here?

src/expr/core/src/sig/udf.rs
```diff
@@ -158,7 +159,7 @@ pub trait UdfImpl: std::fmt::Debug + Send + Sync {
     /// Return the memory size consumed by UDF runtime in bytes.
     ///
     /// If not available, return 0.
-    fn memory_usage(&self) -> usize {
+    async fn memory_usage(&self) -> usize {
```
stdrc, line 162: Is this necessary?

src/expr/impl/src/udf/quickjs.rs
```diff
@@ -32,47 +33,68 @@ static QUICKJS: UdfImplDescriptor = UdfImplDescriptor {
...
-        } else {
-            let res = runtime.add_function(
-                opts.identifier,
-                UdfArrowConvert::default().to_arrow_field("", opts.return_type)?,
-                CallMode::CalledOnNullInput,
-                opts.body.context("body is required")?,
-            );
+        // NOTE: Some function calls such as `add_function()` requires async.
+        // However, since the `Runtime` here is not shared, the async block will never block.
+        futures::executor::block_on(async {
```
stdrc, line 38: What about `now_or_never().expect()`?

src/expr/impl/src/udf/quickjs.rs
```diff
@@ -85,37 +107,39 @@ struct QuickJsFunction {
...
-        self.runtime.finish(&self.identifier, state)
+    async fn call_agg_finish(&self, state: &ArrayRef) -> Result<ArrayRef> {
+        self.runtime.finish(&self.identifier, state).await
     }
 
-    fn memory_usage(&self) -> usize {
-        self.runtime.memory_usage().malloc_size as usize
+    async fn memory_usage(&self) -> usize {
+        self.runtime.inner().memory_usage().await.malloc_size as usize
     }
```
stdrc, line 144: I'm wondering if we can do `self.runtime.inner().memory_usage().now_or_never().unwrap_or_default()` here. Making `fn memory_usage` async looks strange to me.

src/expr/impl/src/udf/quickjs.rs
fuyufjh: IIUC, it can work only if the internal implementation of these async functions return a ready promise. I don't want to rely on such assumption.

src/expr/impl/src/udf/quickjs.rs
fuyufjh: Yeah, it's stupid. It's just for `self.inner.lock().await`, which is introduced by the Rust binding (`rquickjs`) rather than `quickjs`.

For the same reason of https://github.com/risingwavelabs/risingwave/pull/20403#discussion_r1948715142, I tend to use `block_on` here; and because JS Runtime is never shared, the function must never block.

src/expr/impl/src/udf/quickjs.rs
```diff
@@ -85,37 +114,43 @@ struct QuickJsFunction {
 #[async_trait::async_trait]
 impl UdfImpl for QuickJsFunction {
     async fn call(&self, input: &RecordBatch) -> Result<RecordBatch> {
-        self.runtime.call(&self.identifier, input)
+        // TODO(eric): if not batched, call JS function row by row. Otherwise, one row failure will fail the entire chunk.
```
stdrc, line 117: I don't quite get this line. IIUC, the current status is to call JS function row by row if not batched, and failure of one row will fail the chunk, right? What's the todo mean?

stdrc (approved):
Rest LGTM!

src/expr/impl/src/udf/quickjs.rs
BugenZhao: I think it's exactly the issue described in https://github.com/risingwavelabs/risingwave/issues/20473?

e2e_test/udf/js_async_udf.slt
```diff
@@ -0,0 +1,48 @@
...
+----
+42
+
+statement ok
+drop function int_42;
+
+statement ok
+create function fetch_todo(id int) returns varchar language javascript as $$
+export async function fetch_todo(id) {
+    const response = await fetch('https://dummyjson.com/todos/' + id);
```
MrCroxx, line 19: A lesson a day. 🤩

src/expr/core/src/aggregate/user_defined.rs
```diff
@@ -42,8 +42,11 @@ impl AggregateFunction for UserDefinedAggregateFunction {
 
     /// Creates an initial state of the aggregate function.
     fn create_state(&self) -> Result<AggregateState> {
-        let state = self.runtime.call_agg_create_state()?;
-        Ok(AggregateState::Any(Box::new(State(state))))
+        // FIXME(eric): This is bad. Let's make `create_state` async if someday we allow async UDAF
+        futures::executor::block_on(async {
```
MrCroxx, line 46: Is `spawn_blocking` acceptable here?

fuyufjh: It's not in used yet, because the syntax `create aggregate with (async = true)` is not supported yet.

ci/scripts/common.sh
```diff
@@ -105,7 +105,7 @@ function download_and_prepare_rw() {
   echo -e "\033[33mPrepare RiseDev dev cluster\033[0m"
 
   risedev pre-start-dev
-  risedev --allow-private link-all-in-one-binaries
+  risedev link-all-in-one-binaries
```
BugenZhao, line 108: Will this work?

src/sqlparser/src/ast/mod.rs
```diff
@@ -3240,48 +3240,61 @@ impl fmt::Display for CreateFunctionBody {
...
         }
         let mut options = vec![];
-        if let Some(always_retry_on_network_error) = self.always_retry_on_network_error {
-            options.push(format!(
-                "ALWAYS_RETRY_NETWORK_ERRORS = {}",
-                always_retry_on_network_error
-            ));
+        if let Some(v) = self.always_retry_on_network_error {
+            options.push(format!("always_retry_on_network_error = {}", v));
+        }
```
BugenZhao, line 3292: @xiangjinwu This unparsing bug will finally be addressed. 😄

src/sqlparser/src/ast/mod.rs
```diff
@@ -3240,48 +3240,61 @@ impl fmt::Display for CreateFunctionBody {
...
+                        SqlOptionValue::Value(Value::Boolean(true))
+                    ));
+                }
+                "async" => {
+                    options.r#async = Some(matches!(
+                        option.value,
+                        SqlOptionValue::Value(Value::Boolean(true))
+                    ))
+                }
+                "batch" => {
```
BugenZhao, line 3269: Shall we name it `batched` instead? 🤔

BugenZhao (approved):

src/sqlparser/src/ast/mod.rs
stdrc: We also use `batch=True` in external python udf

ci/scripts/common.sh
stdrc: This is to avoid a warning complaining duplicated `--allow-private`

PR #20440, opened 2025-02-10
refactor(log-store): refine synced log store executor implementation

wenym1:

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/executor/sync_kv_log_store.rs
+++ b/src/stream/src/executor/sync_kv_log_store.rs
@@ -158,6 +158,11 @@ enum WriteFuture<S: LocalStateStore> {
     Empty,
 }
 
+enum WriteFutureEvent {
+    UpstreamMessageReceived(Message),
+    ChunkFlushed(FlushedChunkInfo),
+}
+
 impl<S: LocalStateStore> WriteFuture<S> {
     fn flush_chunk(
         stream: BoxedMessageStream,
@@ -193,11 +198,7 @@ impl<S: LocalStateStore> WriteFuture<S> {
 
     async fn next_event(
         &mut self,
-    ) -> StreamExecutorResult<(
-        BoxedMessageStream,
-        LogStoreWriteState<S>,
-        Either<Message, FlushedChunkInfo>,
-    )> {
+    ) -> StreamExecutorResult<(BoxedMessageStream, LogStoreWriteState<S>, WriteFutureEvent)> {
         match self {
             WriteFuture::ReceiveFromUpstream { future, .. } => {
                 let (opt, stream) = future.await;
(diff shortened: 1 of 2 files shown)


src/stream/src/executor/sync_kv_log_store.rs
```diff
@@ -160,271 +255,252 @@ impl<S: StateStore> SyncedKvLogStoreExecutor<S> {
...
-        };
-        match item {
-            LogStoreBufferItem::StreamChunk {
-                chunk, end_seq_id, ..
-            } => {
-                truncation_offset.replace((item_epoch, Some(end_seq_id)));
-                Ok(Some(chunk))
+        match self {
+            ReadFuture::ReadingPersistedStream(_) => {
+                unreachable!("must be Idle when reaching here")
```
kwannoel, line 434:
```suggestion
                unreachable!("must be Idle or reading flushed chunk when reaching here")
```

src/stream/src/executor/sync_kv_log_store.rs
```diff
@@ -160,271 +255,252 @@ impl<S: StateStore> SyncedKvLogStoreExecutor<S> {
...
-        let Some((item_epoch, item)) = buffer.pop_front() else {
-            return Ok(None);
-        };
-        match item {
-            LogStoreBufferItem::StreamChunk {
-                chunk, end_seq_id, ..
-            } => {
-                truncation_offset.replace((item_epoch, Some(end_seq_id)));
-                Ok(Some(chunk))
+        match self {
```
kwannoel, line 432: I thought the sequence should be:
1. ReadingPersistedStream
2. Reading Flushed Chunk Future
3. Reading from buffer.

Seems like we only have 1 and 3

src/stream/src/executor/sync_kv_log_store.rs
```diff
@@ -160,271 +255,252 @@ impl<S: StateStore> SyncedKvLogStoreExecutor<S> {
...
+                    let output = select(write_future, read_future).await;
+                    drop_either_future(output)
+                };
+                match select_result {
+                    Either::Left(result) => {
+                        // drop the future to ensure that the future must be reset later
+                        drop(write_future);
+                        let (stream, mut write_state, either) = result?;
+                        match either {
+                            Either::Left(message) => {
```
kwannoel, line 300: Nit: Not very readable to have two `Either::Left`. Prefer to explicitly encode as an enum. e.g. `WriteFutureEvent`.

kwannoel (commented):
Rest LGTM, only the read future needs some clarification.

src/stream/src/executor/sync_kv_log_store.rs
kwannoel: Oh ic. They are independently matched. So 2&3 won't both occur.

The sequence in code is like 1, 3, 2.

kwannoel: But 3 and 2 are mutually exclusive.

kwannoel (approved):
LGTM, thanks!

Issue #16841, opened 2024-05-20
split connector implementations into crates


xxchan:
BTW, I think it's time to split connector implementations off and design a clear interface for it, as there are now community developers willing to develop new connectors. 🤔

_Originally posted by @wangrunji0408 in https://github.com/risingwavelabs/risingwave/pull/16759#discussion_r1604355825_

xxchan:
From the diff in this PR, it seems the boundary is quite clear and it should be easy to split them out. https://github.com/risingwavelabs/risingwave/pull/17154/

The question is just which plan do we choose: feature flags? separated crates?

PR #16759, opened 2024-05-15
refactor(udf): move UDF implementations to `expr_impl`

wangrunji0408:
Signed-off-by: Runji Wang <wangrunji0408@163.com>I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://gist.github.com/TennyZhuang/f00be7f16996ea48effb049aa7be4d66#file-rw_cla).

## What's changed and what's your intention?

This PR moves UDF implementations from `expr_core` to `expr_impl` and introduces a `distributed_slice` in `expr_core` to collect them at link time. It makes it easier for developers to add new UDF languages. It also moves a lot dependencies out of `expr_core`, which can reduce build time and improve DX.

This PR makes each UDF a feature and they are not enabled by default to avoid building a lot of dependencies. For those who need to develop UDF, please enable them from `./risedev configure`.

## Checklist

- [ ] I have written necessary rustdoc comments
- [ ] I have added necessary unit tests and integration tests
- [ ] I have added test labels as necessary. See [details](https://github.com/risingwavelabs/risingwave/blob/main/docs/developer-guide.md#ci-labels-guide).
- [ ] I have added fuzzing tests or opened an issue to track them. (Optional, recommended for new SQL features #7934).
- [ ] My PR contains breaking changes. (If it deprecates some features, please create a tracking issue to remove them in the future).
- [ ] All checks passed in `./risedev check` (or alias, `./risedev c`)
- [ ] My PR changes performance-critical code. (Please run macro/micro-benchmarks and show the results.)

- [ ] My PR contains critical fixes that are necessary to be merged into the latest release. (Please check out the [details](https://github.com/risingwavelabs/risingwave/blob/main/CONTRIBUTING.md))

## Documentation

- [ ] My PR needs documentation updates. (Please use the **Release note** section below to summarize the impact on users)

## Release note

If this PR includes changes that directly affect users or other significant modifications relevant to the community, kindly draft a release note to provide a concise summary of these changes. Please prioritize highlighting the impact these changes will have on users.

</details>

PR #20441, opened 2025-02-10
chore: correct default features to avoid recompiling

fuyufjh:

Currently, `risedev` uses different feature sets from default `cargo build`:

```
[cargo-make] INFO - Execute Command: "/usr/bin/env" "bash" "/var/folders/57/7b2ql1gs0kg9j0b4pv041p3c0000gn/T/fsio_Ywk8VKIhIZ.sh"
+ RUSTFLAGS=<not set>
+ cargo build -p risingwave_cmd_all --profile dev --features rw-static-link --features external-udf --features wasm-udf --features js-udf --features python-udf --timings
```

which caused the `cargo check/build/clippy...` to recompile, consuming more time and disk space. 

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/cmd_all/Cargo.toml
+++ b/src/cmd_all/Cargo.toml
@@ -8,7 +8,7 @@ license = { workspace = true }
 repository = { workspace = true }
 
 [features]
-default = ["rw-static-link"]
+default = ["rw-static-link", "all-udf"]
 rw-static-link = ["workspace-config/rw-static-link"]
 rw-dynamic-link = ["workspace-config/rw-dynamic-link"]
 openssl-vendored = ["workspace-config/openssl-vendored"]


xxchan (commented):
The feature flags of risedev is configured via `risedev configure`. IIRC it's off by default.

fuyufjh:
> The UDF feature flags of risedev is configured via `risedev configure`. IIRC it's off by default.

Yeah, indeed. My local `risedev` was configured to enable these options, although I have forgotten at all.

On the other hand, I doubt whether we need these `features`. We run all test and release flow with all features enabled, so I tend to use the same build config in my local dev env as well.

xxchan:
Background of the feature flags:
https://github.com/risingwavelabs/risingwave/pull/16759#discussion_r1604349219
> My consideration is that each UDF backend brings many dependencies, and most of our developers don't touch UDF at all. So it would be better to skip them in local development. The same thing could be applied to connectors. For example, I almost never use any connector in my development. Those dependencies could be skipped as well to save my disk.

Actually we also wanted to add flags for connectors: https://github.com/risingwavelabs/risingwave/issues/16841

But maybe we could change to opt-out (enable by default) instead of opt-in

stdrc:
> But maybe we could change to opt-out (enable by default) instead of opt-in

+1 for opt-out

PR #20444, opened 2025-02-11
feat(cdc): support mapping json of pg to jsonb in rw in pg-cdc

KeXiangWang:

As titled

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/common/src/types/jsonb.rs
+++ b/src/common/src/types/jsonb.rs
@@ -589,17 +589,40 @@ impl ToSql for JsonbVal {
 
 impl<'a> FromSql<'a> for JsonbVal {
     fn from_sql(
-        _ty: &Type,
+        ty: &Type,
         mut raw: &'a [u8],
     ) -> Result<Self, Box<dyn std::error::Error + Sync + Send>> {
-        if raw.is_empty() || raw.get_u8() != 1 {
-            return Err("invalid jsonb encoding".into());
-        }
-        Ok(JsonbVal::from(Value::from_text(raw)?))
+        Ok(match *ty {
+            // Here we allow mapping JSON of pg to JSONB of rw. But please note the JSONB and JSON have different behaviors in postgres.
+            // An example of different semantics for duplicated keys in an object:
+            // test=# select jsonb_each('{"foo": 1, "bar": 2, "foo": 3}');
+            //  jsonb_each
+            //  ------------
+            //   (bar,2)
+            //   (foo,3)
+            //  (2 rows)
+            // test=# select json_each('{"foo": 1, "bar": 2, "foo": 3}');
+            //   json_each
+            //  -----------
+            //   (foo,1)
+            //   (bar,2)
+            //   (foo,3)
+            //  (3 rows)
+            Type::JSON => JsonbVal::from(Value::from_text(raw)?),
+            Type::JSONB => {
+                if raw.is_empty() || raw.get_u8() != 1 {
+                    return Err("invalid jsonb encoding".into());
+                }
+                JsonbVal::from(Value::from_text(raw)?)
+            }
+            _ => {
+                bail_not_implemented!("the JsonbVal's postgres decoding for {ty} is unsupported")
+            }
+        })
     }
 
     fn accepts(ty: &Type) -> bool {
-        matches!(*ty, Type::JSONB)
+        matches!(*ty, Type::JSONB | Type::JSON)
     }
 }
 
--- a/src/common/src/types/jsonb.rs
+++ b/src/common/src/types/jsonb.rs
@@ -569,25 +569,29 @@ impl<F: std::fmt::Write> std::io::Write for FmtToIoUnchecked<F> {
 }
 
 impl ToSql for JsonbVal {
-    accepts!(JSONB);
+    accepts!(JSON, JSONB);
 
     to_sql_checked!();
 
     fn to_sql(
         &self,
-        _ty: &Type,
+        ty: &Type,
         out: &mut BytesMut,
     ) -> Result<IsNull, Box<dyn std::error::Error + Sync + Send>>
     where
         Self: Sized,
     {
-        out.put_u8(1);
+        if matches!(*ty, Type::JSONB) {
+            out.put_u8(1);
+        }
         write!(out, "{}", self.0).unwrap();
         Ok(IsNull::No)
     }
 }
 
 impl<'a> FromSql<'a> for JsonbVal {
+    accepts!(JSON, JSONB);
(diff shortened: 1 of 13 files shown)


src/connector/src/connector_common/postgres.rs
```diff
@@ -207,7 +207,7 @@ impl PostgresExternalTable {
             SeaType::Boolean => Ok(PgType::BOOL),
             SeaType::Point => Ok(PgType::POINT),
             SeaType::Uuid => Ok(PgType::UUID),
-            SeaType::JsonBinary => Ok(PgType::JSONB),
+            SeaType::Json | SeaType::JsonBinary => Ok(PgType::JSONB),
```
xiangjinwu, line 210: How is this used in RisingWave? I see `CHAR_ARRAY` and `TIMETZ_ARRAY` here which do not exist in RisingWave.

src/common/src/types/jsonb.rs
```diff
@@ -589,17 +589,25 @@ impl ToSql for JsonbVal {
...
-        _ty: &Type,
+        ty: &Type,
         mut raw: &'a [u8],
     ) -> Result<Self, Box<dyn std::error::Error + Sync + Send>> {
-        if raw.is_empty() || raw.get_u8() != 1 {
-            return Err("invalid jsonb encoding".into());
-        }
-        Ok(JsonbVal::from(Value::from_text(raw)?))
+        Ok(match *ty {
+            Type::JSON => JsonbVal::from(Value::from_text(raw)?),
```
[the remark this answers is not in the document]
KeXiangWang: Added. Will also add it in doc.

src/connector/src/connector_common/postgres.rs
KeXiangWang: Thanks for pointing this out. After checking, these codes are for postgres sink. Since we don't have json type in RW, let's keep sink  to `json` as unsupported.

KeXiangWang: I've reverted this part.

KeXiangWang: This part of codes are used to load the schema of the table to be connected via source or sink.
`timestamptz[]` should be supported in RW?

xiangjinwu: I meant it should be either:
```
    // mapping into RisingWave supported data types
    SeaType::Char(_) | SeaType::Varchar(_) => Ok(PgType::VARCHAR),
    SeaType::Json | SeaType::JsonBinary => Ok(PgType::JSONB)
```
or
```
    // mapping into PostgreSQL data types, regardless of RisingWave support
    SeaType::Varchar(_) => Ok(PgType::VARCHAR),
    SeaType::Char(_) => Ok(PgType::CHAR),
    SeaType::JsonBinary => Ok(PgType::JSONB),
    SeaType::Json => Ok(PgType::JSON),
```
But after this PR it is:
```
    SeaType::Varchar(_) => Ok(PgType::VARCHAR),
    SeaType::Char(_) => Ok(PgType::CHAR),
    SeaType::Json | SeaType::JsonBinary => Ok(PgType::JSONB)
```
That's why I am asking how is this function used / what is its responsibility.

KeXiangWang: It's used by both sink and source. For source, I can confirm it from a source perspective, it should be 
```
    SeaType::Char(_) | SeaType::Varchar(_) => Ok(PgType::VARCHAR),
    SeaType::Json | SeaType::JsonBinary => Ok(PgType::JSONB)
```
cc @kwannoel for suggestion for sink.

kwannoel: Yup correct

KeXiangWang: OKK, I'll update it.

xiangjinwu: BEFORE we make the correction for `char`, is there any way to reproduce a test failure for both source and sink if there is a `char` column?

KeXiangWang: I try for source, even with `SeaType::Char(_) => Ok(PgType::CHAR),` it can still be interpreted as varchar successfully.

KeXiangWang: I've updated this part. For `TIMETZ_ARRAY`, it doesn't work with `TIMESTAMPTZ_ARRAY` for now, I'll open another PR to resolve it.

xiangjinwu (approved):

src/connector/src/connector_common/postgres.rs
xiangjinwu: This function seems only reachable for sink:
* `PostgresExternalTable::discovered_type_to_pg_type()`
  * `PostgresExternalTable::column_name_to_pg_type`
    * `PostgresExternalTable::column_name_to_pg_type()`
      * `PostgresSinkWriter::schema_types`

xiangjinwu: I traced further and have the following findings:
* For source, the result of this mapping is unused. It can be anything - even mapping `SeaType::Json` to `PgType::INT2`
* For sink, the result of this mapping is ineffective when no corresponding logic in `ScalarAdapter::from_scalar` - so in this case, even mapping `SeaType::Json` to `PgType::INT2` makes no difference.
* For sink to actually work, one possible implementation is to update `ToSql for JsonbVal` to omit the leading byte for `PgType::JSON`. Note this `PgType` comes from actual binding rather than output of this function, so it is always `PgType::JSON` rather than `PgType::INT2` above.
* For sink to actually work, alternative implementation is to introduce a corresponding variant in `ScalarAdapter`.

So my suggestions are:
* Keep this file unchanged for source.
* For sink, it does not matter either. But it is supposed to be the real PgType (rather than RW type) of the external table.
* For sink to work, update `ToSql for JsonbVal` similar to this PR's `FromSql for JsonbVal`

KeXiangWang: Update: 
1. Refactored the codes to build `column_name_to_pg_type` for source connector.
2. Renamed several function/type names to make the codes cleaner.
3. Update `ToSql for JsonbVal` similar to this PR's `FromSql for JsonbVal`
4. Just tested and found that both `JSON` and `JSONB` can work.

KeXiangWang: And I also realize, `column_name_to_pg_type` is only for native pg sink (`connector = 'postgres'`)

xiangjinwu (approved):
Thanks for the refactor!

PR #20446, opened 2025-02-11
feat(connector): resolve `$ref` in json schema

### LINKED ISSUE #20351: handle json reference in json schema
It's a common practice in JSON schema that a field refers to content in a `definition` to define its properties. 

https://json-schema.org/draft/2020-12/draft-bhutton-json-schema-00#rfc.section.8.2.3

```
{
    "name": "example",
    "address": { "$ref": "#/definitions/address" },
    "definitions": {
        "address": {
            "city": "Austin",
            "state": "TX"
        }
    }
} 
```

yuhao-su:

close #20351

This PR only resolve the `$ref` in a json schema. Other fields like `$anchor` `$id` etc was not resolved or checked.

This PR does not handle the json schema used by confluent schema registry. 

## Checklist

## Documentation

<details>

</details>


src/connector/codec/src/decoder/json/mod.rs
```diff
@@ -12,11 +12,170 @@
...
+                    let ref_url = id_url.join(ref_string).into_url_parse(ref_string)?;
+                    let mut ref_url_no_fragment = ref_url.clone();
+                    ref_url_no_fragment.set_fragment(None);
+                    let url_schema = ref_url_no_fragment.scheme();
+                    let ref_no_fragment = ref_url_no_fragment.to_string();
+
+                    let mut schema = match self.schema_cache.get(&ref_no_fragment) {
+                        Some(cached_schema) => cached_schema.clone(),
+                        None => {
+                            if url_schema == "http" || url_schema == "https" {
```
xxchan, line 109: Are you sure we want to support schema reference from http/fs?
Should we support retrieve references from schema registry instead? https://docs.confluent.io/platform/current/schema-registry/fundamentals/serdes-develop/serdes-json.html#schema-references-in-json-schemas

xxchan: Is this feature a customer request? Please provide more context, thanks.

yuhao-su: User only requested schema ref resolution without an external link.

I thought the HTTP support can be a general approve to support schema registry. But I just realized there is a version segment after the subject. I'll change this pr to only support schema resolution without external resource for now.

yuhao-su: Actually, we do allow users using http and file. So it won't hurt to add those support.

It's just for the schema registry, we need to investigate more into how it works.

src/connector/codec/src/decoder/json/mod.rs
```diff
@@ -26,7 +185,8 @@ impl crate::JsonSchema {
     /// <https://github.com/mozilla/jsonschema-transpiler/blob/fb715c7147ebd52427e0aea09b2bba2d539850b1/src/jsonschema.rs#L228-L280>
     ///
     /// TODO: examine other stuff like `oneOf`, `patternProperties`, etc.
-    pub fn json_schema_to_columns(&self) -> anyhow::Result<Vec<ColumnDesc>> {
+    pub async fn json_schema_to_columns(&mut self, uri: String) -> anyhow::Result<Vec<ColumnDesc>> {
```
xiangjinwu, line 188:
```suggestion
    pub async fn json_schema_to_columns(&mut self, retrieval: &Url) -> anyhow::Result<Vec<ColumnDesc>> {
```

src/connector/codec/src/decoder/json/mod.rs
```diff
@@ -12,11 +12,170 @@
...
+impl JsonRef {
+    fn new() -> JsonRef {
+        JsonRef {
+            schema_cache: HashMap::new(),
+            reference_key: Some("__reference__".to_owned()),
+        }
+    }
+
+    async fn deref_value(&mut self, value: &mut Value, id: String) -> Result<()> {
+        let id_url = Url::parse(&id).into_url_parse(&id)?.to_string();
```
xiangjinwu, line 84: No need to parse again if we pass in `retrieval: &Url`

src/connector/codec/src/decoder/json/mod.rs
```diff
@@ -12,11 +12,170 @@
...
+        #[from]
+        #[backtrace]
+        anyhow::Error,
+    ),
+}
+
+type Result<T, E = Error> = std::result::Result<T, E>;
+
+#[derive(Debug)]
+pub struct JsonRef {
```
xiangjinwu, line 70: Acknowledge that `$id` and `$anchor` are currently ignored and unsupported.

src/connector/codec/src/decoder/json/mod.rs
```diff
@@ -12,11 +12,170 @@
...
+        let id_url = Url::parse(&id).into_url_parse(&id)?.to_string();
+        self.schema_cache.insert(id_url.clone(), value.clone());
+        self.deref(value, id_url, &vec![]).await?;
+        Ok(())
+    }
+
+    async fn deref(
+        &mut self,
+        value: &mut Value,
+        id: String,
```
xiangjinwu, line 93:
```suggestion
        base: &Url,
```

src/connector/codec/src/decoder/json/mod.rs
```diff
@@ -12,11 +12,170 @@
...
+        used_refs: &Vec<String>,
+    ) -> Result<()> {
+        if let Some(obj) = value.as_object_mut() {
+            if let Some(ref_value) = obj.remove("$ref") {
+                if let Some(ref_string) = ref_value.as_str() {
```
xiangjinwu, line 98:
```suggestion
        if let Some(obj) = value.as_object_mut()
            && let Some(ref_value) = obj.remove("$ref")
                && let Some(ref_string) = ref_value.as_str() {
```

src/connector/codec/src/decoder/json/mod.rs
```diff
@@ -12,11 +12,170 @@
...
+                    let id_url = Url::parse(&id).into_url_parse(&id)?;
```
xiangjinwu, line 99: Unnecessary `parse`

src/connector/codec/src/decoder/json/mod.rs
```diff
@@ -12,11 +12,170 @@
...
+                    new_used_refs.push(ref_url_string);
+
+                    Box::pin(self.deref(&mut schema, ref_no_fragment, &new_used_refs)).await?;
+                    let old_value = mem::replace(value, schema);
+
+                    if let Some(reference_key) = &self.reference_key {
+                        if let Some(new_obj) = value.as_object_mut() {
+                            new_obj.insert(reference_key.clone(), old_value);
+                        }
+                    }
```
xiangjinwu, line 165: Any document on this behavior? It seems `self.reference_key` is constantly set to `Some("__reference__".to_owned())`

yuhao-su: Yes it's unccessary.  It's just when a object contains more than `$ref`. The items other than `$ref` will be ignored and put it in a __reference__ file. It's from the python and rust library `jsonref`

src/connector/codec/src/decoder/json/mod.rs
yuhao-su: I did check the jsonschema library. It's more like designed for validation and is vary hard to use it for dereffing the schema file.

src/connector/codec/src/decoder/json/mod.rs
yuhao-su: > they provide a way to load schemas into an internal schema database

It's true a for a deref library to focus on derefing instead of fetching (in fact it's  `[referencing](https://github.com/Stranger6667/jsonschema/tree/master/crates/jsonschema-referencing)` do). But if we want do do so we will have to ask users to provide all the URL in the schema ref or parse the schema to get all the $ref recurrently before ref.

src/connector/codec/src/decoder/json/mod.rs
tabVersion: shall we move the impl to a standalone crate? I feel that the kernel should not touch the logic of JsonRef.

yuhao-su: No. We may need to hack it to support confluent schema registry. The confluent schema registry should be handled differently. IIUC they can use reference name to retrieve a reference instead of URI.

src/connector/codec/src/decoder/json/mod.rs
xiangjinwu: Seem the dereferencing implementation is from the `jsonref` library (including the repeated unnecessary `parse` + `to_string`), so we need to include its MIT license.

Let's also remove the unused `__reference__` key to avoid confusing people reading this part. As a standalone library it provides flexibility to the caller, but it is just dead code here.

yuhao-su: > so we need to include its MIT license.

Good catch. Can't believe I missed this...

> remove the unused __reference__ key

done

src/connector/codec/src/decoder/json/mod.rs
```diff
@@ -12,11 +12,180 @@
...
+                        ref_fragment: ref_fragment.to_owned(),
+                    })?
+                    .clone();
+            }
+            // Do not deref a url twice to prevent infinite loops
+            if used_refs.contains(&ref_url_string) {
+                return Ok(());
+            }
+            let mut new_used_refs = used_refs.clone();
+            new_used_refs.push(ref_url_string);
```
xiangjinwu, line 175: Shall we add a test case on recursion? This `new_used_refs` is now unused.

yuhao-su: You are right. I mis deleted 1 extra deref line. Test added.

PR #20452, opened 2025-02-11
feat(SBC): Provision proto

CAJan93:

Communication between kernel and the Serverless Backfill Controller (SBC). Proto definitions should be in this repo and not in the SBC repo. 

Equivalent on SBC side:  https://github.com/risingwavelabs/serverless-backfill-controller/pull/76 

## Checklist

## Documentation

<details>

</details>


proto/sbc_provision.proto
```diff
@@ -0,0 +1,24 @@
+syntax = "proto3";
+
+// Provision a node via the serverless backfill controller.
+package sbc_provision;
```
arkbriar, line 4: How about `sbc` only?

proto/sbc_provision.proto
```diff
@@ -0,0 +1,24 @@
...
+
+message ProvisionRequest {
+    // Leave it empty for future extension.
+}
+
+message ProvisionResponse {
+    string label = 1;
+}
+
+service NodeGroupController {
```
arkbriar, line 16: Shall we unify the term, such as `ResourceGroupController`? 

@shanicky PTAL

proto/sbc_provision.proto
```diff
@@ -0,0 +1,24 @@
...
+package sbc_provision;
+
+option optimize_for = SPEED;
+
+message ProvisionRequest {
+    // Leave it empty for future extension.
+}
+
+message ProvisionResponse {
+    string label = 1;
```
arkbriar, line 13: Ditto.

```suggestion
    string resource_group = 1;
```

arkbriar (commented):
+ @hzxa21 @fuyufjh

CAJan93:
I will add the Proto-related changes in https://github.com/risingwavelabs/serverless-backfill-controller/pull/81

proto/sbc_provision.proto
CAJan93: Do we plan to provision something other than node groups? If not we can also be precise and use `NodeGroupController`.

arkbriar: I'm fine with both. `Resource group` is what's used in the kernel, while `node group` is from the operator. They do have a 1-1 mapping.

Issue #19031, opened 2024-10-21
Make state table visible for creating MV


xxchan:
Currently we hide both MV and their state tables during backfilling. However, I think there's no need to hide the state tables:
1. Hiding state tables don't have large benefits, compared with hiding the MV (e.g., for consistency), since state table is an internal thing.
2. For backfill executors' state tables, they are only meaningful during the backfill stage. Their content can also be viewed as an observability tool for backfilling. So it's beneficial to expose them.

related:
- Inspired when working on `SHOW JOBS` (another observability tool) https://github.com/risingwavelabs/risingwave/pull/18925#issuecomment-2425430299
- (also for creating MV, but not the same dimension) https://github.com/risingwavelabs/risingwave/issues/12771

BugenZhao:
Since https://github.com/risingwavelabs/risingwave/pull/17503, we've already broadcast all table catalogs (including both internal tables and the MV) to the frontend immediately when meta starts the creating procedure. You can verify this by `SHOW INTERNAL TABLES`.

Selecting from internal tables is intentionally disabled, by only resolving the "created" table during binding.

https://github.com/risingwavelabs/risingwave/blob/33bca3a61e94975a44d73d7a9743bf8e27ace4c7/src/frontend/src/binder/relation/table_or_source.rs#L177-L178

BugenZhao:
However, as described in #18944, the catalogs received by the frontend during creation are incomplete, with fields like `fragment_id` or `vnode_count` not correctly filled. Performing batch scan during this period may lead to problem.

Perhaps this is the real motivation for the refactor of _only_ notifying the complete catalogs once to the frontends, which is the original idea of #18944:

<img width="519" alt="image" src="https://github.com/user-attachments/assets/f1c884ec-1d4d-4962-8ad2-8f78f8b9c836">

kwannoel:
> However, as described in #18944, the catalogs received by the frontend during creation are incomplete, with fields like `fragment_id` or `vnode_count` not correctly filled. Performing batch scan during this period may lead to problem.
> 
> Perhaps this is the real motivation for the refactor of _only_ notifying the complete catalogs once to the frontends, which is the original idea of #18944:
[...]

PR #20460, opened 2025-02-11
feat(meta): expose internal table catalog during backfill phase to batch engine

kwannoel:

Closes https://github.com/risingwavelabs/risingwave/issues/19031

We notify the MV catalog to the frontend only after the vnode bitmap field is filled up.

We also report the fragments to FE, before the stream job is finished creating, they are also required to run batch queries.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/meta/src/controller/streaming_job.rs
+++ b/src/meta/src/controller/streaming_job.rs
@@ -389,21 +389,6 @@ impl CatalogController {
         }
         txn.commit().await?;
 
-        if job.is_materialized_view() {
-            self.notify_frontend(
-                Operation::Add,
-                Info::ObjectGroup(PbObjectGroup {
-                    objects: incomplete_internal_tables
-                        .iter()
-                        .map(|table| PbObject {
-                            object_info: Some(PbObjectInfo::Table(table.clone())),
-                        })
-                        .collect(),
-                }),
-            )
-            .await;
-        }
-
         Ok(table_id_map)
     }
 
--- a/src/meta/src/controller/streaming_job.rs
+++ b/src/meta/src/controller/streaming_job.rs
@@ -178,8 +178,6 @@ impl CatalogController {
             }
         }
 
-        let mut objects = vec![];
-
         match streaming_job {
             StreamingJob::MaterializedView(table) => {
                 let job_id = Self::create_streaming_job_obj(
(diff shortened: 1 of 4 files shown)


BugenZhao (approved):

src/meta/src/controller/streaming_job.rs
```diff
@@ -497,6 +476,12 @@ impl CatalogController {
                     }
                     .update(&txn)
                     .await?;
+
+                    if is_materialized_view {
```
chenzl25, line 480: Shall we also expose internal tables for other streaming jobs (like sink)?

chenzl25 (approved):

src/meta/src/controller/streaming_job.rs
BugenZhao: Can sink DDL be blocking now?

kwannoel: Sink DDL is already blocking by default during the backfill phase I think 🤔

kwannoel: > Shall we also expose internal tables for other streaming jobs (like sink)?
https://github.com/risingwavelabs/risingwave/issues/20522

src/meta/src/controller/streaming_job.rs
```diff
@@ -198,10 +196,6 @@ impl CatalogController {
                 table.id = job_id as _;
                 let table_model: table::ActiveModel = table.clone().into();
                 Table::insert(table_model).exec(&txn).await?;
-
-                objects.push(PbObject {
-                    object_info: Some(PbObjectInfo::Table(table.to_owned())),
-                });
```
xxchan, line 204: It seems in the old code, we notify the MV table + internal tables, but now we just notify internal tables. Why don't we notify the MV table any more? 🤔 Did I miss sth?

kwannoel: See: https://github.com/risingwavelabs/risingwave/pull/20460/files#r1959043203

src/meta/src/controller/streaming_job.rs
```diff
@@ -537,6 +522,12 @@ impl CatalogController {
 
         txn.commit().await?;
 
+        if !objects.is_empty() {
+            assert!(is_materialized_view);
+            self.notify_frontend(Operation::Add, Info::ObjectGroup(PbObjectGroup { objects }))
```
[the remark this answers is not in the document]
xxchan: Yes I saw this line, but it seems `objects` only contains state tables (in `for state_table_id in state_table_ids`). Do you mean MV is also a state table?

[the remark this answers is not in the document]
kwannoel: Yes, you can see in the deleted code above, for the MV we just send its table catalog to FE. And same in the code here.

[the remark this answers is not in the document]
kwannoel: MV table is just another state table indeed.

PR #20470, opened 2025-02-12
refactor(frontend): release large objects early

zwang28:

This PR aims to optimize memory usage of the frontend node in specific scenarios.
Here is the case:
A query `INSERT INTO VALUES (),(),()` can potentially be very large in size. If this query takes a long time to complete, it can significantly increase the memory usage of the frontend node. This is because the frontend retains various objects derived from the query, such as the SQL string and PlanNode, until the query is finished.

A simple reproduction is as follows:
1. `CREATE TABLE t(v int);`
2. `ALTER TABLE t SET dml_rate_limit to 0;` This will pause the DML executor, thereby preventing any following INSERT operations from completing.
3. Repeatedly call `INSERT INTO t values (),(),()...()`.  So there'll be increasing number of ongoing queries.

Without this PR, the frontend node will OOM quickly.
With this PR, the frontend node memory usage is capped and won't OOM.

Note that this PR doesn't resolve the OOM issue caused by a surge of large incoming queries. [More on this](https://risingwave-labs.slack.com/archives/C07TGGFUNMD/p1739273341047389?thread_ts=1738986297.402499&cid=C07TGGFUNMD).

## Checklist

## Documentation

<details>

</details>


BugenZhao (approved):
Interesting...

zwang28:
> Interesting... Did you locate them by heap profiling?

Yes, reproduce locally and profile.

chenzl25 (approved):
LGTM！

PR #20480, opened 2025-02-12
feat(mongodb): add support for mongodb as a backend for meta

jayakasadev:

* add support for using MongoDB as a backend for meta
* update the MongoDB connector to use the latest driver.

## Checklist

## Documentation

<details>

</details>


jayakasadev:
This is WIP. I'm not creating any new Model structs. I'm creating wrapper structs with custom serialization to read/write from the db. 

I'm using an enum to wrap the database client (ie sql or mongodb) and using pattern matching to read from the backend. There should be an initial performance hit from branching, but this should go away in time with branch prediction and speculative execution.

src/connector/src/connector_common/common.rs
```diff
@@ -841,7 +841,14 @@ pub struct MongodbCommon {
...
     pub(crate) async fn build_client(&self) -> ConnectorResult<mongodb::Client> {
-        let client = mongodb::Client::with_uri_str(&self.connect_uri).await?;
+        let mut options = mongodb::options::ClientOptions::parse(&self.connect_uri).await?;
+        // Set the server_api field of the client_options object to Stable API version 1
+        let server_api = mongodb::options::ServerApi::builder()
+            .version(mongodb::options::ServerApiVersion::V1)
+            .build();
+        options.server_api = Some(server_api);
+        // Create a new client and connect to the server
+        let client = mongodb::Client::with_options(options)?;
```
jayakasadev, line 851: updating the connector to use the latest driver

TennyZhuang:
Thank you for your contribution.
The NoSQL meta backend is not currently on our roadmap, and its subsequent maintenance could also be quite troublesome for us. For each new metadata interface, we would have to add a new implementation along with the corresponding tests.
If you really need such a feature, you might consider implementing your own meta service based on our gRPC proto, which could be simpler than merging the code in.

jayakasadev:
> Thank you for your contribution. The NoSQL meta backend is not currently on our roadmap, and its subsequent maintenance could also be quite troublesome for us. For each new metadata interface, we would have to add a new implementation along with the corresponding tests. If you really need such a feature, you might consider implementing your own meta service based on our gRPC proto, which could be simpler than merging the code in.

is there a doc on this i can follow?

yezizp2012:
> > Thank you for your contribution. The NoSQL meta backend is not currently on our roadmap, and its subsequent maintenance could also be quite troublesome for us. For each new metadata interface, we would have to add a new implementation along with the corresponding tests. If you really need such a feature, you might consider implementing your own meta service based on our gRPC proto, which could be simpler than merging the code in.
> 
> is there a doc on this i can follow?

Unfortunately, we don’t have documentation specifically for implementing a custom meta service. While it’s possible to build one based on our gRPC proto, it’s not a straightforward task. The meta service in RisingWave not only maintains metadata but also plays a crucial role as a coordinator, managing cluster components and task scheduling.  

If you’re looking to support MongoDB, a more practical approach would be to fork the RisingWave repo and reimplement each necessary interface while keeping it in sync with upstream changes. However, NoSQL-based implementations differ significantly from SQL-based ones, meaning additional work beyond just interface adaptation would be required. Given the complexity, we wouldn’t recommend this unless absolutely necessary.

PR #20340, opened 2025-01-29
feat(iceberg): add support for creating new iceberg tables with partition keys

jayakasadev:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## What's changed and what's your intention?

* add partition_by field to iceberg sink when creating new iceberg table
  * allow users to pass a list of fields or function wrapped fields (ie year(date_column)) 

## Checklist

- [x] I have written necessary rustdoc comments.
- [ ]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  My PR contains critical fixes that are necessary to be merged into the latest release. 

## Documentation

- [x]  My PR needs documentation updates. 

<details>
<summary><b>Release note</b></summary>

</details>

PR #20495, opened 2025-02-14
fix(sink): fix iceberg sink partition_by panic 

xxhZs:

In this pr https://github.com/risingwavelabs/risingwave/pull/20340
we support partition_by  for iceberg, But it will panic, when `partition_by` is the column name

this pr fix panic. 
But iceberg-rust doesn't support func now, so this option is currently only supported by column name
https://github.com/apache/iceberg-rust/blob/75c7e8d67942fc6e29413d8011c508bdba95abe9/crates/iceberg/src/spec/partition.rs#L537

## Checklist

## Documentation

iceberg sink
with option: partition_by  Optional. A partition operation is performed based on the value of this column.  When the user sets partition_by  to a value that changes frequently, it will write too many files and lead to commit errors

### DIFF
--- a/src/connector/src/sink/iceberg/mod.rs
+++ b/src/connector/src/sink/iceberg/mod.rs
@@ -331,20 +331,28 @@ impl IcebergSink {
                     }
                     let caps = re.captures_iter(partition_field);
                     for mat in caps {
-                        let (column, transform) = if mat.name("n").is_none() && mat.name("field").is_none()
-                        {
-                            (&mat["transform"], Transform::Identity)
-                        } else {
-                            let mut func = mat["transform"].to_owned();
-                            if func == "bucket" || func == "truncate" {
-                                let n = &mat.name("n").ok_or_else(||{
-                                    SinkError::Iceberg(anyhow!("The `n` must be set with `bucket` and `truncate`"))
-                                })?.as_str();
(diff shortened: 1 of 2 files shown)


src/connector/src/sink/iceberg/mod.rs
```diff
@@ -331,17 +331,20 @@ impl IcebergSink {
...
-                                func = format!("{func}({n})");
+                                let n = &mat.name("n").ok_or_else(||{
+                                    SinkError::Iceberg(anyhow!("The `n` must be set with `bucket` and `truncate`"))
+                                })?.as_str();
+                                func = format!("{func}[{n}]");
                             }
-                            (&mat["field"], Transform::from_str(&func).unwrap())
+                            (&mat["field"], Transform::from_str(&func).map_err(|e| SinkError::Iceberg(anyhow!(e)))?)
                         };
+                        println!("column: {}, transform: {:?}", column, transform);
```
chenzl25, line 347: Remove the print？

chenzl25 (approved):
Rest LGTM， thanks

xxhZs:
Also this pr will have the problem that when the user sets `partition_by ` to a value that changes frequently, it will write too many files and lead to commit errors, but since this is not a correct usage, there is no restriction on this case

PR #20498, opened 2025-02-14
fix(file source): fix rate limit for list executor

wcy-fdu:

In the current logic, the list executor will only handle Pause, which means setting the rate limit to 0. As a result, we have observed some bugs: after setting the rate limit to 0 and then restoring it to default, only the fetch executor took effect, while the list executor remained at 0. Therefore, when the files previously persisted by the fetch executor are read completely, this source stops, which is why we observed the source throughput"sudden drop to 0 after running for a while".

## Checklist

## Documentation

<details>

</details>


kwannoel (commented):
Still a little confused about something. The `new_rate_limit` should be `None` when resuming, i.e. `ALTER SOURCE SET SOURCE_RATE_LIMIT=default` will set rate limit to None.

Which means the changes here are no-op, and don't actually resume the fs_list_executor.

Is there some other explanation?

Could you add the repro as a test case, and does the changes here fix it?

tabVersion:
I don't get why we need this. As Fetch Exec has already been throttled, the new workload will not enter the cluster. The listed files will just write to the downstream state table.

wcy-fdu:
How to reproduce
```
CREATE source s1(
        id int,
    ) WITH (
    connector = 's3',
    match_pattern = '*.parquet',
    s3.region_name = 'custom',
    s3.bucket_name = 'hummock001',
    s3.credentials.access = 'hummockadmin',
    s3.credentials.secret = 'hummockadmin',
    s3.endpoint_url = 'http://hummock001.127.0.0.1:9301',
    refresh.interval.sec = 1,
    ) FORMAT PLAIN ENCODE PARQUET;
create table d(id int);
create sink sink1 into d as select id from s1 with (
      type = 'append-only',
      force_append_only = 'true',
  );
alter sink sink1 set PARALLELISM to 1;
```
upload file1
`ALTER SOURCE s1 SET source_rate_limit TO 0;`
upload file2
`ALTER SOURCE s1 SET source_rate_limit TO default;`
upload file 3 4 5

File 3 4 5 cannot be read, but is in the fetch state table.

PR #20506, opened 2025-02-14
feat(source): update split state after yield barrier on split change for source executor

wenym1:

Previously on source split change (either on explicit split change, or on split migration during scale), we write the state of target splits (splits to read after barrier) before yielding the barrier, so as to include the state of target splits in the checkpoint of the barrier. However, before the barrier, some of the target splits may be owned by other source parallelisms, and the state of the splits may be written by these other source parallelisms, and if we also write the state in the current parallelism, the splits may be written in multiple parallelisms and hit assertion in storage on something like a write conflict. Semantically, in current parallelism, we start to read the split after the barrier, which means it owns the split only after the barrier, and before the barrier, it should not write state for split it has not owned yet.

Therefore, in this PR, on source split change, we will change to write the state of the target splits after yielding the barrier. Besides, when applying split change, we have call `try_recover_from_state_store`, which loads the splits across all parallelisms. However, in current code, we don't wait for the committed epoch, and the loaded state may be stale if the split is owned by other parallelisms. Therefore in this PR, we will wait for the committed epoch before `try_recover_from_state_store` to ensure that we read the latest split state. Since we have yielded the barrier before applying split change, we can ensure that we won't block the checkpoint committed epoch, and we can always eventually see the committed epoch bump up. The idea is the same as https://github.com/risingwavelabs/risingwave/issues/18312.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/executor/source/legacy_fs_source_executor.rs
+++ b/src/stream/src/executor/source/legacy_fs_source_executor.rs
@@ -15,7 +15,7 @@
 #![deprecated = "will be replaced by new fs source (list + fetch)"]
 #![expect(deprecated)]
 
-use std::collections::{HashMap, HashSet};
+use std::collections::HashSet;
 
 use anyhow::anyhow;
 use either::Either;
@@ -143,9 +143,9 @@ impl<S: StateStore> LegacyFsSourceExecutor<S> {
         &mut self,
         source_desc: &LegacyFsSourceDesc,
         stream: &mut StreamReaderWithPause<BIASED, StreamChunk>,
-        mapping: &HashMap<ActorId, Vec<SplitImpl>>,
+        target_splits: Vec<SplitImpl>,
     ) -> StreamExecutorResult<()> {
-        if let Some(target_splits) = mapping.get(&self.actor_ctx.id).cloned() {
+        {
             if let Some(target_state) = self.get_diff(target_splits).await? {
                 tracing::info!(
                     actor_id = self.actor_ctx.id,
@@ -407,12 +407,17 @@ impl<S: StateStore> LegacyFsSourceExecutor<S> {
                             self_paused = false;
                         }
                         let epoch = barrier.epoch;
+                        let mut split_change = None;
 
                         if let Some(ref mutation) = barrier.mutation.as_deref() {
                             match mutation {
                                 Mutation::SourceChangeSplit(actor_splits) => {
-                                    self.apply_split_change(&source_desc, &mut stream, actor_splits)
-                                        .await?
+                                    split_change = actor_splits
+                                        .get(&self.actor_ctx.id)
+                                        .cloned()
+                                        .map(|target_splits| {
+                                            (&source_desc, &mut stream, target_splits)
+                                        });
                                 }
                                 Mutation::Pause => {
                                     command_paused = true;
@@ -423,12 +428,12 @@ impl<S: StateStore> LegacyFsSourceExecutor<S> {
                                     stream.resume_stream()
                                 }
                                 Mutation::Update(UpdateMutation { actor_splits, .. }) => {
-                                    self.apply_split_change(
-                                        &source_desc,
-                                        &mut stream,
-                                        actor_splits,
-                                    )
-                                    .await?;
+                                    split_change = actor_splits
+                                        .get(&self.actor_ctx.id)
+                                        .cloned()
+                                        .map(|target_splits| {
+                                            (&source_desc, &mut stream, target_splits)
+                                        });
                                 }
                                 Mutation::Throttle(actor_to_apply) => {
                                     if let Some(new_rate_limit) =
@@ -446,6 +451,18 @@ impl<S: StateStore> LegacyFsSourceExecutor<S> {
                         self.take_snapshot_and_clear_cache(epoch).await?;
 
                         yield msg;
+
+                        if let Some((source_desc, stream, target_splits)) = split_change {
+                            // do `try_wait_committed_epoch` to ensure that when `apply_split_change`,
+                            // we can read the latest state written by other source executor
+                            self.stream_source_core
+                                .split_state_store
+                                .state_table
+                                .try_wait_committed_epoch(epoch.prev)
+                                .await?;
+                            self.apply_split_change(source_desc, stream, target_splits)
+                                .await?;
+                        }
                     }
                     _ => {
                         // For the source executor, the message we receive from this arm should
--- a/src/stream/src/executor/source/legacy_fs_source_executor.rs
+++ b/src/stream/src/executor/source/legacy_fs_source_executor.rs
@@ -139,14 +139,15 @@ impl<S: StateStore> LegacyFsSourceExecutor<S> {
         Ok(())
     }
 
-    async fn apply_split_change<const BIASED: bool>(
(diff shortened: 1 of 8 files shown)


wenym1:
* **#20493** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20493?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#20506** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20506?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20506?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

wenym1:
The risk is that on source split change, the split state of some newly generated split will not be included in the checkpoint of the barrier of source split change. 

From my understanding, this is acceptable, because split state is the offset information. On source split change, the current parallelism has not owned the split yet before the barrier, and therefore should not have any update on the split offset, and therefore won't have data loss even if the state of the target split is not written.

src/stream/src/executor/source/legacy_fs_source_executor.rs
```diff
@@ -271,10 +275,7 @@ impl<S: StateStore> LegacyFsSourceExecutor<S> {
             core.split_state_store.set_all_complete(completed).await?
         }
         // commit anyway, even if no message saved
-        core.split_state_store
-            .state_table
-            .commit_assert_no_update_vnode_bitmap(epoch)
-            .await?;
+        core.split_state_store.commit(epoch).await?;
```
tabVersion, line 278: So we are committing the offset after meta has collected current epoch's barrier?

tabVersion:
Just checking:

If meeting some errors in the time of waiting committing epoch, we rollback to the latest checkpoint and expect the change split mutation to be sent again?

I am afraid the assumption does not hold as we don't handle the run command error in source manager.  

https://github.com/risingwavelabs/risingwave/blob/7c7e0ad4ed27fbc6c1f7a9b54dd96a80c2f1182f/src/meta/src/stream/source_manager.rs#L454-L486

src/stream/src/executor/source/legacy_fs_source_executor.rs
wenym1: Here is just a refactor, to change the method name we call and no original logic is changed here. 

The correct order should be: `receive barrier` -> `commit state store` -> `yield barrier` -> `meta collect barrier`. The current code is following this correct order.

wenym1:
> Just checking:
> 
> If meeting some errors in the time of waiting committing epoch, we rollback to the latest checkpoint and expect the change split mutation to be sent again?
> 
> I am afraid the assumption does not hold as we don't handle the run command error in source manager.
> 
> https://github.com/risingwavelabs/risingwave/blob/7c7e0ad4ed27fbc6c1f7a9b54dd96a80c2f1182f/src/meta/src/stream/source_manager.rs#L454-L486

Error can happen at any of step of source split change, not just when waiting for committing epoch. If this is a potential bug when source split change gets any error and fail to run the source split change command, we may need to fix it independently in a separate PR, because it's not brought by this PR, and this PR doesn't make it worse though.

xxchan (commented):
The idea generally LGTM. 

> and the state of the splits may be written by these other source parallelisms, and if we also write the state in the current parallelism, the splits may be written in multiple parallelisms and hit assertion in storage on something like a write conflict

Currently we don't have this problem because of the Pause, but it will be a problem after the new config change design. Do I understand it correctly?

wenym1:
> The idea generally LGTM.
> 
> > and the state of the splits may be written by these other source parallelisms, and if we also write the state in the current parallelism, the splits may be written in multiple parallelisms and hit assertion in storage on something like a write conflict
> 
> Currently we don't have this problem because of the Pause, but it will be a problem after the new config change design. Do I understand it correctly?

Yes.

BugenZhao:
> > Currently we don't have this problem because of the Pause, but it will be a problem after the new config change design. Do I understand it correctly?
> 
> Yes.

Do you mean that there's `Pause` for split change? 😳 Can you point me out the code path?

src/stream/src/executor/source/legacy_fs_source_executor.rs
```diff
@@ -163,14 +164,19 @@ impl<S: StateStore> LegacyFsSourceExecutor<S> {
...
+        rhs: Vec<SplitImpl>,
+    ) -> StreamExecutorResult<ConnectorState> {
         let core = &mut self.stream_source_core;
         let all_completed: HashSet<SplitId> = core.split_state_store.get_all_completed().await?;
 
         tracing::debug!(actor = self.actor_ctx.id, all_completed = ?all_completed , "get diff");
 
         let mut target_state: Vec<SplitImpl> = Vec::new();
         let mut no_change_flag = true;
+        let committed_reader = core.split_state_store.new_committed_reader(epoch).await?;
```
BugenZhao, line 179: Can we add some doc / comments?

BugenZhao (commented):
Generally LGTM. Can we have more docs and comments for such subtlety?

wenym1:
> > > Currently we don't have this problem because of the Pause, but it will be a problem after the new config change design. Do I understand it correctly?
> > 
> > 
> > Yes.
> 
> Do you mean that there's `Pause` for split change? 😳 Can you point me out the code path? I didn't find it.
> 
> https://github.com/risingwavelabs/risingwave/blob/10e177aa170aa9b352c09ef82bd0ed5efdfe6f71/src/meta/src/stream/source_manager.rs#L446-L471
> 
> If not, does this mean that these correctness issues might occur under current impl?
> 
> > 1. the splits may be written in multiple parallelisms and hit assertion in storage on something like a write conflict
> > 2. the loaded state may be stale if the split is owned by other parallelisms

The incorrectness only happens when a split is previously owned by another parallelism, and then get migrated to the current parallelism. Currently source split change can happen in either source migration happened during configuration change, or an explicit source split change. For configuration change, the previous pause-change-resume mechanism can avoid both write conflict and loading stale state. For explicit source spit change, if it never schedules a split migration to migrate a split from one parallelism to another, the incorrectness also won't happen. 

So in short, if the explicit source split change won't have split migration, but only have adding or deleting split, the incorrectness won't happen, and if not, our current implementation already has bug.

BugenZhao:
> So in short, if the explicit source split change won't have split migration, but only have adding or deleting split, the incorrectness won't happen, and if not, our current implementation already has bug.

Not sure which one is the case. 🤔 cc @tabVersion @shanicky

tabVersion:
> > So in short, if the explicit source split change won't have split migration, but only have adding or deleting split, the incorrectness won't happen, and if not, our current implementation already has bug.
> 
> Not sure which one is the case. 🤔 cc @tabVersion @shanicky

Source split change only adds splits or delete them. We try our best to keep split assignment balance but always avoid migration. So I think there should be no bug here.

wenym1:
> Generally LGTM. Can we have more docs and comments for such subtlety?

Comments are addressed. Do you have any other concerns to merge this PR? @tabVersion @BugenZhao @xxchan

BugenZhao (approved):
LGTM

PR #20511, opened 2025-02-17
feat(logstore): handle paused stream

kwannoel:

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/executor/sync_kv_log_store.rs
+++ b/src/stream/src/executor/sync_kv_log_store.rs
@@ -261,6 +261,7 @@ impl<S: StateStore> SyncedKvLogStoreExecutor<S> {
             new_log_store_state(self.table_id, local_state_store, self.serde);
         initial_write_state.init(first_write_epoch).await?;
 
+        let mut pause_stream = first_barrier.is_pause_on_startup();
         let mut initial_write_epoch = first_write_epoch;
 
         // We only recreate the consume stream when:
@@ -289,8 +290,15 @@ impl<S: StateStore> SyncedKvLogStoreExecutor<S> {
 
             loop {
                 let select_result = {
-                    let read_future =
-                        read_future.next_chunk(&read_state, &mut buffer, &self.metrics);
+                    let read_future = async {
+                        if pause_stream {
+                            pending().await
+                        } else {
+                            read_future
+                                .next_chunk(&read_state, &mut buffer, &self.metrics)
+                                .await
+                        }
+                    };
                     pin_mut!(read_future);
                     let write_future = write_future.next_event();
                     pin_mut!(write_future);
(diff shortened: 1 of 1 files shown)


kwannoel:
* **#20511** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20511?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20511?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/stream/src/executor/sync_kv_log_store.rs
```diff
@@ -430,6 +451,9 @@ impl<S: StateStoreRead> ReadFuture<S> {
                 *self = ReadFuture::Idle;
             }
             ReadFuture::ReadingFlushedChunk { .. } | ReadFuture::Idle => {}
+            ReadFuture::Paused => {
+                unreachable!("should not be polled after paused")
```
wenym1, line 455: Can just move the `return pending().await` at the beginning to here to avoid an extra match.

src/stream/src/executor/sync_kv_log_store.rs
```diff
@@ -260,6 +260,7 @@ impl<S: StateStore> SyncedKvLogStoreExecutor<S> {
             new_log_store_state(self.table_id, local_state_store, self.serde);
         initial_write_state.init(first_write_epoch).await?;
 
+        let mut pause_stream = false;
```
wenym1, line 263: May set the initial value as `barrier.is_pause_on_startup()`

wenym1 (commented):
Rest LGTM

src/stream/src/executor/sync_kv_log_store.rs

src/stream/src/executor/sync_kv_log_store.rs
```diff
@@ -301,6 +306,18 @@ impl<S: StateStore> SyncedKvLogStoreExecutor<S> {
                             WriteFutureEvent::UpstreamMessageReceived(msg) => {
                                 match msg {
                                     Message::Barrier(barrier) => {
+                                        if let Some(mutation) = barrier.mutation.as_deref() {
+                                            match mutation {
+                                                Mutation::Pause => {
+                                                    pause_stream = true;
```
wenym1, line 312: It seems that we only set the `pause_stream` variable, but the variable never takes effects. Besides, in current implementation, it seems that once `first_barrier.is_pause_on_startup()` is true, the read future will always be `Paused`, which is incorrect.

The correct implementation should be:
* remove the `ReadFuture::Paused` variant
* change the `read_future` input to `select` to the following
```
let read_future = async {
  if pause_stream {
      pending().await
  } else {
      read_future.next_chunk(&read_state, &mut buffer, &self.metrics)
  }
}
```

kwannoel: fixed

wenym1 (approved):
LGTM

PR #20523, opened 2025-02-18
refactor(stream): shortcircuit flushing chunks on barrier

kwannoel:

No need to iterate through the rest of the buffer after hitting a flushed item / chunk. Because the rest of the chunks should already be flushed

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/executor/sync_kv_log_store.rs
+++ b/src/stream/src/executor/sync_kv_log_store.rs
@@ -261,6 +261,7 @@ impl<S: StateStore> SyncedKvLogStoreExecutor<S> {
             new_log_store_state(self.table_id, local_state_store, self.serde);
         initial_write_state.init(first_write_epoch).await?;
 
+        let mut pause_stream = first_barrier.is_pause_on_startup();
         let mut initial_write_epoch = first_write_epoch;
 
         // We only recreate the consume stream when:
@@ -289,8 +290,15 @@ impl<S: StateStore> SyncedKvLogStoreExecutor<S> {
 
             loop {
                 let select_result = {
-                    let read_future =
-                        read_future.next_chunk(&read_state, &mut buffer, &self.metrics);
+                    let read_future = async {
+                        if pause_stream {
+                            pending().await
+                        } else {
+                            read_future
+                                .next_chunk(&read_state, &mut buffer, &self.metrics)
+                                .await
+                        }
+                    };
                     pin_mut!(read_future);
                     let write_future = write_future.next_event();
                     pin_mut!(write_future);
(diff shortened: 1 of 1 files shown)


kwannoel:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20523?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#20523** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20523?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20523?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#20511** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20511?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/stream/src/executor/sync_kv_log_store.rs
```diff
@@ -535,38 +535,40 @@ impl<S: StateStore> SyncedKvLogStoreExecutor<S> {
...
+        // TODO(kwannoel): As an optimization we can also change flushed chunks to be flushed items
+        // to reduce memory consumption of logstore.
 
-        // FIXME(kwannoel): Flush all unflushed chunks
-        // As an optimization we can also change it into flushed items instead.
-        // This will reduce memory consumption of logstore.
+        let epoch = barrier.epoch.prev;
+        if barrier.is_checkpoint() {
+            let mut writer = write_state.start_writer(false);
+            writer.write_barrier(epoch, barrier.is_checkpoint())?;
```
wenym1, line 544: We may need to start the writer and write the barrier anyway instead of only on checkpoint barrier. Otherwise, we may lose the non-checkpoint barrier.

src/stream/src/executor/sync_kv_log_store.rs
```diff
@@ -535,38 +535,40 @@ impl<S: StateStore> SyncedKvLogStoreExecutor<S> {
...
+                        ..
+                    } => {
+                        if !*flushed {
+                            writer.write_chunk(chunk, *epoch, *start_seq_id, *end_seq_id)?;
+                            *flushed = true;
+                        } else {
+                            break;
+                        }
                     }
+                    LogStoreBufferItem::Flushed { .. } => {
```
wenym1, line 562: Not sure if we can break when seeing `Flushed` item. For example, let's say the buffer size is 1 chunk, and then we have the following actions

start from empty buffer: []
add chunk 1: [chunk1] 
add flushed chunk2: [chunk1, flushed2] 
consume chunk1: [flushed2]
add chunk2: [flushed2, chunk2]

On checkpoint barrier, when we should write all unflushed chunks, we will first see flushed2, and then break the for loop. But chunk2 is not flushed yet.

src/stream/src/executor/sync_kv_log_store.rs

src/stream/src/executor/sync_kv_log_store.rs

src/stream/src/executor/sync_kv_log_store.rs
```diff
@@ -530,38 +530,39 @@ impl<S: StateStore> SyncedKvLogStoreExecutor<S> {
...
-                    start_seq_id,
-                    end_seq_id,
-                    flushed,
-                    ..
-                } => {
-                    if !*flushed {
-                        writer.write_chunk(chunk, *epoch, *start_seq_id, *end_seq_id)?;
-                        *flushed = true;
+        if barrier.is_checkpoint() {
+            for (epoch, item) in &mut buffer.buffer {
```
wenym1, line 541: It seems that when we add items to `buffer`, we use `buffer.push_back(...)`, which means a newer item is at the back. Therefore, should we iterate from back to front instead?

wenym1 (approved):

PR #20526, opened 2025-02-18
refactor: move source retry inside reader stream

tabVersion:

replacement of https://github.com/risingwavelabs/risingwave/pull/20295

the intuition is to handle the crash loop when upstream is unavailable and now we only report error when the exec is first executed. 

Source backfill does not need to change because when it is doing backfilling, a recover will lead to cancel the job, instead of rebuilding the actor. So when upstream is unavailable, it does not panic RisingWave in recovery. 

## Checklist

## Documentation

<details>

</details>


src/stream/src/executor/source/source_executor.rs
```diff
@@ -393,12 +404,11 @@ impl<S: StateStore> SourceExecutor<S> {
         );
 
         // Replace the source reader with a new one of the new state.
-        let (reader, _) = self
-            .build_stream_source_reader(source_desc, Some(target_state.clone()), false)
```
wenym1, line 397: It seems that the original `build_stream_source_reader` is not used anymore. May remove it.

PR #20532, opened 2025-02-18
feat(frontend): redirect cast char to varchar

chenzl25:

- Allow to use `cast char` in SQL and redirect it to `cast varchar`, because some ecosystems like metabase will use it to query the system table. 

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/binder/expr/mod.rs
+++ b/src/frontend/src/binder/expr/mod.rs
@@ -939,6 +939,17 @@ impl Binder {
             }
             // Redirect cast char to varchar to make system like Metabase happy.
             // Char is not supported in RisingWave, but some ecosystem tools like Metabase will use it.
+            // Notice that the behavior of `char` and `varchar` is different in PostgreSQL.
+            // The following sql result should be different in PostgreSQL:
+            // ```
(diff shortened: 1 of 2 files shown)


src/frontend/src/binder/expr/mod.rs
```diff
@@ -937,6 +937,9 @@ impl Binder {
                     Err(ErrorCode::BindError(format!("Can't cast {} to regproc", lhs_ty)).into())
                 }
             }
+            // Redirect cast char to varchar to make system like Metabase happy.
+            // Char is not supported in RisingWave, but some ecosystem tools like Metabase will use it.
+            AstDataType::Char(_) => self.bind_cast_inner(expr, DataType::Varchar),
```
xiangjinwu, line 942: What about including this counter-example in the comment / e2e test:
```
test=# select 'a'::char(2) = 'a '::char(2);
 ?column? 
----------
 t
(1 row)

test=# select 'a'::varchar = 'a '::varchar;
 ?column? 
----------
 f
(1 row)
```

xiangjinwu (approved):

PR #20546, opened 2025-02-20
feat(logstore): provide more opportunities to read

kwannoel:

At writer future we will sleep briefly when transitioning from clean state to flushed state. This allows us to avoid flushing, if the read side can catch up.

Clean state is when writer has no flushed items, not reading from state store.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/executor/sync_kv_log_store.rs
+++ b/src/stream/src/executor/sync_kv_log_store.rs
@@ -1009,17 +1009,17 @@ mod tests {
         }
 
         match stream.next().await {
-            Some(Ok(Message::Chunk(chunk))) => {
-                assert_eq!(chunk, chunk_2);
-            }
-            other => panic!("Expected a chunk message, got {:?}", other),
-        }
-
-        match stream.next().await {
             Some(Ok(Message::Barrier(barrier))) => {
                 assert_eq!(barrier.epoch.curr, test_epoch(2));
             }
             other => panic!("Expected a barrier message, got {:?}", other),
+        }
+
+        match stream.next().await {
+            Some(Ok(Message::Chunk(chunk))) => {
+                assert_eq!(chunk, chunk_2);
+            }
+            other => panic!("Expected a chunk message, got {:?}", other),
         }
     }
 
--- a/src/stream/src/executor/sync_kv_log_store.rs
+++ b/src/stream/src/executor/sync_kv_log_store.rs
@@ -154,7 +154,7 @@ struct FlushedChunkInfo {
 enum WriteFuture<S: LocalStateStore> {
     Paused {
         message: Message,
-        sleep_future: Option<Pin<Box<Sleep>>>,
+        sleep_future: Pin<Box<Sleep>>,
         stream: BoxedMessageStream,
         write_state: LogStoreWriteState<S>, // Just used to hold the state
     },
@@ -217,7 +217,7 @@ impl<S: LocalStateStore> WriteFuture<S> {
         write_state: LogStoreWriteState<S>,
     ) -> Self {
         Self::Paused {
-            sleep_future: Some(Box::pin(sleep(duration))),
+            sleep_future: Box::pin(sleep(duration)),
             message,
             stream,
             write_state,
@@ -229,10 +229,7 @@ impl<S: LocalStateStore> WriteFuture<S> {
     ) -> StreamExecutorResult<(BoxedMessageStream, LogStoreWriteState<S>, WriteFutureEvent)> {
         match self {
             WriteFuture::Paused { sleep_future, .. } => {
-                if let Some(sleep_future) = sleep_future {
-                    sleep_future.await;
-                }
-                sleep_future.take();
+                sleep_future.await;
                 must_match!(replace(self, WriteFuture::Empty), WriteFuture::Paused { stream, write_state, message, .. } => {
                     Ok((stream, write_state, WriteFutureEvent::UpstreamMessageReceived(message)))
                 })
@@ -312,7 +309,7 @@ impl<S: StateStore> SyncedKvLogStoreExecutor<S> {
                 max_size: self.max_buffer_size,
                 next_chunk_id: 0,
                 metrics: self.metrics.clone(),
-                flushed: false,
+                flushed_count: 0,
             };
 
             let mut log_store_stream = read_state
@@ -343,8 +340,7 @@ impl<S: StateStore> SyncedKvLogStoreExecutor<S> {
             let mut write_future_state =
                 WriteFuture::receive_from_upstream(input, initial_write_state);
 
-            let is_reading_from_storage = read_future_state.is_reading_from_storage();
-            let mut clean_state = !buffer.flushed() && !is_reading_from_storage;
+            let mut clean_state = matches!(read_future_state, ReadFuture::Idle);
             loop {
                 let select_result = {
                     let read_future = async {
@@ -371,7 +367,10 @@ impl<S: StateStore> SyncedKvLogStoreExecutor<S> {
                             WriteFutureEvent::UpstreamMessageReceived(msg) => {
                                 match msg {
                                     Message::Barrier(barrier) => {
-                                        if clean_state {
+                                        if clean_state
+                                            && buffer.no_flushed_items()
+                                            && barrier.kind.is_checkpoint()
+                                        {
                                             write_future_state = WriteFuture::paused(
                                                 self.pause_duration_ms,
                                                 Message::Barrier(barrier),
@@ -425,27 +424,26 @@ impl<S: StateStore> SyncedKvLogStoreExecutor<S> {
                                         }
                                     }
                                     Message::Chunk(chunk) => {
-                                        if buffer.size() + chunk.cardinality()
-                                            >= self.max_buffer_size
-                                        {
-                                            write_future_state = WriteFuture::paused(
-                                                self.pause_duration_ms,
-                                                Message::Chunk(chunk),
-                                                stream,
-                                                write_state,
-                                            );
-                                            clean_state = false;
-                                        } else {
-                                            let start_seq_id = seq_id;
-                                            seq_id += chunk.cardinality() as SeqIdType;
-                                            let end_seq_id = seq_id - 1;
-                                            let epoch = write_state.epoch().curr;
-                                            if let Some(chunk_to_flush) = buffer.add_or_flush_chunk(
-                                                start_seq_id,
-                                                end_seq_id,
-                                                chunk,
-                                                epoch,
-                                            ) {
+                                        let start_seq_id = seq_id;
+                                        let new_seq_id = seq_id + chunk.cardinality() as SeqIdType;
+                                        let end_seq_id = new_seq_id - 1;
+                                        let epoch = write_state.epoch().curr;
+                                        if let Some(chunk_to_flush) = buffer.add_or_flush_chunk(
+                                            start_seq_id,
+                                            end_seq_id,
+                                            chunk,
+                                            epoch,
+                                        ) {
+                                            if clean_state && buffer.no_flushed_items() {
+                                                write_future_state = WriteFuture::paused(
+                                                    self.pause_duration_ms,
+                                                    Message::Chunk(chunk_to_flush),
+                                                    stream,
+                                                    write_state,
+                                                );
+                                                clean_state = false;
+                                            } else {
+                                                seq_id = new_seq_id;
                                                 write_future_state = WriteFuture::flush_chunk(
                                                     stream,
                                                     write_state,
@@ -454,13 +452,13 @@ impl<S: StateStore> SyncedKvLogStoreExecutor<S> {
                                                     start_seq_id,
                                                     end_seq_id,
                                                 );
-                                            } else {
-                                                write_future_state =
-                                                    WriteFuture::receive_from_upstream(
-                                                        stream,
-                                                        write_state,
-                                                    );
                                             }
+                                        } else {
+                                            seq_id = new_seq_id;
+                                            write_future_state = WriteFuture::receive_from_upstream(
+                                                stream,
+                                                write_state,
+                                            );
                                         }
                                     }
                                     // FIXME(kwannoel): This should truncate the logstore,
@@ -491,6 +489,11 @@ impl<S: StateStore> SyncedKvLogStoreExecutor<S> {
                         }
                     }
                     Either::Right(result) => {
+                        if matches!(read_future_state, ReadFuture::Idle)
+                            && buffer.no_flushed_items()
+                        {
+                            clean_state = true;
+                        }
                         let (chunk, new_truncate_offset) = result?;
                         if let Some(new_truncate_offset) = new_truncate_offset {
                             truncation_offset = Some(new_truncate_offset);
@@ -614,15 +617,6 @@ impl<S: StateStoreRead> ReadFuture<S> {
         *self = ReadFuture::Idle;
         Ok((chunk, Some(truncate_offset)))
     }
-
-    fn is_reading_from_storage(&self) -> bool {
-        match self {
-            ReadFuture::ReadingInitialPersistedStream { .. }
-            | ReadFuture::ReadingPersistedStream(_)
-            | ReadFuture::ReadingFlushedChunk { .. } => true,
-            ReadFuture::Idle => false,
-        }
-    }
 }
 
 // Write methods
@@ -689,17 +683,13 @@ struct SyncedLogStoreBuffer {
     max_size: usize,
     next_chunk_id: ChunkId,
     metrics: KvLogStoreMetrics,
-    flushed: bool,
+    flushed_count: usize,
 }
 
 impl SyncedLogStoreBuffer {
-    fn size(&self) -> usize {
-        self.current_size
-    }
-
     /// Returns true if there are flushed items in the buffer.
-    fn flushed(&self) -> bool {
-        self.flushed
+    fn no_flushed_items(&self) -> bool {
+        self.flushed_count == 0
     }
 
     fn add_or_flush_chunk(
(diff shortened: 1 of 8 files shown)


kwannoel:
* **#20546** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20546?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20546?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/stream/src/executor/sync_kv_log_store.rs
```diff
@@ -350,10 +387,20 @@ impl<S: StateStore> SyncedKvLogStoreExecutor<S> {
                                                 end_seq_id,
                                             );
                                         } else {
-                                            write_future = WriteFuture::receive_from_upstream(
-                                                stream,
-                                                write_state,
-                                            );
+                                            // If buffer 90% full, pause the stream for a while, let downstream do some processing
+                                            // to avoid flushing.
+                                            if buffer.buffer.len() >= self.buffer_size * 9 / 10 {
```
wenym1, line 392: We'd better determine whether to pause by the number of rows rather than the item count.

More importantly, we shouldn't always pause the write future when the buffer is almost full, because in this way, the slowness in the reader side will always block the writer and the up-down stream won't be decoupled.

In my rough design, we should only pause when we transition from a clean in-memory state to flushed state. The clean in-memory state means when no pending data is in storage, and all chunks can be retrieved from the buffer without reading from storage, and the second state is all other circumstances. 

More specifically, we may pause for a while only when we were previously in clean in-memory state, and in either of the following scenarios:
- when receive a chunk and will write a chunk to storage
- when receive a checkpoint barrier and going to write all unflushed chunks to storage.

When this happens and we may store the item we receive when we pause, and then after the pause sleep, re-apply the item to buffer and storage.

And when we are not in the clean in-memory state, we don't have to pause.

kwannoel: > The clean in-memory state means when no pending data is in storage, and all chunks can be retrieved from the buffer without reading from storage, and the second state is all other circumstances.

Seems like we have consider case where there's historical data being read from the logstore, on top of reading from flushed chunk future.

Not sure if there's a simple way to check if there's data for the logstore in the previous checkpointed epoch, on recovery.

src/stream/src/executor/sync_kv_log_store.rs
```diff
@@ -197,10 +208,37 @@ impl<S: LocalStateStore> WriteFuture<S> {
...
         &mut self,
     ) -> StreamExecutorResult<(BoxedMessageStream, LogStoreWriteState<S>, WriteFutureEvent)> {
         match self {
+            WriteFuture::Paused {
+                sleep_future,
+                future,
+                ..
+            } => {
+                sleep_future.await;
+                let (opt, stream) = future.await;
```
wenym1, line 233: When we have passed the `sleep_future.await` and gets pending at `future.await`, if this future of `next_future` is dropped and recreated, we will still be at `WriteFuture::Paused`, and then we will poll the `sleep_future` again. However, the `sleep_future` have been ready, and the behavior of polling it is undefined, may be pending forever, or panic.

src/stream/src/executor/sync_kv_log_store.rs
```diff
@@ -356,10 +398,23 @@ impl<S: StateStore> SyncedKvLogStoreExecutor<S> {
                                                 end_seq_id,
                                             );
                                         } else {
-                                            write_future = WriteFuture::receive_from_upstream(
-                                                stream,
-                                                write_state,
-                                            );
+                                            // If buffer full, pause the stream for a while, let downstream do some processing
```
wenym1, line 401: If we are implementing the idea proposed in https://github.com/risingwavelabs/risingwave/pull/20546#discussion_r1963403734, we should not always set the write future to paused when we have flushed data, but only pause for once when we transit from clean state to flushed state.

The general logic should be like, when we are handling a message that is going to trigger a flush to storage, (either on a chunk when buffer is full, or on a checkpoint barrier when buffer has unflushed chunk), if the state is currently clean, we may pause for a while, and then reapply the message anyway.

The pause is always triggered by an upstream message. Therefore, we can store the upstream message in `WriteFuture::Pause`, and then when the sleep finish, we can yield this upstream message to reapply the message.

The handle logic should be like
```
let mut clean_state = check initial clean state.
select!{
    message = write_future => {
        match message {
            barrier => {
                if clean_state && going to flush some unflushed chunk {
                    write_future = paused(barrier);
                    clean_state = false;
                } else {
                    ...handle the barrier
                }
            }
            chunk => {
                if clean_state && buffer full {
                    write_future = paused(chunk);
                    clean_state = false;
                } else {
                    ...handle the chunk
                }
            }
        }
    }
    chunk = read_future {
        if !clean_state && now the state is clean {
            clean_state = true;
        }
        ...
    }
}
```

kwannoel:
Should have fixed the comments, will have a second look tomorrow.

I also added another read state, to try to fetch a single chunk, so we know if we are in a dirty or clean state initially. https://github.com/risingwavelabs/risingwave/pull/20546/commits/c588e6fd160a0a226b1cae74b3324f07a579dbd5.

kwannoel:
Seems I encountered some other bug. Investigating

src/stream/src/executor/sync_kv_log_store.rs
```diff
@@ -197,10 +210,33 @@ impl<S: LocalStateStore> WriteFuture<S> {
...
+            stream,
+            write_state,
+        }
+    }
+
     async fn next_event(
         &mut self,
     ) -> StreamExecutorResult<(BoxedMessageStream, LogStoreWriteState<S>, WriteFutureEvent)> {
         match self {
+            WriteFuture::Paused { sleep_future, .. } => {
```
wenym1, line 231: Seems no need to make it `Option`? Because we always overwrite it with WriteFuture::Empty after it returns successfully.

src/stream/src/executor/sync_kv_log_store.rs
```diff
@@ -272,120 +308,167 @@ impl<S: StateStore> SyncedKvLogStoreExecutor<S> {
...
-                                            // Apply Vnode Update
-                                            read_state.update_vnode_bitmap(vnode_bitmap);
-                                            initial_write_epoch = barrier_epoch;
-                                            input = stream;
-                                            initial_write_state = write_state;
-                                            continue 'recreate_consume_stream;
-                                        } else {
-                                            write_future = WriteFuture::receive_from_upstream(
+                                        if clean_state {
+                                            write_future_state = WriteFuture::paused(
```
wenym1, line 377: We don't have to always pause and set state as not clean. We can pause only when not in clean_state && checkpoint barrier && the buffer has unflushed chunk.

src/stream/src/executor/sync_kv_log_store.rs
```diff
@@ -272,120 +308,167 @@ impl<S: StateStore> SyncedKvLogStoreExecutor<S> {
...
-                                        let end_seq_id = seq_id - 1;
-                                        let epoch = write_state.epoch().curr;
-                                        if let Some(chunk_to_flush) = buffer.add_or_flush_chunk(
-                                            start_seq_id,
-                                            end_seq_id,
-                                            chunk,
-                                            epoch,
-                                        ) {
-                                            write_future = WriteFuture::flush_chunk(
+                                        if clean_state && buffer.size() + chunk.cardinality()
```
wenym1, line 430: Should be
```
if let Some(chunk) = buffer.add_or_flush_chunk(...) {
    not update seq_id, but only get start_seq_id and end_seq_id
    if clean_state {
        write_future_state = pause(chunk);
    } else {
        write_future_state = flush_chunk(chunk);
        clean_state = false;
        update seq_id
    }
}
```

src/stream/src/executor/sync_kv_log_store.rs
```diff
@@ -404,12 +487,13 @@ impl<S: StateStore> SyncedKvLogStoreExecutor<S> {
...
                                 );
                                 flush_info.report(&self.metrics);
-                                write_future =
+                                write_future_state =
                                     WriteFuture::receive_from_upstream(stream, write_state);
                             }
                         }
                     }
                     Either::Right(result) => {
+                        println!("read_stream progress");
```
wenym1, line 496: It seems that we need to update `clean_state  = true` here when some conditions are met. In current code, we have no where to reset `clean_state  = true`.

src/stream/src/executor/sync_kv_log_store.rs
kwannoel: Fixed, can just try poll on init

src/stream/src/executor/sync_kv_log_store.rs
kwannoel: Fixed, no longer poll the secondary future, thanks to @wenym1's idea below, where we can directly store the message, and replay it after pause.

src/stream/src/executor/sync_kv_log_store.rs
kwannoel: Implemented.

src/stream/src/executor/sync_kv_log_store.rs
kwannoel: Fixed, `add_or_flush_chunk` does not update any state, unless it actually flushes a chunk, so it's safe to call directly, rather than check size before.

src/stream/src/executor/sync_kv_log_store.rs
kwannoel: Fixed, reset clean state, if read future in an idle state, and if the buffer has no flushed items.

kwannoel:
Fixing some planner tests

src/stream/src/executor/sync_kv_log_store.rs
```diff
@@ -423,6 +507,10 @@ impl<S: StateStore> SyncedKvLogStoreExecutor<S> {
 }
 
 enum ReadFuture<S: StateStoreRead> {
+    ReadingInitialPersistedStream {
```
wenym1, line 510: May use `StreamExt::Peekable` to merge the two variants on reading persisted stream?

src/stream/src/executor/sync_kv_log_store.rs
```diff
@@ -272,111 +305,157 @@ impl<S: StateStore> SyncedKvLogStoreExecutor<S> {
...
                                         ) {
-                                            write_future = WriteFuture::flush_chunk(
-                                                stream,
-                                                write_state,
-                                                chunk_to_flush,
-                                                epoch,
-                                                start_seq_id,
-                                                end_seq_id,
-                                            );
+                                            if clean_state && buffer.no_flushed_items() {
```
wenym1, line 437: Seems no need to check `no_flushed_item` here, because we are sure that we may be going to flush a chunk.

src/stream/src/executor/sync_kv_log_store.rs
```diff
@@ -404,12 +483,17 @@ impl<S: StateStore> SyncedKvLogStoreExecutor<S> {
...
                                 );
                                 flush_info.report(&self.metrics);
-                                write_future =
+                                write_future_state =
                                     WriteFuture::receive_from_upstream(stream, write_state);
                             }
                         }
                     }
                     Either::Right(result) => {
+                        if matches!(read_future_state, ReadFuture::Idle)
```
wenym1, line 492: May only check the condition when `clean_state == false`.

src/stream/src/executor/sync_kv_log_store.rs
```diff
@@ -272,111 +305,157 @@ impl<S: StateStore> SyncedKvLogStoreExecutor<S> {
...
-                                            // Apply Vnode Update
-                                            read_state.update_vnode_bitmap(vnode_bitmap);
-                                            initial_write_epoch = barrier_epoch;
-                                            input = stream;
-                                            initial_write_state = write_state;
-                                            continue 'recreate_consume_stream;
-                                        } else {
-                                            write_future = WriteFuture::receive_from_upstream(
+                                        if clean_state
+                                            && buffer.no_flushed_items()
```
wenym1, line 371: I think the condition should be buffer.has_unflushed_items()? Otherwise, even for an empty buffer we will enter the unclean state.

src/stream/src/executor/sync_kv_log_store.rs
```diff
@@ -579,20 +679,27 @@ impl<S: StateStore> SyncedKvLogStoreExecutor<S> {
 
 struct SyncedLogStoreBuffer {
     buffer: VecDeque<(u64, LogStoreBufferItem)>,
+    current_size: usize,
     max_size: usize,
     next_chunk_id: ChunkId,
     metrics: KvLogStoreMetrics,
+    flushed_count: usize,
```
wenym1, line 686: Seems that we didn't update the `flushed_count` properly when we flush a chunk to storage on barrier (when we mark the `flushed` as true for the `StreamChunk` variant).

kwannoel: I intended `flushed_count` to be incremented, just for `FlushedItem`, because even if some chunk was flushed to storage on barrier, the chunk still remains in the buffer, and is not rewritten to a `FlushedItem`. So it can still be read directly from buffer.

Only `FlushedItem` will need to be read from storage.

src/stream/src/executor/sync_kv_log_store.rs
kwannoel: Hmm `futures::StreamExt` has some limitations. If we use its peekable implementation, subsequently when we call `peek` we need to `pin_mut`.

I opt for `tokio_stream::StreamExt::peekable` instead. Since it does not have this limitation. Seems like other modules in our codebase (dispatcher) also use this.

src/stream/src/executor/sync_kv_log_store.rs
kwannoel: Fixed, good catch.

src/stream/src/executor/sync_kv_log_store.rs
kwannoel: Fixed, since only one other area uses this, I just add a new method call: `is_empty`.

Then the predicate is `!buffer.is_empty() && buffer.no_flushed_items()`.

Calling `buffer.has_unflushed_items` seems ambiguous, because it implies having unflushed items, but doesn't indicate that there should be no flushed items.

src/stream/src/executor/sync_kv_log_store.rs
```diff
@@ -404,12 +471,26 @@ impl<S: StateStore> SyncedKvLogStoreExecutor<S> {
...
+                            && buffer.no_flushed_items()
+                        {
+                            clean_state = true;
+
+                            // Let write future resume immediately
+                            if let WriteFuture::Paused { sleep_future, .. } =
+                                &mut write_future_state
+                            {
+                                assert!(buffer.current_size < self.max_buffer_size);
+                                *sleep_future = Box::pin(sleep(Duration::ZERO));
```
wenym1, line 491: Not sure about whether it's illegal to sleep with zero duration. May need to add a ut to simply call `sleep(Duration::ZERO)`. Or can we just make `sleep_future` Option, and when we pull the future, if we see it's None, we can return ready immediately.

src/stream/src/executor/sync_kv_log_store.rs
wenym1: FYI, the implementation at the moment has more than one usages on `no_flushed_items`.

src/stream/src/executor/sync_kv_log_store.rs
```diff
@@ -579,20 +660,31 @@ impl<S: StateStore> SyncedKvLogStoreExecutor<S> {
...
     buffer: VecDeque<(u64, LogStoreBufferItem)>,
+    current_size: usize,
     max_size: usize,
     next_chunk_id: ChunkId,
     metrics: KvLogStoreMetrics,
+    flushed_count: usize,
 }
 
 impl SyncedLogStoreBuffer {
+    /// Returns true if there are flushed items in the buffer.
```
wenym1, line 671: To make the method name more consistent with the implementation, may just be `self.flushed_count`? We can check `self.is_empty()` in where we wanted to check `self.current_size > 0`.

src/stream/src/executor/sync_kv_log_store.rs
```diff
@@ -145,6 +152,12 @@ struct FlushedChunkInfo {
 }
 
 enum WriteFuture<S: LocalStateStore> {
+    Paused {
```
wenym1, line 155: Make add comments about the motivation of introducing the `Paused` item, and some details about the strategy. The strategy may include the condition to following actions
- pause when seeing an upstream data chunk
- pause when seeing an upstream checkpoint barrier
- unpause a paused write future when just consumed a chunk

src/stream/src/executor/sync_kv_log_store.rs
wenym1: The make things simple, I think we can just pause when previously in clean state && the barrier is checkpoint barrier && buffer is non-empty. It means simply avoid checkpoint barrier skip rows with some efforts.

wenym1 (commented):
Rest LGTM! Can merge this PR after comments are resolved.

src/stream/src/executor/sync_kv_log_store.rs
kwannoel: Option sounds cleaner. Let's go with it.

src/stream/src/executor/sync_kv_log_store.rs
kwannoel: Fixed

src/stream/src/executor/sync_kv_log_store.rs
kwannoel: Fixed, thanks

src/stream/src/executor/sync_kv_log_store.rs
kwannoel: Updated

PR #20558, opened 2025-02-21
feat(sink): change sink type to append only when plan is append-only

st1page:

When user does not specify sink type(`append-only`, `force-append-only` or `upsert`), use plan's append only perporty for sink's type.
The sink's type used in sink executor for some " chunk compact with key" behavior. We have found some unexpected compact in the flamegraph which is unnessary for an append-only sink.

## Checklist

## Documentation

<details>

</details>


src/frontend/src/optimizer/plan_node/stream_sink.rs
```diff
@@ -487,7 +487,8 @@ impl StreamSink {
         ) {
             (true, true, _) => Ok(SinkType::AppendOnly),
             (false, true, true) => Ok(SinkType::ForceAppendOnly),
-            (_, false, false) => Ok(SinkType::Upsert),
+            (true, false, false) => Ok(SinkType::AppendOnly),
```
xiangjinwu, line 490: FYI this also covers `format upsert` (in addition to omitting `type = `). So when the upstream is append-only and user explicitly writes `format upsert`, the resulting sink will also be `append-only`.

hzxa21 (approved):
Thanks for the quick fix.

src/frontend/src/optimizer/plan_node/stream_sink.rs
```diff
@@ -465,46 +466,63 @@ impl StreamSink {
...
-        ) {
-            (true, true, _) => Ok(SinkType::AppendOnly),
-            (false, true, true) => Ok(SinkType::ForceAppendOnly),
-            (_, false, false) => Ok(SinkType::Upsert),
-            (false, true, false) => {
-                Err(ErrorCode::SinkError(Box::new(Error::new(
+        if let Some(user_defined_sink_type) = user_defined_sink_type {
+            if user_defined_sink_type == SinkType::AppendOnly {
+                if user_force_append_only {
+                    return Ok(SinkType::ForceAppendOnly);
```
xiangjinwu, line 490: When upstream is append-only, and user specifies both append-only and force-append-only, the `SinkType` changed from `AppendOnly` to `ForceAppendOnly`.

src/frontend/src/optimizer/plan_node/stream_sink.rs
```diff
@@ -465,46 +466,63 @@ impl StreamSink {
...
+                                if syntax_legacy { "WITH" } else { "FORMAT ENCODE" }
+                            ),
+                        )))
+                            .into());
+                } else {
+                    return Ok(SinkType::AppendOnly);
+                }
+            }
+
+            Ok(user_defined_sink_type)
```
xiangjinwu, line 507: Regardless of upstream, when user specifies both upsert and force-append-only, it changed from returning error to `Ok(SinkType::Upsert)`.

This is the CI failure.

src/frontend/src/optimizer/plan_node/stream_sink.rs
```diff
@@ -465,46 +466,63 @@ impl StreamSink {
...
+        } else {
+            if user_force_append_only {
```
xiangjinwu, line 509: If upstream is append-only, and omitting `type = ` will result in `SInkType::AppendOnly` (the intended new behavior), isn't it more natural to tolerate an unnecessary force-append-only?

We can move this check after the match below and only reject for Upsert.

st1page:
Thanks for @xiangjinwu's carefully review and I think it much complex than I think... I convert the PR to draft and will add test to cover the branches all...

tabVersion:
I thought we had enforced the FORMAT ... ENCODE ... syntax.

st1page:
> I thought we had enforced the FORMAT ... ENCODE ... syntax.

Actually it is only good for those sink which RW should determine the encoding. At least it is not suitable for PG sink or RW sink

src/frontend/src/optimizer/plan_node/stream_sink.rs
st1page: To easier handle it... I finally choose to ban user using force append only when it is not nessarry.

tabVersion (approved):

Issue #20462, opened 2025-02-11
BUG: Unable to convert timestamp_ns to date in iceberg table


cyberchen98:
### Describe the bug

When executing a SQL query that involves a date comparison with an interval subtraction on an Iceberg table, the query fails with the following error

### Error message/log

```text
dev=> SELECT id, price, date FROM price WHERE date <= DATE '2025-02-10' - interval '5' day;
ERROR:  Failed to run the query

Caused by these errors (recent errors listed first):
  1: Scheduler error
  2: DataInvalid => Can't convert datum from timestamp_ns type to date type.
```

### To Reproduce

general table
```
dev=> CREATE TABLE price                  
    (
      id            BIGINT,
      price         NUMERIC,    
      date          DATE
    ); 
CREATE_TABLE
dev=> INSERT INTO price (id, price, date) VALUES                        
(1, 100.50, '2025-02-01'),
(2, 200.75, '2025-02-05'),
(3, 150.25, '2025-02-10');
INSERT 0 3
dev=> SELECT id, price, date FROM price WHERE date <= DATE '2025-02-10' - interval '5' day;
 id | price  |    date    
----+--------+------------
  1 | 100.50 | 2025-02-01
  2 | 200.75 | 2025-02-05
(2 rows)
```
iceberg table
```
dev=> CREATE TABLE price
    (
      id            BIGINT,
      price         NUMERIC,    
      date          DATE
    ) ENGINE = ICEBERG; 
CREATE_TABLE
dev=> INSERT INTO price (id, price, date) VALUES
(1, 100.50, '2025-02-01'),
(2, 200.75, '2025-02-05'),
(3, 150.25, '2025-02-10');
INSERT 0 3
dev=> SELECT id, price, date FROM price WHERE date <= DATE '2025-02-10' - interval '5' day;
ERROR:  Failed to run the query

Caused by these errors (recent errors listed first):
  1: Scheduler error
  2: DataInvalid => Can't convert datum from timestamp_ns type to date type.
```

### How did you deploy RisingWave?

RisingWave Cloud

### The version of RisingWave

```
dev=> select version();
                                       version                                        
--------------------------------------------------------------------------------------
 PostgreSQL 13.14.0-RisingWave-2.3.0-alpha (f9e420f8743215eff528f81bdd8478bb8467abf5)
(1 row)
```

xxhZs:
[...]

PR #19221, opened 2024-10-31
refactor(source): enhance parquet file source

wcy-fdu:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://gist.github.com/TennyZhuang/f00be7f16996ea48effb049aa7be4d66#file-rw_cla).

## What's changed and what's your intention?
This PR primarily focuses on the conversion of the `Timestamp` type. The Arrow `Timestamp` includes four types:
```

pub enum TimeUnit {
    /// Time in seconds.
    Second,
    /// Time in milliseconds.
    Millisecond,
    /// Time in microseconds.
    Microsecond,
    /// Time in nanoseconds.
    Nanosecond,
}
```
but previously we only handled one `Microsecond`. This pr makes the comparison of data types more reasonable:
- when doing data type check at column pruning, make `risingwave_common::types::DataType::Timestamp` match `arrow_schema_iceberg::DataType::Timestamp(_, _)`.
- when convert arrow field to rw field, add `Nanosecond`, `Millisecond`, `Second` type.

## Checklist

- [ ] I have written necessary rustdoc comments
- [ ] I have added necessary unit tests and integration tests
- [ ] I have added test labels as necessary. See [details](https://github.com/risingwavelabs/risingwave/blob/main/docs/developer-guide.md#ci-labels-guide).
- [ ] I have added fuzzing tests or opened an issue to track them. (Optional, recommended for new SQL features #7934).
- [ ] My PR contains breaking changes. (If it deprecates some features, please create a tracking issue to remove them in the future).
- [ ] All checks passed in `./risedev check` (or alias, `./risedev c`)
- [ ] My PR changes performance-critical code. (Please run macro/micro-benchmarks and show the results.)

- [ ] My PR contains critical fixes that are necessary to be merged into the latest release. (Please check out the [details](https://github.com/risingwavelabs/risingwave/blob/main/CONTRIBUTING.md))

## Documentation

- [ ] My PR needs documentation updates. (Please use the **Release note** section below to summarize the impact on users)

## Release note

If this PR includes changes that directly affect users or other significant modifications relevant to the community, kindly draft a release note to provide a concise summary of these changes. Please prioritize highlighting the impact these changes will have on users.

</details>

PR #20571, opened 2025-02-21
fix(iceberg): fix iceberg predicate pushdown

chenzl25:

- Only push down predicates to iceberg scan if the arguments of binary expression has the same type, because iceberg-rs doesn't full type conversion between different type of the predicate, so it would cause query failure.
- resolve https://github.com/risingwavelabs/risingwave/issues/20462 

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/optimizer/rule/batch/batch_iceberg_predicate_pushdown.rs
+++ b/src/frontend/src/optimizer/rule/batch/batch_iceberg_predicate_pushdown.rs
@@ -23,7 +23,7 @@ use iceberg::spec::Datum as IcebergDatum;
 use risingwave_common::catalog::Field;
 use risingwave_common::types::{Decimal, ScalarImpl};
 
-use crate::expr::{ExprImpl, ExprType, Literal};
+use crate::expr::{Expr, ExprImpl, ExprType, Literal};
 use crate::optimizer::plan_node::generic::GenericPlanRef;
 use crate::optimizer::plan_node::{BatchFilter, BatchIcebergScan, PlanTreeNodeUnary};
 use crate::optimizer::rule::{BoxedRule, Rule};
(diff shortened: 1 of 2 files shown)


kwannoel (approved):

src/frontend/src/optimizer/rule/batch/batch_iceberg_predicate_pushdown.rs
```diff
@@ -84,8 +84,8 @@ fn rw_literal_to_iceberg_datum(literal: &Literal) -> Option<IcebergDatum> {
             };
             Some(datum)
         }
-        ScalarImpl::Timestamp(t) => Some(IcebergDatum::timestamp_nanos(
-            t.0.and_utc().timestamp_nanos_opt()?,
+        ScalarImpl::Timestamp(t) => Some(IcebergDatum::timestamp_micros(
+            t.0.and_utc().timestamp_micros(),
```
xxchan, line 88: Are we fixing a different bug here? And do we have a test covering this branch 🤪

xxchan (approved):

src/frontend/src/optimizer/rule/batch/batch_iceberg_predicate_pushdown.rs
chenzl25: Just found that timestamp is microsecond precision for both risingwave and iceberg 😆

xxchan: `date <= DATE '2025-02-10' - interval '5' day`

So the right side changes from `timestamp_nanos` to `timestamp_micros`.
And then will we pushdown or not predicate for it after the PR?

I don't fully get the reason of the type change here. 🤔

chenzl25: > date <= DATE '2025-02-10' - interval '5' day
We won't push down this predicate because the lhs with data type `date` and the rhs is `timestamp`.
We can push down the predicate if we change it to `date <= (DATE '2025-02-10' - interval '5' day)::date`

xxchan: > Just found that timestamp is microsecond precision for both risingwave and iceberg 😆

I found iceberg has a `timestamp_ns` type. And we've supported to convert it to RisingWave `timestamp` in #19221

So it seems using either one is ok here...? 🤔

chenzl25: We can support timestamp_ns after this PR https://github.com/risingwavelabs/risingwave/pulls?q=is%3Apr+is%3Aopen+timestamp+ns

PR #20587, opened 2025-02-24
feat(frontend): improve index selection to pushdown limit

kwannoel:

Currently if the query is something like:
```
select * from t where v1 = 1 order by o_key;
```

We will only use select indexes according to `o_key`, when applying the `top_n_on_index_rule`.

We can also use conjunctive point get columns to be an index prefix to allow more index candidates.

We will include these in the order key of top n. This is so it can subsequently be rewritten from partial topn to `BatchLimit`.

Another approach could be to allow partial topn to be rewritten to `BatchLimit` directly, by checking the input `BatchScan`, to see if scan ranges are all point gets. However, this approach feels more hacky, since scan ranges are a lower level representation, specific to scanning storage.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/optimizer/plan_node/logical_scan.rs
+++ b/src/frontend/src/optimizer/plan_node/logical_scan.rs
@@ -181,6 +181,75 @@ impl LogicalScan {
             .collect()
     }
 
+    /// Return indexes can satisfy the required order.
+    /// The `prefix` refers to optionally matching columns of the index
+    /// It is unordered initially.
+    /// If any are used, we will return the fixed `Order` prefix.
+    pub fn indexes_satisfy_order_with_prefix(
+        &self,
+        required_order: &Order,
+        prefix: &HashSet<ColumnOrder>,
+    ) -> Vec<(&Rc<IndexCatalog>, Order)> {
+        let output_col_map = self
+            .output_col_idx()
+            .iter()
+            .cloned()
+            .enumerate()
+            .map(|(id, col)| (col, id))
+            .collect::<BTreeMap<_, _>>();
+        let unmatched_idx = output_col_map.len();
+        let mut index_catalog_and_orders = vec![];
+        for index in self.indexes() {
+            let s2p_mapping = index.secondary_to_primary_mapping();
+            let index_orders: Vec<ColumnOrder> = index
+                .index_table
+                .pk()
+                .iter()
+                .map(|idx_item| {
+                    let idx = match s2p_mapping.get(&idx_item.column_index) {
+                        Some(col_idx) => *output_col_map.get(col_idx).unwrap_or(&unmatched_idx),
+                        // After we support index on expressions, we need to handle the case where the column is not in the `s2p_mapping`.
+                        None => unmatched_idx,
+                    };
+                    ColumnOrder::new(idx, idx_item.order_type)
+                })
+                .collect();
+
+            let mut index_orders_iter = index_orders.iter();
+
+            // First check the prefix
+            let fixed_prefix = {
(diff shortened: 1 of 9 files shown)


src/frontend/src/optimizer/plan_node/logical_scan.rs
```diff
@@ -181,6 +181,75 @@ impl LogicalScan {
...
+            if remaining_len < required_order.column_orders.len() {
+                continue;
+            }
+
+            #[expect(clippy::disallowed_methods)]
+            for (order, other_order) in index_orders_iter.zip(required_order.column_orders.iter()) {
+                if order != other_order {
+                    continue;
+                }
+            }
```
chenzl25, line 246: Maybe we should construct an `Order` to reuse the `satisfy` logic. https://github.com/risingwavelabs/risingwave/blob/053506f9657c8eb56f14cfc77e6b2c19375b85a2/src/frontend/src/optimizer/property/order.rs#L111

src/frontend/src/optimizer/plan_node/logical_scan.rs
```diff
@@ -181,6 +181,75 @@ impl LogicalScan {
             .collect()
     }
 
+    /// Return indexes can satisfy the required order.
+    /// The `prefix` refers to optionally matching columns of the index
+    /// It is unordered initially.
+    /// If any are used, we will return the fixed `Order` prefix.
+    pub fn indexes_satisfy_order_with_prefix(
```
chenzl25, line 188: We can reimplement the `indexes_satisfy_order` method by leveraging the `indexes_satisfy_order_with_prefix` implementation.

src/frontend/src/optimizer/plan_node/logical_scan.rs
kwannoel: We could, but then we need to construct all possible orders.

Then if we have N point gets, we have O(N^2) orders to match.

Using a set here avoids that.

chenzl25: I see. I mean only replace this block with an `Order` (after pruning the prefix), because this code is almost the same with `satisfies `.

kwannoel: I guess we can make the tradeoff and collect remaining orders into a vec.

src/frontend/src/optimizer/plan_node/logical_scan.rs
```diff
@@ -181,6 +181,75 @@ impl LogicalScan {
...
+            // Next check required order
+            let remaining_len = index_orders_iter.len();
+            if remaining_len < required_order.column_orders.len() {
+                continue;
+            }
+
+            #[expect(clippy::disallowed_methods)]
+            for (order, other_order) in index_orders_iter.zip(required_order.column_orders.iter()) {
+                if order != other_order {
+                    continue;
```
chenzl25, line 244: This continue statement should go to the outer loop instead of the inner loop.  I think need to add a testcase including multiple indexes and test whether it can choose the one we want.

kwannoel: fixed

src/frontend/src/optimizer/plan_node/logical_scan.rs
kwannoel: fixed

src/frontend/src/optimizer/plan_node/logical_scan.rs
kwannoel: fxied

PR #20603, opened 2025-02-25
refactor(source): column derivation shall return `Field` rather than `PbColumnDesc`

xiangjinwu:

https://github.com/risingwavelabs/risingwave/blob/4a63ece56e30568d66b8e764ccaf8e4d97933832/src/connector/codec/src/decoder/avro/schema.rs#L75-L79

Complete the TODO above. Actually this function only derives column name and data type, where `Field` fits best. The column id is actually assigned in `create` handler - this is clearer now as `ColumnId::placeholder()` is used in the `frontend` crate rather than `connector` crate.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/common/src/catalog/column.rs
+++ b/src/common/src/catalog/column.rs
@@ -225,7 +225,7 @@ impl ColumnDesc {
     }
 
     pub fn from_field_without_column_id(field: &Field) -> Self {
-        Self::from_field_with_column_id(field, 0)
+        Self::from_field_with_column_id(field, ColumnId::placeholder().into())
     }
 
     pub fn is_generated(&self) -> bool {
--- a/src/connector/codec/src/decoder/avro/schema.rs
+++ b/src/connector/codec/src/decoder/avro/schema.rs
@@ -72,7 +72,7 @@ impl MapHandling {
 
 /// This function expects original schema (with `Ref`).
 /// TODO: change `map_handling` to some `Config`, and also unify debezium.
-pub fn avro_schema_to_column_descs(
+pub fn avro_schema_to_fields(
     schema: &Schema,
     map_handling: Option<MapHandling>,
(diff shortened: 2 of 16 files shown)


src/common/src/catalog/column.rs
```diff
@@ -225,7 +225,7 @@ impl ColumnDesc {
     }
 
     pub fn from_field_without_column_id(field: &Field) -> Self {
-        Self::from_field_with_column_id(field, 0)
+        Self::from_field_with_column_id(field, ColumnId::placeholder().into())
     }
```
BugenZhao, line 229: 🥵 Has this ever been persisted?

src/connector/codec/src/decoder/avro/schema.rs
```diff
@@ -72,11 +72,10 @@ impl MapHandling {
 
 /// This function expects original schema (with `Ref`).
 /// TODO: change `map_handling` to some `Config`, and also unify debezium.
-/// TODO: use `ColumnDesc` in common instead of PB.
 pub fn avro_schema_to_column_descs(
```
BugenZhao, line 75: Shall we also rename the function? Also for json and protobuf.

BugenZhao (approved):

src/connector/codec/src/decoder/avro/schema.rs
xiangjinwu: Good catch! 😂

src/common/src/catalog/column.rs
xiangjinwu: After this PR it is clear that this function is only called by `src/frontend/src/handler/create_source/external_schema/` and the name emphasizes `without_column_id`.

Before this PR, this utility is actually unused ...

PR #16825, opened 2024-05-20
support scaling source backfill

xxchan:
- We must support this? Otherwise scale source will break... We can't migrate actors??
- NOT TESTED

Signed-off-by: xxchan <xxchan22f@gmail.com>

I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://gist.github.com/TennyZhuang/f00be7f16996ea48effb049aa7be4d66#file-rw_cla).

## Checklist

- [ ] I have written necessary rustdoc comments
- [ ] I have added necessary unit tests and integration tests
- [ ] I have added test labels as necessary. See [details](https://github.com/risingwavelabs/risingwave/blob/main/docs/developer-guide.md#ci-labels-guide).
- [ ] I have added fuzzing tests or opened an issue to track them. (Optional, recommended for new SQL features #7934).
- [ ] My PR contains breaking changes. (If it deprecates some features, please create a tracking issue to remove them in the future).
- [ ] All checks passed in `./risedev check` (or alias, `./risedev c`)
- [ ] My PR changes performance-critical code. (Please run macro/micro-benchmarks and show the results.)

- [ ] My PR contains critical fixes that are necessary to be merged into the latest release. (Please check out the [details](https://github.com/risingwavelabs/risingwave/blob/main/CONTRIBUTING.md))

## Documentation

- [ ] My PR needs documentation updates. (Please use the **Release note** section below to summarize the impact on users)

## Release note

If this PR includes changes that directly affect users or other significant modifications relevant to the community, kindly draft a release note to provide a concise summary of these changes. Please prioritize highlighting the impact these changes will have on users.

</details>

PR #20606, opened 2025-02-25
feat: support alter source set parallelism & fix shared source scale bug

### LINKED ISSUE #19508: Support ALTER SOURCE SET PARALLELISM
part of #16003 

We've supported scaling in https://github.com/risingwavelabs/risingwave/pull/16825, but it can only be scaled under adaptive parallelism, since we lack a command for `ALTER SOURCE SET PARALLELISM` now.

xxchan:

close #19508

This is mainly just frontend work to add a new interface to call the scaling API.

We've supported scaling for shared source in #16825. But before this PR, we can only scale by adaptive parallelism (change CN parallelism and reboot).

## Checklist

## Documentation

<details>

Support `ALTER SOURCE SET PARALLELISM [TO | =] n` for shared source.

</details>


src/stream/src/executor/source/source_backfill_executor.rs
```diff
@@ -442,23 +441,7 @@ impl<S: StateStore> SourceBackfillExecutorInner<S> {
...
-                    HummockReadEpoch::Committed(epoch),
-                    TryWaitEpochOptions { table_id },
-                )
-                .await
-                .expect("failed to wait epoch");
-            STATE_TABLE_INITIALIZED.call_once(|| ());
-            tracing::info!("finished waiting for epoch: {}", epoch);
-        });
-
+        let mut state_table_initialized = false;
```
xxchan, line 444: Due to https://github.com/risingwavelabs/risingwave/issues/18312, the first barrier will be yield immediately, and the correctness doesn't hold any more.

xxchan: cc @wenym1 to take a look at the new solution.

src/frontend/src/handler/alter_parallelism.rs
```diff
@@ -76,6 +76,13 @@ pub async fn handle_alter_parallelism(
                 session.check_privilege_for_drop_alter(schema_name, &**table)?;
                 table.id.table_id()
             }
+            StatementType::ALTER_SOURCE => {
+                let (source, schema_name) =
+                    reader.get_source_by_name(db_name, schema_path, &real_table_name)?;
+
+                session.check_privilege_for_drop_alter(schema_name, &**source)?;
```
tabVersion, line 83: Do we need to add a check for whether the source is a shared one.

xxchan: you are right. Thanjs for catching this

src/stream/src/executor/source/source_backfill_executor.rs
```diff
@@ -665,11 +648,25 @@ impl<S: StateStore> SourceBackfillExecutorInner<S> {
...
+                                        // This is for self.backfill_finished() to be safe: wait until this actor can read all actors' written data.
+                                        // We wait for 2nd epoch
+                                        let epoch = barrier_epoch.prev;
+                                        tracing::info!("waiting for epoch: {}", epoch);
+                                        state_store
+                                            .try_wait_epoch(
+                                                HummockReadEpoch::Committed(epoch),
+                                                TryWaitEpochOptions { table_id },
+                                            )
+                                            .await
```
wenym1, line 661: May use `?` instead of `expect` to avoid accidental panic?

src/stream/src/executor/source/source_backfill_executor.rs
```diff
@@ -665,11 +648,25 @@ impl<S: StateStore> SourceBackfillExecutorInner<S> {
                                         .await?;
                                     }
 
+                                    if !state_table_initialized {
+                                        // This is for self.backfill_finished() to be safe: wait until this actor can read all actors' written data.
+                                        // We wait for 2nd epoch
+                                        let epoch = barrier_epoch.prev;
```
wenym1, line 654: The condition for `backfill_finished()` is that all loaded states are at `Finished` state. Is the reason for potential correctness issue that, when we fail to load the initial states written by other actors, we may get something like an empty list and then in the empty list, and the all-finished condition can be satisfied?

src/stream/src/executor/source/source_backfill_executor.rs
```diff
@@ -665,11 +648,25 @@ impl<S: StateStore> SourceBackfillExecutorInner<S> {
...
+                                        tracing::info!("waiting for epoch: {}", epoch);
+                                        state_store
+                                            .try_wait_epoch(
+                                                HummockReadEpoch::Committed(epoch),
```
wenym1, line 658: Currently we may have checkpoint and non-checkpoint barrier. When we have checkpoint frequency more than 1, it's likely that the second barrier is a non-checkpoint barrier. 

Let's say the first epoch is 1, and then the second epoch is 2, which is a non-checkpoint epoch, and then next checkpoint epoch is 5. Here the `try_wait_epoch` on epoch 2 will finish only when the epoch 5 has finished the checkpoint. However, we in-place block the whole executor, which will block handling barrier 3, 4, 5, and further block the checkpoint of epoch 5, and then cause deadlock.

The correct implementation should be waiting on the next checkpoint epoch.

src/stream/src/executor/source/source_backfill_executor.rs
wenym1: I'm thinking that shall we let all source parallelisms be aware of globally all the source splits instead of only the source splits they own? This global source split list can also be updated and maintained through barrier mutation. And then here instead of iter to load the all states written to storage, we do point-get one by one to explicitly see if all source splits exist and have finished.

src/stream/src/executor/source/source_backfill_executor.rs
```diff
@@ -442,23 +441,7 @@ impl<S: StateStore> SourceBackfillExecutorInner<S> {
...
-            let epoch = first_epoch.curr;
-            tracing::info!("waiting for epoch: {}", epoch);
-            state_store
-                .try_wait_epoch(
-                    HummockReadEpoch::Committed(epoch),
-                    TryWaitEpochOptions { table_id },
-                )
-                .await
-                .expect("failed to wait epoch");
-            STATE_TABLE_INITIALIZED.call_once(|| ());
```
wenym1, line 458: Was it correct to use the global static variable `STATE_TABLE_INITIALIZED` previously? 

IIUC, Since it exists globally, it can be called at most once in the whole lifetime of a CN process, even if we have multiple source created, and then it will always be immediately completed for subsequent sources after the first source gets completed.

If it's incorrect, we may need to fix it in previous release.

src/stream/src/executor/source/source_backfill_executor.rs
xxchan: > when we fail to load the initial states written by other actors, we may get something like an empty list and then in the empty list, and the all-finished condition can be satisfied?

exactly

> This global source split list can also be updated and maintained through barrier mutation

I'm a little confused how to implement this :eyes:

src/stream/src/executor/source/source_backfill_executor.rs
xxchan: 🤡 You are right.

xxchan: But this is hard to trigger for serious workload though. Will only happen if the Kafka topic doesn't have new incoming data, and then scaled

src/sqlparser/src/ast/ddl.rs
```diff
@@ -500,6 +519,17 @@ impl fmt::Display for AlterSourceOperation {
             AlterSourceOperation::SwapRenameSource { target_source } => {
                 write!(f, "SWAP WITH {}", target_source)
             }
+            AlterSourceOperation::SetParallelism {
+                parallelism,
+                deferred,
+            } => {
+                write!(
+                    f,
+                    "SET PARALLELISM TO {} {}",
```
BugenZhao, line 528:
```suggestion
                    "SET PARALLELISM TO {}{}",
```

src/sqlparser/src/parser.rs
```diff
@@ -3619,8 +3619,22 @@ impl Parser<'_> {
                 }
             } else if let Some(rate_limit) = self.parse_alter_source_rate_limit(false)? {
                 AlterSourceOperation::SetSourceRateLimit { rate_limit }
+            } else if self.parse_keyword(Keyword::PARALLELISM) {
+                if self.expect_keyword(Keyword::TO).is_err()
+                    && self.expect_token(&Token::Eq).is_err()
+                {
+                    return self.expected("TO or = after ALTER TABLE SET PARALLELISM");
```
BugenZhao, line 3626:
```suggestion
                    return self.expected("TO or = after ALTER SOURCE SET PARALLELISM");
```

src/stream/src/executor/source/source_backfill_executor.rs
wenym1: > > This global source split list can also be updated and maintained through barrier mutation
> 
> I'm a little confused how to implement this 👀

Currently we create actors along with barrier injection, so we can include the list of all source splits at the moment in the create actor information, and the source split change mutation will include all the changes, and then we can update source splits list.

xxchan:
There's no similar logic in source executor. 🤔

src/stream/src/executor/source/source_backfill_executor.rs
xxchan: Oh, you are quite right. I got it

tabVersion: > and the source split change mutation will include all the changes, and then we can update source splits list.

you mean carrying the fragment_id along with the assignments? I don't get why each actor need to maintain the whole upstream split list.

tabVersion (approved):
executor part LGTM

Issue #19373, opened 2024-11-13
feat(table): Not NULL constraint


st1page:
## background
https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-NOT-NULL
As a pg-compatible system, we don't implement the not null constraint. it's hard for us to report errors to the user for data coming into the table from the stream, resulting in pg-like behavior. But for some pg users, their data comes from DML statements like insert, which allows us to check and report errors with low cost.

## Tracking tasks & behaviors
- [ ] check and report errors in DML executors(`BatchInsert` and `BatchUpdate`)
    - A tricky thing is generated column which is generated in streaming execution. But it is such a corner case, I'd like to not implement it and mention it in doc.
- [ ] check and discard the NULL rows silently in StreamingMaterializeExecutor
- [ ] For CDC table I prefer to ban the not null constraint like https://github.com/risingwavelabs/risingwave/issues/19319. But it does not a big issue
- [ ] when a table with NULL constraint is sinked, or user try to use not null constraint when creating table with conn use Notice to warn user

KeXiangWang:
I'm recently working on it.

PR #20611, opened 2025-02-25
feat(dml): add not null support

KeXiangWang:

https://github.com/risingwavelabs/risingwave/issues/19373

## Checklist

## Documentation

<details>

</details>


st1page (approved):

proto/plan_common.proto
```diff
@@ -60,6 +60,8 @@ message ColumnDesc {
   ColumnDescVersion version = 10;
 
   AdditionalColumn additional_column = 11;
+
+  bool nullable = 12;
```
BugenZhao, line 64: What about https://github.com/risingwavelabs/risingwave/blob/e1c0a93350dc5874c5a2b0fc72a0456c549a4d69/proto/data.proto#L64 🤡

KeXiangWang: 🥵 I'll take a look

st1page: 😇 has it used?

KeXiangWang: Seems never used.

xiangjinwu: https://github.com/risingwavelabs/risingwave-legacy/issues/1746

KeXiangWang: Thanks for the information from Xiangjin. As mentioned in [this issue](https://github.com/risingwavelabs/risingwave-legacy/issues/1746#issue-1069133426), if we want to use this field, we would need to change our current simple data types, such as `Boolean`, to more complex types like `Boolean { nullable: bool }`. But `nullable` seems rarely used in our system and it would be quite non-trivial to adapt.

Since we don’t really use it, we should actually consider deprecating it.

PR #20613, opened 2025-02-26
test(backfill): add scale in test for arrangement backfill in background ddl

kwannoel:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## What's changed and what's your intention?

This code path is not well tested. We add a scale in test for background ddl here. Without the upstream commit, this test will fail.

Closes https://github.com/risingwavelabs/risingwave/issues/17015

## Checklist

- [ ] I have written necessary rustdoc comments.
- [ ]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  My PR contains critical fixes that are necessary to be merged into the latest release. 

## Documentation

- [ ]  My PR needs documentation updates. 

<details>
<summary><b>Release note</b></summary>

</details>

PR #20615, opened 2025-02-26
fix(log-store): rebuild decoupled sink on scale to fix data loss

wenym1:

Currently in decoupled sinks, which use KvLogStore, we handle scale in a similar mechanism to online scale in normal streaming jobs. When scale happen, the barrier with scale mutation will arrive at log writer first, and then instead of applying the scale mutation to log reader immediately, the scale mutation will be added to the buffer queue as a special item, and the reader will not apply this item until all previous items before it has been consumed.

However, this works when no sink parallelism exists in the scale. For a sink parallelism that will exit after receiving the barrier with scale mutation, unlike normal streaming jobs, it will exit without waiting for all previous items being consumed. The sink parallelisms that exist previously and start taking charge of the vnodes owned by the exited sink parallelism won't read the pending data of these transferred vnodes from log store, and then these pending data is lost.

The following example shows the data loss

||exec1 |   exec2|
|-|-----------|---------------|
||start with vnode [1] | start with vnode [2] |
||write chunk1 | write chunk2 |
|scale barrier|exit| change to own vnode [1,2]|

When exec1 receives the scale barrier, If the log reader of exec1 has not truncated on chunk1, it will not wait for chunk1 being truncated, and will only write chunk1 to storage and then exit anyway. However, when exec2 receive the scale barrier and start owning a new vnode1, it will not scan the storage on vnode1 for pending data, and therefore chunk1 is lost.

In this PR, we change to reinitialize the log reader during scale. During log reader reinitialization, it will rescan the vnode it owns, so that all pending data in the vnodes can be read. In the example, we will let exec2 scan the storage on vnode1, so that chunk1 won't lose. This is similar to offline scale in normal streaming jobs.

Besides, we also rebuild the sink during scale. This is because, in the lifecycle of a created sink, when receiving a barrier, it may assume that it has received all data at the epoch of the barrier. However, in the scenario above, when receiving the barrier and start owning new vnodes, there might still be pending data in the new vnodes, and if we yield these pending data to the sink, it will break the assumption above. Therefore, to make it simple, we will rebuild the whole sink during scale.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/bench/sink_bench/main.rs
+++ b/src/bench/sink_bench/main.rs
@@ -131,8 +131,8 @@ impl LogReader for MockRangeLogReader {
         Ok(())
     }
 
-    async fn rewind(&mut self) -> LogStoreResult<(bool, Option<Bitmap>)> {
-        Ok((false, None))
+    async fn rewind(&mut self) -> LogStoreResult<()> {
+        Err(anyhow!("should not call rewind"))
     }
 }
 
--- a/src/connector/src/sink/log_store.rs
+++ b/src/connector/src/sink/log_store.rs
@@ -186,12 +186,12 @@ pub trait LogReader: Send + Sized + 'static {
     /// Reset the log reader to after the latest truncate offset
     ///
     /// The return flag means whether the log store support rewind
-    fn rewind(
-        &mut self,
-    ) -> impl Future<Output = LogStoreResult<(bool, Option<Bitmap>)>> + Send + '_;
+    fn rewind(&mut self) -> impl Future<Output = LogStoreResult<()>> + Send + '_;
 }
 
 pub trait LogStoreFactory: Send + 'static {
+    const ALLOW_REWIND: bool;
+    const REBUILD_SINK_ON_UPDATE_VNODE_BITMAP: bool;
     type Reader: LogReader;
     type Writer: LogWriter;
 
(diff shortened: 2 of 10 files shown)


wenym1:
* **#20493** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20493?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#20615** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20615?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20615?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

kwannoel (approved):
LGTM, do we have test for this? I think it maybe important.

I recently added some tests for arrangement backfill also: https://github.com/risingwavelabs/risingwave/pull/20613 for scale in.

wenym1:
> LGTM, do we have test for this? I think it maybe important.
> 
> I recently added some tests for arrangement backfill also: #20613 for scale in.

Yes. Actually the bug was found in the deterministic scale test for sink.

PR #20617, opened 2025-02-26
feat: add some logs for recovery

xxchan:

During oncall I found a customer's cluster recovered, but didn't find any recover related log. Authough they are using an old version (2.0), it seems latest version also don't have these logs.

There's also no useful info in event log:
```
>  select timestamp, info from rw_event_logs  where event_type = 'COLLECT_BARRIER_FAIL'

│ 2025-02-26 05:20:10.745+00:00 │ {"collectBarrierFail": {"curEpoch": "8081375879954432", "error": "get error from control stream, in worker node 2: gRPC request to stream service failed: Unknown error: error reading a body from connection: connection reset", "prevEpoch": "8081375814483968"}} │
```

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/meta/src/barrier/worker.rs
+++ b/src/meta/src/barrier/worker.rs
@@ -220,7 +220,7 @@ impl GlobalBarrierWorker<GlobalBarrierWorkerContextImpl> {
 
             let paused = self.take_pause_on_bootstrap().await.unwrap_or(false);
             let paused_reason = paused.then_some(PausedReason::Manual);
-            tracing::info!("bootstrap recovery start!");
+
             self.recovery(paused_reason, None, RecoveryReason::Bootstrap)
                 .instrument(span)
                 .await;
@@ -490,8 +490,6 @@ impl<C: GlobalBarrierWorkerContext> GlobalBarrierWorker<C> {
 
             let reason = RecoveryReason::Failover(err.clone());
 
-            // We don't have span in release mode, so add an additional log here.
-            tracing::warn!(error = %err.as_report(), "failover recovery start!");
             // No need to clean dirty tables for barrier recovery,
             // The foreground stream job should cleanup their own tables.
             self.recovery(None, Some(err), reason)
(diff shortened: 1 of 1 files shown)


src/meta/src/barrier/worker.rs
```diff
@@ -490,6 +490,8 @@ impl<C: GlobalBarrierWorkerContext> GlobalBarrierWorker<C> {
 
             let reason = RecoveryReason::Failover(err.clone());
 
+            // We don't have span in release mode, so add an additional log here.
```
BugenZhao, line 493: This is not true. 🤔 Spans are only disabled on CI.

xxchan: You are right. I found the tracing span on another customer's cluster. 🤔

src/meta/src/barrier/worker.rs
```diff
@@ -724,7 +727,8 @@ impl<C: GlobalBarrierWorkerContext> GlobalBarrierWorker<C> {
                     ),
                 )
             };
-            if recovery_result.is_err() {
+            if let Err(err) = &recovery_result {
+                tracing::error!(error = %err.as_report(), "recovery failed");
```
BugenZhao, line 731: Not every step is inside the `try` block above, thus we may miss some failures here. Is this intended, or shall we improve this? cc @wenym1

wenym1: Not intended. We include all the steps.

Besides, I notice that logs printed during recovery will include the error as the tracing context and get printed out. However, the full error is printed out, including the long error stack trace, and cause the log output flooded with the stack trace. We may fix this by the way in this PR.

BugenZhao: We use `%` here to prevent the backtrace of `err` from being printed. It's possible that the message itself is constructed from another error and already includes its backtrace (which may not be expected).

Issue #20593, opened 2025-02-25
Parquet Source: Alter source_rate_limit=0 sometimes does not resume after altering back to default


kwannoel:
### Describe the bug

When altering parquet source rate limit to 0, and changing it back to `default`, the parquet source does not resume ingesting sometimes. And the compute node has to be restarted to make it resume.

### Error message/log

```text

```

PR #20622, opened 2025-02-26
fix(file source): clear counter `splits_on_fetch` when updating rate limit

wcy-fdu:

Close https://github.com/risingwavelabs/risingwave/issues/20593
How to reproduce
```
CREATE source s1(
        id int,
    ) WITH (
    connector = 's3',
    match_pattern = '*.parquet',
    s3.region_name = 'custom',
    s3.bucket_name = 'hummock001',
    s3.credentials.access = 'hummockadmin',
    s3.credentials.secret = 'hummockadmin',
    s3.endpoint_url = 'http://hummock001.127.0.0.1:9301',
    refresh.interval.sec = 1,
    ) FORMAT PLAIN ENCODE PARQUET;
create table d(id int);
create sink sink1 into d as select id from s1 with (
      type = 'append-only',
      force_append_only = 'true',
  );
alter sink sink1 set PARALLELISM to 1;
```
upload file1
`ALTER SOURCE s1 SET source_rate_limit TO 0;`
upload file2
`ALTER SOURCE s1 SET source_rate_limit TO default;`
upload file 3 4 5

File 3 4 5 cannot be read, but is in the fetch state table.

## root cause

This is a bug that only occurs when applying rate limits on the file source. After updating the rate limit, we set `need_rebuild_reader` to true, but we do not sync the modification of `splits_on_fetch`. Each time we call `replace_with_new_batch_reader`, it reads the state table of the fetch executor and updates the counter for `splits_on_fetch`. As a result, this leads to pending files being counted twice during this period. Subsequently, because `splits_on_fetch` is not zero, it causes further files to remain pending.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/executor/source/fs_fetch_executor.rs
+++ b/src/stream/src/executor/source/fs_fetch_executor.rs
@@ -248,8 +248,6 @@ impl<S: StateStore, Src: OpendalSource> FsFetchExecutor<S, Src> {
                         Either::Left(msg) => {
                             match msg {
                                 Message::Barrier(barrier) => {
-                                    let mut need_rebuild_reader = false;
-
                                     if let Some(mutation) = barrier.mutation.as_deref() {
                                         match mutation {
                                             Mutation::Pause => stream.pause_stream(),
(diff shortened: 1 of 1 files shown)


tabVersion (approved):

tabVersion:
Shall we add a test for it?

kwannoel (approved):
Great work! Thanks for finding the root cause

kwannoel:
We can add a test as per your repro

wcy-fdu:
Will complete the test in PR https://github.com/risingwavelabs/risingwave/pull/20236/ uniformly

PR #20623, opened 2025-02-26
fix(udf): fix wasm e2e test compile and update build target

wenym1:

As titled. We failed to compile with the newly release chrono version 0.4.40. Therefore, in this PR we will use a fixed version 0.4.39.

Besides, when compiling with target `wasm32-wasi`, it shows the following warning
```
warning: the `wasm32-wasi` target is being renamed to `wasm32-wasip1` and the `wasm32-wasi` target will be removed from nightly in October 2024 and removed from stable Rust in January 2025
```

Therefore, we change `wasm32-wasi` to `wasm32-wasip1` in the whole codebase.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/expr/impl/src/udf/wasm.rs
+++ b/src/expr/impl/src/udf/wasm.rs
@@ -92,6 +92,17 @@ fn create_rust(opts: CreateFunctionOptions<'_>) -> Result<CreateFunctionOutput> 
 
     let wasm_binary = std::thread::spawn(move || {
         let mut opts = arrow_udf_wasm::build::BuildOpts::default();
+        // Use the same chrono feature setting as src/expr/impl/Cargo.toml
+        // Use a fixed chrono version 0.4.39 because the latest 0.4.40 failed to compile.
+        // TODO: may avoid setting a fixed version when succeed to compile with latest chrono version
+        opts.manifest = r#"
+[dependencies]
+chrono = { version = "=0.4.39", default-features = false, features = [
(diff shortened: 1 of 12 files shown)


stdrc (approved):
Thank you very much!

src/expr/impl/src/udf/wasm.rs
```diff
@@ -92,6 +92,17 @@ fn create_rust(opts: CreateFunctionOptions<'_>) -> Result<CreateFunctionOutput>
 
     let wasm_binary = std::thread::spawn(move || {
         let mut opts = arrow_udf_wasm::build::BuildOpts::default();
+        // Use the same chrono feature setting as src/expr/impl/Cargo.toml
+        // Use a fixed chrono version 0.4.39 because the latest 0.4.40 failed to compile.
+        // TODO: may avoid setting a fixed version when succeed to compile with latest chrono version
```
BugenZhao, line 97: 🤔 I would rather we pin the dependencies to specific versions (with `Cargo.lock`) and document them to get more reproducible results.

wenym1: Agree. Maybe we can first support specifying `Cargo.lock` file used to build wasm in `arrow_udf_wasm`, and then we can include workspace `Cargo.lock` using `include_str`. For e2e_test, shall we let it reuse the workspace `Cargo.lock` with soft link to the lock file?

PR #20624, opened 2025-02-26
fix(iceberg): refine iceberg catalog parameter validation

chenzl25:

- Make `catalog.name` totally optional, since we have a default catalog name `risingwave` already.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/src/connector_common/connection.rs
+++ b/src/connector/src/connector_common/connection.rs
@@ -184,7 +184,7 @@ impl Connection for IcebergConnection {
                 let url = Url::parse(warehouse_path);
                 if url.is_err()
                     && let Some(catalog_type) = &self.catalog_type
-                    && (catalog_type == "rest" || catalog_type == "rest_rust")
+                    && matches!(self.catalog_type.as_ref(), Some("rest" | "rest_rust"))
                 {
                     // If the warehouse path is not a valid URL, it could be a warehouse name in rest catalog,
                     // so we allow it to pass here.
@@ -204,7 +204,7 @@ impl Connection for IcebergConnection {
             }
             None => {
                 if let Some(catalog_type) = &self.catalog_type
-                    && (catalog_type == "rest" || catalog_type == "rest_rust")
+                    && matches!(self.catalog_type.as_ref(), Some("rest" | "rest_rust"))
                 {
                     None
                 } else {
--- a/src/connector/src/connector_common/connection.rs
+++ b/src/connector/src/connector_common/connection.rs
@@ -183,10 +183,7 @@ impl Connection for IcebergConnection {
             Some(warehouse_path) => {
                 let url = Url::parse(warehouse_path);
                 if url.is_err()
-                    && matches!(
-                        self.catalog_type.as_ref().map(|x| x.as_str()),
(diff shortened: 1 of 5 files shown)


src/connector/src/connector_common/connection.rs
```diff
@@ -185,7 +184,7 @@ impl Connection for IcebergConnection {
                 let url = Url::parse(warehouse_path);
                 if url.is_err()
                     && let Some(catalog_type) = &self.catalog_type
-                    && catalog_type == "rest"
+                    && (catalog_type == "rest" || catalog_type == "rest_rust")
```
xxchan, line 187: nit: can be written as
```rust
matches!(self.catalog_type.as_ref(), Some("rest" | "rest_rust"))
```

src/connector/src/connector_common/connection.rs
```diff
@@ -205,7 +204,7 @@ impl Connection for IcebergConnection {
             }
             None => {
                 if let Some(catalog_type) = &self.catalog_type
-                    && catalog_type == "rest"
+                    && (catalog_type == "rest" || catalog_type == "rest_rust")
```
xxchan, line 207: ditto

src/connector/src/connector_common/iceberg/mod.rs
```diff
@@ -174,7 +173,7 @@ impl IcebergCommon {
                     }
                 }
                 None => {
-                    if catalog_type != "rest" {
+                    if catalog_type != "rest" || catalog_type != "rest_rust" {
```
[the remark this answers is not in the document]
xxchan: The Copilot's suggestion looks correct

xxchan (approved):
Others LGTM

src/connector/src/connector_common/connection.rs
```diff
@@ -132,15 +132,14 @@ pub struct IcebergConnection {
...
+    /// Path of iceberg warehouse.
     #[serde(rename = "warehouse.path")]
     pub warehouse_path: Option<String>,
     /// Catalog id, can be omitted for storage catalog or when
     /// caller's AWS account ID matches glue id
     #[serde(rename = "glue.id")]
     pub glue_id: Option<String>,
-    /// Catalog name, can be omitted for storage catalog, but
-    /// must be set for other catalogs.
+    /// Catalog name, default value is risingwave.
```
xxchan, line 142: I'm a little confused about the default value though. Will user really create a catalog called risingwave? 🤔

src/connector/src/connector_common/iceberg/mod.rs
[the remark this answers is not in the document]
chenzl25: Indeed

src/connector/src/connector_common/connection.rs
chenzl25: Actually not every catalog needs a catalog name to access it, so for these catalogs, we don't require users to provide us a catalog name. For those catalog required a catalog name like `JdbcCatalog`, they should set the catalog name. If not, we will use `risingwave` as the catalog. (Although `JdbcCatalog` would use `jdbc` as catalog name if it is unspecified).

PR #20627, opened 2025-02-27
fix(barrier): ensure causal order in local barrier manager event

wenym1:

In the event loop of local barrier manager, we need to ensure that events do not break the causal order. For example, let's say, an actor has two event, `collect barrier1` and then `collect barrier2`, each causing event of `barrier1 all collected` and `barrier2 all collected`, and then we should ensure that the local barrier manager must handle `barrier1 all collected` before `barrier2 all collected`.

In current implementation, when we poll the next barrier manager event, we will poll all ready actor events first, and apply them to the graph state, and in the end see if we can yield any `barrier all collected` event. Since we poll all ready actor events and apply them all together to the graph state first, we may lose the causal order of these ready events. Within a partial graph, the causal order of events is ensured by using an epoch queue, in which we will always check the earliest epoch first. However, in snapshot backfill, actor will migrate from the partial graph of creating job to the database partial graph when backfill finishes. Since we don't ensure the causal order between two partial graphs, the two events may be out of causal order. 

For example, let's say a backfilling actor finishes backfill in epoch1, and will migrate to the database partial graph starting from epoch2. The actor handles message very fast, and then two actor events `barrier of epoch1 collected` and `barrier of epoch2 collected` become ready at the same time and apply to the two partial graphs. Since we don't ensure order when we  check two partial graphs, we may see `barrier of epoch2 all collected` first, and then see `barrier of epoch1 all collected`, and then we will handle the epoch2 event earlier than the epoch1 event.

In this PR, we fix this issue by not polling all ready actor events all at once. Instead, when we get a ready actor event, we will apply to the graph and immediately see if there is any new event.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/task/barrier_manager/managed_state.rs
+++ b/src/stream/src/task/barrier_manager/managed_state.rs
@@ -855,10 +855,30 @@ impl DatabaseManagedBarrierState {
             let (actor_id, err) = option.expect("non-empty when tx in local_barrier_manager");
             return Poll::Ready(ManagedBarrierStateEvent::ActorError { actor_id, err });
         }
+        // yield some pending collected epochs
+        for (partial_graph_id, graph_state) in &mut self.graph_states {
+            if let Some(barrier) = graph_state.may_have_collected_all() {
+                if let Some(actors_to_stop) = barrier.all_stop_actors() {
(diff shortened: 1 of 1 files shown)


hzxa21 (approved):
Good catch!

PR #20629, opened 2025-02-27
feat(expr): add expression to reveal license status & content

BugenZhao:

Add a function `rw_license()` (returning a jsonb) to dump the license information.

If the current license is valid, the output will be like this:

```
dev=> select rw_license();
                                                 rw_license
-------------------------------------------------------------------------------------------------------------
 {"cpu_core_limit": null, "exp": 9999999999, "iss": "test.risingwave.com", "sub": "rw-test", "tier": "paid"}
(1 row)
```

Otherwise, it will yield an error indicating what's going wrong (like invalid token, expired license).

~~Note that a valid license does not mean that the paid features are available. Should also check the result of `rw_test_paid_tier()`.~~

## Checklist

## Documentation

<details>

See PR body above.

</details>

### DIFF
--- a/proto/expr.proto
+++ b/proto/expr.proto
@@ -305,6 +305,7 @@ message ExprNode {
     VNODE = 1101;
     TEST_PAID_TIER = 1102;
     VNODE_USER = 1103;
+    LICENSE = 1104;
     // Non-deterministic functions
     PROCTIME = 2023;
     PG_SLEEP = 2024;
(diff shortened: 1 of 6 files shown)


xxchan (approved):

xxchan:
> Note that a valid license does not mean that the paid features are available. Should also check the result of rw_test_paid_tier().

I don't fully get this. We can already see the `tier` info in `rw_license`. Seems no need to further check anything.

stdrc (approved):

BugenZhao:
> > Note that a valid license does not mean that the paid features are available. Should also check the result of rw_test_paid_tier().
> 
> I don't fully get this. We can already see the `tier` info in `rw_license`. Seems no need to further check anything.

Yes, you are right. 😄

PR #20632, opened 2025-02-27
fix(jdbc): remove postgres jdbc sink schema if not specified

chenzl25:

- Currently, we use `public` as the default schema for postgres jdbc sink sql. However, some of database like questdb doesn't support schema. I think we can just remove the schema from the dml if users haven't specified. 

## Checklist

## Documentation

<details>

</details>


tabVersion (commented):
I think the fix may affect the existing jobs?
If a user rely on the behavior to connect to downstream, the fix can make the job fail after upgrade.

chenzl25:
> I think the fix may affect the existing jobs? If a user rely on the behavior to connect to downstream, the fix can make the job fail after upgrade.

I think it is fine, because `public` should be the default schema.

lmatz:
SCR-20250303-mb6

looks like questdb is not alone

let's label this as a breaking change

lmatz (approved):

lmatz:
wonder if it is urgently required by a known customer?
If time allows, it is better to give a 1-version notice in advance before making a breaking change

lmatz:
Got it,
as `release-2.3` is cutoff, let's merge now and make it effective in v2.4

Issue #18372, opened 2024-09-03
jdbc sink stuck on jvm side


wenym1:
### Describe the bug

On a postgresql jdbc sink

log store read epoch does not increase for a long time

await-tree dump 
```
>> Actor 63591
Actor 63591: `xxx` [4584.598s]
  Epoch 7084692759511040 [380.010ms]
    Sink F86700000002 [380.010ms]
      consume_log (sink_id 17099) [!!! 4584.598s]
        log_sinker_send_chunk (chunk 21) [!!! 4579.358s]
        log_sinker_wait_next_response [!!! 4579.358s]
      Merge F86700000001 [380.010ms]
        LocalInput (actor 63596) [380.010ms]
        LocalInput (actor 63595) [380.010ms]
        LocalInput (actor 63594) [380.010ms]
        LocalInput (actor 63593) [380.010ms]
```

jstack
```
"Thread-6964" #7058 prio=5 os_prio=0 cpu=329.34ms elapsed=5299.86s tid=0x0000fffe4786c800 nid=0x14d1 runnable  [0x0000fffbbbab8000]
   java.lang.Thread.State: RUNNABLE
	at sun.nio.ch.Net.poll(java.base@17.0.12/Native Method)
	at sun.nio.ch.NioSocketImpl.park(java.base@17.0.12/NioSocketImpl.java:186)
	at sun.nio.ch.NioSocketImpl.park(java.base@17.0.12/NioSocketImpl.java:195)
	at sun.nio.ch.NioSocketImpl.implWrite(java.base@17.0.12/NioSocketImpl.java:420)
	at sun.nio.ch.NioSocketImpl.write(java.base@17.0.12/NioSocketImpl.java:445)
	at sun.nio.ch.NioSocketImpl$2.write(java.base@17.0.12/NioSocketImpl.java:831)
	at java.net.Socket$SocketOutputStream.write(java.base@17.0.12/Socket.java:1035)
	at sun.security.ssl.SSLSocketOutputRecord.deliver(java.base@17.0.12/SSLSocketOutputRecord.java:345)
	at sun.security.ssl.SSLSocketImpl$AppOutputStream.write(java.base@17.0.12/SSLSocketImpl.java:1308)
	at java.io.BufferedOutputStream.flushBuffer(java.base@17.0.12/BufferedOutputStream.java:81)
	at java.io.BufferedOutputStream.write(java.base@17.0.12/BufferedOutputStream.java:127)
[...]

PR #20641, opened 2025-02-27
fix(sink): reduce jdbc query timeout to 60s

chenzl25:

- Currently, JDBC sink query timeout is 600s while socket timeout is 300s. It would cause an issue if transactions are waiting for a lock, because socket timeout just close the connect from RisingWave side, but the target database connection waiting for a lock doesn't  know it and we have no way to kill the query anymore.
- So in this PR, we reduce the query  timeout from 600s to 60s to make sure it is smaller than 300s (the socket timeout value).
- Related issue https://github.com/risingwavelabs/risingwave/issues/18372

## Checklist

## Documentation

<details>

</details>

- Change jdbc sink `jdbc.query.timeout` default value from 600s to 60s.


hzxa21 (approved):
Thanks for the investigation!

wenym1 (approved):

PR #20649, opened 2025-02-27
feat(frontend): memory throttling for frontend messages

zwang28:

This PR is a follow-up to #20470. It aims to address the frontend OOM issue caused by a surge of large incoming queries, which may occur due to numerous concurrent DML or a few extremely large DML.

A `MessageMemoryManager` has been implemented in the frontend. It tracks memory usage of frontend messages and provides hint of whether this message would cause memory constraint violation. For any message of size n,
- If n <= `min_filter_bytes`, the message won't add to the memory constraint, because the message is too small that it can always be accepted. e.g. `SELECT` is usually small relative in size and should not be rejected.
- If n > `max_filter_bytes`, the message won't add to the memory constraint, because the message is too large that it's better to be rejected immediately.
- Otherwise, the message will add size n to the memory constraint. See `MessageMemoryManager::add`.

The frontend will follow `MessageMemoryManager`'s hint and respond client with error immediately, without processing the message any further.

Tests similar to those in #20470 have demonstrated the effectiveness of this PR.

## Checklist

## Documentation

<details>

</details>


src/utils/pgwire/src/pg_protocol.rs
```diff
@@ -506,15 +592,43 @@ where
                 }
             }
             FeMessage::HealthCheck => self.process_health_check(),
+            FeMessage::ServerThrottle(reason) => match reason {
+                ServerThrottleReason::TooLargeMessage => {
+                    return Err(PsqlError::ServerThrottle(anyhow::anyhow!(format!("frontend_throttling_filter_max_bytes {} has been exceeded, please either reduce the message size or increase the limit", self.message_memory_manager.max_filter_bytes)).into()));
```
BugenZhao, line 597:
```suggestion
                    return Err(PsqlError::ServerThrottle(anyhow::anyhow!("frontend_throttling_filter_max_bytes {} has been exceeded, please either reduce the message size or increase the limit", self.message_memory_manager.max_filter_bytes).into()));
```

src/utils/pgwire/src/pg_protocol.rs
```diff
@@ -506,15 +592,43 @@ where
...
+            PgProtocolState::Regular => {
+                self.stream.read_header().await?;
+                let guard = if let Some(ref header) = self.stream.read_header {
+                    let payload_len = std::cmp::max(header.payload_len, 0) as u64;
+                    let (reason, guard) = self.message_memory_manager.add(payload_len);
+                    if let Some(reason) = reason {
+                        // Release the memory ASAP.
+                        self.stream.skip_body().await?;
+                        return Ok((FeMessage::ServerThrottle(reason), None));
+                    }
```
chenzl25, line 624: Shall we drop the guard to release memory before `skip_body`? Otherwise, no other sql could run during this time.

chenzl25 (approved):
Rest LGTM!

BugenZhao:
I'd like to also take a look next week. 🥵

wcy-fdu (commented):
Motivation LGTM!
Some thoughts:
1. I suggest we add some metrics to track rejected messages, and then implement user facing alerts based on this information. We cannot assume that the user side will be able to catch errors in time, alerting from our side is a good way. @xuhui-lu is working on user facing altering.
2. Shall we  control memory only for write operations(`insert`, `delete`, `update`)? IIUC, The query operation(`select`) consumes more CPU resources. If this is not distinguished, the query may be blocked due to the large amount of writes. cc @chenzl25 for comment.

src/utils/pgwire/src/pg_message.rs
```diff
@@ -289,10 +312,35 @@ impl FeMessage {
...
+            const BUF_SIZE: usize = 1024;
+            let mut buf: Vec<u8> = vec![0; BUF_SIZE];
+            for _ in 0..(payload_len as usize) / BUF_SIZE {
+                stream.read_exact(&mut buf).await?;
+            }
+            let remain = (payload_len as usize) % BUF_SIZE;
+            if remain > 0 {
+                buf.truncate(remain);
+                stream.read_exact(&mut buf).await?;
+            }
```
BugenZhao, line 340: I think skipping with zero allocation can be achieved with a combination of [`tokio::io::copy`](https://docs.rs/tokio/latest/tokio/io/fn.copy.html) and [`tokio::io::sink`](https://docs.rs/tokio/latest/tokio/io/fn.sink.html).

src/common/src/config.rs
```diff
@@ -630,6 +630,18 @@ pub struct BatchConfig {
...
+    #[serde(default = "default::batch::frontend_max_running_message_bytes")]
+    pub frontend_max_running_message_bytes: u64,
+
+    /// A message under this threshold will not be throttled for memory constraints.
+    #[serde(default = "default::batch::frontend_throttling_filter_min_bytes")]
+    pub frontend_throttling_filter_min_bytes: u64,
+
+    /// A message exceeding this threshold will always be throttled for memory constraints.
+    #[serde(default = "default::batch::frontend_throttling_filter_max_bytes")]
+    pub frontend_throttling_filter_max_bytes: u64,
```
BugenZhao, line 644: Why are they organized into the `batch` section? Shall we introduce a new section for frontend (pgwire) servers?

src/utils/pgwire/src/pg_protocol.rs
```diff
@@ -68,6 +70,75 @@ tokio::task_local! {
     pub static CURRENT_SESSION: Weak<dyn Any + Send + Sync>
 }
 
+/// `MessageMemoryManager` tracks memory usage of frontend messages and provides hint of whether this message would cause memory constraint violation.
+///
+/// For any message of size n,
+/// - If n <= `min_filter_bytes`, the message won't add to the memory constraint, because the message is too small that it can always be accepted.
+/// - If n > `max_filter_bytes`, the message won't add to the memory constraint, because the message is too large that it's better to be rejected immediately.
+/// - Otherwise, the message will add size n to the memory constraint. See `MessageMemoryManager::add`.
+pub struct MessageMemoryManager {
```
BugenZhao, line 79: Shall we move this into a separate file?

BugenZhao (commented):
I feel like terms including "running message", "running bytes", "throttling filter" confusing and not intuitive to users. For example, users are likely unaware of what a (pgwire) "message" is and cannot control the size of the "message" either. When it comes to be user-facing, shall we switch to terms that are easier to understand?

src/utils/pgwire/src/pg_protocol.rs
xxchan: > I feel like terms including "running message", "running bytes", "throttling filter" confusing and not intuitive to users. First, it's not clear that "message" here is referencing to a concept in Postgres protocol. Then, users are likely unaware of what a (pgwire) "message" is and cannot control the size of the "message" either. When it comes to be user-facing, shall we switch to terms that are easier to understand?

Very good point. 
Recommended reading: https://developers.google.com/tech-writing/error-messages/target-audience

src/utils/pgwire/src/pg_message.rs
zwang28: > A heap-allocated copy buffer with 8 KB is created to take data from the reader to the writer

tokio::io::copy still need a copy buffer.

zwang28:
> I feel like terms including "running message", "running bytes", "throttling filter" confusing and not intuitive to users. First, it's not clear that "message" here is referencing to a concept in Postgres protocol. Then, users are likely unaware of what a (pgwire) "message" is and cannot control the size of the "message" either. When it comes to be user-facing, shall we switch to terms that are easier to understand?

Agree. I've modified those terms. Please feel free to update them further.

wcy-fdu (approved):
LGTM!

BugenZhao (approved):
LGTM!

PR #17963, opened 2024-08-07
refactor: deprecate old s3 and use use s3_v2 as default

tabVersion:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://gist.github.com/TennyZhuang/f00be7f16996ea48effb049aa7be4d66#file-rw_cla).

## What's changed and what's your intention?

as title

resolve https://github.com/risingwavelabs/risingwave/issues/16645

replace `s3` with `s3_v2` in the frontend so we can migrate the default to `s3`

## Checklist

- [ ] I have written necessary rustdoc comments
- [ ] I have added necessary unit tests and integration tests
- [ ] I have added test labels as necessary. See [details](https://github.com/risingwavelabs/risingwave/blob/main/docs/developer-guide.md#ci-labels-guide).
- [ ] I have added fuzzing tests or opened an issue to track them. (Optional, recommended for new SQL features #7934).
- [ ] My PR contains breaking changes. (If it deprecates some features, please create a tracking issue to remove them in the future).
- [x] All checks passed in `./risedev check` (or alias, `./risedev c`)
- [ ] My PR changes performance-critical code. (Please run macro/micro-benchmarks and show the results.)

- [ ] My PR contains critical fixes that are necessary to be merged into the latest release. (Please check out the [details](https://github.com/risingwavelabs/risingwave/blob/main/CONTRIBUTING.md))

## Documentation

- [x] My PR needs documentation updates. (Please use the **Release note** section below to summarize the impact on users)

## Release note

If this PR includes changes that directly affect users or other significant modifications relevant to the community, kindly draft a release note to provide a concise summary of these changes. Please prioritize highlighting the impact these changes will have on users.

please remove the prev s3 source part and rename s3_v2 into s3

</details>

PR #20658, opened 2025-02-28
feat: fully deprecate legacy s3 source

xxchan:

fs v2 is default since v2.0 https://github.com/risingwavelabs/risingwave/pull/17963

In telemetry, there's no usage of legacy s3 source at all. So it should be safe to remove code now.

## Checklist

## Documentation

May include in release note, but doc do not need change.
<details>
For legacy file source (created before 2.0), they need to be `DROP` and recreated in order for RisingWave to run.
</details>


tabVersion:
@yezizp2012 is there a way to prevent meta doing migration if it has legacy s3 jobs? It is better than fixing things after getting panic.

yezizp2012:
> @yezizp2012 is there a way to prevent meta doing migration if it has legacy s3 jobs? It is better than fixing things after getting panic.

I didn't quite understand, it should have nothing to do with migration? Do you mean to add a migration to do a sanity check? But meta will also panic, which is no different from directly deleting the code.

> In telemetry, there's no usage of legacy s3 source at all. So it should be safe to remove code now.

☝️ The information here should indicate that there is no running cluster with any legacy S3 jobs?

tabVersion:
> > @yezizp2012 is there a way to prevent meta doing migration if it has legacy s3 jobs? It is better than fixing things after getting panic.
> 
> I didn't quite understand, it should have nothing to do with migration? Do you mean to add a migration to do a sanity check? But meta will also panic, which is no different from directly deleting the code.

In my understanding, the happy path is rejecting upgrade cluster if there is legacy s3 jobs running. Let the users drop the jobs first with the old image and then do the upgrade, making everything under control. 

> > In telemetry, there's no usage of legacy s3 source at all. So it should be safe to remove code now.
> 
> ☝️ The information here should indicate that there is no running cluster with any legacy S3 jobs?

The data is from public network. Most of the prod cluster is in private network and cannot send the reports, we need to do handle them well.

xxchan:
> rejecting upgrade cluster

I think there's no such thing as "reject" now. User enters the new image version, and the migration is executed by meta. So if it fails, meta panics. Then perhaps user needs to restore to earlier version with a backup.

tabVersion:
Alright, after all, we can pause the cluster and drop the s3 job. Not a big deal.

src/connector/src/source/util.rs
```diff
@@ -49,3 +49,108 @@ pub fn fill_adaptive_split(
         )))
     }
 }
+
+/// A no-op source, which can be used to deprecate a source.
+/// When removing the code for a source, we can remove the source reader and executor,
+/// but we need to let the meta source manager running normally without panic.
+/// Otherwise the user can DROP the source.
```
xxchan, line 56: Here's an implementation to avoid meta panic

xxchan:
Since 2.3 will branch off soon, maybe we could delay this deprecation to 2.4. We can include a deprecation notice in 2.3 release note first.

src/stream/src/from_proto/source/trad_source.rs
```diff
@@ -190,20 +191,15 @@ impl ExecutorBuilder for SourceExecutorBuilder {
...
-                        system_params,
-                        source.rate_limit,
-                    )?
-                    .boxed()
+                if is_legacy_fs_connector {
+                    // Changed to default since v2.0 https://github.com/risingwavelabs/risingwave/pull/17963
+                    bail!(
+                        "legacy s3 connector is fully deprecated since v2.4.0, please DROP and recreate the s3 source.\nexecutor: {:?}",
+                        params
+                    );
```
BugenZhao, line 202: Will this block the recovery process of the whole cluster? Just to confirm, is there any way for users to drop the source in this situation?

(similar to https://github.com/risingwavelabs/risingwave/issues/17807)

BugenZhao (commented):
Rest LGTM

src/stream/src/from_proto/source/trad_source.rs
xxchan: IIRC failed recovery won't block DROP. @yezizp2012

tabVersion (approved):

src/stream/src/from_proto/source/trad_source.rs
yezizp2012: No, it won't block the DROP command. User can still run drop commands during the recovery.

BugenZhao (approved):

Issue #20625, opened 2025-02-26
Cannot CDC from a PG table with foreign key


xxchan:
### Error message/log

```text

```

cyliu0:
> cc [@lmatz](https://github.com/lmatz) [@cyliu0](https://github.com/cyliu0) Have we tested the privilege related problem for PG CDC?

I think I tested the PG CDC privilege long time ago before the PR merged. There were some problems fixed by Siyuan. But no further testing after Nov 24.

> We both tested it, and the doc probably is (partially?) written by CY
Which doc? I don't really remember the details :(

hzxa21:
> I think I tested the PG CDC privilege long time ago before the PR merged. There were some problems fixed by Siyuan. But no further testing after Nov 24.

I suggest we set up a regular test pipeline for PG CDC (following the doc's instruction) to avoid regression.

cyliu0:
RW: v2.1.4
PG: 16

I prepared the data with a superuser `postgres`. And I created the user according to https://docs.risingwave.com/integrations/sources/postgresql-cdc

The source was created successfully. But I encountered this error when I tried to create the table. Is this a problem of RW?

`INVALID_ARGUMENT: Postgres user must be the owner of table 'warehouse' to create/alter publication`

```
    CREATE USER rw REPLICATION LOGIN CREATEDB PASSWORD 'postgres';
    GRANT CONNECT ON DATABASE postgres TO rw;
    GRANT USAGE ON SCHEMA public TO rw;
    GRANT SELECT ON ALL TABLES IN SCHEMA public TO rw;
    GRANT CREATE ON DATABASE postgres TO rw;
```

<img width="783" alt="Image" src="https://github.com/user-attachments/assets/bcb82cd8-32e3-4ebf-bfb8-09a1fab5110c" />

hzxa21:
> But I encountered this error when I tried to create the table. Is this a problem of RW?
> INVALID_ARGUMENT: Postgres user must be the owner of table 'warehouse' to create/alter publication

Based on the issue description, I think yes, it is the RW problem.

xxchan:
> > But I encountered this error when I tried to create the table. Is this a problem of RW?
> > INVALID_ARGUMENT: Postgres user must be the owner of table 'warehouse' to create/alter publication
> 
> Based on the issue description, I think yes, it is the RW problem.

IIUC, the publication can be either created by RW (needs privilege), or by user (no need privilege, but pass publication name)

[...]

Issue #19319, opened 2024-11-08
Ban user defined default value on CDC 


st1page:
offline discuss and proposed by @StrikeW 
Currently, for cdc table, there are two sources of default value for each column
- from the upstream database
- user defined in the `create table()` statement
- 
Currently, if the user uses create table(*) to automatically map the upstream table, the default value of the upstream table will be used, when the user specifies the column and uses create table(v1 int, v2 int default 5) or the like to create the table, the default value will be completely determined by the table building behavior. value will be completely determined by the table creation behavior.
But things get weird when dealing with schema changes coming from upstream. At this point the default value of all columns will be overwritten by the default value of the upstream table

After some discussion, we decided to simplify the problem by disallowing users from specifying default value when creating cdc table.
At the same time, we need to get the default value from upstream when creating the cdc table when user specify column.

For the PG table
```
postgres=# \d+ t;
                                            Table "public.t"
 Column |  Type   | Collation | Nullable | Default | Storage | Compression | Stats target | Description
--------+---------+-----------+----------+---------+---------+-------------+--------------+-------------
 pk     | integer |           | not null |         | plain   |             |              |
 v1     | integer |           |          |         | plain   |             |              |
 v2     | integer |           |          | 0       | plain   |             |              |
```
If user create table in rw with 
```
create table t(pk int primary key, v1 int, v2 int) from pg_source table 'public.t';
```
The default value of v2 will be 0 which follow the upstream database.

PR #20671, opened 2025-02-28
feat: remove support for default column from CDC

xxchan:

From @BugenZhao:
This feature doesn’t make much sense because…
- all upstream data are already filled default value by CDC driver (if applicable)
- it’s only applicable for performing DML in our side, which might be rare in CDC table’s case
- only the most trivial expressions (that can be parsed by us) are supported, which is quite limited

Another motivation is #20625

apply #20626 in main

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/e2e_test/source_legacy/cdc/cdc.share_stream.slt
+++ b/e2e_test/source_legacy/cdc/cdc.share_stream.slt
@@ -215,16 +215,16 @@ statement ok
 INSERT INTO test_my_default_value VALUES (2, 'jack');
 
 query II
-select * from test_pg_default_value order by id asc;
+select * from test_my_default_value order by id asc;
 ----
 1 bugen Shanghai
 2 jack NULL
 
 statement ok
-INSERT INTO test_pg_default_value_disorderd (id) VALUES (2);
+INSERT INTO test_my_default_value_disorderd (id) VALUES (2);
 
 query II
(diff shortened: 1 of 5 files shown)


BugenZhao (approved):

KeXiangWang:
CC @st1page for visibility.

The default column behavior for CDC was first discussed in [this issue](https://github.com/risingwavelabs/risingwave/issues/19319). If we revert the default column for CDC and also prevent users from specifying default values when creating tables, they will have no way to set default values in RisingWave tables with CDC. This doesn’t seem like a major issue, but I’m not sure if any client is actually relying on this type of default columns.

st1page:
LGTM, btw generated column is needed

xxchan:
generated column is not related with this code path, right?

KeXiangWang:
> generated column is not related with this code path, right?

Yes

tabVersion:
This pr is acutually doing the same thing as https://github.com/risingwavelabs/risingwave/pull/20626
please cherry-pick to the release branch next time

xxchan:
what's the problem?

xxchan:
That PR was created earlier to provide a patched image. And this PR body
mentioned
it's apply #20626 <https://github.com/risingwavelabs/risingwave/pull/20626> in
main

PR #20672, opened 2025-02-28
feat(CLOUD-3655): Frontend communication with serverless backfill controller

CAJan93:

FE calls SBC if user defines serverless backfill. 

I also added a version with `HostAddr` in case you prefer that https://github.com/risingwavelabs/risingwave/pull/20740

## Checklist

## Documentation

<details>

</details>


src/frontend/src/handler/create_mv.rs
```diff
@@ -176,6 +182,36 @@ pub async fn handle_create_mv(
     .await
 }
 
+/// Send a provision request to the serverless backfill controller
+pub async fn provision_serverless_backfill(sbc_addr: String) -> Result<String> {
```
arkbriar, line 186: The name `provision_serverless_backfill` is a bit confusing. `provision_resource_group` is better.

src/frontend/src/handler/create_mv.rs
```diff
@@ -200,35 +236,75 @@ pub async fn handle_create_mv_bound(
     }
 
     let (table, graph, dependencies, resource_group) = {
-        let context = OptimizerContext::from_handler_args(handler_args);
-        let mut with_options = context.with_options().clone();
-
+        let mut with_options = get_with_options(handler_args.clone());
```
arkbriar, line 239: I think you could simply remove the function and lift the `context` retrieval to this line?

```suggestion
        let context = OptimizerContext::from_handler_args(handler_args);
        let mut with_options = context.with_options().clone();
```

src/frontend/src/handler/create_mv.rs
```diff
@@ -200,35 +236,75 @@ pub async fn handle_create_mv_bound(
...
 
-        if resource_group.is_some()
-            && !context
-                .session_ctx()
-                .config()
-                .streaming_use_arrangement_backfill()
-        {
-            return Err(RwError::from(ProtocolError("The session config arrangement backfill must be enabled to use the resource_group option".to_owned())));
+        let is_serverless_backfill = with_options
+            .remove(&CLOUD_SERVERLESS_BACKFILL_ENABLED.to_owned())
```
arkbriar, line 243: Remove?

src/frontend/src/handler/create_mv.rs
```diff
@@ -200,35 +236,75 @@ pub async fn handle_create_mv_bound(
...
+            Some(manager) => manager.env().sbc_address(),
+            None => "",
+        }
+        .to_owned();
+
+        if is_serverless_backfill && sbc_addr.is_empty() {
+            return Err(RwError::from(InvalidInputSyntax(
+                "Serverless Backfill is disabled on-premise. Use RisingWave cloud at https://cloud.risingwave.com/auth/signup to try this feature".to_owned(),
+            )));
+        }
```
arkbriar, line 273: 👍

src/frontend/src/handler/create_mv.rs
```diff
@@ -200,35 +236,75 @@ pub async fn handle_create_mv_bound(
...
+
+        let context = OptimizerContext::from_handler_args(handler_args);
         let has_order_by = !query.order.is_empty();
         if has_order_by {
             context.warn_to_user(r#"The ORDER BY clause in the CREATE MATERIALIZED VIEW statement does not guarantee that the rows selected out of this materialized view is returned in this order.
 It only indicates the physical clustering of the data, which may improve the performance of queries issued against this materialized view.
 "#.to_owned());
         }
 
+        tracing::debug!(resource_group = resource_group, "after calling SBC");
```
arkbriar, line 297: `after calling SBC` is only valid after calling SBC. I think you could simply print a final result here, while having the value from RPC printed on RPC completion.

src/frontend/src/handler/create_mv.rs
CAJan93: We need this. `is_serverless_backfill` is true if the user passed `cloud.serverless_backfill_enabled=true` when creating the materialised view.

arkbriar: Didn't get it. Why we need to remove it?

src/frontend/src/handler/create_mv.rs
arkbriar: Sounds good.

src/frontend/src/handler/create_mv.rs
arkbriar: Get it! Thanks!

src/frontend/src/handler/create_mv.rs
CAJan93: We also remove `let resource_group = with_options.remove(&RESOURCE_GROUP_KEY.to_owned());` because we want to check if we processed all `WITH` statements from the query: 

```
if !with_options.is_empty() {
    // get other useful fields by `remove`, the logic here is to reject unknown options.
    return Err(RwError::from(ProtocolError(format!(
        "unexpected options in WITH clause: {:?}",
        with_options.keys()
    ))));
}
```

CAJan93:
I validated the current setup with the [new rw operator implementation](https://github.com/risingwavelabs/risingwave-operator/pull/816) and the current SBC. It seems to work as expected.

src/frontend/src/session.rs
```diff
@@ -178,6 +178,9 @@ pub(crate) struct FrontendEnv {
 
     /// Memory context used for batch executors in frontend.
     mem_context: MemoryContext,
+
+    /// address of the serverless backfill controller.
+    sbc_addr: String,
```
yezizp2012, line 183: The abbreviation of `sbc` is a bit confusing and hard to maintain; at first glance, I didn't know what it meant. Can we use the full name instead?

src/frontend/src/handler/create_mv.rs
```diff
@@ -200,35 +236,77 @@ pub async fn handle_create_mv_bound(
...
-            return Err(RwError::from(ProtocolError("The session config arrangement backfill must be enabled to use the resource_group option".to_owned())));
+        let is_serverless_backfill = with_options
+            .remove(&CLOUD_SERVERLESS_BACKFILL_ENABLED.to_owned())
+            .unwrap_or_default()
+            .parse::<bool>()
+            .unwrap_or(false);
+
+        if resource_group.is_some() && is_serverless_backfill {
+            return Err(RwError::from(InvalidInputSyntax(
+                "Please do not specify serverless backfilling and resource_group together"
```
yezizp2012, line 250:
```suggestion
                "Please do not specify serverless backfilling and resource group together"
```

src/frontend/src/handler/create_mv.rs
```diff
@@ -200,35 +236,77 @@ pub async fn handle_create_mv_bound(
...
+                    return Err(RwError::from(ProtocolError(format!(
+                        "failed to provision serverless backfill nodes: {}",
+                        e.as_report()
+                    ))));
+                }
+                Ok(val) => Some(val),
+            }
+        } else {
+            resource_group
+        };
```
yezizp2012, line 287:
```suggestion
	if is_serverless_backfill {
            match provision_resource_group(sbc_addr).await {
                Err(e) => {
                    return Err(RwError::from(ProtocolError(format!(
                        "failed to provision serverless backfill nodes: {}",
                        e.as_report()
                    ))));
                }
                Ok(val) => resource_group = Some(val),
            }
        }
```

yezizp2012 (commented):
Rest generally LGTM.

yezizp2012 (approved):

PR #20691, opened 2025-03-03
feat(sink): support alter sink config

xxhZs:

as title, 

## Checklist

## Documentation

In this pr, we support alter sink with option.
we can use it with 
`ALTER  SINK <object name> CONNECTOR WITH ( ... delta props ...);
`

and we only support alter them: 
clickhouse and starrocks and deltalake and iceberg's `commit_checkpoint_interval`
kafka's 
```
"properties.allow.auto.create.topics",
"properties.batch.num.messages",
"properties.batch.size",
"properties.enable.idempotence",
"properties.max.in.flight.requests.per.connection",
"properties.message.max.bytes",
"properties.message.send.max.retries",
"properties.message.timeout.ms",
"properties.queue.buffering.max.kbytes",
"properties.queue.buffering.max.messages",
"properties.queue.buffering.max.ms",
"properties.request.required.acks",
"properties.retry.backoff.ms",
"properties.receive.message.max.bytes"
```


src/connector/src/sink/big_query.rs
```diff
@@ -536,6 +536,11 @@ impl Sink for BigQuerySink {
         self.check_column_name_and_type(big_query_schema)?;
         Ok(())
     }
+
+    fn update_config(&mut self, config: BTreeMap<String, String>) -> Result<()> {
+        self.config = BigQueryConfig::from_btreemap(config)?;
+        Ok(())
+    }
```
tabVersion, line 543: I don't think update partial props on Compute Node is a good practice, we should not rely on the state kept by CN. 
A preferred way is to do the updates on meta and dispatch a full connector props to CN.

proto/meta.proto
```diff
@@ -322,6 +322,13 @@ message RecoverRequest {}
 
 message RecoverResponse {}
 
+message AlterSinkConfigRequest {
+  uint32 sink_id = 1;
+  map<string, string> config = 2;
+}
+
+message AlterSinkConfigResponse {}
```
tabVersion, line 330: shall we make the request for general usage? eg. `AlterConnectorProps`

```proto
message AlterConnectorPropsRequest {
  enum AlterConnectorPropsObject {
    UNSPECIFIED = 0;
    SOURCE = 1;
    SINK = 2;
    CONNECTION = 3;
  }


  uint32 object_id = 1;
  map<string, string> changed_props = 2;
  AlterConnectorPropsObject object_type = 3;
}
```

proto/meta.proto
```diff
@@ -336,6 +343,7 @@ service StreamManagerService {
   rpc ApplyThrottle(ApplyThrottleRequest) returns (ApplyThrottleResponse);
   rpc Recover(RecoverRequest) returns (RecoverResponse);
   rpc ListRateLimits(ListRateLimitsRequest) returns (ListRateLimitsResponse);
+  rpc AlterSinkConfig(AlterSinkConfigRequest) returns (AlterSinkConfigResponse);
```
tabVersion, line 346: ditto

proto/stream_plan.proto
```diff
@@ -129,6 +137,8 @@ message BarrierMutation {
     ThrottleMutation throttle = 10;
     // Drop subscription on mv
     DropSubscriptionsMutation drop_subscriptions = 12;
+    // Alter sink with_option
+    AlterSinkConfigMutation alter_sink_config = 13;
```
tabVersion, line 141: shall we make it `AlterConnectorPropsMutation`, it should contain a mapping from actor_id -> full setting for new connectors

src/connector/src/sink/clickhouse.rs
```diff
@@ -493,8 +493,14 @@ impl Sink for ClickHouseSink {
     type Coordinator = DummySinkCommitCoordinator;
     type LogSinker = DecoupleCheckpointLogSinkerOf<ClickHouseSinkWriter>;
 
+    const SINK_ALTER_CONFIG_LIST: &'static [&'static str] = &["commit_checkpoint_interval"];
```
tabVersion, line 496: specifying a whitelist for each connector is troublesome and not necessary.
How about utilizing the existing props validation step when creating sink? If it successes, apply the change, reject it  otherwise.

src/connector/src/sink/mod.rs
```diff
@@ -643,6 +644,19 @@ pub trait Sink: TryFrom<SinkParam, Error = SinkError> {
         }
     }
 
+    fn validate_alter_config(config: &HashMap<String, String>) -> Result<()> {
```
tabVersion, line 647: 👎

src/frontend/src/handler/alter_sink_config.rs
```diff
@@ -0,0 +1,56 @@
...
+use super::{HandlerArgs, RwPgResponse};
+use crate::Binder;
+use crate::catalog::root_catalog::SchemaPath;
+use crate::error::Result;
+
+pub async fn handle_alter_sink_config(
+    handler_args: HandlerArgs,
+    table_name: ObjectName,
+    config: BTreeMap<String, String>,
+) -> Result<RwPgResponse> {
```
tabVersion, line 30: please add check for:

If the sink is created using CONNECTION, then the in-coming change cannot modify the props defined in the CONNECTION.

src/frontend/src/meta_client.rs
```diff
@@ -130,6 +130,9 @@ pub trait FrontendMetaClient: Send + Sync {
     async fn list_rate_limits(&self) -> Result<Vec<RateLimitInfo>>;
 
     async fn get_meta_store_endpoint(&self) -> Result<String>;
+
+    async fn alter_sink_config(&self, sink_id: u32, config: BTreeMap<String, String>)
```
tabVersion, line 134: please make the interface for more general use.

src/meta/service/src/stream_service.rs
```diff
@@ -461,4 +461,35 @@ impl StreamManagerService for StreamServiceImpl {
...
         Ok(Response::new(ListRateLimitsResponse { rate_limits }))
     }
+
+    async fn alter_sink_config(
+        &self,
+        request: Request<AlterSinkConfigRequest>,
+    ) -> Result<Response<AlterSinkConfigResponse>, Status> {
+        let request = request.into_inner();
+
+        let actor_to_apply = self
```
tabVersion, line 471: please verify the new sink config is valid first.

src/meta/src/controller/streaming_job.rs
```diff
@@ -1605,6 +1609,112 @@ impl CatalogController {
...
+
+        let mut new_config = sink.properties.clone().into_inner();
+        new_config.extend(config.clone());
+
+        {
+            let active_source = sink::ActiveModel {
+                sink_id: Set(sink_id),
+                properties: Set(risingwave_meta_model::Property(new_config)),
+                ..Default::default()
+            };
```
tabVersion, line 1649: should we also update the new sql when doing changes to props? @BugenZhao

src/stream/src/executor/sink.rs
```diff
@@ -294,6 +298,7 @@ impl<F: LogStoreFactory> SinkExecutor<F> {
         mut log_writer: W,
         actor_id: ActorId,
         rate_limit_tx: UnboundedSender<RateLimit>,
+        update_config_tx: UnboundedSender<HashMap<String, String>>,
```
tabVersion, line 301: Given that we dispatch full settings on the connector, how about reusing `rebuild_sink_tx`

tabVersion (commented):
the general idea to impl alter props LGTM

As we discussed earlier, we can make the framework for more general use and follow the syntax 

```
ALTER [ TABLE / SOURCE / SINK / CONNECTION ] <object name> CONNECTOR WITH ( ... delta props ...);
```

src/connector/src/sink/clickhouse.rs
wenym1: May not be a good idea. If we just treat it as recreating the sink, we will not be able to validate the comparison between the existing config and the new config. For example, for kafka sink, it's unlikely for us to change the target topic.

Rust's trait associated const support default value (in this PR we set as &[]), so we don't have to set the whitelist explicitly for all sinks. Only those supporting alter config should set this whitelist.

src/meta/src/controller/streaming_job.rs
```diff
@@ -1605,6 +1609,112 @@ impl CatalogController {
...
+                fragment::Column::StreamNode,
+            ])
+            .filter(fragment::Column::JobId.eq(sink_id))
+            .into_tuple()
+            .all(&txn)
+            .await?;
+        let mut fragments = fragments
+            .into_iter()
+            .map(|(id, mask, stream_node)| (id, mask, stream_node.to_protobuf()))
+            .collect_vec();
```
wenym1, line 1667:
```suggestion
        let sink_fragments = fragment
            .into_iter()
            .filter(|(_, fragment_type_mask, _)| *fragment_type_mask & FragmentTypeFlag::Sink as i32 != 0)
            .map(|(id, _, stream_node)| (id, stream_node.to_protobuf()))
            .exactly_once()
            .map_err(...)?;
```

src/meta/src/controller/streaming_job.rs
```diff
@@ -1605,6 +1609,112 @@ impl CatalogController {
...
+        });
+        assert!(
+            !fragments.is_empty(),
+            "sink id should be used by at least one fragment"
+        );
+        let fragment_ids = fragments.iter().map(|(id, _, _)| *id).collect_vec();
+        for (id, fragment_type_mask, stream_node) in fragments {
+            fragment::ActiveModel {
+                fragment_id: Set(id),
+                fragment_type_mask: Set(fragment_type_mask),
```
wenym1, line 1692: We should not reset the `fragment_type_mask`.

src/stream/src/executor/mod.rs
```diff
@@ -307,6 +308,7 @@ pub enum Mutation {
     Resume,
     Throttle(HashMap<ActorId, Option<u32>>),
     AddAndUpdate(AddMutation, UpdateMutation),
+    AlterSinkConfig(HashMap<ActorId, HashMap<String, String>>),
```
wenym1, line 311: The sink executor can match the alter sink config by its `sink_id` rather than `actor_id`. Therefore, it will be better to make it `AlterSinkConfig(SinkId, HashMap<String, String>)`.

src/meta/src/controller/streaming_job.rs
```diff
@@ -1605,6 +1609,112 @@ impl CatalogController {
             .await
     }
 
+    pub async fn update_sink_config_by_sink_id(
```
wenym1, line 1612: It's unlikely to have multiple sink fragments in a sink, so no need to return a map of multiple fragments.

Besides, if we don't store actor ids in `Mutation::AlterSinkConfig`, we don't even need to return the actor ids here.

src/meta/service/src/stream_service.rs
```diff
@@ -461,4 +461,35 @@ impl StreamManagerService for StreamServiceImpl {
...
+            .metadata_manager
+            .update_sink_config_by_sink_id(request.sink_id as i32, request.config.clone())
```
wenym1, line 473: We should't do the update sink config here. We should apply the config to meta store in `CommandContext::post_collect` after we successfully collect the mutation barrier.

src/connector/src/sink/clickhouse.rs
tabVersion: > we will not be able to validate the comparison between the existing config and the new config

why do we need do the comparison?

> For example, for kafka sink, it's unlikely for us to change the target topic.

Indeed, but I cannot find a reason to reject the user doing such change.
I've met request that they want to switch the existing pipeline the prod, then both kafka broker addr and topic name are to be changed. 

---

My point is that it makes no sense to limit what users want to change the settings, as long as they keep the connector type the same.

src/meta/service/src/stream_service.rs
tabVersion: This approach is acceptable to me, a failed barrier is likely to trigger a recover.
IIRC, we also update metadata first and then push a mutation when `alter rate limit`.

src/connector/src/sink/clickhouse.rs
tabVersion: Then how about supporting altering params in this section https://docs.risingwave.com/integrations/destinations/apache-kafka#additional-kafka-parameters as the target for this pr.

src/meta/service/src/stream_service.rs
xxhZs: do it in update_sink_config_by_sink_id

src/frontend/src/meta_client.rs
tabVersion: I think the answer is yes in this case. 
The frontend always pass the normal props in BTreeMap/HashMap and the secrets in SecretRef. 
We are not altering secret here, just making sure referencing an existing secret can work. 

```
alter sink ... connector with ( some_props = secret <secret name>, ... );
```

src/meta/service/src/stream_service.rs
yezizp2012: Got it, then I'm okay with it.

src/meta/src/stream/source_manager.rs
```diff
@@ -53,6 +53,7 @@ use crate::rpc::metrics::MetaMetrics;
 pub type SourceManagerRef = Arc<SourceManager>;
 pub type SplitAssignment = HashMap<FragmentId, HashMap<ActorId, Vec<SplitImpl>>>;
 pub type ThrottleConfig = HashMap<FragmentId, HashMap<ActorId, Option<u32>>>;
+pub type ConnectorPropsChange = HashMap<u32, HashMap<String, String>>;
```
tabVersion, line 56: please add some comments about the mapping

proto/meta.proto
```diff
@@ -322,6 +323,23 @@ message RecoverRequest {}
...
+    UNSPECIFIED = 0;
+    SOURCE = 1;
+    SINK = 2;
+    CONNECTION = 3;
+  }
+
+  uint32 object_id = 1;
+  map<string, string> changed_props = 2;
+  map<string, secret.SecretRef> changed_secret_refs = 3;
+  optional uint32 connector_conn_ref = 4;
```
tabVersion, line 337: why we need to include the secret's connection in the request? The info is available in source/sink's catalog.

src/frontend/src/handler/alter_sink_props.rs
```diff
@@ -0,0 +1,71 @@
...
+        sink.id.sink_id
+    };
+
+    let meta_client = session.env().meta_client();
+    let (resolved_with_options, _, connector_conn_ref) = resolve_connection_ref_and_secret_ref(
+        WithOptions::try_from(changed_props.as_ref() as &[SqlOption])?,
+        &session,
+        None,
+    )?;
+    let (changed_props, changed_secret_refs) = resolved_with_options.into_parts();
```
tabVersion, line 57: Please also check:

* If the sink is created using a CONNECTION:
  * if yes, the changed props are not in the CONNECTION

src/frontend/src/handler/alter_sink_props.rs
```diff
@@ -0,0 +1,71 @@
...
+    if !changed_secret_refs.is_empty() || connector_conn_ref.is_some() {
+        bail!("ALTER SINK does not support SECRET or CONNECTION now")
```
tabVersion, line 59: I think supporting secrets is must as a delivered feature.

src/stream/src/executor/sink.rs
```diff
@@ -621,6 +641,42 @@ impl<F: LogStoreFactory> SinkExecutor<F> {
                     if notify.send(()).is_err() {
                         warn!("failed to notify rebuild sink");
                     }
+                    log_reader.init().await?;
+                }
+                result = update_config_rx.recv()  => {
```
tabVersion, line 646: shall we push a new sink instance to the `rebuild_sink_tx`, it basically do the same thing as vnode change.

tabVersion (commented):
Thanks for your work! Impressive!

src/frontend/src/handler/alter_sink_props.rs
xxhZs: Will be supported in the next pr

src/frontend/src/handler/alter_sink_props.rs
xxhZs: Will be supported in the next pr with connection

proto/meta.proto
xxhZs: Because this s is a new SecretRef from alter

xxhZs:
can review again cc @tabVersion

src/connector/src/sink/deltalake.rs
```diff
@@ -293,8 +293,14 @@ impl Sink for DeltaLakeSink {
     type Coordinator = DeltaLakeSinkCommitter;
     type LogSinker = DecoupleCheckpointLogSinkerOf<CoordinatedSinkWriter<DeltaLakeSinkWriter>>;
 
+    const SINK_ALTER_CONFIG_LIST: &'static [&'static str] = &["commit_checkpoint_interval"];
     const SINK_NAME: &'static str = DELTALAKE_SINK;
 
+    fn update_config(&mut self, config: BTreeMap<String, String>) -> Result<()> {
+        self.config = DeltaLakeConfig::from_btreemap(config)?;
+        Ok(())
+    }
```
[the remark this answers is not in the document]
tabVersion: It does not make sense to me, we dispatch connector props to the running actor and expect to align the code path with creating sink. Ideally, we recreate the inner part (data writer) without touching the log reader, or at the expense of rewinding to the latest ckpt.

proto/stream_plan.proto
```diff
@@ -110,6 +110,14 @@ message DropSubscriptionsMutation {
   repeated SubscriptionUpstreamInfo info = 1;
 }
 
+message ConnectorPropsInfo {
+  map<string, string> connector_props_info = 1;
+}
+
+message ConnectorPropsChangeMutation {
+  map<uint32, ConnectorPropsInfo> connector_props_infos = 1;
+}
```
tabVersion, line 119: shall we move the `ConnectorPropsInfo` to the inside of `ConnectorPropsChangeMutation`? Like

```
message ConnectorPropsChangeMutation {
  message ConnectorPropsInfo {
    map<string, string> connector_props_info = 1;
  }

  map<uint32, ConnectorPropsInfo> connector_props_infos = 1;
}
```

because the type is relatively simple and won't be reused elsewhere.

tabVersion: & please add some comments about the concept mapping

src/meta/src/controller/streaming_job.rs
```diff
@@ -1656,6 +1659,137 @@ impl CatalogController {
...
+        new_config.extend(props);
+
+        let definition = sink.definition.clone();
+        let entries: &Vec<String> = &new_config
+            .iter()
+            .map(|(k, v)| format!("{} = {}", k, v))
+            .collect();
+        let new_with_definition = format!("WITH ( {} )", entries.join(", "));
+        let re = Regex::new(r"(?i)WITH\s?\([^\)]*\)").unwrap();
+        let new_definition = re.replace_all(&definition, new_with_definition);
```
tabVersion, line 1716: shall we re-generate the sql here using `CreateSinkStatement`, instead of magic.

src/stream/src/executor/sink.rs
```diff
@@ -616,17 +633,54 @@ impl<F: LogStoreFactory> SinkExecutor<F> {
...
+                                            error = %rewind_err.as_report(),
+                                            "fail to rewind log reader for alter sink config "
+                                        );
+                                        Err(anyhow!("fail to rewind log after alter table").into())
+                                    }
+                                }
+                            } else {
+                                sink_param.properties = config.into_iter().collect();
+                                sink = TryFrom::try_from(sink_param.clone()).map_err(|e| StreamExecutorError::from((e, sink_param.sink_id.sink_id)))?;
+                                Err(anyhow!("alter sink config need to rebuild sink.").into())
```
tabVersion, line 668: Does it mean drop and recreate the sink?

xxhZs: yes

xxhZs: I think we should use the validate config method for validation, rather than recreate sink here, because essentially, the behavior here is the same as the previous update config—it's just updating the configuration. One key reason for writing meta first is that, in the absence of a log store, we need to rely on recovery to rebuild the sink.

tabVersion: Then the error message is quite misleading, `alter sink config need to rebuild sink` means drop and recreate, but users are expected to trigger a recovery to apply the change.

src/meta/src/controller/streaming_job.rs
```diff
@@ -1656,6 +1661,157 @@ impl CatalogController {
...
+                add_sql_options
+                    .iter()
+                    .map(|sql_option| (&sql_option.name, sql_option)),
+            );
+            stmt.with_properties.0 = new_sql_options.into_values().cloned().collect();
+        } else {
+            return Err(SinkError::Config(anyhow!(
+                "sink definition is not a create sink statement"
+            ))
+            .into());
```
tabVersion, line 1734: the statement MUST be a create sink statement, panic here is okay

src/sqlparser/src/ast/mod.rs
```diff
@@ -2953,6 +2954,48 @@ impl fmt::Display for SqlOption {
     }
 }
 
+impl TryFrom<(&String, &String)> for SqlOption {
+    type Error = ParserError;
+
+    fn try_from((name, value): (&String, &String)) -> Result<Self, Self::Error> {
+        let mut tokenizer = Tokenizer::new(value);
```
tabVersion, line 2961: shall we reuse `parse_sql_option` for this case?

xxhZs: The string must be constructed with 'with' to be usable, so parsing the map directly seems simpler.

tabVersion: Here we handle CONNECTION specifically so maintaining the logic twice seems error prone to me.

src/sqlparser/src/ast/mod.rs
```diff
@@ -2953,6 +2954,48 @@ impl fmt::Display for SqlOption {
...
+                    }
+                    _ => connection_name,
+                };
+                SqlOptionValue::ConnectionRef(ConnectionRefValue { connection_name })
+            } else {
+                parser.parse_value_and_obj_ref::<false>().map_err(|e| {
+                    ParserError::ParserError(format!("Failed to parse value in option {}", e))
+                })?
+            }
+        };
```
tabVersion, line 2991: maybe a extract it into a individual func and share usage with `parse_sql_option`  because we do some spec logic for connection.

src/sqlparser/src/ast/mod.rs
xxhZs: I found formatting wasn't hard, so I formatted first and then parse

xxhZs:
can review again cc @tabVersion

proto/meta.proto
tabVersion: I dont get it. Are we allowing to change the CONNECTION used by existing SINK?

src/stream/src/executor/mod.rs
```diff
@@ -739,6 +742,24 @@ impl Mutation {
                     )
                     .collect(),
             }),
+            Mutation::ConnectorPropsChange(map) => {
+                PbMutation::ConnectorPropsChange(ConnectorPropsChangeMutation {
+                    connector_props_infos: map
+                        .iter()
+                        .map(|(actor_id, options)| {
+                            (
+                                *actor_id,
```
tabVersion, line 751: the actor_id here is actually sink_id when handling mutation in sink exec, we need a fix for code maintain

src/frontend/src/handler/alter_sink_props.rs
tabVersion: please remember to follow up on this

you can find some useful funcs in https://github.com/risingwavelabs/risingwave/pull/20780

tabVersion:
should add some e2e tests to make the feature functionality on the track

PR #7895, opened 2023-02-14
feat(source): support debezium avro format

hzxa21:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://gist.github.com/TennyZhuang/f00be7f16996ea48effb049aa7be4d66#file-rw_cla).

## What's changed and what's your intention?

Support debezium avro format with schema registery. Example:
```
CREATE table dbz (id integer primary key) with (
    connector = 'kafka',
    topic = 'dbserver1.inventory.customers',
    properties.bootstrap.server = '127.0.0.1:9092',
    scan.startup.mode = 'earliest'
) row format debezium_avro row schema location CONFLUENT SCHEMA REGISTRY 'http://localhost:8081';
```

Known limitations of this PR:

- We must use CREATE TABLE syntax with primary key specified to create dbz avro source
- Schema resolution fails when calling the from_avro_datum even when reader_schema == writer_schame. This is a limitation in apache-avro rust lib. We can ignore it for now since we haven't support source table schema evolution.

Close #7892 #7904 

## Checklist

- [x] I have written necessary rustdoc comments
- [x] I have added necessary unit tests and integration tests
~~- [ ] I have added fuzzing tests or opened an issue to track them. (Optional, recommended for new SQL features).~~
- [x] I have demonstrated that backward compatibility is not broken by breaking changes and created issues to track deprecated features to be removed in the future. (Please refer to the issue)
- [x] All checks passed in `./risedev check` (or alias, `./risedev c`)

### Types of user-facing changes

Please keep the types that apply to your changes, and remove the others.

- Connector (sources & sinks)

### Release note
```
CREATE TABLE [ IF NOT EXISTS ] source_name (
   pk_column_name data_type PRIMARY KEY
) 
WITH (
   connector='kafka',
   field_name='value', ...
) 
ROW FORMAT DEBEZIUM_AVRO
ROW SCHEMA LOCATION CONFLUENT SCHEMA REGISTRY '<schema_registry_url>';
```

PR #20699, opened 2025-03-04
fix(source): debezium avro schema evolution

xiangjinwu:

In short, a table with `format debezium encode avro` did not support schema evolution and would panic due to mix use of reader and writer schema.

When the feature was introduced by #7895, the avro lib (0.14.0) was [unable](https://gist.github.com/xiangjinwu/fa9ad88e11061553abd0d5e8e4c2e84f) to handle schema evolution for debezium (which contains `Ref` within nullable `Union`). The message was decoded with writer schema but the SQL columns were derived from reader schema. Such mismatch can lead to panics.

Now the avro lib (0.16.0) is able to handle schema evolution in such case, so we pass both writer and reader schema to make sure all subsequent processing only needs to handle reader schema.

The implementation on main also confuses key schema and value schema. Refer to the inline comments on how they were wrong.

## Checklist

- [x]  My PR contains critical fixes that are necessary to be merged into the latest release. 

## Documentation

<details>

</details>


src/connector/src/parser/debezium/avro_parser.rs
```diff
@@ -35,34 +35,27 @@ use crate::schema::schema_registry::{
...
+            writer_schema.as_ref(),
+            &mut raw_payload,
+            Some(&self.reader_schema.original_schema),
+        )?);
         Ok(AccessImpl::Avro(AvroAccess::new(
-            self.value.as_mut().unwrap(),
-            // Assumption: Key will not contain reference, so unresolved schema can work here.
-            AvroParseOptions::create(match self.encoding_type {
-                EncodingType::Key => self.key_schema.as_mut().unwrap(),
-                EncodingType::Value => &self.schema.original_schema,
```
xiangjinwu, line 64: The key uses writer schema, with reader schema dropped by `..` below; while the value uses reader schema.

tabVersion (approved):
LGTM

e2e_test/source_inline/kafka/avro/debezium_evolution.slt
```diff
@@ -0,0 +1,173 @@
...
+system ok
+rpk topic produce --schema-key-id=topic --schema-id=topic -f '%k %v\n' dbserver1.inventory.t <<EOF
+{"f1":1,"f3":3} {"before":null,"after":{"dbserver1.inventory.t.Value":{"f1":1,"f2":{"string":"zxc"},"f3":3}},"source":{"version":"2.7.4.Final","connector":"postgresql","name":"dbserver1","ts_ms":1740989137107,"snapshot":{"string":"false"},"db":"test","sequence":{"string":"[\"28875024\",\"29156768\"]"},"ts_us":{"long":1740989137107480},"ts_ns":{"long":1740989137107480000},"schema":"inventory","table":"t","txId":{"long":820},"lsn":{"long":29156768},"xmin":null},"transaction":null,"op":"c","ts_ms":{"long":1740989137572},"ts_us":{"long":1740989137572365},"ts_ns":{"long":1740989137572365000}}
+EOF
+
+statement ok
+create table v1 with (${RISEDEV_KAFKA_WITH_OPTIONS_COMMON}, topic = 'dbserver1.inventory.t') format debezium encode avro (schema.registry = '${RISEDEV_SCHEMA_REGISTRY_URL}');
+
+system ok
+sr_register dbserver1.inventory.t-value AVRO <<EOF
```
tabVersion, line 83: better add some notes about adding `f4` column to the schema.

PR #20715, opened 2025-03-04
feat(frontend,meta,stream): explain analyze stream job

### LINKED ISSUE #20721: Explain Analyze Stream Job
Explain analyze for stream jobs

- [x] Basic implementation of explain analyze: https://github.com/risingwavelabs/risingwave/pull/20715
- [x] Support changing analyze duration.
- [x] Decouple rendering and construction of the graph.
- [ ] Populate merge and dispatcher metrics.
- [x] Include operator epoch as a column, to show lag, where barrier is stuck. Not very useful, removed.
- [ ] Add telemetry: profiling duration, type of jobs being analyzed
- [ ] Add some e2e explain analyze tests
- [ ] Add dispatcher type e.g. HashDispatcher, in the identity column.
- [ ] Add join type (left / right / outer) in the identity column.
- [x] Operator level profiling: join, agg can be bottlenecks. How can we help user to figure out these bottlenecks?

Optional
- [ ] Use tokio idle actor metrics.
- [ ] Add a verbose option, to display slowest actors, and their throughput / backpressure vs avg. Can use to detect skew.
- [ ] Support explain analyze actor id. This will show the metrics for the entire actor, can include idle duration etc...
- [x] Explain analyze await-tree (separate issue)
- [x] Maybe improve GC algorithm for metrics
- [x] or remove the need for GC entirely -- This can be done by adding RPCs to enable / disable profiling + synchronizing channel per actor.
- [ ] Give notice / progress bar on how long more it takes for analyze to complete.

kwannoel:

Closes #20721 

Support explain analyze stream job. See release notes for an example.

<details>

<summary><b>DEMO</b></summary>

https://github.com/user-attachments/assets/90bde9ff-c0d3-40dd-b33f-3aab45291c67

</details>

### Major changes

In this PR every executor will collect profiling metrics, but only 1 set of metrics is allocated per **operator**. For instance, for hash aggregation, we could have 8 parallelism, and hence 8 executors running. But all 8 executors will share a single set of metrics, to avoid high memory consumption when actor counts are high.

When some executor is dropped, it also decrements the metric's reference count. But similar to prometheus metrics, `StreamingMetrics` will still hold a reference to these, and we have hanging metrics as a result. So periodically, for each CN and each metric group, we will run GC on metrics, to check the reference counts and evict entries which are only referenced by the hashmap itself.
We may use `Weak` pointers for some improvement, but the entry itself still needs to be GC'ed eventually.

As such, we can always keep the profiling metric collection on. The alternative would be to add `start` and `stop` RPCs to CNs, and we can reuse the prometheus framework. However we will need some synchronization inside the executor wrappers, to toggle metrics collection. We cannot just leave it on because multiple actors can't share the same metric label. Maintaining the extra RPC + synchronization mechanism seems worst than doing simple GC. The tradeoff is that GC will of course incur CPU usage periodically. Open to suggestions on this part.

Frontend will send an RPC call to meta to list fragments and get the streaming worker nodes.

It will then send an RPC call to the streaming worker nodes to collect streaming stats **twice**. First to get a baseline, second to compute the delta.

Then we will use the fragments from meta to construct a streaming graph, and map metrics to each operator in the graph.

### What's Next

- [ ] Add telemetry: profiling duration, type of jobs being analyzed
- [ ] Use tokio actor metrics.
- [ ] Add some explain analyze tests for the plan node column.
- [ ] Maybe improve GC algorithm for metrics
- [ ] or remove the need for GC entirely -- This can be done by adding RPCs to enable / disable profiling + synchronizing channel per actor.
- [ ] Support changing analyze duration.
- [ ] Give notice / progress bar on how long more it takes for analyze to complete.
- [ ] Populate merge and dispatcher metrics.
- [ ] Add dispatcher type e.g. HashDispatcher, in the identity column.
- [ ] Add join type (left / right / outer) in the identity column.
- [ ] Include operator epoch as a column, to show lag.

## Checklist

## Documentation

<details>

Syntax:
```
EXPLAIN ANALYZE (TABLE <name> | MATERIALIZED VIEW <name> | SINK <name> | INDEX <name> | ID <job_id>)
```

Example:
```
CREATE TABLE t(v1 int);

CREATE MATERIALIZED VIEW m1 as 
  select * from 
    t x join t y using(v1)
      join t z using(v1) 
        join t c using(v1);
```

Output:
```
EXPLAIN ANALYZE MATERIALIZED VIEW m1
 operator_id |                    identity                    |  actor_ids  | output_rps | avg_output_pending_ratio
-------------+------------------------------------------------+-------------+------------+--------------------------
 12884901902 | StreamMaterialize                              | 12,11,10,9  | 41589.1    | 0.00005
 12884901901 | └─ MergeExecutor                               | 12,11,10,9  | 41589.1    | 0.8149749999999999
 4           |    └─ Dispatcher                               |             | 0          | NaN
 17179869196 |       └─ StreamHashJoin                        | 16,15,14,13 | 41560.1    | 0.859225
 17179869195 |          ├─ StreamHashJoin                     | 16,15,14,13 | 152.6      | 0.93265
 17179869194 |          │  ├─ MergeExecutor                   | 16,15,14,13 | 25.6       | 1.6336749999999998
 9           |          │  │  └─ Dispatcher                   |             | 0          | NaN
 38654705673 |          │  │     └─ StreamFilter              | 20,19,18,17 | 0          | 0
 38654715839 |          │  │        └─ StreamTableScan        | 20,19,18,17 | 0          | 0
 38654715838 |          │  │           ├─ BatchPlanNode       | 20,19,18,17 | 0          | 0
 38654705664 |          │  │           └─ Upstream            | 20,19,18,17 | 0          | 0
 17179869192 |          │  └─ MergeExecutor                   | 16,15,14,13 | 0          | 0
 8           |          │     └─ Dispatcher                   |             | 0          | NaN
 34359738375 |          │        └─ StreamFilter              | 28,27,26,25 | 0          | 0
 34359748538 |          │           └─ StreamTableScan        | 28,27,26,25 | 0          | 0
 34359748537 |          │              ├─ BatchPlanNode       | 28,27,26,25 | 0          | 0
 34359738368 |          │              └─ Upstream            | 28,27,26,25 | 0          | 0
 17179869190 |          └─ MergeExecutor                      | 16,15,14,13 | 51.2       | 0.71425
 5           |             └─ Dispatcher                      |             | 0          | NaN
 21474836485 |                └─ StreamHashJoin               | 24,23,22,21 | 51.2       | 0.711325
 21474836484 |                   ├─ MergeExecutor             | 24,23,22,21 | 0          | 0
 7           |                   │  └─ Dispatcher             |             | 0          | NaN
 30064771075 |                   │     └─ StreamFilter        | 36,35,34,33 | 0          | 0
 30064781234 |                   │        └─ StreamTableScan  | 36,35,34,33 | 0          | 0
 30064781233 |                   │           ├─ BatchPlanNode | 36,35,34,33 | 0          | 0
 30064771072 |                   │           └─ Upstream      | 36,35,34,33 | 0          | 0
 21474836482 |                   └─ MergeExecutor             | 24,23,22,21 | 0          | 0
 6           |                      └─ Dispatcher             |             | 0          | NaN
 25769803777 |                         └─ StreamFilter        | 32,31,30,29 | 0          | 0
 25769813933 |                            └─ StreamTableScan  | 32,31,30,29 | 0          | 0
 25769813932 |                               ├─ BatchPlanNode | 32,31,30,29 | 0          | 0
 25769803776 |                               └─ Upstream      | 32,31,30,29 | 0          | 0
(32 rows)
```

This feature has a feature gate:
```
[streaming.developer]
enable_explain_analyze_stats = true # change to false to disable profiling cluster-wide.
```

</details>

### DIFF
--- a/src/config/example.toml
+++ b/src/config/example.toml
@@ -169,6 +169,7 @@ stream_enable_shared_source = true
 stream_switch_jdbc_pg_to_native = false
 stream_max_barrier_batch_size = 1024
 stream_hash_join_entry_state_max_rows = 30000
+enable_explain_analyze_stats = false
 
 [storage]
 share_buffers_sync_parallelism = 1
--- a/src/config/example.toml
+++ b/src/config/example.toml
@@ -169,7 +169,7 @@ stream_enable_shared_source = true
 stream_switch_jdbc_pg_to_native = false
 stream_max_barrier_batch_size = 1024
 stream_hash_join_entry_state_max_rows = 30000
-enable_explain_analyze_stats = false
+enable_explain_analyze_stats = true
 
 [storage]
 share_buffers_sync_parallelism = 1
--- a/src/config/example.toml
+++ b/src/config/example.toml
@@ -169,7 +169,7 @@ stream_enable_shared_source = true
 stream_switch_jdbc_pg_to_native = false
 stream_max_barrier_batch_size = 1024
 stream_hash_join_entry_state_max_rows = 30000
-enable_explain_analyze_stats = true
+stream_enable_explain_analyze_stats = true
 
 [storage]
 share_buffers_sync_parallelism = 1
--- a/src/frontend/src/handler/explain.rs
+++ b/src/frontend/src/handler/explain.rs
@@ -292,6 +292,8 @@ pub async fn handle_explain(
     analyze: bool,
 ) -> Result<RwPgResponse> {
     if analyze {
+        // NOTE(kwannoel): This path is for explain analyze on stream and batch queries.
+        // For existing stream jobs, see the handler module `explain_analyze` instead.
         bail_not_implemented!(issue = 4856, "explain analyze");
     }
     if options.trace && options.explain_format == ExplainFormat::Json {
--- a/src/frontend/src/handler/explain_analyze_stream_job.rs
+++ b/src/frontend/src/handler/explain_analyze_stream_job.rs
@@ -1,4 +1,6 @@
-use std::ops::Deref;
+use std::collections::{HashMap, HashSet};
+
+use risingwave_pb::meta::list_table_fragments_response::FragmentInfo;
 
 use crate::error::Result;
 use crate::handler::{HandlerArgs, RwPgResponse};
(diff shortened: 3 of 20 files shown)


hzxa21:
cc @neverchanje FYI, we may consider incorporating this in the cloud UI after it is merged.

src/frontend/src/handler/explain_analyze_stream_job.rs
```diff
@@ -0,0 +1,426 @@
...
+        .collect::<Vec<_>>();
+    Ok(stream_worker_nodes)
+}
+
+type OperatorId = u64;
+
+#[derive(Default)]
+struct StreamNodeMetrics {
+    operator_id: OperatorId,
+    epoch: u32,
```
kwannoel, line 187: We can fix this in subsequent PRs, when recording epoch.

src/frontend/src/handler/explain.rs
```diff
@@ -292,6 +292,8 @@ pub async fn handle_explain(
     analyze: bool,
 ) -> Result<RwPgResponse> {
     if analyze {
+        // NOTE(kwannoel): This path is for explain analyze on stream and batch queries.
```
kwannoel, line 295: e.g.
```
explain analyze create materialized view m1 as select * from t;
explain analyze select * from t;
```

src/stream/src/task/mod.rs
```diff
@@ -227,15 +227,3 @@ impl SharedContext {
...
 }
-
-/// Generate a globally unique executor id.
-pub fn unique_executor_id(actor_id: u32, operator_id: u64) -> u64 {
-    assert!(operator_id <= u32::MAX as u64);
-    ((actor_id as u64) << 32) + operator_id
-}
-
-/// Generate a globally unique operator id.
-pub fn unique_operator_id(fragment_id: u32, operator_id: u64) -> u64 {
```
kwannoel, line 238: Reused in frontend, for `explain_analyze` to construct the operator graph.

src/stream/src/executor/wrapper/stream_node_metrics.rs
```diff
@@ -0,0 +1,42 @@
...
+
+use futures_async_stream::try_stream;
+use tokio::time::Instant;
+
+use crate::executor::monitor::ProfileMetricsExt;
+use crate::executor::prelude::*;
+
+#[try_stream(ok = Message, error = StreamExecutorError)]
+pub async fn stream_node_metrics(
+    enable_explain_analyze_stats: bool,
```
kwannoel, line 23: Notice how we use `enable_explain_analyze_stats` before this commit: https://github.com/risingwavelabs/risingwave/pull/20715/commits/879c5ad61d90fd6b3bd088e92930e17b51d9e51d.

We had the following structure:
```rs
if enable_explain_analyze_stats {
    #[for_await]
  for m in stream { 
    yield m?; 
  }
} else {
    #[for_await]
  for m in stream { 
    // do profiling
    yield m?; 
  }
}
```

But this leads to stack overflow. I think it prevents some inlining optimizations from happening, and the result is that the generated state machine struct for `stream_node_metrics` stream is really large (maybe x2?). Hence I changed it here to avoid the branch.

kwannoel:
```
  #[for_await]
  for m in stream { 
    // do profiling _conditionally_
    yield m?;
  }
```

kwannoel: No more stack overflow is encountered.

kwannoel: (I used lldb to pinpoint it to this exact function call)

BugenZhao: Would you try `Box::pin` first? This should help.

kwannoel: I think it will move the problem from stack to heap, and consume more memory there. So I try to avoid it with this solution.

src/stream/src/executor/wrapper.rs
```diff
@@ -17,28 +17,37 @@ use crate::executor::prelude::*;
...
 pub struct WrapperExecutor {
+    operator_id: u64,
+
     input: Executor,
 
     actor_ctx: ActorContextRef,
 
     enable_executor_row_count: bool,
+
+    enable_explain_analyze_stats: bool,
```
chenzl25, line 34: cc @BugenZhao Could you please provide insights on how the performance impact of enabling enable_explain_analyze_stats compares to that of enable_executor_row_count based on your experience?

src/stream/src/executor/monitor/in_mem_stats.rs
```diff
@@ -0,0 +1,86 @@
...
+        let inner = Arc::new(RwLock::new(HashMap::new()));
+        #[cfg(all(not(test), not(madsim)))]
+        {
+            let inner = inner.clone();
+            tokio::spawn(async move {
+                loop {
+                    tokio::time::sleep(tokio::time::Duration::from_secs(60)).await;
+                    if Self::should_gc(&inner.read()) {
+                        Self::gc(&mut inner.write());
+                    }
```
chenzl25, line 37: Maybe we can add some logs here to monitor the map size.

src/frontend/src/handler/explain_analyze_stream_job.rs
```diff
@@ -0,0 +1,430 @@
...
+            "│  "
+        };
+        let child_prefix = format!("{}{}", prefix, child_prefix);
+
+        let stats = stats.inner.get(&node_id);
+        let (output_throughput, output_latency) = stats
+            .map(|stats| (stats.total_output_throughput, stats.total_output_pending_ms))
+            .unwrap_or((0, 0));
+        let row = ExplainAnalyzeStreamJobOutput {
+            operator_id: node.operator_id.to_string(),
```
chenzl25, line 410: I think we can hide the operator_id to users, since it is not a user-facing message.

src/frontend/src/handler/explain_analyze_stream_job.rs
```diff
@@ -0,0 +1,430 @@
...
+                .actor_ids
+                .iter()
+                .map(|id| id.to_string())
+                .collect::<Vec<_>>()
+                .join(","),
+            output_rps: (output_throughput as f64 / profiling_duration_secs).to_string(),
+            avg_output_pending_ratio: (Duration::from_millis(output_latency).as_secs_f64()
+                / node.actor_ids.len() as f64
+                / profiling_duration_secs)
+                .to_string(),
```
chenzl25, line 422: Maybe we can rename these 2 metrics to more user-friendly names like `output_rows_per_second` and `back_pressure_ratio`

chenzl25 (approved):
Generally LGTM. Thanks for the work!

src/frontend/src/handler/explain_analyze_stream_job.rs
kwannoel: fixed https://github.com/risingwavelabs/risingwave/pull/20715/commits/cc4bc13811268cf3f6e3e883dfd22bfd001dd53a

src/frontend/src/handler/explain_analyze_stream_job.rs
kwannoel: fixed https://github.com/risingwavelabs/risingwave/commit/cc4bc13811268cf3f6e3e883dfd22bfd001dd53a

src/stream/src/executor/monitor/in_mem_stats.rs
kwannoel: fixed https://github.com/risingwavelabs/risingwave/commit/cc4bc13811268cf3f6e3e883dfd22bfd001dd53a

src/frontend/src/handler/explain_analyze_stream_job.rs
BugenZhao, line 1: nit: I'm not sure whether we are going to introduce more profiling metrics. If so, the current approach might seem difficult to maintain, as the code logics for fetching/mapping/aggregating/displaying metrics are fragmented.

src/stream/src/executor/monitor/in_mem_stats.rs
```diff
@@ -0,0 +1,87 @@
...
+use std::collections::HashMap;
+use std::sync::Arc;
+use std::sync::atomic::AtomicU64;
+
+use parking_lot::RwLock;
+
+pub type Count = Arc<AtomicU64>;
+
+#[derive(Clone)]
+pub struct CountMap(Arc<RwLock<HashMap<u64, Count>>>);
```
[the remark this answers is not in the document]
kwannoel: I think it could be a good fit, but it was updated last 4 years ago.

src/stream/src/executor/monitor/streaming_stats.rs
```diff
@@ -1529,6 +1542,25 @@ impl StreamingMetrics {
                 .with_guarded_label_values(label_list),
         }
     }
+
+    pub fn new_profile_metrics(
+        &self,
+        operator_id: u64,
+        enable_profiling: bool,
+    ) -> ProfileMetricsImpl {
```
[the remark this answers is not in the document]
kwannoel: If we just return an `Option`, inside the `stream_node_metrics` wrapper, the code becomes more unreadable, because we need to deconstruct the pattern match on the `Option<ProfileMetrics>`.

The current approach allows us to dynamically dispatch these calls.

src/sqlparser/src/parser.rs
```diff
@@ -4295,6 +4295,36 @@ impl Parser<'_> {
...
+                    Ok(None)
+                }
+            }
+            if let Some(target) = parse_analyze_target(self)? {
+                let statement = Statement::ExplainAnalyzeStreamJob { target };
+                return Ok(statement);
+            }
+        }
+
         let statement = self.parse_statement()?;
```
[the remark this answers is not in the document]
kwannoel:
```
dev=> explain analyze ka;

Caused by:
  sql parser error: expected SINK, TABLE, MATERIALIZED VIEW, INDEX or a statement after ANALYZE, found: ka
LINE 1: explain analyze ka;

```

Updated

src/frontend/src/handler/explain_analyze_stream_job.rs
kwannoel: Point taken, I will give it more thought later.

src/config/example.toml
```diff
@@ -169,6 +169,7 @@ stream_enable_shared_source = true
 stream_switch_jdbc_pg_to_native = false
 stream_max_barrier_batch_size = 1024
 stream_hash_join_entry_state_max_rows = 30000
+stream_enable_explain_analyze_stats = true
```
hengm3467, line 172: @kwannoel If we intend this feature to be used by users, shall we not include the settings in the `developer` section? Sorry for the belated comment. I was looking at the blog post about `EXPLAIN ANALYZE`. I feel this is an awesome feature and experienced users will find it super useful. 
cc: @hzxa21

PR #20759, opened 2025-03-06
feat(iceberg): support creating namespace and namespace exists for iceberg catalog

chenzl25:

- support creating namespace and namespace exists for iceberg catalog. It is used by iceberg table engine. For example, if a user uses glue asa  catalog, they shouldn't create a namespace by themselves when using a new schema in risingwave.

## Checklist

## Documentation

<details>

</details>


java/connector-node/risingwave-sink-iceberg/src/main/java/com/risingwave/connector/catalog/JniCatalogWrapper.java
```diff
@@ -100,6 +100,43 @@ public boolean tableExists(String tableIdentifier) {
...
+    /**
+     * Checks if a namespace exists in the catalog.
+     *
+     * @param namespaceStr The namespace to check.
+     * @return true if the namespace exists, false otherwise.
+     */
+    public boolean namespaceExists(String namespaceStr) {
+        Namespace namespace;
+        if (namespaceStr == null) {
+            namespace = Namespace.empty();
```
Li0k, line 112: Should we return directly when the namespace is empty?

Li0k (approved):
Rest LGTM

java/connector-node/risingwave-sink-iceberg/src/main/java/com/risingwave/connector/catalog/JniCatalogWrapper.java
chenzl25: I think we can let the underlying system to decide.

PR #20780, opened 2025-03-07
feat: Add support for altering source connector properties

tabVersion:

This commit introduces the ability to alter connector properties for sources using the `ALTER SOURCE ... CONNECTOR WITH` syntax. 

related https://github.com/risingwavelabs/risingwave/pull/20691

## Checklist

## Documentation

<details>

support new syntax 

```
ALTER [ TABLE / SOURCE ] <object name> CONNECTOR WITH ( ... delta props ...);
```

to alter the connector props

Note that the props defined in CONNECTION should be altered by `ALTER CONNECTION`

This is a new feature and we just allow altering props within the whitelist & not change the state table

```
only alter kafka props allowed:

- properties.client.id
- properties.sync.call.timeout
- properties.enable.auto.commit
- properties.enable.ssl.certificate.verification
- properties.fetch.max.bytes
- properties.fetch.queue.backoff.ms
- properties.fetch.wait.max.ms
- properties.message.max.bytes
- properties.queued.max.messages.kbytes
- properties.queued.min.messages
- properties.receive.message.max.bytes
- properties.statistics.interval.ms
- properties.ssl.endpoint.identification.algorithm
- properties.security.protocol
- properties.sasl.mechanism
- properties.sasl.username
- properties.sasl.password
```

</details>


src/meta/src/barrier/context/context_impl.rs
```diff
@@ -145,6 +145,15 @@ impl CommandContext {
                     .unregister_table_ids(unregistered_state_table_ids.iter().cloned())
                     .await?;
             }
+            Command::ConnectorPropsChange(obj_id_map_props) => {
```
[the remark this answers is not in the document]
tabVersion: should be done in the future

src/frontend/src/handler/alter_source_props.rs
```diff
@@ -0,0 +1,172 @@
...
+}
+
+async fn handle_alter_source_props_inner(
+    session: &SessionImpl,
+    alter_props: Vec<SqlOption>,
+    source_id: u32,
+) -> Result<()> {
+    let meta_client = session.env().meta_client();
+    let (resolved_with_options, _, connector_conn_ref) = resolve_connection_ref_and_secret_ref(
+        WithOptions::try_from(alter_props.as_ref() as &[SqlOption])?,
```
tabVersion, line 82: https://github.com/risingwavelabs/risingwave/pull/22085
should use `Ident::from` to convert it back

src/connector/src/with_options.rs
```diff
@@ -231,6 +232,63 @@ impl WithOptionsSecResolved {
...
+                );
+            }
+        }
+
+        // remove legacy key if it's set in both plaintext and secret
+        for k in update_alter_props.keys() {
+            if self.secret_ref.contains_key(k) {
+                let to_remove_sink = self.secret_ref.remove(k).unwrap();
+                to_remove_secret_dep.push(to_remove_sink.secret_id);
+            }
```
hzxa21: nits:
```rust
if let Some(removed_secret) = self.secret_ref.remove(k) {
    to_remove_secret_dep.push(removed_secret.secret_id);
}
```

src/connector/src/with_options.rs
```diff
@@ -231,6 +232,63 @@ impl WithOptionsSecResolved {
...
+        }
+        for (k, v) in &update_alter_secret_refs {
+            if self.inner.contains_key(k) {
+                self.inner.remove(k);
+            }
```
hzxa21: nits:
```rust
self.inner.remove(k);
```

src/connector/src/with_options.rs
```diff
@@ -231,6 +232,63 @@ impl WithOptionsSecResolved {
...
+                if old_secret_ref.secret_id != v.secret_id {
+                    to_remove_secret_dep.push(old_secret_ref.secret_id);
+                } else {
+                    // If the secret ref is the same, we don't need to update it.
+                    continue;
+                }
+            }
+        }
+
+        self.inner.extend(update_alter_props);
```
hzxa21, line 283: Will it result in duplicate keys in `self.inner` given that we don't check the keys against `update_alter_props`?

src/sqlparser/src/ast/ddl.rs
```diff
@@ -239,6 +243,9 @@ pub enum AlterSourceOperation {
         parallelism: SetVariableValue,
         deferred: bool,
     },
+    AlterConnectorProps {
```
hzxa21, line 252: Let's rename `AlterSinkOperation::SetSinkProps` to `AlterSinkOperation::AlterConnectorProps` and `changed_props` to `alter_props` to make the naming consistent between source and sink.

src/frontend/src/handler/alter_source_props.rs
```diff
@@ -0,0 +1,172 @@
...
+    if let Some(conn_id) = connection_id {
+        let conn = reader.get_connection_by_id(db_name, conn_id)?;
+        if let ConnectionParams(params) = &conn.info {
+            for prop in alter_props {
+                let prop_key = prop.name.real_value();
+                if params.properties.contains_key(&prop_key)
+                    || params.secret_refs.contains_key(&prop_key)
+                {
+                    return Err(ErrorCode::InvalidInputSyntax(
+                        "Cannot alter connector properties that are set by CONNECTION. Use `ALTER CONNECTION` to alter them.".to_owned(),
```
hzxa21: Do we support `ALTER CONNECTION` props already? If not, let's not mention it in the error message.

src/meta/src/controller/streaming_job.rs
```diff
@@ -1620,6 +1624,252 @@ impl CatalogController {
...
+            source.with_properties.0.clone(),
+            source
+                .secret_ref
+                .map(|secret_ref| secret_ref.to_protobuf())
+                .unwrap_or_default(),
+        );
+        let (to_add_secret_dep, to_remove_secret_dep) =
+            options_with_secret.handle_update(alter_props, alter_secret_refs)?;
+
+        tracing::debug!(
```
hzxa21: I think this can be an INFO log to ease future troubleshooting.

src/meta/src/controller/streaming_job.rs
```diff
@@ -1620,6 +1624,252 @@ impl CatalogController {
...
+                .then(|| SecretRef::from(options_with_secret.as_secret().clone()))),
+            ..Default::default()
+        };
+        active_source_model.update(&txn).await?;
+
+        if let Some(associate_table_id) = associate_table_id {
+            // update the associated table statement accordly
+            let active_table_model = table::ActiveModel {
+                table_id: Set(associate_table_id),
+                definition: Set(rewrite_sql),
```
hzxa21, line 1871: Both the table and source model are using the same `rewrite_sql` definition. Is this duplication expected?

src/meta/src/manager/metadata.rs
```diff
@@ -703,6 +705,53 @@ impl MetadataManager {
...
+    pub async fn update_source_props_by_source_id(
+        &self,
+        source_id: SourceId,
+        alter_props: BTreeMap<String, String>,
+        alter_secret_refs: BTreeMap<String, SecretRef>,
+    ) -> MetaResult<WithOptionsSecResolved> {
+        self.catalog_controller
+            .update_source_props_by_source_id(source_id, alter_props, alter_secret_refs)
+            .await
+    }
```
hzxa21: This method is not used because `stream_service.rs` call `catalog_controller.update_source_props_by_source_id` directly.

src/meta/src/stream/source_manager.rs
```diff
@@ -162,6 +164,13 @@ impl SourceManagerCore {
                 split_assignment = split_assignment_;
                 dropped_actors = dropped_actors_;
             }
+            SourceChange::UpdateSourceProps {
+                source_id_map_new_props,
+            } => {
+                for (source_id, new_props) in source_id_map_new_props {
+                    reccreate_source_id_map_new_props.push((source_id, new_props));
```
hzxa21: type: reccreate -> recreate

src/meta/src/stream/source_manager/worker.rs
```diff
@@ -306,6 +306,13 @@ impl ConnectorSourceWorker {
                                     tracing::warn!(error = %e.as_report(), "error happened when finish backfill");
                                 }
                             }
+                            SourceWorkerCommand::UpdateProps(new_props) => {
+                                self.connector_properties = new_props;
```
hzxa21, line 309: Do we have some validation phase for the new props to make sure the new values are valid?

src/connector/src/source/mod.rs
```diff
@@ -207,3 +210,31 @@ impl WaitCheckpointTask {
         }
     }
 }
+
+pub static ALTER_CONNECTOR_PROPS_ALLOWED: LazyLock<HashMap<String, HashSet<String>>> =
+    LazyLock::new(|| {
+        hashmap! {
+            "kafka".to_owned() => hashset! {
+                "properties.client.id".to_owned(),
```
wenym1, line 218: Shall we change to declare the allowed props in the source trait, so that this map can be initialized with a `for_all_sources!` macro, and we won't forget to update this map when we implement new sources.

tabVersion: > Shall we change to declare the allowed props in the source trait, so that this map can be initialized with a for_all_sources! macro

This is another thing I want to discuss. 
I think `ALTER_CONNECTOR_PROPS_ALLOWED` is a business policy that controls which properties users can modify at runtime. And the source trait should focus on core data source functionality, not altering permission policies. 

> we won't forget to update this map when we implement new sources.

The hashmap serves as a white-list so we can always add new connectors in separate pr. New sources are not automatically granted modification permissions to any of its props, making the permission safer.

src/connector/src/with_options.rs
tabVersion: calling BtreeMap extend should overwrite the existing key, so we may not worried about dup key

```
The extend method on BTreeMap in Rust allows you to add all elements from an iterator into the map, effectively merging two maps. If a key already exists, the existing value will be overwritten by the new value from the iterator

summarized by Google search
```

src/meta/src/controller/streaming_job.rs
tabVersion: yes, it is expected. When creating a table with connector, we record the same sql in rw_tables and rw_sources. 

```
dev=> create table t  (a bytea)
include timestamp as timestamp  
WITH ( 
  connector = 'kafka', properties.bootstrap.server = 'localhost:9092', 
  scan.startup.mode = 'earliest', 
  topic = 'test4',  
) format plain encode bytes;
CREATE_TABLE
dev=> select definition from rw_tables;
CREATE TABLE t (a BYTEA) INCLUDE timestamp AS timestamp WITH (connector = 'kafka', properties.bootstrap.server = 'localhost:9092', scan.startup.mode = 'earliest', topic = 'test4') FORMAT PLAIN ENCODE BYTES
(1 row)
dev=> select definition from rw_sources;
CREATE TABLE t (a BYTEA) INCLUDE timestamp AS timestamp WITH (connector = 'kafka', properties.bootstrap.server = 'localhost:9092', scan.startup.mode = 'earliest', topic = 'test4') FORMAT PLAIN ENCODE BYTES
(1 row)
```

src/sqlparser/src/ast/ddl.rs
tabVersion: It is a leftover when impl alter sink props. Prefer do it in another pr

tabVersion: https://github.com/risingwavelabs/risingwave/pull/22239

src/meta/src/stream/source_manager/worker.rs
tabVersion: is done by adding new method `validate_source_once` in SourceManager

src/connector/src/source/mod.rs
wenym1: > I think ALTER_CONNECTOR_PROPS_ALLOWED is a business policy that controls which properties users can modify at runtime. And the source trait should focus on core data source functionality, not altering permission policies.

I don't think so. The scope of these two concepts are quite small, and they are declared in the same crate, even the same file, and we don't have to intentionally split them. Besides, currently we have the `SourceProperties` trait as the entrance trait for any source, and at least the trait name indicates that anything related to the source can be declared here. And this is also how we implement alter sink prop.

> The hashmap serves as a white-list so we can always add new connectors in separate pr.

This is not about the flexibility to add new entry to the white-list hashmap. Implementing the trait for a new source can be treated as a todo-list, with which developers can be aware of what should be considered for a new source. If we have a white-list separately, developers without related context will never know that they should consider alter source prop.

tabVersion: Let's keep it for the first version, I will try hack the WithOption trait for more native support.

tabVersion: https://github.com/risingwavelabs/risingwave/pull/22363

e2e_test/source_inline/kafka/alter/alter_connector_props.slt
```diff
@@ -0,0 +1,116 @@
...
+
+sleep 1s
+
+query I retry 3 backoff 5s
+SELECT count(*) from plain_students;
+----
+3
+
+statement error Invalid Parameter Value: Invalid parameter: connector kafka does not support alter property `topic`
+ALTER TABLE plain_students CONNECTOR WITH ( topic = secret s2 );
```
xxchan, line 70: nit about the syntax: 
In PG:
```
CREATE TABLE t(...) WITH ( storage_parameter [= value] [, ... ] ) 
ALTER TABLE t SET ( storage_parameter [= value] [, ... ] )
```
So what about using `SET` instead of `WITH` here?

src/connector/src/source/mod.rs
```diff
@@ -207,3 +210,31 @@ impl WaitCheckpointTask {
...
+                "properties.queued.max.messages.kbytes".to_owned(),
+                "properties.queued.min.messages".to_owned(),
+                "properties.receive.message.max.bytes".to_owned(),
+                "properties.statistics.interval.ms".to_owned(),
+                "properties.ssl.endpoint.identification.algorithm".to_owned(),
+                // need to rebuild source enumerator
+                "properties.security.protocol".to_owned(),
+                "properties.sasl.mechanism".to_owned(),
+                "properties.sasl.username".to_owned(),
+                "properties.sasl.password".to_owned(),
```
xxchan, line 235: Regarding the comment, how are they distinguished? Or we just always rebuild?

tabVersion: always rebuild

e2e_test/source_inline/kafka/alter/alter_connector_props.slt
tabVersion: maybe too late, we have already adopt the `with` style syntax in the alter support for sink, and merged in release 2.4

tabVersion:
merge first to catch up with release-2.5

PR #20681, opened 2025-02-28
fix(frontend): avoid error when `create materialized view if not exists` on existing background mv

kwannoel:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## What's changed and what's your intention?

When background_ddl is used on some MV, and we attempt to recreate the MV with `if not exists`, it errors out before this PR. We fix it by getting all names, not just created ones, to check for duplicates.
See slt test for more details.

## Checklist

- [ ] I have written necessary rustdoc comments.
- [ ]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  My PR contains critical fixes that are necessary to be merged into the latest release. 

## Documentation

- [ ]  My PR needs documentation updates. 

<details>
<summary><b>Release note</b></summary>

</details>

PR #20783, opened 2025-03-07
fix(meta): background ddl wait collect first barrier

wenym1:

Currently in background ddl, we return without waiting for collecting and committing the first barrier. However, we will only retry the background ddl job until the first barrier is collected and committed so that the job info can be marked as creating. If we early return, users may think that the background ddl can be persisted and eventually succeed, but actually it won't.

In this PR, we will change to return until collecting the first barrier.

Besides, in frontend, when seeing a job to create exists but is creating, we will return with an extra `exists but under creation` error message to users. In the future, when handling `create if not exists` and seeing that the job exists but under creation, we will wait until the job is created before return.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/catalog/mod.rs
+++ b/src/frontend/src/catalog/mod.rs
@@ -106,8 +106,21 @@ pub type CatalogResult<T> = std::result::Result<T, CatalogError>;
 pub enum CatalogError {
     #[error("{0} not found: {1}")]
     NotFound(&'static str, String),
-    #[error("{0} with name {1} exists")]
-    Duplicated(&'static str, String),
+    #[error(
+        "{0} with name {1}  exists{under_creation}", under_creation = (.2).then_some(" but under creation").unwrap_or("")
+    )]
+    Duplicated(
+        &'static str,
+        String,
+        // whether the object is under creation
+        bool,
+    ),
+}
+
+impl CatalogError {
+    pub fn duplicated(object_type: &'static str, name: String) -> Self {
+        Self::Duplicated(object_type, name, false)
+    }
 }
 
 impl From<CatalogError> for RwError {
--- a/src/frontend/src/catalog/mod.rs
+++ b/src/frontend/src/catalog/mod.rs
@@ -107,7 +107,7 @@ pub enum CatalogError {
     #[error("{0} not found: {1}")]
     NotFound(&'static str, String),
     #[error(
-        "{0} with name {1}  exists{under_creation}", under_creation = (.2).then_some(" but under creation").unwrap_or("")
+        "{0} with name {1} exists{under_creation}", under_creation = (.2).then_some(" but under creation").unwrap_or("")
     )]
     Duplicated(
         &'static str,
--- a/src/frontend/src/catalog/root_catalog.rs
+++ b/src/frontend/src/catalog/root_catalog.rs
@@ -16,7 +16,7 @@ use std::collections::HashMap;
 use std::sync::Arc;
 
 use itertools::Itertools;
-use risingwave_common::catalog::{CatalogVersion, FunctionId, IndexId, TableId};
+use risingwave_common::catalog::{CatalogVersion, FunctionId, IndexId, StreamJobStatus, TableId};
 use risingwave_common::session_config::{SearchPath, USER_NAME_WILD_CARD};
 use risingwave_common::types::DataType;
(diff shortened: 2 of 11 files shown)


kwannoel (approved):
LGTM, thanks

kwannoel (commented):
IIRC we also report the catalog to FE _before_ the first barrier. I think we need to change that logic as well.

wenym1:
> IIRC we also report the catalog to FE _before_ the first barrier. I think we need to change that logic as well.

In current code, catalog changes are synced to FE asynchronously via notification. For foreground jobs, FE will wait for the notification version returned from the create streaming job response, and for background jobs, it returns with an empty notification version to indicate that it won't wait for any notification version. 

So what would be the expected behavior we need to change to 🤔

kwannoel:
> > IIRC we also report the catalog to FE _before_ the first barrier. I think we need to change that logic as well.
> 
> In current code, catalog changes are synced to FE asynchronously via notification. For foreground jobs, FE will wait for the notification version returned from the create streaming job response, and for background jobs, it returns with an empty notification version to indicate that it won't wait for any notification version.
> 
> So what would be the expected behavior we need to change to 🤔

The current behaviour is to:
1. Immediately update the catalog to FE once it's done with `prepare_streaming_job` step.
2. Respond to FE after the initial prepare steps, without waiting for first barrier.

After this PR I think we should:
1. Update the catalog to FE after first barrier collected.
2. Respond to FE after first barrier collected.

Both steps 1 & 2 are kind of decoupled. The notification to FE is not done in the same path, but separately via a different notification service. We currently report the table catalog in the `prepare_streaming_job` step: https://github.com/risingwavelabs/risingwave/blob/87c7975a7f3d366eb950b220beb5fc551cefb28f/src/meta/src/controller/streaming_job.rs#L392

At this line: https://github.com/risingwavelabs/risingwave/blob/87c7975a7f3d366eb950b220beb5fc551cefb28f/src/meta/src/controller/streaming_job.rs#L525.

This step finishes before the first barrier is collected. So if we don't update this logic, after this PR, it will be:
1. Update the catalog to FE once it's done with `prepare_streaming_job` step.
2. Respond to FE after first barrier.

There's a scenario where the barrier fails before (2), for instance if CN crashed. Then FE will have the catalog, but actually the job has be cancelled due to the error. Then if we attempt to recreate the MV, it will fail, because an existing catalog exists. FE will be in an inconsistent state until it restarts.

wenym1:
> This step finishes before the first barrier is collected. So if we don't update this logic, after this PR, it will be:
>
> 1. Update the catalog to FE once it's done with prepare_streaming_job step.
> 2. Respond to FE after first barrier.
>
> There's a scenario where the barrier fails before (2), for instance if CN crashed. Then FE will have the catalog, but actually the job has be cancelled due to the error. Then if we attempt to recreate the MV, it will fail, because an existing catalog exists. FE will be in an inconsistent state until it restarts.

I observed similar errors in CI recovery test. But I think we shouldn't fix it by delaying notification to FE, since it doesn't completely prevent FR from seeing the inconsistency. FE can still fetch a whole snapshot of catalog when being relaunched quickly enough. 

I will push a fix commit by returning with more information in error message when a job with the same name exists but is still being created. It will show message like `job xxx exists but under creation`. In CI recovery test, previously when we see that the error message contains `exists`, we will treat the job as already created, though it might be still creating. In this fix commit, we will not treat the job as created when the error message contains `under creation`.

In the future, we may add some extra logic when handling create job with `if not exists`. When we see a job with the name exists but is still creating, we may wait until the job is created.

wenym1:
* **#20783** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20783?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20783?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

kwannoel:
> In the future, we may add some extra logic when handling create job with if not exists. When we see a job with the name exists but is still creating, we may wait until the job is created.

Actually this is something I just fixed for a user https://github.com/risingwavelabs/risingwave/pull/20681. I think we need to preserve this. `create if not exists` should not fail.

yezizp2012 (approved):
LGTM

PR #20795, opened 2025-03-10
feat(frontend,meta): support alter fragment rate limit

kwannoel:

This developer-feature is added for two reasons: first, to support altering specific backfill rate limits, currently our granularity is at the **relation level**. Sometimes we need to tune backfill rate limit for specific sources, e.g. dimension table, so this allows us to do that as a last resort option. Next, for backfill order control we will need to be able to alter fragment rate limit for specific operators. Some of the internal functionality of this PR overlaps with that. The major assumption we make is that each backfill operator **exists on its own fragment**.

Provides more granular control over rate limit. Not a stable SQL interface. So do not mark as `user-facing`.

- [x] Add an slt test

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/meta/src/controller/streaming_job.rs
+++ b/src/meta/src/controller/streaming_job.rs
@@ -1514,6 +1514,51 @@ impl CatalogController {
         Ok(fragment_actors)
     }
 
+    async fn mutate_fragment_by_fragment_id(
+        &self,
+        fragment_id: FragmentId,
+        mut fragment_mutation_fn: impl FnMut(&mut i32, &mut PbStreamNode) -> bool,
+        err_msg: &'static str,
+    ) -> MetaResult<HashMap<FragmentId, Vec<ActorId>>> {
+        let inner = self.inner.read().await;
+        let txn = inner.db.begin().await?;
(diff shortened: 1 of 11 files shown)


src/meta/src/controller/streaming_job.rs
```diff
@@ -1514,6 +1514,51 @@ impl CatalogController {
         Ok(fragment_actors)
     }
 
+    async fn mutate_fragment_by_fragment_id(
```
chenzl25, line 1517: It seems only changing the fragment metadata, but no barrier is carrying out the information to the actors. Does it work for background ddl only?

kwannoel: See: https://github.com/risingwavelabs/risingwave/pull/20795/files#diff-66d82e7ca80b658c018a0a5f6407e235d890ba46ecc677cdf83a626f4af7dde7R147-R174

chenzl25 (approved):
LGTM

yezizp2012 (approved):
Rest LGTM.

Issue #20640, opened 2025-02-27
Enhancing the Scalability of Hummock LSM


Li0k:
**Meta Scalability Issues**

1. Single-Core Bottleneck: Meta has a single - core bottleneck that cannot be alleviated by scaling.Global Lock Problem: There is a global lock in version. In fact, multiple control groups (cg) do not conflict. The non-conflicting feature of LSM multi-cg submission is not utilized, and a global lock is not necessary.  

2. CPU Overhead: Meta has a high CPU overhead. There is Vec sort overhead in apply_compact_ssts and HashSet overhead in check_sst_ids_exist. Meta can adopt a more reasonable structure, such as list + map index, to avoid global sorting when applying version delta each time, so as to optimize CPU utilization. High CPU usage is also a bottleneck for compaction efficiency.

**LSM Scalability Issues**

1. Single LSM Size Limit: The size of a single LSM tree is limited. When it exceeds the default configured size, the score calculation will be disordered.

2. Subtree Splitting Granularity: The granularity of LSM subtree splitting is at the table level and cannot be split at a smaller granularity. As a result, the resource utilization of a single LSM is lower than that of multiple LSMs, and the competition to the base level is more severe. From a personal perspective, LSM should not have a size limit. The score-calculating algorithm may need further research. For large-capacity single-table LSMs, splitting subtrees is a better solution. In the short term, it is feasible to support subtree splitting at the vnode level, but note that this cannot reduce the overall compactor utilization. Unreasonable configurations and score-calculating algorithms may expand the jitter of write amplification.

**Large-Scale Ingestion Optimization**

1. Adequate Resource Assurance: Sufficient resources are needed to ensure the compaction speed.
2. New Solution Exploration: Consider whether to adopt a new solution to bypass the write amplification problem of LSM, such as researching a better batch ingestion method, adopting the Append Only solution, and delaying compaction.

zwang28:
> Global Lock Problem: There is a global lock in version.

Meta node operations that require global locks, such as commit_epoch and compaction, can become a source of contention for multiple isolated databases. Implementing a more fine-grained locking mechanism is expected to alleviate this issue.

hzxa21:
Note: the meta scalability issue can be a problem if we have a multi-tenant (multi-DB) RW cluster.

PR #20824, opened 2025-03-11
refactor(compaction): optimize level_insert_ssts

zwang28:

part of #20640

This PR addresses the CPU overhead introduced by the sort in `level_insert_ssts`. 

The CPU overhead can be substantial during intensive compaction, often due to numerous trivial move tasks. [This integration test](https://buildkite.com/risingwave-test/compaction-test/builds/285) using append-only write workload illustrates the issue. The flamegraph shows 35% of CPU is utilized by the `sort of SstableInfo`.
meta-flamegraph-2025-03-11T04-28-49Z

The solution is simply to replace the original sort with an insertion sort. We were concerned whether the additional element movement overhead of insertion sort might outweigh its benefits. Fortunately both the [microbench](https://github.com/risingwavelabs/risingwave/tree/wangzheng/bench) and [integretion test](https://buildkite.com/risingwave-test/compaction-test/builds/287) has shown that the insertion sort is significantly faster in this case. 
- The microbench shows the insertion sort is 10 times faster than original sort.
- The flamegraph indicates that the CPU overhead from sorting has been eliminated.

meta-flamegraph-2025-03-11T06-05-16Z

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/storage/hummock_sdk/src/compaction_group/hummock_version_ext.rs
+++ b/src/storage/hummock_sdk/src/compaction_group/hummock_version_ext.rs
@@ -1356,14 +1356,27 @@ fn level_insert_ssts(operand: &mut Level, insert_table_infos: &Vec<SstableInfo>)
         .iter()
         .map(|sst| sst.uncompressed_file_size)
         .sum::<u64>();
-    operand
-        .table_infos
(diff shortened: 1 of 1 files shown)


Li0k (approved):
LGTM

PR #20840, opened 2025-03-11
refactor(connector): correctly log java exceptions

BugenZhao:

Update `TracingSlf4jAdapter` to not directly implement the `Logger` interface, but extend the `LegacyAbstractLogger` class, so that we only need to handle the single method `handleNormalizedLoggingCall`, where the `Throwable` is already extracted from arguments.

Also record `stackTrace` as a separate tracing field `error` instead of interpolating it into the log message, to make the log more structured.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/jni_core/src/tracing_slf4j.rs
+++ b/src/jni_core/src/tracing_slf4j.rs
@@ -29,6 +29,7 @@ pub(crate) extern "system" fn Java_com_risingwave_java_binding_Binding_tracingSl
     class_name: JString<'_>,
     level: jint,
     message: JString<'_>,
+    stack_trace: JString<'_>,
 ) {
     execute_and_catch(env, move |env| {
         let thread_name = env.get_string(&thread_name)?;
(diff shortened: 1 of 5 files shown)


src/jni_core/src/tracing_slf4j.rs
```diff
@@ -40,13 +41,22 @@ pub(crate) extern "system" fn Java_com_risingwave_java_binding_Binding_tracingSl
...
+        let stack_trace: Option<&str> = stack_trace.as_deref();
+
         macro_rules! event {
             ($lvl:expr) => {
                 tracing::event!(
                     target: TARGET,
                     $lvl,
                     thread = &*thread_name,
                     class = &*class_name,
+                    error = stack_trace,
```
wenym1, line 59: Should it be `stack_trace` rather than `error`?

wenym1 (approved):
LGTM!

src/jni_core/src/tracing_slf4j.rs
BugenZhao: The `stack_trace` will also include the error message and exception class name. Just to follow the convention for recording Rust errors:

image

stdrc (approved):

PR #20822, opened 2025-03-11
feat(stream): support not null in stream

KeXiangWang:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## What's changed and what's your intention?

Followup of https://github.com/risingwavelabs/risingwave/pull/20611. For streaming input, for example, kafka input, or Sink into input, we drop the row violating NOT NULL constraint. And we ban NOT NULL in cdc tables.

## Checklist

- [ ] I have written necessary rustdoc comments.
- [ ]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  My PR contains critical fixes that are necessary to be merged into the latest release. 

## Documentation

- [x]  My PR needs documentation updates. 

<details>
<summary><b>Release note</b></summary>

Support NOT NULL in table schema.
For BatchInsert/Update, RW will throw an error for rows with NULL.
For Streaming, RW will ignore the rows with NULL silently.

PR #20858, opened 2025-03-11
feat(stream): notice for NOT NULL when create table with connector

KeXiangWang:

https://github.com/risingwavelabs/risingwave/issues/19373
Notice when 
1. creating table with connector and the schema contains NOT NULL
2. create sink into a table with NOT NULL. 
Alter table add column with NOT NULL contraint must have default value.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/handler/alter_table_column.rs
+++ b/src/frontend/src/handler/alter_table_column.rs
@@ -315,6 +315,20 @@ pub async fn handle_alter_table_column(
                 ))?
             }
 
+            if new_column
+                .options
+                .iter()
+                .any(|x| matches!(x.option, ColumnOption::NotNull))
+                && !new_column
+                    .options
+                    .iter()
+                    .any(|x| matches!(x.option, ColumnOption::DefaultValue(_)))
+            {
+                return Err(ErrorCode::InvalidInputSyntax(
+                    "alter table add NOT NULL columns must have default value".to_owned(),
+                ))?;
+            }
+
             // Add the new column to the table definition if it is not created by `create table (*)` syntax.
             columns.push(new_column);
 
--- a/e2e_test/ddl/not_null.slt
+++ b/e2e_test/ddl/not_null.slt
@@ -6,6 +6,9 @@ CREATE TABLE test_struct (
   info STRUCT <first_name VARCHAR, last_name VARCHAR> NOT NULL
 );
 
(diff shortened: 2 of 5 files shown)


src/frontend/src/handler/alter_table_column.rs
```diff
@@ -286,6 +286,20 @@ pub async fn handle_alter_table_column(
...
+                .options
+                .iter()
+                .any(|x| matches!(x.option, ColumnOption::NotNull))
+                && !new_column
+                    .options
+                    .iter()
+                    .any(|x| matches!(x.option, ColumnOption::DefaultValue(_)))
+            {
+                return Err(ErrorCode::InvalidInputSyntax(
+                    "alter table add NOT NULL columns must have default value".to_owned(),
```
BugenZhao, line 299: What if the default value is `NULL`? 🤯

BugenZhao:
May I know if the `NOT NULL` property of a column is strictly enforced and might be relied by us as a property, or is it just a best-effort? I'm asking because it requires different level of seriousness when implementing and reviewing the edge cases. 🤔

st1page:
> May I know if the `NOT NULL` property of a column is strictly enforced and might be relied by us as a property, or is it just a best-effort? I'm asking because it requires different level of seriousness when implementing and reviewing the edge cases. 🤔

strictly enforced. Because when user set the constraint, we should maintain the property and user should take responsible for that.
more discussion here
https://risingwave-labs.slack.com/archives/C0327040PK2/p1731475254014369

I believe that directly adding the filter in the table's streaming plan in https://github.com/risingwavelabs/risingwave/pull/20822 can ensure the filter of rows violating the NOT NULL constraint in the table.

src/frontend/src/handler/alter_table_column.rs
KeXiangWang: Hmmm, we should allow this but reject the row violating the NOT NULL constrains, according to PG. However, our current Default expr haven't forbid NULL, which means user can insert `NULL` to NOT NULL column via `default NULL`. We can first keep them aligned.

KeXiangWang: `However, our current Default expr haven't forbid NULL`
I'll leave a TODO to fix it later.

st1page (approved):

PR #20908, opened 2025-03-14
feat(stream): limit sync log store flushed item read size + only pause on checkpoint barrier

kwannoel:

If read size is unbounded, we can end up reading a really large chunk, and lead to barrier latency spike, if it cannot be processed next barrier.

We also only pause on ckpt barrier, because if we pause on chunk buffer full, the state will transition to the read state immediately, and after polling, the chunk buffer will have free space again, and we resume immediately. After that, the state transitions back to the write state, and the whole cycle repeats. In normal circumstances this is fine. But if there's high write pressure, it will backpressure the write side, instead of letting barriers pass.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/executor/sync_kv_log_store.rs
+++ b/src/stream/src/executor/sync_kv_log_store.rs
@@ -549,7 +549,7 @@ impl<S: StateStore> SyncedKvLogStoreExecutor<S> {
                 buffer: VecDeque::new(),
                 current_size: 0,
                 max_size: self.max_buffer_size,
-                max_read_size: self.chunk_size,
+                max_chunk_size: self.chunk_size,
                 next_chunk_id: 0,
                 metrics: self.metrics.clone(),
                 flushed_count: 0,
@@ -918,7 +918,7 @@ struct SyncedLogStoreBuffer {
     buffer: VecDeque<(u64, LogStoreBufferItem)>,
     current_size: usize,
     max_size: usize,
-    max_read_size: u32,
+    max_chunk_size: u32,
     next_chunk_id: ChunkId,
     metrics: SyncedKvLogStoreMetrics,
     flushed_count: usize,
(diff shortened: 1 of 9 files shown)


kwannoel:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20908?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#20908** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20908?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20908?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#20907** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20907?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/stream/src/executor/sync_kv_log_store.rs
```diff
@@ -1036,10 +1029,30 @@ impl SyncedLogStoreBuffer {
...
+            Some((
+                epoch,
+                LogStoreBufferItem::Flushed {
+                    start_seq_id,
+                    end_seq_id,
+                    vnode_bitmap,
+                    chunk_id,
+                },
+            )) => {
+                let end_seq_id_bound = *start_seq_id + self.max_read_size as i32;
```
wenym1, line 1043: Introducing the `max_read_size` seems increasing unnecessary complexity here. I think we can instead limit the max chunk size when we attempt to add a new `Flushed` item and merge with the previous `Flushed` item.

wenym1 (commented):
Rest LGTM

src/stream/src/executor/sync_kv_log_store.rs
kwannoel: Prefer not to do this. Because when adding a chunk, its size might exceed our max chunk size.
In such a case, we still have to break it up into multiple flushed items. And we have the same complexity.

The approach taken here has the same complexity, but is able to use less space for consecutive flush items O(1) vs O(n), where n is the number of consecutive flushed chunks.

wenym1: > Because when adding a chunk, its size might exceed our max chunk size.
In such a case, we still have to break it up into multiple flushed items. 

I don't think this is a case we should optimize on. We use the `chunk_size` in develop config as `max_chunk_size`, which is a chunk size configured globally. We can expect that the chunk we receive from upstream executor will not greatly exceed such `chunk_size`. On the contrary, in some executors a normally sized chunk may break down into some very small chunks, even chunks with only a single row, and that was the reason of introducing the logic of merging consecutive small chunks.

If we don't have to optimize on receiving a large chunk , we can avoid the complexity here. The current implementation looks wired, where we may first merge some too small chunks into an oversized chunk, and then break this oversized chunk into several normally sized chunks when we read from buffer.

kwannoel: updated

wenym1 (approved):
Rest LGTM. Thanks for the PR!

wenym1 (approved):

PR #20909, opened 2025-03-14
feat(sink): fix clickhouse sink datetime64 bug

### LINKED ISSUE #22758: bug(sink): ClickHouse DateTime64 precision mismatch when columns out of order
### Describe the bug

When sinking into a `DateTime64` column of ClickHouse, it is critical to get its precision and preprocess the data accordingly. For example `2001-02-03T04:05:06-07:00` shall be written as:
* `981198306` for `DateTime64(0)`
* `981198306000000000` for `DateTime64(9)`

However, the current logic of passing precision for each column is based on column order, and default to `0` (rather than `None`) when it fails. This can lead to wrong sink data when the upstream (RisingWave) and downstream (ClickHouse) have different column orders.

### Error message/log

```text
No error reported. Writes wrong data silently:

   ┌─id─┬───────────────────scheduled─┬───────────────────────actual─┐
1. │  3 │ 2280-12-05 14:51:01.2345600 │ 1970-01-01 00:00:09.81198306 │
   └────┴─────────────────────────────┴──────────────────────────────┘
```

### To Reproduce

Start a ClickHouse server instance and connect from a client:
```
docker run -d -p 18123:8123 -p19000:9000 -e CLICKHOUSE_PASSWORD=changeme --name some-clickhouse-server --ulimit nofile=262144:262144 clickhouse/clickhouse-server:25.5
docker exec -it some-clickhouse-server clickhouse-client
```

In ClickHouse:
```
create table t (id Int32 primary key, scheduled datetime64(7), actual datetime64(8));
```

Then in RisingWave:
```
create table t (actual timestamptz, id int primary key, scheduled timestamptz);
insert into t values ('2001-02-03T11:05:06.987654Z', 3, '2001-02-03T11:05:06.123456Z');
create sink sk from t with (connector = 'clickhouse', type = 'append-only', force_append_only = true, clickhouse.url = 'http://127.0.0.1:18123', clickhouse.user = 'default', clickhouse.password = 'changeme', clickhouse.database = 'default', clickhouse.table = 't');
```

### Expected behavior

```
   ┌─id─┬───────────────────scheduled─┬───────────────────────actual─┐
1. │  3 │ 2001-02-03 11:05:06.1234560 │ 2001-02-03 11:05:06.98765400 │
   └────┴─────────────────────────────┴──────────────────────────────┘

```
(Obtained via `create sink sk as select id, scheduled, actual from t with (connector = ...);`)

_No response_

### The version of RisingWave

2.4

### Additional context

_No response_

xxhZs:

This PR fixes ClickHouse sink handling for `DateTime64` columns after relaxing the earlier one-to-one column correspondence assumption between ClickHouse and RisingWave.

The previous implementation relied on positional matching through a vector, which could pick the wrong ClickHouse column metadata when the schemas were not aligned exactly. This PR changes the lookup to use a column map, so validation and encoding use the metadata for the intended target column.

## Checklist

## Documentation

<details>

Fixes ClickHouse sink schema handling for `DateTime64` columns when ClickHouse and RisingWave columns are not in a strict one-to-one positional layout.

</details>

### DIFF
--- a/src/connector/src/sink/clickhouse.rs
+++ b/src/connector/src/sink/clickhouse.rs
@@ -22,9 +22,10 @@ use clickhouse::{Client as ClickHouseClient, Row as ClickHouseRow};
 use itertools::Itertools;
 use phf::{Set, phf_set};
 use risingwave_common::array::{Op, StreamChunk};
-use risingwave_common::catalog::Schema;
+use risingwave_common::catalog::{FieldLike, Schema};
 use risingwave_common::row::Row;
 use risingwave_common::types::{DataType, Decimal, ScalarRefImpl, Serial};
+use risingwave_common::util::iter_util::ZipEqDebug;
 use serde::ser::{SerializeSeq, SerializeStruct};
 use serde::{Deserialize, Serialize};
 use serde_with::{DisplayFromStr, serde_as};
@@ -81,7 +82,6 @@ impl EnforceSecret for ClickHouseCommon {
         "clickhouse.password", "clickhouse.user"
     };
 }
-
 
 #[derive(Debug)]
 enum ClickHouseEngine {
@@ -603,7 +603,6 @@ impl Sink for ClickHouseSink {
 }
 pub struct ClickHouseSinkWriter {
     pub config: ClickHouseConfig,
-    #[expect(dead_code)]
     schema: Schema,
     #[expect(dead_code)]
     pk_indices: Vec<usize>,
@@ -611,7 +610,7 @@ pub struct ClickHouseSinkWriter {
     #[expect(dead_code)]
     is_append_only: bool,
     // Save some features of the clickhouse column type
-    column_correct_vec: Vec<ClickHouseSchemaFeature>,
+    column_correct_map: HashMap<String, ClickHouseSchemaFeature>,
     rw_fields_name_after_calibration: Vec<String>,
     clickhouse_engine: ClickHouseEngine,
     inserter: Option<Insert<ClickHouseColumn>>,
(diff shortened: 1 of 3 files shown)


src/connector/src/sink/clickhouse.rs
```diff
@@ -731,13 +737,15 @@ impl ClickHouseSinkWriter {
                 self.rw_fields_name_after_calibration.clone(),
             )?);
         }
+        let fields_name = self.schema.fields().iter().cloned().collect_vec();
```
xiangjinwu: Why do we need a dedicated vec of clones?

src/connector/src/sink/clickhouse.rs
```diff
@@ -915,14 +923,27 @@ enum ClickHouseFieldWithNull {
...
+        };
+
         if data.is_none() {
-            if !clickhouse_schema_feature.can_null {
+            if clickhouse_schema_feature.is_none() {
+                return Err(SinkError::ClickHouse(
+                    "clickhouse's nested can not insert null".to_owned(),
+                ));
+            }
+            if !clickhouse_schema_feature.unwrap().can_null {
```
xiangjinwu: Avoid `is_none()` followed by `.unwrap()`. Use `let Some(..) = .. else { .. };`

src/connector/src/sink/clickhouse.rs
```diff
@@ -1007,11 +1031,17 @@ impl ClickHouseFieldWithNull {
             }
             ScalarRefImpl::Struct(v) => {
                 let mut struct_vec = vec![];
-                for (index, field) in v.iter_fields_ref().enumerate() {
+                for ((field, struct_field_name), struct_field_type) in v
+                    .iter_fields_ref()
+                    .zip_eq_debug(rw_type.as_struct().names())
+                    .zip_eq_debug(rw_type.as_struct().types())
```
xiangjinwu: `rw_type.as_struct().iter()` rather than getting `names()` and `types()` individually

chenzl25 (approved):

PR #20568, opened 2025-02-21
.

xxchan:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## Checklist

- [ ] I have written necessary rustdoc comments.
- [ ]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  My PR contains critical fixes that are necessary to be merged into the latest release. 

## Documentation

- [ ]  My PR needs documentation updates. 

<details>
<summary><b>Release note</b></summary>

</details>

PR #20911, opened 2025-03-14
refactor(state table): add an API for 1 row table

xxchan:

reuse code for dynamic filter / now, and planned to be used for iceberg list exec

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/common/table/state_table.rs
+++ b/src/stream/src/common/table/state_table.rs
@@ -1377,6 +1377,34 @@ where
         Ok(stream.map_ok(|(_, row)| row))
     }
 
+    /// Get the row from a state table with only 1 row.
+    pub async fn get_from_one_row_table(&self) -> StreamExecutorResult<Option<OwnedRow>> {
+        let sub_range: &(Bound<OwnedRow>, Bound<OwnedRow>) = &(Unbounded, Unbounded);
+        let stream = self
+            .iter_with_prefix(row::empty(), sub_range, Default::default())
+            .await?;
+        pin_mut!(stream);
+
+        if let Some(res) = stream.next().await {
+            let value = res?.into_owned_row();
+            assert!(stream.next().await.is_none());
+            Ok(Some(value))
+        } else {
+            Ok(None)
+        }
+    }
+
+    /// Get the row from a state table with only 1 row, and the row has only 1 col.
+    ///
+    /// `None` can mean either the row is never persisted, or is a persisted `NULL`,
+    /// which does not matter in the use case.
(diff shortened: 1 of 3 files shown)


xxchan:
* **#20568** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20568?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#20911** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20911?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20911?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/stream/src/common/table/state_table.rs
```diff
@@ -1377,6 +1377,22 @@ where
         Ok(stream.map_ok(|(_, row)| row))
     }
 
+    /// Get the row from a state table with only 1 row.
+    pub async fn get_from_one_row_table(&self) -> StreamExecutorResult<Option<OwnedRow>> {
+        self.get_row(None)
```
xxchan, line 1382: Original impl is like this. But I found simple Agg is as simple as `get_row(None)`..
```rust
let sub_range: &(Bound<OwnedRow>, Bound<OwnedRow>) = &(Unbounded, Unbounded);
let stream = self
    .iter_with_prefix(row::empty(), sub_range, Default::default())
    .await?;
pin_mut!(stream);

if let Some(res) = stream.next().await {
    let value = res?.into_owned_row();
    assert!(stream.next().await.is_none());
    Ok(Some(value))
} else {
    Ok(None)
}
```

st1page (approved):
LGTM. I think it should be correct theoretically... So LGTM if it pass all test.

src/stream/src/common/table/state_table.rs
```diff
@@ -1377,6 +1377,22 @@ where
         Ok(stream.map_ok(|(_, row)| row))
     }
 
+    /// Get the row from a state table with only 1 row.
+    pub async fn get_from_one_row_table(&self) -> StreamExecutorResult<Option<OwnedRow>> {
```
BugenZhao, line 1381: I find it a bit ambiguous: having a table with exactly "one row" doesn't necessarily mean the key is empty. If we expect the exactly same behavior as before, a range scan seems still necessary. 🤔

xxchan: I think the case should not exist: if the key is non-empty and has only one row, then it's just a constant. 🤔

PR #20918, opened 2025-03-14
feat: add recovery events

shanicky:

This PR adds 6 types of events, including

#### For GLOBAL recovery
* `GLOBAL_RECOVERY_START` This event is attached next to the "recovery start" log with reason
* `GLOBAL_RECOVERY_SUCCESS` This event is attached next to the "recovery success" log with reason and recovery duration in seconds
* `GLOBAL_RECOVERY_FAILURE` This event is attached next to the "recovery failure" log with reason and error

#### For DATABASE recovery
* `DATABASE_RECOVERY_START` This event is attached next to the database `entering_recovery` from `running`
* `DATABASE_RECOVERY_SUCCESS` This event is attached next to the database `entering_running`
* `DATABASE_RECOVERY_FAILURE` This event is attached next to the database `entering_running` from `initializing`

The current functionality of this PR is usable, but a few issues need to be clarified.

1. Global recovery will recover databases that are in the `database_recovery_start` state. These databases will not enter the running state, which could lead to event loss. Therefore, at the end of global recovery, the `database_recovery_success` event will be triggered for databases that started in the Recovering state in global recovery.
2. When a database enters the recovery state, manually executing recover will also succeed. Is this expected behavior?
3. Which term is more appropriate: `database_recovery_start`, `database_entering_recovering`, or `database_recovery_recovering`? Do we need to retain the `recovery` keyword for querying in `rw_event_logs` using like `%recovery%`?
4. Do we need a mechanism to ensure when to enter database-level recovery and when to enter global-level recovery? The current behavior seems a bit unpredictable, making it difficult to write accurate simulation tests.

## Checklist

### DIFF
--- a/src/meta/src/barrier/checkpoint/recovery.rs
+++ b/src/meta/src/barrier/checkpoint/recovery.rs
@@ -303,7 +303,7 @@ impl DatabaseStatusAction<'_, EnterReset> {
         barrier_complete_output: Option<BarrierCompleteOutput>,
         control_stream_manager: &mut ControlStreamManager,
     ) {
-        let event_log_manager_ref = control_stream_manager.env.event_log_manager_ref();
+        let event_log_manager_ref = self.control.env.event_log_manager_ref();
         if let Some(output) = barrier_complete_output {
             self.control.ack_completed(output);
         }
@@ -486,9 +486,9 @@ impl DatabaseStatusAction<'_, EnterInitializing> {
 pub(crate) struct EnterRunning;
 
 impl DatabaseStatusAction<'_, EnterRunning> {
-    pub(crate) fn enter(self, control_stream_manager: &ControlStreamManager) {
+    pub(crate) fn enter(self) {
         info!(database_id = ?self.database_id, "database enter running");
-        let event_log_manager_ref = control_stream_manager.env.event_log_manager_ref();
+        let event_log_manager_ref = self.control.env.event_log_manager_ref();
         event_log_manager_ref.add_event_logs(vec![Event::Recovery(
             EventRecovery::database_recovery_success(self.database_id.database_id),
         )]);
--- a/proto/meta.proto
+++ b/proto/meta.proto
@@ -792,6 +792,8 @@ message EventLog {
     message GlobalRecoverySuccess {
       string reason = 1;
       float duration_secs = 2;
+      repeated uint32 running_database_ids = 3;
+      repeated uint32 recovering_database_ids = 4;
     }
     message GlobalRecoveryFailure {
       string reason = 1;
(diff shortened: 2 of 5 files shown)


wenym1:
> Global recovery will recover databases that are in the database_recovery_start state. These databases will not enter the running state, which could lead to event loss. Therefore, at the end of global recovery, the database_recovery_success event will be triggered for databases that started in the Recovering state in global recovery.

I think we can include the database id of successfully recovered databases and also the blocked databases in the GLOBAL_RECOVERY_SUCCESS  event, and then we don't need a separate database_recovery_success for each recovered database.

> When a database enters the recovery state, manually executing recover will also succeed. Is this expected behavior?

Indeed seems confusing, but currently the recover command is used mostly for operational purpose to manually trigger a global recovery, and I think it's ok to have behavior that is a bit confusing.

> Which term is more appropriate: database_recovery_start, database_entering_recovering, or database_recovery_recovering? Do we need to retain the recovery keyword for querying in rw_event_logs using like %recovery%?

Personally I prefer to be consistent to the global recovery.

> Do we need a mechanism to ensure when to enter database-level recovery and when to enter global-level recovery? The current behavior seems a bit unpredictable, making it difficult to write accurate simulation tests.

For error happened outside meta node (e.g. worker crash, actor exit), it's expected to trigger per database recovery only. Global recovery should only be triggered on meta node internal error (e.g. fail to access meta store).

Currently the inject barrier error is not handled in a per database manner, and should be fixed in a later PR.

wenym1:
Should we also have a `DATABASE_RECOVERY_FAILURE `, which is for database recovery failure?

wenym1:
> Do we have any way to track database-level recovery failures?
We can recognize a database-level recovery failure on ` DatabaseStatusAction<'_, EnterReset>::enter`. If we are going to enter Reset from a Running state, it's a recovery starts. If we are going to enter Reset from a Initializing state, it should be recovery failure.

We can have a try to include the concrete failure in the event. If it's too complicated to record the failure reason based on current code, we can at least record a database recovery failure event.

> Additionally, I noticed that I might not have tracked `global_database_failure`. Do you know when the "recovery failure" log typically is printed?

For global recovery, it's printed every time we failed to recover due to any error in `recover_inner`.

src/meta/src/barrier/checkpoint/recovery.rs
```diff
@@ -478,8 +486,12 @@ impl DatabaseStatusAction<'_, EnterInitializing> {
 pub(crate) struct EnterRunning;
 
 impl DatabaseStatusAction<'_, EnterRunning> {
-    pub(crate) fn enter(self) {
+    pub(crate) fn enter(self, control_stream_manager: &ControlStreamManager) {
```
wenym1, line 489: Looks dirty to pass the `control_stream_manager ` just for accessing the event log manager. We can store the `env` in `CheckpointControl` and then we can access the `env` from `self`. And also in `DatabaseStatusAction<'_, EnterReset> ` we can use `self` to access the `env`.

wenym1 (approved):
LGTM!

PR #20920, opened 2025-03-14
feat: introduce `rw_resource_groups` view 

shanicky:

This PR adds the `rw_resource_groups` view, which aggregates the number of workers in the cluster, total parallelism, and the number of databases and streaming jobs bound to this resource group.

Currently, this is implemented through a view, but in the future, if resource groups become a type of resource, we may need to turn this into an RPC operation.

```
dev=> select * from rw_resource_groups;
  name   | streaming_workers | parallelism | databases | streaming_jobs
---------+-------------------+-------------+-----------+----------------
 test    |                 2 |           8 |         0 |              1
 default |                 1 |           4 |         1 |              1
(2 rows)
```

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/catalog/system_catalog/rw_catalog/rw_resource_groups.rs
+++ b/src/frontend/src/catalog/system_catalog/rw_catalog/rw_resource_groups.rs
@@ -1,0 +1,63 @@
+// Copyright 2025 RisingWave Labs
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


src/frontend/src/catalog/system_catalog/rw_catalog/rw_resource_groups.rs
```diff
@@ -0,0 +1,63 @@
...
+             LEFT JOIN database_stats d ON ag.resource_group = d.resource_group
+             LEFT JOIN streaming_job_stats j ON ag.resource_group = j.resource_group"
+)]
+#[derive(Fields)]
+struct RwResourceGroup {
+    name: String,
+    streaming_workers: i64,
+    parallelism: i64,
+    databases: i64,
+    streaming_jobs: i64,
```
st1page, line 62: I guess it could be very large. Should we remove this field and let users join by themselves when they need it?
If we still like to maintain it, I am not sure if our optimizer can pruning it when it is not referred. Would you please send a explain result of `select * except (streaming_jobs);` here, thanks?

st1page: thanks

st1page (approved):

yezizp2012 (approved):

PR #17321, opened 2024-06-18
feat(sink): minor improvements for starrocks sink

ly9chee:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://gist.github.com/TennyZhuang/f00be7f16996ea48effb049aa7be4d66#file-rw_cla).

## What's changed and what's your intention?

This PR adds 3 minor improvements for StarRocks sink

#### 1. Support upserting `Aggregation` and `Unique` models
We currently only support upserting the `Primary Key` model, which is sufficient in most cases. However, the `Primary Key` model does not allow the table's primary key to exceed 128 bytes, which may be too restrictive in some cases. For example, in our environment, there are 10+ tables whose primary key exceed 128 bytes, a typical case is a MV's primary key is composed of 4 uuid strings, resulting in a primary key size of 144 bytes. As a result, we cannot sink these tables into StarRocks within RisingWave. The adhoc workaround is we sink these tables to kafka, and then we use flink to consume these topics and sink them to StarRocks.

Below is a demonstration of a **unique model** table `demo_unique`, with its primary key consisting of (pk1, pk2, pk3, pk4). The op field signifies the record's current operation, for example, `u` stands for `upsert` and `d` stands for `delete`.
| pk1                                       | pk2                                       | pk3                                       | pk4                                       | op |
|-------------------------------------------|-------------------------------------------|-------------------------------------------|-------------------------------------------|----|
| 7e672c80-6d31-45d4-84ce-5353f332d6e8      | 87202e74-b020-4676-b13f-2d87e06a114e      | 852045ec-3605-4adc-86a5-8f20a0d2e155      | 706cd645-ad43-42a6-9bae-5490a64472d7      | u  |
| 9c5b79d6-49b6-4f10-ac5c-1f359c1a4be4      | 28556f11-7f26-49b2-a095-5b82326aef8c      | b2aedcca-122e-4b57-a1f9-bdea5b60d642      | a1272ee7-5a90-4d8d-8b7c-c61fe4ec4f9c      | d  |

Under this PR, we can sink this MV to StarRocks directly, with no additional workaround needed:
```sql
CREATE SINK demo_unique_sink
FROM demo_upsert WITH (
    connector = 'starrocks',
    type = 'upsert',
    starrocks.host = 'starrocks-fe',
    starrocks.mysqlport = '9030',
    starrocks.httpport = '8030',
    starrocks.user = 'root',
    starrocks.password = '123456',
    starrocks.database = 'demo',
[...]

PR #20925, opened 2025-03-16
fix(sink): correctly get primary keys from starrocks aggregate table

ly9chee:

This PR fixes the pk validation issue for aggregate table.

Currently, we support customized upsert unique key table through the [changelog](https://github.com/risingwavelabs/risingwave/pull/17321), however, the aggregate table should be taken into account. Because the unique key table is just a special case of aggregate table where `agg_type` of all columns is set to `REPLACE`.

In the sink creation phase, we retrieve the primary keys from `tables_config.primary_key` field, which is true for the primary key table and unique key table, but the primary keys of an aggregate table should be retrieved from `sort_key` field. Before starrocks 3.3.0, the key columns of unique key, duplicate key and aggregate table were [coupled](https://docs.starrocks.io/docs/3.2/table_design/table_types/table_capabilities/) with sort key. Since starrocks 3.3.0, the order of key columns may differ from sort key columns, but the columns remain the same. Thus, it is safe to use the `sort_key` as the primary key for sink validation.

## Checklist

## Documentation

<details>

</details>


ly9chee:
@xxhZs @wenym1 @hzxa21 PTAL

wenym1:
Thanks for raising this PR! Could you give an example in what will be changed in the behavior? Like a sink that you used to be unable to create and able to create with this PR.

xxhZs (approved):

ly9chee:
> Thanks for raising this PR! Could you give an example in what will be changed in the behavior? Like a sink that you used to be unable to create and able to create with this PR.

Sure, PR description updated.

wenym1 (approved):

PR #20937, opened 2025-03-17
fix(snapshot-backfill): fix panic in cross db snapshot backfill recovery

wenym1:

Hits the following panic for cross db snapshot backfill after recovery.
```
thread 'rw-streaming' panicked at /Users/william/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/itertools-0.14.0/src/zip_eq_impl.rs:50:17:
itertools: .zip_eq() reached end of one iterator before the other
```

We used the wrong column size to resolve the consume progress from the progress rows loaded from progress state table. The bug will be reproduced when backfill recover from half of an epoch.

Add an e2e test to cover the case. Before this PR the e2e will failed with the panic above.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/ci/scripts/run-backfill-tests.sh
+++ b/ci/scripts/run-backfill-tests.sh
@@ -410,6 +410,16 @@ test_scale_in() {
   risedev clean-data
 }
 
+test_cross_db_snapshot_backfill() {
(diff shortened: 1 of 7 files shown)


yezizp2012 (approved):
LGTM

kwannoel (approved):

PR #20948, opened 2025-03-18
refactor(meta): exclude table change log from time travel metadata

zwang28:

Previously the table change log metadata was included in the time travel metadata, because there was a possibility we'd leverage time travel to improve table change log management.

Since the potential improvement does not appear to be planned for the near future, this PR removes table change logs from time travel metadata, to reduce overhead which can be substantial with a long subscription retention period.

[Storage log iterator](https://github.com/risingwavelabs/risingwave/blob/ec899adc3f7e8b23f555657972368dc7ef6ba6c3/src/storage/src/hummock/store/hummock_storage.rs#L709) will never use time travel HummockVersion anyway.

## Checklist

## Documentation

<details>

</details>


src/storage/hummock_sdk/src/version.rs
```diff
@@ -579,35 +579,49 @@ where
...
             .collect()
     }
 
+    pub fn newly_added_sst_ids_exclude_change_log(&self) -> HashSet<HummockSstableObjectId> {
+        self.newly_added_sst_infos_exclude_change_log()
+            .map(|sst| sst.sst_id())
+            .collect()
+    }
+
     pub fn newly_added_sst_infos(&self) -> impl Iterator<Item = &'_ T> {
```
wenym1, line 600: Shall we just add a new parameter named like `exclude_change_log` instead of having a new method for the case?

wenym1 (approved):
Rest LGTM

PR #20848, opened 2025-03-11
feat(snapshot-backfill): support recoverable snapshot backfill (meta part)

wenym1:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## What's changed and what's your intention?

In this PR, we will support recoverable snapshot backfill in meta node. During recovery, we will recover the ongoing snapshot backfill states by resolving the lagged epochs from hummock table change log epoch.

## Checklist

- [ ] I have written necessary rustdoc comments.
- [ ]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  My PR contains critical fixes that are necessary to be merged into the latest release. 

## Documentation

- [ ]  My PR needs documentation updates. 

<details>
<summary><b>Release note</b></summary>

</details>

PR #20847, opened 2025-03-11
feat(snapshot-backfill): support recoverable snapshot backfill (executor part)

wenym1:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## Checklist

- [ ] I have written necessary rustdoc comments.
- [ ]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  My PR contains critical fixes that are necessary to be merged into the latest release. 

## Documentation

- [ ]  My PR needs documentation updates. 

<details>
<summary><b>Release note</b></summary>

</details>

PR #20952, opened 2025-03-18
refactor(snapshot-backfill): refine snapshot backfill tracking logic in meta

wenym1:

Before the [PR](https://github.com/risingwavelabs/risingwave/pull/20848) to support recoverable snapshot backfill, we will refine the current logic of tracking snapshot backfill in global barrier manager, so that we can simplify the implementation in the subsequent PR.

We will make the following changes:
- Previously, we store initial mutation and new actors to create in the tracker of snapshot backfill(`CreatingStreamingJobControl`), and take them out to include it when later injecting the first barrier. In this PR we will change to injecting the first barrier along with creating a new `CreatingStreamingJobControl`, so that we don't have to store the initial mutation and new actors.
- We handle the transition from `ConsumingSnapshot` to `ConsumingLogStore` when we collect a new `CompleteBarrierResponse` and see that all actors have finished consuming the snapshot. During this transition, we will inject all the pending upstream barriers immediately in this `collect_barrier` method. In this PR, we will change to store the pending upstream barriers in `ConsumingLogStore` first, and then when handling the next upstream barrier, we then inject all the pending barriers. The reason for this change is to avoid injecting barrier in the `collect` of `CreatingStreamingJobControl`.
- extract the logic of collecting initial barriers during recovery to a `DatabaseInitialBarrierCollector` and provide some util methods
- simplify some fields for the structs that track snapshot backfill. For example, previously we store a `StreamJobFragment` as a field, but we actually only access its `job_id`, and then we change to only store a `job_id`

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/meta/src/barrier/checkpoint/control.rs
+++ b/src/meta/src/barrier/checkpoint/control.rs
@@ -48,8 +48,7 @@ use crate::barrier::utils::{
     NodeToCollect, collect_creating_job_commit_epoch_info, is_valid_after_worker_err,
 };
 use crate::barrier::{
-    BarrierKind, Command, CreateStreamingJobCommandInfo, CreateStreamingJobType,
-    InflightSubscriptionInfo, SnapshotBackfillInfo, TracedEpoch,
+    BarrierKind, Command, CreateStreamingJobType, InflightSubscriptionInfo, TracedEpoch,
 };
 use crate::manager::MetaSrvEnv;
 use crate::rpc::metrics::GLOBAL_META_METRICS;
@@ -129,14 +128,13 @@ impl CheckpointControl {
     pub(crate) fn barrier_collected(
         &mut self,
         resp: BarrierCompleteResponse,
-        control_stream_manager: &mut ControlStreamManager,
         periodic_barriers: &mut PeriodicBarriers,
     ) -> MetaResult<()> {
         let database_id = DatabaseId::new(resp.database_id);
         let database_status = self.databases.get_mut(&database_id).expect("should exist");
         match database_status {
             DatabaseCheckpointControlStatus::Running(database) => {
-                database.barrier_collected(resp, control_stream_manager, periodic_barriers)
+                database.barrier_collected(resp, periodic_barriers)
             }
             DatabaseCheckpointControlStatus::Recovering(state) => {
                 state.barrier_collected(database_id, resp);
@@ -375,11 +373,7 @@ impl CheckpointControl {
                 .values()
             {
                 progress.extend([(
-                    creating_job
-                        .info
-                        .stream_job_fragments
-                        .stream_job_id()
-                        .table_id,
+                    creating_job.job_id.table_id,
                     creating_job.gen_ddl_progress(),
                 )]);
             }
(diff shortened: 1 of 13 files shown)


wenym1:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20952?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#20952** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20952?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20952?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#20847** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20847?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#20848](https://github.com/risingwavelabs/risingwave/pull/20848) <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20848?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

src/meta/src/barrier/checkpoint/control.rs
```diff
@@ -441,16 +435,27 @@ impl CheckpointControl {
...
     pub(crate) fn inflight_infos(
         &self,
-    ) -> impl Iterator<Item = (DatabaseId, &InflightSubscriptionInfo, &InflightDatabaseInfo)> + '_
-    {
+    ) -> impl Iterator<
+        Item = (
+            DatabaseId,
+            &InflightSubscriptionInfo,
+            &InflightDatabaseInfo,
+            impl Iterator<Item = &InflightStreamingJobInfo> + '_,
```
kwannoel, line 443: This is a fix for snapshot backfill inflight info, separate from database info. So we didn't recover it previously.

src/meta/src/barrier/checkpoint/control.rs
```diff
@@ -129,14 +128,13 @@ impl CheckpointControl {
...
         &mut self,
         resp: BarrierCompleteResponse,
-        control_stream_manager: &mut ControlStreamManager,
         periodic_barriers: &mut PeriodicBarriers,
     ) -> MetaResult<()> {
         let database_id = DatabaseId::new(resp.database_id);
         let database_status = self.databases.get_mut(&database_id).expect("should exist");
         match database_status {
             DatabaseCheckpointControlStatus::Running(database) => {
-                database.barrier_collected(resp, control_stream_manager, periodic_barriers)
```
kwannoel, line 139: We remove `control_Stream_manager` because it's only used to inject barrier, and we no longer need to do so.

src/meta/src/barrier/checkpoint/control.rs
```diff
@@ -441,16 +435,27 @@ impl CheckpointControl {
...
-                )
-            })
+            database
+                .database_state()
+                .map(|(database_state, creating_jobs)| {
+                    (
+                        *database_id,
+                        &database_state.inflight_subscription_info,
+                        &database_state.inflight_graph_info,
+                        creating_jobs
```
kwannoel, line 454: As per above, we need to recover the graph of creating jobs as well.

src/meta/src/barrier/checkpoint/creating_job/barrier_control.rs
```diff
@@ -43,7 +44,7 @@ pub(super) struct CreatingStreamingJobBarrierControl {
     // key is prev_epoch of barrier
     inflight_barrier_queue: BTreeMap<u64, CreatingStreamingJobEpochState>,
     backfill_epoch: u64,
-    initial_epoch: Option<u64>,
+    is_first_committed: bool,
```
kwannoel, line 47: nit: `is_first_committed` reads really similar to `is_first_commit`. However the usage is different. I think the intention is to track if the first epoch has been committed.

`first_epoch_committed` could be a better name.

kwannoel (approved):
LGTM

PR #20967, opened 2025-03-18
feat(meta,frontend,streaming): support fixed backfill order control

kwannoel:

In this PR we support **fixed** backfill orders like so:
```sql
create materialized view m1 with (backfill_order = FIXED(t -> t2)) as select t.v1 from t join t2 using(v1);
```

The backfill orders will initially be specified as a list of edges, between table / source names being scanned. This will be parsed and bound to a list of edges between table / source ids.

Then we will transform it to an adjacency list and send this info to the meta node. The meta node constructs a mapping between table ids to the corresponding fragments containing the table scan. Next, it converts the adjacency list of table ids to an adjacency list of fragments.

Finally, it will build the initial backfill order state, which consists of the root nodes, and remaining nodes.

To execute the backfill order, each time an actor is finished with backfill, we will update the progress of the corresponding fragment. Once all actors are finished for a fragment, we will update its downstreams, and queue them if all their dependencies are finished.

### Frontend

We implement it inside the `with_options` clause, rather than as a session variable.  This is because the backfill order should be parsed as a graph of `object_names`. So it's a little hard to specify that inside a session variable, since there will be a dependency introduced between the `common` module, which contains the `session_config` module, to the `sqlparser::ast` module.

This PR changes the frontend, to introduce fixed ordering strategy, supporting the parsing of it, and also binding of it into physical relation ids. We add a field to `StreamFragments`, so this info can be propagated together with the create stream job rpc call.

### Meta

Inside the meta node, we add a new field to `CreateMViewProgressTracker`, the `backfill_order_state`. This will track the fragment level progress of backfill. Once a node's dependencies finished backfilling, we will schedule it for backfill. This continues until all nodes are backfilled. We piggyback on barriers with no commands to schedule a command which triggers the backfill for the backfill node(s).

### Streaming

Inside the backfill executors (no shuffle and arrangement backfill), we handle the start backfill barrier command, which will trigger backfilling for a backfill node, based on the fragment id.

We ended up not using rate limit to pause / resume backfilling, because it conflicts with backfill order control. When backfilling node is not ready for backfilling, because its dependencies are not done, altering its rate limit could inadvertently resume it. So we separate backfill pause and rate limit pause.

## Performance

I tested with `backfill_order.slt`, using the release build. `car_regions` and `car_info` are dimension tables. `car_sales` is the fact table.

- [ ] Test out left join 4-5 dimension table.

| order | total test duration (ms) |
| - | - |
| car_regions -> car_sales, car_info -> car_sales | 22392 |
| car_sales -> car_regions, car_sales -> car_regions | 30687 |
| None | 20000-30000 (fluctuates) |

That's a 33% improvement over the worst case.

## Future Work

We don't support (yet):
- [ ] Auto-derive backfill order
- [ ] Observe backfill progress per node
- [ ] Backfill order for sinks
- [ ] Backfill order control under background_ddl
- [ ] Backfill order control for snapshot backfill
- [ ] Backfill order control for stream source
- [x] Cycle detection

## Checklist

## Documentation

<details>

</details>


src/frontend/src/binder/backfill_order_strategy.rs
```diff
@@ -0,0 +1,99 @@
...
+};
+use risingwave_sqlparser::ast::{BackfillOrderStrategy, ObjectName};
+
+use crate::Binder;
+use crate::catalog::CatalogError;
+use crate::catalog::schema_catalog::SchemaCatalog;
+use crate::error::Result;
+use crate::session::SessionImpl;
+
+// FIXME(kwannoel): we can flatten the strategy earlier
```
kwannoel, line 32: Flatten before merging

src/frontend/src/binder/backfill_order_strategy.rs
```diff
@@ -0,0 +1,99 @@
...
+    BackfillOrderFixed, BackfillOrderStrategy as PbBackfillOrderStrategy, BackfillOrderUnspecified,
+};
+use risingwave_sqlparser::ast::{BackfillOrderStrategy, ObjectName};
+
+use crate::Binder;
+use crate::catalog::CatalogError;
+use crate::catalog::schema_catalog::SchemaCatalog;
+use crate::error::Result;
+use crate::session::SessionImpl;
+
```
kwannoel, line 31: Add some docs: why we don't use Binder, because binder stores relation context, and will lead to conflicts. It also binds a lot of extra stuff like system catalog etc..

src/stream/src/executor/backfill/no_shuffle_backfill.rs
```diff
@@ -453,6 +468,11 @@ where
                         Mutation::Resume => {
                             paused = false;
                         }
+                        Mutation::StartFragmentBackfill { fragment_ids } => {
```
kwannoel, line 471: Only check if currently paused

kwannoel:
* **#20967** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20967?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/20967?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lmatz:
> I tested with backfill_order.slt, using the release build. car_regions and car_info are dimension tables. car_sales is the fact table.
That's a 33% improvement over the worst case.

I guess with this size of data, everything is cached in memory?
If it cannot be fully cached, the improvement could be even larger.

proto/stream_plan.proto
```diff
@@ -129,6 +133,12 @@ message BarrierMutation {
     ThrottleMutation throttle = 10;
     // Drop subscription on mv
     DropSubscriptionsMutation drop_subscriptions = 12;
+    // Start backfilling for specific fragments
+    // This is separated from `ThrottleMutation::NO_RATE_LIMIT`.
+    // This is because user may concurrently rate limit + use backfill order control.
+    // If we use rate limit to pause / resume backfill fragments, if user manually
+    // resumes some fragments, this will overwrite the backfill order configuration.
```
BugenZhao, line 140: I think this is not bad, if the user intends some intervention? As a result, we don't have to introduce a new control plane.

proto/stream_plan.proto
```diff
@@ -1029,6 +1042,19 @@ message StreamContext {
...
 }
 
+message BackfillOrderUnspecified {}
+
+message BackfillOrderFixed {
+  map<uint32, common.Uint32Vector> order = 1;
+}
+
+message BackfillOrderStrategy {
+  oneof strategy {
```
BugenZhao, line 1052: `oneof` itself can be `None`. So do we still need `unspecified`?

src/frontend/src/binder/backfill_order_strategy.rs
```diff
@@ -0,0 +1,102 @@
...
+    let reader = session.env().catalog_reader().read_guard();
+
+    match schema_name {
+        Some(name) => {
+            let schema_catalog = reader.get_schema_by_name(&db_name, &name)?;
+            bind_source_or_table(schema_catalog, &rel_name)
+        }
+        None => {
+            let search_path = session.config().search_path();
+            for path in search_path.path() {
```
BugenZhao, line 77: Shall we use `SchemaPath::try_find` for DRY?

src/sqlparser/src/ast/mod.rs
```diff
@@ -1535,6 +1535,10 @@ pub enum Statement {
         local: bool,
         value: SetTimeZoneValue,
     },
+    /// `SET BACKFILL_ORDER_STRATEGY TO { DEFAULT | NONE | AUTO | FIXED <object_name> -> <object_name>, ... }
```
BugenZhao, line 1538: According to the PR body, shall we remove this?

e2e_test/backfill/backfill_order_control.slt
```diff
@@ -0,0 +1,83 @@
...
+    region_id_gen.id as region_id,
+    sales_id_gen.id as price
+  FROM generate_series(1, 100000) as sales_id_gen(id)
+    CROSS JOIN generate_series(1, 5) as car_id_gen(id)
+    CROSS JOIN generate_series(1, 5) as region_id_gen(id);
+
+# Create an MV that groups sales by a range of prices
+statement ok
+create materialized view m1
+    with (backfill_order = FIXED car_regions -> car_sales, car_info -> car_sales)
```
BugenZhao, line 49: Shall we encapsulate the order value into a string, so that we don't break the Postgres syntax too much? For example, this can be more friendly to formatters.

BugenZhao, line 49: Also, is there any reason for putting the `WITH` clause before `AS`? 🥵 This seems inconsistent with `CREATE SINK`.

src/meta/src/barrier/backfill_order_control.rs
```diff
@@ -0,0 +1,149 @@
...
+    // The order plan.
+    current_backfill_nodes: HashMap<FragmentId, BackfillNode>,
+    // Remaining nodes to finish
+    remaining_backfill_nodes: HashMap<FragmentId, BackfillNode>,
+    // The mapping between actors and fragment_ids
+    actor_to_fragment_id: HashMap<ActorId, FragmentId>,
+}
+
+// constructor
+impl BackfillOrderState {
```
BugenZhao, line 51: What if users accidentally specify a cycle and we cannot do topological sort here?

src/meta/src/barrier/progress.rs
```diff
@@ -312,8 +336,11 @@ impl CreateMviewProgressTracker {
                 backfill_upstream_types.insert(actor, backfill_upstream_type);
             }
 
+            let backfill_order_state = backfill_order_map.remove(&creating_table_id);
```
BugenZhao, line 339: Can you explain this line?

src/meta/src/barrier/checkpoint/control.rs
```diff
@@ -609,6 +610,9 @@ pub(crate) struct DatabaseCheckpointControl {
     create_mview_tracker: CreateMviewProgressTracker,
 
     metrics: DatabaseCheckpointControlMetrics,
+
```
wenym1, line 613: We can store the field in `CreateMviewProgressTracker`, so that the logic can be naturally ported to snapshot backfill. When handling new barriers, we can inspect the specific field in `CreateMviewProgressTracker` rather than this field in `DatabaseCheckpointControl`.

e2e_test/backfill/backfill_order_control.slt
kwannoel: It's just the current way we parse `WITH` options for MV. I didn't change the behaviour. I also prefer it to be after the MV statement. But this change is out of scope.

src/meta/src/barrier/backfill_order_control.rs
kwannoel: Yeah I previously thought of this as well, and mentioned it in the future work section (Cycle detection).

The PR is large at the moment. I will refine this later.

src/meta/src/barrier/checkpoint/control.rs
kwannoel: Makes sense, thanks for the suggestion

proto/stream_plan.proto
kwannoel: We can introduce separate path / features for that, rather than overlap into rate limit.

e2e_test/backfill/backfill_order_control.slt
kwannoel: Yeah, I will do something like `FIXED("car_regions -> car_sales, ...")` instead.

kwannoel: `FIXED(car_regions -> car_sales, ...)`.

Because the tokenizer will take `'car_regions -> car_sales, ...'` as a single token.

src/meta/src/barrier/backfill_order_control.rs
kwannoel: https://github.com/risingwavelabs/risingwave/pull/20967/commits/9b4ac0c570a87b5d4a93e863a1e9376e7752e11e

kwannoel: Added

src/meta/src/barrier/progress.rs
kwannoel: Backfill orders are tracked per job. So we just extract the backfill order graph for a specific job (`creating_table_id`) and use it for recovery.

src/meta/src/barrier/checkpoint/control.rs
kwannoel: https://github.com/risingwavelabs/risingwave/pull/20967/commits/39760bd58bdf2e328aa7928c15818cbe9e771118

kwannoel:
Bump

src/meta/src/barrier/rpc.rs
```diff
@@ -526,7 +526,13 @@ impl ControlStreamManager {
                     (job_id, (definition.clone(), stream_job_fragments))
                 })
         });
-        let tracker = CreateMviewProgressTracker::recover(background_mviews, hummock_version_stats);
+        // FIXME(kwannoel): recover this for background ddl
+        let recovered_backfill_order_map = HashMap::new();
```
chenzl25, line 530: What will happen if the creating MV with backfill order recovers with an empty state?

chenzl25: IIUC, it will be treated as if without any backfill order.

chenzl25:
If a table is used in the SQL multiple times, e.g., `create materialized view mv with(backfill_order = FIXED (t1 -> t2)) as select count(*) from t1 as x, t1 as y, t2 where .... `. Will these tables backfill at the same time?

src/sqlparser/src/ast/mod.rs
```diff
@@ -3548,6 +3552,47 @@ impl fmt::Display for DiscardType {
...
+// so we can choose strategies that make the most sense.
+#[derive(Debug, Default, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub enum BackfillOrderStrategy {
+    #[default]
+    Default,
+    None,
+    Auto,
+    Fixed(Vec<(ObjectName, ObjectName)>),
+}
```
chenzl25, line 3565: What's the difference among defult, none, and auto?

chenzl25 (commented):
Rest LGTM

kwannoel:
> In this example, `t -> t2` meaning t depends on t2 or meaning first backfilling t then t2? IIUC, the code implementation looks like the prior one. We should document it for better understanding, including the syntax.
> 
> ```
> create materialized view m1 with (backfill_order = FIXED t -> t2 ) as select t.v1 from t join t2 using(v1);
> ```

This means `t` backfill first then `t2`.

kwannoel:
> If a table is used in the SQL multiple times, e.g., `create materialized view mv with(backfill_order = FIXED (t1 -> t2)) as select count(*) from t1 as x, t1 as y, t2 where .... `. Will these tables backfill at the same time?

Yes they will. The implementation is such that we will bind `t1` using the session context to a table id.
Then all scans on the same table id will be scheduled together.

src/frontend/src/binder/backfill_order_strategy.rs
```diff
@@ -0,0 +1,142 @@
...
+use crate::catalog::root_catalog::SchemaPath;
+use crate::catalog::schema_catalog::SchemaCatalog;
+use crate::error::Result;
+use crate::session::SessionImpl;
+
+// FIXME(kwannoel): we can flatten the strategy earlier
+/// We don't use query binder directly because its binding rules are different from a query.
+/// We only bind tables, materialized views and sources.
+/// Queries won't bind duplicate relations in the same query context.
+/// But backfill order strategy can have duplicate relations.
```
kwannoel, line 37: initially just t->t2, after this phase table_id(t) -> table_id(t2)`

src/meta/src/barrier/backfill_order_control.rs
```diff
@@ -0,0 +1,149 @@
...
+        stream_job_fragments: &StreamJobFragments,
+    ) -> Self {
+        tracing::debug!(?backfill_orders, "initialize backfill order state");
+        let actor_to_fragment_id = stream_job_fragments.actor_fragment_mapping();
+
+        let mut backfill_nodes: HashMap<FragmentId, BackfillNode> = HashMap::new();
+
+        for fragment in stream_job_fragments.fragments() {
+            if fragment.fragment_type_mask
+                & (FragmentTypeFlag::StreamScan as u32 | FragmentTypeFlag::SourceScan as u32)
```
chenzl25, line 63: Does backfill order control support source backfill? I don't see any changes to the `SourceBackfillExecutor`.

kwannoel: Not in the initial parts, but eventually we should. So this PR only implements some easy parts for it.
Detailed in the Future Work section of the PR sescription

src/meta/src/barrier/backfill_order_control.rs
```diff
@@ -0,0 +1,149 @@
...
+}
+
+// state transitions
+impl BackfillOrderState {
+    pub fn finish_actor(&mut self, actor_id: ActorId) -> Vec<FragmentId> {
+        // Find the fragment_id of the actor.
+        let fragment_id = self.actor_to_fragment_id.get(&actor_id).unwrap();
+        // Decrease the remaining_actor_count of the operator.
+        // If the remaining_actor_count is 0, add the operator to the current_backfill_nodes.
+        let node = self.current_backfill_nodes.get_mut(fragment_id).unwrap();
```
chenzl25, line 123: If that's the case, I think we can't unwrap the `current_backfill_nodes`, because source backfill could finish at any time.

kwannoel: I have banned it in the binder first.

kwannoel: https://github.com/risingwavelabs/risingwave/pull/20967/commits/217c501af6345c918d4a64270748a82e05decf6f

kwannoel:
blocked by flaky test https://github.com/risingwavelabs/risingwave/issues/21355

kwannoel:
previously passed CI.

src/meta/src/barrier/context/context_impl.rs
```diff
@@ -321,6 +321,7 @@ impl CommandContext {
             }
             Command::DropSubscription { .. } => {}
             Command::MergeSnapshotBackfillStreamingJobs(_) => {}
+            Command::StartFragmentBackfill { .. } => {}
```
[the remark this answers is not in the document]
kwannoel: No need to. Recovery is not the focus of this PR.

src/stream/src/executor/mod.rs
```diff
@@ -314,6 +314,9 @@ pub enum Mutation {
         /// `subscriber` -> `upstream_mv_table_id`
         subscriptions_to_drop: Vec<(u32, TableId)>,
     },
+    StartFragmentBackfill {
+        fragment_ids: Vec<FragmentId>,
```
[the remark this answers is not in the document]
kwannoel: fixed

e2e_test/backfill/backfill_order_control.slt
```diff
@@ -0,0 +1,101 @@
...
+create materialized view m1
+  with (backfill_order = FIXED(car_sales -> car_regions, car_info -> car_sales, car_regions -> car_sales))
+as select count(*) from car_sales join car_info
+  on car_sales.car_id = car_info.id
+  join car_regions
+    on car_sales.region_id = car_regions.id;
+
+# Create an MV that groups sales by a range of prices
+statement ok
+create materialized view m1
```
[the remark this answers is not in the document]
kwannoel: fixed

src/meta/src/barrier/progress.rs
```diff
@@ -465,12 +508,20 @@ impl CreateMviewProgressTracker {
         self.pending_finished_jobs.push(finished_job);
     }
 
+    pub(super) fn queue_backfill(&mut self, backfill_nodes: Vec<FragmentId>) {
```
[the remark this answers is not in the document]
kwannoel: fixed

src/meta/src/barrier/backfill_order_control.rs
```diff
@@ -0,0 +1,149 @@
...
+            current_backfill_nodes,
+            remaining_backfill_nodes,
+            actor_to_fragment_id,
+        }
+    }
+}
+
+// Getters
+impl BackfillOrderState {
+    pub fn get_initial_nodes(&self) -> Vec<FragmentId> {
```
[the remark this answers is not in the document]
kwannoel: It makes sense from the implementation perspective, because the initial nodes to backfill are equivalent to the initial nodes currently.

The intention is that we can clearly read this meaning at the call site:

```
                        if let Some(order) = &info.backfill_order_state {
                            self.queue_backfill(order.get_initial_nodes());
                        }
```

We don't have to understand how it gets the initial nodes, since that's an implementation detail.

src/meta/src/barrier/progress.rs
```diff
@@ -429,7 +467,12 @@ impl CreateMviewProgressTracker {
         let new_tracking_job_info =
             if let Some(Command::CreateStreamingJob { info, job_type, .. }) = command {
                 match job_type {
-                    CreateStreamingJobType::Normal => Some((info, None)),
+                    CreateStreamingJobType::Normal => {
```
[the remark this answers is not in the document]
kwannoel: Yes, if backfill_order is specified.

src/frontend/src/binder/backfill_order_strategy.rs
```diff
@@ -0,0 +1,145 @@
...
+            }
+            stack.remove(&node);
+        }
+        false
+    }
+
+    let mut visited = HashSet::new();
+    let mut stack = HashSet::new();
+    for &start in order.keys() {
+        if !visited.contains(&start) && dfs(start, order, &mut visited, &mut stack) {
```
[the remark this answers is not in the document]
kwannoel: Beginning of dfs only includes:
```
        if stack.contains(&node) {
            return true; // Cycle detected
        }
```

src/meta/src/barrier/progress.rs
```diff
@@ -614,16 +668,19 @@ impl CreateMviewProgressTracker {
                     for actor in o.get().0.actors() {
                         self.actor_map.remove(&actor);
                     }
-                    Some(o.remove().1)
+                    UpdateProgressResult::Finished(o.remove().1)
```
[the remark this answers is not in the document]
kwannoel: fixed

src/meta/src/barrier/backfill_order_control.rs
```diff
@@ -0,0 +1,149 @@
...
+/// This is the "global" `fragment_id`.
+/// The local `fragment_id` is namespaced by the `fragment_id`.
+pub type ActorId = u32;
+
+#[derive(Clone, Debug, Default)]
+pub struct BackfillNode {
+    fragment_id: FragmentId,
+    /// How many more actors need to finish,
+    /// before this fragment can finish backfilling.
+    remaining_actor_count: usize,
```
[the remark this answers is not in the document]
kwannoel: fixed https://github.com/risingwavelabs/risingwave/pull/20967/commits/37a95c7b27b75e9dc4c040ed455a3ccfc6bcceb1

src/meta/src/barrier/backfill_order_control.rs
```diff
@@ -0,0 +1,149 @@
...
+            }
+        }
+
+        for (fragment_id, children) in backfill_orders {
+            for child in &children {
+                let child_node = backfill_nodes.get_mut(child).unwrap();
+                child_node.remaining_dependency_count += 1;
+            }
+
+            let backfill_node = backfill_nodes.get_mut(&fragment_id).unwrap();
```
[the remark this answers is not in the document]
kwannoel: fixed https://github.com/risingwavelabs/risingwave/pull/20967/commits/477bcad3113081d035f56d173d8ac75627c2acce

src/meta/src/barrier/progress.rs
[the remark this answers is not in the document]
kwannoel: Moved this logic to `add` in https://github.com/risingwavelabs/risingwave/pull/20967/commits/4ad647d40f95f44e5d6fd340d4a8c4d09ccab5a4

src/meta/src/stream/stream_manager.rs
```diff
@@ -97,6 +97,8 @@ pub struct CreateStreamingJobContext {
     pub option: CreateStreamingJobOption,
 
     pub streaming_job: StreamingJob,
+
+    pub backfill_order_state: Option<BackfillOrderState>,
```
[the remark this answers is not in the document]
kwannoel: Fixed https://github.com/risingwavelabs/risingwave/pull/20967/commits/4ad647d40f95f44e5d6fd340d4a8c4d09ccab5a4

src/meta/src/barrier/command.rs
```diff
@@ -177,6 +178,7 @@ pub struct CreateStreamingJobCommandInfo {
     pub create_type: CreateType,
     pub streaming_job: StreamingJob,
     pub internal_tables: Vec<Table>,
+    pub backfill_order_state: Option<BackfillOrderState>,
```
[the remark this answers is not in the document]
kwannoel: Fixed https://github.com/risingwavelabs/risingwave/pull/20967/commits/4ad647d40f95f44e5d6fd340d4a8c4d09ccab5a4

src/frontend/src/optimizer/plan_node/stream_table_scan.rs
```diff
@@ -283,6 +283,13 @@ impl StreamTableScan {
     ) -> SchedulerResult<PbStreamNode> {
         use risingwave_pb::stream_plan::*;
 
+        let backfill_paused = self
```
wenym1, line 286: Currently we set the initial `backfill_paused` in frontend and then persist this value and never update it later. But as described in the other comment, after recovery if we don't reset this value to false, the paused backfill may never be unpaused.

So I am thinking about a cleaner implementation. Here in frontend, we may always initial the value to false, so that the persisted value is always false and after recovery all backfill nodes are not initially paused. The work of setting this `backfill_paused` field can be handled by barrier manager. After it resolve the initial `BackfillOrderState`, it set `backfill_paused=false` for nodes without dependency, and as `true` for nodes with dependency. We can update the node field via `visit_stream_node_mut`.

e2e_test/backfill/backfill_order_control.slt
```diff
@@ -0,0 +1,106 @@
...
+        max(price) as sales_max,
+        approx_percentile(0.5) WITHIN GROUP (ORDER BY price) as sales_est_median,
+        approx_percentile(0.01) WITHIN GROUP (ORDER BY price) as sales_est_bottom_1_percent,
+        approx_percentile(0.99) WITHIN GROUP (ORDER BY price) as sales_est_top_1_percent
+      FROM
+        price_ranges
+      GROUP BY name, price_range;
+
+query I
+select count(*) from m1;
```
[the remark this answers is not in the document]
kwannoel: Nice technique of using `except`

src/frontend/src/optimizer/plan_node/stream_table_scan.rs
kwannoel: Addressed in https://github.com/risingwavelabs/risingwave/pull/21474

src/meta/src/barrier/rpc.rs
```diff
@@ -505,7 +505,13 @@ impl ControlStreamManager {
...
+        let tracker = CreateMviewProgressTracker::recover(
+            background_mviews,
+            hummock_version_stats,
+            recovered_backfill_order_map,
```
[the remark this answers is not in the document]
kwannoel: Addressed in: https://github.com/risingwavelabs/risingwave/pull/21474

src/frontend/src/binder/backfill_order_strategy.rs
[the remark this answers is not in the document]
kwannoel: Fixed

src/meta/src/barrier/backfill_order_control.rs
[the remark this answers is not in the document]
kwannoel: Removed in https://github.com/risingwavelabs/risingwave/pull/21474/commits/e5f972c98a63d3c51a90e362d734f0778d5fbd34

src/meta/src/barrier/progress.rs
```diff
@@ -299,9 +322,11 @@ impl CreateMviewProgressTracker {
     pub fn recover(
         mviews: impl IntoIterator<Item = (TableId, (String, &StreamJobFragments))>,
         version_stats: &HummockVersionStats,
+        mut backfill_order_map: HashMap<TableId, BackfillOrderState>,
```
wenym1, line 325: Better to include the `BackfillOrderState` to the `mviews` iter.

src/meta/src/barrier/progress.rs
```diff
@@ -328,6 +357,7 @@ impl CreateMviewProgressTracker {
             progress_map,
             actor_map,
             pending_finished_jobs: Vec::new(),
+            pending_backfill_nodes,
```
wenym1, line 360: Seems more straight forward to avoid initially pausing all backfill nodes when recovery and just initialize with an empty `pending_backfill_nodes`.

src/meta/src/barrier/progress.rs
```diff
@@ -338,6 +368,7 @@ impl CreateMviewProgressTracker {
     /// and then it will just report progress like newly created executors.
     fn recover_progress(
         states: HashMap<ActorId, BackfillState>,
+        backfill_order_state: Option<BackfillOrderState>,
```
wenym1, line 371: It seems that in this PR we haven't recovered the specified backfill order control yet in recovery. It so, the backfill_order_state here should always be None.

Given that this PR is already very large, we may remove this logic of passing the recovered backfill_order_map and include it in the PR that supports recovering the backfill order.

src/meta/src/stream/stream_graph/fragment.rs
```diff
@@ -682,6 +696,74 @@ impl StreamFragmentGraph {
...
+    /// Collect the mapping from table / `source_id` -> `fragment_id`
+    pub fn collect_backfill_mapping(&self) -> HashMap<u32, Vec<FragmentId>> {
+        let mut mapping = HashMap::new();
+        for (fragment_id, fragment) in &self.fragments {
+            let fragment_id = fragment_id.as_global_id();
+            let fragment_mask = fragment.fragment_type_mask;
+            let candidates = [
+                FragmentTypeFlag::CrossDbSnapshotBackfillStreamScan,
+                FragmentTypeFlag::SnapshotBackfillStreamScan,
+                FragmentTypeFlag::StreamScan,
```
wenym1, line 709: `StreamScan` should have included all `CrossDbSnapshotBackfillStreamScan` and `SnapshotBackfillStreamScan`.

src/meta/src/stream/stream_graph/fragment.rs
```diff
@@ -682,6 +696,74 @@ impl StreamFragmentGraph {
...
+                FragmentTypeFlag::SourceScan,
```
wenym1, line 710: Though `SourceScan` is considered here, we haven't supported backfill order control in source backfill executor yet. Is this expected?

wenym1: If not, shall we return with error if the user create mv ddl specifies backfill order control on any source?

src/stream/src/executor/mod.rs
```diff
@@ -316,6 +320,9 @@ pub enum Mutation {
         /// `subscriber` -> `upstream_mv_table_id`
         subscriptions_to_drop: Vec<(u32, TableId)>,
     },
+    StartFragmentBackfill {
```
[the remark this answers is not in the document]
kwannoel: https://github.com/risingwavelabs/risingwave/pull/20967/commits/d447492cddf440dc76e52d295f46c8fb06cf665e

proto/stream_plan.proto
```diff
@@ -30,6 +30,8 @@ message AddMutation {
   // TODO: we may allow multiple mutations in a single barrier.
   bool pause = 4;
   repeated SubscriptionUpstreamInfo subscriptions_to_add = 5;
+  // nodes which should start backfill
+  common.Uint32Vector backfill_nodes_to_start = 6;
```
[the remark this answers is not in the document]
kwannoel: Fixed in https://github.com/risingwavelabs/risingwave/pull/20967/commits/5543490f33cd75859e05bb202c40a87c13556d39

src/meta/model/src/streaming_job.rs
```diff
@@ -30,6 +33,9 @@ pub struct Model {
     pub specific_resource_group: Option<String>,
 }
 
+#[derive(Clone, Debug, PartialEq, Eq, FromJsonQueryResult, Serialize, Deserialize, Default)]
+pub struct BackfillOrders(pub HashMap<u32, Vec<u32>>);
```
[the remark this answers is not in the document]
kwannoel: Fixed in https://github.com/risingwavelabs/risingwave/pull/20967/commits/603ccc6160601d83a272f0a2d5e1028be4612ce5

src/meta/src/stream/stream_graph/fragment.rs
kwannoel: Fixed https://github.com/risingwavelabs/risingwave/pull/20967/commits/d5b053f003f71ed81dd3e552c20afe97e5ae2b6b

src/meta/src/stream/stream_graph/fragment.rs
kwannoel: We currently ban it in binding phase. I have refined the error message subsequently and method names https://github.com/risingwavelabs/risingwave/pull/20967/commits/9810a40ed2806ec238f1913f8308e0c36dd50434

src/meta/src/barrier/progress.rs
kwannoel: https://github.com/risingwavelabs/risingwave/pull/20967/commits/2f8679b5652c3783d55d3ac19bf95d3e0fa890a1 removed recovery logic, so we no longer have this paramter

src/meta/src/barrier/progress.rs
kwannoel: https://github.com/risingwavelabs/risingwave/pull/20967/commits/2f8679b5652c3783d55d3ac19bf95d3e0fa890a1 removed recovery logic, so we no longer have this paramter

src/meta/src/barrier/checkpoint/creating_job/mod.rs
```diff
@@ -123,6 +125,8 @@ impl CreatingStreamingJobControl {
             // we assume that when handling snapshot backfill, the cluster must not be paused
             pause: false,
             subscriptions_to_add: Default::default(),
+            // TODO: support backfill-order-control for snapshot backfill
```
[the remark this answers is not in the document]
kwannoel: No, because backfill nodes are paused within the executor, and I didn't implement any pausing logic inside snapshot backfill executor.

src/meta/src/barrier/progress.rs
kwannoel: Fixed by https://github.com/risingwavelabs/risingwave/pull/20967/commits/305fab707d07e84e9e9f9ce5b51a90b528cf6a14

and https://github.com/risingwavelabs/risingwave/pull/20967/commits/a8f13b2806484eeffb6261855f50a3306e5e37ca

src/stream/src/executor/mod.rs
```diff
@@ -530,6 +538,23 @@ impl Barrier {
...
+                ..
+            }))
+            | Some(Mutation::AddAndUpdate(
+                AddMutation {
+                    backfill_nodes_to_start,
+                    ..
+                },
+                _,
+            )) => !backfill_nodes_to_start.contains(&backfill_fragment_id),
+            _ => true,
```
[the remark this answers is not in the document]
kwannoel: After https://github.com/risingwavelabs/risingwave/pull/20967/commits/a8f13b2806484eeffb6261855f50a3306e5e37ca, we won't pause a node in these scenarios (scale / recovery), since their AddMutations will have an empty `backfill_nodes_to_pause`.

src/meta/src/stream/stream_graph/fragment.rs
kwannoel: Also removed all related source scan code: https://github.com/risingwavelabs/risingwave/pull/20967/commits/7c15dd44bf3d2e520f65e73002fe39be2b2b2c26

kwannoel:
Received a weird error in PR backfill tests:
```
2025-04-27T12:43:32.681278927Z  WARN risingwave_storage::hummock::event_handler::hummock_event_handler: cannot acquire lock for all read version pending_count=1 total_count=3
2025-04-27T12:43:34.118065407Z  WARN risingwave_storage::hummock::event_handler::hummock_event_handler: cannot acquire lock for all read version pending_count=1 total_count=4
2025-04-27T12:43:36.013967633Z  WARN risingwave_storage::hummock::event_handler::hummock_event_handler: cannot acquire lock for all read version pending_count=1 total_count=4
2025-04-27T12:43:46.031088558Z  WARN risingwave_storage::hummock::event_handler::hummock_event_handler: cannot acquire lock for all read version pending_count=1 total_count=4
2025-04-27T12:43:49.031780243Z  WARN risingwave_storage::hummock::event_handler::hummock_event_handler: cannot acquire lock for all read version pending_count=1 total_count=4
2025-04-27T12:43:52.953474268Z  WARN risingwave_storage::hummock::event_handler::hummock_event_handler: cannot acquire lock for all read version pending_count=1 total_count=4
2025-04-27T12:44:15.109883272Z  WARN risingwave_storage::hummock::event_handler::hummock_event_handler: cannot acquire lock for all read version pending_count=1 total_count=4
2025-04-27T12:44:22.559643063Z  WARN risingwave_storage::hummock::event_handler::hummock_event_handler: cannot acquire lock for all read version pending_count=1 total_count=4
2025-04-27T12:44:23.38246524Z  WARN risingwave_storage::hummock::event_handler::hummock_event_handler: cannot acquire lock for all read version pending_count=1 total_count=4
2025-04-27T12:44:25.228013259Z  WARN risingwave_storage::hummock::event_handler::hummock_event_handler: cannot acquire lock for all read version pending_count=1 total_count=4
2025-04-27T12:44:34.250162546Z  WARN risingwave_storage::hummock::event_handler::hummock_event_handler: cannot acquire lock for all read version pending_count=1 total_count=4
2025-04-27T12:44:43.162539534Z  WARN risingwave_storage::hummock::event_handler::hummock_event_handler: cannot acquire lock for all read version pending_count=1 total_count=4
2025-04-27T12:44:47.13186442Z  WARN risingwave_storage::hummock::event_handler::hummock_event_handler: cannot acquire lock for all read version pending_count=1 total_count=4
2025-04-27T12:44:57.057529904Z  WARN risingwave_storage::hummock::event_handler::hummock_event_handler: cannot acquire lock for all read version pending_count=1 total_count=4
2025-04-27T12:44:58.159557197Z  WARN risingwave_storage::hummock::event_handler::hummock_event_handler: cannot acquire lock for all read version pending_count=1 total_count=4
2025-04-27T12:44:59.050858039Z  WARN risingwave_storage::hummock::event_handler::hummock_event_handler: cannot acquire lock for all read version pending_count=1 total_count=4
2025-04-27T12:45:04.227780695Z  WARN risingwave_storage::hummock::event_handler::hummock_event_handler: cannot acquire lock for all read version pending_count=1 total_count=4
2025-04-27T12:45:15.413387237Z  WARN risingwave_storage::hummock::event_handler::hummock_event_handler: cannot acquire lock for all read version pending_count=1 total_count=4
2025-04-27T12:45:24.362448364Z  WARN risingwave_storage::hummock::event_handler::hummock_event_handler: cannot acquire lock for all read version pending_count=1 total_count=4
2025-04-27T12:45:38.750638942Z  WARN risingwave_storage::hummock::event_handler::hummock_event_handler: cannot acquire lock for all read version pending_count=1 total_count=4
2025-04-27T12:45:51.383654739Z  WARN risingwave_storage::hummock::event_handler::hummock_event_handler: cannot acquire lock for all read version pending_count=1 total_count=4
```

src/stream/src/executor/mod.rs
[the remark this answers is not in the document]
wenym1: For other scenario, their mutation is not even AddMutation. That's why I said adding an assertion or warning log when this method is called, when the mutation is not `AddMutation` or `AddAndUpdate `.

wenym1 (approved):
Rest LGTM. Thanks for the PR!

src/stream/src/executor/mod.rs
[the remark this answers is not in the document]
kwannoel: https://github.com/risingwavelabs/risingwave/pull/20967/commits/05ad81e285fef782d4344a48008db6006a0372ed

Added

PR #18571, opened 2024-09-18
feat(cdc): support default column in auto schema mapping

StrikeW:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://gist.github.com/TennyZhuang/f00be7f16996ea48effb049aa7be4d66#file-rw_cla).

## What's changed and what's your intention?

support default column in auto schema mapping for mysql and postgres.
The implementation only support constant expression, if the default value cannot be parsed it will be **ignored**.

close https://github.com/risingwavelabs/risingwave/issues/18359

## Checklist

- [x] I have written necessary rustdoc comments
- [x] I have added necessary unit tests and integration tests
- [ ] I have added test labels as necessary. See [details](https://github.com/risingwavelabs/risingwave/blob/main/docs/developer-guide.md#ci-labels-guide).
- [ ] I have added fuzzing tests or opened an issue to track them. (Optional, recommended for new SQL features #7934).
- [ ] My PR contains breaking changes. (If it deprecates some features, please create a tracking issue to remove them in the future).
- [x] All checks passed in `./risedev check` (or alias, `./risedev c`)
- [ ] My PR changes performance-critical code. (Please run macro/micro-benchmarks and show the results.)

- [ ] My PR contains critical fixes that are necessary to be merged into the latest release. (Please check out the [details](https://github.com/risingwavelabs/risingwave/blob/main/CONTRIBUTING.md))

## Documentation

- [ ] My PR needs documentation updates. (Please use the **Release note** section below to summarize the impact on users)

## Release note

If this PR includes changes that directly affect users or other significant modifications relevant to the community, kindly draft a release note to provide a concise summary of these changes. Please prioritize highlighting the impact these changes will have on users.

</details>

PR #21024, opened 2025-03-20
refactor(cdc): remove support for external default values

BugenZhao:

Follow-up of https://github.com/risingwavelabs/risingwave/pull/20671. See motivations there.

In that PR, we only avoided fetching metadata when users manually specify the schema. But for auto-derived schema, we still call `ExternalTableImpl::connect` to fetch the metadata, where we attempt to resolve upstream default values.

This effectively reverted #18571, #18322.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/common/src/catalog/column.rs
+++ b/src/common/src/catalog/column.rs
@@ -15,12 +15,10 @@
 use std::borrow::Cow;
 
 use itertools::Itertools;
-use risingwave_common::types::Datum;
 use risingwave_pb::expr::ExprNode;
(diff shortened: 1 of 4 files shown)


xxchan (approved):
The test cases should be kept to demonstrate the behavior. Others LGTM

BugenZhao:
I just realized that, even that Debezium will fill the default value for us, is it only applicable to incremental data? After removing the support, a new column with non-NULL default value will have NULL value for existing data in RW, which is inconsistent with the upstream database. :hot_face:

PR #16816, opened 2024-05-19
feat(sink): decouple starrocks commit from risingwave commit

ly9chee:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://gist.github.com/TennyZhuang/f00be7f16996ea48effb049aa7be4d66#file-rw_cla).

## What's changed and what's your intention?

The current StarRocks sink implementation using `StreamLoad` API writes data to StarRocks. It works well when the upstream does not change frequently.  However, if the upstream changes frequently, many versions of the table being written to are generated, which will impact the query performance and bring significant pressure to the `BE Compactor`.

### Implementation details

This PR attempts to solve this problem by leveraging the Starrocks [Stream Load Transaction Interface](https://docs.starrocks.io/docs/loading/Stream_Load_transaction_interface/). The idea is inspired by RFC [Support sink coordinator](https://github.com/risingwavelabs/rfcs/pull/61) and RFC [Decouple iceberg sink commit from risingwave checkpoint](https://github.com/risingwavelabs/rfcs/pull/78). Despite these two RFCs being designed for data lakes, but after doing some research, I found OLAP System like StarRocks can apply the same pattern as described in these RFCs as well, i.e., Changing the current `StreamLoad` to the `Stream Load Transaction Interface`.

> If you run a load job by using a program, the Stream Load transaction interface allows you to merge multiple mini-batches of data on demand and then send them all at once within one transaction by calling the /api/transaction/commit operation. As such, fewer data versions need to be loaded, and load performance is improved.

As the doc says, after a transaction is created, we can write multiple mini-batches to the transaction through `/api/transaction/load` interface. StarRocks will take care of merging these batches into a larger one when the `/api/transaction/commit` is called. So in `SinkWriter` implementation, we can write data to the transaction at every barrier, but only commit the transaction when a checkpoint barrier is received, which is delayed by `commit_checkpoint_interval`. In fact, the `commit` is requested after the coordinator collects the metadata from each `SinkWriter`.

Assume the barrier interval is 1 second and the checkpoint frequency is 1. The current sink implementation will commit data to StarRocks at every barrier, thus producing a version every second. But this PR will commit data to StarRocks only when the user-specified `commit_checkpoint_interval` is reached. If we set `commit_checkpoint_interval` to 10, a version is created roughly every 10 seconds.

#### Minor refactors
1. URL encoding is applied to the username and password during the construction of `mysql_conn_uri` to allow special characters to be included.
[...]

PR #21118, opened 2025-03-26
refactor(sink): remove sink coordination from starrocks sink

ly9chee:

Previously, in #16816, we decoupled the StarRocks sink commit from the RisingWave commit. However, sink coordination was also introduced in that PR, which seems unnecessary for the at-least-once semantics that we currently support. Unlike data lakes, StarRocks doesn't support loading a table concurrently within a single transaction; instead, it supports multiple transactions against one table. According to [flink connector](https://github.com/StarRocks/starrocks-connector-for-apache-flink/blob/fcc130ca1eb640bc1b28cefdd0ee3b420bd72d8a/starrocks-stream-load-sdk/src/main/java/com/starrocks/data/load/stream/v2/TransactionTableRegion.java#L411) source code, they also implement this in "one transaction one writer" style rather than "one transaction multiple writers" style.

This PR removes sink coordination-related code, eliminating the overhead between meta and sink writers at every checkpoint.

## Checklist

## Documentation

<details>

</details>


ly9chee:
@xxhZs @wenym1 PTAL

src/connector/src/sink/starrocks.rs
```diff
@@ -257,8 +251,8 @@ impl StarrocksSink {
 }
 
 impl Sink for StarrocksSink {
-    type Coordinator = StarrocksSinkCommitter;
-    type LogSinker = DecoupleCheckpointLogSinkerOf<CoordinatedSinkWriter<StarrocksSinkWriter>>;
+    type Coordinator = DummySinkCommitCoordinator;
```
wenym1, line 254: We don't have to explicitly specify the coordinator type here, since it's the default associated type.

src/connector/src/sink/starrocks.rs
```diff
@@ -583,42 +549,41 @@ impl SinkWriter for StarrocksSinkWriter {
         }
     }
 
-    async fn barrier(&mut self, is_checkpoint: bool) -> Result<Option<SinkMetadata>> {
-        if self.client.is_some() {
+    async fn barrier(&mut self, is_checkpoint: bool) -> Result<()> {
+        if let Some(client) = self.client.take() {
```
wenym1, line 553: Not relate to this PR. Previously, should we only take and finish the client only when `is_checkpoint` is true?

src/connector/src/sink/starrocks.rs
```diff
@@ -583,42 +549,41 @@ impl SinkWriter for StarrocksSinkWriter {
...
-                Ok(Some(StarrocksWriteResult(Some(txn_label)).try_into()?))
-            } else {
-                // no data was written within previous epoch
-                Ok(Some(StarrocksWriteResult(None).try_into()?))
+        if is_checkpoint && let Some(txn_label) = self.curr_txn_label.take() {
+            if let Err(err) = self.prepare_and_commit(txn_label.clone()).await {
+                self.txn_client
+                    .rollback(txn_label.clone())
+                    .await
+                    .map_err(|err| {
```
wenym1, line 567: On rollback error, we may only add a warning log, and otherwise we will lose the information in the error of `prepare_and_commit`.

src/connector/src/sink/starrocks.rs
ly9chee: here client is only related to a single transaction load, we should finish the transaction load at every barrier. Thus when checkpoint_interval > 1, there are multiple transaction loads will be made within a checkpoint. when `is_checkpoint == true` we commit those transaction loads to SR. The previous `is_some` check seems a little bit redundant

src/connector/src/sink/starrocks.rs
ly9chee: It seems that the associated type `Coordinator`  does not have a default type declared 🤔 Other non-coordinated sinks(kafka, mongodb, postgres, ...) also declared this `type Coordinator = DummySinkCommitCoordinator;` in their Sink impl

src/connector/src/sink/starrocks.rs
```diff
@@ -527,6 +479,23 @@ impl StarrocksSinkWriter {
             chrono::Utc::now().timestamp_micros()
         )
     }
+
+    async fn prepare_and_commit(&self, txn_label: String) -> Result<()> {
+        tracing::debug!(?txn_label, "prepare transaction");
+        let txn_label_res = self.txn_client.prepare(txn_label.clone()).await?;
+        assert_eq!(
```
wenym1, line 486: May change from assert to return error. Otherwise, the RW compute node may be crashed by potential bug in external SR server.

wenym1 (approved):
Rest LGTM.

Thanks for the contribution!

src/connector/src/sink/starrocks.rs
ly9chee: Make sense, the assertions are inherited from previous impl. I'll do some refactoring when I have time

ly9chee: fixed

fahadullahshah261:
Hi @ly9chee,

I am Fahad from RisingWave, and I am editing our monthly newsletter. We have recently introduced the Community Corner section that features contributors and community members. Not much space, but goes out to 8k subscribers. 
We really appreciate your contribution and would love to highlight it in that section! Could you share a few sentences about yourself and what motivated you to contribute? Would you be okay if we include the photo?
p.s. I am attaching how that looked in the last edition of the newsletter. 
Thank you again for your contribution! 

Regards, 
Fahad

Screenshot 2025-09-26 at 12 18 17

PR #21121, opened 2025-03-26
refactor(meta): extract common logic of collecting initial barrier in recovery

wenym1:

Part of https://github.com/risingwavelabs/risingwave/pull/20952

* extract the logic of collecting initial barriers during recovery to a DatabaseInitialBarrierCollector and provide some util methods
* simplify some fields for the structs that track backfill. For example, previously we store a StreamJobFragment as a field, but we actually only access its job_id, and then we change to only store a job_id
* include the `InflightStreamingJobInfo` of creating snapshot backfill job in the `inflight_info`, which is used to generate the initial partial graph infos when initially establishing a new control stream to CN.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/meta/src/barrier/checkpoint/control.rs
+++ b/src/meta/src/barrier/checkpoint/control.rs
@@ -1147,6 +1147,7 @@ impl DatabaseCheckpointControl {
                         .to_mutation(false)
                         .expect("should have some mutation in `CreateStreamingJob` command");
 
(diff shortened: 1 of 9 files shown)


wenym1:
Reviewed and approved in https://github.com/risingwavelabs/risingwave/pull/20952.

Close it for duplication.

PR #3745, opened 2022-07-08
feat: deprecate catalog ref id

yezizp2012:
I hereby agree to the terms of the [Singularity Data, Inc. Contributor License Agreement](https://gist.github.com/skyzh/0663682a70b0edde7ae991492f2314cb#file-s9y_cla).

## What's changed and what's your intention?

As title, deprecate database/schema/table_ref_id.

Please explain **IN DETAIL** what the changes are in this PR and why they are needed:

- Summarize your change (**mandatory**)
- How does this PR work? Need a brief introduction for the changed logic (optional)
- Describe clearly one logical change and avoid lazy messages (optional)
- Describe any limitations of the current code (optional)

## Checklist

- [x] I have written necessary rustdoc comments
- [x] I have added necessary unit tests and integration tests
- [x] All checks passed in `./risedev check` (or alias, `./risedev c`)

## Documentation

If your pull request contains user-facing changes, please specify the types of the changes, and create a release note. Otherwise, please feel free to remove this section.

### Types of user-facing changes

Please keep the types that apply to your changes, and remove those that do not apply.

* Installation and deployment 
* Connector (sources & sinks)
* SQL commands, functions, and operators
* RisingWave cluster configuration changes
* Other (please specify in the release note below)

### Release note

Please create a release note for your changes. In the release note, focus on the impact on users, and mention the environment or conditions where the impact may occur.

PR #21133, opened 2025-03-27
refactor(proto): cleanup unused fields to reduce confusion

xiangjinwu:

Highest Release Version: 2.3

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/prost/src/lib.rs
+++ b/src/prost/src/lib.rs
@@ -511,15 +511,10 @@ impl std::fmt::Debug for meta::SystemParams {
 impl std::fmt::Debug for data::DataType {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         let data::DataType {
-            precision,
-            scale,
-            interval_type,
             field_type,
             field_names,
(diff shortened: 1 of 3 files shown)


src/prost/src/lib.rs
```diff
@@ -688,12 +677,12 @@ mod tests {
...
-        let data_type: DataType = DataType::default();
-        let new_data_type = DataType {
-            is_nullable: data_type.get_is_nullable(),
+        let chunk = DataChunk::default();
+        let new_chunk = DataChunk {
+            cardinality: chunk.get_cardinality(),
             ..Default::default()
         };
-        assert!(!new_data_type.is_nullable);
+        assert_eq!(new_chunk.cardinality, 0);
```
xiangjinwu, line 685: Historical context on what this test is really about: #3745

xiangjinwu:
Not sure if `is_nullable` should be removed.

It is currently unused and always hardcoded to `true`.

But once removed it cannot be added back - we can only add `is_not_null` because missing fields will default to `false`.

PR #21149, opened 2025-03-28
feat(ctl): add API to detect bottleneck actors

KeXiangWang:

Highest Release Version: 2.3

Add an API in risectl to using await tree to locate the bottleneck in streaming graphs. Currently, the API will first pull for a new await tree dump and analyze on it. Ideally, it should also be able to read from an existing diagnose file or await tree file. Will add this later.

## Checklist

## Documentation

<details>

</details>


kwannoel (commented):
It will be better to decouple the API for fetching the await-tree json, and the API to detect bottleneck actors.

That's because our bottleneck-detecting strategy may not be perfect. So we may wish to fetch the await-tree json from the user cluster first.

Then we can iterate on the bottleneck detecting algorithm, and tweak the tool to suit the await-tree.

Besides that, great work!

src/meta/src/dashboard/mod.rs
```diff
@@ -351,7 +351,10 @@ pub(super) mod handlers {
 
         for worker_node in worker_nodes {
             let client = compute_clients.get(worker_node).await.map_err(err)?;
-            let result = client.stack_trace().await.map_err(err)?;
+            let result = client
+                .stack_trace(StackTraceRequest::default())
+                .await
+                .map_err(err)?;
```
xxchan, line 357: We may also allow specifying format here

src/meta/src/manager/diagnose.rs
```diff
@@ -626,7 +626,7 @@ impl DiagnoseCommand {
         let compute_clients = ComputeClientPool::adhoc();
         for worker_node in &worker_nodes {
             if let Ok(client) = compute_clients.get(worker_node).await
-                && let Ok(result) = client.stack_trace().await
+                && let Ok(result) = client.stack_trace(StackTraceRequest::default()).await
```
xxchan, line 629: I'm thinking maybe json can be better for diagnose dump 🤔

Reason: we have periodical diagnose dump on cloud. So we could use that json to analyze more easily.

But the diagnose report is also human readable, so I'm a little hesitant.

xxchan:
Actually I prefer a js/web based tool to analyze the bottleneck. 🤔 That would be faster to iterate and demonstrate. But it's also fine to have one in risectl.

src/ctl/src/lib.rs
```diff
@@ -72,6 +72,9 @@ enum Commands {
     /// Dump the await-tree of compute nodes and compactors
     #[clap(visible_alias("trace"))]
     AwaitTree,
+    /// Detect the bottleneck actors from the await-tree of compute nodes
+    #[clap(visible_alias("trace-analyze"))]
+    AwaitTreeAnalyze,
```
xxchan, line 77: Prefer to make this a subcommand of `AwaitTree`

kwannoel:
Tested with this SQL:
```
create table t(id int, v1 int);

create table t2(id int, v1 int);

insert into t select id, 1 from generate_series(1, 10000) t(id);

insert into t2 select id, 1 from generate_series(1, 10000) t(id);

flush;

create materialized view mv as select t.id, count(t.v1) from t join t2 on t.v1 = t2.v1 group by t.id;
```

Works well for me:
```
>> Actor 129
Actor 129: `mv` [21.285s]  <== current
  Epoch 8251479171792896 [!!! 21.283s]
    Materialize 8100000007 [!!! 21.283s]
      Project 8100000006 [!!! 21.273s]
        HashAgg 8100000005 [!!! 21.273s]
          Merge 8100000004 [0.001s]
[Detached 3]
  LocalInput (actor 134) [0.001s]

>> Actor 130
Actor 130: `mv` [21.285s]  <== current
  Epoch 8251479171792896 [!!! 21.284s]
    Materialize 8200000007 [!!! 21.284s]
      Project 8200000006 [!!! 21.273s]
        HashAgg 8200000005 [!!! 21.273s]
          Merge 8200000004 [0.009s]
[Detached 3]
  LocalInput (actor 134) [0.009s]

>> Actor 131
Actor 131: `mv` [21.285s]  <== current
  Epoch 8251479171792896 [!!! 21.283s]
    Materialize 8300000007 [!!! 21.283s]
      Project 8300000006 [!!! 21.279s]
        HashAgg 8300000005 [!!! 21.279s]
          Merge 8300000004 [0.007s]
[Detached 3]
  LocalInput (actor 134) [0.007s]

>> Actor 132
Actor 132: `mv` [21.285s]
  Epoch 8251479171792896 [!!! 21.283s]
    Materialize 8400000007 [!!! 21.283s]
      Project 8400000006 [!!! 21.280s]
        HashAgg 8400000005 [!!! 21.280s]
          Merge 8400000004 [0.000s]  <== current

```

kwannoel:
>  Ideally, it should also be able to read from an existing diagnose file or await tree file.

+1 for this, it takes a long time for users to upgrade.

src/meta/src/manager/diagnose.rs
KeXiangWang: I'll add an API to convert JSON to human-readable text. This is mainly to assist with manual review in cases where bottleneck detect produces inaccurate analysis.

src/meta/src/dashboard/mod.rs
KeXiangWang: > It will be better to decouple the API for fetching the await-tree json, and the API to detect bottleneck actors.
That's because our bottleneck-detecting strategy may not be perfect. So we may wish to fetch the await-tree json from the user cluster first.

@kwannoel read my mind. I choose to write this in risectl is for decoupling dumping and analyzing. But of course we can allow one to specify format through meta's API.

await-tree.txt
xxchan, line 1: delete this

src/compute/src/rpc/service/monitor_service.rs
```diff
@@ -77,12 +77,23 @@ impl MonitorService for MonitorServiceImpl {
...
+                .map(|(k, v)| {
+                    (
+                        k.0,
+                        if req.actor_traces_format.is_some()
+                            && req.actor_traces_format.as_ref().unwrap() == "text"
+                        {
+                            v.to_string()
+                        } else {
+                            serde_json::to_string(&v).unwrap()
+                        },
```
xxchan, line 94: The logic is different from the comment: "// Allowed values: "json", "text". Default is "text"."

Here it defaults to `json`.

Maybe better to define an enum instead of using string

src/ctl/src/lib.rs
```diff
@@ -847,7 +872,15 @@ async fn start_impl(opts: CliOpts, context: &CtlContext) -> Result<()> {
...
         }
-        Commands::AwaitTree => cmd_impl::await_tree::dump(context).await?,
+        Commands::AwaitTree(AwaitTreeCommands::Dump {
+            actor_traces_format,
+        }) => cmd_impl::await_tree::dump(context, actor_traces_format).await?,
+        Commands::AwaitTree(AwaitTreeCommands::Analyze { path }) => {
+            cmd_impl::await_tree::bottleneck_detect(context, path).await?
+        }
+        Commands::AwaitTree(AwaitTreeCommands::Transcribe { path }) => {
+            cmd_impl::await_tree::transcribe(path)?
```
xxchan, line 882: still slightly prefer to put these logic in a separate repo. (It's ok to let risectl include them, but they can be developed separately and iterate faster. Also no need to compile risectl which means compiling whole risingwave to develop)

src/ctl/src/cmd_impl/await_tree/analyze.rs
```diff
@@ -0,0 +1,185 @@
...
+
+    if !bottleneck_actors_found {
+        println!("No bottleneck actors detected.");
+    }
+
+    Ok(())
+}
+
+fn bottleneck_detect_from_file(path: String) -> anyhow::Result<()> {
+    let actor_traces = extract_actor_traces(&path)
```
xxchan, line 161: hmmm, is `extract_actor_traces` for TEXT only, and should not be applied to JSON?

xxchan (commented):
Thanks for the work! Mainly just nit picking.

What about we first get the changes to get JSON format merged (maybe can also cherry-picked). And refine the analyzer part later?

src/ctl/src/cmd_impl/await_tree/utils.rs
```diff
@@ -0,0 +1,253 @@
...
+
+    Ok(actor_traces)
+}
+
+/// The process of converting the tree to text is not lossless—information such as
+/// `node_id` will be lost. Consequently, this function can only restore information
+/// to the best extent possible. Fields like `current` and `node_id` cannot be
+/// recovered, but this loss does not affect our bottleneck detection. In the
+/// function, we will set all `node_id` values to 0 and `current` to 100.
+pub fn parse_tree_view_from_text(input: &str) -> anyhow::Result<TreeView> {
```
BugenZhao, line 84: TBH I didn't expect us to write a function to work atop of the human-readable format. But as it's proven feasible, I'm okay with it and we can keep it until json format is supported in all active RisingWave releases.

Agree on @xxchan that it would be better to maintain this in a separate repo (and thus a separate tool) for faster iteration.

src/ctl/src/cmd_impl/await_tree/transcribe.rs
```diff
@@ -0,0 +1,28 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::cmd_impl::await_tree::tree::TreeView;
+use crate::cmd_impl::await_tree::utils::extract_actor_traces;
+
+pub fn transcribe(path: String) -> anyhow::Result<()> {
```
BugenZhao, line 18: I'm thinking that this functionality can be provided by the library itself.

src/ctl/src/cmd_impl/await_tree/analyze.rs
```diff
@@ -0,0 +1,185 @@
...
+    ///     Union 9E20200000007 [1.000s]
+    ///       Merge 9E20200000003 [1.000s]
+    ///         LocalInput (actor 647689) [1.000s]
+    /// ```
+    /// Note:
+    /// In this case, although the bottleneck actor(BN Tree) throttles the whole graph,
+    /// the bottleneck actor is still yielding output to downstream actors. A typical
+    /// case is JOIN amplification. So the corresponding actors are actively processing
+    /// the data but the EPOCH span is blocked.
+    pub fn is_bottleneck(&self) -> bool {
```
BugenZhao, line 87: Hmmm, the algorithm looks much easier than I thought initially. So there's no need to traverse between actors (based on the `Input` and `Output` spans) at all?

src/ctl/src/cmd_impl/await_tree/utils.rs
KeXiangWang: > TBH I didn't expect us to write a function to work atop of the human-readable format

Me neither. But AI can do it. 🤣

src/ctl/src/cmd_impl/await_tree/analyze.rs
KeXiangWang: `extract_actor_traces` is for both JSON and TEXT. For either JSON or TEXT, the actor traces are encoded as 
```
--- Actor Traces ---
>> Actor {actor_id}
{trace}

...
```
So we can use one unified funtion to extract  `extract_actor_traces` from the diagnose file.

src/ctl/src/cmd_impl/await_tree/analyze.rs
KeXiangWang: To be honest, I'm not entirely sure this algorithm works in every situation. However, in most cases where there's only one bottleneck actor, it should perform as expected. I can think of a few edge cases:

1. If the bottleneck's span doesn't call `instrument_await`, the bottleneck's tree might not exhibit the typical "slow parent with fast children" signal.
2. If an MV or the entire graph has multiple fragments acting as bottlenecks, I'm uncertain what the outcome would be.

At present, this tool doesn't completely eliminate manual effort. It definitely requires ongoing optimization and the addition of new rules.

src/ctl/src/cmd_impl/await_tree/transcribe.rs
KeXiangWang: I'm OK with that.

src/ctl/src/cmd_impl/await_tree/dump.rs
```diff
@@ -0,0 +1,65 @@
...
+        let response = client.stack_trace(req.clone()).await?;
+        all.merge_other(response);
+    }
+
+    if all.actor_traces.is_empty()
+        && all.rpc_traces.is_empty()
+        && all.compaction_task_traces.is_empty()
+        && all.inflight_barrier_traces.is_empty()
+    {
+        println!("No traces found. No actors are running, or `--async-stack-trace` not set?");
```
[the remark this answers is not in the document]
KeXiangWang: You mean the all the traces? or the output like `No traces found. No actors are running...`?

src/ctl/src/lib.rs
KeXiangWang: updated

src/ctl/src/cmd_impl/await_tree/dump.rs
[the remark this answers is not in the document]
BugenZhao: Maybe print to `stderr` instead?

src/ctl/src/cmd_impl/await_tree/analyze.rs
BugenZhao: > 2\. I'm uncertain what the outcome would be.

Shall we output all of them, ordering by some sort of confidence or probablity?

KeXiangWang: Tend to leave it for future. At present, we don't have too much experience dealing with these edge cases. We can add some new rules when we can summarize some tips for them. For now, if the current `bottleneck_detect` return nothing, the on-caller should just use `transcribe` to convert the json await tree to text and mannually analyse the text await tree.

src/ctl/src/cmd_impl/await_tree/dump.rs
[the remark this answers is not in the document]
KeXiangWang: done

src/ctl/src/lib.rs
KeXiangWang: Done. https://github.com/risingwavelabs/rw-diagnose-tools

src/meta/src/manager/diagnose.rs
KeXiangWang: >I'm thinking maybe json can be better for diagnose dump 

Update: now it's json.

src/ctl/Cargo.toml
```diff
@@ -31,6 +31,7 @@ risingwave_pb = { workspace = true }
 risingwave_rpc_client = { workspace = true }
 risingwave_storage = { workspace = true }
 risingwave_stream = { workspace = true }
+rw-diagnose-tools = { git = "https://github.com/risingwavelabs/rw-diagnose-tools" }
```
KeXiangWang, line 34: Use git url for now. Need @BugenZhao 's help to publish the repo in crates.io.

await-tree.txt
xxchan, line 1: delete this file

src/compute/src/rpc/service/monitor_service.rs
xxchan: this is not resolved

src/meta/src/dashboard/mod.rs
```diff
@@ -351,7 +351,12 @@ pub(super) mod handlers {
 
         for worker_node in worker_nodes {
             let client = compute_clients.get(worker_node).await.map_err(err)?;
-            let result = client.stack_trace().await.map_err(err)?;
+            let result = client
+                .stack_trace(StackTraceRequest {
+                    actor_traces_format: Some("text".to_owned()),
```
xxchan, line 356: Please add a parameter to the meta dashboard `/await_tree` call instead. Can default to `text`

src/meta/src/manager/diagnose.rs
```diff
@@ -626,7 +626,11 @@ impl DiagnoseCommand {
         let compute_clients = ComputeClientPool::adhoc();
         for worker_node in &worker_nodes {
             if let Ok(client) = compute_clients.get(worker_node).await
-                && let Ok(result) = client.stack_trace().await
+                && let Ok(result) = client
+                    .stack_trace(StackTraceRequest {
+                        actor_traces_format: Some("json".to_owned()),
```
xxchan, line 631: ditto, please add a parameter to the `/diagnose` API instead.

xxchan (commented):
Others LGTM. Thanks for the work

kwannoel (approved):
After address @xxchan comments, rest LGTM.

await-tree.txt
KeXiangWang: done

src/compute/src/rpc/service/monitor_service.rs
KeXiangWang: Updated.

src/meta/src/dashboard/mod.rs
KeXiangWang: Added. Example of a request:
```
curl "http://localhost:5691/api/monitor/await_tree/?format=text" > diagnose.txt
```

src/meta/src/manager/diagnose.rs
KeXiangWang: Example:
```
curl http://localhost:5691/api/monitor/diagnose/\?actor_traces_format\=json > diagnose.txt
```

xxchan (approved):
Generally LGTM, Thanks!  Should we cherry-pick to earlier releases?

BugenZhao (approved):
LGTM

PR #21166, opened 2025-03-28
refactor(frontend,stream): store metrics at an executor level rather than operator level

kwannoel:

Highest Release Version: 2.3

This is done to avoid GC and it avoids synchronization between actors due to them updating the same metric.
It allows us to track metrics at a more granular level as well: We can look at min, max, avg for each metric, and identify the slowest executor for a given operator.

The trade-off is that more memory is utilized per executor. But the memory used per executor is very little. Each metric occupies 8 bytes of memory, with the `Arc` pointer, that's another 8 bytes. Considering the entry in the hash map and the value stored, we can assert another 16bytes. So in total 8x4=32 bytes approximately.
If we have 1000 operators, with parallelism 64, that's 1000 * 64 * 32 = 2048KB ~ 2MB, which still seems acceptable.

When dropping actors, we also need to lock their corresponding metric map, and drop the entry. So there will be some overhead. But I think this should be minimal.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/common/metrics/src/monitor/in_mem.rs
+++ b/src/common/metrics/src/monitor/in_mem.rs
@@ -16,74 +16,111 @@
 //! It is intentionally decoupled from Prometheus.
 
 use std::collections::HashMap;
-use std::sync::Arc;
 use std::sync::atomic::AtomicU64;
+use std::sync::{Arc, Weak};
 
-use parking_lot::RwLock;
-
-use crate::MetricLevel;
+use parking_lot::Mutex;
 
 pub type Count = Arc<AtomicU64>;
 
+pub struct GuardedCount {
+    id: u64,
+    pub count: Count,
+    parent: Weak<Mutex<InnerCountMap>>,
+}
+
+impl GuardedCount {
+    pub fn new(id: u64, parent: &Arc<Mutex<InnerCountMap>>) -> (Count, Self) {
+        let guard = GuardedCount {
+            id,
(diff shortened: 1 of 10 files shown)


kwannoel:
* **#21166** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/21166?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/21166?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/common/metrics/src/monitor/in_mem.rs
```diff
@@ -16,74 +16,105 @@
...
+}
+
+impl Drop for GuardedCount {
+    fn drop(&mut self) {
+        if let Some(parent) = self.parent.upgrade() {
+            let mut map = parent.lock();
+            map.inner.remove(&self.id);
+        } else {
+            tracing::warn!("parent dropped unexpectedly");
+        }
```
BugenZhao, line 50: I think there's no need to warn about this, as there's no resource leakage in any case.

kwannoel: Fixed

BugenZhao (approved):
LGTM

PR #21169, opened 2025-03-28
feat(streaming): asof join without cache

yuhao-su:

Highest Release Version: 2.3

Rewrite the Asof join. Now the we do not use the operator cache to reduce the unnecessary fetch of data.

## Checklist

## Documentation

<details>

</details>


kwannoel (commented):
Could you run some benchmarks for before vs after:
- Performance when eq join has high selectivity but ineq cond has low selectivity
- Performance when eq join has low selectivity and ineq cond has low selectivity

src/stream/src/executor/join/hash_join.rs
```diff
@@ -1196,6 +978,182 @@ where {
     }
 }
 
+pub struct AsOfJoinHashMap<S: StateStore> {
```
kwannoel, line 981: We should decouple this. We can have a module for `as_of_join`, independent of `hash_join`.

src/stream/src/executor/asof_join.rs
```diff
@@ -897,37 +788,20 @@ impl<K: HashKey, S: StateStore, const T: AsOfJoinTypePrimitive> AsOfJoinExecutor
...
-                        side_update.ht.delete_row(key, row)?;
+                        side_update.ht.delete(row)?;
                     }
                 }
             } else {
                 // Row which violates null-safe bitmap will never be matched so we need not
                 // store.
                 // Noop here because we only support left outer AsOf join.
             }
-            join_matched_rows_metrics.observe(join_matched_rows_cnt as _);
```
kwannoel, line 917: Why don't we monitor high join amp anymore?

src/stream/src/executor/join/hash_join.rs
```diff
@@ -1196,6 +978,182 @@ where {
...
+        let row = pinned_table_iter.next().await.transpose()?;
+        Ok(row)
+    }
+
+    pub async fn table_iter_by_inequality_with_jk_prefix<'a>(
+        &'a self,
+        join_key: &'a impl Row,
+        range: &'a (Bound<impl Row>, Bound<impl Row>),
+    ) -> StreamExecutorResult<impl RowStream<'a>> {
+        // let pk_prefix = join_key.chain(inequality_key);
```
kwannoel, line 1093:
```suggestion
```

src/stream/src/executor/asof_join.rs
```diff
@@ -183,18 +148,14 @@ impl<K: HashKey, S: StateStore, const T: AsOfJoinTypePrimitive> AsOfJoinExecutor
         input_r: Executor,
         params_l: JoinParams,
         params_r: JoinParams,
-        null_safe: Vec<bool>,
```
kwannoel, line 186: I forgot the purpose of `null_safe`. Why is it removed, should we also remove it for hash join?

kwannoel (commented):
Generally LGTM. I think we need to add some tests also to make sure our new matching logic is sound.

Issue #21174, opened 2025-03-30
failed test: iceberg_engine.slt stuck


xxchan:
https://buildkite.com/risingwavelabs/main-cron/builds/4703#0195e2a2-0789-4340-ad3b-ef21223137fe
<img width="1088" alt="Image" src="https://github.com/user-attachments/assets/a1a62463-44da-45a4-87de-1c0b780d17e3" />

logs: (notice timestamp)

FE log:
```
2025-03-29T16:21:36.959471308Z  INFO handle_query: risingwave_jni_core::jvm_runtime: initialize JVM successfully mode="simple query" session_id=10 sql=CREATE TABLE t (id INT PRIMARY KEY, name CHARACTER VARYING) WITH (commit_checkpoint_interval = 1) ENGINE = ICEBERG
2025-03-29T16:21:37.018897058Z  INFO handle_query: risingwave_jni_core::jvm_runtime: register native methods for jvm successfully mode="simple query" session_id=10 sql=CREATE TABLE t (id INT PRIMARY KEY, name CHARACTER VARYING) WITH (commit_checkpoint_interval = 1) ENGINE = ICEBERG
2025-03-29T16:21:37.105616296Z  INFO handle_query: risingwave_telemetry_event: Telemetry tracking_id is not set, event reporting disabled mode="simple query" session_id=10 sql=CREATE TABLE t (id INT PRIMARY KEY, name CHARACTER VARYING) WITH (commit_checkpoint_interval = 1) ENGINE = ICEBERG
2025-03-29T16:21:38.27263672Z  INFO handle_query: risingwave_telemetry_event: Telemetry tracking_id is not set, event reporting disabled mode="simple query" session_id=10 sql=CREATE TABLE t (id INT PRIMARY KEY, name CHARACTER VARYING) WITH (commit_checkpoint_interval = 1) ENGINE = ICEBERG
2025-03-29T16:21:38.30645836Z  INFO handle_query: risingwave_telemetry_event: Telemetry tracking_id is not set, event reporting disabled mode="simple query" session_id=10 sql=CREATE TABLE t (id INT PRIMARY KEY, name CHARACTER VARYING) WITH (commit_checkpoint_interval = 1) ENGINE = ICEBERG
2025-03-29T16:22:38.369368672Z  INFO handle_query: pgwire_slow_query_log: slow query elapsed=60000ms mode="simple query" session_id=10 sql=FLUSH
[...]

PR #21179, opened 2025-03-31
chore: increase timeout for end-to-end iceberg test (release)

wcy-fdu:

Highest Release Version: 2.3

Since exactly once iceberg sink is enable, there will be some performance loss when commit:
1. There will be an additional process of writing system tables when committing
2. When the commit epoch > the checkpoint completion epoch, the commit will wait until the `committed_epoch` increases before committing, this is for consistency.

Thus, this pr set a larger timeout for the previous test.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/ci/workflows/main-cron.yml
+++ b/ci/workflows/main-cron.yml
@@ -430,7 +430,7 @@ steps:
           <<: *docker-compose
           run: iceberg-test-env
       - ./ci/plugins/upload-failure-logs
-    timeout_in_minutes: 12
+    timeout_in_minutes: 20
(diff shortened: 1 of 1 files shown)


xxchan (commented):
Do you mean #21174? But it looks like a stuck, instead of a timeout 🤔

wcy-fdu:
> Do you mean #21174? But it looks like a stuck, instead of a timeout 🤔

yes #21174, just want to increase timeout to see whether it's a stuck. Seems committed_epoch is not increase.

PR #21200, opened 2025-04-01
fix(snapshot-backfill): reproduce snapshot backfill test timeout

wenym1:

Highest Release Version: 2.3

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/ci/scripts/run-backfill-tests.sh
+++ b/ci/scripts/run-backfill-tests.sh
@@ -424,32 +424,32 @@ test_cross_db_snapshot_backfill() {
 
 main() {
   set -euo pipefail
-  test_snapshot_and_upstream_read
-  test_backfill_tombstone
-  test_replication_with_column_pruning
-  test_sink_backfill_recovery
+#  test_snapshot_and_upstream_read
+#  test_backfill_tombstone
+#  test_replication_with_column_pruning
+#  test_sink_backfill_recovery
   test_snapshot_backfill
 
-  test_scale_in
+#  test_scale_in
 
-  test_cross_db_snapshot_backfill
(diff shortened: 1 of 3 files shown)


kwannoel (commented):
> However, when calling this method, the dispatcher changes on upstream actor is taken, but will not be used, since the mutation will not be applied to the upstream actors, which causes snapshot backfill actors fail to connect to upstream actors.

Why does it cause snapshot backfill actors to fail to connect to upstream actors

wenym1:
> > However, when calling this method, the dispatcher changes on upstream actor is taken, but will not be used, since the mutation will not be applied to the upstream actors, which causes snapshot backfill actors fail to connect to upstream actors.
> 
> Why does it cause snapshot backfill actors to fail to connect to upstream actors

In previous refactor, we change to compose the edges between actors in global barrier manager. The edges are stored in a separate result struct. When we compose a new add mutation, we will take the related edges out to fill the add mutation.

In snapshot backfill, we should include the dispatcher changes of upstream actors in the add mutation of barrier injected to upstream. Previously we use the same `Command::to_mutation` when we compose the initial add mutation of the newly creating job, and take the dispatchers changes of upstream actors by accident. This mutation is in a barrier only injected to downstream newly creating job, and will not be applied to upstream actors. When later the upstream jobs call the same `Command::to_mutation`, the dispatchers changes information has been taken, and upstream actors won't receive the expected dispatcher changes, and then it won't try to establish a dispatch connection to downstream actors of newly creating job, and the snapshot backfill actors will fail to connect to upstream actors (blocked at trying to receive the first upstream barrier)

kwannoel (approved):

PR #21201, opened 2025-04-01
feat: retry for requests in schema registry

tabVersion:
…t using exponential backoff

Highest Release Version: 2.3

as title

adding the following options for controlling retry when requiring schema, spec to confluent schema registry

* `schema.registry.max.delay.sec`
* `schema.registry.backoff.duration.ms`
* `schema.registry.backoff.factor`
* `schema.registry.retries.max`

It serves as internal fine-tune options so not plan to add to docs. 

## Checklist

## Documentation

<details>

</details>


src/connector/src/schema/schema_registry/client.rs
```diff
@@ -135,12 +138,25 @@ impl Client {
             client: self.inner.clone(),
             path: path.iter().map(|p| p.to_string()).collect_vec(),
         });
+
+        let retry_strategy = ExponentialBackoff::from_millis(100)
```
hzxa21, line 142: How about making the backoff duration, factor and attempts configurable via a developer config so that we can tune the config without rebuilding an image just in case?

src/connector/src/schema/schema_registry/client.rs
```diff
@@ -135,12 +138,25 @@ impl Client {
...
+            .factor(2)
+            .max_delay(Duration::from_secs(1))
+            .take(3);
```
hzxa21, line 145: Let's add a jitter here as well because we want to reduce the peak QPS to schema registry. Example: https://github.com/risingwavelabs/risingwave/blob/60ebe13ccf40cb69ced1849c2ca28ff3c393a909/src/stream/src/executor/source/mod.rs#L172

I am also thinking whether we should add a few ms delay with some randomness for the initial request before the retry kicks in to reduce the QPS for the 1st attempt.

tabVersion:
image

manually tested in local

hzxa21 (approved):

PR #21219, opened 2025-04-02
refactor: increase default time_travel_version_cache_capacity

zwang28:

Highest Release Version: 2.3

A small `time_travel_version_cache_capacity` can result in frequent evictions during batch queries that involve multiple lookup joins.
Since the time travel version only contains SstableInfo for specific table, the overall version size would be much smaller than the original. So it's safe to increase `time_travel_version_cache_capacity` a bit.

## Checklist

## Documentation

<details>

</details>


Li0k (approved):
LGTM, Are there any tests that show the difference in memory usage? However, I think the results are strongly correlated with the workload.

Issue #20946, opened 2025-03-18
System config to ban plain-text secrets


kwannoel:
`ALTER SYSTEM SET ENABLE_PLAIN_TEXT_SECRETS=FALSE;`.

By default it should be disabled, and users may enable it manually.

kwannoel:
any updaes?

tabVersion:
impl in https://github.com/risingwavelabs/risingwave/pull/21248

PR #21248, opened 2025-04-05
feat: enforce some props to be secret on RisingWave Cloud

tabVersion:

Highest Release Version: 2.3

as title, relate https://linear.app/risingwave-labs/issue/CLOUD-3897/%5Bdiscussion%5D-enforce-secret-for-sensitive-fields-on-cloud

Enforce sensitive fields to be SECRET when create SOURCE/SINK(in #21269)/CONNECTION

For cluster matching BOTH condition:

* hosted on cloud: by setting ENV `RISINGWAVE_CLOUD`
* system config: `enforce_secret_on_cloud` set to `true`

Then the kernel will do such enforcement. 

manually test loacally

```
set_env RISINGWAVE_CLOUD "1"

dev=> create table t1( a int, b varchar )  with ( connector = 'kafka', properties.bootstrap.server = 'localhost:9092',
  topic = 'test2', properties.sasl.password = 'demo') format plain encode json ;

  1: connector error
  2: Enforce Secret Error
  3: properties.sasl.password is enforced to be a SECRET on RisingWave Cloud, please use `CREATE SECRET` first
```

resolve https://github.com/risingwavelabs/risingwave/issues/20946

## Checklist

## Documentation

<details>

Introduce new system setting `enforce_secret_on_cloud` (default to `false` for open source, on-prem and byoc, default to `true` for cloud-hosted instance)

---

detect ENV `RISINGWAVE_CLOUD` as the signal for the instance is running on RisingWave Cloud

* enforces some props have to be SECRET when creating source/table. 

enforce SECRET props:

```
### AWS (mainly for loading schema from aws s3)
- access_key
- aws.credentials.access_key_id
- s3.access.key
- secret_key
- aws.credentials.secret_access_key
- s3.secret.key
- session_token
- aws.credentials.session_token

### Kafka
- properties.ssl.key.pem
- properties.ssl.key.password
- properties.sasl.password

### Pulsar
- pulsar.auth.token

### Kinesis
- kinesis.credentials.access
- kinesis.credentials.secret
- kinesis.credentials.session_token

### NATS
- password
- jwt
- nkey

### Iceberg
- s3.access.key
- s3.secret.key
- gcs.credential
- catalog.credential
- catalog.token
- catalog.oauth2_server_uri

### MQTT
- tls.client_cert
- tls.client_key
- password

### Filesystem (S3)
- s3.credentials.access
- s3.credentials.secret

### Filesystem (GCS)
- gcs.credential
- gcs.service_account

### Filesystem (Azblob)
- azblob.credentials.account_key
- azblob.credentials.account_name

### Google Pubsub
- pubsub.credentials

### Iceberg (Catalog)
- catalog.jdbc.password

### TestSource
- No specific secrets enforced (placeholder)

### DummyProperties
- No specific secrets enforced (placeholder)
```

comes from #21352

enforce secrets for new create CONNECTION

```
### Kafka: the same as Kafka source

### IcebergConnection: 
- s3.access.key
- s3.secret.key
- gcs.credential
- catalog.token

### ConfluentSchemaRegistryConnection
- schema.registry.password

### ElasticsearchConnection:
- elasticsearch.password
```

</details>


src/common/src/telemetry/mod.rs
```diff
@@ -39,8 +39,13 @@ pub const TELEMETRY_CLUSTER_TYPE_KUBERNETES: &str = "kubernetes";
 pub const TELEMETRY_CLUSTER_TYPE_SINGLE_NODE: &str = "single-node";
 pub const TELEMETRY_CLUSTER_TYPE_DOCKER_COMPOSE: &str = "docker-compose";
 const TELEMETRY_CLUSTER_TYPE_TEST: &str = "test";
+pub use risingwave_telemetry_event::{
+    TELEMETRY_RISINGWAVE_CLOUD_UUID, get_telemetry_risingwave_cloud_uuid,
+};
 
-pub use risingwave_telemetry_event::get_telemetry_risingwave_cloud_uuid;
+pub fn is_cloud_hosted() -> bool {
+    get_telemetry_risingwave_cloud_uuid().is_some()
```
BugenZhao, line 47: Wondering which is the most appropriate way to decide if we are running on Cloud. FYI, we also use `RISINGWAVE_CLOUD` here: https://github.com/risingwavelabs/risingwave/blob/4a1100e359469e401e044a542f64d40911938844/src/common/src/util/deployment.rs#L31-L42

cc @arkbriar

src/connector/src/enforce_secret_on_cloud.rs
```diff
@@ -0,0 +1,50 @@
...
+use risingwave_common::telemetry::is_cloud_hosted;
+
+use crate::error::ConnectorResult as Result;
+
+#[derive(Debug, thiserror::Error)]
+pub enum EnforceSecretOnCloudError {
+    #[error("{} is enforced to be a SECRET on RisingWave Cloud, please use `CREATE SECRET` first", .0)]
+    SecretProperty(String),
+}
+
```
BugenZhao, line 25:
```suggestion
#[derive(Debug, thiserror::Error)]
#[error("{key} is enforced to be a SECRET on RisingWave Cloud, please use `CREATE SECRET` first")]
pub struct EnforceSecretOnCloudError {
    key: String,
}

```

src/connector/src/error.rs
```diff
@@ -81,6 +82,7 @@ def_anyhow_newtype! {
 
     openssl::error::ErrorStack => "OpenSSL error",
     risingwave_common::secret::SecretError => "Secret error",
+    EnforceSecretOnCloudError => "Enforce Secret Error",
```
BugenZhao, line 85:
```suggestion
    EnforceSecretOnCloudError => transparent,
```

src/common/src/telemetry/mod.rs
arkbriar: I believe `RISINGWAVE_CLOUD` is the right way.

src/connector/Cargo.toml
```diff
@@ -85,6 +85,7 @@ parking_lot = { workspace = true }
 parquet = { workspace = true }
 paste = "1"
 pg_bigdecimal = { git = "https://github.com/risingwavelabs/rust-pg_bigdecimal", rev = "0b7893d88894ca082b4525f94f812da034486f7c" }
+phf = { version = "0.11", features = ["macros"] }
```
kwannoel, line 88: cool

kwannoel (commented):
Is there anyway to disable this on cloud, if some unforseen circumstance is encountered, and for already created source, will this affect backwards compatbility?

tabVersion:
> Is there anyway to disable this on cloud, if some unforseen circumstance is encountered, and for already created source, will this affect backwards compatbility?

I am considering following the issue suggested `alter system`. 
The check only happens in create object stage so it should not have backwards compatibility issue.

tabVersion:
also do we want to enforce secret when CREATE CONNECTION? If we just want to reuse some sensitive fields, we can directly ask for the privilege of CONNECTION, no bother enforcing anything. 
cc @lmatz @kwannoel

kwannoel:
> also do we want to enforce secret when CREATE CONNECTION? If we just want to reuse some sensitive fields, we can directly ask for the privilege of CONNECTION, no bother enforcing anything. cc @lmatz @kwannoel

Does this mean that in RWC, they must always create connection for source and sinks, and cannot fill in the fields inline?

tabVersion:
> > also do we want to enforce secret when CREATE CONNECTION? If we just want to reuse some sensitive fields, we can directly ask for the privilege of CONNECTION, no bother enforcing anything. cc @lmatz @kwannoel
> 
> Does this mean that in RWC, they must always create connection for source and sinks, and cannot fill in the fields inline?

I don't mean that. Using CONNECTION and filling inline are two alternatives.

lmatz:
I support enforcing secret also when creating CONNECTION. 
The motivation is to eliminate the possibility of leaking users' credentials as much as possible.
Because once the leak happens, it becomes a mess, and it is hard to explain.

kwannoel (approved):

xiangjinwu (approved):
For Cargo.lock changes

PR #21256, opened 2025-04-07
refactor(stream): unify remote and local output

wenym1:

Highest Release Version: 2.3

Currently, the logic of `RemoteOutput` and `LocalOutput` is basically the same, except that
- the instrument await tree message is different
- remote output calls `chunk.compact()` for `StreamChunk` before adding to `Sender`.

In this PR, we will unify the two `Output`s. The difference in instrument await can be unified without much concern. For calling `chunk.compact()`, we delay calling it in the handler of exchange service.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/executor/exchange/output.rs
+++ b/src/stream/src/executor/exchange/output.rs
@@ -12,12 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::fmt::Debug;
-
-use async_trait::async_trait;
 use await_tree::InstrumentAwait;
 use educe::Educe;
-use risingwave_common::util::addr::is_local_address;
 
 use super::error::ExchangeChannelClosed;
 use super::permit::Sender;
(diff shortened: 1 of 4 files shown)


wenym1:
* **#21257** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/21257?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#21256** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/21256?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/21256?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/stream/src/executor/exchange/output.rs
```diff
@@ -98,57 +34,38 @@ pub struct RemoteOutput {
     ch: Sender,
 }
 
-impl RemoteOutput {
+impl Output {
     pub fn new(actor_id: ActorId, ch: Sender) -> Self {
         Self {
             actor_id,
-            span: await_tree::span!("RemoteOutput (actor {:?})", actor_id).verbose(),
+            span: await_tree::span!("Output (actor {:?})", actor_id).verbose(),
```
BugenZhao, line 41: Shall we still preserve the information of `Local` or `Remote`? This can be helpful when debugging stuck caused by gRPC.

BugenZhao (commented):
Rest LGTM

src/stream/src/executor/exchange/output.rs
wenym1: Actually the purpose of this series of PRs is to avoid broadcasting and maintaining the global actor info, which is used in dispatch and merge executor. Previously in the dispatch executor side, we get the actor info only to distinguish between remote and local output. Now the difference between them is just the await tree span message, and it's not worth it to maintain the global actor info just to provide more detailed spawn message.

Besides, we provide the actor id in the span message, and we can actually infer whether it's remote or local output by checking if the downstream actor id is in the same CN as the current actor id. So we won't lose any debug information after this PR. It just make this information not straightforward, which I think it's acceptable.

BugenZhao (approved):

PR #21257, opened 2025-04-07
refactor(stream): store shared context in local barrier manager

wenym1:

Highest Release Version: 2.3

Currently when we spawn an actor, we pass two shared structs actor, `SharedContext` and `LocalBarrierManager`. The lifecycle of the two structs is basically the same: they are bind to a database, and get created along with having a new running database, and get cleared when the database is reset or global recovery is triggered. 

Previously we store the `SharedContext` separately to `DatabaseManagedBarrierState`, while the `LocalBarrierManager` is stored in `DatabaseManagedBarrierState`. This is because, `SharedContext` is used to store the exchange channel between actors, and the creation of new running database is asynchronous, and it may happen that the exchange request from other CN is received before the request to create a new running database. As a result, we may create the `SharedContext` of a database before the database is created, and we store the `SharedContext` separately to `DatabaseManagedBarrierState`.

In this PR, we will try to store the `SharedContext` in `LocalBarrierManager`, so that we only need to pass the `LocalBarrierManager` when we spawn a new actor. This means that, the `SharedContext` will be created along with `LocalBarrierManager` when we have a new request to create a new running `DatabaseManagedBarrierState`. To handle the early-arrived exchange requests, we introduce a new variant `DatabaseStatus::ReceivedExchangeRequests`. When receiving exchange request on a database that is not created yet, we will mark the database status as `ReceivedExchangeRequests`, and store this pending requests. When the database is later created, it will create the `SharedContext`, and handle pending requests in the `SharedContext`.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/executor/chain.rs
+++ b/src/stream/src/executor/chain.rs
@@ -115,11 +115,13 @@ mod test {
     use super::ChainExecutor;
     use crate::executor::test_utils::MockSource;
     use crate::executor::{AddMutation, Barrier, Execute, Message, Mutation, PkIndices};
-    use crate::task::{CreateMviewProgressReporter, LocalBarrierManager};
+    use crate::task::CreateMviewProgressReporter;
+    use crate::task::barrier_test_utils::LocalBarrierTestEnv;
 
     #[tokio::test]
     async fn test_basic() {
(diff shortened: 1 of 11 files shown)


yezizp2012 (approved):
LGTM

Issue #21259, opened 2025-04-07
bug: SINK INTO TBALE does not handle partial columns well


tabVersion:
### Describe the bug

```
dev=> create table t ( a int, b varchar , c int ) ;
CREATE_TABLE
dev=> create table t1 ( d int , e int ) ;
CREATE_TABLE
dev=> create sink s3 into t1 ( d ) as select a from t with (type = 'append-only', force_append_only='true' ) ; ;
CREATE_SINK
dev=> show create sink s3 ;
   Name    |                                            Create Sql
-----------+---------------------------------------------------------------------------------------------------
 public.s3 | CREATE SINK s3 INTO t1 AS SELECT a FROM t WITH (type = 'append-only', force_append_only = 'true')
(1 row)
```

the show create sink statement lacks the column info `(d)` and lead to possible failure if using the normalized sql

### Error message/log

```text

```

Issue #20713, opened 2025-03-04
Tracking: `Display` of sqlparser AST generates invalid SQL


xiangjinwu:
TODOs:
The removed lines are valid inputs. The added lines are invalid actual output.
```diff
- CREATE TABLE T ("FULL" INT) ON CONFLICT DO UPDATE FULL WITH VERSION COLUMN("FULL")
+ CREATE TABLE T ("FULL" INT) ON CONFLICT DO UPDATE FULL WITH VERSION COLUMN(FULL)

- INSERT INTO t VALUES(1,3), (2,4) RETURNING *, a, a as aaa
+ INSERT INTO t VALUES (1, 3), (2, 4) RETURNING (*, a, a AS aaa)

- select * from t1 left join t2 FOR SYSTEM_TIME AS OF NOW() on a1 = a2;
+ SELECT * FROM t1 LEFT JOIN t2 FOR SYSTEM_TIME AS OF NOW() - 0 SECOND ON a1 = a2

- alter materialized view m2 set backfill_rate_limit=default;
+ ALTER MATERIALIZED VIEW m2 SET BACKFILL_RATE_LIMIT TO -1

- CREATE FUNCTION sleep_always_retry(INT) RETURNS INT AS 'sleep' USING LINK 'http://localhost:8815' WITH ( always_retry_on_network_error = true );
+ CREATE FUNCTION sleep_always_retry(INT) RETURNS INT AS 'sleep' USING LINK 'http://localhost:8815' WITH ( ALWAYS_RETRY_NETWORK_ERRORS = true )

- ALTER TABLE diamonds SET source_rate_limit TO DEFAULT;
+ ALTER TABLE diamonds SET SOURCE_RATE_LIMIT TO -1

- CREATE SINK s_postgres_0 FROM mv_remote_0 WITH (
-  connector='jdbc',
-  jdbc.url='jdbc:postgresql://db:5432/test',
-  user='test',
-  password='connector',
-  table.name='t_remote_0',
-  primary_key='id',
-  type='upsert'
-  )
- ;
+ CREATE SINK s_postgres_0 FROM mv_remote_0 WITH (connector = 'jdbc', jdbc.url = 'jdbc:postgresql://db:5432/test', user = 'test', password = [REDACTED], table.name = 't_remote_0', primary_key = [REDACTED], type = 'upsert')

- explain (distsql, trace, verbose) select 1;
+ EXPLAIN VERBOSE,TRACE,TEXTSELECT 1

[...]

PR #21260, opened 2025-04-07
fix: `SHOW CREATE SINK` missing column spec

tabVersion:

Highest Release Version: 2.3

fix https://github.com/risingwavelabs/risingwave/issues/21259

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/sqlparser/src/ast/statement.rs
+++ b/src/sqlparser/src/ast/statement.rs
@@ -517,14 +517,16 @@ impl ParseTo for CreateSinkStatement {
         impl_parse_to!(if_not_exists => [Keyword::IF, Keyword::NOT, Keyword::EXISTS], p);
         impl_parse_to!(sink_name: ObjectName, p);
 
+        let mut columns = Vec::new();
         let into_table_name = if p.parse_keyword(Keyword::INTO) {
             impl_parse_to!(into_table_name: ObjectName, p);
+
+            // we only allow specify columns when creating sink into a table
+            columns = p.parse_parenthesized_column_list(IsOptional::Optional)?;
             Some(into_table_name)
         } else {
             None
         };
-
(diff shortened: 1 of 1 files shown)


BugenZhao (approved):
LGTM

cc @xiangjinwu  To prevent similar issues from happening, shall we consider:

1. Also cover `CreateSinkStatement` in https://github.com/risingwavelabs/risingwave/issues/20713
2. Ensure that the unparsed statement is not only parsable but also yields the exact same result when parsed.

xiangjinwu:
> cc @xiangjinwu To prevent similar issues from happening, shall we consider:
> 
> 1. Also cover `CreateSinkStatement` in [Tracking: `Display` of sqlparser AST generates invalid SQL #20713](https://github.com/risingwavelabs/risingwave/issues/20713)
> 2. Ensure that the unparsed statement is not only parsable but also yields the exact same result when parsed.

Definitely yes. It just takes a lot of efforts to fix existing issues ...

> yields the exact same result 

I also considered an alternative, to relax it to being "equivalent". That is, wrap the top level `parse` function to always `parse-display-parse`, so that developers always see the re-parsed results and would notice the unexpected result. However it can also cause confusion when it produces that equivalent AST.

src/sqlparser/src/ast/statement.rs
```diff
@@ -572,6 +577,9 @@ impl fmt::Display for CreateSinkStatement {
         if let Some(into_table) = &self.into_table_name {
             impl_fmt_display!([Keyword::INTO], v);
             impl_fmt_display!([into_table], v);
+            if !self.columns.is_empty() {
+                v.push(format!("({})", display_comma_separated(&self.columns)));
+            }
```
xiangjinwu, line 582: Always accompany a test case that reproduces the previous bug

Issue #20028, opened 2025-01-06
more efficient array builder


lmatz:
eliminate dynamic dispatch as much as possible

and maybe bulk append the rows so that static dispatch and dynamic dispatch (if not possible to remove) are amortized. This does not necessarily contradict the per-row handling logic in the join operator.

checkout context: https://risingwave-labs.slack.com/archives/C034TRPKN1F/p1733464753437989

hzxa21:
@kwannoel and @st1page have discussed on this topic and think we should focus on improving our vectorization execution instead.

PR #21289, opened 2025-04-08
feat: allow user to force filter pushed into join in streaming query

lmatz:

We have a known issue #20028  that building rows into arrays, especially certain data types, this building process takes quite some time. Filtering unnecessary rows earlier in the join can reduce the cost.

Apart from building rows into arrays, materializing possibly many rows only to get them filtered in the next executor is just wasteful.

However, it could be generally hard to decide which one is better, as the existing comment suggest that
```
// We do so as the filter operator can apply the non-equal condition batch-wise (vectorized)
// as opposed to the HashJoin, which applies the condition row-wise.
```
vectorization may help.

Since the overall efficiency could depend on the join selectivity, array building efficiency and vectorization benefits,
we make it a session variable instead of deciding automatically by the optimizer.

We make its default behavior the same as the current one.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/planner_test/tests/testdata/output/join.yaml
+++ b/src/frontend/planner_test/tests/testdata/output/join.yaml
@@ -756,3 +756,43 @@
     ├── distribution key: [ 0, 1, 3, 4, 5, 6 ]
     └── read pk prefix len hint: 6
 
+- name: Force filter pushed down into join
+  sql: |
+    set streaming_force_filter_inside_join = true;
+    create table t (src int, dst int);
+    select t1.src, t2.dst from t t1, t t2 where t1.src = t2.dst and t1.dst < t2.src;
+  batch_plan: |-
+    BatchExchange { order: [], dist: Single }
+    └─BatchHashJoin { type: Inner, predicate: t.src = t.dst AND (t.dst < t.src), output: [t.src, t.dst] }
+      ├─BatchExchange { order: [], dist: HashShard(t.src) }
+      │ └─BatchScan { table: t, columns: [t.src, t.dst], distribution: SomeShard }
+      └─BatchExchange { order: [], dist: HashShard(t.dst) }
+        └─BatchScan { table: t, columns: [t.src, t.dst], distribution: SomeShard }
+  stream_plan: |-
+    StreamMaterialize { columns: [src, dst, t._row_id(hidden), t._row_id#1(hidden)], stream_key: [t._row_id, t._row_id#1, src], pk_columns: [t._row_id, t._row_id#1, src], pk_conflict: NoCheck }
+    └─StreamExchange { dist: HashShard(t.src, t._row_id, t._row_id) }
+      └─StreamHashJoin { type: Inner, predicate: t.src = t.dst AND (t.dst < t.src), output: [t.src, t.dst, t._row_id, t._row_id] }
+        ├─StreamExchange { dist: HashShard(t.src) }
+        │ └─StreamTableScan { table: t, columns: [t.src, t.dst, t._row_id], stream_scan_type: ArrangementBackfill, stream_key: [t._row_id], pk: [_row_id], dist: UpstreamHashShard(t._row_id) }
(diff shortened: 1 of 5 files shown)


lmatz:
wonder if we are aware of any known difficulties regarding:
https://github.com/risingwavelabs/risingwave/blob/main/src/stream/src/executor/hash_join.rs#L1259-L1287

src/frontend/planner_test/tests/testdata/output/join.yaml
```diff
@@ -756,3 +756,43 @@
...
+    BatchExchange { order: [], dist: Single }
+    └─BatchHashJoin { type: Inner, predicate: t.src = t.dst AND (t.dst < t.src), output: [t.src, t.dst] }
+      ├─BatchExchange { order: [], dist: HashShard(t.src) }
+      │ └─BatchScan { table: t, columns: [t.src, t.dst], distribution: SomeShard }
+      └─BatchExchange { order: [], dist: HashShard(t.dst) }
+        └─BatchScan { table: t, columns: [t.src, t.dst], distribution: SomeShard }
+  stream_plan: |-
+    StreamMaterialize { columns: [src, dst, t._row_id(hidden), t._row_id#1(hidden)], stream_key: [t._row_id, t._row_id#1, src], pk_columns: [t._row_id, t._row_id#1, src], pk_conflict: NoCheck }
+    └─StreamExchange { dist: HashShard(t.src, t._row_id, t._row_id) }
+      └─StreamHashJoin { type: Inner, predicate: t.src = t.dst AND (t.dst < t.src), output: [t.src, t.dst, t._row_id, t._row_id] }
```
kwannoel, line 774: Why not push down to scan?

kwannoel: Oh we can't.. nevermind.

kwannoel: Hmm wait a second this also does not make sense to me:
```
t.src = t.dst AND (t.dst < t.src)
```

This is always false.

kwannoel: Maybe we can add a more realistic test case, so it won't distract others who see this test 😅

kwannoel (approved):
Rest LGTM, also I think we may do a simple slt test, to just compare runtime of some scenario where this can help.

Just need to update the test case, so we won't confuse others.

lmatz:
a contrived example
```
dev=> create table t(v1 int, v2 int);
dev=> create table t2(v3 int, v4 int);
dev=> insert into t select 1, id as v2 from generate_series(1, 2000) t(id);
dev=> insert into t2 select 1, id as v4 from generate_series(1, 2000) t(id);

dev=> set streaming_force_filter_inside_join = false;

dev=> create materialized view mv as select * from t, t2 where t.v1 = t2.v3 and t.v2 > t2.v4 + 1000;
CREATE_MATERIALIZED_VIEW
Time: 22553.773 ms (00:22.554)

dev=> create materialized view mv as select * from t, t2 where t.v1 = t2.v3 and t.v2 > t2.v4 + 1000;
CREATE_MATERIALIZED_VIEW
Time: 22911.186 ms (00:22.911)

dev=> set streaming_force_filter_inside_join = true;

dev=> create materialized view mv as select * from t, t2 where t.v1 = t2.v3 and t.v2 > t2.v4 + 1000;
CREATE_MATERIALIZED_VIEW
Time: 17850.257 ms (00:17.850)


dev=> create materialized view mv as select * from t, t2 where t.v1 = t2.v3 and t.v2 > t2.v4 + 1000;
CREATE_MATERIALIZED_VIEW
Time: 18196.095 ms (00:18.196)
```

src/frontend/planner_test/tests/testdata/output/join.yaml
lmatz: cool, let me use the example below

PR #21294, opened 2025-04-08
feat: let `ALTER SYSTEM set <session_parameter>` take effect in the current session

shanicky:

This PR modifies the behavior of using "ALTER SYSTEM" to change session parameters. The previous code did not apply the changes to the current session after the alteration, which should align with PostgreSQL. This PR attempts to directly apply the changes to the current session after the modification is completed.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/handler/alter_system.rs
+++ b/src/frontend/src/handler/alter_system.rs
@@ -51,15 +51,15 @@ pub async fn handle_alter_system(
             .await?;
 
         if let Some(value) = value {
-            builder.notice(format!(
-                "The config #{param_name} of the current session has already been set to #{value}"
+            builder = builder.notice(format!(
+                "The config {param_name} of the current session has already been set to {value}"
             ));
             handler_args
                 .session
                 .set_config(param_name.as_str(), value)?;
         } else {
-            builder.notice(format!(
-                "The config #{param_name} of the current session has already been reset to default"
+            builder = builder.notice(format!(
+                "The config {param_name} of the current session has already been reset to default"
             ));
             handler_args.session.reset_config(param_name.as_str())?;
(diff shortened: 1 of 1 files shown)


kwannoel (approved):

kwannoel:
Mirroring some discussion offline with @xiangjinwu and @yezizp2012 for visibility:

This is a deviation from PG's default behaviour stated [here](https://www.postgresql.org/docs/current/sql-altersystem.html):
> Values set with ALTER SYSTEM will be effective after the next server configuration reload, or after the next server restart in the case of parameters that can only be changed at server start. A server configuration reload can be commanded by calling the SQL function pg_reload_conf(), running pg_ctl reload, or sending a SIGHUP signal to the main server process.

Most of our session variables are RisingWave specific so this change is acceptable.
For non-specific RisingWave variables we have `timezone`.

It is very unlikely this will break any **important** tools. Since tools will be unlikely to depend on `alter system set timezone` NOT changing the timezone for the current session.

So we will go ahead and make this change since it has a lot more benefits.

Users should expect the defaults for `session_config` to take effect in their current session, as soon as they set it via `alter system set <session_config>`.

But this is not the current behaviour. So it can lead to some logic error, where user configured a session variable, but it did not actually take effect.

yuhao-su (approved):

BugenZhao:
> The previous code did not apply the changes to the current session after the alteration, which should align with PostgreSQL

Shall we update the PR body then?

kwannoel:
> > The previous code did not apply the changes to the current session after the alteration, which should align with PostgreSQL
> 
> Shall we update the PR body then?

Let's add a release note

PR #21301, opened 2025-04-08
feat(test): support mocked exactly once iceberg sink in simulation environment

wcy-fdu:

This pr depends on the previous PR #21214 , thanks @wenym1  for supporting the testing of the coordinated sink in simulation test.

This PR is designed to test the exactly-once semantics of the Iceberg sink within the simulation framework.
Note that it specifically targets only the exactly-once aspect. For testing other functionalities of the Iceberg sink, you can refer to the `test_decoupled_coordinated_sink` introduced by @wenym1.

### mock iceberg exactly once commit
Currently, out read Iceberg commit is done using `snapshot_id`, which is why we constructed a `snapshot_id_to_data: BTreeMap<snapshot_id, HashMap<id, name>>` in this pr to simulate an external Iceberg. Each time we want to commit a batch of data to Iceberg, a unique `snapshot_id` is used to insert into the BTreeMap. The check for whether a commit has been made is also consistent with the real environment, using the snapshot_id to verify its existence. In addition to mocking an Iceberg, the other parts are consistent with the real exactly-once implementation (such as using the meta store's system tables for pre-commit and the entire commit process, etc.).

Based on the following Commit Timeline and corresponding actions, we support injecting errors at five specific points in time to trigger recovery. The err_rate_list is used to control the sequence of different errors, simulating some hard-to-detect corner cases. 

Currently, two tests have been introduced:

- Basic Test: A batch of unique IDs flows from the source. Errors are injected with a 10% probability at five stages. After the sink completes, if no duplicates are found in the mocked external sink, it confirms the exactly-once guarantee.
- Should Panic Test: This test simulates corner cases(like [this](https://github.com/risingwavelabs/risingwave/pull/19771#discussion_r2000561951)). When should_panic is enabled, certain parts use erroneous implementations, which should intentionally cause duplicates, thereby confirming that the corner case has been detected.

### Iceberg Commit Timeline

| time | Action                                                                                   |
|------|------------------------------------------------------------------------------------------|
|  0   | Collect metadata from various parallelism.                                              |
|  1   | Prepare 2pc: write pre-commit into system table.                                       |
|  2   | Do iceberg commit.                                                                       |
|  3   | Mark the record with `epoch = 2` in the system table as committed.                     |
|  4   | Delete the record with `epoch < 2` in the system table.      

### Recovery Action
The error can happen at any of the above points, let's see how to handle each case when recovery: 

| **Crash Scenario**                                   | **Actions**                                                                                     | **Log Store Operations**                      |
|-----------------------------------------------------|------------------------------------------------------------------------------------------------|-----------------------------------------------|
| **Crash before 0**                                  | Just rewind log store is fine.                                                                 | Rewind to truncate offset.                    |
| **Crash during (0, 1)**                             | Just rewind log store is fine.                                                                 | Rewind to truncate offset.                    |
| **Crash during (1, 2)**                             | Find record with `epoch = 2` not in Iceberg (snapshot does not exist), it's `committed` is false. Re-commit this record. | Rewind from epoch 2 (consume data after epoch = 2) since we can to commit based on system table, so log store can be read from epoch>2               |
| **Crash during (2, 3)**                             | Find record with `epoch = 2` in system table and Iceberg, committed is false. Set `committed` to true. | Rewind from epoch 2 (consume data after epoch2). |
| **Crash during (3, 4)**                             | Find record with `epoch = 2` in system table and Iceberg, committed is false. Skip; this record will be deleted when committing a larger epoch later. | Rewind from epoch 2 (consume data after epoch 2). |
| **Crash after 4**                                   | Do nothing since everything is fine.                                                           | Rewind from epoch 2 (consume data after epoch 2). |

## Checklist

## Documentation

<details>

</details>


src/tests/simulation/tests/integration_tests/sink/exactly_once_iceberg.rs
```diff
@@ -0,0 +1,135 @@
...
+    test_exactly_once_sink_inner(err_rate_list, false).await
+}
+
+#[tokio::test]
+#[should_panic]
+async fn test_exactly_once_sink_should_panic() {
+    // If, during a re-commit, a new snapshot_id is generated instead of using the previously persisted snapshot_id, it can lead to duplicate data when consecutive "23" appears in err_events.
+    let err_rate_list = vec![0.0, 0.0, 0.3, 0.3, 0.0];
+    let _ = test_exactly_once_sink_inner(err_rate_list, true).await;
+}
```
chenzl25, line 44: If the error rate is not 1, how can it ensure the test must panic?

wcy-fdu: After some offline discussion, removed the should panic test.

chenzl25 (approved):
LGTM!

hzxa21 (commented):
All the changes in cargo lock are version downgrade. Is it expected?

wcy-fdu:
> All the changes in cargo lock are version downgrade. Is it expected?



> All the changes in cargo lock are version downgrade. Is it expected?

Manually remove the unrelated changes.

src/tests/simulation/tests/integration_tests/sink/exactly_once_utils.rs
```diff
@@ -0,0 +1,915 @@
...
+impl SimulationTestIcebergCommitter {
+    async fn persist_pre_commit_metadata(
+        &self,
+        db: DatabaseConnection,
+        start_epoch: u64,
+        end_epoch: u64,
+        pre_commit_metadata: Vec<u8>,
+        snapshot_id: i64,
+    ) -> Result<()> {
+        let m = exactly_once_iceberg_sink::ActiveModel {
```
wenym1, line 655: It seems that we are still accessing the internal meta store table manually in the test. In this way, the logic of accessing meta store table in the original implementation is not tested at all. Can we possibly reuse these logic of accessing meta store table in the original implementation and only mock the access to iceberg? 

In our offline discussion, we once mentioned that we can extract all access to iceberg table to a common trait, and then we can mock accessing iceberg via implementing this trait.

src/stream/src/common/log_store_impl/kv_log_store/mod.rs
```diff
@@ -1209,10 +1209,7 @@ mod tests {
             // Though we don't truncate reader2 with epoch1, we have truncated reader1 with epoch1, and with align_init_epoch
             // set to true, we won't receive the following commented items.
             match reader.next_item().await.unwrap() {
-                (epoch, LogStoreReadItem::StreamChunk { chunk, .. }) => {
-                    assert_eq!(epoch, epoch1);
-                    assert!(check_stream_chunk_eq(&chunk1_2, &chunk));
-                }
+                (_epoch, LogStoreReadItem::StreamChunk { chunk: _, .. }) => {}
```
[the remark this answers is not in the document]
wcy-fdu: It was changed by mistake when resolving conflicts, and has been restored.

src/tests/simulation/tests/integration_tests/sink/exactly_once_utils.rs
wcy-fdu: I functionalized the logic of the operating system table after enabling iceberg exactly once, and then used the same set of logic in the test and real environments.

hzxa21 (approved):
LGTM for Cargo.lock

wcy-fdu:
After some offline discussions, I realized that the mocked part can be further narrowed.
I will merge this PR first and add this part of the logic in another PR later.

PR #21303, opened 2025-04-09
feat(ctl): add an IO-bound rule for await tree analyze

KeXiangWang:

See https://github.com/risingwavelabs/rw-diagnose-tools/compare/main%40%7B1day%7D...main for more impl information.
This PR makes some changes:
1. Add another rule to detect actors blocked by IO operations, such as `store_flush` and `store_get`
2. Refactor the code, place output part (`println`) in riseclt instead of `rw-diagnose-tools`.
3. Support analyze await tree dump file from meta dashboard, the format of which is slightly different from the format of dump file by risectl.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/ctl/src/cmd_impl/await_tree.rs
+++ b/src/ctl/src/cmd_impl/await_tree.rs
@@ -12,14 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use anyhow::Context;
 use risingwave_common::util::StackTraceResponseExt;
 use risingwave_common::util::addr::HostAddr;
 use risingwave_pb::common::WorkerType;
 use risingwave_pb::monitor_service::stack_trace_request::ActorTracesFormat;
 use risingwave_pb::monitor_service::{StackTraceRequest, StackTraceResponse};
 use risingwave_rpc_client::{CompactorClient, ComputeClientPool};
-use rw_diagnose_tools::await_tree::TreeView;
+use rw_diagnose_tools::await_tree::AnalyzeSummary;
 
 use crate::CtlContext;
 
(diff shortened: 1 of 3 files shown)


src/ctl/src/cmd_impl/await_tree.rs
```diff
@@ -73,14 +72,16 @@ pub async fn dump(context: &CtlContext, actor_traces_format: Option<String>) ->
 }
 
 pub async fn bottleneck_detect(context: &CtlContext, path: Option<String>) -> anyhow::Result<()> {
-    if let Some(path) = path {
-        rw_diagnose_tools::await_tree::bottleneck_detect_from_file(path)
+    let summary = if let Some(path) = path {
```
kwannoel, line 75: In the upstream, I notice that `AnalyzeSummary` has the following format:
```rs
pub struct AnalyzeSummary {
    has_fast_children_actors: HashMap<u32, TreeView>,
    /// IO bound rule usually match a lot of Trees once the storage is unavailable, as a
    /// result, too many trees are outputed. We only output the actor ids here.
    io_bound_actors: Vec<u32>,
}
```

Initially this implementation is alright, I guess we can `ctrl-F` ourselves to locate the span.
But subsequently I would like us to have an option to **aggregate by reason**

So the analyze summary should look like:
```rs
type IoInfo = String;

pub struct AnalyzeSummary {
    has_fast_children_actors: HashMap<u32, TreeView>,
    /// IO bound rule usually match a lot of Trees once the storage is unavailable, as a
    /// result, too many trees are outputed. We only output the actor ids here.
    io_bound_actors: HashMap<IoInfo, Set<u32>>,
}
```

kwannoel (approved):

BugenZhao (approved):
LGTM

MrCroxx (approved):

Issue #19641, opened 2024-12-02
MQTT source with wildcard topic periodically errors in logs


florissmit10:
### Describe the bug

When there is a wildcard in the MQTT topic, the connection periodically gets interupted or taken over. This causes errors in the logs.

### Error message/log

```text
2024-12-02T10:02:36.177851778Z ERROR actor{otel.name="Actor 34" actor_id=34 prev_epoch=7595526839271424 curr_epoch=7595526904807424}:executor{otel.name="Source 2200002715" barrier}:source_parse_chunk{actor_id=34 source_id=54}: risingwave_connector::source::mqtt::source::reader: Failed to poll mqtt eventloop: Mqtt state: Server sent disconnect with reason `None` and code 'SessionTakenOver' 
2024-12-02T10:02:36.183927124Z ERROR actor{otel.name="Actor 30" actor_id=30 prev_epoch=7595526904807424 curr_epoch=7595526970343424}:executor{otel.name="Source 1E00002715" barrier}:source_parse_chunk{actor_id=30 source_id=51}: risingwave_connector::source::mqtt::source::reader: Failed to poll mqtt eventloop: Mqtt state: Server sent disconnect with reason `None` and code 'SessionTakenOver' 
2024-12-02T10:02:36.18508886Z ERROR actor{otel.name="Actor 34" actor_id=34 prev_epoch=7595527560167424 curr_epoch=7595527625703424}:executor{otel.name="Source 2200002715" barrier}:source_parse_chunk{actor_id=34 source_id=54}: risingwave_connector::source::mqtt::source::reader: Failed to poll mqtt eventloop: Mqtt state: Server sent disconnect with reason `None` and code 'SessionTakenOver' 
2024-12-02T10:02:36.18909765Z ERROR actor{otel.name="Actor 30" actor_id=30 prev_epoch=7595527560167424 curr_epoch=7595527625703424}:executor{otel.name="Source 1E00002715" barrier}:source_parse_chunk{actor_id=30 source_id=51}: risingwave_connector::source::mqtt::source::reader: Failed to poll mqtt eventloop: Mqtt state: Server sent disconnect with reason `None` and code 'SessionTakenOver'
```

### To Reproduce

There are 3 devices publising data every 10s on topic "tele/{d1,d2,d3}/SENSOR" Create table:
```
CREATE TABLE "dev"."public"."sensor_raw" (
    Time TIMESTAMP,
    ENERGY STRUCT <
[...]

PR #21323, opened 2025-04-09
fix(connector) Dont resubscibe mqtt client and propagate error

florissmit10:

A while back we encountered some errors in our environment (see https://github.com/risingwavelabs/risingwave/issues/19641 )
After investigating we thing we pinpointed the error and this will fix it.
The biggest change is that the MQTT client no longer resubscribes to topics on error. This is normal because the subscription is saved in the client, even though there are errors. Polling the client will handle the reconnect.

## Checklist

## Documentation

<details>

Increased the stability of the MQTT Source connector

</details>


src/connector/src/source/mqtt/source/reader.rs
```diff
@@ -29,11 +28,6 @@ use crate::source::{BoxSourceChunkStream, Column, SourceContextRef, SourceMessag
 
 pub struct MqttSplitReader {
     eventloop: rumqttc::v5::EventLoop,
-    client: rumqttc::v5::AsyncClient,
-    qos: QoS,
-    splits: Vec<MqttSplit>,
-    #[expect(dead_code)]
-    properties: MqttProperties,
```
tabVersion, line 36: let's keep the props for future use (maybe), just allow dead code here should be ok.

src/connector/src/source/mqtt/enumerator/mod.rs
```diff
@@ -29,10 +29,6 @@ use crate::error::ConnectorResult;
 use crate::source::{SourceEnumeratorContextRef, SplitEnumerator};
 
 pub struct MqttSplitEnumerator {
-    #[expect(dead_code)]
-    topic: String,
-    #[expect(dead_code)]
-    client: rumqttc::v5::AsyncClient,
```
tabVersion, line 35: ditto

tabVersion (approved):
rest LGTM

florissmit10:
What is needed to merge this?

PR #21351, opened 2025-04-11
feat(iceberg): support hosted SQL catalog for iceberg engine

chenzl25:

- Support hosted SQL catalog for iceberg engine. Users can specify `hosted_catalog = true` in their iceberg engine connection to enable the hosted iceberg catalog. The current hosted iceberg catalog implementation is a SQL catalog backed by the meta store (typically a Postgres).

## Example

```
create connection my_hosted_catalog_conn
with (
    type = 'iceberg',
    s3.region = 'us-west-2',
    hosted_catalog = true
);
```

## Checklist

## Documentation

<details>

- Add option `hosted_catalog` to the iceberg connection. Users can use `hosted_catalog=true` to enable the hosted catalog (currently, which is a JDBC/SQL catalog managed by RisingWave). This option could only be used in iceberg table engine, but not iceberg source and iceberg sink.

</details>


chenzl25:
The ci is using a old version of minio `Version: RELEASE.2023-05-04T21-44-30Z (go1.20.3 linux/amd64)` which cause the following error.
```
Exception in thread "Thread-8" software.amazon.awssdk.services.s3.model.S3Exception: The provided 'x-amz-content-sha256' header does not match what was computed. (Service: S3, Status Code: 400, Request ID: 18354C689AFE1AAB, Extended Request ID: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855) (SDK Attempt Count: 1)
```

chenzl25:
After upgrading the CI minio, it passes the test.

src/frontend/src/handler/create_table.rs
```diff
@@ -1627,6 +1627,22 @@ pub async fn create_iceberg_engine_table(
                 with_common.insert("connector".to_owned(), "iceberg".to_owned());
                 with_common.insert("database.name".to_owned(), iceberg_database_name.to_owned());
                 with_common.insert("table.name".to_owned(), iceberg_table_name.to_owned());
+
+                if let Some(s) = params.properties.get("hosted_catalog") {
+                    if s.eq_ignore_ascii_case("true") {
+                        with_common.insert("catalog.type".to_owned(), "jdbc".to_owned());
+                        with_common.insert("catalog.uri".to_owned(), catalog_uri.to_owned());
```
xxchan, line 1634: Maybe we can delete the `#[allow(dead_code)]` branch above now? 🤔

xxchan (approved):
others LGTM

src/frontend/src/handler/create_table.rs
```diff
@@ -1627,6 +1627,22 @@ pub async fn create_iceberg_engine_table(
...
+                        with_common
+                            .insert("catalog.jdbc.user".to_owned(), meta_store_user.to_owned());
+                        with_common.insert(
+                            "catalog.jdbc.password".to_owned(),
+                            meta_store_password.clone(),
+                        );
+                        with_common
+                            .insert("catalog.name".to_owned(), iceberg_catalog_name.to_owned());
```
xxchan, line 1642: Maybe we should check the config is not set by user here.

src/frontend/src/handler/create_table.rs
chenzl25: Yes. Let me remove it.

src/connector/src/connector_common/connection.rs
```diff
@@ -265,6 +273,27 @@ impl Connection for IcebergConnection {
...
+                bail!("`catalog.name` must not be set when `hosted_catalog` is set");
+            }
+            if self.jdbc_user.is_some() {
+                bail!("`catalog.jdbc.user` must not be set when `hosted_catalog` is set");
+            }
+            if self.jdbc_password.is_some() {
+                bail!("`catalog.jdbc.password` must not be set when `hosted_catalog` is set");
+            }
+            return Ok(());
+        }
```
chenzl25, line 295: > Maybe we should check the config is not set by user here, which will be overrideen by us.

We check it when creating iceberg connection cc @xxchan

xxchan: LGTM 👍

Issue #20525, opened 2025-02-18
test: better e2e test framework


stdrc:
```
# file structure
e2e_test
    |- streaming
        |- main.py # discover all tests in `tests`
        |- risedev.yml
        |- tests
            |- a.slt
            |- b.slt
    |- batch
        |- ...
```

```sh
risedev e2e streaming
```

xxchan:
What problem does it solve?

xxchan:
Some discussion: SLT itself isn't a very good DSL. 
- The original motivation may be reuse existing tests from other DBs. But it's not our need now.
- It lacks many abilities, and introduced quite a few features. But we still want some more from time to time. Mainly some kind of "templating" 
    * reuse tests for different config https://risingwave-labs.slack.com/archives/C04S2HVU99V/p1744009501065689?thread_ts=1744006963.857519&cid=C04S2HVU99V
    * non-exact match for results https://github.com/risingwavelabs/risingwave/pull/21354#discussion_r2048258148
    * conditionals, etc.
    * https://github.com/risinglightdb/sqllogictest-rs/issues/143
- Moreover, we may want some kind of per-test orchestration (different `risedev.yml` cluster setup, different test discovery, etc.)
- At the end of day, why do we need a DSL? We can still do data-driven tests as a library, using rust or python.
    * stream executor integration tests (#9787), and Madsim integration tests https://github.com/risingwavelabs/risingwave/tree/main/src/tests/simulation/tests/integration_tests are basically just this.
    * The library provide some machinery to run basic sqls very easily. (Well, or some kind of DSL...)
    * When we need special handing of the results, we just use the programming language's native ability.

BugenZhao:
Looking forward to further design. From my own perspective, however, I still believe the flexibility is always a trade-off: with limited expressiveness, we still gain some benefits with `sqllogictest-rs`.

xxchan:
I agree for (maybe most) simple cases, DSL may be convenient. SLT + system commands are working fine. Maybe I just want a less verbose DSL, e.g., using markers like `>` to run SQL. 🤣 Besides, it can support some built-in commands, instead of full-fledged bash, as you dislike/

xxchan:
Maybe we can experimentally support an alternative syntax in `sqllogictest-rs`...

xxchan:
#21082 is kind of this

xxchan:
# DRAFT: E2E Test Framework Improvement Plan

By Claude Code (to be reviewed)

[...]

PR #21354, opened 2025-04-11
feat: support `DESCRIBE PLAN (options)`

### LINKED ISSUE #21254: Introduce a sytax sugar for checking a created MV's plan (`DESCRIBE PLAN mv`)
Currently it's doable via `EXPLAIN CREATE MATERIALIZED VIEW` (which won't check name exists). But it needs to fetch the MV's sql manually first. It's is more user friendly to have a dedicated syntax for this purpose.

[Slack Message](https://risingwave-labs.slack.com/archives/C04S2HVU99V/p1744008921557039?thread_ts=1744006963.857519&cid=C04S2HVU99V)

xxchan:
close #21254

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/.cursor/rules/build-run-test.mdc
+++ b/.cursor/rules/build-run-test.mdc
@@ -9,10 +9,15 @@ You may need to learn how to build and test RisingWave when implementing feature
 
 - Use `./risedev b` to build the project.
 - Use `./risedev c` to check if the code follow rust-clippy rules, coding styles, etc.
-- Use `./risedev d` to run a RisingWave instance. If the project is not built, this will build the project before running. The instance will run in backgroud, so you don't need to start a new terminal to run this command. You can connect to the instance right after this command is finished.
+- Use `./risedev d` to run a RisingWave instance. If the project is not built, this will build the project before running. The instance will run in background, so you don't need to start a new terminal to run this command. You can connect to the instance right after this command is finished.
 - Use `./risedev k` to stop a RisingWave instance started by `./risedev d`.
 - The log files of a running RisingWave are located in `.risingwave/log` folder in the workspace. You can check them (`tail` is suggested because they can be quite long) if things go wrong.
 - When a RisingWave instance is running, you can use `./risedev psql -c "<your query>"` to run SQL queries in RW.
 - When a RisingWave instance is running, you can use `./risedev slt './path/to/e2e-test-file.slt'` to run end-to-end SLT tests. File globs like `/**/*.slt` is allowed.
 - Integration tests and unit tests are valid Rust/Tokio tests, you can locate and run those related in a standard Rust way.
 - `./risedev` command is safe to run automatically.
+
+## Testing
+
+The preferred way to write tests is to write tests in SQLLogicTest format.
+Tests are put in `./e2e_test` folder.
--- a/e2e_test/batch/describe_plan.slt
+++ b/e2e_test/batch/describe_plan.slt
@@ -1,0 +1,293 @@
+# Test DESCRIBE PLAN with complex relations and schemas
+
+# Setup test schema
+statement ok
+CREATE SCHEMA describe_plan_test;
+
+statement ok
+CREATE TABLE describe_plan_test.tbl (
+    id INT PRIMARY KEY,
+    name VARCHAR NOT NULL,
+    age INT,
+    created_at TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP,
+    data JSONB
+);
+
+# Test materialized view with complex query
+statement ok
+CREATE MATERIALIZED VIEW describe_plan_test.complex_mv AS
+SELECT
+    t.name,
+    COUNT(*) as count,
+    AVG(t.age) as avg_age,
+    jsonb_agg(t.data) as all_data
+FROM describe_plan_test.tbl t
+WHERE t.age > 18
+GROUP BY t.name
+HAVING COUNT(*) > 1;
+
+# Test index with included columns and custom distribution
+statement ok
+CREATE INDEX idx ON describe_plan_test.tbl (name DESC, age)
+INCLUDE (created_at)
+DISTRIBUTED BY (name);
+
+# Test source with complex schema and options
+statement ok
+CREATE SOURCE describe_plan_test.src (
+    id INT,
+    event_time TIMESTAMP,
+    payload STRUCT<name VARCHAR, value INT>,
+    metadata MAP(VARCHAR, VARCHAR)
+) WITH (
+    connector = 'datagen',
+);
+
+# Test sink with transformation
+statement ok
+CREATE SINK describe_plan_test.snk AS
+SELECT
+    id,
+    name,
+    age,
+    created_at,
+    data->>'type' as type
+FROM describe_plan_test.tbl
+WHERE age >= 21
+WITH (connector ='blackhole')
+
+# Test view with subquery and joins
+statement ok
+CREATE VIEW describe_plan_test.complex_view AS
+WITH aged_users AS (
+    SELECT name, age
+    FROM describe_plan_test.tbl
+    WHERE age > 25
+)
+SELECT
+    t.name,
+    t.age,
+    m.count as appearance_count
+FROM aged_users t
+LEFT JOIN describe_plan_test.complex_mv m ON t.name = m.name;
+
+# Test special characters in relation names
+statement ok
+CREATE TABLE describe_plan_test."table.with.dots" (id INT PRIMARY KEY, value VARCHAR);
+
+statement ok
+CREATE VIEW describe_plan_test."view-with-dashes" AS SELECT * FROM describe_plan_test."table.with.dots";
+
+
+# Test DESCRIBE PLAN for each relation
+query ?
+DESCRIBE PLAN describe_plan_test.tbl;
+----
+StreamMaterialize { columns: [id, name, age, created_at, data], stream_key: [id], pk_columns: [id], pk_conflict: Overwrite }
+└─StreamFilter { predicate: IsNotNull(name) }
+  └─StreamUnion { all: true }
+    └─StreamExchange { dist: HashShard(id) }
+      └─StreamDml { columns: [id, name, age, created_at, data] }
+        └─StreamSource
+
+
+# Test DESCRIBE PLAN for each relation
+query ?
+DESCRIBE PLAN (logical) describe_plan_test.tbl;
+----
+LogicalSource
+
+
+
+# Should show materialized view creation plan with complex aggregations
+query ?
+DESCRIBE PLAN describe_plan_test.complex_mv;
+----
+StreamMaterialize { columns: [name, count, avg_age, all_data], stream_key: [name], pk_columns: [name], pk_conflict: NoCheck }
+└─StreamProject { exprs: [tbl.name, count, (sum(tbl.age)::Decimal / count(tbl.age)::Decimal) as $expr1, jsonb_agg(tbl.data)] }
+  └─StreamFilter { predicate: (count > 1:Int32) }
+    └─StreamHashAgg { group_key: [tbl.name], aggs: [count, sum(tbl.age), count(tbl.age), jsonb_agg(tbl.data)] }
+      └─StreamExchange { dist: HashShard(tbl.name) }
+        └─StreamFilter { predicate: (tbl.age > 18:Int32) }
+          └─StreamTableScan { table: tbl, columns: [name, age, data, id] }
+
+
+
+# Should show materialized view creation plan with complex aggregations
+query ?
+DESCRIBE PLAN (logical, trace) describe_plan_test.complex_mv;
+----
+Begin:
+(empty)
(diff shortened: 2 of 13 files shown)


stdrc:
just to confirm, is this describe the existing plan or re-explain the sql?

stdrc:
> re-explain the sql

I'm a bit concern here, since this is not "describe".

xxchan:
What about changing that to REPLAN

tabVersion:
> What about changing that to REPLAN

It sounds like a modification to existing jobs.

xxchan:
Well, what about adding a warning string in each plan's output?

kwannoel:
> > re-explain the sql
> 
> I'm a bit concern here, since this is not "describe".

I think if we want to capture exactly the current physical plan, one way is to pretty print the fragment graph of the old stream job.

BugenZhao:
> What about changing that to REPLAN

What about `EXPLAIN SHOW CREATE MATERIALIZED VIEW`? 😆

kwannoel:
> > What about changing that to REPLAN
> 
> What about `EXPLAIN SHOW CREATE MATERIALIZED VIEW`? 😆

We already have `explain create materialized view` 😆 I think the extra `SHOW` keyword does not help 😅

BugenZhao:
> I think the extra `SHOW` keyword does not help 😅

Didn't quite get this. 👀

stdrc:
> > > What about changing that to REPLAN
> > 
> > 
> > What about `EXPLAIN SHOW CREATE MATERIALIZED VIEW`? 😆
> 
> We already have `explain create materialized view` 😆 I think the extra `SHOW` keyword does not help 😅

I suppose Bugen means `EXPLAIN SHOW CREATE MATERIALIZED VIEW mv;`? In which, `show` is to display the original definition of `mv`? Correct me if I'm wrong...

kwannoel:
> > > > What about changing that to REPLAN
> > > 
> > > 
> > > What about `EXPLAIN SHOW CREATE MATERIALIZED VIEW`? 😆
> > 
> > 
> > We already have `explain create materialized view` 😆 I think the extra `SHOW` keyword does not help 😅
> 
> I suppose Bugen means `EXPLAIN SHOW CREATE MATERIALIZED VIEW mv;`? In which, `show` is to display the original definition of `mv`? Correct me if I'm wrong...
 
Oh I see, `explain (show create materialzied view mv)`. Thanks

xxchan:
> EXPLAIN SHOW CREATE MATERIALIZED VIEW

Acceptable to me.

One benefit of `DESCRIBE` is that, we don't need to specify the relation type `MATERIALIZED VIEW` (as pointed out by @kwannoel ). 😌
... and it's shorter

> The scope should expand to all relations not just mvs, since our relations should have unique names.

BugenZhao:
> since our relations should have unique names.

Not sure if this is always true. IIRC creating table with connector will create both a table and a source with the same name.

xxchan:
> > since our relations should have unique names.
> 
> Not sure if this is always true. IIRC creating table with connector will create both a table and a source with the same name.

The source is not user accessible 🤔

kwannoel:
> > > re-explain the sql
> > 
> > 
> > I'm a bit concern here, since this is not "describe".
> 
> I think if we want to capture exactly the current physical plan, one way is to pretty print the fragment graph of the stream job.
> 
> Because we don't persist the logical or stream plans. I think persisting them is a lot of extra work so -1.
> 
> Pretty print the fragment graph has the advantage that we can support it in a backwards compatible way. Because jobs which are created have their fragment graphs already persisted.
> 
> I suppose print fragment graph may make more sense, we can identify specific fragments in the pretty printed output. So when debugging, we can reference parts of the fragment graph for a stream job.

Any thoughts on this suggestion?

xxchan:
I agree we can support `DESCRIBE FRAGMENT`. I also mentioned some thoughts here https://github.com/risingwavelabs/risingwave/issues/21254#issuecomment-2796560723

But it's already available on meta dashboard. Isn't that more useful than a text format? 🤔 


> reference parts of the fragment graph 

I don't get what does this mean :eyes: 


> > since our relations should have unique names.
> 
> Not sure if this is always true. IIRC creating table with connector will create both a table and a source with the same name.

Besides, we already have `DESCRIBE t`... So it shouldn't be a problem.

kwannoel:
> I agree we can support `DESCRIBE FRAGMENT`. I also mentioned some thoughts here [#21254 (comment)](https://github.com/risingwavelabs/risingwave/issues/21254#issuecomment-2796560723)
> 
> But it's already available on meta dashboard. Isn't that more useful than a text format? 🤔

Meta dashboard is not fully designed for end-user. More for RisingWave Developers. For instance, cloud users don't have access to it.

An SQL based interface is accessible by all our users.

> > reference parts of the fragment graph
> 
> I don't get what does this mean 👀

The graph might look something like (similar to distsql):
```
    Fragment 0
    StreamSink { type: upsert, columns: [v1, v2, v3, v5, t1.v4(hidden)], downstream_pk: [t1.v1, t1.v2] }
    ├── tables: [ Sink: 0 ]
    ├── output: [ t1.v1, t1.v2, t1.v3, t1.v5, t1.v4 ]
    ├── stream key: [ t1.v3, t1.v4 ]
    └── StreamExchange Hash([0, 1]) from 1
        ├── output: [ t1.v1, t1.v2, t1.v3, t1.v5, t1.v4 ]
        └── stream key: [ t1.v3, t1.v4 ]

    Fragment 1
    StreamTableScan { table: t1, columns: [t1.v1, t1.v2, t1.v3, t1.v5, t1.v4], stream_scan_type: ArrangementBackfill, stream_key: [t1.v3, t1.v4], pk: [v3, v4], dist: UpstreamHashShard(t1.v3, t1.v4) }
    ├── tables: [ StreamScan: 1 ]
    ├── output: [ t1.v1, t1.v2, t1.v3, t1.v5, t1.v4 ]
    ├── stream key: [ t1.v3, t1.v4 ]
    ├── Upstream { output: [ v1, v2, v3, v5, v4 ], stream key: [] }
    └── BatchPlanNode { output: [ v1, v2, v3, v5, v4 ], stream key: [] }
```

Maybe user has a log like: `sink_error{fragment_id:0}`

They can easily compare it using the ids of the stream graph, and see what that looks like.

They can also use `DESCRIBE FRAGMENT <job_id>`. But in cases like dirty agg + join amp, the event might surface at the agg fragment, so it's useful to view the whole graph.

xxchan:
> Meta dashboard is not fully designed for end-user. More for RisingWave Developers. For instance, cloud users don't have access to it.

I don't think PLAN/FRAGMENT explain is for users. It's mainly for dev troubleshooting perf.
Asking users to understand the PLAN sounds too difficult. 🤔

xxchan:
Hmm, actually fragment definition is already available in `select node from rw_fragments`. 🤔

kwannoel:
> > Meta dashboard is not fully designed for end-user. More for RisingWave Developers. For instance, cloud users don't have access to it.
> 
> I don't think PLAN/FRAGMENT explain is for users. It's mainly for dev troubleshooting perf. Asking users to understand the PLAN sounds too difficult. 🤔

I think not to understand the whole plan, but specific parts can be highlighted via our docs:
- predicate pushdown
- optimize over window -> top n
- join ordering

---

Even for dev troubleshooting, the meta dashboard is not a portable format. User can dump the `describe plan` output, we can store it on notion etc...

It also only gives a high-level view of things, rather than allow us to introspect things like the schema of join, the join type (left/right/inner, predicates, etc...).

Further, meta dashboard requires us to port-forward + dump the fragment graph, which can take a while.

kwannoel:
> Hmm, actually fragment definition is already available in `select jsonb_pretty(node) from rw_fragments`. 🤔

I think the `json` is more designed to be machine-readable. Is not really readable 😅 . Hence I'd prefer a pretty printer with options to toggle verbosity.

xxchan (commented):
Update:
- Added a note: `Note: The result above is a newly generated plan based on the same SQL statement, which might be different from the job's actual plan.`
- Temporarily Support `DESCRIBE FRAGMENT` based on `DESCRIBE PLAN (distsql)`. We can switch the implementation to a good one later.

kwannoel (approved):
Thanks!

stdrc:
I personally -2 for this. To me, the behavior of the syntax is unacceptably confusing. I believe a user may have the need to re-explain the query of an existing MV, but she/he won't like to do it in the form of `DESCRIBE PLAN mv`. If we have to do this, `EXPLAIN SHOW CREATE MATERIALIZED VIEW` looks more acceptable to me...

stdrc:
Considering this is a user-facing feature, especially, a new SQL syntax, I'd like to involve product guys to comment.

xxchan:
> Considering this is a user-facing feature

This don't need to be user-facing. Can hide from doc. And it can be removed/changed in the future. No stability guarantee at all. 

> If we have to do this, EXPLAIN SHOW CREATE MATERIALIZED VIEW looks more acceptable to me...

I don't think user can understand this syntax.

If we really want to discuss user's mindset, I think they care syntax for these needs at all. At least the NOTE should be more effective than the syntax difference. 

And making it longer will lose the purpose of introducing a syntax sugar.

kwannoel:
How about just support `describe fragments` first. I don't think there's any contention about that, and it should cover most functionality of `describe plan`. Readability can be tweaked by configuring the fragment pretty printer.

Its output will mirror `explain plan (distsql)`.

stdrc:
> How about just support `describe fragments` first. I don't think there's any contention about that, and it should cover all the functionality of `describe plan`. Readability can be tweaked by configuring the fragment pretty printer.
> 
> Its initial implementation will mirror `explain (distsql)`, and will just be a re-explain. But eventually it can fetch fragments from the meta store and render them.

I still don't understand. If it finally will fetch fragments from meta store, why not directly do it in that way. Is this a really urgent need? If yes, why? If no, why introducing mental overhead/tech debt for a non-urgent issue?

---

> > Considering this is a user-facing feature
> 
> This don't need to be user-facing. Can hide from doc. And it can be removed/changed in the future. No stability guarantee at all.
> 
> A practical plan is to remove this after #21432 is done
> 
> > If we have to do this, EXPLAIN SHOW CREATE MATERIALIZED VIEW looks more acceptable to me...
> 
> I don't think user can understand this syntax.

So I really want a clear claim that this is or is not a user-facing feature? If it's just for DX, then we'd better to make it some `INTERNAL` thing (interestingly even `SHOW INTERNAL TABLES` is expected to be used by our users).

> And making it longer will lose the purpose of introducing a syntax sugar.

What's the reasonale behind this? I mean, why is it reasonable to introduce a short syntax suger in our SQL syntax just for debugging purpose of only one or several of our developers? SQL syntax is not `risedev`... If it's reasonable, then everyone can introduce their only shortcut to do their own workflow. Do you think that's acceptable? What if I want to introduce `CREATE MV` for creating an MV, or even `CT t (foo int)` for creating a table? What's the difference between these kind of ridiculous syntax suger and the one in this PR?

stdrc:
> For me, I think `SHOW CREATE MATERIALIZED VIEW` + copy-paste SQL + `EXPLAIN SQL` is a typical pattern and it's troublesome. The trouble include typing MATERIALIZED VIEW without auto-completiton (unlike CREATE MATERIALIZED VIEW)

If this is really the case, I suggest `DEBUG EXPLAIN mv;`

xxchan:
DESCRIBE FRAGMENTS is done. Now you all can be happy

stdrc (approved):
looks great to me

kwannoel (approved):
LGTM, thanks for your work

e2e_test/batch/describe_fragments.slt
```diff
@@ -0,0 +1,241 @@
...
+    t.age,
+    m.count as appearance_count
+FROM aged_users t
+LEFT JOIN describe_plan_test.complex_mv m ON t.name = m.name;
+
+# Test DESCRIBE FRAGMENTS for each relation
+# Use psql_validate.py since the fragment id is not deterministic
+
+system ok
+psql_validate.py --sql "DESCRIBE FRAGMENTS describe_plan_test.tbl" \
```
BugenZhao, line 78: A bit sad to see system command being abused in this way. 😢 Perhaps we can introduce some sort of non-exact matching in SLT.

xxchan: Yeah. But it needs a bit design and iterate.. It's harder than I imagined and I spent >1h to debug yesterday. So prefer to keep it here to iterate faster.

xxchan: e.g., one difficulty is the error message when match failed..

xxchan: 🤔 Actually, I'm thinking stuff like PLAN is already not suitable for (original) SLT.

xxchan: In the meantime, maybe we don't want SLT any more https://github.com/risingwavelabs/risingwave/issues/20525#issuecomment-2811901092

Issue #21218, opened 2025-04-02
support per query query_mode


zwang28:
### Is your feature request related to a problem? Please describe.

Currently to customize the query_mode for a query, I need
1. `set query_mode to distributed;` -- temporarily set the query_mode for current connection
2. run the query, e.g. `select * from t;`
3. `set query_mode to default; `-- revert the query_mode for current connection

This is inconvenient especially when using a sql library. It would be handy if I could write something like
`select * from t with query_mode = distributed;`

Issue #17472, opened 2024-06-26
store "pure" SQL in the Source/Table's definition


st1page:
The original idea is proposed by @BugenZhao in https://github.com/risingwavelabs/risingwave/issues/9828 Initially, the intention was to simplify code complexity, but for some new fact such as CDC table's auto schema revolution, this matter has become more urgent.

## Background
Currently, we have two completely incompatible ways of determining the schema for tables/sources:

1. **Schema Defined by the CREATE Statement**: In this approach, the schema is defined by the user at the time of creating the table or source. Users can specify the columns and their types within the column definitions. In this case, the definition field in the catalog fully contains the SQL information for creating the table or source.

2. **Schema Derived from External Sources**: This method was initially designed for external connectors that use Protobuf or Avro for encoding. These encodings typically manage schema information externally. When creating a table or source, a URL or other retrieval method is provided, and RisingWave fetches and parses the schema from the external source. 

These two usage patterns are currently in conflict within the implementation. If you expect to use an external schema, specifying columns in the CREATE statement is not allowed, which means it's a pattern of "one or the other," not both.

- The "one or the other pattern" results in some limitations. For schema changes, if the schema is defined by the CREATE Statement, then its schema changes can only be made by statements like add column. If the schema is derived from external sources, then it can only be refreshed externally. This is reasonable for encodings like protobuf or Avro. However, it becomes unreasonable when applied to CDC tables.

- Even for encodings like protobuf or Avro, it is meaningful to declare some columns in the table definition. https://github.com/risingwavelabs/risingwave/issues/12199 But the current code structure and organization, which forces a choice between the two, make it difficult to implement. 

In the catalog for tables and sources, we have a field called definition that stores the SQL statement used to create the table or source. This field should always remain consistent with the source or table, so it needs to be updated accordingly during ALTER TABLE or ALTER SOURCE operations. Currently, it also serves as the source of truth for some information. For example, when executing CREATE SINK INTO TABLE, the SQL statement must be fully processed again to generate a plan. In ALTER SOURCE REFRESH SCHEMA, the table's format and encoding are obtained. And If the table/source's schema is derived externally, the catalog's definition only stores the original SQL without any sql definition.

## Design
In the design, if a table/source should be able to obtain a schema externally and, at the same time, specify some columns in the CREATE statement, and it can support a full set of features such as refresh schema, add/drop column, etc., such a table/source needs to meet these points:

[...]

Issue #11929, opened 2023-08-29
discussion: shall we introduce `WITH` properties for materialized views?


fuyufjh:
Shall we introduce a `WITH` clause for `create materialized view` statement?

We have `WITH` for tables, source, and sink to define some additional parameters. However, currently there is no such place for materialized views.

For example, `rate_limit` should be a per-stream property and should be able to be viewed and modified after MV creation. See also https://github.com/risingwavelabs/risingwave/pull/11919#issuecomment-1696731719. But this motivation seems not strong enough, because it's designed as a makeshift in case of creating MV failure.

kwannoel:
Minor improvement to UX: clearer semantics for session variables associated to MV.
Things like `RW_TWO_PHASE_AGG`, `STREAMING_PARALLELISM` can be tagged to specific MV, as per their semantics.

fuyufjh:
- `RW_TWO_PHASE_AGG` may be not the case? It is an optimizer hint and applies to batch query as well.
- `STREAMING_PARALLELISM`. Yeah, we can move the scale out/in commands here. Modifying the parallelism means scaling.

kwannoel:
> RW_TWO_PHASE_AGG may be not the case? It is an optimizer hint and applies to batch query as well.

My take is that as with global variables, setting session variables means the user may mistakenly use `two_phase_agg` with all queries in the session.

If they can scope it to a materialized view, it is less likely they make this mistake.

st1page:
maybe all session variable can be defined here for the query to overwrite the session varible such as timezone?

fuyufjh:
> maybe all session variable can be defined here for the query to overwrite the session varible such as timezone?

Good point. At least timezone sounds good to me.

Another random thought: how about capturing these session variables into the `WITH` clause automatically? It seems to be an elegant way to persist these variables (I don't know how it was done in current implementation). Additionally, when users run `show create materialized view`, the context can be printed as well.

BugenZhao:
Just FYI, we already have a context structure for each streaming job:

https://github.com/risingwavelabs/risingwave/blob/b823b0ce7b68f032c00e03d4cdda95bdd6a9f1b4/proto/stream_plan.proto#L723-L728

We can consolidate them in one place, such as storing all session-local variables here or in the `WITH` clause of the definition.

BugenZhao:
[...]

Issue #7491, opened 2023-01-19
Tracking: optimizer session variables and hints


kwannoel:
Session variables:
- [ ] `SET FORCE_DELTA_JOIN TO true`
- [ ] `SET FORCE_TWO_PHASE_AGG TO true`
- [ ] `SET DISABLE_TWO_PHASE_AGG TO true`
- [ ] `SET FORCE_LEFT_DEEP_JOIN TO true`
- [ ] `SET FORCE_BUSHY_JOIN TO true`

Optimizer hints:
- https://github.com/ossc-db/pg_hint_plan#hints-list
- https://nightlies.apache.org/flink/flink-docs-master/docs/dev/table/sql/queries/hints/#broadcast

chenzl25:
`DISABLE_TWO_PHASE_AGG` and `FORCE_TWO_PHASE_AGG` seem can't be both set to true. A priority needs to be introduced.

lmatz:
Suppose a query can utilize multiple indexes to accelerate itself, is it possible for the current optimizer to choose a suboptimal one?

I guess this can be supported only after hints are introduced if let's say this is necessary

PR #21356, opened 2025-04-11
feat(frontend): add settings keyword for runtime parameter

zwang28:

This PR adds the `SETTINGS` keyword for `Statement::Query`, to allow a query to overwrite its session runtime parameters.

The core changes are minor in terms of lines of code (LOC). Unfortunately I have to change all usage of `session.config().xxx()` to `running_sql_runtime_parameters(RuntimeParameters::xxx)` in order to expose `RuntimeParameters` inside `SessionImpl::running_sql_runtime_parameters` of `ArcSwap<RwLock>` type, which increases changes in LOC.

#21218

https://risingwave-labs.slack.com/archives/C034U0NH5ND/p1744175012756489

## Checklist

## Documentation

<details>

</details>


chenzl25:
Can we use the same syntax as pg_hint_plan? That is supporting SQL hint.
For example:
```
/*+ Set(query_mode local) */ SELECT * FROM table1 t1 WHERE key = 'value';
```

Once we have the SQL hint, we can extend it to support more features, e.g., changing the query plan.

 https://github.com/ossc-db/pg_hint_plan/blob/master/docs/hint_table.md#guc-parameters-set-during-planning

chenzl25:
Related issue https://github.com/risingwavelabs/risingwave/issues/7491

BugenZhao:
This somehow reminds me of...
- https://github.com/risingwavelabs/risingwave/issues/11929#issuecomment-1698433215
- https://github.com/risingwavelabs/risingwave/issues/17472
- https://github.com/risingwavelabs/risingwave/pull/21354

In short, we may consider persisting all session variables referenced by the optimizer into the job definition to achieve better reproducibility when replanning it.

src/sqlparser/src/keywords.rs
```diff
@@ -755,6 +758,7 @@ pub const RESERVED_FOR_COLUMN_OR_TABLE_NAME: &[Keyword] = &[
     Keyword::RIGHT,
     // Keyword::SELECT,
     // Keyword::SESSION_USER,
+    Keyword::SETTINGS,
```
xiangjinwu, line 761: **Reserving** any new keyword is a breaking change. If a user has existing mviews / tables / sources / columns named `settings`, their `definition` stored in meta catalog would become invalid.

In an old version, a user can:
```
create table t (settings struct<key varchar, value varchar>[]);
```

In the new version, if we do
```
alter table t add column foo int;
```

It would result in:
```

  1: unable to parse definition sql
  2: sql parser error: syntax error at or near settings
```

Because to create the same table, the correct SQL in new version is:
```
create table t ("settings" struct<key varchar, value varchar>[]);
```

e2e_test/batch/basic/runtime_parameter.slt.part
```diff
@@ -0,0 +1,37 @@
...
+----
+db error: ERROR: Failed to run the query
+
+Caused by these errors (recent errors listed first):
+  1: Catalog error
+  2: table or source not found: t
+
+
+statement ok
+EXPLAIN SELECT * FROM t SETTINGS search_path=schema_2 ,query_mode=auto;
```
BugenZhao, line 28: Shall we follow the syntax of `WITH` clause for better consistency? Kindly ask @xiangjinwu for insights.

src/sqlparser/src/keywords.rs
xiangjinwu: Just something on top of my mind: **one of** the the least intrusive ways to extend the `Query` syntax is probably the `WITH cte` clause at the beginning.

In addition to recognizing either the `RECURSIVE` keyword or a non-reserved identifier as cte_name, we could peek for another already reserved keyword (e.g. `USING` or other words that reads better) for this extended settings clause. For example:
```
WITH USING (search_path=schema_2 ,query_mode=auto)
WITH foo (select 1)
SELECT * FROM foo;
```
https://github.com/risingwavelabs/risingwave/blob/017748e455d8935cd2741ebc4ef443ee92766b16/src/sqlparser/src/parser.rs#L4419-L4426

**Alternatively**, @bugen is likely suggesting a trailing `WITH` for consistency with existing mview / sink syntax:
```
SELECT * FROM foo WITH (...);
CREATE MATERIALIZED VIEW foo WITH (...) AS query; -- existing syntax, also in PostgreSQL
CREATE SINK foo AS query WITH (...); -- existing syntax
```

However, *consistency* can also mean *ambiguity* here. If we add a trailing `WITH` as part of `Query`, it would result in:
```
CREATE MATERIALIZED VIEW foo WITH (..mview..) AS SELECT * FROM foo WITH (..query..);
CREATE SINK foo AS SELECT * FROM foo WITH (..query..) WITH (..sink..);
```

The latter cast is ambiguous when the WITH-of-query is omitted. The `parse_query` function may consume the `WITH (..sink..)` part mistakenly.

The former case, as well as the leading `WITH USING` syntax proposed above, reads a little bit unfriendly (which parameter should I write in which parentheses) but unambiguous - thanks to the `AS` keyword and existing support of WITH-cte.
```
CREATE MATERIALIZED VIEW foo WITH (...) AS WITH USING (...) SELECT * FROM foo;
CREATE SINK foo AS WITH USING (...) SELECT * FROM foo WITH (...);
```

Further iterations on this rough idea are welcomed @BugenZhao @chenzl25

chenzl25: How about using the PostgreSQl syntax for materialized view https://www.postgresql.org/docs/current/sql-creatematerializedview.html like:
```
 create materialized view mv with(xxx = yyy) as with cte as (select 1, 2, 3) select * from cte;
```

xiangjinwu: The materialized view syntax is already supported. The discussion here is about the syntax when it is not a materialized view - and whether it should look like the existing materialized view syntax

chenzl25: Considering no matter for batch queries or streaming queries, we want to capture the session variables/parameters, and we don't want to introduce a new keyword which would cause a breaking change, beside that using `with` keyword is a bit confused to users since `with` options are in the different position of our MV and sink syntax, furthermore, batch queries don't support `with` options. I think maybe we can just use `using session parameter (...)`  to capture the session parameters to declare the purpose precisely without ambiguity. 

Example:

```
select * from t using session parameter (xxx = yyy);
create materialized view v as select * from t using session parameter (xxx = yyy);
create materialized view mv as with cte as (select 1, 2, 3) select * from cte using session parameter (xxx = yyy);
```

chenzl25: I just realized that `SETTINGS` is a keyword of clickhouse, and they have a session setting syntax similar to this PR, so I think we can use `SETTINGS` as a keyword as well.
  https://clickhouse.com/docs/operations/settings/settings

PR #21360, opened 2025-04-11
fix: add under creation info when checking name duplicate in meta

### LINKED ISSUE #21336: During high barrier latency, create table takes a while to complete, and returns incorrect error messages
### Describe the bug

When triggering a create table under high barrier latency, it will notify the meta-node, and its catalog gets committed.

In another session, when attempting to recreate the same table, we will see the following error:
```
  1: Catalog error
  2: table with name test exists
```

This appears to mean that the table already created, which is not quite right. Users can't query it.

We should instead disable a different error message, like:
```
  1: Catalog error
  2: table with name test is being created
```

### Error message/log

```text

```

### To Reproduce

First, trigger high barrier latency in one session:

```
create table t(v1 int);
insert into t select 1 from generate_series(1, 10000);
create materialized view m1 as select x.v1 from t x join t y using(v1);
```

Second, create a table in a second session:
```
create table test(v1 int);
```

Third, try the following in another session:
```
create table test(v1 int);

error message:
  1: Catalog error
  2: table with name test exists
```

Fourth, attempt to drop the "existing table":
```
drop table test;

  1: Catalog error
  2: table not found: test
```

### Expected behavior

_No response_

_No response_

### The version of RisingWave

v2.2

### Additional context

_No response_

yezizp2012:

Fixed #21336 . After this PR, the error message for the case in #21336 would be as follows:
```
  1: Catalog error
  3: table with name test exists but under creation
```

Error message is consistent with FE: 
https://github.com/risingwavelabs/risingwave/blob/4847896a5a564ea5f9b06a0401b864b29033da1a/src/frontend/src/catalog/mod.rs#L109-L117

## Checklist

## Documentation

<details>

</details>


kwannoel (approved):
Could we add an slt test corresponding to the error in the linked issue?

kwannoel:
Thanks for the fix!

yezizp2012:
> Could we add an slt test corresponding to the error in the linked issue?

An slt test might not be suitable in CI. Let me add a simulation test for it.

wenym1 (approved):

e2e_test/slow_tests/ddl/creation_check.slt
```diff
@@ -0,0 +1,22 @@
+statement ok
+create table t(v1 int);
+
+statement ok
+insert into t select 1 from generate_series(1, 10000);
+
+# trigger high barrier latency
+system ok
```
BugenZhao, line 8: Instead of using system commands, can we adopt `connection <tag>` to achieve this?

src/meta/src/controller/utils.rs
```diff
@@ -513,7 +524,7 @@ where
         .await?;
     if count > 0 {
         assert_eq!(count, 1);
-        return Err(MetaError::catalog_duplicated("user", name));
+        return Err(MetaError::catalog_duplicated("user", name, false));
```
BugenZhao, line 527: Personally I would suggest making it new variant of error for better readability.

BugenZhao (commented):
Rest LGTM

e2e_test/slow_tests/ddl/creation_check.slt
yezizp2012: Unfortunately no, in scenarios with high barrier latency, the initial creation of table `t2` will be blocked. We have to spawn and make it running in background.

Issue #18601, opened 2024-09-19
improve `json_populate_record/json_populate_recordset` to accept `null::<table>` as the first parameter


lmatz:
PG:
```
CREATE TABLE IF NOT EXISTS t (host text, message text, payload jsonb);


SELECT * FROM jsonb_populate_record(null::t, '{"host": "abc", "message": "bcd", "payload": {"d": 4, "e": "zzz"}}');
```

RW:
```
dev=> SELECT * FROM jsonb_populate_record(null::t, '{"host": "abc", "message": "bcd", "payload": {"d": 4, "e": "zzz"}}');
ERROR:  Failed to run the query

Caused by:
  Feature is not yet implemented: unsupported data type: t
No tracking issue yet. Feel free to submit a feature request at https://github.com/risingwavelabs/risingwave/issues/new?labels=type%2Ffeature&template=feature_request.yml
```

RW can use `struct<>` to replace `t`, but in the case of integration, it is required to support using table to indicate the row type:
https://github.com/vectordotdev/vector/pull/21248/files#diff-e940b31386309d22cca11fff761d2467ff4a590c8110378b7a6d2a8747859031R123

fuyufjh:
+1

Even the requirement is from intergration of https://github.com/vectordotdev/vector, It looks like a general requirement to me - converting JSON records to rows of table.

lmatz:
https://github.com/postgres/postgres/blob/master/src/test/regress/sql/json.sql#L456-L624

xiangjinwu:
After some attempts we realized the issue is NOT as trivial as replacing table name with a struct of table columns:
* PostgreSQL does NOT allow it to be a CTE name: `with k as (select 2) select null::k;`
  * When both CTE and table named `k` exists, it will skip CTE and bind to table - this means adding/removing support for CTE is a breaking change.
* PostgreSQL allows it to be a (non-materialized) view, but in RisingWave views are re-bind every time being referenced. We may skip supporting views until our view support is no longer hacky.
* What happens to `expr::t` when we `ALTER TABLE t`? In other words, do we consider this usage as a downstream dependency of `t` and reject such `ALTER`s if columns are dropped? ([PG](https://gist.github.com/xiangjinwu/ed8ebc6a4db957a2ce7b3e758698fc8b))

andrescaroc:
@lmatz Are you still working on this? It seems you are so close to achieve the integration with Vector.dev and @jorgehermo9
seems to be willing to help.

[...]

PR #21375, opened 2025-04-14
Support josnb_populate_record cast to table.

lin0303-siyuan:

- Change: cast custom datatype to table in `bind_cast`, if it is a table, instead of throwing error, returning a struct containing the type and field.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/binder/expr/mod.rs
+++ b/src/frontend/src/binder/expr/mod.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use iceberg::spec::Struct;
 use itertools::Itertools;
 use risingwave_common::catalog::PG_CATALOG_SCHEMA_NAME;
 use risingwave_common::types::{DataType, MapType, StructType};
@@ -19,11 +20,11 @@ use risingwave_common::util::iter_util::zip_eq_fast;
 use risingwave_common::{bail_no_function, bail_not_implemented, not_implemented};
 use risingwave_sqlparser::ast::{
     Array, BinaryOperator, DataType as AstDataType, EscapeChar, Expr, Function, JsonPredicateType,
-    ObjectName, Query, TrimWhereField, UnaryOperator,
+    ObjectName, Query, TableFactor, TrimWhereField, UnaryOperator,
 };
 
-use crate::binder::Binder;
 use crate::binder::expr::function::is_sys_function_without_args;
+use crate::binder::{Binder, BoundBaseTable};
 use crate::error::{ErrorCode, Result, RwError};
 use crate::expr::{Expr as _, ExprImpl, ExprType, FunctionCall, InputRef, Parameter, SubqueryKind};
 use crate::handler::create_sql_function::SQL_UDF_PATTERN;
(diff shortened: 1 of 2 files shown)


src/frontend/src/binder/expr/mod.rs
```diff
@@ -12,18 +12,19 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use iceberg::spec::Struct;
```
xiangjinwu, line 15: How is this related to iceberg?

src/frontend/src/binder/expr/mod.rs
```diff
@@ -950,6 +951,57 @@ impl Binder {
...
+                    Err(err) => {
+                        // Try to bind the table.
+                        let idents = qualified_type_name
+                            .0
+                            .iter()
+                            .map(|n| n.real_value())
+                            .collect_vec();
+                        let table_name = idents[0].as_str();
+                        use risingwave_sqlparser::ast::Ident;
+                        let ident = Ident::new_unchecked(table_name);
```
xiangjinwu, line 967: Why are we extracting `real_value` from `ast::{ObjectName,Ident}`, and then wrapping it back into a `ast::Ident` (with bugs)? Both `qualified_type_name` and `qualified_relation_name` are `ObjectName`s.

src/frontend/src/binder/expr/mod.rs
```diff
@@ -950,6 +951,57 @@ impl Binder {
...
+                        let table_factor = TableFactor::Table {
+                            name: ObjectName(vec![ident]),
+                            alias: None,
+                            as_of: None,
+                        };
+                        let relation = self.bind_table_factor(table_factor);
```
xiangjinwu, line 973: Unfortunately `self.bind_table_factor` cannot be used here to **lookup** a table by possibly schema-qualified name from catalog. It has a side effect (`bind_table_to_context`) that populates the current binding context:
```
dev=> create table t (x int, y int);
CREATE_TABLE

dev=> select row(2, 3)::t;
  row  
-------
 (2,3)
(1 row)

dev=> select row(2, 3)::t, x; -- x is column 0 of t
ERROR:  Panicked when handling the request: insert at index 0 exceeds fixbitset size 0
```

Expect behavior: column `x` not found because table `t` is not in current binding context.

The proper implementation shall refactor `bind_relation_by_name_inner` and split it into 2 parts: first part is the pure lookup that can be reused here, and the second part populates binding context.

src/frontend/src/binder/expr/mod.rs
```diff
@@ -950,6 +951,57 @@ impl Binder {
...
+                                    return Err(err);
+                                }
+                            },
+                            _ => {
+                                return Err(err);
+                            }
+                        }
+                    }
+                }
+            }
```
xiangjinwu, line 1004: Avoid deep indentation by handling simple cases with early return

src/frontend/src/binder/expr/mod.rs
lin0303-siyuan: My bad, maybe the editor generates automatically for me.

src/frontend/src/binder/expr/mod.rs
lin0303-siyuan: Am I supposed to do this refactor, or someone is taking care of this? Or is there other workaround, for example, first create a temp binding context, and then discard it.

src/frontend/src/binder/expr/mod.rs
lin0303-siyuan: This part is actually copied from codes in `bind_data_type` that will deal with `AstDataType::Custom`.

xiangjinwu: Yes the `real_value` part is same as `bind_data_type` but it shall not be wrapped back to `Ident` with `new_unchecked`. Refer to the discussion below on how to lookup the table given a `ObjectName`

src/frontend/src/binder/expr/mod.rs
xiangjinwu: If the refactor is easy to achieve, do it here. The workaround is also possible but is not preferred when we can do it the right way. Assume there is another place later wants this behavior, and the developer can fall into the same issue as we did here - forgetting to create a temp binding context.

scrogson:
Just curious about any future progress on this PR? It looks like it's been stalled for a bit. Support for `json_populate_recordset` is something I'm hoping to use in the short term if possible.

src/frontend/src/binder/expr/mod.rs
xiangjinwu: Seems non-trivial: https://github.com/risingwavelabs/risingwave/issues/18601#issuecomment-2961440504

Besides `bind_relation_by_name_inner`, the function `resolve_table_relation` also has a side effect of tracking dependencies via `included_relations` - we need to clarify all these before proceeding, rather than "delivery whichever easy to implement".

hzxa21:
Hi @lin0303-siyuan, there are a few unaddressed comments. Are you still working on this PR?

xiangjinwu:
> Hi @lin0303-siyuan, there are a few unaddressed comments. Are you still working on this PR?

There are some undefined behaviors in https://github.com/risingwavelabs/risingwave/issues/18601#issuecomment-2961440504 and matching PostgreSQL is not trivial.

PR #21380, opened 2025-04-14
feat(common): add a new dummy data type

xiangjinwu:

Adding a new builtin data type into the system affects a lot of areas. This PR is the minimal boilerplate required to make the code compile. It does not implement any concrete semantic of the data type and just use `todo!("VECTOR_PLACEHOLDER")`. With a few exceptions:
* The **code** name of this data type is called `vector`.
* It can NOT be the key data type of a `map` data type. In other words `map(vector(_), _)` is invalid.
* It is **displayed** as `vector(n)`, and uses this form to parse testing chunk literal (`DataChunk::from_pretty`).
* In `pg_catalog.pg_type`, its `typlen` is `-1` (`varlena`) and **`typname`** is `vector`.
* In our expr macros, the **pattern** `vector` (similar to `anyarray`) will match the type regardless of parameter value.
* `to_jsonb` will not treat it specially and just fallback to text format.

Notably:
* It is parameterized by a required positive integer. Making it neither `simple!` nor `composite!`.
* Unlike `simple!` data types:
  * `ArrayBuilder::new` can NOT be used
  * it can NOT be inferred by expr signatures like `cast(varchar) -> vector`
  * `lhs.typename == rhs.typename` does NOT imply `lhs == rhs`
* Unlike `composite!` data types:
  * there is no nested data type to alter or assign a column id

One may think `vector` is composite as it will contain `f32`, but we never need to recursively pass a calling function to its `f32` elements. SimIlarly `int256` is NOT composite of multiple `i128` or `i64`s; `interval`/`date`/`time` is NOT composite of integral months or seconds.

## Checklist

## Documentation

<details>

</details>


src/expr/impl/src/scalar/to_jsonb.rs
```diff
@@ -73,6 +73,7 @@ impl ToJsonb for ScalarRefImpl<'_> {
             Struct(v) => v.add_to(ty, builder),
             List(v) => v.add_to(ty, builder),
             Map(v) => v.add_to(ty, builder),
+            Vector(v) => v.add_to(ty, builder),
```
xiangjinwu, line 76: * #22030

src/tests/sqlsmith/src/sql_gen/types.rs
```diff
@@ -53,6 +53,7 @@ pub(super) fn data_type_to_ast_data_type(data_type: &DataType) -> AstDataType {
                 .collect(),
         ),
         DataType::List(typ) => AstDataType::Array(Box::new(data_type_to_ast_data_type(typ))),
+        DataType::Vector(_) => todo!("VECTOR_PLACEHOLDER"),
```
xiangjinwu, line 56: * #22032

PR #21389, opened 2025-04-14
feat(compactor): optimize compactor pull policy

Li0k:

A fixed algorithm based on the **number of CPU cores** may lead to the situation where CPU resources cannot be fully utilized in an IO-bound task scenario. On the other hand, directly estimating with CPU usage is likely to cause an overload. Therefore, a new idea is introduced, which allows for boosting when the CPU usage is low.

- Refactor pull task logic with an encapsulated structure
- A new policy that allows for boosting when the CPU usage is low.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12145,6 +12145,7 @@ dependencies = [
  "sled",
  "spin 0.10.0",
  "sync-point",
+ "sysinfo",
  "tempfile",
  "thiserror 1.0.63",
  "thiserror-ext",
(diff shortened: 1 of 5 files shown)


stdrc (approved):
approve on behalf of cargo.lock

zwang28 (approved):

hzxa21:
Do we still need to merge this PR or we leave it as it is and merge it until we do meet the issue?

PR #21390, opened 2025-04-14
chore(udf): control all udf languages/runtimes by a single feature/env var

stdrc:

This PR removes `ENABLE_EXTERNAL_UDF`, `ENABLE_PYTHON_UDF`, `ENABLE_JS_UDF` and `ENABLE_WASM_UDF` environment variables that were used to enable each UDF language feature, instead, replacing them with a single `ENABLE_UDF` env var (and `risedev configure` item). This can simplify the dev and testing process of UDF feature.

## Checklist

## Documentation

<details>

</details>


stdrc:
* **#21390** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/21390?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/21390?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/expr/impl/src/udf/mod.rs
```diff
@@ -22,14 +22,14 @@ use risingwave_expr::sig::{
...
 mod external;
-#[cfg(feature = "python-udf")]
+#[cfg(feature = "udf")]
 mod python;
-#[cfg(feature = "js-udf")]
+#[cfg(feature = "udf")]
 mod quickjs;
-#[cfg(feature = "wasm-udf")]
+#[cfg(feature = "udf")]
 mod wasm;
```
xxchan, line 33: Do you mean we want to lift the support to all UDFs to the same level? Otherwise maybe separate feature flags are still good?

stdrc: What do you mean by "lift the support to all UDFs to the same level"? IIUC currently we always build all UDF languages, I don't think we may have some day only build JS UDF support or something like that.

xxchan: not for production, but may only want to develop some of the udfs. (reduce compile time)

xxchan: I think it's the same rationale for adding flags for each connectors

stdrc: That's exactly my motivation. We have plenty of optional modules that are not related to the development of other modules. I don't think it's that necessary to split UDF feature into parts. From my experience, when I want to develop UDF related things, I always want to enable all the features.

PR #21399, opened 2025-04-14
feat(storage): implement in memory FLAT vector index

wenym1:

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/storage/src/vector/mod.rs
+++ b/src/storage/src/vector/mod.rs
@@ -12,7 +12,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+pub mod distance;
+pub use distance::DistanceMeasurement;
+
+pub mod utils;
+
 use std::sync::Arc;
+
+use crate::vector::utils::BoundedNearest;
 
 pub type VectorItem = f32;
 #[derive(Clone, Copy, Debug, PartialEq)]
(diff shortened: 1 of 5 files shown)


wenym1:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/21399?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#21399** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/21399?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/21399?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#21398** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/21398?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#21397** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/21397?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

hzxa21 (approved):
LGTM

src/storage/src/vector/mod.rs
```diff
@@ -0,0 +1,537 @@
...
+    }
+}
+
+impl<'a> MeasureDistance<'a> for L1DistanceMeasure<'a> {
+    fn measure(&self, other: VectorRef<'_>) -> VectorDistance {
+        // TODO: use some library with simd support
+        let len = self.0.0.len();
+        assert_eq!(len, other.0.len());
+        // In this implementation, we don't take the square root to avoid unnecessary computation, because
+        // we only want comparison rather than the actual distance.
```
xxhZs, line 168: It seems that for L1, even the final result doesn’t require taking the square root.

src/storage/src/vector/mod.rs
```diff
@@ -0,0 +1,537 @@
...
+}
+
+impl<'a> MeasureDistance<'a> for CosineDistanceMeasure<'a> {
+    fn measure(&self, other: VectorRef<'_>) -> VectorDistance {
+        // TODO: use some library with simd support
+        let len = self.target.0.len();
+        assert_eq!(len, other.0.len());
+        let magnitude_mul = other.magnitude() * self.magnitude;
+        (0..len)
+            .map(|i| self.target.0[i] * other.0[i] / magnitude_mul)
```
xxhZs, line 252: Cosine Distance appears to be 1 - xxx. Could this reverse the sorting results here?

wenym1: Seems not. Cosine distance is the cosine value on the angle of two vectors. You may refer to https://en.wikipedia.org/wiki/Cosine_similarity

When the two vectors are normalized, the cosine simarility is equal to inner product of them.

wenym1: Oh wait. Indeed, when two vector is the same, their cosine similarity is 1, but they should be at the nearest distance. It's true that we should use the `1-similarity` as the distance. Thanks for pointing it out.

PR #21400, opened 2025-04-14
feat(iceberg): support iceberg jdbc catalog views

chenzl25:

- Support `rw_catalog.iceberg_tables` and `rw_catalog.iceberg_namespace_properties` catalog views. `rw_catalog.iceberg_tables` will redirect the query to the meta store table `iceberg_tables` and it will check the user privilege to determine whether to show iceberg table records.
- Users can use other query engines they like to connect to the RisingWave SQL catalog.

## Checklist

## Documentation

<details>

- Support `iceberg_tables` and `iceberg_namespace_properties` system view. The goal is to use RisingWave as a JDBC catalog for other query engines to query the hosted iceberg tables.

</details>

### DIFF
--- a/src/frontend/src/catalog/system_catalog/rw_catalog/iceberg_tables.rs
+++ b/src/frontend/src/catalog/system_catalog/rw_catalog/iceberg_tables.rs
@@ -1,4 +1,4 @@
-// Copyright 2024 RisingWave Labs
+// Copyright 2025 RisingWave Labs
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -87,7 +87,6 @@ async fn read(reader: &SysCatalogReaderImpl) -> Result<Vec<IcebergTables>> {
         meta_store_host, meta_store_port, meta_store_user, meta_store_password, meta_store_database
     );
 
-    println!("conn_str: {}", conn_str);
     let (client, conn) = tokio_postgres::connect(&conn_str, tokio_postgres::NoTls)
         .await
         .map_err(|e| anyhow!(e))?;
--- a/src/frontend/src/catalog/system_catalog/rw_catalog/iceberg_tables.rs
+++ b/src/frontend/src/catalog/system_catalog/rw_catalog/iceberg_tables.rs
@@ -1,0 +1,138 @@
+// Copyright 2024 RisingWave Labs
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use anyhow::{anyhow, Context};
+use risingwave_batch::executor::postgres_row_to_owned_row;
+use risingwave_common::acl::AclMode;
+use risingwave_common::bail;
+use risingwave_common::catalog::{Field, Schema};
+use risingwave_common::types::DataType::Varchar;
+use risingwave_common::types::Fields;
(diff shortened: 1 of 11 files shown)


src/frontend/src/catalog/system_catalog/rw_catalog/iceberg_tables.rs
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    let conn_str = format!(
+        "host={} port={} user={} password={} dbname={}",
+        meta_store_host, meta_store_port, meta_store_user, meta_store_password, meta_store_database
+    );
+
+    let (client, conn) = tokio_postgres::connect(&conn_str, tokio_postgres::NoTls)
+        .await
+        .map_err(|e| anyhow!(e))?;
```
xxchan, line 92: Hmmm, this looks concerning. Does frontend pod have access to the meta backend RDS? (should it have?)

xxchan: Maybe having a meta RPC to do the query is preferred.

src/frontend/src/catalog/system_catalog/rw_catalog/iceberg_tables.rs
```diff
@@ -0,0 +1,155 @@
...
+struct IcebergTables {
+    pub catalog_name: String,
+    pub table_namespace: String,
+    pub table_name: String,
+    pub metadata_location: Option<String>,
+    pub previous_metadata_location: Option<String>,
+    pub iceberg_type: Option<String>,
+}
+
+#[system_catalog(table, "rw_catalog.iceberg_tables")]
```
xxchan, line 48: `iceberg_tables` sound a bit general, but it's only for hosted catalog (unlike `rw_iceberg_files` which is for all sources).

What about using a new system catalog (e.g., `hosted_iceberg_catalog`?)

chenzl25: For JDBC, not every database has schema, so we need to create `iceberg_tables` at the default schema level so that, other query engines can access this view. Therefore, we should keep it in `rw_catalog`.

chenzl25:
> The goal is to use RW as a JDBC catalog for other query engines to query the hosted tables, right? 🤔

Exactly

src/frontend/src/catalog/system_catalog/rw_catalog/iceberg_tables.rs
xxchan: Why can't we make `hosted_iceberg_catalog` also at the default schema level?

xxchan: Oh, since users will be able to select without schema directly, so there's no too much difference whether it's in `rw_catalog` or  `hosted_iceberg_catalog ` 🤣

chenzl25: Just confirm with @yezizp2012 that we need to add a migration script to add the newly created schema to all existing created databases, which seems a bit overkill.

src/frontend/src/catalog/system_catalog/rw_catalog/iceberg_tables.rs
chenzl25: I just refactor this part and use a meta RPC instead

src/frontend/src/meta_client.rs
```diff
@@ -134,6 +135,8 @@ pub trait FrontendMetaClient: Send + Sync {
     async fn list_rate_limits(&self) -> Result<Vec<RateLimitInfo>>;
 
     async fn get_meta_store_endpoint(&self) -> Result<String>;
+
+    async fn list_iceberg_tables(&self) -> Result<Vec<IcebergTable>>;
```
xxchan, line 139: nitpicking:
```suggestion
    async fn list_hosted_iceberg_tables(&self) -> Result<Vec<IcebergTable>>;
```
might be more clear

In meta part it's put in `HostedIcebergCatalogService`, but seems in meta_client here, we don't have that context

xxchan (approved):
LGTM

PR #21407, opened 2025-04-15
feat: check SecretManagement available when enforce secret

tabVersion:

relate to https://github.com/risingwavelabs/risingwave/pull/21248

and make the enforcement default to `true` in examples

We are enforcing `enforce_secret_on_cloud = true` for new clusters for on-prem/byoc/cloud hosted, the variable default to `false` for backward compatibility, cc @lmatz and @kwannoel 

## Checklist

## Documentation

<details>

</details>


src/connector/src/enforce_secret_on_cloud.rs
```diff
@@ -27,9 +26,6 @@ pub trait EnforceSecretOnCloud {
     const ENFORCE_SECRET_PROPERTIES_ON_CLOUD: Set<&'static str> = phf_set! {};
 
     fn enforce_secret_on_cloud<'a>(prop_iter: impl Iterator<Item = &'a str>) -> Result<()> {
-        if !Deployment::is_cloud_hosted() {
-            return Ok(());
-        }
```
xxchan, line 32: Why to remove this check? (Do you mean to remove the `on_cloud` part from `enforce_secret_on_cloud`?)

tabVersion:
> I see, so rather than rely in the env var, since that only applies to cloud, we only check for license instead, since that applies to cloud, on prem and byoc

Yes

kwannoel (approved):

PR #21416, opened 2025-04-15
fix(source): handling the case of reading empty files

wcy-fdu:

Since v2.3, the logic of determining whether a file has been read and deleted from the state table becomes to determine whether the reader's stream has ended. If it has ended, a None will be inserted. The fetch executor will delete the corresponding file after encountering None.

However, if a file is empty, it will also be None, and its file name cannot be extracted from the chunk, so an empty string will be deleted. This pr fixes this problem by initializing the delete file name when handling list stream.

## Checklist

## Documentation

<details>

</details>


wenym1 (approved):
LGTM!

PR #21429, opened 2025-04-16
feat(lru): add `remove` method to LruCache and ManagedLruCache

stdrc:

Our LRU cache doesn't have `remove` method before, and that caused some problem because developer may forget to remove entries from it when deleting a corresponding row from state table.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/common/src/lru.rs
+++ b/src/common/src/lru.rs
@@ -176,6 +176,39 @@ where
         }
     }
 
+    pub fn remove(&mut self, key: &K) -> Option<V> {
+        unsafe {
+            let hash = self.hash_builder.hash_one(key);
+
+            if let Some(ptr) = self.map.find(hash, |p| p.as_ref().key() == key) {
+                let ptr = *ptr;
+
+                // Detach the entry from the LRU list
+                self.detach(ptr);
+
+                // Remove entry from the hash table
+                match self
+                    .map
+                    .entry(hash, |p| p.as_ref().key() == key, |p| p.as_ref().hash)
+                {
+                    Entry::Occupied(o) => {
+                        o.remove();
+                    }
+                    Entry::Vacant(_) => {}
+                }
+
+                // Extract entry from the box and return its value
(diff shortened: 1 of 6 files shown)


stdrc:
* **#21429** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/21429?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/21429?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/common/src/lru.rs
```diff
@@ -176,6 +176,39 @@ where
...
+                // Remove entry from the hash table
+                match self
+                    .map
+                    .entry(hash, |p| p.as_ref().key() == key, |p| p.as_ref().hash)
+                {
+                    Entry::Occupied(o) => {
+                        o.remove();
+                    }
+                    Entry::Vacant(_) => {}
+                }
```
MrCroxx, line 198: Is it possible to replace `entry()` and `find()` with `entry()` only here?

stdrc: Good catch. Done.

MrCroxx (approved):
LGTM

src/common/src/lru.rs
```diff
@@ -323,7 +352,7 @@ where
         }
     }
 
-    fn detach(&mut self, mut ptr: NonNull<LruEntry<K, V>>) {
+    fn detach(mut ptr: NonNull<LruEntry<K, V>>) {
```
BugenZhao, line 355: I'm not entirely sure but previously `&mut self` ensured that caller has a mutable reference to the cache. Will removing it somehow affects optimization and leads to unsoundness?

stdrc: During all the calls to this method, the caller `self` is held as a mut ref, so I think nothing is changed.

PR #21431, opened 2025-04-16
feat(sink): Support more template format for Redis sink

xxhZs:

Previously, our Redis template format {} did not support nested , and this PR support it.
## Checklist

For  template format ,  we can use the \{ or \} method to represent {} the original meaning .
 example : 
 v1 = 1 ;  v2 = 1;
 "{v1}xxx{v2}"  => 1xxx1
"{{v1}{v2}}" => error
"\{{v1}sad{v2}\}" => {1sad1}


wenym1 (approved):

src/connector/src/sink/encoder/template.rs
```diff
@@ -419,3 +434,213 @@ impl<T: SerTo<Vec<u8>>> SerTo<RedisSinkPayloadWriterInput> for T {
...
+            );
+        }
+
+        // Test invalid templates
+        let invalid_templates = vec![
+            "user:no_braces",        // No braces
+            "user:{invalid_column}", // Non-existent column
+            "user:{id",              // Unclosed brace
+            "user:id}",              // Unopened brace
+            "sadsadsad{}qw4e2ewq21", // Empty braces
```
xiangjinwu, line 504: Is `foo{id}bar{name` considered valid or invalid?
* if it is valid, the current description is wrong, as the unpaired `{` does not need escaping - only paired braces require escaping with backslash
* If it is invalid, the current implementation is wrongly accepting it - outputting `foo123bar{`

What about partially-escaped as in `foo\{id}bar` or `foo{id\}bar`?

As the template syntax gets more complicated, I would suggest finding an existing template library, rather than reinventing with our own buggy syntax and implementation. It maybe a breaking change in templating syntax, so better done earlier before it's too late.

PR #21439, opened 2025-04-16
refactor(frontend): reduce memory usage of `LogicalJoin::to_stream`

xxchan:
related: https://github.com/risingwavelabs/risingwave/issues/21355#issuecomment-2808083547

by returning PlanRef in function calls directly. Otherwise, `LogicalJoin::to_stream` needs to allocate stack memory for these temporary variables.

`LogicalJoin::to_stream` is somehow an important path (especially in deep plan), so it may worth the micro optimization

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/optimizer/plan_node/logical_join.rs
+++ b/src/frontend/src/optimizer/plan_node/logical_join.rs
@@ -925,6 +925,7 @@ impl LogicalJoin {
         Ok((left, right))
     }
 
+    #[inline(never)]
     fn to_stream_hash_join(
         &self,
         predicate: EqJoinPredicate,
@@ -999,24 +1000,26 @@ impl LogicalJoin {
         }
     }
 
+    #[inline(never)]
     fn to_stream_temporal_join_with_index_selection(
         &self,
         predicate: EqJoinPredicate,
         ctx: &mut ToStreamContext,
-    ) -> Result<StreamTemporalJoin> {
+    ) -> Result<PlanRef> {
         // Index selection for temporal join.
         let right = self.right();
         // `should_be_temporal_join()` has already check right input for us.
         let logical_scan: &LogicalScan = right.as_logical_scan().unwrap();
 
         // Use primary table.
-        let mut result_plan = self.to_stream_temporal_join(predicate.clone(), ctx);
+        let mut result_plan: Result<StreamTemporalJoin> =
+            self.to_stream_temporal_join(predicate.clone(), ctx);
         // Return directly if this temporal join can match the pk of its right table.
         if let Ok(temporal_join) = &result_plan
             && temporal_join.eq_join_predicate().eq_indexes().len()
                 == logical_scan.primary_key().len()
         {
-            return result_plan;
+            return result_plan.map(|x| x.into());
         }
         let indexes = logical_scan.indexes();
         for index in indexes {
@@ -1040,7 +1043,7 @@ impl LogicalJoin {
             }
         }
 
-        result_plan
+        result_plan.map(|x| x.into())
     }
 
     fn check_temporal_rhs(right: &PlanRef) -> Result<&LogicalScan> {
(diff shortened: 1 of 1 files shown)


xxchan:
I'm not 100% sure though 🤡

src/frontend/src/optimizer/plan_node/logical_join.rs
```diff
@@ -999,24 +1000,26 @@ impl LogicalJoin {
...
+    ) -> Result<PlanRef> {
         // Index selection for temporal join.
         let right = self.right();
         // `should_be_temporal_join()` has already check right input for us.
         let logical_scan: &LogicalScan = right.as_logical_scan().unwrap();
 
         // Use primary table.
-        let mut result_plan = self.to_stream_temporal_join(predicate.clone(), ctx);
+        let mut result_plan: Result<StreamTemporalJoin> =
+            self.to_stream_temporal_join(predicate.clone(), ctx);
```
stdrc, line 1016: I guess there's no difference to return a `PlanRef` or `StreamTemporalJoin`? 😄 IIUC, when appropriete optimization is on, the returned `StreamTemporalJoin` should be allocated in the frame of the caller.

src/frontend/src/optimizer/plan_node/logical_join.rs
```diff
@@ -925,6 +925,7 @@ impl LogicalJoin {
         Ok((left, right))
     }
 
+    #[inline(never)]
```
stdrc, line 928: What's the benefit of this line?

src/frontend/src/optimizer/plan_node/logical_join.rs
xxchan: I though the same, but see https://github.com/risingwavelabs/risingwave/issues/21355#issuecomment-2808083547 , the stack size in release mode seems to be contrary to this assumption.

I also checked the assembly from `cargo asm`, after this change, the stack size indeed reduced a few hundred bytes...

src/frontend/src/optimizer/plan_node/logical_join.rs
xxchan: If these functions are inlined into the `to_stream` call, the local variables used in these functions will be counted to `to_stream`'s frame, causing larger reserved stack size. 🤔

src/frontend/src/optimizer/plan_node/logical_join.rs
stdrc: interesting, good to know that

src/frontend/src/optimizer/plan_node/logical_join.rs
stdrc: But the total(maximum) stack size used when running this code path is not affected?

xxchan: Oh, you are right. So basically this only changes stack size from `max(calls)` to `min(calls)` 🤔

xxchan: It's just like `large_enum_variant` optimization. But I checked and found these functions' size do not differ too much. So maybe not worth it.

stdrc (approved):

src/frontend/src/optimizer/plan_node/logical_join.rs
BugenZhao: Does it mean that the plan node is somehow directly allocated in heap?

xxchan: Directly allocation in heap seems not possible:
- https://users.rust-lang.org/t/how-to-create-large-objects-directly-in-heap/26405
- https://users.rust-lang.org/t/is-there-equivalent-of-placement-new-in-rust/86611

PR #21424, opened 2025-04-15
test: run fe stack overflow test only in release mode

xxchan:
Signed-off-by: xxchan <xxchan22f@gmail.com>

I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## What's changed and what's your intention?
See https://github.com/risingwavelabs/risingwave/issues/21355#issuecomment-2806876248

future work: #21427

## Checklist

- [ ] I have written necessary rustdoc comments.
- [ ]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  I have checked the [Release Timeline](https://github.com/risingwavelabs/rw-commits-history/blob/main/release_timeline.md) and [Currently Supported Versions](https://docs.risingwave.com/changelog/release-support-policy#support-end-dates-for-recent-releases) to determine which release branches I need to cherry-pick this PR into. 

## Documentation

- [ ]  My PR needs documentation updates. 

<details>
<summary><b>Release note</b></summary>

</details>

PR #21454, opened 2025-04-17
revert: "test: run fe stack overflow test only in release mode"

xxchan:
Reverts risingwavelabs/risingwave#21424

https://github.com/risingwavelabs/risingwave/pull/21424#issuecomment-2812700084

Although I think if the test fail in dev mode, it's a not big deal. But this might be a good way to track increased memory usage in frontend. (until we cannot manage it...)

related #21355

### DIFF
--- a/e2e_test/streaming/bug_fixes/stack_overflow_17342.slt
+++ b/e2e_test/streaming/bug_fixes/stack_overflow_17342.slt
@@ -1,3 +1,5 @@
+# If this test failed, may take a look at https://github.com/risingwavelabs/risingwave/issues/21355
+
 statement ok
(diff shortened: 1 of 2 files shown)


e2e_test/streaming/bug_fixes/stack_overflow_17342.slt
```diff
@@ -1,5 +1,3 @@
-# Why this is release mode only: https://github.com/risingwavelabs/risingwave/issues/21355#issuecomment-2806876248
```
kwannoel, line 1: Can we keep link to issue? So if we encounter some regression again, we have a starting point to debug.

BugenZhao (approved):

PR #21472, opened 2025-04-19
fix(cdc): check database name when creating tables from cdc source

KeXiangWang:

Previously, the following SQL can be executed without errors but the table doesn't really ingest any data. The root cause is that we didn't check whether the `database.name`, `mydb` in this case, matchs the db name in the `from` clause of the `CREATE TABLE` statement, `yourdb` in this case. The `CdcFilter` executer will filter all debezium messages based on the table name from `from` clause, which contains the database name. When this mismatching happens, all the useful events will be filtered out in the `CdcFilter` executer, so the created cdc table is never updated. Currently, only `mysql-cdc ` and `sqlserver-cdc` are affected by this issue.

This PR fixes this issue by adding a database name check. 

```
CREATE SOURCE mssql_source WITH (
    connector = 'sqlserver-cdc',
    hostname = 'xxxxxxx.database.windows.net',
    port = '1433',
    username = 'my_sa',
    password = 'YourPassword123',
    database.name = 'mydb',
);

CREATE TABLE shared_orders (
    order_id INT,
    order_date BIGINT,
    customer_name VARCHAR,
    price DECIMAL,
    product_id INT,
    order_status SMALLINT,
    PRIMARY KEY (order_id)
)
from mssql_source table 'yourdb.dbo.orders';
```

## Checklist

## Documentation

<details>

</details>


src/frontend/src/handler/create_table.rs
```diff
@@ -938,6 +942,18 @@ fn derive_with_options_for_cdc_table(
...
+                // We allow multiple database names in the source definition
+                if !source_database_name
+                    .split(',')
+                    .map(|s| s.trim())
+                    .any(|name| name == db_name)
+                {
+                    return Err(anyhow!(
+                        "The database name `{}` in the FROM clause does not contain the database name `{}` in source definition",
+                        db_name,
+                        source_database_name
```
BugenZhao, line 954: Is the error message backwards?

BugenZhao (commented):
Rest LGTM. Shall we also cherry pick?

BugenZhao (approved):
Please also fix the e2e test for error messages.

PR #11345, opened 2023-08-01
ci: add back auto-retry for main and docker pipeline

huangjw806:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://gist.github.com/TennyZhuang/f00be7f16996ea48effb049aa7be4d66#file-rw_cla).

## What's changed and what's your intention?

In #11304 , the auto-retry(exit code 255) of pull-request workflow was added back.

Here I add back the retry of main, main-cron and docker workflows, and also retry when the exit code is -1.

## Checklist

- [ ] I have written necessary rustdoc comments
- [ ] I have added necessary unit tests and integration tests
- [ ] I have added fuzzing tests or opened an issue to track them. (Optional, recommended for new SQL features #7934).
- [ ] My PR contains breaking changes. (If it deprecates some features, please create a tracking issue to remove them in the future).
- [ ] All checks passed in `./risedev check` (or alias, `./risedev c`)
- [ ] My PR changes performance-critical code. (Please run macro/micro-benchmarks and show the results.)

- [ ] My PR contains critical fixes that are necessary to be merged into the latest release. (Please check out the [details](https://github.com/risingwavelabs/risingwave/blob/main/CONTRIBUTING.md))

## Documentation

- [ ] My PR contains user-facing changes.

PR #21475, opened 2025-04-21
chore(ci): add `exit_status: -1` as buildkite auto-retry condition

BugenZhao:

It's observed that `agent_lost` is not retried in recent main-cron runs (https://buildkite.com/risingwavelabs/main-cron/builds/4973#019653ee-6c34-45fc-8f1b-41218534b599).

According to the doc (https://buildkite.com/resources/changelog/193-signal-and-signal-reason-in-automatic-retry-rules/), we may also include `exit_status: -1` as a retry condition.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/ci/scripts/gen-integration-test-yaml.py
+++ b/ci/scripts/gen-integration-test-yaml.py
@@ -77,6 +77,9 @@ auto-retry: &auto-retry
     # Agent terminated because the AWS EC2 spot instance killed by AWS.
     - signal_reason: agent_stop
       limit: 3
+    - exit_status: -1
+      signal_reason: none
+      limit: 3
 
 steps: {pipeline_steps}
 YAML
--- a/ci/workflows/docker-arm-fast.yml
+++ b/ci/workflows/docker-arm-fast.yml
@@ -3,7 +3,9 @@ auto-retry: &auto-retry
     # Agent terminated because the AWS EC2 spot instance killed by AWS.
     - signal_reason: agent_stop
       limit: 3
-
+    - exit_status: -1
+      signal_reason: none
+      limit: 3
 steps:
   - label: "docker-build-push: aarch64"
     if: build.env("SKIP_TARGET_AARCH64") != "true"
--- a/ci/workflows/docker.yml
+++ b/ci/workflows/docker.yml
@@ -3,6 +3,9 @@ auto-retry: &auto-retry
     # Agent terminated because the AWS EC2 spot instance killed by AWS.
     - signal_reason: agent_stop
       limit: 3
+    - exit_status: -1
+      signal_reason: none
+      limit: 3
 
 steps:
   - label: "docker-build-push: amd64"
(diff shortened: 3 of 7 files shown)


kwannoel (approved):

ci/scripts/gen-integration-test-yaml.py
```diff
@@ -77,6 +77,9 @@ def format_pipeline_yaml_cmd(pipeline_steps):
     # Agent terminated because the AWS EC2 spot instance killed by AWS.
     - signal_reason: agent_stop
       limit: 3
+    - exit_status: -1
+      signal_reason: none
+      limit: 3
```
xxchan, line 82: I found discussion here: https://github.com/risingwavelabs/risingwave/pull/11345#discussion_r1280315011

xxchan: IIUC the idea is we don't want to retry timeout'd tests. (BTW, would -1 also occur for other bash errors?)

BugenZhao: The timeouts I observed recently were all with exit code 143, which is `SIGTERM`.

Even for the case mentioned in https://github.com/risingwavelabs/risingwave/pull/11305#issuecomment-1656004765, there's a `signal_reason` of `SIGKILL`, while we only match `none` here.

BugenZhao:
BTW, I just realized `integration test (madsim) - sink` has been failing with "agent lost" for several consecutive days. There might be an issue with this step itself. 🤡

xxchan:
> BTW, I just realized `integration test (madsim) - sink` has been failing with "agent lost" for several consecutive days. There might be an issue with this step itself. 🤡

I guess "agent lost" can also be sth like OOM 🤡

ci/scripts/gen-integration-test-yaml.py
xxchan: buildkite.com/risingwavelabs/main-cron/builds/4973#019653ee-6c34-45fc-8f1b-41218534b599


This sais Timeout with status -1. Did I miss sth?
image

BugenZhao: It's `1` instead of `-1`, and I believe it's because of this error:

image

xxchan: 🤣

xxchan: Oh! I think `signal_reason: none` plus `-1` can mean "agent lost"?

xxchan: none: No signal was sent to the process.

xxchan: Oh, according to the doc https://buildkite.com/docs/agent/v3#exit-codes
The exit code of the script will be transformed by the buildkite agent

xxchan (approved):

ci/scripts/gen-integration-test-yaml.py
BugenZhao: > I believe it's because of this error (no `junit` xml found)

Seems it's because sqllogictest did not exit gracefully in 30 seconds, so there's no chance to dump the junit file.

Issue #19755, opened 2024-12-11
Support altering fields of struct-type column


fuyufjh:
Let's prioritize this a bit because of the wider adoption of schema registry. Schema registry users have been used to use deeply nested structure, and we can't prevent users from updating the schema. A related issue https://github.com/risingwavelabs/risingwave/issues/19736.

related discussion: https://github.com/risingwavelabs/risingwave/issues/14319

PR #21499, opened 2025-04-22
refactor(meta): avoid using frontend generated `ColumnIndexMapping` when constructing new dispatcher for replacement

BugenZhao:

Previously, when performing schema change, the frontend will generate a `ColumnIndexMapping` between the old and the new schema. It will be used by the meta node to rewrite the `output_indices` field of the dispatchers, to maintain the same schema from the view of existing downstream jobs.

Things start to get complicated since we support altering the column type (https://github.com/risingwavelabs/risingwave/issues/19755): instead of using a simple `output_indices` to transform the data chunks in the dispatcher, we need to introduce some sort of expression evaluation to transform between between two different but compatible composite types.

Note that altering type can be called multiple times on a column. Instead of rewriting the transforming expression with an enhanced `FieldIndexMapping`, it's obvious that generating the transformation between the latest upstream & downstream schema each time can be much easier to implement. This is the motivation behind this PR.

However, we still cannot deprecate the frontend generated `ColumnIndexMapping` now, because it's also used to rewrite the index expressions. ~Considering that the persisted information is all about column index instead of ID, I still haven't come up with a solution to easily update the functional indexes after altering column type.~ UPDATE: it will be removed in an upcoming PR after we eliminate the need for using it to rewrite index expressions:
- #21681
- #21685

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/e2e_test/ddl/alter_table_column_type.slt
+++ b/e2e_test/ddl/alter_table_column_type.slt
@@ -65,11 +65,12 @@ NULL
 
 # If a column is currently being referenced by any downstream job, the column cannot be altered.
 # TODO: support altering referenced columns
-statement ok
-create materialized view my_mv_a as select a from my_table;
+# TODO: supporting in progress, temporarily disable this test
+# statement ok
+# create materialized view my_mv_a as select a from my_table;
 
-statement error unable to drop or alter the column due to being referenced by downstream materialized views or sinks
-alter table my_table alter column a type struct<v double, w varchar, x int>;
+# statement error unable to drop or alter the column due to being referenced by downstream materialized views or sinks
+# alter table my_table alter column a type struct<v double, w varchar, x int>;
 
 statement ok
 drop table my_table cascade;
--- a/e2e_test/source_legacy/cdc_inline/alter/cdc_table_alter.slt
+++ b/e2e_test/source_legacy/cdc_inline/alter/cdc_table_alter.slt
@@ -210,7 +210,7 @@ psql -c "
   ALTER TABLE shipments1 DROP COLUMN destination;
 "
 
-statement error unable to drop or alter the column due to being referenced by downstream materialized views or sinks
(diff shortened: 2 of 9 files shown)


BugenZhao:
* **#21499** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/21499?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/21499?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

xxchan (commented):
Do you mean we want to switch from calculating in FE to calculating in meta? 

What was the original reason to calculate in FE? 🤔

BugenZhao:
> Do you mean we want to switch from calculating in FE to calculating in meta?

Exactly



> What was the original reason to calculate in FE? 🤔

Cannot remember clearly.



For example, there's an upstream table on its schema version $i$, and we create a downstream job from it. Then we now have a column mapping $m(up_i \rightarrow down)$ in the dispatcher. When we are going to alter the upstream table to schema version $i+1$

- (current approach) the frontend generates $m(up_{i+1} \rightarrow up_i)$, the meta node applies it to the existing mapping $m(up_i \rightarrow down)$ in the dispatcher, to get

	$$m(up_{i+1} \rightarrow down) = m(up_{i+1} \rightarrow up_i) \cdot m(up_i \rightarrow down)$$

- (approach after this PR) the meta node directly resolve the schema of $up_{i+1}$ and compare it against $down$, to generate $m(up_{i+1} \rightarrow down)$.

The previous approach eliminated the need for the meta node to resolve table schema information, which might have been considered simpler to implement when table schema changes were first introduced.

However, the key point is that index mapping can be flattened regardless of how many mappings are applied. This does hold any more if we are going to work on complicated mappings involving expressions (for altering column type).

Issue #21487, opened 2025-04-22
Frontend crashes after creating a postgres-cdc source


cyberchen98:
### Describe the bug

Creating a PostgreSQL CDC source(pg17) causes the frontend to crash and enter a crash loop.


### Error message/log

```text
thread 'rw-batch-local' panicked at src/sqlparser/src/ast/mod.rs:1677:45:
normalized SQL should be parsable: ParserError("CDC source cannot define columns and constraints\nLINE 1: CREATE SOURCE pg_source_2 (payload JSONB) WITH (connector = 'postgres-cdc', hostname = 'xxx', port = '5432', username = 'postgres', password = 'postgres', database.name = 'risingwave', schema.name = 'public') FORMAT PLAIN ENCODE JSON\n                                                                                                                                                                                                                                                                                                             ^")
stack backtrace:
   0:     0xaaaae9827c44 - std::backtrace_rs::backtrace::libunwind::trace::h68fa4f2ef0c30713
                               at /rustc/f9e0239a7bc813b4aceffc7f069f4797cde3175c/library/std/src/../../backtrace/src/backtrace/libunwind.rs:117:9
   1:     0xaaaae9827c44 - std::backtrace_rs::backtrace::trace_unsynchronized::h27df4f6f457f4fd5
                               at /rustc/f9e0239a7bc813b4aceffc7f069f4797cde3175c/library/std/src/../../backtrace/src/backtrace/mod.rs:66:14
   2:     0xaaaae9827c44 - std::sys::backtrace::_print_fmt::h18ce864b39d76b45
                               at /rustc/f9e0239a7bc813b4aceffc7f069f4797cde3175c/library/std/src/sys/backtrace.rs:66:9
   3:     0xaaaae9827c44 - <std::sys::backtrace::BacktraceLock::print::DisplayBacktrace as core::fmt::Display>::fmt::hff8abd39e1f90e87
                               at /rustc/f9e0239a7bc813b4aceffc7f069f4797cde3175c/library/std/src/sys/backtrace.rs:39:26
   4:     0xaaaadd2a9404 - core::fmt::rt::Argument::fmt::h9058bde717e94412
                               at /rustc/f9e0239a7bc813b4aceffc7f069f4797cde3175c/library/core/src/fmt/rt.rs:177:76
   5:     0xaaaadd2a9404 - core::fmt::write::h86ac288c85fd4a0a
[...]

PR #21514, opened 2025-04-23
fix(frontend): fallback to persisted SQL if purified SQL is unparsable

BugenZhao:

So that we can tolerate hard error like https://github.com/risingwavelabs/risingwave/issues/21487.

Note that this is not a fix to #21487; it's simply intended to make purification more robust. Instead, the problems we should still address are:

- `purify` may not specify the single `JSONB` column for CDC source
- checks for non-empty column definition for CDC source should not be done in `sqlparser`.

## Checklist

## Documentation

<details>

</details>


BugenZhao:
* **#21521** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/21521?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#21514** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/21514?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/21514?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

xxchan (approved):
> Note that this is not a fix to #21487; 

I don't fully get this. Do you mean this *CANNOT* fix that, or this *does not intent to* fix that (although *CAN* fix that).

BugenZhao:
> > Note that this is not a fix to #21487;
> 
> I don't fully get this. Do you mean this _CANNOT_ fix that, or this _does not intent to_ fix that (although _CAN_ fix that).

Fixed, but as a workaround

src/sqlparser/src/ast/mod.rs
```diff
@@ -1666,22 +1666,61 @@ pub enum DescribeKind {
     Fragments,
 }
 
+task_local::task_local! {
+    /// A flag indicating whether to validate the unparsed SQL to be parsable
+    /// in [`Statement::fmt`] implementation. True by default.
+    static VALIDATE_PARSABLE: bool
```
xiangjinwu, line 1672: What's the benefit of using this task_local? Why not simply move all previous logic (except `expect`) to `try_to_string` and call it from `Display::fmt`?

xiangjinwu (approved):

src/sqlparser/src/ast/mod.rs
xiangjinwu: Oh it's for passing down extra info to recursive call without changing the signature 😂

PR #21516, opened 2025-04-23
fix(agg): fix the error of append-only approx_count_distinct

stdrc:

The PR replaces the base-2 logarithm used in the linear-counting branch with the natural logarithm. In the original HyperLogLog paper the bias-correction for small cardinalities is `E = m × ln(m / V)` where V is the number of zero registers. Using log₂ instead of ln multiplies the estimate by 1/ln 2 ≈ 1.44. Consequently every time the algorithm fell into the “small-range” path (roughly n ≲ 2.5 m, i.e. < 163 840 for m = 65 536) it over-estimated the true cardinality by about 44 %. That matches the 40 %-plus inflation we observed around 20 000 distinct values. Restoring ln brings the estimator back to the theoretical error envelope (≈ 0.4 % standard deviation for m = 65 536) and eliminates the systematic high bias at low counts.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/expr/impl/src/aggregate/approx_count_distinct/updatable.rs
+++ b/src/expr/impl/src/aggregate/approx_count_distinct/updatable.rs
@@ -14,7 +14,7 @@
 
 use super::*;
 
-#[derive(Clone, Default, Debug)]
+#[derive(Clone, Default, Debug, PartialEq, Eq)]
 struct SparseCount {
     inner: Vec<(u8, u64)>,
 }
@@ -87,7 +87,7 @@ impl EstimateSize for SparseCount {
     }
 }
 
-#[derive(Clone, Debug, EstimateSize)]
+#[derive(Clone, Debug, EstimateSize, PartialEq, Eq)]
 pub(super) struct UpdatableBucket<const DENSE_BITS: usize = 16> {
     dense_counts: [u64; DENSE_BITS],
     sparse_counts: SparseCount,
--- a/src/expr/impl/src/aggregate/approx_count_distinct/updatable.rs
+++ b/src/expr/impl/src/aggregate/approx_count_distinct/updatable.rs
@@ -182,10 +182,10 @@ mod tests {
         agg.update(1.into(), false).unwrap();
         agg.update(2.into(), false).unwrap();
         agg.update(3.into(), false).unwrap();
-        assert_eq!(agg.calculate_result(), 4);
+        assert_eq!(agg.calculate_result(), 3);
 
         agg.update(3.into(), false).unwrap();
-        assert_eq!(agg.calculate_result(), 4);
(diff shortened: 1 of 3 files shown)


stdrc:
* **#21516** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/21516?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/21516?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

kwannoel:
Great investigation and fix

yuhao-su (approved):

kwannoel (commented):
I'm also thinking if we can add some property tests, to compare results between approx count and normal count. To make sure the estimates don't exceed the theoretical margin of error.

stdrc:
> I'm also thinking if we can add some property tests, to compare results between approx count and normal count. To make sure the estimate count don't exceed the theoretical margin of error.

Do you mean adding such test for our existing e2e workload or something?

src/expr/impl/src/aggregate/approx_count_distinct/updatable.rs
```diff
@@ -182,10 +182,10 @@ mod tests {
         agg.update(1.into(), false).unwrap();
         agg.update(2.into(), false).unwrap();
         agg.update(3.into(), false).unwrap();
-        assert_eq!(agg.calculate_result(), 4);
+        assert_eq!(agg.calculate_result(), 3);
 
         agg.update(3.into(), false).unwrap();
-        assert_eq!(agg.calculate_result(), 4);
+        assert_eq!(agg.calculate_result(), 3);
```
stdrc, line 188: For this UT, doing this change just for passing the test. I'm still investigating on the updatable implementation of HLL. Will do better UT in next PR.

kwannoel:
> > I'm also thinking if we can add some property tests, to compare results between approx count and normal count. To make sure the estimate count don't exceed the theoretical margin of error.
> 
> Do you mean adding such test for our existing e2e workload or something?

Might not be simple to add to e2e workload.

If we can just isolate the algorithm, it can be added as a property test via `quickcheck` or some other fuzz test framework.
Then we can do the following:
- Randomly generate a count according to some range policy (extreme values, normal values)
- Partition the count into parts. E.g. given the count 100,000, it can be partitioned into 1, 2, 99997.
- Run our approx count algo on these parts.

yuhao-su:
Should we doc this feature now?

stdrc:
> Should we doc this feature now?

No hurry. Let me do it after refine the retractable one.

PR #19688, opened 2024-12-05
perf(stream): optimize native pg sink

kwannoel:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://gist.github.com/TennyZhuang/f00be7f16996ea48effb049aa7be4d66#file-rw_cla).

## What's changed and what's your intention?

Just measure the throughput for both via grafana.

The performance improvement of unoptimized to optimized native pg sink went from 30r/s to 10K r/s.

The performance is around the same as JDBC. LHS is for native pg sink. RHS is for jdbc sink. Toggle the screenshot below.

---

<details>
<summary>screenshot</summary>

Screenshot 2024-12-06 at 2 47 15 PM

</details>

---

Implementation details:
1. We batch inserts and deletes in this PR, with up to 32768 parameters, since that's the max parameter supported by PG.
2. We use transactions to delimit each batch.
3. We remove `updateInsert` handling path, since compaction will normalizze `updateInsert` to `Insert`.
4. We use `LogSinker` instead of `SinkWriter` to avoid redundant flushes to the log store.

## Checklist

- [ ] I have written necessary rustdoc comments
- [ ] I have added necessary unit tests and integration tests
- [ ] I have added test labels as necessary. See [details](https://github.com/risingwavelabs/risingwave/blob/main/docs/developer-guide.md#ci-labels-guide).
- [ ] I have added fuzzing tests or opened an issue to track them. (Optional, recommended for new SQL features #7934).
- [ ] My PR contains breaking changes. (If it deprecates some features, please create a tracking issue to remove them in the future).
- [ ] All checks passed in `./risedev check` (or alias, `./risedev c`)
- [ ] My PR changes performance-critical code. (Please run macro/micro-benchmarks and show the results.)

- [ ] My PR contains critical fixes that are necessary to be merged into the latest release. (Please check out the [details](https://github.com/risingwavelabs/risingwave/blob/main/CONTRIBUTING.md))

## Documentation

- [ ] My PR needs documentation updates. (Please use the **Release note** section below to summarize the impact on users)

## Release note

If this PR includes changes that directly affect users or other significant modifications relevant to the community, kindly draft a release note to provide a concise summary of these changes. Please prioritize highlighting the impact these changes will have on users.

</details>

PR #21523, opened 2025-04-23
fix(connector): use `insert on conflict do update` for postgres sink

kwannoel:

If there is either:
- recovery
- multiple sinks writing to same table

There can be duplicate entries for the same PK. We should use `on conflict do update` for non append only sinks.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/src/sink/postgres.rs
+++ b/src/connector/src/sink/postgres.rs
@@ -420,6 +420,7 @@ impl PostgresSinkWriter {
             Op::Insert,
             &mut transaction,
             &self.schema,
+            &self.config.schema,
             &self.config.table,
             &self.pk_indices,
             parameters,
@@ -460,6 +461,7 @@ impl PostgresSinkWriter {
             Op::Delete,
             &mut transaction,
             &self.schema,
+            &self.config.schema,
             &self.config.table,
             &self.pk_indices,
             delete_parameters,
@@ -472,6 +474,7 @@ impl PostgresSinkWriter {
             Op::Insert,
             &mut transaction,
             &self.schema,
+            &self.config.schema,
             &self.config.table,
             &self.pk_indices,
             insert_parameters,
@@ -488,6 +491,7 @@ impl PostgresSinkWriter {
         op: Op,
         transaction: &mut tokio_postgres::Transaction<'_>,
         schema: &Schema,
+        schema_name: &str,
         table_name: &str,
         pk_indices: &[usize],
         parameters: Vec<Vec<Option<ScalarAdapter>>>,
@@ -512,12 +516,14 @@ impl PostgresSinkWriter {
             let statement_str = match op {
                 Op::Insert => {
                     if append_only {
-                        create_insert_sql(schema, table_name, rows_length)
+                        create_insert_sql(schema, schema_name, table_name, rows_length)
                     } else {
-                        create_upsert_sql(schema, table_name, pk_indices, rows_length)
+                        create_upsert_sql(schema, schema_name, table_name, pk_indices, rows_length)
                     }
                 }
-                Op::Delete => create_delete_sql(schema, table_name, pk_indices, rows_length),
+                Op::Delete => {
+                    create_delete_sql(schema, schema_name, table_name, pk_indices, rows_length)
+                }
                 _ => unreachable!(),
             };
             let statement = transaction.prepare(&statement_str).await?;
@@ -538,12 +544,14 @@ impl PostgresSinkWriter {
             let statement_str = match op {
                 Op::Insert => {
                     if append_only {
-                        create_insert_sql(schema, table_name, rows_length)
+                        create_insert_sql(schema, schema_name, table_name, rows_length)
                     } else {
-                        create_upsert_sql(schema, table_name, pk_indices, rows_length)
+                        create_upsert_sql(schema, schema_name, table_name, pk_indices, rows_length)
                     }
                 }
-                Op::Delete => create_delete_sql(schema, table_name, pk_indices, rows_length),
+                Op::Delete => {
+                    create_delete_sql(schema, schema_name, table_name, pk_indices, rows_length)
+                }
                 _ => unreachable!(),
             };
             tracing::trace!("binding statement: {:?}", statement_str);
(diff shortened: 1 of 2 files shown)


kwannoel:
> LGTM. Looks like optimize native pg sink (#19688) this PR used insert instead of upsert? I remembered the first version we have the upsert logic.

The first version has upsert logic, but it's only used for `updateInsert`. So also not implemented correctly. A correct implementation will use it for both `insert` and `updateInsert`, since duplicate records maybe sinked over recovery, multiple writers to the base table, or existing data in the PG table.

zwang28 (approved):

src/connector/src/sink/postgres.rs
```diff
@@ -506,7 +510,13 @@ impl PostgresSinkWriter {
                 schema.fields(),
             );
             let statement = match op {
-                Op::Insert => create_insert_sql(schema, table_name, rows_length),
+                Op::Insert => {
+                    if append_only {
+                        create_insert_sql(schema, table_name, rows_length)
+                    } else {
+                        create_upsert_sql(schema, table_name, pk_indices)
+                    }
```
xiangjinwu, line 518: I am not familiar with how it works previously, but these 2 branches look imbalanced:
* `create_insert_sql` generates `format!("${}", i * number_of_columns + j + 1)`
* `create_upsert_sql` generates `format!("${}", i + 1)`

xiangjinwu: And FYI PostgreSQL supports a special table name `excluded`:
```
ON CONFLICT (did) DO UPDATE SET dname = EXCLUDED.dname;
```
https://www.postgresql.org/docs/current/sql-insert.html
This can be helpful when handling multi-row insertion.

Another issue not limited to this PR: **interpolating SQL string is error-prone**. For example, the table and column names shall be quoted (`"foo"`) or uppsercase letters are folded to lowercase by PostgreSQL - more tricky is that quotes shall be doubled (e.g. when `field.name` is `abC"Def` the correct quoting is `"abC""Def"`)

src/connector/src/sink/postgres.rs
```diff
@@ -662,4 +725,24 @@ mod tests {
...
+            Field {
+                data_type: DataType::Int32,
+                name: "b".to_owned(),
+            },
+        ]);
+        let table_name = "test_table";
+        let sql = create_upsert_sql(&schema, table_name, &[1], 3);
+        check(
+            sql,
+            expect!["INSERT INTO test_table (a, b) VALUES ($1, $2), ($3, $4), ($5, $6) on conflict (b) do update set (a = $1), (a = $3), (a = $5)"],
```
xiangjinwu, line 745:
```
test=# INSERT INTO test_table (a, b) VALUES (1, 2), (3, 4), (5, 6) on conflict (b) do update set (a = 1), (a = 3), (a = 5);
ERROR:  syntax error at or near "="
LINE 1: ..., (3, 4), (5, 6) on conflict (b) do update set (a = 1), (a =...
                                                             ^

test=# INSERT INTO test_table (a, b) VALUES (1, 2), (3, 4), (5, 6) on conflict (b) do update set a = 1, a = 3, a = 5;
ERROR:  multiple assignments to same column "a"
```

src/connector/src/sink/postgres.rs
kwannoel: > And FYI PostgreSQL supports a special table name excluded:
ON CONFLICT (did) DO UPDATE SET dname = EXCLUDED.dname;
https://www.postgresql.org/docs/current/sql-insert.html
This can be helpful when handling multi-row insertion.

Thanks, this is helpful.

> Another issue not limited to this PR: interpolating SQL string is error-prone. For example, the table and column names shall be quoted ("foo") or uppsercase letters are folded to lowercase by PostgreSQL - more tricky is that quotes shall be doubled (e.g. when field.name is abC"Def the correct quoting is "abC""Def")

Yeah I guess we need to audit PG CDC as well for this. For PG Native Sink, I don't think we have that quoting in place.

kwannoel:
* **#21543** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/21543?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#21523** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/21523?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/21523?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/connector/src/sink/postgres.rs
kwannoel: Fixed

chenzl25 (approved):
LGTM

src/connector/src/sink/postgres.rs
```diff
@@ -522,17 +544,29 @@ impl PostgresSinkWriter {
                 0,
                 "flattened parameters are unaligned"
             );
-            let statement = match op {
-                Op::Insert => create_insert_sql(schema, table_name, rows_length),
-                Op::Delete => create_delete_sql(schema, table_name, pk_indices, rows_length),
+            let statement_str = match op {
```
xiangjinwu, line 547: nit: looks same as (or similar to) the snippet above. Consider sharing the code so that there is no need to keep them in sync.

xiangjinwu (approved):
Lgtm. Feel free to defer or ignore the nitpicks

src/connector/src/sink/postgres.rs
kwannoel: Addressed it in https://github.com/risingwavelabs/risingwave/pull/21566/files

src/connector/src/sink/postgres.rs
```diff
@@ -505,14 +513,28 @@ impl PostgresSinkWriter {
...
+                Op::Delete => {
+                    create_delete_sql(schema, schema_name, table_name, pk_indices, rows_length)
+                }
                 _ => unreachable!(),
             };
-            let statement = transaction.prepare(&statement).await?;
+            let statement = transaction
+                .prepare(&statement_str)
+                .await
+                .with_context(|| format!("failed to run statement: {}", statement_str))?;
```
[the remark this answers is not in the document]
kwannoel: https://github.com/risingwavelabs/risingwave/pull/21566/commits/5f4078f9bb051f9d52a15c3f4d741739e6145793

src/connector/src/sink/postgres.rs
```diff
@@ -505,14 +513,28 @@ impl PostgresSinkWriter {
...
             for parameter in parameters {
-                transaction.execute_raw(&statement, parameter).await?;
+                transaction
+                    .execute_raw(&statement, parameter)
+                    .await
+                    .with_context(|| format!("failed to run statement: {}", statement_str,))?;
```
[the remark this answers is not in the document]
kwannoel: https://github.com/risingwavelabs/risingwave/pull/21566/commits/5f4078f9bb051f9d52a15c3f4d741739e6145793

kwannoel:
Merged via https://github.com/risingwavelabs/risingwave/pull/21543

Issue #19075, opened 2024-10-23
bug: window TVF can not handle month and year interval correctly


st1page:
```
dev=> create table t1 (v int, created_at timestamp);
insert into t1 values
    (1, '2021-01-01 10:00:00'),
    (2, '2000-01-01 10:05:00'),
    (3, '2012-01-01 10:14:00'),
    (4, '2042-01-01 10:22:00');
CREATE_TABLE
INSERT 0 4

dev=> select * from tumble(t1, created_at, interval '1 year');
 v |     created_at      |    window_start     |     window_end
---+---------------------+---------------------+---------------------
 3 | 2012-01-01 10:14:00 | 2011-05-26 00:00:00 | 2012-05-26 00:00:00
 1 | 2021-01-01 10:00:00 | 2020-04-08 00:00:00 | 2021-04-08 00:00:00
 2 | 2000-01-01 10:05:00 | 1999-07-28 00:00:00 | 2000-07-28 00:00:00
 4 | 2042-01-01 10:22:00 | 2041-12-14 00:00:00 | 2042-12-14 00:00:00
(4 rows)

dev=> select * from tumble(t1, created_at, interval '1 month');
 v |     created_at      |    window_start     |     window_end
---+---------------------+---------------------+---------------------
 3 | 2012-01-01 10:14:00 | 2011-12-22 00:00:00 | 2012-01-22 00:00:00
 4 | 2042-01-01 10:22:00 | 2041-12-14 00:00:00 | 2042-01-14 00:00:00
 1 | 2021-01-01 10:00:00 | 2020-12-04 00:00:00 | 2021-01-04 00:00:00
 2 | 2000-01-01 10:05:00 | 1999-12-25 00:00:00 | 2000-01-25 00:00:00
(4 rows)
```

fuyufjh:
This case should probably use `date_trunc()`?

st1page:
> This case should probably use `date_trunc()`?

I recommend user use that, but I'm not sure there will be a window length that more than one day. 
If we're not sure of the appropriate behavior, I guess we can ban them first so they don't get misused

PR #21559, opened 2025-04-24
feat(expr): add date_bin

KeXiangWang:

Impl https://www.postgresql.org/docs/current/functions-datetime.html#FUNCTIONS-DATETIME-BIN

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/e2e_test/batch/functions/date_bin.part
+++ b/e2e_test/batch/functions/date_bin.part
@@ -42,3 +42,14 @@ SELECT date_bin(
 );
 ----
 2021-12-31 23:45:00+00:00
+
+statement ok
+set timezone = 'US/Pacific';
+
+query T
+select date_bin(interval '1' day, '2021-09-08T18:05:04-07:00'::timestamptz, '2021-01-01'::timestamp);
+----
+2021-09-08 01:00:00-07:00
+
+statement ok
+set timezone = 'UTC';
--- a/e2e_test/batch/functions/date_bin.part
+++ b/e2e_test/batch/functions/date_bin.part
@@ -1,0 +1,44 @@
+query T
+SELECT date_bin(
+  '2 days'::interval,
+  '2021-01-02 12:00:00+08'::TIMESTAMPTZ,
+  '2021-01-01 00:00:00+00'::TIMESTAMPTZ
+);
+----
+2021-01-01 00:00:00+00:00
+
+query T
+SELECT date_bin(
+  '3 hours'::interval,
+  '2021-01-01 05:45:00'::TIMESTAMP,
+  '2021-01-01 00:00:00'::TIMESTAMP
+);
+----
+2021-01-01 03:00:00
+
+query T
+SELECT date_bin(
+  '30 minutes'::interval,
+  '2021-01-01 02:17:00-05'::TIMESTAMPTZ,
+  '2021-01-01 01:00:00-05'::TIMESTAMPTZ
+);
+----
+2021-01-01 07:00:00+00:00
+
+query T
+SELECT date_bin(
+  '1 hour'::interval,
+  '2021-06-15 13:30:00'::TIMESTAMP,
+  '2021-06-15 00:00:00+02'::TIMESTAMPTZ
+);
+----
+2021-06-15 13:00:00+00:00
+
+query T
+SELECT date_bin(
+  '15 minutes'::interval,
+  '2021-12-31 23:59:59+00'::TIMESTAMPTZ,
+  '2021-12-31 22:00:00'::TIMESTAMP
+);
+----
+2021-12-31 23:45:00+00:00
--- a/src/expr/impl/src/scalar/date_bin.rs
+++ b/src/expr/impl/src/scalar/date_bin.rs
@@ -1,0 +1,97 @@
+// Copyright 2025 RisingWave Labs
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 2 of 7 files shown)


src/expr/impl/src/scalar/date_bin.rs
```diff
@@ -0,0 +1,97 @@
...
+}
+
+#[function("date_bin(interval, timestamp, timestamptz) -> timestamptz")]
+pub fn date_bin_ts_tstz(
+    stride: Interval,
+    source: Timestamp,
+    origin: Timestamptz,
+) -> Result<Timestamptz> {
+    let source_us = source.0.and_utc().timestamp_micros(); // source to microseconds
+    let origin_us = origin.timestamp_micros(); // origin to microseconds
```
xiangjinwu, line 52: Mixed usage of timestamp and timestamptz is wrong. PostgreSQL only defines 2 overloads: `select * from pg_proc where proname = 'date_bin';`

When the input is mixed, PostgreSQL (and RisingWave) tries to add implicit cast to find a best match (#3393). In this case, the timestamp argument will be casted into timestamptz **in the session zone** (NOT `0.and_utc()`).

The fix is also simple: just remove the implementation of these 2 overloads.

e2e_test/batch/functions/date_bin.part
```diff
@@ -0,0 +1,44 @@
...
+2021-06-15 13:00:00+00:00
+
+query T
+SELECT date_bin(
+  '15 minutes'::interval,
+  '2021-12-31 23:59:59+00'::TIMESTAMPTZ,
+  '2021-12-31 22:00:00'::TIMESTAMP
+);
+----
+2021-12-31 23:45:00+00:00
```
xiangjinwu, line 44: Some alternative test cases to demonstrate the expected behavior of timezone handling:
```
set timezone = 'US/Pacific';

select date_bin(interval '1' day, '2021-09-08T18:05:04-07:00'::timestamptz, '2021-01-01'::timestamp);
----
2021-09-08 01:00:00-07:00

set timezone = 'UTC';
```

This case actually tests 2 things:
* `'2021-01-01'::timestamp` is interpreted in session time zone `US/Pacific` thus `1609488000000000` (rather than in UTC `1609459200000000`)
* `1 day` got no special treatment regarding DST. It is always 24 hours instead of 23 or 25. So it is NOT `2021-09-08 00:00:00-07:00` but `2021-09-08 01:00:00-07:00`

stdrc:
`date_bin` seems to be a stronger version of `tumble_start`, isn't it? Can we reuse the code in either of the two?

xiangjinwu:
> `date_bin` seems to be a stronger version of `tumble_start`, isn't it? Can we reuse the code in either of the two?

Yes I also thought the same but did not bring it up to limit the scope of this PR. We need to clarify the expected behavior of `TUMBLE` in some cases:
* Do we intend to support month? [#19075](https://github.com/risingwavelabs/risingwave/issues/19075)
* Should it handle DST? If yes, then it would require different implementation from `date_bin`
* Why did we define the optional `offset` as an `interval` rather than a `timestamp/timestamptz` here?

KeXiangWang:
Will continue to perfect the PR this weekend.

src/expr/impl/src/scalar/date_bin.rs
KeXiangWang: Updated. Thanks!

src/expr/impl/src/scalar/date_bin.rs
```diff
@@ -0,0 +1,97 @@
...
+
+fn date_bin_inner(stride: Interval, source_us: i64, origin_us: i64) -> Result<i64> {
+    if stride.months() != 0 {
+        // PostgreSQL doesn't allow months in the interval for date_bin.
+        return Err(ExprError::InvalidParam {
+            name: "stride",
+            reason: "stride interval with months not supported in date_bin".into(),
+        });
+    }
+    let stride_us = stride.usecs() + (stride.days() as i64) * (24 * 3600 * 1000000); // stride width in microseconds
```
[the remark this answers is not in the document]
KeXiangWang: done

e2e_test/batch/functions/date_bin.part
KeXiangWang: Added. Thanks for the test case!

src/expr/impl/src/scalar/date_bin.rs
```diff
@@ -0,0 +1,71 @@
...
+) -> Result<Timestamptz> {
+    let source_us = source.timestamp_micros(); // source to microseconds
+    let origin_us = origin.timestamp_micros(); // origin to microseconds
+
+    let binned_source_us = date_bin_inner(stride, source_us, origin_us)?;
+    Ok(Timestamptz::from_micros(binned_source_us))
+}
+
+fn date_bin_inner(stride: Interval, source_us: i64, origin_us: i64) -> Result<i64> {
+    if stride.months() != 0 {
```
stdrc, line 46: Seems we also check `stride.yeas()`? PG says:

```
> select date_bin(interval '1 year', timestamp '2025-02-28 01:00:00', timestamp '1970-01-01 00:00:00');

Query Error: timestamps cannot be binned into intervals containing months or years
```

xiangjinwu: 1 year is always 12 months and thus stored in `months` as well

stdrc: thanks for clarify

PR #21568, opened 2025-04-25
fix(meta): Don’t need to notify the frontend to clean up subscriptions upon recovery.

xxhZs:

in this pr https://github.com/risingwavelabs/risingwave/pull/21479
We only created_subscriptions that will be sent to the frontend, so we don’t need to notify the frontend to clean up subscriptions upon recovery.

## Checklist

## Documentation

<details>

</details>


src/meta/src/controller/catalog/mod.rs
```diff
@@ -385,23 +385,12 @@ impl CatalogController {
...
             .filter(filter_condition)
             .exec(&txn)
             .await?;
         txn.commit().await?;
-        let object_group = build_object_group_for_delete(objects);
-
-        let _version = self
-            .notify_frontend(NotificationOperation::Delete, object_group)
-            .await;
+        // We don't need to notify the frontend, because the Init subscription is not send to frontend.
```
kwannoel, line 393: When do we send subscription catalog to FE? Is it only after successful creation?

kwannoel: I assume so.

kwannoel (approved):
LGTM

kwannoel:
We need to cherry-pick this in as well?

xxhZs:
> We need to cherry-pick this in as well?

yesyes

Issue #20799, opened 2025-03-10
create mysql 5.x cdc table error


litZz:
### Describe the bug

when creating a MySQL 5.7 cdc table in RisingWave, an error message appears when the DDL contains a timestamp type field with a default value

### Error message/log

```text
ERROR:  Failed to run the query
Caused by these errors (recent errors listed first):
  1: failed to auto derive table schema
  2: input contains invalid characters
  3: failed to parse mysql timestamp value
```

### To Reproduce

1. create a mysql 5.7 cluster
2. Create a  table in mysql, the ddl contains a timestamp type field with a default value
     CREATE TABLE `test_rw`.`t1` (
       `id` varchar(50) NOT NULL,
        `create_timestamp` timestamp(3) NOT NULL DEFAULT CURRENT_TIMESTAMP(3)
      );
4. create mysql cdc source in RisingWave
      CREATE SOURCE mysql_cdc_t1 with (
        connector = 'mysql-cdc',
        hostname = '127.0.0.1',
        port = '3306',
        username = 'xxxxxx',
        password = 'xxxxxx',
        database.name = 'test_rw',
        server.id = '7992323'
      );
6.   create table in RisingWave
       REATE TABLE t1 (
         id string PRIMARY KEY,
        create_timestamp timestamptz
        )
       INCLUDE timestamp AS us
       WITH (
          snapshot='true'
        )

the error msg shows as below:

ERROR:  Failed to run the query

Caused by these errors (recent errors listed first):
  1: failed to auto derive table schema
  2: input contains invalid characters
  3: failed to parse mysql timestamp value

### Expected behavior

I expect the table to be created successfully.

### How did you deploy RisingWave?

local & RisingWave operator

### The version of RisingWave

select version();
                    version
------------------------------------------------
 PostgreSQL 13.14.0-RisingWave-2.2.1 (9cb476af)

### Additional context

when create mysql ExternalTableImpl, MySqlExternalTable::connect will use sea schema to obtain column information and parse it.
sea schema map mysql 5.x timestamp default to ColumnDefault::String.
So when data_type = Timestamptz and ColumnDefault = String, the timestamp(3) string will be parsed as time, resulting in an error.

litZz:
when create mysql ExternalTableImpl,  MySqlExternalTable::connect will use sea schema to obtain column information and parse it.
sea schema map mysql 5.x timestamp default to ColumnDefault::String.
[...]

PR #21569, opened 2025-04-25
fix(cdc): mysql default value conversion failure should be tolerated

BugenZhao:

Fix https://github.com/risingwavelabs/risingwave/issues/20799.

Note: we also have tried to remove the support in https://github.com/risingwavelabs/risingwave/pull/21024, but there's some problem. See discussions here: https://risingwave-labs.slack.com/archives/C064SBT0ASF/p1742465000639249?thread_ts=1741859862.262589&cid=C064SBT0ASF

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/src/source/cdc/external/mysql.rs
+++ b/src/connector/src/source/cdc/external/mysql.rs
@@ -26,7 +26,7 @@ use mysql_common::value::Value;
 use risingwave_common::bail;
 use risingwave_common::catalog::{CDC_OFFSET_COLUMN_NAME, ColumnDesc, ColumnId, Schema};
 use risingwave_common::row::OwnedRow;
-use risingwave_common::types::{DataType, Decimal, F32, ScalarImpl};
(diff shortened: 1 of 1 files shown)


src/connector/src/source/cdc/external/mysql.rs
```diff
@@ -115,65 +115,17 @@ impl MySqlExternalTable {
...
-                            );
-                            None
-                        }
-                    },
-                    ColumnDefault::String(mut val) => {
-                        // mysql timestamp is mapped to timestamptz, we use UTC timezone to
-                        // interpret its value
-                        if data_type == DataType::Timestamptz {
-                            val = timestamp_val_to_timestamptz(val.as_str())?;
-                        }
```
BugenZhao, line 161: Previously the implementation tried to handle all errors and fallback to `NULL`, but missed this line.

In this PR, we refactor the process into a separate function and handle errors through a unified path.

PR #21591, opened 2025-04-27
fix(meta): try reconnect to worker node on create new streaming job

wenym1:

Currently, when global recovery is not triggered, we won't retry creating control stream to failed CN worker in most scenarios. We only try recreating the stream on per-database failure recovery. However, we don't retry on all failed CN, but only for CN on which there are some existing jobs of the recovering database running.

As a result, when a worker without any job running gets disconnected on its control stream, any per-database failure recovery won't retry creating control stream to the worker, until a global recovery is triggered. When this happens, any create streaming job command that schedules any actor to this worker will fail with `unconnected worker`, even though the worker has been brought up for a while.

In this PR, we will change to also try recreating the control stream when we handle new create streaming job command. When we see that the worker node of an actor of the job is not connected with control stream, we will try to recreate the control stream to it.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/meta/src/barrier/worker.rs
+++ b/src/meta/src/barrier/worker.rs
@@ -43,8 +43,8 @@ use crate::barrier::context::{GlobalBarrierWorkerContext, GlobalBarrierWorkerCon
 use crate::barrier::rpc::{ControlStreamManager, merge_node_rpc_errors};
 use crate::barrier::schedule::{MarkReadyOptions, PeriodicBarriers};
 use crate::barrier::{
-    BarrierManagerRequest, BarrierManagerStatus, BarrierWorkerRuntimeInfoSnapshot, RecoveryReason,
-    schedule,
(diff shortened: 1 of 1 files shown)


hzxa21 (approved):
LGTM. Has this issue caused any test failure or did we spot the issue by ourselves?

wenym1:
> LGTM. Has this issue caused any test failure or did we spot the issue by ourselves?

Yes. It's observed in the test of @CAJan93.

yezizp2012 (approved):

CAJan93:
@wenym1 Thank you very much! 

@hzxa21 yes, we encountered this error multiple times. You can find discussion and logs here: https://risingwave-labs.slack.com/archives/C0340PNL68P/p1745396331888519
I recently wrote an e2e test that restarts the tenant and on Azure we often run into this issue. Not sure why Azure, maybe they just have more network issues...

PR #21594, opened 2025-04-27
feat: support partition by for iceberg table engine

chenzl25:

- This PR reuses the `partition_by` options as how it used by the iceberg sink, so users can define their partition by columns.
- We restrict the partition by column should be the prefix of the primary key.
- Also, we will try to drop the table if the internal iceberg sink and iceberg source creation fail.

## Checklist

## Documentation

<details>

</details>


chenzl25:
Could you please take a look? cc @Li0k @xxchan @xxhZs

src/frontend/src/handler/create_table.rs
```diff
@@ -1709,6 +1714,82 @@ pub async fn create_iceberg_engine_table(
...
+    if let Some(partition_by) = &partition_by {
+        // captures column, transform(column), transform(n,column), transform(n, column)
+        let re =
+            Regex::new(r"(?<transform>\w+)(\(((?<n>\d+)?(?:,|(,\s)))?(?<field>\w+)\))?").unwrap();
+        if !re.is_match(partition_by) {
+            bail!(format!(
+                "Invalid partition fields: {}\nHINT: Supported formats are column, transform(column), transform(n,column), transform(n, column)",
+                partition_by
+            ))
+        }
```
xxchan, line 1731: These code are copy-pasted from sink? 🤔

xxchan: Not sure whether we need keep both of them. If needed, maybe better to have a common `fn(String) ->  Vec<UnboundPartitionField>` instead

e2e_test/iceberg/test_case/pure_slt/iceberg_engine.slt
```diff
@@ -269,10 +269,38 @@ statement ok
...
+
+statement ok
+create table t_partition3(c1 int, c2 int, c3 int, primary key(c1, c2, c3)) with(commit_checkpoint_interval = 1, partition_by='bucket(4, c1),c2') engine = iceberg;
+
+statement ok
+create table t_partition4(c1 int, c2 int, c3 int, primary key(c1, c2, c3)) with(commit_checkpoint_interval = 1, partition_by='c1,truncate(8, c2)') engine = iceberg;
+
+# the partition key should be the prefix of the primary key
+statement error
+create table t_partition5(c1 int, c2 int, c3 int, primary key(c1, c2, c3)) with(commit_checkpoint_interval = 1, partition_by='c2,c3') engine = iceberg;
```
xxchan, line 290: Ideally validating partitions here. But maybe after support sth like https://iceberg.apache.org/docs/nightly/spark-queries/#partitions (https://github.com/apache/iceberg-rust/issues/823)

xxchan (approved):

Issue #21601, opened 2025-04-28
CI: When building patch branches, ensure the base branch has the same semver


kwannoel:
For example, consider `v2.2.3-patch-some-thing`.
The branch the image is built on should contain the tagged commit: `v2.2.3`.

xxchan:
We have some check like this:
https://github.com/risingwavelabs/risingwave/blob/e05848715d6f41a25e65d625496eea73d63a0979/ci/scripts/release.sh#L59-L68

We can move that check to earlier stage (docker.sh)

PR #21626, opened 2025-04-28
feat(ci): support workflow for docker image building

kwannoel:

Closes https://github.com/risingwavelabs/risingwave/issues/21601.

This PR reduces margin for human error due to manual tagging of images. It also makes tagging entirely optional. We can automatically generate semantically correct tagging after this PR:
```
v<X.Y.Z[-alpha]>--<label>--<commit_sha>--<git-branch>
```

We use `--`, to clearly delimit different sections. Since hyphens `-` can be used within each segment as well. It will only be a problem if developer uses `-` as a prefix or suffix. I think that's unlikely.

Test workflow dispatch via cli (actual use will just be via github UI for workflows, see screenshot below):
```sh
 gh workflow run 'Build Docker Image' --ref kwannoel/workflow-update-branch --field label='mylabel'
```

[example run](https://github.com/risingwavelabs/risingwave/actions/runs/14808598756/job/41580436686)

After this PR is merged, we can have a similar UX to this (we won't actual fill in a version, just an arbitrary tag, ignore that part):

Screenshot 2025-04-29 at 4 52 32 PM

This gives us a "form-filling experience", as requested by @BugenZhao. It's inspired by the work by @arkbriar for image building in our other repositories.

Subsequently, I will add docs for how to use this workflow in our internal and developer docs.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/.github/workflows/build-docker-image.yml
+++ b/.github/workflows/build-docker-image.yml
@@ -47,16 +47,11 @@ jobs:
           echo "enable_alpha_build=${{ github.event.inputs.enable_alpha_build }}"
       - name: 'Verify release branch'
         id: verify_release_branch
-        # FIXME
-        # if: github.event.inputs.enable_patch_build == 'false' && github.event.inputs.enable_alpha_build == 'false'
+        if: github.event.inputs.enable_patch_build && github.event.inputs.enable_alpha_build
         run: |
           # Get current branch name
           echo "GITHUB_REF=$GITHUB_REF"
-          
-          # BRANCH_NAME=$(basename $GITHUB_REF)
-          # FIXME
-          BRANCH_NAME=release-8888.8888
-          
+          BRANCH_NAME=$(basename $GITHUB_REF)
           echo "BRANCH_NAME=$BRANCH_NAME"
           echo "BRANCH_NAME=$BRANCH_NAME" >> $GITHUB_ENV
           
@@ -69,18 +64,14 @@ jobs:
           fi
       - name: 'Get base version'
         id: get_base_version
-        # FIXME
-        # if: github.event.inputs.enable_alpha_build == 'false'
+        if: !github.event.inputs.enable_alpha_build
         run: |
-          # FIXME
-          VERSION=8888.8888.0
-          # VERSION=$(grep -m 1 '^version' Cargo.toml | cut -d '"' -f 2) 
-          # Check its not an alpha build
+          VERSION=$(grep -m 1 '^version' Cargo.toml | cut -d '"' -f 2) 
+          echo "VERSION=$VERSION"
           if [[ $VERSION == *"alpha"* ]]; then
-            echo "Error: Branch $BRANCH_NAME is has an alpha version: $VERSION"
+            echo "Error: Branch $BRANCH_NAME has an alpha version"
             exit 1
           fi
-          echo "VERSION=$VERSION"
           echo "VERSION=$VERSION" >> $GITHUB_ENV
       - name: 'Generate image tag'
         id: generate_image_tag
--- a/.github/workflows/build-docker-image.yml
+++ b/.github/workflows/build-docker-image.yml
@@ -4,7 +4,7 @@ on:
   workflow_dispatch:
     inputs:
       label:
-        description: 'Generate image tag + build Docker image: v<X.Y.Z>:<label>:<commit_sha>:<branch>'
+        description: 'Generate image tag + build Docker image: v<X.Y.Z>--<label>--<commit_sha>--<branch>'
         required: true
         type: string
         default: 'unlabeled'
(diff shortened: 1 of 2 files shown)


.github/workflows/build-docker-image.yml
```diff
@@ -0,0 +1,106 @@
...
+          echo "TAG=${{ github.event.inputs.tag }}"
+          echo "enable_patch_build=${{ github.event.inputs.enable_patch_build }}"
+          echo "enable_alpha_build=${{ github.event.inputs.enable_alpha_build }}"
+      - name: 'Verify release branch'
+        id: verify_release_branch
+        if: github.event.inputs.enable_patch_build == 'false' && github.event.inputs.enable_alpha_build == 'false'
+        run: |
+          # Get current branch name
+          echo "GITHUB_REF=$GITHUB_REF"
+          BRANCH_NAME=$(basename $GITHUB_REF)
```
BugenZhao, line 54: What if my branch is named `bz/release-2.4` 🤔 Is there any better way to get the branch name?

.github/workflows/build-docker-image.yml
```diff
@@ -0,0 +1,106 @@
+# To test: gh workflow run 'Build Docker Image' --ref kwannoel/workflow-update-branch
+name: Build Docker Image
+on:
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'Tag for the Docker image: vX.Y.Z-<tag>-<commit_sha>'
+        required: true
+        type: string
+        default: 'untagged'
```
BugenZhao, line 10: Will `untagged` appear verbatim in the docker image tag? Is this desired?

.github/workflows/build-docker-image.yml
```diff
@@ -0,0 +1,106 @@
+# To test: gh workflow run 'Build Docker Image' --ref kwannoel/workflow-update-branch
+name: Build Docker Image
+on:
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'Tag for the Docker image: vX.Y.Z-<tag>-<commit_sha>'
```
BugenZhao, line 7: I think the term "tag" is already widely recognized in the scope of Docker, to represent the full `vX.Y.Z-foo-1234567`, rather than just `foo`. Can we change to another term to avoid confusion?

.github/workflows/build-docker-image.yml
```diff
@@ -0,0 +1,106 @@
...
+          TAG=${{ github.event.inputs.tag }}
+          echo "TAG=$TAG"
+
+          # Generate the image tag
+          if [[ ${{ github.event.inputs.enable_alpha_build }} == 'true' ]]; then
+            # If alpha build is enabled, use the commit SHA and tag
+            IMAGE_TAG="$COMMIT_SHA-$TAG"
+          else
+            # Use the version from Cargo.toml, commit SHA, and tag
+            IMAGE_TAG="v$VERSION-$TAG-$COMMIT_SHA"
```
BugenZhao, line 93: `COMMIT_SHA` goes first in alpha build while it becomes `TAG` in release build. Shall we normalize this?

BugenZhao (commented):
Generally LGTM

.github/workflows/build-docker-image.yml
kwannoel: Yeah I prefer it to always have the same structure.

.github/workflows/build-docker-image.yml
kwannoel: This is kind of a hack to prevent patching alpha build versions to production cluster in Rw Cloud. Since cloud will do semantic version checks, and alpha build versions will fail that if they use this workflow.

kwannoel: Perhaps we can enhance the check on the cloud side rather than here

.github/workflows/build-docker-image.yml
kwannoel: Fixed, used `git rev-parse` instead.

.github/workflows/build-docker-image.yml
kwannoel: Fixed, I removed both checks. I think it's straightforward to spot `alpha` in the version, and I included `branch` in the image tag as well, so we can verify the source branch elsewhere.

This pipeline will just focus on automating the image tag generation to keep it simple and stupid

kwannoel: https://buildkite.com/risingwavelabs/docker/builds/17282
https://github.com/risingwavelabs/risingwave/actions/runs/14807210370/job/41577305338

.github/workflows/build-docker-image.yml
```diff
@@ -0,0 +1,57 @@
...
+          echo "COMMIT_SHA=$COMMIT_SHA"
+          echo "COMMIT_SHA=$COMMIT_SHA" >> $GITHUB_ENV
+
+          # Get the label from the input
+          LABEL=${{ github.event.inputs.label }}
+          echo "LABEL=$LABEL"
+          echo "LABEL=$LABEL" >> $GITHUB_ENV
+
+          # Build the image tag
+          IMAGE_TAG="v$VERSION--$LABEL--$COMMIT_SHA--$NO_SLASH_BRANCH_NAME"
```
BugenZhao, line 46: Is it expected to have 2 dash here?

kwannoel: Yes, because branch name can also include dash. I want to clearly delimit each segment. So the tag name is machine readable, and further validation can be done.

kwannoel:
Any further comments?

BugenZhao (approved):

xxchan (commented):
Maybe we should also only allow images build from the action here and disallow manual building? (unless with some flag)

PR #14753, opened 2024-01-23
feat(expr): support VARIADIC function arguments

wangrunji0408:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://gist.github.com/TennyZhuang/f00be7f16996ea48effb049aa7be4d66#file-rw_cla).

## What's changed and what's your intention?

resolve #14705

This PR adds VARIADIC argument support for the following functions:

```sql
select format('%s %s', variadic array['Hello', 'World']); -> "Hello World"
select concat(variadic array['abcde', '2', NULL, '22']); -> "abcde222"
select concat_ws(',', variadic array['abcde', 2, NULL, 22] :: varchar[]); -> "abcde,2,22"
select jsonb_build_array(variadic array[1, 2, 4, 5]); -> [1, 2, 4, 5]
select jsonb_build_object(variadic array['foo', '1', '2', 'bar']); -> {"2": "bar", "foo": 1}
select jsonb_extract_path('{"a": {"b": ["foo","bar"]}}', variadic array['a', 'b', '1']); -> bar
select jsonb_extract_path_text('{"a": {"b": ["foo","bar"]}}', variadic array['a', 'b', '1']); -> bar
```

It adds new functions `format_variadic`, `concat_ws_variadic`... on the backend, which take a fixed `anyarray` as the last argument. On the frontend, we rewrite the variadic calls into new functions.

## Checklist

- [x] I have written necessary rustdoc comments
- [x] I have added necessary unit tests and integration tests
- [ ] I have added test labels as necessary. See [details](https://github.com/risingwavelabs/risingwave/blob/main/docs/developer-guide.md#ci-labels-guide).
- [ ] I have added fuzzing tests or opened an issue to track them. (Optional, recommended for new SQL features #7934).
- [ ] My PR contains breaking changes. (If it deprecates some features, please create a tracking issue to remove them in the future).
- [x] All checks passed in `./risedev check` (or alias, `./risedev c`)
- [ ] My PR changes performance-critical code. (Please run macro/micro-benchmarks and show the results.)

- [ ] My PR contains critical fixes that are necessary to be merged into the latest release. (Please check out the [details](https://github.com/risingwavelabs/risingwave/blob/main/CONTRIBUTING.md))

## Documentation

- [x] My PR needs documentation updates. (Please use the **Release note** section below to summarize the impact on users)

## Release note

support VARIADIC arguments for the following functions:

- format
- concat_ws
- jsonb_build_array
- jsonb_build_object
[...]

PR #21640, opened 2025-04-29
feat(expr): implement `array_flatten`

stdrc:
Many query engines/databases have builtin function `array_flatten` or `flatten` to flatten an array of arrays. And it is also requested by our user. This PR adds it for RisingWave.

For the sake of simplicity and clarity, we implement this function following [Snowflake](https://docs.snowflake.com/en/sql-reference/functions/array_flatten?utm_source=chatgpt.com), [Trino](https://trino.io/docs/current/functions/array.html#flatten), [DuckDB](https://database.guide/flattening-nested-lists-with-duckdbs-flatten-function/?utm_source=chatgpt.com), which means, this function works in a non-recursive way, only flattening one level of nesting.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/expr/impl/src/scalar/array_flatten.rs
+++ b/src/expr/impl/src/scalar/array_flatten.rs
@@ -81,10 +81,11 @@ fn array_flatten(array: ListRef<'_>, ctx: &Context) -> Result<ListValue> {
             reason: Box::from("expected the argument to be an array of arrays"),
         });
     }
+    let inner_elem_type = inner_type.as_list();
 
     // Collect all inner array elements and flatten them into a single array
     Ok(ListValue::from_datum_iter(
-        inner_type,
+        inner_elem_type,
         array
             .iter()
             // Filter out NULL inner arrays
--- a/src/expr/impl/src/scalar/array_flatten.rs
+++ b/src/expr/impl/src/scalar/array_flatten.rs
@@ -77,7 +77,7 @@ fn array_flatten(array: ListRef<'_>) -> Result<ListValue> {
         array
             .iter()
             // Filter out NULL inner arrays
-            .filter_map(|inner_array| inner_array)
+            .flatten()
             // Flatten all inner arrays
             .flat_map(|inner_array| {
                 if let ScalarRefImpl::List(inner_list) = inner_array {
--- a/src/expr/impl/src/scalar/array_flatten.rs
+++ b/src/expr/impl/src/scalar/array_flatten.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use itertools::Itertools;
+use itertools::Either;
 use risingwave_common::array::{ListRef, ListValue};
 use risingwave_common::types::ScalarRefImpl;
 use risingwave_expr::{Result, function};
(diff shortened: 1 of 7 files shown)


src/expr/impl/src/scalar/array_flatten.rs
```diff
@@ -0,0 +1,91 @@
...
+/// select array_flatten(null::int[][]);
+/// ----
+/// NULL
+/// ```
+#[function("array_flatten(anyarray) -> anyarray")]
+fn array_flatten(array: ListRef<'_>) -> Result<ListValue> {
+    // The elements of the array must be arrays themselves
+    // Create a new list by flattening all inner arrays
+    let array_data_type = array.data_type();
+    let inner_type = array_data_type.as_list();
```
BugenZhao, line 72: Can we assert that `inner_type` is also an array?

src/expr/impl/src/scalar/array_flatten.rs
```diff
@@ -0,0 +1,91 @@
...
+/// query T
+/// select array_flatten(null::int[][]);
+/// ----
+/// NULL
+/// ```
+#[function("array_flatten(anyarray) -> anyarray")]
+fn array_flatten(array: ListRef<'_>) -> Result<ListValue> {
+    // The elements of the array must be arrays themselves
+    // Create a new list by flattening all inner arrays
+    let array_data_type = array.data_type();
```
BugenZhao, line 71: What about extracting data type from the expression context? `ListRef::data_type` is somehow specific to the list type only and kind of leaky abstraction to me.

BugenZhao (commented):
LGTM

src/expr/impl/src/scalar/array_flatten.rs
```diff
@@ -0,0 +1,91 @@
...
+
+/// Flattens a nested array by concatenating the inner arrays into a single array.
+/// Only the outermost level of nesting is removed. For deeper nested arrays, call
+/// `array_flatten` multiple times.
+///
+/// Examples:
+///
+/// ```slt
+/// query T
+/// select array_flatten(array[array[1, 2], array[3, 4]]);
```
xiangjinwu, line 28: Just realized `array_flatten(arr)` is equivalent to `genuine_array_concat(VARIADIC arr)`. However:
* This hypothetical `genuine_array_concat` is different from PostgreSQL/RisingWave `array_cat`, which has a wired behavior on >= 2d array due to its weak array typing (`int[][]` is same as `int[]`)
* [#14753](https://github.com/risingwavelabs/risingwave/pull/14753) implements `VARIADIC` as separate exprs but in PostgreSQL there is a single `concat_ws` / `jsonb_extract_path`

Just sharing some related ideas. Nothing to change in this PR.

src/expr/impl/src/scalar/array_flatten.rs
```diff
@@ -0,0 +1,91 @@
...
+            .iter()
+            // Filter out NULL inner arrays
+            .flatten()
+            // Flatten all inner arrays
+            .flat_map(|inner_array| {
+                if let ScalarRefImpl::List(inner_list) = inner_array {
+                    Either::Left(inner_list.iter())
+                } else {
+                    // This shouldn't happen but handle it gracefully
+                    Either::Right(std::iter::empty())
```
xiangjinwu, line 87: Maybe too "graceful". If not panic we shall return `Err`

BugenZhao (approved):

PR #21643, opened 2025-04-29
refactor(common): rename `DataType::as_list` to `as_list_element_type`

stdrc:
The inner data of the `DataType::List` variant is quite different from `MapType` in `DataType::Map` and `StructType` in `DataType::Struct`, so the name should be `as_list_element_type` instead of `as_list`. The old name is very confusing.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/common/src/array/list_array.rs
+++ b/src/common/src/array/list_array.rs
@@ -749,13 +749,13 @@ impl ListValue {
                 }
                 self.skip_whitespace();
                 if self.try_consume('}') {
-                    return Ok(ListValue::empty(self.data_type.as_list()));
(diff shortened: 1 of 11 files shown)


BugenZhao:
> The old name is very confusing.

+1 for this. Long awaited.

BTW, have you also considered introducing a new type `List` that wraps the inner element type?

BugenZhao (approved):
LGTM

xiangjinwu (approved):

BugenZhao (approved):

BugenZhao (approved):

PR #21646, opened 2025-04-29
fix(file source): do not persist empty file info in state table

wcy-fdu:

Previously in https://github.com/risingwavelabs/risingwave/pull/21434 we handled the empty file case. In order to clean up empty files in the state table, the reading file will be set when listing. However, we should only get the file name from the hidden column of the chunk and assign it to the reading file when reading the file.

This pr is changed to not store the empty file information in the fetch exectuor's state table.

## Checklist

## Documentation

<details>

</details>


wenym1 (approved):
LGTM as a hotfix.

Discussed offline. This PR is a temporarily hot fix. We will have a cleaner refactor to reset the `reading_file` on `Self::replace_with_new_batch_reader` and avoid reading and writing split offset to state table on every chunk.

PR #21656, opened 2025-04-30
feat(frontend): add a session config to allow disabling sql purification

BugenZhao:

SQL Purification itself may have some bugs (https://github.com/risingwavelabs/risingwave/pull/21638) and may also expose some bugs in parser (https://github.com/risingwavelabs/risingwave/issues/21487).

Let's add a toggle first to allow users to bypass SQL purification and prevent any potential regression, if necessary.

## Checklist

## Documentation

<details>

</details>


src/common/src/session_config/mod.rs
```diff
@@ -392,6 +392,12 @@ pub struct SessionConfig {
     /// The max buffer size for sync logstore, before we start flushing.
     #[parameter(default = 2048_usize)]
     streaming_sync_log_store_buffer_size: usize,
+
+    /// Whether to disable purifying the definition of the table or source upon retrieval.
+    /// Only set this if encountering issues with functionalities like `SHOW` or `ALTER TABLE/SOURCE`.
+    /// This config may be removed in the future.
+    #[parameter(default = false, flags = "NO_ALTER_SYS")]
```
BugenZhao, line 399: Since this is only a workaround, we don't want users to leave it disabled cluster-wide by default, so that we can easily remove it in the future.

xiangjinwu (approved):

kwannoel (approved):

PR #21661, opened 2025-04-30
feat(streaming): support mapping column type in dispatcher

BugenZhao:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## Checklist

- [ ] I have written necessary rustdoc comments.
- [ ]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  I have checked the [Release Timeline](https://github.com/risingwavelabs/rw-commits-history/blob/main/release_timeline.md) and [Currently Supported Versions](https://docs.risingwave.com/changelog/release-support-policy#support-end-dates-for-recent-releases) to determine which release branches I need to cherry-pick this PR into. 

## Documentation

- [ ]  My PR needs documentation updates. 

<details>
<summary><b>Release note</b></summary>

</details>

PR #21662, opened 2025-04-30
feat(meta): consider types when resolve dispatcher output mapping 

BugenZhao:

In last PR (#21661), we added the actual execution logic for transforming types when dispatching. However, that path is unreachable since we never construct `DispatchOutputMapping::TypeMapping`.

This PR implements the meta part to resolve the `output_mapping` using both column IDs and their respective types.

- If all columns have their types matched, we fallback to `Simple` and directly map by indices.
- If there's any type change, we yield a `TypePair` indicating how the column should be transformed.

To persist this information in meta store, a new column is introduced to `fragment_relation` table.

Actually after this PR, `ALTER TABLE ALTER COLUMN TYPE` can function with downstream jobs. However, some additional refinement may still be needed. Will enrich e2e tests progressively.

## Checklist

## Documentation

<details>

</details>


yezizp2012 (approved):
LGTM

PR #21714, opened 2025-05-05
fix: assertion caused by missing check during sink into table recovery

shanicky:

Introduced by PR https://github.com/risingwavelabs/risingwave/pull/20548, we need to skip non-materialized fragments through a conditional check here.

## Checklist


yezizp2012 (approved):

BugenZhao (approved):
Is it possible for us to add recovery tests covering this?

PR #21738, opened 2025-05-06
feat(sqlsmith): support generation of unnest table function

### LINKED ISSUE #4540: sqlsmith: Generate table function
- [x] `generate_series`
- [x] `unnest`
- [ ] `regexp_match`
- [x] `range`
- [x] `jsonb_array_elements`
- [x] `jsonb_array_elements_text`
- [x] `jsonb_each`
- [x] `jsonb_each_text`
- [x] `jsonb_object_keys`

See: https://github.com/risingwavelabs/risingwave/blob/main/src/expr/impl/src/table_function/mod.rs

cardigan1008:

Add support for generating `unnest` table functions in Sqlsmith.

It currently support randomly generation of one single array to be the argument, whose type is also randomly chosen. 

Potential limitations: unnest should support a set of array as the argument according to [doc](https://docs.risingwave.com/sql/functions/array#unnest). But expr [defined here](https://github.com/risingwavelabs/risingwave/blob/main/src/expr/impl/src/table_function/unnest.rs#L20) restricts to one single array.

Fixes #4540 

## Checklist

## Documentation

<details>

</details>


src/tests/sqlsmith/src/sql_gen/table_functions.rs
```diff
@@ -127,6 +146,20 @@ impl<R: Rng> SqlGenerator<'_, R> {
             _ => unreachable!(),
         }
     }
+
+    fn pick_random_scalar_type(&mut self) -> DataType {
```
kwannoel, line 150: You can reuse `sql_gen::expr::gen_list_data_type`

kwannoel (approved):
Generally LGTM, seems like currently we generate a flat array. To provide more coverage, we can consider generating more nested array structures as an extension.

src/tests/sqlsmith/src/sql_gen/table_functions.rs
cardigan1008: Fixed. Thanks for catching this :)

kwannoel (approved):
LGTM!

PR #21749, opened 2025-05-07
docs: add more docs on different keys in RisingWave

### LINKED ISSUE #9942: Write a doc on different types of keys in our database
Currently there are docs about various types of keys. Some of these have overlapping names but slightly different meaning in different context.

E.g. `primary_key` vs `logical_pk` in plan.

We should have a doc for different types of keys and how they relate to each other: storage, streaming etc...
We should also port over some information from quip too.

We can also link existing documentation in the repo within this doc for further reference.

Feel free to append useful docs to this issue.

kwannoel:

Closes #9942 

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/docs/dev/src/design/keys.md
+++ b/docs/dev/src/design/keys.md
@@ -2,6 +2,20 @@
 
 Document the different Keys in RisingWave.
 
+## Order Key
+
+The order key is user-specified, in SQL. For example:
+
+```sql
+create materialized view m1 as select * from t1 order by i, id;
+```
+
+The order key is `i, id`, which is the order of the columns in the `order by` clause.
+
+It is used to ensure locality of records in storage, and locality of record updates in streaming.
+To ensure storage locality, when deriving the storage key, we will use the order key as a prefix.
+To ensure streaming locality, we will use the order key as a prefix of the stream key.
+
 ## Stream Key
 
 The key which can identify records in the RisingWave stream.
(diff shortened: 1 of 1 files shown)


docs/dev/src/design/keys.md
```diff
@@ -2,9 +2,28 @@
...
+The order key is `i, id`, which is the order of the columns in the `order by` clause.
+
+It is used to ensure locality of records in storage, and locality of record updates in streaming.
+To ensure storage locality, when deriving the storage key, we will use the order key as a prefix.
+To ensure streaming locality, we will use the order key as a prefix of the stream key.
+
+## Group Key
+
+Similar to the order key, the group key is also user-specified in SQL.
+We also apply the same derivations of group key in stream key, to ensure storage and streaming locality.
```
xxchan, line 22: I think order key and group key are not really the same level of concept as the following keys.

They are user-facing concepts. And the following are how things work internally.

docs/dev/src/design/keys.md
```diff
@@ -23,15 +42,39 @@ The record corresponding to key `(0, 1)` has been inserted with `(0, 1, 2, 3)`.
...
+## Primary Key
+
+### User Specified Primary Key
+
+When the user specifies primary key in their `create table` statement,
+it is used to ensure that the records are unique in the table.
+We will always use the user-specified primary key as the stream key and storage primary key
+
+For other streaming jobs such as sinks, materialized views and indexes,
+we will use the user-specified primary key in our derivation of the stream key and storage primary key.
```
xxchan, line 67: ditto

xxchan: I think to understand storage primary key, need to first understand what "Table" means.. (quite messy)

A table = TableCatalog = a hummock table = internal state table/ MV (i.e., the state table for MaterializeExec) / ...

docs/dev/src/design/keys.md
```diff
@@ -23,15 +42,39 @@ The record corresponding to key `(0, 1)` has been inserted with `(0, 1, 2, 3)`.
...
 for instance `group key` could be part of the stream key, to specify the distribution of records.
 
-## Primary Key (Storage)
+## Distribution key
+
+This key specifies the distribution of the data streams, as mentioned in [Consistent Hash - Streaming](../design/consistent-hash.md#streaming).
+We want data to be distributed in a way that minimizes data skew,
+and maximizes data locality, for more efficient stateful processing.
+
+To ensure data consistency of updates (U-, U+),
```
BugenZhao, line 51: It doesn't have to be `U-` `U+`. The order of updates in any form for records with the same stream key should be preserved.

docs/dev/src/design/keys.md
```diff
@@ -23,15 +42,39 @@ The record corresponding to key `(0, 1)` has been inserted with `(0, 1, 2, 3)`.
...
+the distribution key must always be a prefix of the stream key.
```
BugenZhao, line 52: I think there's no need to be a prefix? A subset is okay.

docs/dev/src/design/keys.md
```diff
@@ -2,9 +2,28 @@
...
 
+## Order Key
+
+The order key is user-specified, in SQL. For example:
+
+```sql
+create materialized view m1 as select * from t1 order by i, id;
+```
+
+The order key is `i, id`, which is the order of the columns in the `order by` clause.
```
BugenZhao, line 13: To be more precise, the `order_key` also specifies an `OrderType` for each column, including direction (`ASC` `DESC`) and nulls ordering.

docs/dev/src/design/keys.md
kwannoel: Maybe I'll include a section for them inside stream_key and storage_pk?
Since they are used in the derivation step.

xxchan: Indeed, a section about their derivation (maybe with examples) would be nice to have, and the user facing part should also go there.

kwannoel:
Update: I need to address the review comments.

PR #21781, opened 2025-05-08
test: test log store scale-in / out

kwannoel:

Kill some nodes, wait for worker to be expired, which triggers scale-in. Restart the worker nodes after scale-in, to trigger scale-out again.

Finally, check that the data in the logstore is correct.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/config/ci-sim-log-store.toml
+++ b/src/config/ci-sim-log-store.toml
@@ -1,0 +1,10 @@
+[server]
+telemetry_enabled = false
+metrics_level = "Disabled"
+
+[system]
+max_concurrent_creating_streaming_jobs = 0
+
+[meta]
+meta_leader_lease_secs = 10
+max_heartbeat_interval_secs = 5
--- a/src/tests/simulation/tests/integration_tests/log_store/scale.rs
+++ b/src/tests/simulation/tests/integration_tests/log_store/scale.rs
@@ -12,16 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::io::Write;
-use std::sync::Arc;
 use std::sync::atomic::Ordering::Relaxed;
 use std::time::Duration;
 
 use anyhow::Result;
-use itertools::Itertools;
 use madsim::runtime::init_logger;
-use rand::{Rng, rng as thread_rng};
-use risingwave_common::bail;
 use risingwave_common::hash::WorkerSlotId;
 use risingwave_simulation::cluster::{Cluster, ConfigPath, Configuration, KillOpts};
 use risingwave_simulation::ctl_ext::predicate::identity_contains;
@@ -39,7 +34,7 @@ use crate::log_store::utils::*;
 // ./risedev sit-test test_recover_synced_log_store >out.log 2>&1
 // ```
 #[tokio::test]
-async fn test_recover_synced_log_store() -> Result<()> {
+async fn test_scale_in_synced_log_store() -> Result<()> {
     init_logger();
     let mut cluster = start_sync_log_store_cluster().await?;
     cluster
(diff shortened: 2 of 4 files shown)


src/config/ci-sim-log-store.toml
```diff
@@ -0,0 +1,10 @@
+[server]
+telemetry_enabled = false
+metrics_level = "Disabled"
+
+[system]
+max_concurrent_creating_streaming_jobs = 0
+
+[meta]
+meta_leader_lease_secs = 10
+max_heartbeat_interval_secs = 5
```
kwannoel, line 10: Copied from `ci-sim`, but with this additional configuration, to force worker node to be evicted.

src/tests/simulation/tests/integration_tests/log_store/recovery.rs
```diff
@@ -27,27 +27,7 @@ use risingwave_simulation::cluster::{Cluster, ConfigPath, Configuration, KillOpt
 use risingwave_simulation::ctl_ext::predicate::identity_contains;
 use tokio::time::sleep;
```
kwannoel, line 29: Refactored utilies from this file to `utils.rs`

src/config/ci-sim-log-store.toml
wenym1: If this heartbeat value was too large previously, should we set this value directly in the origin `ci-sim` and reuse the same config file so that more corner case can be tested in the integration tests and recovery tests?

wenym1 (approved):

src/config/ci-sim-log-store.toml
kwannoel: Okay, we can change it in a subsequent PR to avoid scope creep

kwannoel: https://github.com/risingwavelabs/risingwave/issues/21792

kwannoel:
* **#21781** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/21781?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/21781?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/tests/simulation/tests/integration_tests/log_store/scale.rs
```diff
@@ -0,0 +1,136 @@
...
+            cluster.wait_for_recovery().await?;
+            tracing::info!("cluster recovered");
+
+            cluster.wait_for_scale(10).await?;
+            assert_parallelism_eq(&mut session, 10).await;
+            assert_lag_in_log_store(&mut cluster, UNALIGNED_MV_NAME, result_count).await?;
+            tracing::info!("cluster scaled back to 10 parallelism");
+
+            run_amplification_workload(&mut cluster, dimension_count).await?;
+            tracing::info!("ran insert amplification workload");
```
kwannoel, line 111: Guarantee the lag in logstore is preserved between test iterations.

kwannoel:
`log_store` tests unexpectedly run under scale tests. Fixing: https://github.com/risingwavelabs/risingwave/pull/22039

BugenZhao:
This increases the CI runtime to 50 minutes.. 😕

kwannoel:
> This increases the CI runtime to 50 minutes.. 😕

You mean main-cron or pull request? I think it's kind of inevitable, given that this case only surfaces under high join amplification workload and requires `sleep` (which may take a long time in madsim). If just for PR tests, I can try to exclude it with regex.

BugenZhao:
> If just for PR tests, I can try to exclude it with regex.

Yeah I'm mainly referring to pull-request. Sounds good.

PR #21786, opened 2025-05-08
refactor(sink): add feature flag for deltalake sink

xxchan:
deltalake has heavy dependency, incl. datafusion. So it would be nice to have an option to disable it to faster compilation.

- Introduced feature flag `sink-deltalake`, and move impl under the gate.
- By default, the flag is enabled, to ensure production build is not affected.
- Added configure option in `risedev configure` to disable default features

related #16841

## Checklist

## Documentation

<details>

</details>


Makefile.toml
```diff
@@ -64,6 +65,12 @@ env_scripts = [
     set_env RISINGWAVE_FEATURE_FLAGS "${flags} --features udf"
   end
 
+  flags = get_env RISINGWAVE_FEATURE_FLAGS
+
+  if ${is_default_features_disabled}
+    set_env RISINGWAVE_FEATURE_FLAGS "${flags} --no-default-features"
+  end
+
```
xxchan, line 73: The feature flags mgmt here is a mess. Wanted to refactor it later

xxchan:
sounds good for the deltalake case. But in the long term I still want to add feature gate for most connectors. https://github.com/risingwavelabs/risingwave/issues/16841

How do you think about the goal and the solution?

src/cmd_all/Cargo.toml
```diff
@@ -31,6 +32,8 @@ risingwave_cmd = { workspace = true }
 risingwave_common = { workspace = true }
 risingwave_compactor = { workspace = true }
 risingwave_compute = { workspace = true }
+# put here just to tweak feature flags
+risingwave_connector = { workspace = true, optional = true }
```
BugenZhao, line 36: No need to make it `optional`?

src/connector/Cargo.toml
```diff
@@ -7,6 +7,12 @@ keywords = { workspace = true }
 license = { workspace = true }
 repository = { workspace = true }
 
+[features]
+default = []
+all-sinks = ["sink-deltalake"]
+all-sources = []
+sink-deltalake = ["deltalake"]
```
BugenZhao, line 14: To be more clear...

```suggestion
sink-deltalake = ["dep:deltalake"]
```

wenym1:
> sounds good for the deltalake case. But in the long term I still want to add feature gate for most connectors. #16841
> 
> So this PR is also like an experiment for that
> 
> How do you think about the goal and the solution?

LGTM. Besides adding the feature flag, we should still avoid enabling the `datafusion` feature of `deltalake-rs` globally, since it affects the actual release build.

Moreover, We have labels `run-e2e-xxx-sink-test` to explicitly enable e2e test for specific sinks. When the label is not selected, the specific sinks will not be run in the CI. So we can also optimize the build time in CI by checking the labels and only including the sinks with label enabled.

src/connector/src/sink/mod.rs
```diff
@@ -1005,8 +1010,9 @@ impl From<ClickHouseError> for SinkError {
...
 }
 
-impl From<DeltaTableError> for SinkError {
-    fn from(value: DeltaTableError) -> Self {
+#[cfg(feature = "sink-deltalake")]
+impl From<::deltalake::DeltaTableError> for SinkError {
+    fn from(value: ::deltalake::DeltaTableError) -> Self {
         SinkError::DeltaLake(anyhow!(value))
     }
 }
```
BugenZhao, line 1018: Shall we move this into `imp.rs`?

src/connector/src/sink/deltalake/mod.rs
BugenZhao, line 1: I think we have to minimize boilerplate code as much as possible to prevent maintenance from becoming more difficult over time.

src/connector/src/sink/deltalake/mod.rs
```diff
@@ -0,0 +1,187 @@
...
+    const ERROR_MESSAGE: &str = "RisingWave is not compiled with feature `sink-deltalake`";
+
+    #[derive(Debug)]
+    pub struct DeltaLakeSinkName;
+
+    impl TrivialSinkName for DeltaLakeSinkName {
+        const SINK_NAME: &'static str = DELTALAKE_SINK;
+    }
+
+    impl Sink for DeltaLakeSink {
```
wenym1, line 148: Shall we make this dummy sink implementation to be more general so that in the future when we want to support optionally disabling more sinks, we can easily reuse this dummy impl?

src/connector/src/sink/mod.rs
xxchan: I think putting it here should be fine

xxchan:
* **#21786** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/21786?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#21810](https://github.com/risingwavelabs/risingwave/pull/21810) <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/21810?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#21811](https://github.com/risingwavelabs/risingwave/pull/21811) <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/21811?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/21786?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

wenym1 (approved):
Rest LGTM

src/connector/src/sink/utils.rs
```diff
@@ -28,3 +28,87 @@ pub fn chunk_to_json(chunk: StreamChunk, encoder: &JsonEncoder) -> Result<Vec<St
...
+
+    pub fn err_feature_not_enabled(sink_name: &'static str) -> SinkError {
+        SinkError::Config(anyhow!(
+            "RisingWave is not compiled with feature `sink-{}`",
+            sink_name
+        ))
+    }
+
+    /// Implement this trait will bring a dummy `impl Sink` for the type which always returns an error.
+    pub trait FeatureNotEnabledSinkMarker:
```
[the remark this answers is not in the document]
xxchan: That violates orphan rule 😇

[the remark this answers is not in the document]
wenym1: If so, maybe we can introduce a `FeatureNotEnabledSink`. Instead of doing `impl<S: FeatureNotEnabledSinkMarker> Sink for S {}`, we can `impl<S: FeatureNotEnabledSinkMarker> Sink for FeatureNotEnabledSink<S> {}`. A trivial `TryFrom`, `Debug` and `EnforceSecret` can also be impl for `FeatureNotEnabledSink<S>`.

In the `deltalake/mod.rs`, it can be as simple as 
```
pub const DELTALAKE_SINK: &str = "deltalake";

cfg_if::cfg_if! {
    if #[cfg(feature = "sink-deltalake")] {
        mod imp;
        pub use imp::DeltaLakeSink;
    } else {
        use crate::sink::utils::dummy::{FeatureNotEnabledSinkMarker, FeatureNotEnabledSink};
        pub struct DeltalakeNotEnabled;
        impl FeatureNotEnabledSinkMarker for DeltalakeNotEnabled {
            const SINK_NAME: &'static str = super::DELTALAKE_SINK;
        }
        pub type DeltaLakeSink = FeatureNotEnabledSink<DeltalakeNotEnabled>;
    }
}
```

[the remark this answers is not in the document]
wenym1: Just pushed a new commit 95896c5a882ba5d161b973e09d80672bd2193465 for the suggested solution. Feel free to revert the commit if you prefer the original implementation.

BugenZhao (approved):

Issue #21788, opened 2025-05-08
Iceberg table engine error with Postgres CDC auto schema change


n14zio:
### Describe the bug

See steps to reproduce.
Use case: Correctly replicate Postgres table in Iceberg, with schema evolution.

### Error message/log

```text
SQL Error [XX000]: ERROR: Failed to execute the statement

Caused by these errors (recent errors listed first):
  1: failed to auto derive table schema
  2: Postgres table should define the primary key for non-append-only tables


Error position:
```

### To Reproduce

**Postgres Table DDL**
` CREATE TABLE ingestion.transactions (
	id bigserial NOT NULL,
	merchant_id int4 NOT NULL,
	amount numeric(12, 2) NULL,
	created_at timestamp NOT NULL,
	last_updated_at timestamp NULL,
	new_col int4 NULL,
	CONSTRAINT transactions_pkey PRIMARY KEY (id)
);
`

**RW Source Creation**
`
CREATE SOURCE ingestion.api_db_ingestion_auto_schema WITH (
    connector='postgres-cdc',
    hostname='redacted.eu-west-1.rds.amazonaws.com',
    port='5432',
    username='redacted',
    password='redacted',
    database.name='redacted',
    schema.name='ingestion', 
    auto.schema.change = 'true'
);
`

**RW Connection creation for Iceberg**
`CREATE CONNECTION public.iceberg_conn WITH (
    type = 'iceberg',
    catalog.type='rest',
    catalog.uri= 'xxxx...',
    etc.
);
SET iceberg_engine_connection = 'public.iceberg_conn';
ALTER system SET iceberg_engine_connection = 'public.iceberg_conn';
`

**RW Table**
`
CREATE TABLE ingestion.pg_transactions_auto_internal_stg  (*) 
FROM ingestion.api_db_ingestion_auto_schema TABLE 'ingestion.transactions'
ENGINE = iceberg;     
`

### How did you deploy RisingWave?

Helm chart

### The version of RisingWave

PostgreSQL 13.14.0-RisingWave-2.3.0 (e8f053e09651e3f956a9944063dccc746794dfe2)

chenzl25:
@n14zio May I ask whether it failed to create a table with the iceberg engine in this case, even if you didn't do any schema change? We found a bug, and we are working on fixing it. Related PR: [fix(iceberg): fix iceberg table engine pk retrieving #21789](https://github.com/risingwavelabs/risingwave/pull/21789)

[...]

PR #21789, opened 2025-05-09
fix(iceberg): fix iceberg table engine pk retrieving

chenzl25:

- Retrieve the pk information from the table proto instead of column constraints, because column constraints could be empty when users' sql doesn't explicitly specify it.
- Related issue. https://github.com/risingwavelabs/risingwave/issues/21788

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/handler/create_table.rs
+++ b/src/frontend/src/handler/create_table.rs
@@ -1423,8 +1423,6 @@ pub async fn handle_create_table(
                 table,
                 graph,
                 job_type,
-                column_defs,
-                constraints,
                 table_name,
             )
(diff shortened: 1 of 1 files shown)


Li0k (approved):

xxhZs (approved):

BugenZhao (commented):
Can we have an e2e test for this?

chenzl25:
> Can we have an e2e test for this?

Considering adding more connector tests for the iceberg engine, but probably in the following PRs. This fix needed to be cherrpick to v2.3 and v2.4.

PR #21802, opened 2025-05-09
feat(iceberg): support interval type for iceberg engine

chenzl25:

- Support interval type for the iceberg engine. RisingWave has an interval type, however, iceberg specification doesn't have an interval type, so when users want to create an iceberg engine table with an internal type, we need to convert the internal type to a string type in iceberg and convert it back to the interval type from the iceberg read path.

## Checklist

## Documentation

<details>

</details>


src/common/src/array/arrow/arrow_iceberg.rs
```diff
@@ -102,7 +104,7 @@ impl ToArrow for IcebergArrowConvert {
             DataType::Time => self.time_type_to_arrow(),
             DataType::Timestamp => self.timestamp_type_to_arrow(),
             DataType::Timestamptz => self.timestamptz_type_to_arrow(),
-            DataType::Interval => self.interval_type_to_arrow(),
+            DataType::Interval => self.varchar_type_to_arrow(),
```
xiangjinwu, line 107: What are the pros and cons of doing it this way compared to overriding `<IcebergArrowConvert as ToArrow>::interval_type_to_arrow` to return `arrow_schema::DataType::Utf8`?

src/common/src/array/arrow/arrow_iceberg.rs
```diff
@@ -168,6 +170,13 @@ impl ToArrow for IcebergArrowConvert {
             .map_err(ArrayError::from_arrow)?;
         Ok(Arc::new(array) as ArrayRef)
     }
+
+    fn interval_to_arrow(
```
xiangjinwu, line 174: Out of curiosity, why is it unnecessary to have a counterpart in the `FromArrow` implementation below to convert from varchar back to interval?

xiangjinwu (commented):
No objection to the current implementation, but have we considered labeling it as an extension type similar to https://github.com/arrow-udf/arrow-udf/pull/17? Although technically it is unnecessary to extend arrow - we are extending iceberg's way of using arrow.

src/common/src/array/arrow/arrow_iceberg.rs
chenzl25: Because once we convert an interval type to varchar type, we can't tell what this varchar arrow array solely should be converted to. Currently, we will cast this varchar array on the operator (using a project to cast the varchar back to interval) level.

src/common/src/array/arrow/arrow_iceberg.rs
chenzl25: I think there is no difference.

xiangjinwu: Then I would prefer updating `interval_type_to_arrow`, to prevent from accidentally calling it. The current implementation means its functionality is broken and should never be called.

It also means keeping `interval_type_to_arrow` and `interval_to_arrow` consistent.

src/common/src/array/arrow/arrow_iceberg.rs
xiangjinwu: > we can't tell what this varchar arrow array solely should be converted to

Sounds like the extension type annotation can help. But we do not have to go this way if that is a big refactor and has little other benefits.

PR #21803, opened 2025-05-09
feat: introduce new scheduling algorithm

shanicky:

This PR attempts to introduce a new scheduling algorithm, primarily aimed at ensuring the affinity between vnodes and workers during multiple scheduling/rescheduling processes.

This PR attempts to use weighted rendezvous hashing for scheduling. [Rendezvous hashing](https://en.wikipedia.org/wiki/Rendezvous_hashing) is a algorithm similar to consistent hashing, used to assign items to specific workers.

Currently, our algorithm only guarantees affinity between vnodes and actors. This is no longer sufficient for the current scenarios. As more clusters utilize disk caches or other block caches, we need to ensure that vnodes maintain their affinity with workers across multiple reschedules as much as possible.

Furthermore, in our future plans to deprecate the meta store's actor table, a stable and evenly distributed algorithm that can guarantee affinity is essential. This PR aims to replace the existing simple scheduling algorithm with a more robust approach.

Currently, we have two extreme scenarios:

1. A large number of streaming jobs with many actors, but each streaming job has relatively low traffic. In this case, we can use a less uniform algorithm that ensures overall fairness based on the number of jobs, allowing us to focus more on maintaining affinity.  
2. Very few streaming jobs with few actors, but each job carries massive traffic. Under these circumstances, even slight imbalance can lead to significant traffic skew, so we need to ensure perfect uniformity in scheduling.

This PR specifically introduces a multi-level scheduling pattern involving actor -> vnode -> worker:

1. Assigns actors to workers using rendezvous hashing, distributed according to the workers' weights.  
2. Assigns vnodes to workers based on either the number of actors (acting as weight) or the original worker weights.  
3. Assigns vnodes to actors.

This PR still requires more testing to confirm corner cases and needs additional time to evaluate scheduling issues under different workloads.

### Allocation Matrix: `CapacityMode` × `BalancedBy`

| **Actor → Worker (`CapacityMode`) \ VNode → Worker (`BalancedBy`)** | **RawWorkerWeights**<br><sub>VNode weight = original worker weight</sub> | **ActorCounts**<br><sub>VNode weight = #actors on each worker</sub> |
| --- | --- | --- |
| **Weighted**<br><sub>Actors capped by worker weight</sub> | *Actor stage* strictly respects weights (e.g. W1:W2 = 1:3 → actors ≈ 1:3).<br>*VNode stage* again uses original weights (with a lower-bound of actor count).<br>Result: heavier workers get **more VNodes *per actor*** – machine-oriented load. | *Actor stage* still follows raw weights.<br>*VNode stage* uses actor counts as weights, so each actor gets roughly the **same number of VNodes**.<br>Good when you want per-actor fairness but still honor worker capacities. |
| **Unbounded**<br><sub>No actor cap; rendezvous hash only</sub> | *Actor stage* is weight-biased but unbounded – a light worker can still win actors via hashing.<br>*VNode stage* keeps raw weights, so a worker that “lucks out” with many actors might end up **CPU heavy but VNode light**. | *Actor stage* is free + biased.<br>*VNode stage* mirrors actor counts, giving near-equal VNodes per actor.<br>This combo yields the **fairest per-actor load** while letting hashing absorb cluster churn. |

#### Cheat-Sheet

* **CapacityMode::Weighted**  
  * Hard cap: actors ≤ worker weight (scale = 1).  
  * Highly predictable, mirrors hardware sizing.

* **CapacityMode::Unbounded**  
  * Actors limited only by rendezvous hash probability.  
  * Useful when worker capacity is unknown or elastic.

* **BalancedBy::RawWorkerWeights**  
  * VNodes skew toward heavier machines; each actor on a “big” worker handles more VNodes.  
  * Best when storage/IO throughput truly scales with weight.

* **BalancedBy::ActorCounts**  
  * VNodes proportional to actor counts, so every actor sees similar load.  
  * Ideal when **actor-level fairness** matters more than raw machine weight.

## Checklist

### DIFF
--- a/src/meta/src/stream/stream_graph/assignment.rs
+++ b/src/meta/src/stream/stream_graph/assignment.rs
@@ -25,7 +25,7 @@ use anyhow::{Context, Result, anyhow, ensure};
 /// using a three-phase algorithm:
 ///
 /// # Type Parameters
-/// - `C`: Container identifier. Must implement `Ord + Hash + Eq + Copy + Debug`.
+/// - `C`: Container identifier. Must implement `Ord + Hash + Eq + Clone + Debug`.
 /// - `I`: Item type. Must implement `Hash + Eq + Copy + Debug`.
 /// - `S`: Salt type for tie-breaking. Must implement `Hash + Copy`.
 ///
@@ -100,7 +100,7 @@ pub fn assign_items_weighted_with_scale_fn<C, I, S>(
     capacity_scale_factor_fn: impl Fn(&BTreeMap<C, NonZeroUsize>, &[I]) -> Option<ScaleFactor>,
 ) -> BTreeMap<C, Vec<I>>
 where
-    C: Ord + Hash + Eq + Copy + Clone + Debug,
+    C: Ord + Hash + Eq + Clone + Debug,
     I: Hash + Eq + Copy + Clone + Debug,
     S: Hash + Copy,
 {
@@ -116,15 +116,15 @@ where
         "Sum of container weights must be non-zero"
     );
 
-    struct QuotaInfo<C> {
-        container: C,
+    struct QuotaInfo<'a, C> {
+        container: &'a C,
         quota: usize,
         rem_part: u128,
     }
 
-    let mut infos: Vec<QuotaInfo<C>> = containers
+    let mut infos: Vec<QuotaInfo<'_, C>> = containers
         .iter()
-        .map(|(&container, &weight)| {
+        .map(|(container, &weight)| {
             // Use saturating multiplication to prevent overflow, even though saturation is highly unlikely in practice.
             let ideal_num = (items.len() as u128).saturating_mul(weight.get() as u128);
             QuotaInfo {
@@ -150,7 +150,7 @@ where
 
     // Apply capacity scaling
     let scale_factor = capacity_scale_factor_fn(containers, items);
-    let quotas: HashMap<C, usize> = infos
+    let quotas: HashMap<&C, usize> = infos
         .into_iter()
         .map(|info| match scale_factor {
             Some(f) => {
@@ -171,12 +171,12 @@ where
 
     // Assign each item using Weighted Rendezvous
     for &item in items {
-        let mut best: Option<(C, f64)> = None;
-        for (&container, &weight) in containers {
-            let assigned = assignment.get(&container).map(Vec::len).unwrap_or(0);
+        let mut best: Option<(&C, f64)> = None;
+        for (container, &weight) in containers {
+            let assigned = assignment.get(container).map(Vec::len).unwrap_or(0);
 
-            debug_assert!(quotas.contains_key(&container));
-            let quota = quotas.get(&container).copied().unwrap_or(0);
+            debug_assert!(quotas.contains_key(container));
+            let quota = quotas.get(container).copied().unwrap_or(0);
             if assigned >= quota {
                 continue;
             }
@@ -203,7 +203,7 @@ where
         // quotas sum (possibly scaled) always >= items.len(), so best is always Some
         let (winner, _) = best.expect("Invariant violation: no eligible container");
         assignment
-            .entry(winner)
+            .entry(winner.clone())
             .and_modify(|v| v.push(item))
             .or_insert_with(|| vec![item]);
     }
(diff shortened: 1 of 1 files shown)


src/meta/src/stream/stream_graph/assignment.rs
```diff
@@ -0,0 +1,2127 @@
...
+    Weighted,
+
+    /// Ignore per-container quotas entirely—every container can take an unlimited number of items.
+    Unbounded,
+}
+
+/// Defines the vnode distribution strategy for hierarchical assignment.
+///
+/// - `RawWorkerWeights`: Distribute vnodes across workers using the original worker weight values.
+/// - `ActorCounts`: Distribute vnodes based on the number of actors assigned to each worker.
```
BugenZhao, line 278: Am I understanding it correctly: if there are a lot of workers/actors in the cluster, the results for different variant should be quite similar in theory, since the number of actors assigned should follow the weights?

src/meta/src/stream/stream_graph/assignment.rs
```diff
@@ -0,0 +1,2127 @@
...
+        let result = assign_items_weighted_with_scale_fn(&containers, &items, 5u8, weighted_scale);
+        let x_count = result.get(&"X").map(Vec::len).unwrap_or(0);
+        let y_count = result.get(&"Y").map(Vec::len).unwrap_or(0);
+        assert_eq!(x_count + y_count, items.len(), "All items must be assigned");
+        assert!(
+            (x_count == 2 && y_count == 1) || (x_count == 1 && y_count == 2),
+            "One container should get 2 items, the other 1, but got {} and {}",
+            x_count,
+            y_count
+        );
```
BugenZhao: I think the results should be deterministic? Shall we simply demonstrate the expected result in the tests?

src/meta/src/stream/stream_graph/assignment.rs
shanicky: Yes, if the cluster is large enough, these two are roughly the same.

It only needs careful consideration when the number of actors is very small but the traffic is very high, to prevent traffic skew.

src/meta/src/stream/stream_graph/assignment.rs
shanicky: Yes, I'll make a change.

BugenZhao (approved):
Good to see that this new algorithm provides minimal disruption when rescheduling while being stateless!

1. Can you demonstrate which strategy (in the matrix) should be used under different circumstances?
2. The only drawback I can think of is that, vnode assigned to the same actor/worker node is now completely discrete (not continuous). As a result, we cannot effectively leverage the `high_ranges` to merge iterators when performing range scans, and the performance when there's a relatively small amount of data might be degraded.

shanicky:
> Good to see that this new algorithm provides minimal disruption when rescheduling while being stateless!
> 
> 1. Can you demonstrate which strategy (in the matrix) should be used under different circumstances?
> 2. The only drawback I can think of is that, vnode assigned to the same actor/worker node is now completely discrete (not continuous). As a result, we cannot effectively leverage the `high_ranges` to merge iterators when performing range scans, and the performance when there's a relatively small amount of data might be degraded.

I came up with a solution. Is it possible to split vnodes into chunks in certain situations? For example, with 64 vnodes and 16 actors, dividing every 4 vnodes into a vnode chunk would ensure that the vnodes within each actor are contiguous. This should work, but it only guarantees that the vnodes within a worker are not too scattered.

We could even go further by chunking the actors, which might ensure that both the vnodes on the actors and on the workers are ordered.

This way, in some cases, we can trade off some migration affinity (since the basic unit of migration shifts from a vnode to multiple vnodes) to achieve contiguous vnodes.

I'll give it a try.

shanicky:
The vnode chunking mode has already been implemented. You can find more information by checking `VnodeChunkingStrategy`, and you can also refer to the unit test `test_maximize_contiguity_non_divisible_vnodes`. cc @BugenZhao

BugenZhao:
> The vnode chunking mode has already been implemented. You can find more information by checking `VnodeChunkingStrategy`, and you can also refer to the unit test `test_maximize_contiguity_non_divisible_vnodes`. cc @BugenZhao

Cool!

src/meta/src/stream/stream_graph/assignment.rs
```diff
@@ -0,0 +1,2497 @@
...
+///    - If `ActorCounts`: set each worker’s weight = its actor count.
+///    - Run `assign_items_weighted_with_scale_fn` on vnodes vs. the computed weights,
+///      yielding `vnode_to_worker: BTreeMap<W, Vec<V>>`.
+///
+/// 3. **`VNodes` → Actors**
+///    - For each worker, take its vnode list and assign them to actors in simple round-robin:
+///      iterate vnodes in order, dispatching index `% actor_list.len()`.
+///    - Collect into final `BTreeMap<W, BTreeMap<A, Vec<V>>>`.
+pub fn assign_hierarchical<W, A, V, S>(
+    workers: &BTreeMap<W, NonZeroUsize>,
```
yezizp2012, line 383: Due to workers now going offline will actively unregister itself, this will cause the worker id to change. In this case, unnecessary vnode migration might occur IIUC. Is it possible to use `advertise addr` + `port` as the parameter?

yezizp2012: Cc @BugenZhao WDYT?

shanicky:
The worker parameters have already been updated to support Clone. 

I plan to merge this PR first.

yezizp2012 (approved):

PR #21813, opened 2025-05-11
feat(frontend): plan backfill order strategy after the planner phase, instead of binding phase

kwannoel:

This PR is in preparation for https://github.com/risingwavelabs/risingwave/pull/21814. To auto-derive backfill order, we need the stream plan, which is produced only after the planning phase.

So we do two things here:
1. Refactor the planning for backfill order strategy after binding.
2. Move the module for backfill order strategy planning into the **optimizer**. This makes more sense, since this is no longer purely binding user input. Rather we are optimizing the backfill order.

We intentionally keep this PR separate from #21814, since we refactor file paths. To keep diffs clean, we should have the files in separate PRs.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/optimizer/backfill_order_strategy.rs
+++ b/src/frontend/src/optimizer/backfill_order_strategy.rs
@@ -35,7 +35,7 @@ use crate::session::SessionImpl;
 /// We only bind tables, materialized views and sources.
 /// Queries won't bind duplicate relations in the same query context.
 /// But backfill order strategy can have duplicate relations.
-pub fn bind_backfill_order_strategy(
+pub fn plan_backfill_order_strategy(
     session: &SessionImpl,
     backfill_order_strategy: BackfillOrderStrategy,
 ) -> Result<PbBackfillOrderStrategy> {
--- a/src/frontend/src/binder/mod.rs
+++ b/src/frontend/src/binder/mod.rs
@@ -25,7 +25,6 @@ use risingwave_sqlparser::ast::{Expr as AstExpr, SelectItem, SetExpr, Statement}
 
 use crate::error::Result;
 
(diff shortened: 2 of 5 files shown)


kwannoel:
* **#21814** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/21814?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#21813** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/21813?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/21813?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/frontend/src/optimizer/backfill_order_strategy.rs
```diff
@@ -35,7 +35,7 @@ use crate::session::SessionImpl;
 /// We only bind tables, materialized views and sources.
 /// Queries won't bind duplicate relations in the same query context.
 /// But backfill order strategy can have duplicate relations.
-pub fn bind_backfill_order_strategy(
+pub fn plan_backfill_order_strategy(
```
chenzl25, line 38: Need to adjust the comments above?

chenzl25 (approved):
LGTM

src/frontend/src/optimizer/backfill_order_strategy.rs
kwannoel: Will fix it in the downstream PR, otherwise leads to a lot of conflicts

PR #21820, opened 2025-05-12
impl

wenym1:

## Checklist

## Documentation

<details>

</details>


wenym1:
* **#22190** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22190?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#22022** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22022?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#21648** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/21648?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#21820** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/21820?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/21820?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#22103** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22103?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/storage/src/hummock/event_handler/hummock_event_handler.rs
```diff
@@ -785,6 +785,21 @@ impl HummockEventHandler {
                         error!("unable to send get_min_uncommitted_sst_id result: {:?}", e);
                     });
             }
+            HummockEvent::RegisterVectorWriter {
+                table_id,
+                init_epoch,
+            } => self.uploader.register_vector_writer(table_id, init_epoch),
```
hzxa21, line 785: We can only have one vector writer per table_id, right? In other words, the vector wrtier is single parallelism.

src/storage/src/hummock/vector/writer/mod.rs
```diff
@@ -114,9 +113,10 @@ impl VectorWriterImpl {
         }))
     }
 
-    #[expect(dead_code)]
     pub(crate) async fn flush(&mut self) -> HummockResult<usize> {
-        if let Some((file_info, _blocks, _meta)) = self.vector_file_builder.finish().await? {
+        if let Some((file_info, blocks, meta)) = self.vector_file_builder.finish().await? {
```
hzxa21, line 117: It seems that the flush for vector data is synchornous. Will it blocks streaming or the `flush` method is called on barrier collection instead of inside the streaming executor?

src/storage/src/hummock/event_handler/hummock_event_handler.rs
```diff
@@ -785,6 +785,21 @@ impl HummockEventHandler {
...
+            HummockEvent::VectorWriterSealEpoch {
+                table_id,
+                next_epoch,
+                add,
```
hzxa21, line 789: Since `add` is provided here, uploader is not responsible for the actual uploading. Do we still need to make uploader be aware of vector?

src/storage/src/hummock/event_handler/hummock_event_handler.rs
wenym1: Yes, it's singleton.

src/storage/src/hummock/vector/writer/mod.rs
wenym1: This is by design for simplicity on initial implementation. In this future we can further optimize it to asynchronous flush on uploader.

src/storage/src/hummock/event_handler/hummock_event_handler.rs
wenym1: The `add` should be carried in the result of `sync`. Currently, the `tx` of `SyncResult` is passed to the uploader, and after the upload tasks of the sync have finished, inside the uploader we will send the `SyncResult` via the `tx`. Therefore it's simpler to pass the `add` to `uploader` so that it can carry the `add` in the `SyncResult`.

hzxa21 (approved):

PR #21823, opened 2025-05-12
feat(iceberg): add iceberg compactor in risingwave

xxhZs:

in this pr ,we support iceberg compacor depended on https://github.com/nimtable/IC/pulls/

Moreover, I noticed that none of the versions of ​45 are built on DataFusion. Therefore, we currently need to maintain two separate DataFusion instances, and we will consolidate them once the ​Iceberg Rust implementation is upgraded to support integration.

## Checklist

## Documentation

<details>

</details>


src/common/Cargo.toml
```diff
@@ -30,7 +30,7 @@ chrono = { version = "0.4", default-features = false, features = [
     "clock",
     "std",
 ] }
-chrono-tz = { version = "0.10", features = ["case-insensitive"] }
+chrono-tz = { version = "0.9", features = ["case-insensitive"] }
```
xiangjinwu, line 33: There is a better workaround than downgrading the version. I am going to include it in a version bump PR soon.

https://gist.github.com/xiangjinwu/65c310adf84c381c17d99cb4439e1908

src/common/src/hash/key.rs
```diff
@@ -80,7 +80,7 @@ const_assert_eq!(
 
 const_assert_eq!(
     std::mem::size_of::<HeapNullBitmap>(),
-    std::mem::size_of::<usize>() * 4,
+    std::mem::size_of::<usize>() * 3,
```
xiangjinwu, line 83: Already handled by #21853

src/common/Cargo.toml
xiangjinwu: #21860

src/storage/src/hummock/compactor/mod.rs
```diff
@@ -284,6 +419,7 @@ pub fn start_compactor(
     hummock_meta_client: Arc<dyn HummockMetaClient>,
     sstable_object_id_manager: Arc<SstableObjectIdManager>,
     compaction_catalog_manager_ref: CompactionCatalogManagerRef,
+    compactor_type: CompactorType,
```
Li0k, line 422: how about reuse the CompactorMode?

Li0k: We can follow the design of this [PR](https://github.com/risingwavelabs/risingwave/pull/21818/files) and directly determine the startup mode via `CompactorMode`.

src/storage/src/hummock/compactor/mod.rs
```diff
@@ -359,21 +522,11 @@ pub fn start_compactor(
 
                 let running_task_parallelism = running_task_parallelism.clone();
                 let request_sender = request_sender.clone();
-                let event: Option<Result<SubscribeCompactionEventResponse, _>> = tokio::select! {
+                let event: Option<Result<CompactionEventResponses, _>> = tokio::select! {
```
Li0k, line 525: I'm still considering whether we should use the same loop to dispatch both Iceberg and Hummock logic. The reasons are: the workflows for these two services might be completely different. For example:  
1. Heartbeat behavior  
2. Tick intervals  
3. Whether to execute reporting  
4. Runner behavior  
5. Resource estimation and limits  
...and so on.


Of course, the above-mentioned refactoring does not need to be implemented in a single PR. The current workaround is already sufficient to support the ver 0.1 Iceberg compactor service.

 @hzxa21 Any suggestion?

src/storage/src/hummock/iceberg_compactor_runner.rs
```diff
@@ -0,0 +1,118 @@
...
+                    iceberg::spec::DataContentType::Data => {
+                        all_data_file_size += i.data_file().file_size_in_bytes() as u32;
+                    }
+                    iceberg::spec::DataContentType::EqualityDeletes => {}
+                    iceberg::spec::DataContentType::PositionDeletes => {}
+                }
+            }
+        }
+
+        Ok((all_data_file_size / MAX_SIZE_PER_PARTITION) + 1)
```
Li0k, line 85: That's incorrect. When exact division occurs, the calculated number of partitions will be 1 more than expected.

```rust
let partitions = all_data_file_size.div_ceil(MAX_SIZE_PER_PARTITION);
```

src/storage/src/hummock/iceberg_compactor_runner.rs
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    pub async fn compact_iceberg(self, shutdown_rx: Receiver<()>, parallelism: usize) {
+        let compact = async move {
+            let compaction_config = Arc::new(CompactionConfig {
```
Li0k, line 90: We must consider `target_file_size`, or add a doc and implement it in the next PR.

src/storage/src/hummock/iceberg_compactor_runner.rs
```diff
@@ -0,0 +1,118 @@
...
+                batch_parallelism: Some(parallelism),
+                target_partitions: Some(parallelism),
+                data_file_prefix: None,
+            });
+            let compaction = Compaction::new(compaction_config, self.catalog);
```
Li0k, line 95: From my personal perspective, it would be clearer to have the runtime return the result and let the compactor (caller) perform the commit. Is there any additional context here? @chenzl25

src/storage/src/hummock/compactor/mod.rs
```diff
@@ -424,40 +566,77 @@ pub fn start_compactor(
...
+                                    .fetch_add(parallelism, Ordering::SeqCst);
+                                executor.spawn(async move {
+                                    let (tx, rx) = tokio::sync::oneshot::channel();
+                                    shutdown.lock().unwrap().insert(task_id, tx);
+
+                                    icebert_runner.compact_iceberg(
+                                        rx,
+                                        parallelism as usize,
+                                    )
+                                    .await;
```
chenzl25, line 625: I think we should run expire snapshot here as well, otherwise, there will be more and more snapshots.

src/storage/src/hummock/iceberg_compactor_runner.rs
xxhZs: I don't think this is a good approach. Both committing and file writing are integral parts of compaction. Separating these two interrelated operations into different repositories is not advisable.

src/storage/src/hummock/compactor/mod.rs
xxhZs: "I think executing them together is acceptable. Apart from some differences in specific computational implementations, the logic layers are consistent. What Hummock has, Iceberg should also have in the future.

src/storage/src/hummock/compactor/mod.rs
```diff
@@ -424,40 +564,77 @@ pub fn start_compactor(
...
+                            Some(event) => event,
+                            None => continue 'consume_stream,
+                        };
+                        // todo!: add metrics
+                        let shutdown = shutdown_map.clone();
+                        match event {
+                            risingwave_pb::iceberg_compaction::subscribe_iceberg_compaction_event_response::Event::CompactTask(iceberg_compaction_task) => {
+                                let task_id = iceberg_compaction_task.task_id;
+                                let(parallelism,icebert_runner) = match async move {
+                                    let icebert_runner = IcebergCompactorRunner::new(
```
chenzl25, line 584: typo

src/storage/src/hummock/compactor/mod.rs
chenzl25: Considering the iceberg comapctor just handling pull task event, let's make this logic decoupled with the hummock at the very beginning. It would be much clearer and easier to evolve.

src/meta/src/manager/iceberg_compaction.rs
```diff
@@ -112,7 +114,11 @@ impl IcebergCompactionHandle {
         }
     }
 
-    pub async fn send_compact_task(mut self, compactor: Arc<IcebergCompactor>) -> MetaResult<()> {
+    pub async fn send_compact_task(
+        mut self,
+        compactor: Arc<IcebergCompactor>,
+        env: &MetaSrvEnv,
```
Li0k, line 120: env -> task_id ?

MrCroxx (commented):
The tonic 0.11 seems not related? Could you please help check it?

MrCroxx (commented):
Checked again. Imo the none of iceberg-related dependencies updates are intentional?

MrCroxx (approved):

PR #21824, opened 2025-05-12
feat(compaction): introduce iceberg compaction pull strategy

Li0k:

introduce iceberg compaction pull strategy base on `CommitInfo`

- support `handle_pull_task_event` function for `IcebergCompactionEventDispatcher`
- support compaction trigger with fixed interval

tips:  Whether compaction is triggered needs to be supported by `sink-param`. The pr will be initiated to support compaction settings per sink

## Checklist

## Documentation

<details>

</details>


chenzl25 (approved):
LGTM！

src/meta/src/manager/iceberg_compaction.rs
```diff
@@ -97,20 +102,47 @@ impl IcebergCompactionManager {
...
+        // For now, we use a fixed interval of 1 hour
+        const MIN_SINK_COMMIT_INTERVAL: u64 = 3600; // 1 hour
+
+        sink_ids
             .iter()
-            .max_by_key(|&(_, count)| count)
-            .map(|(sink_id, _)| *sink_id)
+            .filter(|(_, commit_info)| {
+                Instant::now().duration_since(commit_info.first_commit_time)
+                    > std::time::Duration::from_secs(MIN_SINK_COMMIT_INTERVAL)
```
hzxa21, line 139: After the initial 1 hour is passed since the sink first commit, are we going to aggressively trigger compaction? I thought we will still trigger compaction based on some interval instead of every time `get_top_n_iceberg_commit_sink_ids` is called..

Li0k: https://github.com/risingwavelabs/risingwave/blob/af755dff8be2784c7966deb95f5e982f2d581d7e/src/meta/src/hummock/manager/compaction/compaction_event_loop.rs#L664

You’re right, so we need to actively clean up this record after sending the msg to avoid repeated execution. I don’t think this is very intuitive. 
I also considered recording **last_compaction_ts**, but this still depends on the external call.

Li0k: PTAL @hzxa21

hzxa21 (approved):

src/meta/src/manager/iceberg_compaction.rs
```diff
@@ -43,8 +47,123 @@ type CompactorChangeTx = UnboundedSender<(u32, Streaming<SubscribeIcebergCompact
...
+}
+
+pub struct IcebergCompactionHandle {
+    sink_id: SinkId,
+    inner: Arc<RwLock<IcebergCompactionManagerInner>>,
+    metadata_manager: MetadataManager,
+    handle_success: bool,
+
+    /// The commit info of the iceberg compaction handle for recovery.
+    commit_info: Option<CommitInfo>,
```
chenzl25, line 90: If it is always a `Some`, we can use clone() directly.

chenzl25 (approved):

xxhZs (approved):

PR #21353, opened 2025-04-11
fix(ci): remove e2e tests with in-memory state store

BugenZhao:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## What's changed and what's your intention?

There are several reasons why we may want to remove the `in-memory` tests:

1. It was added to the CI pipeline a long time ago, when Hummock was not stable. At that time, it helped us determine if the error was caused by the storage module. I believe this is unnecessary now.

2. It uses a `BTreeMap` as the backend, which is quite inefficient and not used anywhere else. For example, `risedev d default` and `risedev p` uses in-memory Hummock, which is still Hummock backend, but with in-memory object store.

3. Another point of this configuration is that, we can test that one compute node may not access the state written (or maintained) by another compute node in most cases. This assumption is much weaker now, as scaling and shared arrangement (like arrangement backfill) is more widely adopted. Also, since #20665, there's only 1 compute node in this configuration, making this completely meaningless.

4. As there are more and more usability features, we start to have a lot of `skipif in-memory` hacks in SLT files, which is a burden for maintainence.

5. It has been timing out frequently lately, though I'm still unsure about the cause.

## Checklist

- [x] I have written necessary rustdoc comments.
- [ ]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  I have checked the [Release Timeline](https://github.com/risingwavelabs/rw-commits-history/blob/main/release_timeline.md) and [Currently Supported Versions](https://docs.risingwave.com/changelog/release-support-policy#support-end-dates-for-recent-releases) to determine which release branches I need to cherry-pick this PR into. 

## Documentation

- [ ]  My PR needs documentation updates. 

<details>
<summary><b>Release note</b></summary>

</details>

PR #21847, opened 2025-05-13
refactor: unify `hummock+memory` with `hummock+memory-shared`

BugenZhao:

Follow up of https://github.com/risingwavelabs/risingwave/pull/21353.

This PR unifies `hummock+memory` with `hummock+memory-shared`: we will always use the singleton in-memory object store in the process. This is because I cannot find an actual use case for differentiating them after removing the distributed in-memory e2e tests.

This PR also fixes a bug that in `playground` (or `single-node` with `--in-memory`) mode, we specify the state store url as non-shared while still starting a compactor, which cannot access the objects as a result.

```
2025-05-13T17:30:23.558657+08:00 ERROR rw-standalone-compactor risingwave_object_store::object: read failed error=NotFound error: no object at path 'hummock_001/13964.data'
2025-05-13T17:30:23.558679+08:00 ERROR rw-standalone-compactor risingwave_object_store::object: read failed error=NotFound error: no object at path 'hummock_001/13917.data'
2025-05-13T17:30:23.558793+08:00  WARN           rw-compaction risingwave_storage::hummock::compactor::compactor_runner: Compaction task 6264 failed with error error=Foyer error: NotFound error: no object at path 'hummock_001/14027.data'
```

After this PR, an embedded compactor will be started by the compute node service. So there's no need to start a compactor service separately any more.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/cmd_all/src/single_node.rs
+++ b/src/cmd_all/src/single_node.rs
@@ -119,6 +119,8 @@ pub struct NodeSpecificOpts {
     pub compaction_worker_threads_number: Option<usize>,
 }
 
+const HUMMOCK_IN_MEMORY: &str = "hummock+memory";
+
 pub fn map_single_node_opts_to_standalone_opts(opts: SingleNodeOpts) -> ParsedStandaloneOpts {
     // Parse from empty strings to get the default values.
     // Note that environment variables will be used if they are set.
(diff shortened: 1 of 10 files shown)


kwannoel:
Any specific reason for this:
>  After this PR, an embedded compactor will be started by the compute node service. So there's no need to start a compactor service separately any more.

BugenZhao:
> Any specific reason for this:
> 
> > After this PR, an embedded compactor will be started by the compute node service. So there's no need to start a compactor service separately any more.

Actually this is the current behavior remaining untouched.

kwannoel (approved):

xxchan (approved):

PR #21854, opened 2025-05-14
refactor(kafka): use StreamConsumer in Enumerator to prevent poll

tabVersion:
…outdated SASL connection notes

as title

BaseConsumer will not poll kafka client, which leads to not refresh IAM auth token
StreamConsumer has a thread to poll internally, should handle this case. 

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/src/source/kafka/enumerator.rs
+++ b/src/connector/src/source/kafka/enumerator.rs
@@ -21,7 +21,7 @@ use async_trait::async_trait;
 use moka::future::Cache as MokaCache;
 use moka::ops::compute::Op;
 use rdkafka::admin::{AdminClient, AdminOptions};
-use rdkafka::consumer::{BaseConsumer, Consumer};
+use rdkafka::consumer::{Consumer, StreamConsumer};
 use rdkafka::error::KafkaResult;
 use rdkafka::{ClientConfig, Offset, TopicPartitionList};
 use risingwave_common::bail;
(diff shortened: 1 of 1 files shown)


src/connector/src/source/kafka/enumerator.rs
```diff
@@ -227,17 +227,6 @@ async fn build_kafka_client(
...
-    // rd_kafka_poll(), rd_kafka_consumer_poll(), rd_kafka_queue_poll(), etc, in order to cause retrieval
-    // of an initial token to occur.
-    // https://docs.confluent.io/platform/current/clients/librdkafka/html/rdkafka_8h.html#a988395722598f63396d7a1bedb22adaf
-    if properties.connection.is_aws_msk_iam() {
-        #[cfg(not(madsim))]
-        client.poll(Duration::from_secs(10)); // note: this is a blocking call
-        #[cfg(madsim)]
-        client.poll(Duration::from_secs(10)).await;
-    }
     Ok(Arc::new(client))
```
yuhao-su, line 230: StreamConsumer only setup the polling interval at `max.poll.interval.ms`. We might need to keep this for initial polling

tabVersion: But `StreamConsumer` does not expose the `poll` interface 🥵 https://docs.rs/rdkafka/latest/rdkafka/consumer/struct.StreamConsumer.html

yuhao-su: Yeah you did ask why using BaseConsumer in our discussion. Now I remember

tabVersion: Let's see if it works.

tabVersion: No 😒

PR #21856, opened 2025-05-14
fix: add async poll for Kafka client in enumerator

tabVersion:

as title,

The BaseConsumer has no internal thread for polling, leading to auth token expire. So we start a poll task each time calling a list_split, like every 10s. 

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/src/source/kafka/enumerator.rs
+++ b/src/connector/src/source/kafka/enumerator.rs
@@ -200,6 +200,18 @@ impl SplitEnumerator for KafkaSplitEnumerator {
             })
             .collect();
 
+        let arc_client = self.client.clone();
+        {
+            #[cfg(not(madsim))]
+            tokio::task::spawn_blocking(move || {
+                arc_client.poll(Duration::from_secs(10));
+            });
+            #[cfg(madsim)]
+            tokio::spawn(async move {
+                arc_client.poll(Duration::from_secs(10)).await;
+            });
+        }
+
         Ok(ret)
     }
 


src/connector/src/source/kafka/enumerator.rs
```diff
@@ -200,6 +200,18 @@ impl SplitEnumerator for KafkaSplitEnumerator {
             })
             .collect();
 
+        let arc_client = self.client.clone();
+        {
+            #[cfg(not(madsim))]
+            tokio::task::spawn_blocking(move || {
+                arc_client.poll(Duration::from_secs(10));
```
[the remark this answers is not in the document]
tabVersion: It is for triggering callback functions and we do not care about the data it carries. Enumerator does not actually consume data.

tabVersion:
manual cherry pick to release-2.2 https://github.com/risingwavelabs/risingwave/pull/21857

src/connector/src/source/kafka/enumerator.rs
```diff
@@ -200,6 +200,18 @@ impl SplitEnumerator for KafkaSplitEnumerator {
             })
             .collect();
 
+        let arc_client = self.client.clone();
+        {
+            #[cfg(not(madsim))]
+            tokio::task::spawn_blocking(move || {
```
hzxa21, line 206: I noticed that in other places we call the poll method only when aws msk iam is used. Should we follow the same pattern here? https://github.com/risingwavelabs/risingwave/blob/578349a5498784e02f1a8146a59613bd51def963/src/connector/src/connector_common/connection.rs#L121

kwannoel: +1 I think we should gate it, in case there's unexpected behaviour from polling. It won't affect users who don't use aws_msk_iam. Better to be defensive here. Especially when we do debugging in the future, we can be sure this code path is not executed with non aws_msk_iam users and ignore it.

PR #21865, opened 2025-05-14
feat: add ENV to control Kafka sdk log level

tabVersion:
- Introduced a new function `read_kafka_log_level` to read the log level from the environment variable `RISINGWAVE_KAFKA_LOG_LEVEL`.
- Updated `KafkaConnection`, `KafkaSinkWriter`, and `KafkaSplitReader` to utilize the dynamic log level configuration.
- Ensured that an invalid log level falls back to the default level of INFO with a warning log.

proposed by @arkbriar 

introduce `RISINGWAVE_KAFKA_LOG_LEVEL` to control the log level of kafka sdk

accept
```
DEBUG
INFO
WARN
ERROR
CRITICAL
EMERG
ALERT
NOTICE
```

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/src/connector_common/connection.rs
+++ b/src/connector/src/connector_common/connection.rs
@@ -20,6 +20,7 @@ use opendal::Operator;
 use opendal::services::{Azblob, Gcs, S3};
 use phf::{Set, phf_set};
 use rdkafka::ClientConfig;
+use rdkafka::config::RDKafkaLogLevel;
 use rdkafka::consumer::{BaseConsumer, Consumer};
 use risingwave_common::bail;
 use risingwave_common::secret::LocalSecretManager;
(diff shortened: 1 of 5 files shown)


arkbriar (approved):
LGTM! Thanks!

src/connector/src/sink/kafka.rs
```diff
@@ -477,7 +477,9 @@ impl KafkaSinkWriter {
             .await?;
             let producer_ctx = RwProducerContext::new(ctx_common);
             // Generate the producer
-            c.create_with_context(producer_ctx).await?
+            c.set_log_level(read_kafka_log_level())
```
xxchan, line 480: I think we can set sth like `RUST_ENV="librdkafka=trace"`, why need a separate env var? :eyes:

src/connector/src/connector_common/connection.rs
```diff
@@ -98,6 +98,30 @@ impl Connection for KafkaConnection {
     }
 }
 
+use rdkafka::config::RDKafkaLogLevel;
+
+pub fn read_kafka_log_level() -> RDKafkaLogLevel {
+    let log_level =
+        std::env::var("RISINGWAVE_KAFKA_LOG_LEVEL").unwrap_or_else(|_| "INFO".to_owned());
```
xxchan, line 105: If we need to add this, prefer returning an `Option` instead of INFO here. As mentioned below, it can also be configured via `RUST_LOG`

Issue #21724, opened 2025-05-06
Test sink into table recovery


kwannoel:
- [ ] Allow kill nodes when creating sink into table in recovery tests
- [ ] Allow background_ddl for sink into table in recovery tests

kwannoel:
cc @BugenZhao @shanicky

Chiro11:
More tests will be added before release-2.7

PR #17203, opened 2024-06-11
feat: support altering the target table’s columns of the sink.

shanicky:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://gist.github.com/TennyZhuang/f00be7f16996ea48effb049aa7be4d66#file-rw_cla).

## What's changed and what's your intention?

Title: Enhance Sink Behavior to Handle Table Alterations and Add Associated Tests

Summary:

This pull request implements enhancements to the system's behavior concerning sinks when table alterations occur. The changes span across several areas of the codebase, including test scripts, protobuf definitions, and Rust source code. Here's a breakdown of the key additions and modifications:

- A new end-to-end test script `alter_column.slt` has been added under `e2e_test/sink/sink_into_table`. This script validates the system's ability to manage sinks during various table alteration operations, including creating tables and sinks, inserting values, and adding or attempting to drop columns. It further checks if the sink tables reflect the correct state after these operations.

- Removal of a specific test case from `basic.slt` indicates an important change in the behavior of the system. The removed test case previously checked for an error when altering a table by adding a column with active incoming sinks. This change implies that the system may now support adding columns to a populated sink table.

- The protobuf definition for a `Sink` within `catalog.proto` has been updated to incorporate a `repeated field original_target_columns`. This new field stores the initial structure of a sink's target table, facilitating the management of table alterations in relation to sinks.

- The main body of Rust source code changes revolves around handling table alterations when sinks are involved:
  - The modules `sink/catalog/desc.rs` and `sink/catalog/mod.rs` have been updated to manage the new `original_target_columns` data.
  - A new function called `default_column_exprs` has been added to `frontend/src/catalog/table_catalog.rs`. This function generates default expressions for columns, which is essential for populating default values in rows added to altered tables.
  - The code within `frontend/src/handler/alter_table_column.rs` has been augmented to enable table alterations in the presence of sinks, highlighting improved support for adding columns to tables with active sinks.

- The pull request also notes the use of Rust `HashMap` and `Vec` collections to maintain the newly introduced data points, signaling improvements in data handling, particularly for append-only tables engaged in sink operations.

This pull request presents a comprehensive enhancement to the system, focusing on improved sink interaction with table alterations. Reviewers can delve into the individual elements of this update for a more detailed understanding, and further discussions are encouraged to fine-tune these changes.

Please review these changes and provide feedback on any aspects that may benefit from further clarification

## Checklist

- [x] I have written necessary rustdoc comments
[...]

PR #21872, opened 2025-05-14
fix: try to enable recovery test for sink into table and replace table

yezizp2012:

part of #21724 . Fixed several bugs.

## Checklist

## Documentation

<details>

</details>


src/meta/src/controller/catalog/mod.rs
```diff
@@ -429,11 +429,37 @@ impl CatalogController {
...
+            vec![]
+        };
 
         if dirty_job_objs.is_empty() {
-            if changed {
+            if !updated_table_objs.is_empty() {
                 txn.commit().await?;
+
+                // Notify the frontend to update the table info.
+                self.notify_frontend(
```
yezizp2012, line 448: When the `incoming_sink` of the table was fixed, the table's catalog was not notified to the frontend.

src/meta/src/controller/catalog/util.rs
```diff
@@ -417,8 +421,23 @@ impl CatalogController {
 
                     visit_stream_node_cont_mut(&mut pb_stream_node, |node| {
                         if let Some(NodeBody::Union(_)) = node.node_body {
-                            node.input.retain_mut(|input| {
-                                if let Some(NodeBody::Merge(merge_node)) = &mut input.node_body {
+                            node.input.retain_mut(|input| match &mut input.node_body {
```
yezizp2012, line 424: The `merge` branch of the upstream sink in the "create sink into table" process might be located within the `project`, which the current clean logic does not account for. This can result in a dirty table graph that cannot be recovered.

src/meta/src/rpc/ddl_controller.rs
```diff
@@ -750,8 +750,6 @@ impl DdlController {
             .try_into()
             .expect("Target table should exist in sink into table");
 
-        assert_eq!(table_catalog.incoming_sinks, target_table.incoming_sinks);
```
yezizp2012, line 753: The `incoming_sinks` in table catalog could be fixed during recovery when issuing a `drop` command, we should return error rather than panic here.

src/meta/src/controller/catalog/util.rs
kwannoel: Why specific to project? I'm thinking if Merge can be contained by other plan nodes as well.
Anyway this is a clear fix, just writing first, in case we need to fix more cases. cc @BugenZhao.

kwannoel: cc @chenzl25

src/meta/src/controller/catalog/util.rs
```diff
@@ -328,14 +328,14 @@ impl CatalogController {
```
kwannoel, line 328: Nit: We can add some docs about what the return values indicate.

kwannoel (approved):
Great work, I think this fixes many sink into table bugs we encountered in our releases. Thanks for your efforts!

src/meta/src/manager/streaming_job.rs
```diff
@@ -75,93 +74,11 @@ impl Default for StreamingJobType {
     }
 }
 
-impl StreamingJob {
-    pub fn mark_created(&mut self) {
```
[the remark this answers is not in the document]
yezizp2012: Unused, there were used only for v1 metastore.

src/meta/src/controller/catalog/util.rs
yezizp2012: That's by designed when introducing the `create sink into table`, Cc @shanicky for more context.

yezizp2012: It's for alter table column with sink upstream. See #17203 for more info.

Issue #7825, opened 2023-02-10
SQL Smith Coverage


jon-chuang:
### Is your feature request related to a problem? Please describe.

We currently do not understand the coverage of SQLSmith of our batch and stream system

### Describe the solution you'd like

I noticed that the plans generated often evaluate to very idiosyncratic queries. Hence, I am worried about the actual coverage. I believe the coverage for frontend/optimizer side may be ok, but am more worried about the e2e testing (batch and stream part). I am suggested several ideas to get a better understanding of the e2e coverage:

- [ ] Test how many columns/table are in the underlying tables referenced in the generated SQL v.s. columns/tables in the final optimized query plan's output. 
  - If a column does not appear in the final query or in a condition, it is very likely to be pruned away, so we do not get coverage.
  - This metric should take into account the fact that certain columns may reference columns via input ref. As long as these input refs can be traced to an underlying table/MV, that table/MV's column should be counted as being referenced.
  - If a column is derived from aggregating InputRef, then it should be considered referenced.
- [ ] fraction of expressions that reference input variables after they have been optimized in frontend.
  - Another issue is that with enough expressions, a condition always ends up being either false or true. Hence I would suggest not putting too many literals in expressions. The metric here is

xxchan:
We want to implement coverage-guided fuzzing? 😲

jon-chuang:
@xxchan not sure. The first step is to measure how good our coverage is before determining if we need to improve the generation. The automatic improvement part is optional, tuning it manually is also an option @kwannoel

kwannoel:
A simple thing to do is enable `codecov`: https://about.codecov.io/.
We currently already use it in our ci pipeline. We can upload a separate set of reports for sqlsmith codecov (deterministic test).

kwannoel:
> How do we measure the coverage in general now? We used to have similar discussions and the final conclusion was that the coverage is hard to be calculated if taking e2e tests into account. E2e is not covered by general Rust coverage tools.

AFAIK, for unit tests we use `codecov`. E2E test coverage might be possible, if we run tests with all-in-one binary / deterministic-simulation.

jon-chuang:
> E2e is not covered by general Rust coverage tools.

I think we can still get coverage via deterministic test. Secondly, we can use LLVM to instrument the deterministic test all in one binary to get a better coverage.

xxchan:
> the final conclusion was that the coverage is hard to be calculated

[...]

PR #21873, opened 2025-05-14
feat(sqlsmith): support decoupling weights from generators

### LINKED ISSUE #3936: Sqlsmith: Decouple weights from generators
- Allows us to test specific classes of expressions
- Allows us to do sanity checks when we support new expr generators. Can write unit tests to check if certain expression is generated.

cardigan1008:

This PR decouples weight configuration from the generator logic in Sqlsmith, enabling more flexible and modular control over SQL generation behavior.

It introduces a YAML-based configuration mode and supports specifying a custom config file via the CLI. Each feature in the config file can be assigned two fields: weight and enabled.

For example, when adding a new feature such as `eowc`, a developer can set its weight to 100 to focus fuzzing on it. The generation logic can then be gated with a call like `self.should_generate("eowc")`, which checks the configuration at runtime. To test the feature, the developer simply adds the following entry to the config file:

```yaml
eowc:
  weight: 100
  enabled: true
```

We have added basic configuration support for controlling the generation of select operators in batch tests, including WHERE, JOIN, and aggregation clauses.

This change simplifies targeted testing for new features and lays the foundation for future coverage-driven fuzzing strategies.

Limitation:

This PR is primarily intended for ad-hoc testing, allowing developers to manually enable and prioritize new features via the configuration file. However, it does not yet support automated weight adjustment based on coverage feedback, as discussed in [#7825](https://github.com/risingwavelabs/risingwave/issues/7825).

Future Work:

- Extend configuration coverage to include more SQL features as part of ongoing coverage evaluation efforts.
- As suggested by @kwannoel, consider integrating this feature into CI workflows, using PR labels to automatically trigger targeted tests.

Fixes #3936 

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/tests/sqlsmith/config.yml
+++ b/src/tests/sqlsmith/config.yml
@@ -1,0 +1,9 @@
+where:
+  weight: 50
+  enabled: true
+agg:
+  weight: 50
+  enabled: true
+join:
+  weight: 75
+  enabled: true
--- a/src/tests/sqlsmith/config.yml
+++ b/src/tests/sqlsmith/config.yml
@@ -1,9 +1,10 @@
-where:
-  weight: 50
-  enabled: true
-agg:
-  weight: 50
-  enabled: true
-join:
-  weight: 75
-  enabled: true
+config:
+  where:
+    weight: 50
+    enabled: true
+  agg:
+    weight: 50
+    enabled: true
+  join:
+    weight: 75
+    enabled: false
--- a/src/tests/sqlsmith/develop.md
+++ b/src/tests/sqlsmith/develop.md
@@ -8,7 +8,7 @@ See [Running with Madsim](./README.md#running-with-madsim).
 ### Focused property testing
 
 When introducing a new feature, you can customize [config.yml](./config.yml) to
-set its generation weight and toggle its availability for targeted testing.
+set its weight to specify the generation probability (0-100%) and toggle its availability for targeted testing.
 
 To see how to control feature generation programmatically, refer to `config.rs` and
 the usage of the should_generate() function.
--- a/src/tests/sqlsmith/develop.md
+++ b/src/tests/sqlsmith/develop.md
@@ -5,6 +5,14 @@
 Suggest testing it with madsim, so you can reproduce the same results each time.
 See [Running with Madsim](./README.md#running-with-madsim).
 
+### Focused property testing
+
+When introducing a new feature, you can customize [config.yml](./config.yml) to
+set its generation weight and toggle its availability for targeted testing.
+
+To see how to control feature generation programmatically, refer to `config.rs` and
+the usage of the should_generate() function.
+
 ## Ignoring errors
 
 Sometimes you may encounter unrelated errors, since SqlSmith frontend tests run per PR.
--- a/src/tests/sqlsmith/src/config.rs
+++ b/src/tests/sqlsmith/src/config.rs
@@ -13,12 +13,33 @@
 // limitations under the License.
 
 use std::collections::BTreeMap;
+use std::fmt;
 
 use rand::Rng;
 use serde::Deserialize;
+
+#[derive(Debug, Clone, Copy, Ord, PartialOrd, PartialEq, Eq, Hash, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum Feature {
+    Where,
+    Agg,
+    Join,
+}
+
+impl fmt::Display for Feature {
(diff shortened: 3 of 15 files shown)


src/tests/sqlsmith/develop.md
```diff
@@ -5,6 +5,14 @@
 Suggest testing it with madsim, so you can reproduce the same results each time.
 See [Running with Madsim](./README.md#running-with-madsim).
 
+### Focused property testing
+
+When introducing a new feature, you can customize [config.yml](./config.yml) to
+set its generation weight and toggle its availability for targeted testing.
```
kwannoel, line 11: We'd better be more precise and remark that it's a probability (0-100%) rather than an arbitrary weight.

src/tests/sqlsmith/src/config.rs
```diff
@@ -0,0 +1,77 @@
...
+impl Default for Configuration {
+    fn default() -> Self {
+        Self::new("config.yml")
+    }
+}
+
+impl Configuration {
+    pub fn new(path: &str) -> Configuration {
+        let data = std::fs::read_to_string(path).unwrap();
+        let config: Configuration = serde_yaml::from_str(&data).unwrap();
```
kwannoel, line 44: We can add some additional validation step, to make sure valid weights [0, 100] are provided.

kwannoel (approved):
Rest LGTM, thanks!

src/tests/sqlsmith/src/config.rs
cardigan1008: Fixed. Thanks!

src/tests/sqlsmith/config.yml
```diff
@@ -0,0 +1,10 @@
+config:
+  where:
+    weight: 50
+    enabled: true
+  agg:
+    weight: 50
+    enabled: true
+  join:
+    weight: 75
+    enabled: false
```
stdrc, line 10: Actually I'm a bit concerned about how we determine the granularity of such features to be tested. For example `eowc` is not a simply a SQL keyword, you have can only add the keyword to a certain type of queries, and should be able to allow planner error when the query is not `eowc`-able. For another example, `topn` feature is not only about `rank` function call and `where`, but about a certain combination pattern of the two, together with subquery. So I'm curious how we gonna adjust the weight of such features with `config.yml`, for instance how do we "disable" the generation of TopN queries.

stdrc: I'm asking this because we basically don't have a good way to achieve this kind of control (disable the generation of TopN, agg calls that only support append-only stream, etc). Since we are introducing a new mechanism that is intended to allow weight controlling, I think this is worth considering.

cardigan1008: Thanks for raising this! I actually discussed this with Noel earlier about the design. We also had concerns about the granularity issue, especially when it comes to enabling dynamic control in fuzzing.

Noel suggested that we could start by supporting a subset of cases, using it primarily as a tool for targeted testing of newly added features. This way, we can validate the mechanism early on and gradually refine it as more use cases emerge. 

cc @kwannoel

src/tests/sqlsmith/src/sql_gen/query.rs
```diff
@@ -195,7 +195,7 @@ impl<R: Rng> SqlGenerator<'_, R> {
     }
 
     fn gen_select_list(&mut self, num_select_items: usize) -> (Vec<SelectItem>, Vec<Column>) {
-        let can_agg = self.flip_coin();
+        let can_agg = self.should_generate("agg");
```
stdrc, line 198: Shall we add some `const FEAT_KEY_AGG: String = ...` things or an auto-generated `enum`?

kwannoel: Good suggestion. Suggest enum for now, until we have a more concrete form for the configurations. Otherwise, we may need to refactor the macro several times.

cardigan1008:
There’s a [SQLGeneratorContext](https://github.com/risingwavelabs/risingwave/blob/main/src/tests/sqlsmith/src/sql_gen/mod.rs#L96) that only serves to disable agg at top level. I think it’s duplicate to this decouple design, as they both target at controlling some features at top level. After discussing with @kwannoel , we agreed it would be cleaner to merge the two mechanisms. This will require a lot of refactoring, so I’ll address it in a follow-up PR.

stdrc (approved):
LGTM

PR #21886, opened 2025-05-15
fix: make batch scan adhere to runtime param batch_parallelism

zwang28:

Previously, the runtime parameter `batch_parallelism` [doesn't apply to the batch scan](https://github.com/risingwavelabs/risingwave/blob/b34ead99cea83cf861952bba326ca2c1bff2efe0/src/frontend/src/scheduler/plan_fragmenter.rs#L1008). Therefore the batch scan will use all available serving parallleism up to 256. The might cause problem when high QPS introduces too many batch tasks in serving nodes.

This PR makes batch scan adhere to `batch_parallelism`, in order to limit batch scan task parallelism to lower number when necessary. The implementation idea is to utilize only partial parallelism of all when `place_vnode` in meta node.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/meta/src/serving/mod.rs
+++ b/src/meta/src/serving/mod.rs
@@ -26,6 +26,7 @@ use tokio::sync::oneshot::Sender;
 use tokio::task::JoinHandle;
 
 use crate::controller::fragment::FragmentParallelismInfo;
+use crate::controller::session_params::SessionParamsControllerRef;
 use crate::manager::{LocalNotification, MetadataManager, NotificationManagerRef};
 use crate::model::FragmentId;
 
@@ -47,6 +48,7 @@ impl ServingVnodeMapping {
         &self,
         streaming_parallelisms: HashMap<FragmentId, FragmentParallelismInfo>,
         workers: &[WorkerNode],
+        max_serving_parallelism: Option<u64>,
     ) -> (HashMap<FragmentId, WorkerSlotMapping>, Vec<FragmentId>) {
         let mut serving_vnode_mappings = self.serving_vnode_mappings.write();
         let mut upserted: HashMap<FragmentId, WorkerSlotMapping> = HashMap::default();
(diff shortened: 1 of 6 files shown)


src/meta/src/serving/mod.rs
```diff
@@ -58,7 +60,8 @@ impl ServingVnodeMapping {
                     FragmentDistributionType::Unspecified => unreachable!(),
                     FragmentDistributionType::Single => Some(1),
                     FragmentDistributionType::Hash => None,
-                };
+                }
+                .or_else(|| max_serving_parallelism.map(|p| p as usize));
```
chenzl25, line 64: IIUC, the current strategy is using 1 parallelism for single distribution and max_serving_parallelism for hash distribution.

src/meta/src/serving/mod.rs
```diff
@@ -158,13 +171,43 @@ pub async fn start_serving_vnode_mapping_worker(
...
+            tracing::debug!(
+                "Update serving vnode mapping snapshot for fragments {:?}.",
+                mappings.keys()
+            );
+            if !failed.is_empty() {
+                tracing::warn!(
+                    "Fail to update serving vnode mapping for fragments {:?}.",
+                    failed
+                );
+            }
```
chenzl25, line 203: Under what circumstances would the failed branch be hit?

chenzl25 (approved):
Rest LGTM

src/meta/src/serving/mod.rs
zwang28: Only on some bug.
The branch is added because it calls the pure function `place_vnode` can fail, e.g. due to incorrect arguments.

src/meta/src/serving/mod.rs
zwang28: Yes.

PR #21891, opened 2025-05-15
fix: poll the kafka consumer in KafkaSplitReader before reading meta data

yuhao-su:

Use a new consumer to kafka fetch meta data in splitreader.
To set up the oauth token, we need to poll the kafka consumer in KafkaSplitReader before reading meta data.

This will incur one more consumer built for each KafkaSplitReader.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/src/source/kafka/enumerator.rs
+++ b/src/connector/src/source/kafka/enumerator.rs
@@ -212,7 +212,7 @@ impl SplitEnumerator for KafkaSplitEnumerator {
     }
 }
 
-async fn build_kafka_client(
+pub async fn build_kafka_client(
     config: &ClientConfig,
     properties: &KafkaProperties,
 ) -> ConnectorResult<Arc<KafkaConsumer>> {
--- a/src/connector/src/source/kafka/source/reader.rs
+++ b/src/connector/src/source/kafka/source/reader.rs
@@ -33,6 +33,7 @@ use crate::parser::ParserConfig;
 use crate::source::base::SourceMessage;
 use crate::source::kafka::{
     KAFKA_ISOLATION_LEVEL, KafkaContextCommon, KafkaProperties, KafkaSplit, RwConsumerContext,
+    build_kafka_client,
 };
 use crate::source::{
     BackfillInfo, BoxSourceChunkStream, Column, SourceContextRef, SplitId, SplitImpl,
@@ -88,7 +89,7 @@ impl SplitReader for KafkaSplitReader {
             // thread consumer will keep polling in the background, we don't need to call `poll`
             // explicitly
             Some(source_ctx.metrics.rdkafka_native_metric.clone()),
-            properties.aws_auth_props,
+            properties.aws_auth_props.clone(),
             properties.connection.is_aws_msk_iam(),
         )
         .await?;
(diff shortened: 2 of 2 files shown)


kwannoel (dismissed):
LGTM. @xxchan @tabVersion can you guys leave a review too?

tabVersion:
Seems the pr do the things:

* Meta: basically the same as `SHARED_KAFKA_CONSUMER` by adding a tokio thread to keep polling. Why not poll the client in the `list_splits` call (https://github.com/risingwavelabs/risingwave/pull/21856), basically we just need a timer for the polling 
* Compute: Introduce the shared client to do the metadata call, but this approach cannot refresh the auth token inside Stream Consumer, which actually reads data. 
  * In the sdk impl, the kafka client only call the callback func when consume data [ref](https://docs.confluent.io/platform/current/clients/librdkafka/html/classRdKafka_1_1OAuthBearerTokenRefreshCb.html#:~:text=The%20SASL%2FOAUTHBEARER%20token%20refresh%20callback,sasl.oauthbearer.config). 
  * Consider the case, the topic has no new messages for a while and token expires, the client is required to re-auth from the broker or it can meet the `SASL authentication failed` and the consume process fails, leading to be unable to generate new token, which is a dead loop.

I think for meta part, the pr introduce too much complexity
and for compute node, it only handles a part of metadata request logic, and still facing the scene of the auth dead loop above.

tabVersion:
also some take away from chatgpt

---

🔑 Key Points:

1.  Token Refresh Is Triggered by poll() by Default
In librdkafka, the OAuth token refresh callback is only triggered when you call poll() (or consume()).
If your app stops polling and the token expires, it won’t be automatically refreshed.

2. No Polling = No Token Refresh
Even if new messages arrive, they won't be consumed if:

* The token has expired.
* You haven't called poll() to refresh it.

3. Broker May Not Immediately Disconnect
If connections.max.reauth.ms is not set on the broker, it may not force reauthentication.
But fetch requests with an expired token can fail with an authentication error, and your client may enter a reconnect loop without refreshing the token.

4. New Messages Don’t Trigger Refresh Automatically
Kafka is pull-based. If your client isn’t polling, new messages won’t trigger any internal activity, including token renewal.

---

✅ Solutions:
1. Call poll() Regularly
Even if there are no new messages, periodically call:
```
consumer.poll(Duration::from_secs(0));
```
This keeps the internal state alive and triggers token refresh when needed.

2. Enable Background Callbacks in librdkafka (Advanced)
Use:

```
rd_kafka_conf_enable_sasl_queue(conf, true);
rd_kafka_sasl_background_callbacks_enable(rk);
```
This lets librdkafka refresh tokens in a background thread — no polling needed.
⚠️ Rust-rdkafka does not yet expose this easily — may require FFI or upstream support.

3. Use a Timer to Force Polling
Use a background task or timer (e.g. tokio::interval) to periodically poll, even if idle.

---

✅ Best Practices (Summary):

* Keep polling, even when idle.
* Enable background token refresh if possible.
* Make sure your token callback is fast, reliable, and idempotent.

On the broker, consider setting connections.max.reauth.ms to avoid stale long-lived connections.

tabVersion:
close as prefer fix inside SDK https://github.com/madsim-rs/madsim/pull/243

PR #21892, opened 2025-05-15
fix: use new tokio thread for iam request to avoid stuck meta

tabVersion:

as title

The origin impl will stuck the whole meta service if load credentials from EC2 env. 
If providing aksk in sql, it runs smoothly. I dont know why. 

* why stuck?

in prev impl, it use the same runtime as the main thread, so when it calls block_on in a new physical thread, seems there is a race condition and make things worse. 

---

manual tested on ec2

## Checklist

## Documentation

<details>

</details>


yuhao-su:
If using iam role does the kafka read work now?

tabVersion:
> If using iam role does the kafka read work now?

Tested on EC2 and yes.

yuhao-su (approved):

hzxa21:
I suspect there are other places in our code base that requires calling an async function inside a non-async function so I had an offline discussion with @chenzl25 and @wenym1 on that. It turns out we do have such use cases in FE and met similar stuckness issue before. A good practice of solving this issue is not creating a separate current thread runtime, but use `block_in_place` together with `block_on` in the existing runtime.

Example: https://github.com/risingwavelabs/risingwave/blob/578349a5498784e02f1a8146a59613bd51def963/src/frontend/src/expr/table_function.rs#L479

kwannoel:
> I suspect there are other places in our code base that requires calling an async function inside a non-async function so I had an offline discussion with @chenzl25 and @wenym1 on that. It turns out we do have such use cases in FE and met similar stuckness issue before. A good practice of solving this issue is not creating a separate current thread runtime, but use `block_in_place` together with `block_on` in the existing runtime.
> 
> Tokio doc: https://docs.rs/tokio/latest/tokio/task/fn.block_in_place.html
> 
> Example:
> 
> https://github.com/risingwavelabs/risingwave/blob/578349a5498784e02f1a8146a59613bd51def963/src/frontend/src/expr/table_function.rs#L479

In Frontend that might work, however we frequently 

> I suspect there are other places in our code base that requires calling an async function inside a non-async function so I had an offline discussion with @chenzl25 and @wenym1 on that. It turns out we do have such use cases in FE and met similar stuckness issue before. A good practice of solving this issue is not creating a separate current thread runtime, but use `block_in_place` together with `block_on` in the existing runtime.
> 
> Tokio doc: https://docs.rs/tokio/latest/tokio/task/fn.block_in_place.html
> 
> Example:
> 
> https://github.com/risingwavelabs/risingwave/blob/578349a5498784e02f1a8146a59613bd51def963/src/frontend/src/expr/table_function.rs#L479

I'm not sure if it's suitable for this case. In FE, it occurs in a one-off way. So it's acceptable to block the task. However, in source executor, if we block the task, it may lead to starvation of other code running in that task.

hzxa21:
> I'm not sure if it's suitable for this case. In FE, it occurs in a one-off way, e.g. during binding. So it's acceptable to block the task. However, in source executor, if we block the task, it may lead to starvation of other code running in that task.

Correct me if I am wrong, the reason why we encounter the stuckness issue prior to this PR is because
1. Each `block_on` call will occupy the tokio worker thread for async task
2. Inside the `block_on` call, we call an async fn `generate_auth_token_from_credentials_provider`, which may spawn a tokio task inside or require the tokio async worker thread to be available in order to drive the future to completion
3. If all tokio worker thread are occupied because of 1, then 2 will never succeed, which creates a circular dependency.

IIUC, there are two thread pool inside each tokio runtime, one for async task, another for bloking task. `block_in_place` is introduced to solve the above issue because the blocking call passed to `block_in_place` will occupy the thread pool for blocking task, which is independent of the thread pool for async task. In this case, the blocking function (`block_on`) won't block the async worker threads and the circular dependency is broken.

src/connector/src/source/kafka/client_context.rs
```diff
@@ -142,8 +144,9 @@ impl KafkaContextCommon {
                         )
                         .await
                     })
-                });
-                handle.join().unwrap()??
+                })
+                .unwrap()
+                .unwrap()
```
kwannoel, line 149: There's still `unwrap`

tabVersion: Oh yet needs another fix...

kwannoel: Any update

tabVersion: https://github.com/risingwavelabs/risingwave/pull/22113

PR #21900, opened 2025-05-16
fix: poll stream consumer for triggering callback func

tabVersion:

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/src/source/kafka/source/reader.rs
+++ b/src/connector/src/source/kafka/source/reader.rs
@@ -150,6 +150,9 @@ impl SplitReader for KafkaSplitReader {
 
         consumer.assign(&tpl)?;
 
+        // poll consumer to trigger callback functions
+        let _ = tokio::time::timeout(Duration::from_millis(1000), consumer.recv()).await;
+
         // The two parameters below are only used by developers for performance testing purposes,
         // so we panic here on purpose if the input is not correctly recognized.
         let bytes_per_second = match properties.bytes_per_second {
--- a/src/connector/src/source/kafka/source/reader.rs
+++ b/src/connector/src/source/kafka/source/reader.rs
@@ -151,6 +151,7 @@ impl SplitReader for KafkaSplitReader {
         consumer.assign(&tpl)?;
 
         // poll consumer to trigger callback functions
+        // We don't need to wait for the result, because recv may block and wait for next new message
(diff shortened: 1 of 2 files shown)


src/connector/src/source/kafka/source/reader.rs
```diff
@@ -150,6 +150,10 @@ impl SplitReader for KafkaSplitReader {
 
         consumer.assign(&tpl)?;
 
+        // poll consumer to trigger callback functions
+        // We don't need to wait for the result, because recv may block and wait for next new message
+        let _ = tokio::time::timeout(Duration::from_millis(1000), consumer.recv()).await;
+
```
yuhao-su, line 156: Will this read a new message and ignore it?

tabVersion: after adjusting the code position, the `recv` call will not dismiss the first message.

tabVersion: https://github.com/risingwavelabs/risingwave/pull/21868/commits/4d50e8d055ee0b916d8c9c7d76e6e1bafc5db315 from https://github.com/risingwavelabs/risingwave/pull/21868 also do the same change. already contained in image [v2.2.6-patch-kafka-poll-log-level-3](https://github.com/orgs/risingwavelabs/packages/container/risingwave/417466703?tag=v2.2.6-patch-kafka-poll-log-level-3)

src/connector/src/source/kafka/source/reader.rs
```diff
@@ -100,6 +100,10 @@ impl SplitReader for KafkaSplitReader {
             .await
             .context("failed to create kafka consumer")?;
 
+        // poll consumer to trigger callback functions
+        // We don't need to wait for the result, because recv may block and wait for next new message
+        let _ = tokio::time::timeout(Duration::from_millis(100), consumer.recv()).await;
```
hzxa21, line 105: Is it because we call `recv` before `assign` (in L155) that `recv` becomes a function purely for side-effect (trigger the callback) without fetching any message?

hzxa21: Also, IIUC `consumer.recv` will be blocked indefinitely here. how do we make sure that the callback function must be triggered before the 100ms timeout?

PR #21811, opened 2025-05-10
feat(iceberg): skip write to hummock for APPEND ONLY iceberg table

xxchan:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## What's changed and what's your intention?
I initially tried to eliminate the `Materialize` plan node completely (#21810), but it needs to change many places (I haven't finished debugging it).

I think the main problem of that approach is that we don't have a dedicated stuff corresponding to a "user created table", or a standalone `TableCatalog`. `TableCatalog` is always coupled with a state table, and user facing table is always coupled with a Materialize. 
Besides, we have assumptions like the leaf node of a job should always be Materialize.

On the other hand, the solution of hacking MaterializeExec to skip write to the table looks good enough. The cost of having an empty hummock table should be ignorable. 

## Checklist

- [ ] I have written necessary rustdoc comments.
- [ ]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  I have checked the [Release Timeline](https://github.com/risingwavelabs/rw-commits-history/blob/main/release_timeline.md) and [Currently Supported Versions](https://docs.risingwave.com/changelog/release-support-policy#support-end-dates-for-recent-releases) to determine which release branches I need to cherry-pick this PR into. 

## Documentation

- [ ]  My PR needs documentation updates. 

<details>
<summary><b>Release note</b></summary>

</details>

PR #21901, opened 2025-05-18
fix(iceberg): fix CDC table with iceberg engine

chenzl25:

- https://github.com/risingwavelabs/risingwave/pull/21811 stopped passing the job type to iceberg, but it should be, otherwise, we can't create an iceberg table with postgre cdc.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/e2e_test/iceberg/test_case/cdc/load.slt
+++ b/e2e_test/iceberg/test_case/cdc/load.slt
@@ -50,6 +50,27 @@ set sink_decouple = true
 
 # iceberg table engine with CDC table
 statement ok
+create secret my_secret with (
+  backend = 'meta'
+) as 'hummockadmin';
+
+
+statement ok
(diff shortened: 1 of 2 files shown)


xxchan (approved):

chenzl25:
cc @xxchan an unexpected failed test for mysql-cdc. https://buildkite.com/risingwavelabs/pull-request/builds/74769#0196e708-1290-4211-9349-37bf486f9d9c

chenzl25:
Another corner case  https://buildkite.com/risingwavelabs/pull-request/builds/74783#0196e7a7-8d2a-4762-9e62-b2438395a631 , when dropping the shared source cascade, if the iceberg engine referring to this source, dropping cascade will drop only the hummock table and iceberg sink, but leaving a dangling iceberg source.

PR #21916, opened 2025-05-19
feat(storage): filter out GC objects by listing time travel versions and deltas

wenym1:

When doing GC, we will filter out the objects pinned by time travel. Previously we do the filtering by reading the `hummock_sstable_info` meta store table. This works for the current code that only has sst object to track. However, we will have vector index file in the future, and therefore reading `hummock_sstable_info` is not enough.

In this PR, we will change to list the whole `hummock_time_travel_version`  and `hummock_time_travel_version_delta` table to figure out the object id pinned by time travel version, so that in the future we can support GC of objects other than sst.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/common/src/config.rs
+++ b/src/common/src/config.rs
@@ -562,6 +562,19 @@ pub struct MetaDeveloperConfig {
     #[serde(default = "default::developer::hummock_time_travel_filter_out_objects_batch_size")]
     pub hummock_time_travel_filter_out_objects_batch_size: usize,
 
+    #[serde(default = "default::developer::hummock_time_travel_filter_out_objects_v1")]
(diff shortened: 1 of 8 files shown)


wenym1:
* **#21916** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/21916?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/21916?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

zwang28 (approved):

PR #21929, opened 2025-05-20
feat(ai): support AI model functions with a special UDF runtime

stdrc:

Now some databases/streaming processing platforms have builtin AI model support like text embedding. This PR:

1. Implement a new special UDF runtime as the basis of such AI models.
2. Implement the `OpenaiEmbedding` function.

Example:

```sql
CREATE SECRET OPENAI_API_KEY WITH (backend = 'meta') AS 'the-key';

CREATE FUNCTION my_openai_embedding(text) RETURNS float4[]
RUNTIME _ai_model
AS openai_embedding
WITH (
    api_key = SECRET OPENAI_API_KEY,
    model = 'text-embedding-3-small'
);

SELECT my_openai_embedding('Hello, world!');

CREATE TABLE inputs (id INT, text TEXT);
CREATE MATERIALIZED VIEW embeddings AS
SELECT id, text, my_openai_embedding(text) as embedding FROM inputs;
```

Note that this is now a internal preview feature and is not intended to be announced publicly.

Later we may introduce new syntax for creating such functions.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -882,6 +882,43 @@ dependencies = [
 ]
 
 [[package]]
+name = "async-openai"
+version = "0.28.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "14d76e2f5af19477d6254415acc95ba97c6cc6f3b1e3cb4676b7f0fab8194298"
+dependencies = [
+ "async-openai-macros",
+ "backoff",
+ "base64 0.22.1",
+ "bytes",
+ "derive_builder",
+ "eventsource-stream",
+ "futures",
+ "rand 0.8.5",
+ "reqwest 0.12.15",
+ "reqwest-eventsource",
+ "secrecy",
+ "serde",
+ "serde_json",
+ "thiserror 2.0.12",
+ "tokio",
+ "tokio-stream 0.1.17",
+ "tokio-util",
+ "tracing",
+]
+
+[[package]]
+name = "async-openai-macros"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0289cba6d5143bfe8251d57b4a8cac036adf158525a76533a7082ba65ec76398"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 2.0.101",
+]
+
+[[package]]
 name = "async-recursion"
 version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1644,6 +1681,20 @@ dependencies = [
  "tower-layer",
  "tower-service",
  "tracing",
+]
+
+[[package]]
+name = "backoff"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b62ddb9cb1ec0a098ad4bbf9344d0713fa193ae1a80af55febcff2627b6a00c1"
+dependencies = [
+ "futures-core",
+ "getrandom 0.2.11",
+ "instant",
+ "pin-project-lite",
+ "rand 0.8.5",
+ "tokio",
 ]
 
 [[package]]
@@ -2585,7 +2636,7 @@ dependencies = [
  "serde_json",
  "thread_local",
  "tokio",
- "tokio-stream 0.1.16",
+ "tokio-stream 0.1.17",
  "tonic",
  "tracing",
  "tracing-core",
@@ -2599,7 +2650,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e4a16ff120e0b2d07fdcfda1cc2c1c72b76d3b2fe7cc5ec82bf7b42769b2e73c"
 dependencies = [
  "auto_enums",
- "darling 0.20.9",
+ "darling 0.20.11",
  "itertools 0.12.1",
  "proc-macro2",
  "quote",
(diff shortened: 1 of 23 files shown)


stdrc:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/21929?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#21929** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/21929?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/21929?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#21928** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/21928?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

hzxa21 (approved):
LGTM

BugenZhao:
From my understanding, basically we are doing some kind of function currying?

What if we provide some builtin functions like
```sql
openai_embedding(api_key, model, input..)
```

Then simply create `SQL` UDF as
```sql
CREATE FUNCTION my_openai_embedding(text) RETURNS float4[]
LANGUAGE sql AS
$$ select openai_embedding(SECRET OPENAI_API_KEY, 'text-embedding-3-small', text) $$
```

So that we don't need to introduce any form of new concepts or syntax?

hzxa21:
> From my understanding, basically we are doing some kind of function currying?
> 
> What if we provide some builtin functions like
> 
> ```sql
> openai_embedding(api_key, model, input..)
> ```
> 
> Then simply create `SQL` UDF as
> 
> ```sql
> CREATE FUNCTION my_openai_embedding(text) RETURNS float4[]
> LANGUAGE sql AS
> $$ select openai_embedding(SECRET OPENAI_API_KEY, 'text-embedding-3-small', text) $$
> ```
> 
> So that we don't need to introduce any form of new concepts or syntax?

That's a good point. +1 on this.

wenym1:
> From my understanding, basically we are doing some kind of function currying?
> 
> What if we provide some builtin functions like
> 
> ```sql
> openai_embedding(api_key, model, input..)
> ```
> 
> Then simply create `SQL` UDF as
> 
> ```sql
> CREATE FUNCTION my_openai_embedding(text) RETURNS float4[]
> LANGUAGE sql AS
> $$ select openai_embedding(SECRET OPENAI_API_KEY, 'text-embedding-3-small', text) $$
> ```
> 
> So that we don't need to introduce any form of new concepts or syntax?

I think this new syntax is more straightforward and user friendly

Besides, for functionality and correctness, function currying is doable. In actual implementation, if we process the input texts one by one, the two are the same. However, when we want to handle texts in batch, they are different, in whether or not we can assume that the hyper parameter(secret key, model type) are a fixed one.

In what we implement in this PR, we will handle the batch in the following way.
```
openai_embedding(api_key, model, 
    [
        'text1',
        'text2',
        'text3',
        'text4',
        ...
    ]
)
```

If we implement it as function currying, we will have to handle the batch in the following way
```
openai_embedding( [
        (api_key1, model1, 'text1'),
        (api_key2, model2, 'text2'),
        (api_key3, model3, 'text3'),
        (api_key4, model4, 'text4'),
        ...
    ]
)
```

In the rust sdk of openai, in the [CreateEmbeddingRequest](https://docs.rs/async-openai/0.28.2/async_openai/types/struct.CreateEmbeddingRequest.html), it supports passing a batch of texts and processing them all at once. If we implement it as function currying, we will not be able to leverage this, unless we can resolve it the optimizer.

```
pub enum EmbeddingInput {
    String(String),
    StringArray(Vec<String>),
    IntegerArray(Vec<u32>),
    ArrayOfIntegerArray(Vec<Vec<u32>>),
}
```

hzxa21:
I think both interfaces can handle batches bound to fixed model and api key like:
```
openai_embedding(api_key, model, 
    [
        'text1',
        'text2',
        'text3',
        'text4',
        ...
    ]
)
```

optimizer should have sufficient knowledge on whether the input `model` and `api_key` parameter is an immutable const. I don't have a strong preference on one option over another but in general not introducing new syntax seems to make things simpler.

BugenZhao:
> optimizer should have sufficient knowledge on whether the input `model` and `api_key` parameter is an immutable const

This is true. Providing a function with signature `openai_embedding(api_key, model, input..)` does not mean that we have to create a `client` and issue an HTTP request for each row.

Expressions related to regex is a very similar example to take reference to: we don't necessarily build the regex each row.

https://github.com/risingwavelabs/risingwave/blob/b7893145dc7e5626584cc3e1e420f598b6914ac2/src/expr/macro/src/gen.rs#L271-L306

https://github.com/risingwavelabs/risingwave/blob/b7893145dc7e5626584cc3e1e420f598b6914ac2/src/expr/impl/src/scalar/regexp.rs#L169-L176

stdrc:
> From my understanding, basically we are doing some kind of function currying?
> 
> What if we provide some builtin functions like
> 
> ```sql
> openai_embedding(api_key, model, input..)
> ```
> 
> Then simply create `SQL` UDF as
> 
> ```sql
> CREATE FUNCTION my_openai_embedding(text) RETURNS float4[]
> LANGUAGE sql AS
> $$ select openai_embedding(SECRET OPENAI_API_KEY, 'text-embedding-3-small', text) $$
> ```
> 
> So that we don't need to introduce any form of new concepts or syntax?

I actually thought about this approach, which is relatively an apparent one. But was stopped by what @wenym1 said. Good to know that `regexp_count` already achieved this.

stdrc:
Superseded by #22242

PR #17933, opened 2024-08-06
feat(source): use the default_provider for AWS-related connectors

fuyufjh:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://gist.github.com/TennyZhuang/f00be7f16996ea48effb049aa7be4d66#file-rw_cla).

## What's changed and what's your intention?

See background at #17851

## Checklist

- [ ] I have written necessary rustdoc comments
- [ ] I have added necessary unit tests and integration tests
- [ ] I have added test labels as necessary. See [details](https://github.com/risingwavelabs/risingwave/blob/main/docs/developer-guide.md#ci-labels-guide).
- [ ] I have added fuzzing tests or opened an issue to track them. (Optional, recommended for new SQL features #7934).
- [ ] My PR contains breaking changes. (If it deprecates some features, please create a tracking issue to remove them in the future).
- [ ] All checks passed in `./risedev check` (or alias, `./risedev c`)
- [ ] My PR changes performance-critical code. (Please run macro/micro-benchmarks and show the results.)

- [ ] My PR contains critical fixes that are necessary to be merged into the latest release. (Please check out the [details](https://github.com/risingwavelabs/risingwave/blob/main/CONTRIBUTING.md))

## Documentation

- [ ] My PR needs documentation updates. (Please use the **Release note** section below to summarize the impact on users)

## Release note

If this PR includes changes that directly affect users or other significant modifications relevant to the community, kindly draft a release note to provide a concise summary of these changes. Please prioritize highlighting the impact these changes will have on users.

</details>

PR #21938, opened 2025-05-21
feat(iceberg): refine enable config load for iceberg

chenzl25:

- Respect the env variable `DISABLE_DEFAULT_CREDENTIAL` introduced by https://github.com/risingwavelabs/risingwave/pull/17933
- In a cloud environment, DISABLE_DEFAULT_CREDENTIAL should be set to true. However, for a self-hosted environment, we allow users to enable config load to load warehouse credentials from the environment.
- Support `enable_config_load` for iceberg sink, source, iceberg engine and connection.

## Checklist

## Documentation

<details>

- Support `enable_config_load` for iceberg sink, source, iceberg engine and connection.
- This parameter set to true will load warehouse credentials from the environment. Only allowed to be used in a self-hosted environment.

</details>

### DIFF
--- a/src/connector/src/connector_common/connection.rs
+++ b/src/connector/src/connector_common/connection.rs
@@ -23,6 +23,7 @@ use rdkafka::ClientConfig;
 use rdkafka::consumer::{BaseConsumer, Consumer};
 use risingwave_common::bail;
 use risingwave_common::secret::LocalSecretManager;
+use risingwave_common::util::env_var::env_var_is_true;
 use risingwave_pb::catalog::PbConnection;
 use serde_derive::Deserialize;
 use serde_with::serde_as;
@@ -30,6 +31,7 @@ use tonic::async_trait;
 use url::Url;
 use with_options::WithOptions;
 
+use crate::connector_common::common::DISABLE_DEFAULT_CREDENTIAL;
 use crate::connector_common::{
     AwsAuthProps, IcebergCommon, KafkaConnectionProps, KafkaPrivateLinkCommon,
 };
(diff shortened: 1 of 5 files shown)


yuhao-su:
I'm thinking if `enable_config_load` is necessary. Is there any case users need to decide if they want to use the config load or not for iceberg? For any other connectors, we don't have such setting.

chenzl25:
> I'm thinking if `enable_config_load` is necessary. Is there any case users need to decide if they want to use the config load or not for iceberg? For any other connectors, we don't have such setting.

Some of our self-hosted users have already used `enable_config_load` to enable config load for iceberg, so at least in terms of compatibility, we should keep it.

yuhao-su (approved):

src/connector/src/connector_common/connection.rs
```diff
@@ -212,6 +214,10 @@ pub struct IcebergConnection {
     #[serde(rename = "catalog.jdbc.password")]
     pub jdbc_password: Option<String>,
 
+    /// Enable config load. This parameter set to true will load warehouse credentials from the environment. Only allowed to be used in a self-hosted environment.
+    #[serde(default, deserialize_with = "deserialize_optional_bool_from_string")]
+    pub enable_config_load: Option<bool>,
```
xxchan, line 219: Do we still need this variable? Can we just us `DISABLE_DEFAULT_CREDENTIAL`？ 🤔

xxchan: Oh, discussed above

xxchan (approved):

Li0k (approved):

Issue #20428, opened 2025-02-10
Benchmark data shuffling vs compaction


kwannoel:
Network bandwidth is a finite resource. Before shuffling a stream chunk, we can decide whether or not to `compact` it, based on the visibility ratio, chunk dimensions.

To determine optimal numbers to run compaction, we should run some benchmarks which shuffle stream chunks, and check the performance graph, to understand the min and max points.

lyang24:
This is very cool may i work on this issue?

kwannoel:
Another related optimization is no-op chunk compaction. In the case where there is a large chunk with a lot of no-op updates, we can compress these updates before processing. For instance when doing an exchange to a single parallelism.

lyang24:
> Another related optimization is no-op chunk compaction. In the case where there is a large chunk with a lot of no-op updates, we can compress these updates before processing. For instance when doing an exchange to a single parallelism.

noob question how do we know if an update is no-op without checking the underlying data?

kwannoel:
> > Another related optimization is no-op chunk compaction. In the case where there is a large chunk with a lot of no-op updates, we can compress these updates before processing. For instance when doing an exchange to a single parallelism.
> 
> noob question how do we know if an update is no-op without checking the underlying data?

We have to definitely check the data. But we can just check the `keys` and `operations`, don't need to check the `values`.

lyang24:
### 🧪 Testing Method

1. **Enable release mode** in your local RisingWave setup.
2. Start a 3 compute node cluster with recovery:
   ```bash
   ./risedev d ci-3cn-1fe-with-recovery
3. `./risedev sql-bench run shuffle_or_compact_vis{10,50,90} -u postgresql://root@localhost:4566/dev -d` (i set run =1)
4.  run with https://github.com/lyang24/risingwave/tree/compact_shuffle_benchmark for compact before dispatch and https://github.com/lyang24/risingwave/tree/shuffle_benchmark for dispatch

### 🔍 Performance Comparison

| Visibility | No Compact (s) | Compact (s) | Difference (s) | % Improvement |
|------------|----------------|-------------|----------------|----------------|
| 10%        | 6.806          | 4.304       | 2.502          | 36.76%         |
| 50%        | 5.054          | 5.055       | -0.001         | -0.02%         |
| 90%        | 5.807          | 5.805       | 0.002          | 0.03%          |

lyang24:
list testing steps:
[...]

PR #21941, opened 2025-05-21
feat(sql_bench): add a sql benchmark framework using `hyperfine`, `risedev` and `python`

kwannoel:

### Motivation

There are a lot of features and optimizations which require benchmarking to measure their performance change. Currently there's high overhead to add performance benchmarks to our nexmark suite. However, we want a fast feedback loop to iterate on these features. The result is that we write adhoc benchmarks. These are often discarded after the feature/optimization is merged, and they may not measure the performance accurately. Further, we have to write our own benchmark scripts from scratch each time.

After the benchmark framework is added, I will add benchmarks for backfill order control.

Requirements:
- Support local iterative benchmarking
- Support cloud benchmarking
- Store benchmark tests somewhere
- Simplicity

Non-goals:
- We don't want to persist the results. It leads to more maintenance burden due to extra infrastructure. Until we have a concrete need, we can avoid this.
- Instead my suggested approach is to add a `baseline` and the actual `benchmark`. The `baseline` will run in the absence of the optimization. The `benchmark` will run with the optimization.

Solution:
1. A place to add and track benchmarks, which is simpler to update than `nexmark-bench`. The best place would be directly in the RisingWave repository.
2. A simple framework to begin, which can be further improved subsequently, when we have clearer benchmarking patterns.

Some examples of sql-based performance benchmarks that should be done:
- Fixed backfill order control
- Backfill Order Control (make sure auto backfill order derivation improves performance)
- Join isolation (make sure barrier latency decreases)
- Join cache optimization

Comparison with nexmark bench:
| feature | nexmark-bench | sql-bench | remarks |
| - | - | - | - |
| local benchmarking | ❌ | ✅ | Can't run nexmark benchmark locally, you can run sql-bench locally, it will use risedev d. |
| auto provision resource | ✅ | ✅ (for local), ❌ (for remote) | sql-bench only provisions local environment, no cloud env. |
| iteration complexity | ❌ | ✅ | There's a lot of configuration knobs for kube-bench, and each benchmark runs for a long time. For sql-bench we can write shorter benchmarks, run them locally, once we're satisfied, we can run against nexmark-bench. |

Future plans:
- We can incorporate resource provisioning into the benchmark framework.
- We can support persisting benchmark results.

### Changes

Add a simple benchmarking wrapper around `hyperfine` and a template for others to add their own SQL benchmarks.

Check the `README.md` for details.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/develop/sql_bench/README.md
+++ b/develop/sql_bench/README.md
@@ -64,16 +64,16 @@ Run a benchmark using either:
 
 ```bash
 # Against a local RisingWave instance
-python main.py run <benchmark_name>
+risedev sql-bench run <benchmark_name>
 
 # Against a specific PostgreSQL URL
-python main.py run <benchmark_name> -u <postgres_url>
+risedev sql-bench run <benchmark_name> -u <postgres_url>
 
 # Show detailed output
-python main.py run <benchmark_name> -d
+risedev sql-bench run <benchmark_name> -d
 
 # Use a specific RisingWave profile
-python main.py run <benchmark_name> -p <profile>
+risedev sql-bench run <benchmark_name> -p <profile>
 ```
 
 ## Debugging


kwannoel:
A better approach is to improve our existing `nexmark_bench` framework, and make it more maintainable and usable by kernel developers.

The SQLs and test workflow should reside on the kernel side, and `nexmark_bench` can just define interfaces to:
1. Upload performance results.
2. Provision infrastructure for benchmarking.

sql_bench/README.md
```diff
@@ -0,0 +1,31 @@
...
+## Setup a new benchmark
+
+```bash
+cp template.sh <benchmark>.sh
+```
+
+## Run a benchmark
+
+```bash
+PSQL_URL=<...>
```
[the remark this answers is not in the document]
xxchan: I think `risedev d` can also support s3 👀

sql_bench/README.md
```diff
@@ -0,0 +1,31 @@
+# Local e2e SQL Benchmarks
```
xxchan, line 1: nit: I think we should avoid putting it on the top-level directory, we have already way too many, e.g., `scripts`, `standalone`, `develop`, ... 

(The cleanest I've seen is https://github.com/tigerbeetle/tigerbeetle )

Maybe we can put this in `scripts` or `develop` (Maybe many can be put under `develop`

sql_bench/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+This folder contains SQL benchmarks which are ran with [`hyperfine`](https://github.com/sharkdp/hyperfine).
+We can provide a template that wraps the `hyperfine` command,
+and runs it as a bash script.
+
+It's a little premature to design a full benchmark suite for now.
+
+You can provision a cloud instance and run the benchmarks there,
+or run them locally.
```
xxchan, line 10: BTW, we had https://github.com/risingwavelabs/risedev-bench  (unmaintained) for cloud bench

sql_bench/README.md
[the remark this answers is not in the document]
kwannoel: S3 is just one part, we also have to consider things like `disk cache` configuration, now that it's part of the cloud offering, whether local machine has enough space to handle the entire benchmark workload, (in cases like join, likely not).

Anyway this is more like an improvement. We can support it on demand later.

sql_bench/README.md
```diff
@@ -0,0 +1,31 @@
+# Local e2e SQL Benchmarks
+
```
xxchan, line 2: Can put the motivation in the pr description into the README.
e.g., it's most useful for ad-hoc benchmarking for developing optimizations

sql_bench/README.md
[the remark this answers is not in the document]
kwannoel: Thinking twice -- the adhoc benchmarking part is important. Since we can use it to do some preliminary benchmarking, before trying on production environment.

We can add it to the framework.

sql_bench/README.md
kwannoel: good suggestion, I think under `develop` makes sense.

sql_bench/README.md
kwannoel: True, usually in early stages we need to iterate fast. Benchmarking in local env makes sense.

sql_bench/README.md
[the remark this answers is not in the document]
xxchan: Actually, I mean I want to invoke the command via sth like `risedev bench` (like `risedev psql`, so it can bench `risedev d` (can also override the env vars if not started by risedev, or add a special risedev profile for remote cluster

sql_bench/template.sh
```diff
@@ -0,0 +1,122 @@
...
+          left join users on orders.user_id = users.id
+        group by users.name;
+  "
+  run_psql -c "${bench_workload}"
+}
+
+export -f optimized_benchmark
+
+hyperfine --shell=bash --runs 1 --show-output \
+  --prepare 'conclude; prepare' --conclude 'conclude' baseline_benchmark \
```
xxchan, line 121: I think the `--prepare` is for sth like "cleanup cache", the "prepare data" step can be shared by runs, and should run before `hyperfine` (I would prefer this option), or maybe use `--setup`

```
  -s, --setup <CMD>
          Execute CMD before each set of timing runs. This is useful for
          compiling your software with the provided parameters, or to do any
          other work that should happen once before a series of benchmark runs,
          not every time as would happen with the --prepare option.
  -p, --prepare <CMD>
          Execute CMD before each timing run. This is useful for clearing disk
          caches, for example.
          The --prepare option can be specified once for all commands or
          multiple times, once for each command. In the latter case, each
          preparation command will be run prior to the corresponding benchmark
          command.
```

sql_bench/README.md
```diff
@@ -0,0 +1,31 @@
...
+You can provision a cloud instance and run the benchmarks there,
+or run them locally.
+
+**NOTE(kwannoel): These benchmarks are not ran in CI.**
+
+## Setup a new benchmark
+
+```bash
+cp template.sh <benchmark>.sh
```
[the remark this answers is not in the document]
xxchan: BTW, I would prefer other languages than `bash` e.g., python for scripting (similar to https://github.com/risingwavelabs/risingwave/issues/20525)

Makefile.toml
```diff
@@ -565,18 +565,6 @@ dependencies = [
...
-description = "Download necessary tools to deploy a benchmark env"
-dependencies = [
-    "download-minio",
-    "download-mcli",
-    "download-grafana",
-    "download-prometheus",
-    "download-tempo",
-    "download-redis",
-]
-
```
kwannoel, line 579: Doesn't seem used anywhere. Removing.

xxchan (commented):
GEnerally LGTM. Want others to have a look too

stdrc (approved):
LGTM

kwannoel:
Any further comments @xxchan @BugenZhao

kwannoel:
Merge in first, so community user can use framework to do benchmarking for https://github.com/risingwavelabs/risingwave/issues/20428

kwannoel:
Any improvements can be done incrementally.

PR #21944, opened 2025-05-21
Kwannoel/release 2.2 kafka fixes

kwannoel:

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/src/source/kafka/client_context.rs
+++ b/src/connector/src/source/kafka/client_context.rs
@@ -158,11 +158,9 @@ impl KafkaContextCommon {
                         )
                         .await
                     })
-                })
-                .unwrap()
-                .unwrap()
+                })??
             };
-            tracing::info!(?expiration_time_ms, "generated token");
+            tracing::debug!(?expiration_time_ms, "generated token");
             Ok(OAuthToken {
                 token,
                 principal_name: "".to_owned(),
--- a/src/connector/src/source/kafka/client_context.rs
+++ b/src/connector/src/source/kafka/client_context.rs
@@ -164,6 +164,7 @@ impl KafkaContextCommon {
                 .unwrap()
                 .unwrap()
             };
+            tracing::info!(?expiration_time_ms, "generated token");
             Ok(OAuthToken {
                 token,
                 principal_name: "".to_owned(),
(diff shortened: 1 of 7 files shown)


src/connector/src/source/kafka/client_context.rs
```diff
@@ -132,9 +158,11 @@ impl KafkaContextCommon {
                         )
                         .await
                     })
-                });
-                handle.join().unwrap()??
+                })
+                .unwrap()
+                .unwrap()
```
tabVersion, line 163: shall we avoid possible panic? just inspect errors and print them

src/connector/src/source/kafka/client_context.rs
```diff
@@ -132,9 +158,11 @@ impl KafkaContextCommon {
...
             };
+            tracing::info!(?expiration_time_ms, "generated token");
```
tabVersion, line 165: info -> debug? a little annoying

src/tests/simulation/tests/integration_tests/scale/sink.rs
```diff
@@ -103,6 +103,10 @@ async fn test_sink_append_only() -> Result<()> {
             }
 
             consumer.assign(&tpl).unwrap();
+
+            // poll consumer to trigger callback functions
+            let _ = tokio::time::timeout(Duration::from_millis(1000), consumer.recv()).await;
```
tabVersion, line 108: I think we should remove this part, and rely on `spawn_consumer_poll_task`

src/tests/simulation/tests/integration_tests/scale/sink.rs
```diff
@@ -185,6 +189,10 @@ async fn test_sink_debezium() -> Result<()> {
...
```
tabVersion, line 194: REMOVE this, it can cause data loss.

tabVersion (commented):
rest LGTM

src/connector/src/source/kafka/client_context.rs
kwannoel: I think the only way is to return the errors. We need to either return some value or an error, according to the function signature.

tabVersion:
related https://github.com/risingwavelabs/risingwave/pull/21912

src/connector/src/source/kafka/client_context.rs
tabVersion: I dont think we need handle it delicately, at least we need to know what the error is.

tabVersion (approved):
LGTM

PR #17744, opened 2024-07-18
fix(storage): use per-table committed epoch in read version update

wenym1:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://gist.github.com/TennyZhuang/f00be7f16996ea48effb049aa7be4d66#file-rw_cla).

## What's changed and what's your intention?

In HummockReadVersion update, we still use the global max committed epoch. In this PR we will change to use the per table committed epoch

## Checklist

- [ ] I have written necessary rustdoc comments
- [ ] I have added necessary unit tests and integration tests
- [ ] I have added test labels as necessary. See [details](https://github.com/risingwavelabs/risingwave/blob/main/docs/developer-guide.md#ci-labels-guide).
- [ ] I have added fuzzing tests or opened an issue to track them. (Optional, recommended for new SQL features #7934).
- [ ] My PR contains breaking changes. (If it deprecates some features, please create a tracking issue to remove them in the future).
- [x] All checks passed in `./risedev check` (or alias, `./risedev c`)
- [ ] My PR changes performance-critical code. (Please run macro/micro-benchmarks and show the results.)

- [ ] My PR contains critical fixes that are necessary to be merged into the latest release. (Please check out the [details](https://github.com/risingwavelabs/risingwave/blob/main/CONTRIBUTING.md))

## Documentation

- [ ] My PR needs documentation updates. (Please use the **Release note** section below to summarize the impact on users)

## Release note

If this PR includes changes that directly affect users or other significant modifications relevant to the community, kindly draft a release note to provide a concise summary of these changes. Please prioritize highlighting the impact these changes will have on users.

</details>

PR #21946, opened 2025-05-21
fix(storage): correctly apply committed table watermark

wenym1:

We apply the table watermark of committed version to staging table watermark in read version in the following way.

```
 if let Some(committed_watermarks) = self
        .committed
        .version()
        .table_watermarks
        .get(&self.table_id)
    {
        // apply
    }
}
```

Before https://github.com/risingwavelabs/risingwave/pull/17744, we set `self.committed = new_committed_version` before applying to staging table watermark, and therefore the `self.committed` we used is the `new_committed_version`. However, in https://github.com/risingwavelabs/risingwave/pull/17744, we incorrectly changed to set `self.committed = new_committed_version` at the end, and therefore we actually used the table watermark in the previous committed version as the latest committed table watermark, which causes using incorrect table watermark in query.

The consequence is that the start range will be wider than the correct one. The correctness can be satisfied for sink log store, because it requires only at-least-once semantic. However, in the newly supported sync log store, which requires exactly-once semantic, it becomes incorrect. The bug is found in its test https://github.com/risingwavelabs/risingwave/pull/21781.

In this PR, along with the fix, we add a new unit test to cover the case. Before this PR, the unit test will fail.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/storage/benches/bench_table_watermarks.rs
+++ b/src/storage/benches/bench_table_watermarks.rs
@@ -26,7 +26,8 @@ use risingwave_common::util::epoch::test_epoch;
 use risingwave_hummock_sdk::HummockEpoch;
 use risingwave_hummock_sdk::compaction_group::StaticCompactionGroupId;
 use risingwave_hummock_sdk::table_watermark::{
-    TableWatermarks, TableWatermarksIndex, VnodeWatermark, WatermarkDirection, WatermarkSerdeType,
+    PkPrefixTableWatermarksIndex, TableWatermarks, VnodeWatermark, WatermarkDirection,
+    WatermarkSerdeType,
 };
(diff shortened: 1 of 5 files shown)


wenym1:
* **#21946** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/21946?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/21946?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

kwannoel (approved):

stdrc:
A user encountered an issue caused by this bug on v2.4. I think we also need to cherry-pick this to release-2.4.

PR #21957, opened 2025-05-22
fix(iceberg): dropping database and dropping source cascade with iceberg tables.

chenzl25:

- Ban dropping database and dropping source cascade with iceberg tables.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/meta/src/controller/catalog/drop_op.rs
+++ b/src/meta/src/controller/catalog/drop_op.rs
@@ -139,11 +139,9 @@ impl CatalogController {
             }
         }
 
-        // 1. Detect the case where the iceberg table created from a shared source and drop source cascade.
-        // Currently, iceberg engine table doesn't support atomic ddl. To drop the source, we need to drop the iceberg table first.
-        //
+        // 1. Detect when an Iceberg table is part of the dependencies.
         // 2. Drop database with iceberg tables in it is not supported.
-        if object_type == ObjectType::Source || drop_database {
+        if object_type != ObjectType::Table || drop_database {
             for obj in &removed_objects {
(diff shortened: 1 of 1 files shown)


src/meta/src/controller/catalog/drop_op.rs
```diff
@@ -138,6 +139,28 @@ impl CatalogController {
             }
         }
 
+        // 1. Detect the case where the iceberg table created from a shared source and drop source cascade.
+        // Currently, iceberg engine table doesn't support atomic ddl. To drop the source, we need to drop the iceberg table first.
+        //
+        // 2. Drop database with iceberg tables in it is not supported.
+        if object_type == ObjectType::Source || drop_database {
```
xxchan, line 146: Can we just check `if object_type != ObjectType::Table && drop_mode == DropMode::Cascade` here?

chenzl25: I think only the source and database can have dependencies on tables.

xxchan: But what we actually care is only "the iceberg table is dropped as a dependency", we don't care who depends on the table. So I think checking `object_type != ObjectType::Table` is slightly more robust

chenzl25: > But what we actually care is only "the iceberg table is dropped as a dependency", we don't care who depends on the table. So I think checking object_type != ObjectType::Table is slightly more robust

yep, it makes sense. Let me change it to `object_type != ObjectType::Table`

Li0k (approved):

PR #21960, opened 2025-05-22
feat(iceberg): init and backup hosted iceberg system table

chenzl25:

- Previously, we created the hosted iceberg catalog system related tables on demand; however, our backup handling logic needs to specify models, so we need to initialize the tables at the beginning.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/meta/model/src/iceberg_namespace_properties.rs
+++ b/src/meta/model/src/iceberg_namespace_properties.rs
@@ -1,0 +1,34 @@
+// Copyright 2025 RisingWave Labs
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use sea_orm::entity::prelude::*;
+use sea_orm::{DeriveEntityModel, DeriveRelation, EnumIter};
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Debug, PartialEq, DeriveEntityModel, Eq, Default, Serialize, Deserialize)]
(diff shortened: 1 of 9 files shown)


src/meta/model/src/iceberg_namespace_properties.rs
```diff
@@ -0,0 +1,34 @@
+// Copyright 2025 RisingWave Labs
```
xxchan, line 1: I don't get why we need this. Is it related with #21400? Isn't that a frontend only view?

zwang28 (commented):
Please add `insert_models` in src/meta/src/backup_restore/restore_impl/v2.rs, referring to https://github.com/risingwavelabs/risingwave/pull/21939.

We need add `insert_models` for new models in addition to defining them in `for_all_metadata_models_v2`, because insert order manners, e.g. foreign key constraints.

src/meta/model/src/iceberg_namespace_properties.rs
chenzl25: It is related. `iceberg_namespace_properties` is system table used for JDBC catalog, I think we need this initialization for backup, even though currently, it is an empty table. In case we need this table in the future.

xxchan: > we need this initialization for backup

> Previously, we created the hosted iceberg catalog system related tables on demand

I still don't get it. IIUC, `iceberg_namespace_properties` and `iceberg_tables` are not created in meta. Why we need them for backup?

chenzl25: The JDBC catalog skd could help us create these 2 tables if they do not exist (i.e., an empty table), but once we write data to them, we are responsible for maintaining them (including backup).

xxchan: Oh, I finally get it.. So we are talking about meta backend as JDBC catalog for RW usage, which does need to store some metadata. 

RW as JDBC catalog (the user-visible system tables) is also just expose this. I thought it's constructed from some other info..

Makes sense that they need backup.

chenzl25: > RW as JDBC catalog (the user-visible system tables) is also just expose this. I thought it's constructed from some other info..

Yes, RW as JDBC catalog is read-only, unless users can access the meta store directly (like self-hosted customers)

zwang28 (approved):

Li0k (approved):
LGTM

Issue #21846, opened 2025-05-13
SQL Server CDC should do schema check and return error when encountering an unsupported type


wcy-fdu:
### Describe the bug

As title.

### Error message/log

```text

```

PR #21961, opened 2025-05-22
fix(mssql-cdc): change the default behavior of schema checks for unsupported data types to reject

wcy-fdu:

We used to do schema check, while for unsupported data type, it's default behavior is `true`, which means all unsupported data type will pass into backend.

This pr fix the default behavior for unsupported data type, and add the missing check for `xml` and `money`.

Close https://github.com/risingwavelabs/risingwave/issues/21846

## Checklist

## Documentation

<details>

</details>


src/connector/src/source/cdc/external/mod.rs
```diff
@@ -362,7 +365,7 @@ impl ExternalTableImpl {
                 .await?,
             )),
             CdcSourceType::SqlServer => Ok(ExternalTableImpl::SqlServer(
-                SqlServerExternalTable::connect(config).await?,
+                SqlServerExternalTable::connect(config, defined_column_descs).await?,
```
xiangjinwu: How does other CDC tables (PostgreSQL / MySQL) verify user specified column data types against derived ones? In Java part?

src/connector/src/source/cdc/external/sql_server.rs
```diff
@@ -121,10 +131,15 @@ impl SqlServerExternalTable {
...
                         let col_type: &str = row.try_get(1)?.unwrap();
+
                         column_descs.push(ColumnDesc::named(
                             col_name,
                             ColumnId::placeholder(),
-                            mssql_type_to_rw_type(col_type, col_name)?,
+                            check_mssql_type_and_convert_to_rw_type(
+                                col_type,
+                                col_name,
+                                column_name_to_data_type.clone(),
```
xiangjinwu: It is better to do the lookup here and just pass `Option<&DataType>` inside.

src/frontend/src/handler/create_table.rs
```diff
@@ -1141,10 +1141,17 @@ pub(super) async fn handle_create_table_plan(
...
+                        .map(|catalog| catalog.column_desc.clone())
+                        .collect();
                     // read default value definition from external db
                     let (options, secret_refs) = cdc_with_options.clone().into_parts();
-                    let _config = ExternalTableConfig::try_from_btreemap(options, secret_refs)
+                    let config = ExternalTableConfig::try_from_btreemap(options, secret_refs)
                         .context("failed to extract external table config")?;
+                    ExternalTableImpl::connect(config, Some(defined_column_descs))
+                        .await
+                        .context("failed to auto derive table schema")?;
```
xiangjinwu: failed to validate table schema

xiangjinwu (commented):
Can we add a test case for this?

wcy-fdu:
> Can we add a test case for this?

Sure, working on it.

src/connector/src/source/cdc/external/mod.rs
wcy-fdu: Yes, actually those cdc(pg, mysql, mssql) all implement the schema check logic in` xxxValidater.java`, but somehow the sqlserver cdc bypass this check and I just found we need to do double check here.

e2e_test/source_inline/cdc/sql_server/sql_server_cdc.slt.serial
```diff
@@ -172,7 +172,7 @@ CREATE TABLE shared_orders (
 )  from mssql_source table 'orders';
 
 # invalid table name
-statement error Sql Server table 'dbo'.'wrong_orders' doesn't exist
+query error
```
xiangjinwu: Only needs to update `doesn't exist` to `not found` for the test to pass. And we'd better unify the error message generated from different places on the same issue later.

e2e_test/source_inline/cdc/sql_server/sql_server_cdc.slt.serial
```diff
@@ -194,9 +202,17 @@ CREATE TABLE shared_orders (
...
+
+Caused by these errors (recent errors listed first):
+  1: connector error
+  2: Sql Server table 'wrong_dbo'.'orders' not found in 'mydb'
+
+
 
 # wrong database name
-statement error The database name `wrong_mydb` in the FROM clause is not the same as the database name `mydb` in source definition
+query error
```
xiangjinwu: Unnecessary change

e2e_test/source_inline/cdc/sql_server/sql_server_cdc.slt.serial
```diff
@@ -230,9 +263,18 @@ CREATE TABLE shared_orders (
...
+Caused by these errors (recent errors listed first):
+  1: gRPC request to meta service (call `/ddl_service.DdlService/CreateTable`) failed: Internal error
+  2: source cannot pass validation
+  3: INVALID_ARGUMENT: Column 'wrong_order_date' not found in the upstream database
+
+
 
 # column data type mismatch
-statement error INVALID_ARGUMENT: Incompatible data type of column order_date
+query error
```
xiangjinwu: Seems we had test case for incompatible data type. Why was it not enough?

src/connector/src/source/cdc/external/mod.rs
xiangjinwu: > somehow the sqlserver cdc bypass this check and I just found we need to do double check here

Could you elaborate on `somehow`? Is such double checking already neccessary for pg / mysql?

xiangjinwu (commented):
The recent update did not add any new test cases. It only updates the error message wording in 3 cases:
* `doesn't exist` -> `not found`: similar error now captured by an earlier check
* `Incompatible data type of column` -> `Type mismatch for column`: the "missing" check we are targeting
* Others are unnecessary changes that only introduce noises for review. Please avoid such changes.

wcy-fdu:
cc @xiangjinwu PTAL❤️

xiangjinwu (approved):
Remember to update the PR description accordingly. We had a check, but it was allowing arbitrary match for unknown types ...

PR #21984, opened 2025-05-24
feat: add support for rw_uint256 data type

chriswessels:

*Last updated: May 31st 2025*

This pull request introduces a new `rw_uint256` data type to RisingWave, enabling the storage and manipulation of unsigned 256-bit integers. This enhancement is primarily aimed at supporting use cases involving very large non-negative numbers, such as those found in blockchain data (e.g., token amounts, transaction IDs), high-precision counters, and other scenarios where standard integer types are insufficient.

**Key Goals:**

*   Provide native support for unsigned 256-bit integers within the RisingWave type system.
*   Enable full SQL functionality, including DDL, DML, arithmetic operations, comparisons, and aggregate functions for `rw_uint256`.
*   Ensure sensible integration with common data sources and sinks, with appropriate type mappings and conversions.
*   Maintain robust error handling for operations like overflow, underflow, and invalid casts.

### Detailed Changes and Considerations:

**1. Core Type System & Representation:**

*   **New `DataType` Variant:** `DataType::UInt256` has been added to represent the unsigned 256-bit integer type.
*   **Scalar Representation:** `ScalarImpl::UInt256` and `ScalarRefImpl::UInt256` are introduced for handling individual `rw_uint256` values.
*   **Underlying Representation:** The `UInt256` struct is implemented, wrapping `ethnum::u256` (from the `ethnum` crate, already a dependency) to provide the core arithmetic and storage capabilities, similar to how `Int256` utilizes `ethnum::i256`.
*   **Array Structures:** `UInt256Array` and `UInt256ArrayBuilder` are now available for efficient, vectorized processing of `rw_uint256` data.
*   **Protobuf Definitions:** The `data.proto` file has been updated to include `UINT256` in the `DataType.TypeName` and `ArrayType` enums, ensuring compatibility for serialization and communication within the distributed system.
*   **Storage Encoding (ValueEncoding):** `rw_uint256` values are serialized as 32-byte little-endian byte arrays, aligning with the existing `Int256` serialization for consistency.
*   **Memcomparable Encoding:** For efficient sorting and indexing (e.g., in `ORDER BY` clauses, join keys), `rw_uint256` values are serialized using big-endian byte order to ensure correct lexicographical comparison.

**2. SQL Surface and Semantics:**

*   **Type Declaration:** Users can now define columns of type `rw_uint256` in DDL statements (e.g., `CREATE TABLE example (id INT, large_unsigned_val rw_uint256);`).
*   **Literal Parsing:**
    *   `rw_uint256` values are primarily input as string literals (e.g., `'123456789...'::rw_uint256`).
    *   These string literals can represent numbers in decimal format.
    *   Hexadecimal string literals (e.g., `'0x...'::rw_uint256`) are also supported for input, providing flexibility for blockchain-related data.
*   **Arithmetic Operations:**
    *   Standard operators (`+`, `-`, `*`, `/`, `%`) are implemented for `rw_uint256` operands.
    *   All arithmetic operations include robust overflow and underflow checks. Any operation resulting in a value outside the `[0, 2^256 - 1]` range will raise a "Numeric value out of range" error.
    *   Division by zero (`x / 0` or `x % 0`) will result in a "division by zero" error.
*   **Comparison Operators:** All standard comparison operators (`=`, `!=`, `<`, `<=`, `>`, `>=`) are implemented and behave as expected for unsigned integers.
*   **Unary Minus:** The unary minus operator (`-`) is not directly applicable to `rw_uint256` as it's an unsigned type. Attempting `SELECT -some_uint256_column` will result in a parser error. The `neg(uint256)` function is provided, which returns 0 if the input is 0, and errors for any other positive input, reflecting the impossibility of representing a negative result.
*   **`abs()` function:** Returns the input `rw_uint256` value, as unsigned integers are always non-negative.
*   **Type Promotion:**
    *   When `rw_uint256` is used in operations with smaller integer types (`SMALLINT`, `INT`, `BIGINT`), the result is promoted to `rw_uint256`. An error will occur if the smaller integer is negative.
    *   Operations involving `rw_uint256` and `FLOAT8` will promote the result to `FLOAT8`.
    *   Operations involving `rw_uint256` and `DECIMAL` will promote the result to `DECIMAL`.
*   **System Catalog (`pg_catalog`) Updates:**
    *   `pg_type`: New entries have been added for `rw_uint256` (OID 1305) and its array type `_rw_uint256` (OID 1306). This includes metadata like `typinput`, `typoutput`, `typbasetype`, `typlen`, `typalign`, and `typcategory`.
    *   `pg_cast`: The catalog has been updated to include definitions for all supported implicit (`i`) and explicit (`e`) casts involving `rw_uint256` and other compatible types (e.g., `int2`, `int4`, `int8`, `numeric`, `varchar`, `float8`).

**3. Casting Behavior:**

*   **To `rw_uint256`:**
    *   From `INT2`, `INT4`, `INT8`, `SERIAL`: Supported (implicit and explicit). Raises an error if the source value is negative.
    *   From `VARCHAR`: Supported (explicit). Parses the string as a decimal or hexadecimal number. Errors on invalid format or if the value is outside the representable range of `rw_uint256`.
    *   From `DECIMAL`: Supported (explicit). The decimal value is rounded to the nearest integer. Errors if the value is negative or exceeds the `rw_uint256` range.
    *   From `INT256`: Explicit cast; errors if the `INT256` value is negative.
*   **From `rw_uint256`:**
    *   To `VARCHAR`: Converts the `rw_uint256` to its string representation.
    *   To `INT256`: Explicit cast. Errors if the `rw_uint256` value is too large to be represented as a signed 256-bit integer (i.e., if the most significant bit is set, which would indicate a negative number in `Int256`).
    *   To `FLOAT8`: Converts to a double-precision floating-point number. Potential precision loss may occur for very large `rw_uint256` values.
    *   To `DECIMAL`: Converts to a decimal number.
    *   To `INT2`, `INT4`, `INT8`: Explicit casts. Errors if the `rw_uint256` value is too large to fit within the target integer type's range.

**4. Aggregate Functions:**

*   `COUNT(rw_uint256)`: Standard count behavior.
*   `SUM(rw_uint256)`: Returns `rw_uint256`. Internally uses a wider accumulator type (currently `u512`) to mitigate overflow during intermediate calculations. An error will be raised if the final sum exceeds `UInt256::MAX`.
*   `AVG(rw_uint256)`: Returns `FLOAT8`.
*   `MIN(rw_uint256)`, `MAX(rw_uint256)`: Return `rw_uint256`.
*   `STDDEV_POP(rw_uint256)`, `STDDEV_SAMP(rw_uint256)`, `VAR_POP(rw_uint256)`, `VAR_SAMP(rw_uint256)`: All return `FLOAT8`.

**5. Connector Interactions:**

*   **PostgreSQL Source/Sink:**
    *   `rw_uint256` is mapped to PostgreSQL's `NUMERIC` type.
    *   Data is exchanged as strings to preserve precision and handle the full range.
    *   `PostgresSink::type_derive()` and `ScalarAdapter` have been updated to handle `rw_uint256`.
*   **MySQL & SQL Server Sources:**
    *   `DECIMAL` and `NUMERIC` columns from these sources can now be correctly mapped to `rw_uint256` (parsed from string format).
*   **JSON Source/Sink:**
    *   `rw_uint256` values are represented as strings in JSON payloads (e.g., `"12345"`). This ensures compatibility with JSON parsers that might not handle numbers of this magnitude directly.
    *   When parsing JSON, string values are preferred for conversion to `rw_uint256`.
*   **Apache Arrow (Internal & IPC):**
    *   Internally, when data is processed in Arrow format (e.g., during certain vectorized operations or for inter-process communication), `rw_uint256` is represented as `FixedSizeBinary(32)` with big-endian byte order. This allows for efficient storage and preserves correct sorting order.
*   **Parquet/Iceberg Sinks:**
    *   When creating table schemas for Iceberg (which typically uses Parquet for storage), `rw_uint256` columns are mapped to `VARCHAR` (string) type. This ensures broader compatibility with tools and systems reading Parquet files, which may not have native support for 256-bit unsigned integers but can handle large numbers as strings.
*   **Other Sinks (Avro, Protobuf, Generic JDBC):**
    *   Direct native support for `rw_uint256` in formats like Avro and Protobuf is generally not available.
    *   For these sinks, and for generic JDBC sinks where the target database type compatibility is uncertain, it is recommended to explicitly cast `rw_uint256` values to `VARCHAR` within the `CREATE SINK` statement to ensure data integrity and avoid potential errors.

**6. Testing:**

*   **New Test Files:** Added `e2e_test/batch/basic/uint256.slt` and `e2e_test/batch/types/uint256.slt.part` for comprehensive testing of `rw_uint256` functionalities.
*   **Existing Test Modifications:** Updated `to_jsonb.slt.part`, `pg_catalog.slt`, `pg_type.slt`, `pg_cast.slt`, and relevant connector tests (PostgreSQL, MySQL, SQL Server sources; PostgreSQL sink, Iceberg sink scenarios) to include `rw_uint256` test cases.
*   **Internal Unit Tests:** Implemented comprehensive unit tests for `UInt256` type operations such as arithmetic, parsing, serialization, and boundary condition handling.

**7. Potential Failure Modes & Considerations:**

*   **Arithmetic Errors:** Standard overflow, underflow, and division by zero errors will be raised.
*   **Invalid Casts:**
    *   Casting negative numeric values or non-numeric strings to `rw_uint256` will result in errors.
    *   Casting an `rw_uint256` value that exceeds the maximum value of a signed 256-bit integer to `rw_int256` will fail.
    *   Casting to smaller integer types (`INT2`, `INT4`, `INT8`) will fail if the `rw_uint256` value exceeds their respective maximums.
*   **Sink Compatibility:** Users should verify the target data type compatibility when sinking `rw_uint256` values to external systems. For sinks without direct support, explicit casting to `VARCHAR` in the sink definition is the safest approach.

This addition aims to provide a robust and flexible solution for handling large unsigned integers in RisingWave. We believe these changes will be valuable for users working with specific data domains.

## Checklist

- [x] I have checked the [Release Timeline](https://github.com/risingwavelabs/rw-commits-history/blob/main/release_timeline.md) and [Currently Supported Versions](https://docs.risingwave.com/changelog/release-support-policy#support-end-dates-for-recent-releases) to determine which release branches I need to cherry-pick this PR into.

## Documentation

<details>

### ✨ New Data Type: `rw_uint256` for Unsigned 256-bit Integers

RisingWave now supports the `rw_uint256` data type, designed for handling extremely large unsigned integer values, ranging from 0 up to 2^256 - 1. This is particularly beneficial for applications dealing with blockchain data, high-precision counters, or other scenarios requiring extended-range unsigned integers.

**Key Features and Behavior:**

*   **SQL Functionality:**
    *   **Declaration:** You can now create tables with columns of type `rw_uint256`.
        ```sql
        CREATE TABLE token_balances (
            address VARCHAR,
            balance rw_uint256
        );
        ```
    *   **Literals:** Insert values using string literals, which can be in decimal or hexadecimal (prefixed with `0x`) format.
        ```sql
        INSERT INTO token_balances VALUES
        ('0xAlice', '123456789012345678901234567890'::rw_uint256),
        ('0xBob', '0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff0000'::rw_uint256);
        ```
    *   **Arithmetic Operations:** Supports `+`, `-`, `*`, `/`, `%` with overflow and division-by-zero checks.
    *   **Comparison Operators:** Standard comparison operators (`=`, `!=`, `<`, `<=`, `>`, `>=`) work as expected.
    *   **Aggregate Functions:** `SUM`, `AVG` (returns `FLOAT8`), `MIN`, `MAX`, `COUNT`, and statistical aggregates like `STDDEV_POP` are supported.
    *   **Casting:**
        *   Cast from smaller integer types (`SMALLINT`, `INT`, `BIGINT`), `DECIMAL`, and `VARCHAR`. Casting negative values will result in an error.
        *   Cast to `VARCHAR`, `FLOAT8`, `DECIMAL`, and other integer types (with potential range errors).
*   **Connector Integration:**
    *   **PostgreSQL Source/Sink:** Maps to PostgreSQL's `NUMERIC` type, facilitating data exchange via string representation.
    *   **MySQL & SQL Server Sources:** `DECIMAL` and `NUMERIC` columns can be read as `rw_uint256`.
    *   **JSON Source/Sink:** `rw_uint256` values are represented as strings.
    *   **Iceberg/Parquet Sinks:** For table creation with Iceberg, `rw_uint256` columns will be mapped to `VARCHAR` (string) type in the Parquet files. Internally, for Arrow processing, `FixedSizeBinary(32)` is used.
    *   **Other Sinks (e.g., Avro, Protobuf, other JDBC):** Direct native support for `rw_uint256` may be limited. It's recommended to cast `rw_uint256` to `VARCHAR` in your `CREATE SINK` statement for these connectors.

**Example Usage:**

```sql
-- Create a table to store cryptocurrency balances
CREATE TABLE crypto_wallets (
    wallet_address VARCHAR PRIMARY KEY,
    token_balance_wei rw_uint256
);

-- Insert some sample data
INSERT INTO crypto_wallets VALUES
    ('0x123...', '150000000000000000000'::rw_uint256), -- 150 Ether in Wei
    ('0xabc...', '7500000000000000000000'::rw_uint256);  -- 7500 Ether in Wei

-- Calculate total supply in Wei
SELECT SUM(token_balance_wei) FROM crypto_wallets;

-- Find wallets with more than 1000 Ether (10^21 Wei)
SELECT wallet_address, token_balance_wei
FROM crypto_wallets
WHERE token_balance_wei > '1000000000000000000000'::rw_uint256;
```

**Use Cases:**

*   Storing and processing cryptocurrency balances (e.g., Wei for Ethereum).
*   Tracking large identifiers or counters in distributed systems.
*   Any application requiring integer precision beyond standard 64-bit integers without resorting to `NUMERIC` for whole numbers.

**Important Notes:**
*   `rw_uint256` strictly represents non-negative integers. Attempting to cast negative values to `rw_uint256` will result in an error.
*   Arithmetic operations that would result in a value outside the `0` to `2^256 - 1` range (e.g., `0::rw_uint256 - 1`) will cause an error.

</details>


lmatz:
Hi @chriswessels, thanks very much for the contribution

>Connector integration: Source ingestion from PostgreSQL, JSON, and Avro; sink support for major systems

Could you add a few source/sink test cases in https://github.com/risingwavelabs/risingwave/tree/main/e2e_test/source_inline and https://github.com/risingwavelabs/risingwave/tree/main/e2e_test/sink?

chriswessels:
Thanks for the quick engagement here @lmatz! I've pushed additional testing of the type (and fixed the CAST_TABLE).

Currently, the tests reveal an issue with `rw_uint256` ordering.

**Problem**: The `UInt256Ref::memcmp_serialize` implementation produces incorrect ordering when used in SQL `ORDER BY` clauses. Large 256-bit values (requiring high word) are sorted before smaller values that fit in 64 bits.

**Likely Root Cause**: Naive implementation serializes as `(u128, u128)` tuple:
```rust
let (hi, lo) = self.0.into_words();
(hi, lo).serialize(serializer)
```
The memcomparable crate's tuple encoding doesn't preserve correct 256-bit precedence relationships.

**Test Failure**: Values like `100000000000000000000000000000000` appear before `18446744073709551615` in ascending order.

**Solutions Attempted**:
1. ✗ Cast high word to `i128`: `(hi as i128, lo).serialize(serializer)`
2. ✗ Sequential serialization: `hi.serialize(); lo.serialize()`  
3. ✗ Big-endian byte array: `self.0.to_be_bytes().serialize(serializer)`
4. ✗ Raw byte-by-byte: `for byte in self.0.to_be_bytes() { serializer.serialize_u8(byte) }`

All approaches still produce the same incorrect ordering. The issue may be deeper in how memcomparable handles large integer comparisons or we're missing something fundamental about the encoding requirements.

**Question**: What's the recommended approach for implementing memcomparable serialization for unsigned 256-bit integers?

chriswessels:
## UInt256 Implementation Update

### 1. ORDER BY Correctness for Large UInt256 Values

**Problem**: UInt256 values >= 2^255 were sorting incorrectly. The root cause was the Arrow conversion using `Decimal256Array`, which interprets values as signed integers. This caused large unsigned values (with MSB set) to be treated as negative numbers.

**Solution**: Implemented a dual representation approach:
- **Internal**: Convert to `FixedSizeBinary(32)` with big-endian encoding for Arrow representation
- **External**: Maintain string representation for sink connectors (Iceberg, Parquet, etc.)

**Why FixedSizeBinary?**
- Arrow lacks a native unsigned 256-bit integer type
- `Decimal256Array` is inherently signed, causing misinterpretation of values >= 2^255
- `FixedSizeBinary(32)` with big-endian encoding provides correct unsigned integer ordering through byte-wise comparison
- This approach is used by other systems handling large unsigned integers (e.g., Ethereum addresses)
- Maintains the same memory efficiency as the previous representation

This ensures correct byte-wise comparison for sorting while preserving compatibility with external systems.

### 2. Missing Arithmetic Operators

Added comprehensive arithmetic support between UInt256 and signed integers (int2, int4, int8):
- `add(uint256, int*)` and `add(int*, uint256)`
- `subtract(uint256, int*)` and `subtract(int*, uint256)` 
- `multiply(uint256, int*)` and `multiply(int*, uint256)`
- `modulus(uint256, int*)` and `modulus(int*, uint256)`

Note: Division operators were already implemented in previous commits.

### 3. Sink Connector Support

Fixed JSON encoder to properly handle Int256/UInt256 scalars by converting them to string representation, ensuring compatibility with external systems.

## Technical Details

### Arrow Conversion Changes

```rust
// Before (incorrect for large values)
fn uint256_to_arrow(&self, array: &UInt256Array) -> Result<ArrayRef, ArrayError> {
    Ok(Arc::new(Decimal256Array::from(array)))
}

// After (correct unsigned ordering)
fn uint256_to_arrow(&self, array: &UInt256Array) -> Result<ArrayRef, ArrayError> {
    // Convert to FixedSizeBinary(32) with big-endian bytes
    let mut builder = FixedSizeBinaryBuilder::with_capacity(array.len(), 32);
    for value in array.iter() {
        match value {
            Some(uint256) => builder.append_value(uint256.to_be_bytes())?,
            None => builder.append_null(),
        }
    }
    Ok(Arc::new(builder.finish()))
}
```

### Cast Table Implications

The conservative cast table (requiring explicit casts from signed to unsigned) remains unchanged. This design choice prevents runtime failures by catching type mismatches at planning time.

## Testing

- Added comprehensive ORDER BY tests with edge cases around 2^255 boundary
- Verified correct sorting for values: 0, 1, 2^64-1, 2^128-1, 2^255-1, 2^255, 2^255+1, 2^256-1
- Updated existing tests to use explicit casts where required by the conservative cast table
- Replaced `rowsort` with explicit `ORDER BY` in tests because `rowsort` performs lexicographic sorting on string representations, which doesn't match numeric ordering for large integers
- All tests pass successfully

chriswessels:
Hey @lmatz! I've added further test coverage and the full suite is now passing. I've updated the [PR description](https://github.com/risingwavelabs/risingwave/pull/21984#issue-3088616289) with a fresh overview of the changes.

src/common/src/types/num256.rs
```diff
@@ -227,6 +260,36 @@ impl Int256 {
...
+impl UInt256Ref<'_> {
+    pub fn memcmp_serialize(
+        &self,
+        serializer: &mut memcomparable::Serializer<impl bytes::BufMut>,
+    ) -> memcomparable::Result<()> {
+        // Big-endian rendering of the full 256-bit value.
+        // We push the 32 bytes one-by-one with `serialize_u8`, which the
+        // docs guarantee writes a *single* raw byte – no length tags,
+        // no terminators. :contentReference[oaicite:0]{index=0}
+        for byte in self.0.to_be_bytes() {
```
lmatz: For Int256, it is serialized by 
```
let (hi, lo) = self.0.into_words();
(hi, lo as u128).serialize(serializer)
```
This is the same as how int256 is compared with each other in the ethum crate:
https://github.com/nlordell/ethnum-rs/blob/main/src/int/cmp.rs#L34-L36
```
let (ahi, alo) = self.into_words();
let (bhi, blo) = other.into_words();
(ahi, alo as u128).cmp(&(bhi, blo as u128))
```

curious why uint256 is not done in the same way by following how it is compared with each other in the ethum crate:
https://github.com/nlordell/ethnum-rs/blob/main/src/uint/cmp.rs#L33
```
self.into_words().cmp(&other.into_words())
```

chriswessels: Thanks for catching this @lmatz!

I did start with your suggested approach, and agree it is more consistent and preferred.

I think I landed up with the bytes iterator approach (which I believe is functionally equivalent) after trying to chase down a sorting bug that landed up being unrelated and in the Arrow type conversions.

I've pushed this update:
```rust
let (hi, lo) = self.0.into_words();
(hi, lo).serialize(serializer)
```

lmatz: Thanks!

src/expr/impl/tests/sig.rs
lmatz, line 1: `least()` and `greatest()`
include `rw_int256`, is it necessary to include `rw_uint256`?

src/frontend/src/catalog/system_catalog/rw_catalog/rw_types.rs
```diff
@@ -54,11 +54,14 @@ macro_rules! impl_pg_type_data {
...
             // with PostgreSQL.
             (25, "text", "textin",0,1009),
-            (1301, "rw_int256", "rw_int256_in",0,0),
+            (1301, "rw_int256", "rw_int256_in",0,1302),
+            (1305, "rw_uint256", "rw_uint256_in",0,1306),
             // Note: Here is only to avoid some components of psql from not being able to find relevant results, causing errors. We will not use it in the RW.
             $(
             ($oid_array, concat!("_", stringify!($name)), "array_in", $oid, 0),
             )*
+            (1302, "_rw_int256", "array_in", 1301, 0),
```
lmatz, line 63: What's the effect of adding these two lines besides the changes of the result set in `e2e_test/batch/catalog/pg_type.slt.part`?

e2e_test/batch/types/uint256.slt.part
```diff
@@ -0,0 +1,405 @@
...
+select v::varchar, pg_typeof(v) from (select 1::rw_uint256 + 1::rw_uint256) as t(v);
+----
+2 rw_uint256
+
+# Edge case: subtraction that would go negative should fail
+statement error
+select 0::rw_uint256 - 1::rw_uint256;
+
+# Edge case: overflow should fail
+statement error
```
lmatz, line 262: the detailed error message can also be tested

example: https://github.com/risingwavelabs/risingwave/blob/main/e2e_test/ddl/table/generated_columns.slt.part#L17

java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/MySqlValidator.java
```diff
@@ -299,7 +299,12 @@ private boolean isDataTypeCompatible(String mysqlDataType, Data.DataType.TypeNam
             case "double":
                 return val == Data.DataType.TypeName.DOUBLE_VALUE;
             case "decimal":
-                return val == Data.DataType.TypeName.DECIMAL_VALUE;
+                return val == Data.DataType.TypeName.DECIMAL_VALUE
```
lmatz, line 302: how about adding a few corresponding test cases for MySQL also?

java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/SqlServerValidator.java
```diff
@@ -315,7 +315,12 @@ private boolean isDataTypeCompatible(String ssDataType, Data.DataType.TypeName t
                 return val == Data.DataType.TypeName.DOUBLE_VALUE;
             case "decimal":
             case "numeric":
-                return val == Data.DataType.TypeName.DECIMAL_VALUE;
+                return val == Data.DataType.TypeName.DECIMAL_VALUE
```
lmatz, line 318: also for sqlserver, thanks
the existing `e2e_test/source_line` include both mysql and sqlserver

lmatz:
Hi @chriswessels 

> ORDER BY Correctness for Large UInt256 Values
Problem: UInt256 values >= 2^255 were sorting incorrectly. The root cause was the Arrow conversion using Decimal256Array, which interprets values as signed integers. This caused large unsigned values (with MSB set) to be treated as negative numbers.

I built the current branch by reverting back to
```
// Before (incorrect for large values)
fn uint256_to_arrow(&self, array: &UInt256Array) -> Result<ArrayRef, ArrayError> {
    Ok(Arc::new(Decimal256Array::from(array)))
}
```

both `e2e_test/batch/basic/uint256.slt.part` and `e2e_test/batch/types/uint256.slt.part` can pass,
Could you point out the test case that the old conversion failed to pass?

src/frontend/src/catalog/system_catalog/rw_catalog/rw_types.rs
chriswessels: As far as I understand, they provide PostgreSQL-compatible type metadata that client tools expect from the PG wire protocol. It allows tools like psql to correctly display and format these types (e.g. arrays of uint256), and JDBC drivers to provide proper metadata about result types.

PR #22019, opened 2025-05-27
feat(common): support input/output of new data type `vector(n)`

xiangjinwu:

The new data type `vector(n)` is expected to match most of the behavior of `vector` from [pgvector](https://github.com/pgvector/pgvector/blob/v0.8.0/README.md). Compared to existing `real[]`:
* The length parameter in the data type is NOT [ignored](https://www.postgresql.org/docs/17/arrays.html#:~:text=the%20current%20implementation%20ignores%20any%20supplied%20array%20size%20limits).
* Elements can NOT be `null`, `nan`, `inf` or `-inf`.

Different from pgvector, the length parameter is always required. In pgvector it is optional for simple operations but required when building index.

This PR fills in some todos left by #21380 to support basic usage:
* convert to / from protobuf
* convert to / from pgwire text format and varchar
  * literal in `explain` output
* assign pg_type oids
* convert to / from value encoding
* parse and bind `vector(n)` from input SQL
  * convert back to ast
* column name derivation for `select null::vector(3);`

Future PR:
* generic operations on all data types with tests, and return unsupported error rather than panic with todo

This implementation uses an inner `ListArray` to mimic the expected semantic. We may consider switching to a more direct implementation that could potentially BREAK existing value encoding (persisted data).

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/common/src/array/vector_array.rs
+++ b/src/common/src/array/vector_array.rs
@@ -304,7 +304,7 @@ impl<'a> VectorRef<'a> {
     }
 
     /// Get the slice of floats in this vector.
-    pub fn into_slice(self) -> &'a [crate::types::F32] {
-        self.inner.as_primitive_slice().unwrap()
+    pub fn into_slice(self) -> &'a [f32] {
+        crate::types::F32::inner_slice(self.inner.as_primitive_slice().unwrap())
     }
 }
--- a/src/common/src/types/ordered_float.rs
+++ b/src/common/src/types/ordered_float.rs
@@ -106,6 +106,13 @@ impl<T: Float> OrderedFloat<T> {
     pub fn into_inner(self) -> T {
         self.0
     }
+
+    pub fn inner_slice(slice: &[Self]) -> &[T] {
+        let ptr = slice.as_ptr() as *const T;
+
+        // Safety: OrderedFloat is #[repr(transparent)] and has no invalid values.
+        unsafe { std::slice::from_raw_parts(ptr, slice.len()) }
+    }
 }
 
 impl ToSql for OrderedFloat<f32> {
@@ -1111,5 +1118,17 @@ mod tests {
     fn test_ordered_float_estimate_size() {
         let ordered_float = OrderedFloat::<f64>::from(5_i64);
         assert_eq!(ordered_float.estimated_size(), 8);
+    }
+
+    #[test]
+    fn test_inner_slice() {
+        use crate::types::F32;
+
+        assert_eq!(std::mem::size_of::<F32>(), std::mem::size_of::<f32>());
+        assert_eq!(std::mem::align_of::<F32>(), std::mem::align_of::<f32>());
+
+        let slice = &[F32::from(1.0), 2.0.into(), 3.0.into()];
+        let inner = F32::inner_slice(slice);
+        assert_eq!(inner, &[1.0f32, 2.0, 3.0]);
     }
 }


src/frontend/src/expr/type_inference/cast.rs
```diff
@@ -184,6 +184,11 @@ pub fn cast_ok(source: &DataType, target: &DataType, allows: CastContext) -> boo
 /// Both `source` and `target` must be base types, i.e. not struct or array.
 pub fn cast_ok_base(source: &DataType, target: &DataType, allows: CastContext) -> bool {
     matches!(CAST_TABLE.get(&(source.into(), target.into())), Some(context) if *context <= allows)
+    // TODO(VECTOR_PLACEHOLDER): not in CAST_TABLE because
+    // * `DataType::try_from(DataTypeName::Vector).unwrap().to_oid()` panics
+    // * `DataTypeName::Vector.to_oid()` is better but `to_oid` does not work for `DataTypeName::List`
+    || matches!((source, target), (DataType::Varchar, DataType::Vector(_)) if CastContext::Explicit <= allows)
+    || matches!((source, target), (DataType::Vector(_), DataType::Varchar) if CastContext::Assign <= allows)
```
xiangjinwu, line 191: * #22031

src/common/src/util/value_encoding/column_aware_row_encoding.rs
```diff
@@ -174,7 +174,7 @@ mod new_serde {
             DataType::Struct(struct_def) => new_deserialize_struct(struct_def, data)?,
             DataType::List(item_type) => new_deserialize_list(item_type, data)?,
             DataType::Map(map_type) => new_deserialize_map(map_type, data)?,
-            DataType::Vector(_) => todo!("VECTOR_PLACEHOLDER"),
+            DataType::Vector(_) => plain::deserialize_value(ty, data)?,
```
xiangjinwu, line 177: @BugenZhao is this correct if `vector(n)` cannot be altered?

src/frontend/src/handler/create_table/col_id_gen.rs
```diff
@@ -197,7 +197,12 @@ impl ColumnIdGenerator {
                 Some((original_column_id, original_data_type)) => {
                     // Only check the type name (discriminant) for compatibility check here.
                     // For nested fields, we will check them recursively later.
-                    if original_data_type.type_name() != data_type.type_name() {
+                    let incompatible = original_data_type.type_name() != data_type.type_name()
+                        || matches!(
+                            (original_data_type, &data_type),
+                            (DataType::Vector(old), DataType::Vector(new)) if old != new,
+                        );
+                    if incompatible {
```
xiangjinwu, line 205: @BugenZhao This rejects alter for vector (as struct field rather than top-level column, which is part of `can_alter`).

btw I did not update the error message so it may look wired for vector, but being concise for the common case of struct/array/map:

|err msg style|case 1|case 2|
|-|-|-|
|type_name (current)|incompatible data type change from Vector to Vector at path "foo.bar"`|incompatible data type change from List to Map|
|type|incompatible data type change from vector(3) to vector(4) at path "foo.bar"`|incompatible data type change from struct<u struct<v int, w double>>[] to map(varchar, varchar)|

src/common/src/array/vector_array.rs
```diff
@@ -33,41 +36,57 @@ impl ArrayBuilder for VectorArrayBuilder {
...
+            elem_size: self.elem_size,
+        }
     }
 }
 
 #[derive(Debug, Clone)]
-pub struct VectorArray {}
+pub struct VectorArray {
+    inner: ListArray,
+    elem_size: usize,
```
BugenZhao, line 84: Can this actually be derived from `inner.flatten().len() / inner.len()`?

src/frontend/src/handler/create_table/col_id_gen.rs
BugenZhao: LGTM

src/common/src/array/vector_array.rs
xiangjinwu: No in the following cases:
* `inner.len() == 0`
* `inner` contains `null::vector(n)`

src/common/src/types/mod.rs
```diff
@@ -432,7 +432,9 @@ impl DataType {
                 // Same as List<Struct<K,V>>
                 pb.field_type = vec![datatype.clone().into_struct().to_protobuf()];
             }
-            DataType::Vector(_) => todo!("VECTOR_PLACEHOLDER"),
+            DataType::Vector(size) => {
+                pb.precision = *size as _;
```
wenym1, line 436: Should we fill in a fixed `pb.field_type = vec![DataType::Float32]`?

src/expr/impl/src/scalar/cast.rs
```diff
@@ -195,6 +198,14 @@ fn str_to_list(input: &str, ctx: &Context) -> Result<ListValue> {
     ListValue::from_str(input, &ctx.return_type).map_err(|err| ExprError::Parse(err.into()))
 }
 
+#[function("cast(varchar) -> vector", type_infer = "unreachable")]
```
wenym1, line 201: Will we implement `cast(real[]) -> vector`?

xiangjinwu: Yes as one of the many exprs `vector` support. Let me open a tracking issue for the `vector` data type after this.

src/common/src/types/mod.rs
xiangjinwu: It is unnecessary. `vector` is only parameterized by a number. The Float16 variant later will be named `halfvec` following pgvector.

PR #22021, opened 2025-05-27
feat: fd control for kafka source

tabVersion:

replace https://github.com/risingwavelabs/risingwave/pull/21734

expose the fd can be used in kafka related sdks in the developer config, default to `40,000`, thanks for the idea @hzxa21 

and count down the fd permit only in kafka source, which can report `not enough fd` error and tell the users to spread the actors across more compute nodes. 

* FD usage for Kafka Consumer (in Chinese) [ref](https://chatgpt.com/s/dr_68364431530c81919cf94dd7d79be4fa)

Kafka Sink now does not count into this, because of the sdk constraints, the producers cannot access metadata [ref](https://docs.rs/rdkafka/latest/rdkafka/producer/base_producer/struct.ThreadedProducer.html#method.send). Need to find another way in the future. 

* FD usage for Kafka Producer (in Chinese) [ref](https://chatgpt.com/s/dr_683643ee23548191894594f6f8ec9e2f)

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/common/src/config.rs
+++ b/src/common/src/config.rs
@@ -1157,6 +1157,10 @@ pub struct StreamingDeveloperConfig {
     #[serde(default = "default::developer::stream_chunk_size")]
     pub chunk_size: usize,
 
+    /// The maximum number of file descriptors that can be used by kafka source/sink.
+    #[serde(default = "default::developer::kafka_max_fd_count")]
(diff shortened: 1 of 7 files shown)


tabVersion:
close for @shanicky is working on shared kafka client on CN.

PR #22026, opened 2025-05-27
refactor(sqlsmith): replace SqlGeneratorContext with config

cardigan1008:

This PR removes the use of `SqlGeneratorContext` and replaces it with the configuration mechanism introduced in #21873 

The original `SqlGeneratorContext` exists solely to determine whether aggregation is allowed during SQL generation in SQLSmith. However, this control is conceptually similar to what the new configuration already supports — with the added benefit that configuration **provides more extensibility and centralized control over generation features**.

### Limitation

Currently, every call to `gen_expr()` requires an explicit `SqlGeneratorContext` to be passed, which becomes repetitive and inconvenient, especially since `gen_expr()` is a high-level and frequently used API. In practice, the only place where the context value truly affects behavior is inside `gen_agg()`. Therefore, we can move the decision logic directly into that method and retrieve the control flag from `self.config` instead of passing around a separate context object.

For example, the current implementation of `gen_having()` has to explicitly construct a new `context`:

```rust
fn gen_having(&mut self, have_group_by: bool) -> Option<Expr> {
    if have_group_by && self.flip_coin() {
        let context = SqlGeneratorContext::new();
        Some(self.gen_expr(&DataType::Boolean, context))
    } else {
        None
    }
}
```

After this PR, `gen_expr()` no longer requires a context argument, and control over aggregation generation becomes declarative and centralized.

This also paves the way for future extensions. For instance, in scenarios that previously relied on modifying the context, we can now use configuration toggles like:

```rust
self.config.set_enabled(Feature::Agg, false);
```

This simplification is particularly evident in cases like:

```rust
fn gen_bool_with_tables(&mut self, tables: Vec<Table>) -> Expr {
    let old_context = self.new_local_context();
    self.add_relations_to_context(tables);
    let expr = self.gen_expr(&Boolean, SqlGeneratorContext::new_with_can_agg(false));
    self.restore_context(old_context);
    expr
}
```

In short, this change removes unnecessary boilerplate, aligns aggregation control with our existing feature configuration system, and makes `gen_expr()` easier to use and extend.

### Plan

There're two fields in `SqlGeneratorContext`: `can_agg` and `inside_agg`. **`can_agg` can be easily mapped to `agg` in config**. We can replace the original random choice of `can_agg` by setting the weight to 50. 

However, `inside_agg` is more complex. For example, one of the only two places using it is in `gen_col`:

```rust
fn gen_col(&mut self, typ: &DataType, context: SqlGeneratorContext) -> Expr {
        let columns = if context.is_inside_agg() {
```

The reason for using it is that `gen_col` needs to decide if the used columns can come from only one table or not. If agg is used, multiple sources are not allowed, like `sum(t1.a + t2.b)`. 

So the key point here is that if we enable agg, `inside_agg` should immediately be set to true, which means **we can map `inside_agg` to `enabled` in config**. This can be done by using `set_enbaled` in config. `inside_agg` can be read by using `enable_generate`, which this PR will introduce.  

The other one is `gen_bool_with_tables` shown above. The original `new_with_can_agg()` set `inside_agg` to false. So we can just set enabled to true in `gen_bool_with_tables`. 

## Checklist

## Documentation

<details>

</details>


cardigan1008:
The original refactor will introduce one problem: when calling `new_with_can_agg()`, we will set `inside_agg` to false by making `enabled` to false. However, this also means the agg feature is completely disabled, which is decided by `enabled` first. So the refactor logic cannot cover the situation when we have `{ "can_agg": true, "inside_agg": false }`.

cardigan1008:
The context is reserved for future extension. The refactor mainly introduces the following changes:
- replace `can_agg` with config
- refactors construction and setter methods to be more explicit and intuitive:
  - `new()` now requires specifying key context values instead of relying on implicit defaults.
  - `set_inside_agg` is no longer a blind flag flip; it now accepts a boolean argument, making the context state more readable and intentional.

src/tests/sqlsmith/src/sql_gen/agg.rs
```diff
@@ -40,8 +40,8 @@ impl<R: Rng> SqlGenerator<'_, R> {
             return self.gen_simple_scalar(ret);
         }
 
-        let context = SqlGeneratorContext::new();
-        let context = context.set_inside_agg();
+        let mut context = SqlGeneratorContext::new(false);
+        context.set_inside_agg(true);
```
kwannoel:
```suggestion
        let context = SqlGeneratorContext::new(true);
```

src/tests/sqlsmith/src/sql_gen/expr.rs
```diff
@@ -90,7 +91,11 @@ impl<R: Rng> SqlGenerator<'_, R> {
         // - `a1 >= a2 IN b`
         // ...
         // We just nest compound expressions to avoid this.
-        let range = if context.can_gen_agg() { 100 } else { 50 };
+        let range = if context.can_gen_agg(self.should_generate(Feature::Agg)) {
```
kwannoel: Seems more readable to just flatten it.
```suggestion
        let range = if !context.is_inside_agg() && self.should_generate(Feature::Agg) {
```

kwannoel:
> Seems like `SqlGeneratorContext` is still present. The PR description needs to be updated, since we keep it for `inside_agg`.

Saw the reply here: https://github.com/risingwavelabs/risingwave/pull/22026#issuecomment-2934285773. LGTM

kwannoel (approved):
Rest LGTM

src/tests/sqlsmith/src/sql_gen/expr.rs
cardigan1008: Remove `can_gen_agg() ` from SqlGeneratorConext now. Thanks!

PR #22029, opened 2025-05-28
feat(expr): `l2_distance` between 2 `vector`s

xiangjinwu:

Adding an expression for the vector index poc. The implementation may be optimized further.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/expr/type_inference/func.rs
+++ b/src/frontend/src/expr/type_inference/func.rs
@@ -660,6 +660,40 @@ fn infer_type_for_special(
                 .into()),
             }
         }
+        ExprType::L2Distance => {
+            ensure_arity!("l2_distance", | inputs | == 2);
+            let (known_idx, unknown_idx) = match (inputs[0].is_untyped(), inputs[1].is_untyped()) {
+                (true, true) => return Err(ErrorCode::BindError(
+                    "function l2_distance(unknown, unknown) is not unique; you might need to add explicit type casts".to_owned()
+                ).into()),
+                (false, false) => match (inputs[0].return_type(), inputs[1].return_type()) {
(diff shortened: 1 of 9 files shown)


src/frontend/src/expr/type_inference/func.rs
```diff
@@ -660,6 +660,40 @@ fn infer_type_for_special(
...
         }
+        ExprType::L2Distance => {
+            ensure_arity!("l2_distance", | inputs | == 2);
+            let (known_idx, unknown_idx) = match (inputs[0].is_untyped(), inputs[1].is_untyped()) {
+                (true, true) => return Err(ErrorCode::BindError(
+                    "function l2_distance(unknown, unknown) is not unique; you might need to add explicit type casts".to_owned()
+                ).into()),
+                (false, false) => match (inputs[0].return_type(), inputs[1].return_type()) {
+                    (DataType::Vector(l), DataType::Vector(r)) => return match l == r {
+                        true => Ok(Some(DataType::Float64)),
```
wenym1, line 671: Given that vector item is `f32`, should the return type be also `f32`?

wenym1 (approved):
rest LGTM

src/frontend/src/expr/type_inference/func.rs
xiangjinwu: This follows pgvector, even the calculation detail of using f32 as intermediate result and only cast to f64 before sqrt
https://github.com/pgvector/pgvector/blob/v0.8.0/sql/vector.sql#L34
https://github.com/pgvector/pgvector/blob/v0.8.0/src/vector.c#L573

wenym1: Good to know that. I will change the type of VectorDistance to f64 also.

PR #22038, opened 2025-05-28
chore(doc): Fixed some introduction docs

Chiro11:

- Some outdated information I encountered while reading the documentation recently.
- Some additional notes.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/docs/dev/src/build-and-run/intro.md
+++ b/docs/dev/src/build-and-run/intro.md
@@ -20,6 +20,8 @@ sudo apt install make build-essential cmake protobuf-compiler curl postgresql-cl
 curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
 ```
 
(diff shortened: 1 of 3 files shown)


docs/dev/src/build-and-run/intro.md
```diff
@@ -20,6 +20,8 @@ sudo apt install make build-essential cmake protobuf-compiler curl postgresql-cl
 curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
 ```
 
+> If you encounter CMake error when compiling some third-party crates, it may be because the CMake version installed by package manager (ie. `brew/apt`) is too high and is incompatible with old CMake files in these crates. You can first uninstall, then manually download the required version from the [official source](https://cmake.org/download/).
```
wenym1, line 23: Can we provide a rough range of CMake version?

wenym1 (approved):
Rest LGTM. Thanks for the refinement!

PR #22039, opened 2025-05-28
fix(ci): fix integration test filter

kwannoel:

It currently uses `substring` matching from `cargo test`. As a result we may run unexpected tests.
For this prefix: `scale::`, [we end up having `sink::scale::*` run as well:](https://buildkite.com/risingwavelabs/main-cron/builds/5302#019714d4-2d1e-42c1-97e5-5a745a0e3748)
Screenshot 2025-05-28 at 2 59 49 PM

Since we are using `cargo nextest`, we can just use [`filtersets`](https://nexte.st/docs/filtersets/reference/), and only run the tests matching the prefix regex.

## Checklist

## Documentation

<details>

</details>


xxchan (approved):

ci/scripts/deterministic-it-test.sh
```diff
@@ -16,7 +16,17 @@ tar -xvf simulation-it-test.tar.zst
 mkdir target/sim
 mv target/ci-sim target/sim
 
-TEST_PATTERN="$@"
+if [ -z "$1" ]; then
+  TEST_PATTERN=""
+else
+  TEST_PATTERN="-E test(/^$1\/)"
```
kwannoel, line 22:
```suggestion
  TEST_PATTERN="-E 'test(/^$1\/)'"
```

ci/scripts/deterministic-it-test.sh
```diff
@@ -16,7 +16,17 @@ tar -xvf simulation-it-test.tar.zst
 mkdir target/sim
 mv target/ci-sim target/sim
 
-TEST_PATTERN="$@"
+if [ -z "$1" ]; then
+  TEST_PATTERN=""
+else
+  TEST_PATTERN="-E 'test(/^$1\/)'"
```
kwannoel, line 22:
```suggestion
  TEST_PATTERN="-E 'test(/^$1/)'"
```

PR #22040, opened 2025-05-28
feat(iceberg): support alter iceberg table connector with

xxhZs:

Previously, we implemented the `ALTER SINK CONNECTOR` syntax, but it lacked support for directly modifying Iceberg tables. Since Iceberg table definitions are not stored via Sink statements, attempts to alter them using Sink syntax would inevitably fail. ​This PR introduces `ALTER TABLE` support for Iceberg tables, enabling direct modifications to their associated Sink configurations.

Previously, we synchronized configurations by pulling the full config from the Meta service and dispatching it to the Sink. However, we observed that the Meta-stored configurations were incomplete. ​This PR refactors the mechanism to incremental updates: only modified configurations are pushed to Compute Nodes and dynamically merged with the Sink's existing configurations.

## Checklist

## Documentation

<details>

</details>


proto/meta.proto
```diff
@@ -345,7 +351,10 @@ message AlterConnectorPropsRequest {
   map<string, string> changed_props = 2;
   map<string, secret.SecretRef> changed_secret_refs = 3;
   optional uint32 connector_conn_ref = 4;
-  AlterConnectorPropsObject object_type = 5;
+  oneof object_type {
+    AlterConnectorPropsObject alter_connector_props_object = 6;
+    AlterIcebergTablePropsObject alter_iceberg_table_props_object = 7;
+  }
```
chenzl25, line 357: We'd better use a new RPC instead of reusing the altering sink RPC to reduce the coupling. Current implementation is a bit invasive.

proto/meta.proto
```diff
@@ -345,7 +350,10 @@ message AlterConnectorPropsRequest {
...
```
chenzl25: I think we should keep the original `object_type` as is and introduce an additional optional field to pass the iceberg sink and source ids.

src/connector/src/sink/iceberg/mod.rs
```diff
@@ -519,7 +519,7 @@ impl Sink for IcebergSink {
     }
 
     async fn new_log_sinker(&self, writer_param: SinkWriterParam) -> Result<Self::LogSinker> {
-        let table = self.create_and_validate_table().await?;
+        let table = self.create_and_validate_table().await.unwrap();
```
chenzl25: why using `unwrap()`?

src/meta/src/manager/metadata.rs
```diff
@@ -711,6 +712,24 @@ impl MetadataManager {
         let new_props = self
             .catalog_controller
             .update_sink_props_by_sink_id(sink_id, props)
+            .await
+            .unwrap();
```
chenzl25: dittio

chenzl25:
cc @xxhZs any updates?

src/frontend/src/handler/alter_table_props.rs
```diff
@@ -0,0 +1,114 @@
...
+        let schema_path = SchemaPath::new(schema_name.as_deref(), &search_path, user_name);
+        let (sink, _schema_name) =
+            reader.get_sink_by_name(db_name, schema_path, &real_sink_name, false)?;
+        let (source, _schema_name) =
+            reader.get_source_by_name(db_name, schema_path, &real_source_name)?;
+        let (table, schema_name) =
+            reader.get_table_by_name(db_name, schema_path, &real_table_name, false)?;
+        if sink.target_table.is_some() {
+            bail!("ALTER iceberg table config is not for sink into table")
+        }
```
chenzl25: Why do we need this check?

src/frontend/src/handler/alter_table_props.rs
```diff
@@ -0,0 +1,114 @@
...
+    let meta_client = session.env().meta_client();
+    let (resolved_with_options, _, connector_conn_ref) = resolve_connection_ref_and_secret_ref(
+        WithOptions::try_from(changed_props.as_ref() as &[SqlOption])?,
+        &session,
+        None,
+    )?;
+    let (changed_props, changed_secret_refs) = resolved_with_options.into_parts();
+    if !changed_secret_refs.is_empty() || connector_conn_ref.is_some() {
+        bail!("ALTER ICEBERG TABLE does not support SECRET or CONNECTION now")
+    }
```
chenzl25, line 98: Considering our e2e_test has used secret and connector, can we support it or is it hard to support? I think if we can add an e2e_test for alter table properties, it would be great. https://github.com/risingwavelabs/risingwave/blob/b234e00d218d8e12b3abf10b3855d7b72db719f8/e2e_test/iceberg/test_case/pure_slt/iceberg_engine.slt#L1-L24

chenzl25: Oh it means altering the secret and connection instead of alter a iceberg table using secret and connection.

PR #22043, opened 2025-05-28
feat(iceberg): enable iceberg compaction for iceberg table and add license check

chenzl25:

- Enable iceberg compaction for iceberg table and add license check
- Set default compaction interval secs to 3600s.

## Checklist

## Documentation

<details>

</details>


src/frontend/src/handler/create_table.rs
```diff
@@ -1714,6 +1714,14 @@ pub async fn create_iceberg_engine_table(
...
     sink_with.insert("is_exactly_once".to_owned(), "true".to_owned());
 
+    sink_with.insert(
+        "enable_compaction".to_owned(),
+        Feature::IcebergCompaction
+            .check_available()
+            .is_ok()
+            .to_string(),
+    );
+
```
Li0k, line 1724: add `compaction_interval_sec`, REST LGTM

Li0k (approved):

src/frontend/src/handler/create_table.rs
```diff
@@ -1714,6 +1714,14 @@ pub async fn create_iceberg_engine_table(
 
     sink_with.insert("is_exactly_once".to_owned(), "true".to_owned());
 
+    sink_with.insert(
```
Li0k, line 1717: check `with_options.enable_compaction`

src/frontend/src/handler/create_table.rs
```diff
@@ -1714,6 +1714,56 @@ pub async fn create_iceberg_engine_table(
 
     sink_with.insert("is_exactly_once".to_owned(), "true".to_owned());
 
+    if let Some(enable_compaction) = handler_args.with_options.get("enable_compaction") {
```
Li0k, line 1717: Here are two issues:  
1. This logic is implemented in a verbose way. We can encapsulate the logic for checking `enable_compaction` within `WithOptions` to reduce if branches.  
2. When `enable_compaction == None`, we don't need to check the license and should set `enable_compaction=false`.

Issue #21556, opened 2025-04-24
feat: allow different configurations for different database/resource group


lmatz:
### Is your feature request related to a problem? Please describe.

since barriers are decoupled among different databases/resources group,
is it possible to configure different barrier_interval for different databases/resource group?

or more generally, what configs (the ones in toml file) are allowed to be configured separately

wenym1:
Currently we have two system parameters `barrier_interval_ms` and `checkpoint_frequency` to configure the periodic barrier injection.

To support the proposed feature, we may need supporting the following SQL syntax:
* create database with specified `barrier_interval_ms` and `checkpoint_frequency`
```
CREATE DATABASE IF NOT EXISTS travel
    WITH 
        OWNER = travel_admin
        BARRIER_INTERVAL_MS = '250ms'
        CHECKPOINT_FREQUENCY = 5;
```
if not specified, use the configuration in system parameter.
* alter database with specified `barrier_interval_ms` and `checkpoint_frequency`
```
ALTER DATABASE database_name
    SET BARRIER_INTERVAL_MS = '250ms' ;
```
* For existing databases, use the configuration in system parameter.

Have assigned @Chiro11 to work on this feature.

PR #22052, opened 2025-05-29
feat(ddl): Implement database-level barrier and checkpoint configuration - Part 1

Chiro11:

- issue #21556 

## Checklist

## Documentation

<details>

</details>


src/frontend/src/handler/alter_barrier.rs
```diff
@@ -0,0 +1,152 @@
...
+
+fn check_privileges_for_alter_database(session: &SessionImpl, db_name: &str) -> Result<DatabaseId> {
+    let user_reader = session.env().user_info_reader().read_guard();
+    let catalog_reader = session.env().catalog_reader().read_guard();
+    let database = catalog_reader.get_database_by_name(db_name)?;
+
+    // The user should be super user or owner to alter the database.
+    session.check_privilege_for_drop_alter_db_schema(database)?;
+
+    // Non-superuser owners must also have the CREATEDB privilege.
```
wenym1, line 75: Any reason to specially forbidden it? The `check_privilege_for_drop_alter_db_schema` seems enough.

src/frontend/src/handler/alter_barrier.rs
```diff
@@ -0,0 +1,152 @@
...
+use risingwave_common::catalog::DatabaseParam;
+use risingwave_sqlparser::ast::ObjectName;
+
+use super::{HandlerArgs, RwPgResponse};
+use crate::Binder;
+use crate::catalog::DatabaseId;
+use crate::error::{ErrorCode, Result};
+use crate::session::SessionImpl;
+
+pub async fn handle_set_barrier_interval(
```
wenym1, line 25: Can merge the two `handle_set_xxx` to a single `handle_set_database_param`, and pass `AlterDatabaseParam`.

src/frontend/src/handler/create_database.rs
```diff
@@ -30,6 +30,8 @@ pub async fn handle_create_database(
     if_not_exist: bool,
     owner: Option<ObjectName>,
     resource_group: Option<SetVariableValue>,
+    barrier_interval_ms: Option<u32>,
```
wenym1, line 33: Shall we have a struct `DatabaseParam` with these two fields, so that we can pass the two fields together? Seems they always appear together in the path of create database.

src/sqlparser/src/ast/ddl.rs
```diff
@@ -32,6 +32,8 @@ use crate::tokenizer::Token;
 pub enum AlterDatabaseOperation {
     ChangeOwner { new_owner_name: Ident },
     RenameDatabase { database_name: ObjectName },
+    SetBarrierIntervalMs { barrier_interval_ms: Option<u32> },
```
wenym1, line 35: Can be merged into a `SetParam`.

src/sqlparser/src/keywords.rs
```diff
@@ -97,6 +97,7 @@ define_keywords!(
     AUTO,
     AVG,
     BACKFILL,
+    BARRIER_INTERVAL_MS,
```
wenym1, line 100: Seems unnecessary to preserve two keywords simply for database configuration. Can we possibly avoid this?

proto/ddl_service.proto
```diff
@@ -548,6 +548,20 @@ message AutoSchemaChangeRequest {
...
+message AlterDatabaseBarrierRequest {
+  enum BarrierParamType {
+    UNSPECIFIED = 0;
+    BARRIER_INTERVAL = 1;
+    CHECKPOINT_FREQUENCY = 2;
+  }
+
+  uint32 database_id = 1;
+  BarrierParamType param_type = 2;
+  optional uint64 value = 3;
```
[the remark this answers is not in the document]
Chiro11: `oneof` cannot wrap optional field.

src/sqlparser/src/keywords.rs
Chiro11: Do you mean just treating it as a literal string? I'm not sure, are there any other examples of configuration being treated in a similar way?

proto/ddl_service.proto
```diff
@@ -548,6 +548,20 @@ message AutoSchemaChangeRequest {
 
 message AutoSchemaChangeResponse {}
 
+message AlterDatabaseBarrierRequest {
```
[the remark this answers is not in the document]
wenym1: Alter resource group is not supported yet. For altering name and owner, they have different syntax, which does not follow the `SET xxx TO xxx` way, so we don't need to treat the altering name and owner as altering database param, but some special property. In the future when we support alter database resource group, we can reuse the same path.

src/frontend/src/handler/alter_barrier.rs
Chiro11: https://docs.risingwave.com/sql/commands/sql-alter-database 
I see that `CREATEDB` privilege is required for other operations.

proto/ddl_service.proto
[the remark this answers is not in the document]
Chiro11: OK, I'll modify those names.

src/sqlparser/src/keywords.rs
wenym1: All with-options of create sink are handled as literal string. 

In your current parser implementation, you can avoid depending on the `parse_one_of_keywords`, which accepts only keywords, but use the inner `peek_token` and resolve the `Token::Word(w)`.

proto/ddl_service.proto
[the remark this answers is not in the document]
wenym1: You can add support for `OptionalUint32` and `OptionalUint64` in `common.proto`, similar to the `Uint32Vector` there.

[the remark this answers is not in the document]
Chiro11: That seems good. Thank you.

src/frontend/src/handler/alter_barrier.rs
wenym1: I think the requirement on `CREATEDB` is special for rename and alter owner. The requirement is following the requirement from [alter database in Postgres](https://www.postgresql.org/docs/current/sql-alterdatabase.html). It says

> ALTER DATABASE changes the attributes of a database.
> 
> The first form changes certain per-database settings. (See below for details.) Only the database owner or a superuser can change these settings.
> 
> The second form changes the name of the database. Only the database owner or a superuser can rename a database; non-superuser owners must also have the CREATEDB privilege. The current database cannot be renamed. (Connect to a different database if you need to do that.)

Alter database param is the first form of altering, and therefore checking whether is database owner or superuser should be enough.

src/sqlparser/src/ast/ddl.rs
wenym1: Maybe we can try only define it as `SetParam { option: String, value: String}` in AST, and later resolve the `option` and `value` string in frontend.

src/sqlparser/src/parser.rs
```diff
@@ -2129,11 +2129,27 @@ impl Parser<'_> {
             }
         }
 
+        let barrier_interval_ms = if self.parse_word("BARRIER_INTERVAL_MS") {
```
wenym1: Does it mean users should write the options in the strict order as [OWNER|RESOURCE_GROUP] -> BARRIER_INTERVAL_MS -> CHECKPOINT_FREQUENCY? 

Doesn't look like a good design. Can we support specifying the config in arbitary order?

src/frontend/src/catalog/catalog_service.rs
```diff
@@ -70,6 +72,8 @@ pub trait CatalogWriter: Send + Sync {
         db_name: &str,
         owner: UserId,
         resource_group: &str,
+        barrier_interval_ms: Option<u32>,
```
wenym1, line 74: Any comment on https://github.com/risingwavelabs/risingwave/pull/22052#discussion_r2115282611?

src/meta/service/src/ddl_service.rs
```diff
@@ -1172,6 +1172,31 @@ impl DdlService for DdlServiceImpl {
...
     }
+
+    async fn alter_database_param(
+        &self,
+        request: Request<AlterDatabaseParamRequest>,
+    ) -> Result<Response<AlterDatabaseParamResponse>, Status> {
+        let req = request.into_inner();
+        let database_id = req.database_id;
+
+        if let Some(config_param) = req.param {
```
wenym1: We can assert that the `req.param` must be `Some`. Usually we do `req.param.unwrap()` here.

src/meta/service/src/ddl_service.rs
```diff
@@ -1172,6 +1172,31 @@ impl DdlService for DdlServiceImpl {
...
+            let param = match config_param {
+                alter_database_param_request::Param::BarrierIntervalMs(value) => {
+                    AlterDatabaseParam::BarrierIntervalMs(value.value)
+                }
+                alter_database_param_request::Param::CheckpointFrequency(value) => {
+                    AlterDatabaseParam::CheckpointFrequency(value.value)
+                }
+            };
+            self.ddl_controller
```
wenym1: We may follow other ddl, to return a notification version for frontend to wait for.

wenym1 (commented):
We shall also support seeing the two new params in the `rw_catalog.rw_databases` system table. You can search for `rw_catalog.rw_databases` in the codebase to find where to modify.

src/meta/service/src/ddl_service.rs
Chiro11: Why should we wait? This will change later in meta, and frontend does not need to be aware of it immediately. Besides, I refer to the implementation of `AlterResourceGroup`.

src/frontend/src/catalog/catalog_service.rs
Chiro11: I think it's OK. Anyway, there are relatively few parameters now, and it's quite clear to list them. If there are more parameters in the future, we can consider combining them together.

src/meta/service/src/ddl_service.rs
wenym1: If we want to query the latest param value from `rw_catalog.rw_databases`, we should wait for the version. As for alter resource group, if we expect to be able to query the latest resource group on frontend, the current implementation is not correct.

Chiro11:
@wenym1 The changes have been squashed into the latest commit. Please Review. https://github.com/risingwavelabs/risingwave/pull/22052/commits/37a93b810b96438c8866017aed740fddc0b6cc7c

e2e_test/ddl/database.slt
```diff
@@ -76,6 +76,30 @@ default
 statement ok
 drop database db_with_keyword_default_resource_group;
 
+statement ok
+create database db_barrier_custom with barrier_interval_ms = 1500 checkpoint_frequency = 2;
```
wenym1, line 80: Can add test on querying the `rw_databases` system table.

wenym1 (approved):
Rest LGTM. Thanks for the PR! 

@yezizp2012 Can you review the meta store and migration part?

e2e_test/ddl/database.slt
Chiro11: Added.

yezizp2012 (approved):
The meta part LGTM.

PR #22054, opened 2025-05-29
fix(common): ignore field ids when comparing `struct` for backward compatibility

### LINKED ISSUE #22041: bug: `sink into table` always fail on `struct` column equality check
### Describe the bug

For `sink into table` we require the upstream streaming job to have same data type with the downstream table. However the equality on `struct` data type always fail due to internal invisible differences - even when the field names and data types match.

`lhs.equals_datatype(rhs)` shall be used instead of `lhs == rhs`

https://github.com/risingwavelabs/risingwave/blob/ad53aac274f297f4b6cb72b116641b2d7c9ed8ff/src/frontend/src/handler/create_sink.rs#L660-L669

### Error message/log

```text

Caused by:
  column type mismatch: Struct(StructType { fields: [("foo", Int32)] }) vs Struct(StructType { fields: [("foo", Int32)] }), column name: "$expr10001"
```

### To Reproduce

```sql
create table t (f1 struct<foo int>);
create sink sk into t as select row(3)::struct<foo int> as f1;
```

### Expected behavior

The sink shall be created successfully.

Note that as sink-into-table is working within SQL rather than to an external system, it matches fields by position rather than by name. For example, `struct<a int, b real>` does not match `struct<b real, a int>`.

_No response_

### The version of RisingWave

2.4.0

### Additional context

Tables or sinks created before upgrading to 2.4.0 are not affected. They do not have field ids assigned and the struct cannot be altered.

xiangjinwu:

Fixes #22041.

## Checklist

## Documentation

<details>

</details>


src/common/src/types/struct_type.rs
```diff
@@ -45,7 +45,7 @@ impl Debug for StructType {
     }
 }
 
-#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[derive(Clone, Debug, PartialOrd, Ord)]
```
BugenZhao, line 48: We may also override the derived impl of `Ord`. The behavior should be consistent with `Eq`.

src/common/src/types/struct_type.rs
```diff
@@ -63,6 +63,18 @@ struct StructTypeInner {
...
+    fn eq(&self, other: &Self) -> bool {
+        self.fields == other.fields
+    }
+}
+impl Eq for StructTypeInner {}
+impl std::hash::Hash for StructTypeInner {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        self.fields.hash(state);
+    }
+}
```
BugenZhao: Shall we use `#[educe(PartialEq(ignore), Hash(ignore))]`, and add some explicit comments?

Issue #19707, opened 2024-12-07
MQTT source with wildcard topic may lose data.


KeXiangWang:
### Describe the bug

As titled.
Found it while investigating https://github.com/risingwavelabs/risingwave/issues/19641.

### Error message/log

```text
No erro message.
```

### To Reproduce

```
set sink_decouple = false;

CREATE TABLE
  personnel (id integer, name varchar);

CREATE TABLE mqtt_source_table
(
  id integer,
  name varchar
)
WITH (
    connector='mqtt',
    url='tcp://localhost:1883',
    topic= 'tele/+/SENSOR',
    qos = 'at_least_once',
    max_packet_size =  200000
) FORMAT PLAIN ENCODE JSON;

CREATE SINK mqtt_sink_1
AS
SELECT 
  id,
  name
FROM
  personnel
WITH
(
    connector='mqtt',
    url='tcp://localhost:1883',
    topic= 'tele/d1/SENSOR',
    type = 'append-only',
    retain = 'true',
    qos = 'at_least_once',
) FORMAT PLAIN ENCODE JSON (
    force_append_only='true',
);

CREATE SINK mqtt_sink_2
AS 
SELECT 
  id + 10 as id,
  name
FROM
  personnel
WITH
(
    connector='mqtt',
    url='tcp://localhost:1883',
    topic= 'tele/d2/SENSOR',
    type = 'append-only',
    retain = 'true',
    qos = 'at_least_once',
) FORMAT PLAIN ENCODE JSON (
    force_append_only='true',
);


INSERT INTO
  personnel
VALUES
  (1, 'Alice'),
  (2, 'Bob'),
  (3, 'Tom'),
  (4, 'Jerry'),
  (5, 'Araminta'),
  (6, 'Clover'),
  (7, 'Posey'),
  (8, 'Waverly');

FLUSH;
```

### Expected behavior

You should be able to see 16 rows from `select * from mqtt_source_table`. 
However, you can only observe only 2 rows.

### How did you deploy RisingWave?

Anyway

KeXiangWang:
### Root Cause:

Typically, our source operates as follows:

Meta: Periodically lists splits and assigns them to actors, enabling dynamic discovery of splits.
CN: Scans data from the assigned splits.
However, MQTT is a relatively primitive protocol:

1. It lacks a list API.
[...]

PR #22074, opened 2025-06-02
feat(mqtt): fix MQTT data loss

Boudewijn26:

This PR does 2 things, separated in commits (it can be helpful to review them separatedly).

1. It fixes the MQTT data loss (https://github.com/risingwavelabs/risingwave/issues/19707) by moving the processing to a single split (and client)
2. It allows multiple MQTT topics to be specified by separating with `,`s

I intentionally chose to keep 1 split, even when using multiple MQTT topics to prevent message duplication in the case of overlapping topics. e.g. subscribing to `/a/#` and `/a/b/#` should only result in 1 message on topic `a/b/c`. Splitting multiple topics could result in message duplication, if one client subscribed to `/a/#` and the other subscribed to `/a/b/#`

## Checklist

## Documentation

<details>

MQTT sources now allow for multiple topics by separating with `,`s

</details>

### DIFF
--- a/src/connector/src/source/mqtt/enumerator/mod.rs
+++ b/src/connector/src/source/mqtt/enumerator/mod.rs
@@ -12,16 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::collections::HashSet;
-use std::sync::Arc;
-use std::sync::atomic::AtomicBool;
-
 use async_trait::async_trait;
-use risingwave_common::bail;
-use rumqttc::Outgoing;
-use rumqttc::v5::{ConnectionError, Event, Incoming};
-use thiserror_ext::AsReport;
-use tokio::sync::RwLock;
 
 use super::MqttProperties;
 use super::source::MqttSplit;
@@ -29,13 +20,7 @@ use crate::error::ConnectorResult;
 use crate::source::{SourceEnumeratorContextRef, SplitEnumerator};
 
 pub struct MqttSplitEnumerator {
-    #[expect(dead_code)]
     topic: String,
-    #[expect(dead_code)]
-    client: rumqttc::v5::AsyncClient,
-    topics: Arc<RwLock<HashSet<String>>>,
-    connected: Arc<AtomicBool>,
-    stopped: Arc<AtomicBool>,
 }
 
 #[async_trait]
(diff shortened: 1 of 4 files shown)


src/connector/src/source/mqtt/enumerator/mod.rs
```diff
@@ -45,95 +30,56 @@ impl SplitEnumerator for MqttSplitEnumerator {
...
-            client,
-            topics,
             topic: properties.topic,
-            connected,
-            stopped,
         })
     }
 
     async fn list_splits(&mut self) -> ConnectorResult<Vec<MqttSplit>> {
-        if !self.connected.load(std::sync::atomic::Ordering::Relaxed) {
```
tabVersion: we also utilize list_splits func as connectivity test
if setting to a fixed value, the create source statement always succeeds, whether we can connect to the broker or not.

tabVersion (commented):
sorry for late review, I still have some concern about removing the enumerator.

src/connector/src/source/mqtt/enumerator/mod.rs
```diff
@@ -48,44 +43,20 @@ impl SplitEnumerator for MqttSplitEnumerator {
...
 
         let stopped = Arc::new(AtomicBool::new(false));
         let stopped_clone = stopped.clone();
 
-        let topics_clone = topics.clone();
         tokio::spawn(async move {
             while !stopped_clone.load(std::sync::atomic::Ordering::Relaxed) {
                 match eventloop.poll().await {
-                    Ok(Event::Outgoing(Outgoing::Subscribe(_))) => {
+                    Ok(Event::Incoming(Incoming::ConnAck(_))) => {
```
Boudewijn26, line 57: I'm just testing the connection. I could test the subscribe, but then I would also get the messages. Also the current implementation tests the subscribe going out, not the subscribe acknowledge coming in, so this is comparable

Boudewijn26:
Sorry for the delay, I was on holiday for 2.5 weeks

src/connector/src/source/mqtt/source/reader.rs
```diff
@@ -64,8 +64,8 @@ impl SplitReader for MqttSplitReader {
             .subscribe_many(
                 splits
                     .iter()
-                    .cloned()
-                    .map(|split| Filter::new(split.topic, qos)),
+                    .flat_map(|split| split.topic.split(","))
```
thexiay, line 67: Is it possible to split the topic on the CN side without having to split it? Because it is a static topic string, it is directly divided on the meta side, allowing multiple cns to consume different topics and improve throughput

src/connector/src/source/mqtt/enumerator/mod.rs
```diff
@@ -48,44 +43,20 @@ impl SplitEnumerator for MqttSplitEnumerator {
         context: SourceEnumeratorContextRef,
     ) -> ConnectorResult<MqttSplitEnumerator> {
         let (client, mut eventloop) = properties.common.build_client(context.info.source_id, 0)?;
-
         let topic = properties.topic.clone();
```
thexiay, line 46: If wildcards are not supported, whether to add a check to `MqttSplitEnumerator` to display the user should not use a topic with wildcards

thexiay: Synchronously modify comments on `MqttProperties#topic`

src/connector/src/source/mqtt/source/reader.rs
Boudewijn26: Possible yes, though I intentionally chose not to: (from the PR description)
> I intentionally chose to keep 1 split, even when using multiple MQTT topics to prevent message duplication in the case of overlapping topics. e.g. subscribing to /a/# and /a/b/# should only result in 1 message on topic a/b/c. Splitting multiple topics could result in message duplication, if one client subscribed to /a/# and the other subscribed to /a/b/#

thexiay:
LGTM

thexiay (approved):

florissmit10:
Is this ready to be merged? 
It would be helpful to us to have these changes in the nightly build. Thanks!

tabVersion (approved):
thanks for contribution!

PR #22075, opened 2025-06-02
feat(webhook): add support for webhook batching

### LINKED ISSUE #22066: Support batching in webhook

Currently, `webhook` source ingest data per row. We have a use case where we generate events on client side, queue it in memory, and after certain thresholds, flush it downstream in a single batched http call. We do this to minimize roundtrip, as each flush can contain up to hundreds of small events.

The easiest way I can think of to support this is by using [JSONL](https://jsonlines.org/), where each row is separated by a newline. One way I can think of is to introduce is as a separate connector called `webhook_batched`:

```sql
CREATE TABLE wbhtable (
  data JSONB
) WITH (
  connector = 'webhook_batched'
) VALIDATE SECRET test_secret AS secure_compare(
  headers->>'x-signature',
  encode(hmac(test_secret, data, 'sha1'), 'hex')
);
```

The alternative is to modify `webhook` to support multiple rows. But using JSONL will break backwards compatibility, as it's natural for webhook JSON payload to contain newline.

### Additional context

If this idea makes sense, I am willing to contribute.

hasyimibhar:

This PR adds support for sending multiple rows in a single batched request to webhook table. This is enabled via `is_batched` config. If set to true, the request body is expected to be valid [JSON Lines](https://jsonlines.org/).

```sql
CREATE TABLE wbhtable (
  data JSONB
) WITH (
  connector = 'webhook',
  is_batched = true,
) VALIDATE AS secure_compare(
  headers->>'x-signature',
  encode(hmac('secret', data, 'sha1'), 'hex')
);
```

Send request:

```sh
curl -X POST \
-H "Content-Type: application/jsonl" \
-H "x-signature: ab1ea42f2e0ea2459f5b26fd30ce42bb9b425bf3" \
--data-binary $'{"foo":"bar"}\n{"foo":"baz"}' \
0.0.0.0:4560/webhook/dev/public/wbhtable
```

2 rows inserted:

```
dev=> select * from wbhtable;
      data
----------------
 {"foo": "baz"}
 {"foo": "bar"}
(2 rows)
```

This PR resolves #22066.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/ci/workflows/pull-request.yml
+++ b/ci/workflows/pull-request.yml
@@ -137,7 +137,7 @@ steps:
     plugins:
       - docker-compose#v5.5.0: *docker-compose
       - ./ci/plugins/upload-failure-logs
-    timeout_in_minutes: 9
+    timeout_in_minutes: 12
     retry: *auto-retry
 
   - label: "meta backup test"
--- a/e2e_test/sink/sink_into_table/issue_22041.slt
+++ b/e2e_test/sink/sink_into_table/issue_22041.slt
@@ -1,0 +1,16 @@
+statement ok
+create table t (f1 struct<foo int>);
+
+statement ok
+create sink sk into t as select row(3)::struct<foo int> as f1;
+
+query
+select (f1).foo from t;
+----
+3
+
+statement ok
+drop sink sk;
+
+statement ok
+drop table t;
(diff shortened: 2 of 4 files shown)


hasyimibhar:
@lmatz Any idea why the CI is failing? It was passing before I rebased to main.

lmatz:
SCR-20250607-lhn
I think it is unrelated to this PR.
Let me increase the timeout in a separate PR and then rebase this PR on main

PR #22085, opened 2025-06-03
fix(meta, frontend): convert `real_value` back to `ast::Ident` properly

BugenZhao:

Successor of #19032.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/sqlparser/tests/sqlparser_postgres.rs
+++ b/src/sqlparser/tests/sqlparser_postgres.rs
@@ -471,7 +471,7 @@ fn parse_show() {
     assert_eq!(
         stmt,
         Statement::ShowVariable {
-            variable: vec!["all".into(), "all".into()]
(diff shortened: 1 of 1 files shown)


src/sqlparser/src/ast/mod.rs
```diff
@@ -121,6 +121,7 @@ pub struct Ident {
 impl Ident {
     /// Create a new identifier with the given value and no quotes.
     /// the given value must not be a empty string.
+    // FIXME: should avoid using this function unless it's a literal or for testing.
     pub fn new_unchecked<S>(value: S) -> Self
```
xiangjinwu, line 125: Is it possible to `pub(crate)` this?

xiangjinwu (approved):
Thank you!

src/sqlparser/src/ast/mod.rs
BugenZhao: Unfortunately it's also used by integration tests.

PR #22090, opened 2025-06-03
feat(batch): collect stats for each query

zwang28:

This PR adds a general approach to collect statistics from executors and aggregate into one during a batch query.
The collected stats is output as a new field in the existing log line when completing the query, e.g.
`2025-06-04T10:55:49.396985+08:00  INFO handle_query{mode="simple query" session_id=0 sql=SELECT * FROM t stats=row_scan_count=2}: pgwire_query_log: status="ok" time=16ms`

Currenly the only stats is `row_scan_count` of RowSeqScanExecutor, which aims to help identify queries that scan excessive rows and cause CPU saturation.

Implementation details:
- For query of distributed mode, the query stats are collected via Executor(serving node) -> StageRunner(frontend node) -> QueryExecution(frontend node).
- For query of local mode, the query stats are collected via Executor(serving node) -> GenericExchangeExecutor/MergeSortExchangeExecutor(frontend node) -> LocalQueryExecution(frontend node) 

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/compute/tests/cdc_tests.rs
+++ b/src/compute/tests/cdc_tests.rs
@@ -381,6 +381,7 @@ async fn test_cdc_backfill() -> StreamResult<()> {
         None,
         None,
         None,
+        None,
     ));
 
     // check result
--- a/src/compute/tests/integration_tests.rs
+++ b/src/compute/tests/integration_tests.rs
@@ -275,6 +275,7 @@ async fn test_table_materialize() -> StreamResult<()> {
         None,
         None,
         None,
+        None,
     ));
     let mut stream = scan.execute();
     let result = stream.next().await;
@@ -343,6 +344,7 @@ async fn test_table_materialize() -> StreamResult<()> {
         test_batch_query_epoch(),
         1024,
         "RowSeqScanExecutor2".to_owned(),
+        None,
         None,
         None,
         None,
@@ -426,6 +428,7 @@ async fn test_table_materialize() -> StreamResult<()> {
         None,
         None,
         None,
+        None,
     ));
 
     let mut stream = scan.execute();
@@ -502,6 +505,7 @@ async fn test_row_seq_scan() -> StreamResult<()> {
         test_batch_query_epoch(),
         1,
         "RowSeqScanExecutor2".to_owned(),
+        None,
         None,
         None,
         None,


zwang28:
@chenzl25 Could you please help review this PR? Thank you.

chenzl25:
> @chenzl25 Could you please help review this PR? Thank you.

Sure. Will review it today.

src/utils/pgwire/src/pg_protocol.rs
```diff
@@ -309,6 +309,7 @@ where
             mode,
             session_id,
             sql = tracing::field::Empty, // record SQL later in each `process` call
+            stats = tracing::field::Empty, // record stats for the query
```
chenzl25, line 312: Shall we put it in the `PGWIRE_QUERY_LOG` target instead of `PGWIRE_ROOT_SPAN_TARGET`?

chenzl25 (approved):
Rest LGTM!

proto/task_service.proto
```diff
@@ -34,6 +38,7 @@ message TaskInfoResponse {
   TaskStatus task_status = 2;
   // Optional error message for failed task.
   string error_message = 3;
+  optional TaskStats task_stats = 4;
```
BugenZhao, line 41: I think a message field should already be optional.

src/batch/executors/src/executor/row_seq_scan.rs
```diff
@@ -236,9 +244,15 @@ impl<S: StateStore> RowSeqScanExecutor<S> {
 
         // the number of rows have been returned as execute result
         let mut returned = 0;
+        fn collect_table_stats(task_stats: Option<TaskStatsRef>, returned: u64) {
+            if let Some(ref t) = task_stats {
+                t.row_scan_count.fetch_add(returned, Ordering::Relaxed);
+            }
+        }
```
BugenZhao, line 251: I don't think this is a robust way. 🥵 Can we try leveraging some sort of RAII to achieve this?

src/frontend/src/scheduler/distributed/query.rs
```diff
@@ -189,7 +190,15 @@ impl QueryExecution {
...
                 // Not trace the error here, it will be processed in scheduler.
-                tokio::spawn(async move { runner.run().instrument(span).await });
+                tokio::spawn(
+                    async move {
+                        let query_stats = runner.run().instrument(span).await;
+                        // Record the stats field in the handle_query span.
+                        tracing::Span::current()
+                            .record("stats", tracing::field::display(&query_stats));
+                    }
+                    .instrument(tracing::Span::current()),
```
BugenZhao, line 200: Suggest passing in the span object explicitly.

zwang28:
> I think this PR may need a more detailed design, as it aims to be a general framework for per-query statistics. 

I'll reference the implementation of EXPLAIN ANALYZE for streaming.

PR #22097, opened 2025-06-04
fix: redact sql option in diagnosis report according to keywords

zwang28:

Previously frontend node will redact SQL options according on configurable keywords, while meta node's diagnosis report will always redact all SQL options.
In this PR we make the diagnosis report redaction use the same configurable keywords as well.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/meta/src/manager/diagnose.rs
+++ b/src/meta/src/manager/diagnose.rs
@@ -28,7 +28,7 @@ use risingwave_pb::common::WorkerType;
 use risingwave_pb::meta::EventLog;
 use risingwave_pb::meta::event_log::Event;
 use risingwave_pb::monitor_service::stack_trace_request::ActorTracesFormat;
-use risingwave_sqlparser::ast::{CompatibleFormatEncode, Statement, Value};
+use risingwave_sqlparser::ast::RedactSqlOptionKeywordsRef;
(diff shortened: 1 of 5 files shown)


src/meta/src/manager/diagnose.rs
```diff
@@ -787,51 +790,13 @@ fn merge_prometheus_selector<'a>(selectors: impl IntoIterator<Item = &'a str>) -
...
-        if let Some(options) = options.0 {
-            for option in options {
-                option.value = Value::SingleQuotedString("[REDACTED]".into()).into();
-            }
-        }
-        if let Some(options) = options.1 {
-            for option in options {
-                option.value = Value::SingleQuotedString("[REDACTED]".into()).into();
-            }
-        }
```
xiangjinwu, line 821: Oh so this caller was setting `'[REDACTED]'` itself and not calling `to_redacted_string` (to be fixed by #22096).

xiangjinwu (commented):
Could you cherry-pick 389d677cd6857021b3ad342bdefa0df3549fd501 and rearrange it **BEFORE** your actual fix commit? So that we can be confident that:
* That test correctly reflect a panic we are trying to fix (not a weak test that always passes)
* The fix in this PR changed the test result from failing to passing

PR #22108, opened 2025-06-04
fix(sink): fix starrrocks https host error

xxhZs:

This pull request introduces two new utility methods to handle host formatting in the `StarrocksConfig` struct and updates the `StarrocksSink` and `StarrocksSinkWriter` implementations to use these methods. These changes improve code readability and ensure consistent handling of host strings with or without URL schemes.

## Checklist

## Documentation

<details>

</details>


src/connector/src/sink/starrocks.rs
```diff
@@ -146,6 +146,26 @@ impl StarrocksConfig {
         }
         Ok(config)
     }
+
+    pub fn get_host_with_scheme(&self) -> String {
+        let host = self.common.host.trim();
+        if host.starts_with("http://") || host.starts_with("https://") {
+            host.to_string()
+        } else {
+            format!("http://{}", host)
```
hzxa21: I feel that we should be clear in the error message and documentation to let user fill in the correct http scheme instead of us implicitly using `http` if no schema is provided.

We can use the [url](https://docs.rs/url/latest/url/) crate for url parsing.

xxhZs: Here it's mainly for compatibility considerations. The previous logic was to accept the host with http head, so we added the HTTP head.

hzxa21: I get it now. The user specified `host` is not only used by the http protocol but also used by the mysql protocol. In this case, I think letting user to specify the http scheme is confusing because we need to manually stripped out the http schema and feed the url to mysql protocol.

How about just adding a with option `use_https` (default to `false`)?

hzxa21 (approved):

PR #22113, opened 2025-06-04
fix: improve error handling for AWS MSK IAM token generation

tabVersion:

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/src/source/kafka/client_context.rs
+++ b/src/connector/src/source/kafka/client_context.rs
@@ -23,6 +23,7 @@ use rdkafka::consumer::ConsumerContext;
 use rdkafka::message::DeliveryResult;
 use rdkafka::producer::ProducerContext;
 use rdkafka::{ClientContext, Statistics};
+use thiserror_ext::AsReport;
 use tokio::runtime::Runtime;
 
 use super::private_link::{BrokerAddrRewriter, PrivateLinkContextRole};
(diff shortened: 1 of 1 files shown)


src/connector/src/source/kafka/client_context.rs
```diff
@@ -144,9 +145,11 @@ impl KafkaContextCommon {
...
                         .await
                     })
-                })
-                .unwrap()
-                .unwrap()
+                });
+                let timeout_result =
+                    result.map_err(|_e| "generating AWS MSK IAM token timeout".to_owned())?;
+                timeout_result
+                    .map_err(|e| format!("generating AWS MSK IAM token error: {}", e.as_report()))?
```
kwannoel:
```suggestion
                    .map_err(|e| anyhow!(e)).context("failed to generate AWS MSK IAM token")?
```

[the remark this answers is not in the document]
kwannoel: Should use this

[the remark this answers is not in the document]
kwannoel: https://github.com/risingwavelabs/risingwave/pull/22113/files#r2127825290

src/connector/src/source/kafka/client_context.rs
```diff
@@ -144,9 +144,11 @@ impl KafkaContextCommon {
                         )
                         .await
                     })
-                })
-                .unwrap()
-                .unwrap()
+                });
+                result
+                    .map_err(|_e| "generating AWS MSK IAM token timeout".to_owned())?
```
kwannoel, line 149: We should retain the original error if it provides time elapsed. It will be helpful for debugging.

kwannoel (approved):

kwannoel:
Thanks!

src/connector/src/source/kafka/client_context.rs
tabVersion: The error message here just tells the request timeout (how much time elapsed since beginning), so should be safe to ignore.

PR #22128, opened 2025-06-05
fix(mysql-cdc): use connection pool to avoid data loss caused by unsafe cancellation

wcy-fdu:

In the implementation of MySQL CDC, we use mysql_async to read historical data from MySQL. We utilize the exec_drop interface, but there is a cancellation safety issue: if a drop occurs while exec_drop is running, and if the connection is not rebuilt, subsequent calls to read_stream will return empty. At this point, the upper layer will mistakenly believe that the backfill has ended, leading to data loss.

One way to simulate and reproduce this error is to construct a case where exec_drop is dropped:
```
 let query  = conn.exec_drop("SET time_zone = if(not sleep(2), \"+00:00\", \"\")", ());

        match tokio::time::timeout(Duration::from_secs(1), query).await {
            Ok(result) => {
                println!("Operation completed: {:?}", result);
            },
            Err(_) => {
                println!("Operation timed out");
            }
        };
```
In the above example, after exec_drop times out and is dropped, using this connection afterwards will result in an error: `Io(Io(Custom { kind: UnexpectedEof, error: "can't parse: buf doesn't have enough data" }))`

After reviewing the official [documentation](https://crates.io/crates/mysql_async) for mysql_async, I believe we should use the recommended connection pool. When we need a connection, we can use `get_conn` and then drop the connection with `drop(conn)` once we're done. Additionally, after backfilling and reading all historical data, we should first run `pool.disconnect` before dropping snapshot_stream. This approach can avoid cancellation safety issues and elegantly reclaim resources.

Note that, it's a bit hard to construct a scenario where a drop occurs exactly during the execution of `exec_drop`, so we constructed a unit test using minimal reproducible code and ran it using the MySQL environment in `e2e-source-test`.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/src/source/cdc/external/mod.rs
+++ b/src/connector/src/source/cdc/external/mod.rs
@@ -197,12 +197,12 @@ pub struct DebeziumSourceOffset {
 
 pub type CdcOffsetParseFunc = Box<dyn Fn(&str) -> ConnectorResult<CdcOffset> + Send>;
 
-pub trait ExternalTableReader {
+pub trait ExternalTableReader: Sized {
     async fn current_cdc_offset(&self) -> ConnectorResult<CdcOffset>;
 
     // Currently, MySQL cdc uses a connection pool to manage connections to MySQL, and other CDC processes do not require the disconnect step for now.
     #[allow(clippy::unused_async)]
-    async fn disconnect(&self) -> ConnectorResult<()> {
+    async fn disconnect(self) -> ConnectorResult<()> {
         Ok(())
     }
 
--- a/src/connector/src/source/cdc/external/mysql.rs
+++ b/src/connector/src/source/cdc/external/mysql.rs
@@ -375,8 +375,8 @@ impl ExternalTableReader for MySqlExternalTableReader {
         self.snapshot_read_inner(table_name, start_pk, primary_keys, limit)
     }
 
-    async fn disconnect(&self) -> ConnectorResult<()> {
-        self.pool.clone().disconnect().await.map_err(|e| e.into())
+    async fn disconnect(self) -> ConnectorResult<()> {
+        self.pool.disconnect().await.map_err(|e| e.into())
     }
 }
 
--- a/src/stream/src/executor/backfill/cdc/upstream_table/snapshot.rs
+++ b/src/stream/src/executor/backfill/cdc/upstream_table/snapshot.rs
@@ -44,7 +44,7 @@ pub trait UpstreamTableRead {
         &self,
     ) -> impl Future<Output = StreamExecutorResult<Option<CdcOffset>>> + Send + '_;
 
-    async fn disconnect(&self) -> StreamExecutorResult<()>;
+    async fn disconnect(self) -> StreamExecutorResult<()>;
 }
 
 #[derive(Debug, Clone)]
@@ -244,7 +244,7 @@ impl UpstreamTableRead for UpstreamTableReader<ExternalStorageTable> {
         Ok(Some(binlog))
     }
 
-    async fn disconnect(&self) -> StreamExecutorResult<()> {
+    async fn disconnect(self) -> StreamExecutorResult<()> {
         self.reader.disconnect().await?;
         Ok(())
     }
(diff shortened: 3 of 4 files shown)


wcy-fdu:
CI failed, under investigation

wcy-fdu:
cc @hzxa21 for reviewing, @yuhao-su I heard you found another potential data loss situation. This PR is caused by `mysql_async`, just to let you know.

src/connector/src/source/cdc/external/mysql.rs
```diff
@@ -375,10 +374,15 @@ impl ExternalTableReader for MySqlExternalTableReader {
     ) -> BoxStream<'_, ConnectorResult<OwnedRow>> {
         self.snapshot_read_inner(table_name, start_pk, primary_keys, limit)
     }
+
+    async fn disconnect(&self) -> ConnectorResult<()> {
```
hzxa21: You can pass ownership of self here so that you don't need to clone the pool: `async fn disconnect(self)`

src/connector/src/source/cdc/external/mysql.rs
```diff
@@ -375,10 +374,15 @@ impl ExternalTableReader for MySqlExternalTableReader {
...
+        self.pool.clone().disconnect().await?;
+        Ok(())
```
hzxa21: I think you can just do
```
self.pool.clone().disconnect().await?
```

src/stream/src/executor/backfill/cdc/upstream_table/snapshot.rs
```diff
@@ -241,6 +243,11 @@ impl UpstreamTableRead for UpstreamTableReader<ExternalStorageTable> {
         let binlog = binlog.await?;
         Ok(Some(binlog))
     }
+
+    async fn disconnect(&self) -> StreamExecutorResult<()> {
+        self.reader.disconnect().await?;
+        Ok(())
+    }
```
hzxa21: ditto

src/connector/src/source/cdc/external/mod.rs
```diff
@@ -200,6 +200,12 @@ pub type CdcOffsetParseFunc = Box<dyn Fn(&str) -> ConnectorResult<CdcOffset> + S
 pub trait ExternalTableReader {
     async fn current_cdc_offset(&self) -> ConnectorResult<CdcOffset>;
 
+    // Currently, MySQL cdc uses a connection pool to manage connections to MySQL, and other CDC processes do not require the disconnect step for now.
+    #[allow(clippy::unused_async)]
```
hzxa21, line 204: Why do we need this? I thought it is always used.

hzxa21:
@cyliu0 @wcy-fdu I remembered we have a test pipeline for cdc. Given that this is a critical correctness fix, can we trigger the test pipeline before mering this PR?

src/connector/src/source/cdc/external/mod.rs
wcy-fdu: This is because other cdc(pg, mssql..) does not need this method.

src/connector/src/source/cdc/external/mysql.rs
wcy-fdu: A bit hard because we need to `drop(upstream_table_reader);` after `upstream_table_reader.disconnect()`, since other cdc's table_reader need manually drop.

wcy-fdu: Talked offline, passing self ownership means drop automatically.

wcy-fdu:
> Failed: 0 Passed:8 FailedExecutions:
> 
> PassedExecutions: https://buildkite.com/risingwave-test/ch-benchmark-mysql-cdc-shared-source/builds/227 https://buildkite.com/risingwave-test/ch-benchmark-mysql-cdc-shared-source/builds/227 https://buildkite.com/risingwave-test/ch-benchmark-pg-cdc-shared-source/builds/221 https://buildkite.com/risingwave-test/ch-benchmark-pg-cdc-shared-source/builds/221 https://buildkite.com/risingwave-test/sysbench-cdc/builds/1247 https://buildkite.com/risingwave-test/sysbench-cdc/builds/1247 https://buildkite.com/risingwave-test/tpcc-mysql-cdc-shared-source/builds/98 https://buildkite.com/risingwave-test/tpcc-pg-cdc-shared-source/builds/92

All test passed.

wcy-fdu:
cc @hzxa21 PTAL❤️

PR #22124, opened 2025-06-05
feat(catalog): add fragment backfill progress view

kwannoel:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## What's changed and what's your intention?

In this PR, we introduced the fragment backfill progress view. This view will display the backfill progress per fragment, in a similar format to `show jobs`.

### Motivation

After backfill order is introduced, we need to also provide a way for users to observe the backfill progress per backfill node. Currently users can only view this progress at a job level, and not at a fragment level via show jobs.

The user flow will be something like:
1. `DESCRIBE FRAGMENTS creating_job_name` or `explain <query>`
2. After seeing the fragment_ids of their stream job / backfill table names, they can run `select * from rw_catalog.rw_fragment_backfill_progress where job_id = ...` to check the progress of their fragments.

### Implementation

We have several pieces which we use to support this.
1. A table function to query the current row count: https://github.com/risingwavelabs/risingwave/pull/22122
2. A rw_stream_scan table to query the upstream table ids. https://github.com/risingwavelabs/risingwave/pull/22123, used for (3) to filter out only backfill table stats. It also queries the fragment id of the backfill node so we can join (1) and (3).
3. Reusing rw_table_stats to query total row count for backfill tables so we can compute a (%) progress.

## Checklist

- [ ] I have written necessary rustdoc comments.
- [ ]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  I have checked the [Release Timeline](https://github.com/risingwavelabs/rw-commits-history/blob/main/release_timeline.md) and [Currently Supported Versions](https://docs.risingwave.com/changelog/release-support-policy#support-end-dates-for-recent-releases) to determine which release branches I need to cherry-pick this PR into. 

## Documentation

- [x]  My PR needs documentation updates. 

The system catalog `rw_fragment_backfill_progress` provides a fragment level backfill progress, to allow you to track how each of the backfill partitions are progressing.

Here's an example of how it can be used:
```
drop materialized view if exists mv1;
drop table if exists t1;
drop table if exists t2;
[...]

PR #22129, opened 2025-06-05
feat(frontend): let DESCRIBE FRAGMENTS describe creating stream jobs

kwannoel:

### Changes

This PR has the following changes:

1. Introduced a new `bind_streaming_relation_id_by_name` function that can bind any streaming relation (table, sink, index, materialized view) by name, including those being created.

2. Added a test case for describing fragments of a backfilling materialized view to ensure the command works correctly during the creation process.

### Motivation

Viewing fragments of a creating stream job is necessary for users to observe their backfill order progress. They can check the fragments which contain the stream scan nodes, and from there obtain the fragment ids.

Using the fragment ids, they can check the backfill progress of a specific backfill fragment after https://github.com/risingwavelabs/risingwave/pull/22124 is merged.

These capabilities are important for **backfill order control**, since the ordering means that not all backfills of a stream job will happen concurrently. And are instead scheduled per fragment. To know where the progress is at, the user can use the above workflow to observe it.

### Implementation Notes

Before this PR, we had more specific error messages for relations which were not stream jobs. But after this PR, we only return a `NotFound` error, if the relation exists, but is not a stream job.

This is intentional. There were two directions we could take to implement the binding of `creating jobs`:
1. Add a flag to toggle binding creating jobs / only created jobs. This approach would make the binding function for relation even more complex. It is already quite complex. But the benefit is that it preserves the original hints and error UX.
2. (Recommended) Add a new binder function which only binds streaming relations. This function can be much simpler. It will also explicitly tell the user which specific job types are looked up in the error message. While it's not as explicit as the previous approach, I think it's clear enough. This also keeps the function small and maintainable. Further, it's unlikely for users to run `describe fragments` on shared source and system catalog. 

## Checklist

### DIFF
--- a/e2e_test/batch/describe_fragments.slt
+++ b/e2e_test/batch/describe_fragments.slt
@@ -256,6 +256,14 @@ set backfill_rate_limit=1;
 statement ok
 create materialized view describe_plan_test.backfilling_mv as select * from describe_plan_test.t;
 
+skipif madsim
+statement ok
+set background_ddl=false;
+
+skipif madsim
+statement ok
+set backfill_rate_limit=default;
+
 statement ok
 flush;
 


kwannoel:
* **#22129** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22129?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22129?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

e2e_test/batch/describe_fragments.slt
```diff
@@ -95,10 +95,9 @@ DESCRIBE FRAGMENTS describe_plan_test.src;
 ----
 db error: ERROR: Failed to run the query
 
-Caused by:
-  Not supported: non shared source has no fragments to describe
-HINT: Use `DESCRIBE` instead.
-
+Caused by these errors (recent errors listed first):
+  1: Catalog error
+  2: materialized view, index, table or sink not found: describe_plan_test.src
```
xxchan: I think previous error msgs are more user-friendly. Can we keep that?

kwannoel: I elaborated in my PR description for this part.

kwannoel: I can add more description, tell users not to describe fragments their `source` and `view` I think that's probably the most likely.

xxchan:
> These capabilities are important for backfill order control, since the ordering means that not all backfills of a stream job will happen concurrently. And are instead scheduled per fragment. To know where the progress is at, the user can use the above workflow to observe it.

Regarding this motivation, I think the final goal should be that user can see the progress and fragment graph integrated into one command. (Maybe `DESCRIBE FRAGMENTS`, or `show jobs`, or a new command) `DESCRIBE + select from backfill_progress` looks not good user experience.

kwannoel:
> > These capabilities are important for backfill order control, since the ordering means that not all backfills of a stream job will happen concurrently. And are instead scheduled per fragment. To know where the progress is at, the user can use the above workflow to observe it.
> 
> Regarding this motivation, I think the final goal should be that user can see the progress and fragment graph integrated into one command. (Maybe `DESCRIBE FRAGMENTS`, or `show jobs`, or a new command) `DESCRIBE + select from backfill_progress` looks not good user experience.

True, should technically be possible based on my experience with `explain analyze`. We may use the same structure to do it.

e2e_test/batch/describe_fragments.slt
kwannoel: Fixed: https://github.com/risingwavelabs/risingwave/pull/22129/commits/9ed131f74db2303b06dbdc4786c5afde2ba9df52. I decide to follow your suggestion and keep the old error UX.

> I can add more description, tell users not to describe fragments their source and view I think that's probably the most likely.

Keep everything in the same message also is unclear, if they have actually missing stream jobs, and they see this message, it will mislead the user.

e2e_test/batch/describe_fragments.slt
```diff
@@ -97,8 +97,7 @@ db error: ERROR: Failed to run the query
 
 Caused by:
   Not supported: non shared source has no fragments to describe
-HINT: Use `DESCRIBE` instead.
-
+HINT: USE `DESCRIBE` instead
```
chenzl25, line 100: Shall we tell users to "USE `DESCRIBE` solely without `FRAGMENTS`  instead"?

chenzl25 (approved):
Rest LGTM

e2e_test/batch/describe_fragments.slt
kwannoel: I think we can do:
```
HINT: USE `DESCRIBE` instead of `DESCRIBE FRAGMENTS`
```

kwannoel:
Implemented a simpler version here: https://github.com/risingwavelabs/risingwave/pull/22205.

The implementation in this PR is not very clean, for instance the new binding function inside the `Binder` returns the error message with a hint telling users to use `DESCRIBE` instead. But that's specific to `DESCRIBE FRAGMENTS`, rather than for any call to that function.

PR #22173, opened 2025-06-10
feat(kafka): add multiplexed reader support to Kafka source connector

shanicky:

**Summary**
This PR introduces Kafka multiplexed reader support for the source connector, enabling multiple split readers to share a single underlying Kafka consumer per connection. It also adds a generic async resource cleanup mechanism and enriches source context metadata, alongside dependency updates for simulation compatibility.

**Details**

- KafkaMuxReader implementation allows source splits for the same Kafka connection to share a consumer, reducing resource usage and broker load, with messages distributed via dedicated channels.
- Introduces an opt-in `enable_mux_reader` flag in KafkaProperties, ensuring backward compatibility and configurable activation.
- A global, async-safe registry manages mux reader lifecycles, enforcing singleton semantics and enabling clean resource release via newly introduced ReleaseHandle types.
- Extends SourceContext with optional source metadata for better contextual management and future extensibility.
- Updates simulation and test dependencies to a forked madsim suite, ensuring deterministic Kafka testing and alignment with the multiplexing logic.
- Refactors batch executor, sink, and test infrastructure to propagate new configuration and cleanup mechanisms.

```
dev=> SET streaming_use_shared_source TO true;
SET_VARIABLE
dev=> create connection c with ( type = 'kafka', properties.bootstrap.server = 'localhost:29092' );
CREATE_CONNECTION
dev=> CREATE source s (v int) WITH (
    connector = 'kafka',
    topic = 't',
    enable.mux.reader = true, 
    connection = c,
) FORMAT PLAIN ENCODE JSON;
CREATE_SOURCE
dev=> CREATE source s1 (v int) WITH (
    connector = 'kafka',
    topic = 't1',
    enable.mux.reader = true, 
    connection = c,
) FORMAT PLAIN ENCODE JSON;
CREATE_SOURCE

```
In this scenario, the source jobs for s and s1 will share the consumer, but the source backfill job will still use a separate consumer. S2 will also use an individual consumer.

## Checklist

### DIFF
--- a/src/connector/src/source/kafka/source/mux_reader.rs
+++ b/src/connector/src/source/kafka/source/mux_reader.rs
@@ -245,7 +245,9 @@ impl KafkaMuxReader {
         tpl: TopicPartitionList,
     ) -> anyhow::Result<mpsc::Receiver<Vec<OwnedMessage>>> {
         if tpl.count() == 0 {
-            anyhow::bail!("splits list is empty");
+            // Nothing to register; return an empty channel so caller can proceed gracefully.
+            let (_tx, rx) = mpsc::channel(1);
+            return Ok(rx);
         }
 
         tracing::info!(
@@ -253,22 +255,17 @@ impl KafkaMuxReader {
             tpl,
             self.key
         );
-        {
-            let map = self.senders.read().await;
-            for element in tpl.elements() {
-                let key = (element.topic().to_owned(), element.partition());
-                if map.contains_key(&key) {
-                    anyhow::bail!("split ({:?}) already registered", key);
-                }
-            }
-        }
         // sender / receiver
         let (tx, rx) = mpsc::channel(128);
         {
             let mut map = self.senders.write().await;
             for element in tpl.elements() {
+                let key = (element.topic().to_owned(), element.partition());
+                if map.contains_key(&key) {
+                    tracing::warn!("re-registering split {:?} in mux reader {}", key, self.key);
+                }
                 map.insert(
-                    (element.topic().to_owned(), element.partition()),
+                    key,
                     tx.clone(),
                 );
             }
--- a/src/connector/src/source/kafka/source/reader.rs
+++ b/src/connector/src/source/kafka/source/reader.rs
@@ -183,13 +183,28 @@ impl SplitReader for KafkaSplitReader {
         let mux_enabled = properties.enable_mux_reader.unwrap_or(false);
 
         let (message_reader, backfill_info) = if mux_enabled {
-            let connection_id = connection_id.ok_or_else(|| {
-                ConnectorError::from(anyhow!(
-                    "Kafka mux reader requires a connection bound via WITH CONNECTION when `enable.mux.reader` is true"
-                ))
-            })?;
+            tracing::info!(
+                source_id = ?source_ctx.source_id,
+                actor_id = ?source_ctx.actor_id,
+                ?connection_id,
+                topic = %properties.common.topic,
+                "Kafka mux reader connection info"
+            );
+            // Prefer the explicit connection id; fallback to a stable key derived from brokers+topic.
+            let mux_key = if let Some(id) = connection_id {
+                id.to_string()
+            } else {
+                let key = format!("{}|{}", properties.connection.brokers, properties.common.topic);
+                tracing::warn!(
+                    source_id = ?source_ctx.source_id,
+                    actor_id = ?source_ctx.actor_id,
+                    %key,
+                    "Kafka mux reader missing connection_id, falling back to broker/topic key"
+                );
+                key
+            };
             let reader = KafkaMuxReader::get_or_create(
-                connection_id.to_string(),
+                mux_key,
                 properties.clone(),
                 source_ctx.clone(),
             )
--- a/src/frontend/src/catalog/source_catalog.rs
+++ b/src/frontend/src/catalog/source_catalog.rs
@@ -62,6 +62,15 @@ impl SourceCatalog {
         self.definition.clone()
     }
 
+    /// Returns a cloned `StreamSourceInfo` with connector `connection_id` populated when missing.
+    pub fn info_with_connection(&self) -> StreamSourceInfo {
+        let mut info = self.info.clone();
+        if info.connection_id.is_none() {
+            info.connection_id = self.connection_id;
+        }
+        info
+    }
+
     /// Returns the parsed SQL definition when the source was created.
     ///
     /// Returns error if it's invalid.
--- a/src/frontend/src/optimizer/plan_node/stream_fs_fetch.rs
+++ b/src/frontend/src/optimizer/plan_node/stream_fs_fetch.rs
@@ -102,6 +102,7 @@ impl StreamNode for StreamFsFetch {
         let source_inner = source_catalog.map(|source_catalog| {
             let (with_properties, secret_refs) =
                 source_catalog.with_properties.clone().into_parts();
+            let source_info = source_catalog.info_with_connection();
             PbStreamFsFetch {
                 source_id: source_catalog.id,
                 source_name: source_catalog.name.clone(),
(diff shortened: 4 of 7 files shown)


Cargo.toml
```diff
@@ -335,7 +335,7 @@ redundant_explicit_links = "allow"
 lto = "off"
 # use parallel frontend to speed up build
 # TODO: may consider applying to release/production profile as well
-rustflags = ["-Z", "threads=8"]
+#rustflags = ["-Z", "threads=8"]
```
shanicky: will restore

src/connector/Cargo.toml
```diff
@@ -81,6 +81,7 @@ mysql_common = { version = "0.32", default-features = false, features = [
 ] }
 nexmark = { version = "0.2", features = ["serde"] }
 num-bigint = "0.4"
+once_cell = "1.21.3"
```
tabVersion, line 103: Lets move to the Cargo.toml in the root path

src/connector/src/sink/kafka.rs
```diff
@@ -289,6 +289,7 @@ impl From<KafkaConfig> for KafkaProperties {
             aws_auth_props: val.aws_auth_props,
             group_id_prefix: None,
             unknown_fields: Default::default(),
+            enable_mux_reader: None,
```
tabVersion, line 304: shall we move the field to `KafkaCommon`? It feels like valid for both source and sink

src/connector/src/source/kafka/source/mux_reader.rs
```diff
@@ -0,0 +1,271 @@
...
+
+            for ((topic, partition), messages) in grouped_messages {
+                let key = (topic.to_owned(), partition);
+                if let Some(sender) = senders_guard.get(&key) {
+                    if sender.send(messages).await.is_err() {
+                        // todo
+                        break;
+                    }
+                }
+            }
```
tabVersion: is it possible to directly iter over `owned_msgs` to avoid clone each message

src/connector/src/source/kafka/source/reader.rs
```diff
@@ -51,6 +95,114 @@ pub struct KafkaSplitReader {
...
+        self: Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+    ) -> Poll<Option<Self::Item>> {
+        self.project().inner.poll_next(cx)
+    }
+}
+
+#[pinned_drop]
+impl<S> PinnedDrop for StreamWithSyncCleanup<S> {
+    fn drop(self: Pin<&mut Self>) {
```
tabVersion: please add some comment on what happens in drop

tabVersion (commented):
too complex, I surrender 😢

tabVersion:
> This PR attempts to add a shared Kafka consumer within the compute. When the mux_reader option is enabled, the root job of shared sources under the same connection will share this consumer. New splits will be added by calling incremental_assign on the shared consumer, rather than creating new Kafka consumers.

So the create source should satisfy both 1) create from an existing connection; 2) spec `enable.mux_reader = true` in with clause to enable the feature? Creating purely from with clause will not enable it?

> Currently, it has not been decided whether to use session variables or the with attribute to specify enabling the mux reader. The specific details still need to be discussed.

Personally vote for in with clause because the feature is mainly designed for kafka, so far we dont see similar fd issue on other connectors. So it is designed for kafka and we don't intend to make it open to other connectors, leave it in with clause makes it connector spec.

src/connector/src/source/kafka/source/mux_reader.rs
```diff
@@ -0,0 +1,271 @@
...
+
+            if owned_msgs.is_empty() {
+                continue;
+            }
+
+            let mut grouped_messages: HashMap<(&str, i32), Vec<OwnedMessage>> = HashMap::new();
+
+            for msg in &owned_msgs {
+                let route = (msg.topic(), msg.partition());
+                grouped_messages.entry(route).or_default().push(msg.clone());
```
shanicky: clone will be fixed

BugenZhao:
> WIP State

Is this ready for review?

src/connector/src/source/kafka/source/mux_reader.rs
```diff
@@ -0,0 +1,288 @@
...
+            consumer,
+            senders: RwLock::new(HashMap::new()),
+        });
+
+        let max_chunk_size = source_ctx.source_ctrl_opts.chunk_size;
+
+        Self::registry()
+            .write()
+            .await
+            .insert(connection_id, Arc::clone(&reader));
```
BugenZhao: This doesn't seems to ensure single-flight. Shall we consider `moka`?

src/connector/src/source/kafka/source/reader.rs
```diff
@@ -51,6 +95,114 @@ pub struct KafkaSplitReader {
...
+        let handle = std::mem::replace(
+            // Safety: we are in drop, it's OK to get a mutable reference
+            unsafe { self.map_unchecked_mut(|s| &mut s.handle) }.get_mut(),
+            ReleaseHandle::noop(),
```
BugenZhao: Since `handle` is not pinned, I guess you can safely get a mutable reference from `self.project()`?

src/stream/src/executor/source/source_executor.rs
```diff
@@ -381,6 +307,10 @@ impl<S: StateStore> SourceExecutor<S> {
             target_state
         );
 
+        if source_desc.source.config.enable_mux_reader() {
+            stream.drop_data_stream();
+        }
+
```
BugenZhao: Why do we need this step? Isn't this included in `replace_data_stream`?

shanicky: This situation requires handling the case where splits are added, which could (very likely) involve changing from split (1) to splits (1,2). The original logic is to create a new split reader and then replace the old one. However, if we create the new split reader first, the underlying mux reader will need to register offsets for split 1 twice, leading to conflicts. This is a key challenge in this PR (and we haven't found an elegant solution yet). 

We need to fully drop the old reader (unregister the split) before creating the new one. I previously considered exposing a release handle in the split reader to unregister, and then calling this handle here. But the source logic is too complex, and I haven't thought of a particularly good approach.

So the current approach is to use pinned drop to forcibly drop the old split reader here, synchronously unregister the splits, and then create a new one.

src/connector/src/source/kafka/source/mux_reader.rs
tabVersion: +1

src/stream/src/executor/source/source_executor.rs
BugenZhao: Oh I see. So the main point is that we must drop the old data stream before calling `self.stream_reader_builder`. LGTM.

src/stream/src/executor/stream_reader.rs
```diff
@@ -106,6 +106,16 @@ impl<const BIASED: bool, M: Send + 'static> StreamReaderWithPause<BIASED, M> {
...
     }
 
+    pub fn drop_data_stream(&mut self) {
+        // Take the barrier receiver arm.
+        let data_receiver_arm = std::mem::replace(
+            self.inner.get_mut().1,
+            futures::stream::once(async { unreachable!("placeholder") }).boxed(),
+        );
+
+        std::mem::drop(data_receiver_arm);
```
BugenZhao:
```suggestion
        // Take the barrier receiver arm.
        let _ = std::mem::replace(
            self.inner.get_mut().1,
            futures::stream::once(async { unreachable!("must call `replace_data_stream` after `drop_data_stream` before use") }).boxed(),
        );
```

shanicky:
The madsim framework we rely on doesn't support incremental_assign, so we also need to disable the mux reader during simulation tests.

kwannoel:
>  S2 will also use an individual consumer.

Didn't see s2 definition in the description above, although it's referenced.

shanicky:
There are still a bunch of runtime errors. I'll fix them and open a new PR.

PR #22175, opened 2025-06-10
fix(mysql-cdc): enhance mysql schema check

wcy-fdu:

Previously, unsupported types were matched successfully by default. This PR changed the default behavior to false and added the missing data type.
For bit type
- bit(1) match bool
- bit(>1) match bytea

## Checklist

## Documentation

<details>

</details>


e2e_test/source_inline/tvf/mysql_query.slt
```diff
@@ -73,12 +89,12 @@ create source mysql_cdc_source with (
 query
 select * from mysql_query('$MYSQL_HOST', '$MYSQL_TCP_PORT', '$RISEDEV_MYSQL_USER', '$MYSQL_PWD', 'tvf', 'select * from test;');
 ----
-1 t 1 2 3 4 5 6 7 1.08 1.09 1.10 1.11 char varchar \x000a \x16 \x17 \x18 \x19 2021-01-01 12:34:56 2021-01-01 12:34:56+00:00 {"key1": 1, "key2": "abc"} NULL
+1 t 1 2 3 4 5 6 7 1.08 1.09 1.10 1.11 char varchar \x000a \x16 \x17 \x18 \x19 2021-01-01 12:34:56 2021-01-01 12:34:56+00:00 {"key1": 1, "key2": "abc"} NULL t 2023 b 12345 \x74657874 \x74696e7974657874 \x6d656469756d74657874 \x6c6f6e6774657874
```
tabVersion: shall we add one more test case against `v24 bit(1)`, with exact one bit into a `bytea` column

java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/MySqlValidator.java
```diff
@@ -292,6 +311,18 @@ private boolean isDataTypeCompatible(String mysqlDataType, Data.DataType.TypeNam
...
                 return val == Data.DataType.TypeName.INT64_VALUE;
+            case "boolean":
+            case "bool":
+                return val == Data.DataType.TypeName.BOOLEAN_VALUE;
+            case "enum":
+                return val == Data.DataType.TypeName.VARCHAR_VALUE;
+            case "text":
+            case "tinytext":
+            case "mediumtext":
+                return val == Data.DataType.TypeName.VARCHAR_VALUE;
```
xiangjinwu, line 324: nit: put together with char and varchar below

java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/MySqlValidator.java
```diff
@@ -292,6 +311,18 @@ private boolean isDataTypeCompatible(String mysqlDataType, Data.DataType.TypeNam
...
+            case "longtext":
+                return val == Data.DataType.TypeName.BYTEA_VALUE
```
xiangjinwu, line 326: Do we known when is longtext -> bytea needed?

e2e_test/source_inline/tvf/mysql_query.slt
```diff
@@ -32,7 +32,15 @@ CREATE TABLE test (
     v20 time,
     v21 timestamp,
     v22 json,
-    v23 int
+    v23 int,
+    v24 bit(1),
+    v25 year,
+    v27 enum('a', 'b', 'c'),
```
xiangjinwu: What was v26?

xiangjinwu (commented):
I notice 3 lists of data types:
* the names matched in Java code
* the columns in test
* the mapping table on [doc website](https://docs.risingwave.com/ingestion/sources/mysql/mysql-cdc#data-type-mapping)

Ideally these 3 lists shall match, and **this PR makes great progress towards it**. But there are still some missing pieces:
* The doc separates `FLOAT[(P)]` from `FLOAT(M,D)`, while the latter is not in code or test. (It is [deprecated](https://dev.mysql.com/doc/refman/8.0/en/floating-point-types.html) by MySQL.)
* `datetime` is in doc and code but not test
* `numeric` is in doc and test but not code. If they are just synonyms of the same type, we shall also merge them to 1 line in doc (just like bool/boolean, int/integer)
* more to be discovered by you. If all 3 lists are sorted the same the comparison would be much easier.

wcy-fdu:
> * the columns in test

Good catch! I think the source of truth should be out implementation, instead of doc. I plan to first cover all supported data type, and cover all of it in test and finally refine doc.

e2e_test/source_inline/tvf/mysql_query.slt
wcy-fdu: v26 is binary(4), but a bit wield I found it can not be shown in CI env, but in local env it works. So i just remove it, any idea about the garbled characters?
```

[Diff] (-expected\|+actual)
--
  | 2025-06-26 09:39:57 UTC | -   1 t 1 2 3 4 5 6 7 1.08 1.09 1.10 1.11 char varchar \x000a \x16 \x17 \x18 \x19 2021-01-01 12:34:56 2021-01-01 12:34:56+00:00 {"key1": 1, "key2": "abc"} NULL t 2023 \x01020304 b 12345 \x74657874 \x74696e7974657874 \x6d656469756d74657874 \x6c6f6e6774657874
  | 2025-06-26 09:39:57 UTC | +   1 t 1 2 3 4 5 6 7 1.08 1.09 1.10 1.11 char varchar \x000a \x16 \x17 \x18 \x19 2021-01-01 12:34:56 2021-01-01 12:34:56+00:00 {"key1": 1, "key2": "abc"} NULL t 2023  b 12345 \x74657874 \x74696e7974657874 \x6d656469756d74657874 \x6c6f6e6774657874


```

xiangjinwu: I located the error: `pg_typeof(v26)` is varchar from batch source tvf `mysql_query` rather than bytea. The four characters results from decoding `[0x01, 0x02, 0x03, 0x04]` as string.

e2e_test/source_inline/tvf/mysql_query.slt
wcy-fdu: After some offline discussion, we decide to ban bit(1) -> bytea, which is compatible with [debezium](https://debezium.io/documentation/reference/stable/connectors/mysql.html#mysql-basic-types) behavior.

xiangjinwu (approved):
* Update PR description for bit(1) -> bool
* Let's add back the `binary` type and open an issue for `mysql_query`

Rest LGTM

e2e_test/source_inline/tvf/mysql_query.slt
wcy-fdu: Fix by https://github.com/risingwavelabs/risingwave/pull/22442

xiangjinwu:
Let's cherry-pick after/together with #22442

PR #22178, opened 2025-06-10
feat(ddl): Implement database-level barrier and checkpoint configuration - Part 2

Chiro11:

## Checklist

## Documentation

<details>

</details>


src/meta/src/barrier/manager.rs
```diff
@@ -86,6 +87,22 @@ impl GlobalBarrierManager {
...
     }
 
+    pub fn update_database_barrier(
+        &self,
+        database_id: DatabaseId,
+        barrier_interval_ms: Option<u32>,
+        checkpoint_frequency: Option<u64>,
+    ) -> MetaResult<()> {
+        self.request_tx
+            .send(BarrierManagerRequest::UpdateDatabaseBarrier {
```
wenym1, line 98: Shall we include a oneshot channel tx in request, and wait on the rx, to ensure that the alter param takes effect before we return?

src/meta/src/barrier/checkpoint/control.rs
```diff
@@ -184,12 +170,13 @@ impl CheckpointControl {
         control_stream_manager: &mut ControlStreamManager,
     ) -> Option<HashMap<DatabaseId, MetaError>> {
         let NewBarrier {
+            database_id,
```
wenym1, line 173: Since we only handle the barrier of a single database, we may change to return `MetaResult<()>`, which represents the result of a single database.

src/meta/src/barrier/checkpoint/control.rs
```diff
@@ -289,7 +256,6 @@ impl CheckpointControl {
                 span.clone(),
                 control_stream_manager,
                 &self.hummock_version_stats,
```
wenym1, line 256: After this PR, it's unnecessary for us to handle all databases in the for loop below any more.

src/meta/src/barrier/schedule.rs
```diff
@@ -320,63 +323,209 @@ pub struct ScheduledBarriers {
...
     }
 
-    pub(super) fn set_min_interval(&mut self, min_interval: Duration) {
-        let set_new_interval = min_interval != self.min_interval.period();
-        if set_new_interval {
-            let mut min_interval = tokio::time::interval(min_interval);
-            min_interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Delay);
-            self.min_interval = min_interval;
+    /// Make the `checkpoint` of the next barrier must be true.
+    pub fn force_checkpoint_in_next_barrier(&mut self) {
```
wenym1: Seems this method is called only within the scope of `DatabaseCheckpointControl`, which means we can specify the `database_id` we want to force checkpoint on, instead of all databases.

src/meta/src/barrier/schedule.rs
```diff
@@ -320,63 +323,209 @@ pub struct ScheduledBarriers {
...
+        let new_barrier = select! {
             biased;
             scheduled = context.next_scheduled() => {
-                self.min_interval.reset();
-                let checkpoint = scheduled.command.need_checkpoint() || checkpoint;
+                let database_id = scheduled.database_id;
+                // Check if the database exists.
+                assert!(self.databases.contains_key(&database_id), "database {} not found in periodic barriers", database_id);
+                assert!(self.timer_streams.contains_key(&database_id), "timer stream for database {} not found in periodic barriers", database_id);
+                // New command will trigger the barriers for all other databases, so reset all timers.
```
wenym1: This is not true if with the comment above.

src/meta/src/barrier/schedule.rs
```diff
@@ -320,63 +323,209 @@ pub struct ScheduledBarriers {
...
-        let checkpoint = self.try_get_checkpoint();
-        let scheduled = select! {
+        let new_barrier = select! {
             biased;
             scheduled = context.next_scheduled() => {
-                self.min_interval.reset();
-                let checkpoint = scheduled.command.need_checkpoint() || checkpoint;
+                let database_id = scheduled.database_id;
+                // Check if the database exists.
+                assert!(self.databases.contains_key(&database_id), "database {} not found in periodic barriers", database_id);
```
wenym1, line 487: The assertion here assumes that the database should have been initialized before receiving the command. 

However, when we create the database, we don't wait to ensure that the `UpdateDatabaseBarrier` request is already handled. The `UpdateDatabaseBarrier` request is sent in a separate channel to the command channel, which means, theoretically it's possible that the command is handled earlier than the request. In this case, when we handle the command, the database is not initialized, and then we will hit this assertion.

src/meta/src/barrier/schedule.rs
```diff
@@ -320,63 +323,209 @@ pub struct ScheduledBarriers {
...
+
+        // If the database already has a timer stream, reset it with the new interval.
+        let duration = if let Some(ms) = barrier_interval_ms {
+            Duration::from_millis(ms as u64)
+        } else {
+            self.sys_barrier_interval
+        };
+        self.timer_streams.insert(
+            database_id,
+            IntervalStream::new(tokio::time::interval(duration)),
```
wenym1: Seems that we didn't do `interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Delay)` like other usages.

Shall we have a unified function to create the interval stream from duration?

src/meta/src/barrier/schedule.rs
```diff
@@ -320,63 +323,209 @@ pub struct ScheduledBarriers {
...
+                    database_id: scheduled.database_id,
+                    command: Some((scheduled.command, scheduled.notifiers)),
                     span: scheduled.span,
                     checkpoint,
                 }
             },
-            _ = self.min_interval.tick() => {
+            // If there is no database, we won't wait for `Interval`, but only wait for command.
+            // Normally it-branch will not be true because there is always at least one database.
+            next_timer = self.timer_streams.next(), if !self.timer_streams.is_empty() => {
```
wenym1: Can use `pending_on_none(self.timer_streams.next())`

src/meta/src/barrier/schedule.rs
Chiro11: I am actually considering that: if there are too many commands, it may cause starvation of barriers in other databases (select with bias is used here, so the first branch has a higher priority). Do you think this will not happen?

src/meta/src/barrier/schedule.rs
Chiro11: Indeed. We should wait up there first.

src/meta/src/barrier/schedule.rs
wenym1: We can assume that command is not that frequent, and the handling of each command should be fast enough, since the logic does not involve async method.

Chiro11: Ok.

wenym1:
Please also update the release note in the `Documentation` section of the PR description, since this PR is a user facing change.

src/meta/src/barrier/worker.rs
```diff
@@ -473,29 +473,26 @@ impl<C: GlobalBarrierWorkerContext> GlobalBarrierWorker<C> {
...
+                                    database_id,
+                                    e.as_report()
+                                )
                             );
                         }
                         let result: MetaResult<_> = try {
                             if !self.enable_per_database_isolation() {
-                                let errs = failed_databases.iter().map(|(database_id, e)| (database_id, e.as_report())).collect_vec();
-                                let err = anyhow!("failed to inject barrier for databases: {:?}", errs);
+                                let err = anyhow!("failed to inject barrier for databases: {:?}", (database_id, e.as_report()));
```
wenym1:
```suggestion
                                let err = anyhow!("failed to inject barrier to database {}: {:?}", database_id, e.as_report());
```

wenym1 (approved):
LGTM.

The added unit tests look flaky. May need some way to make it more stable.

PR #21943, opened 2025-05-21
fix(cdc/batch): special char during debezium props substitution and `postgres_query`

xiangjinwu:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## What's changed and what's your intention?

Fixes #21932 by parsing the properties file first and do substitution in-memory. This avoid the substituted text from being handled by the properties-parser.

Also fixes a similar issue in `postgres_query` (batch PostgreSQL source). We shall pass parameters structurally rather than via a interpolated connection string, which would also require complicated escaping to be parsed back losslessly.

Also identified issues in our dev/test tools:
* `risedev` (using `cargo-make`/`envmnt`) environment variable passing [#21974](https://github.com/risingwavelabs/risingwave/issues/21974). Dodged the issue by testing with `\t` rather than `\n`.
* `sqllogictest` substitution: https://gist.github.com/xiangjinwu/f93c11de855cfa38fc484cafad224ecd

## Checklist

- [ ] I have written necessary rustdoc comments.
- [x]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [x]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [x]  I have checked the [Release Timeline](https://github.com/risingwavelabs/rw-commits-history/blob/main/release_timeline.md) and [Currently Supported Versions](https://docs.risingwave.com/changelog/release-support-policy#support-end-dates-for-recent-releases) to determine which release branches I need to cherry-pick this PR into. 

## Documentation

- [ ]  My PR needs documentation updates. 

<details>
<summary><b>Release note</b></summary>

</details>

PR #22182, opened 2025-06-10
fix(ci): update main-cron PostgreSQL meta backend password

### LINKED ISSUE #22167: ci: main-cron cannot connect to postgres meta store
Likely due to password update in #21943

xiangjinwu:

Fixes #22167. The password was updated in #21943 to test for special character handling, but the update was incomplete and missed some test suites in main-cron.

Also fixes the handling when passing meta password to iceberg jdbc. `Url::password` returns the percent encoded results and we need to decode ourselves to get back the original value.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/handler/create_table.rs
+++ b/src/frontend/src/handler/create_table.rs
@@ -21,6 +21,7 @@ use clap::ValueEnum;
 use either::Either;
 use fixedbitset::FixedBitSet;
 use itertools::Itertools;
+use percent_encoding::percent_decode_str;
 use pgwire::pg_response::{PgResponse, StatementType};
 use prost::Message as _;
 use risingwave_common::catalog::{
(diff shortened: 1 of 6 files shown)


src/frontend/src/handler/create_table.rs
```diff
@@ -1472,10 +1473,11 @@ pub async fn create_iceberg_engine_table(
     let meta_store_user = meta_store_endpoint.username().to_owned();
     let meta_store_password = meta_store_endpoint
         .password()
+        .and_then(|p| percent_decode_str(p).decode_utf8().ok())
```
tabVersion, line 1476: how do we handle `post\tgres` if not decode here

xiangjinwu: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=18c32db594c367bfdbf552511dc9cb62

chenzl25 (approved):
rubber stamp

BugenZhao (approved):

PR #22198, opened 2025-06-11
refactor(s3 connector): expose `enable_config_load` to user for s3 source and sink

wcy-fdu:

Previously users are required to configure an access key and access ID to access, which is intended to isolate Hummock's S3. However, for on-prem users, it is more reasonable to expose this configuration to them. Users can set `disable_config_load` to allow the S3 source and sink to directly obtain authentication information from the pod's environment. 
This PR places `disable_config_load` within the with option, and the default behavior is set to true.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/src/sink/file_sink/s3.rs
+++ b/src/connector/src/sink/file_sink/s3.rs
@@ -17,11 +17,13 @@ use anyhow::anyhow;
 use opendal::Operator;
 use opendal::layers::{LoggingLayer, RetryLayer};
 use opendal::services::S3;
+use risingwave_common::util::env_var::env_var_is_true;
 use serde::Deserialize;
 use serde_with::serde_as;
 use with_options::WithOptions;
 
 use super::opendal_sink::{BatchingStrategy, FileSink};
+use crate::connector_common::DISABLE_DEFAULT_CREDENTIAL;
 use crate::deserialize_optional_bool_from_string;
 use crate::sink::file_sink::opendal_sink::OpendalSinkBackend;
 use crate::sink::{Result, SINK_TYPE_APPEND_ONLY, SINK_TYPE_OPTION, SINK_TYPE_UPSERT, SinkError};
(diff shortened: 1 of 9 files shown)


src/connector/src/sink/file_sink/s3.rs
```diff
@@ -34,6 +34,8 @@ pub struct S3Common {
     /// The directory where the sink file is located.
     #[serde(rename = "s3.path", alias = "snowflake.s3_path", default)]
     pub path: Option<String>,
+    #[serde(rename = "s3.disable.config.load", default)]
+    pub disable_config_load: Option<bool>,
```
chenzl25: How about making it consistent with iceberg configuration? we change the the name to `enable_config_load` and don't use `rename` since dot is misused. https://github.com/risingwavelabs/risingwave/blob/04b295347d442a82e85d8a3572f9eb229ff3c539/src/connector/src/connector_common/iceberg/mod.rs#L99

src/connector/src/sink/file_sink/s3.rs
```diff
@@ -82,26 +86,20 @@ impl<S: OpendalSinkBackend> FileSink<S> {
...
         }
 
         if let Some(assume_role) = config.common.assume_role {
             builder = builder.role_arn(&assume_role);
         }
-        builder = builder.disable_config_load();
+        // Default behavior is disable loading config from environment.
+        if !config.common.enable_config_load.unwrap_or(false) {
+            builder = builder.disable_config_load();
+        }
```
chenzl25: Please extract a method and consider the env DISABLE_DEFAULT_CREDENTIAL like this https://github.com/risingwavelabs/risingwave/blob/436862f37b0963302af25d21913f7f06866fc6bc/src/connector/src/connector_common/iceberg/mod.rs#L165-L171

wcy-fdu: `DISABLE_DEFAULT_CREDENTIAL` is not a an ENV variable defined in OpenDAL, will cloud side read it? If not, seems no need to check it.

chenzl25: The cloud side would set it.

chenzl25 (approved):
LGTM

PR #22200, opened 2025-06-11
fix(pg-cdc): enhance postgres cdc schema check

wcy-fdu:

This PR fixes a series of issues with the previous schema checks and specifies the latest supported schema list. It performs strict schema checks when creating Postgres CDC tables. The following types are noteworthy:
- Supports 1-dimensional arrays
- Supports enum types

There are many corner cases with data types, and this PR cannot guarantee that the schema checks are completely comprehensive (for example, there are some user-defined types and situations where it's difficult to distinguish between 1-dimensional and multi-dimensional arrays). However, this is a good start.

## Checklist

## Documentation

<details>

</details>


e2e_test/source_inline/cdc/postgres/postgres_schema_check.slt
```diff
@@ -0,0 +1,296 @@
...
+  v45 bigint,
+  v46 bigint,
+  v47 character varying
+) from s table 'public.t';
+
+sleep 2s
+
+system ok
+psql -c "
+    INSERT INTO t (v2, v3, v4, v5, v6, v7, v8, v9, v10, v11, v12, v13, v14, v15, v16, v17, v18, v19, v20, v21, v22, v23, v24, v25, v26, v27, v28, v29, v31, v32, v33, v34, v35, v36, v37, v38, v39, v40, v41, v43, v44, v45, v46, v47) VALUES (
```
tabVersion: skipping v1?

java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/PostgresValidator.java
```diff
@@ -652,55 +666,144 @@ public void close() throws Exception {
...
+                } else {
+                    // bit(n>1) is not supported
+                    return false;
+                }
             case "smallint":
-                return Data.DataType.TypeName.INT16_VALUE <= val
-                        && val <= Data.DataType.TypeName.INT64_VALUE;
+            case "smallserial":
+                // SMALLINT, SMALLSERIAL -> SMALLINT
+                return val == Data.DataType.TypeName.INT16_VALUE;
```
tabVersion: Is it banned to map int16 into a int32 column?

e2e_test/source_inline/cdc/postgres/postgres_schema_check.slt
wcy-fdu: v1 is auto generated. (`DEFAULT gen_random_uuid()`)

java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/PostgresValidator.java
wcy-fdu: Yes.

java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/PostgresValidator.java
```diff
@@ -212,25 +211,40 @@ private void validateTableSchema() throws SQLException {
...
+
+            Map<String, ColumnInfo> schema = new HashMap<>();
             while (res.next()) {
                 var field = res.getString(1);
                 var dataType = res.getString(2);
-                schema.put(field.toLowerCase(), dataType);
+                Long charMaxLength =
+                        res.getObject(3) == null ? null : ((Number) res.getObject(3)).longValue();
+                var udtName = res.getString(4);
+                schema.put(field.toLowerCase(), new ColumnInfo(dataType, charMaxLength, udtName));
```
xiangjinwu: The existing `toLowerCase` may not be correct. In PostgreSQL:
```
create table t (a int, "A" varchar);
```

java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/PostgresValidator.java
```diff
@@ -212,25 +211,40 @@ private void validateTableSchema() throws SQLException {
...
-                var dataType = schema.get(e.getKey().toLowerCase());
-                if (dataType == null) {
+                var colInfo = schema.get(e.getKey().toLowerCase());
+                if (colInfo == null) {
                     throw ValidatorUtils.invalidArgument(
                             "Column '" + e.getKey() + "' not found in the upstream database");
                 }
-                if (!isDataTypeCompatible(dataType, e.getValue())) {
+                if (!isDataTypeCompatible(
+                        colInfo.dataType, e.getValue(), colInfo.charMaxLength, colInfo.udtName)) {
```
xiangjinwu: nit: why not pass `colInfo` as a whole, in case there are more details in the future?

java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/PostgresValidator.java
```diff
@@ -652,55 +666,144 @@ public void close() throws Exception {
...
+                    // bit(1) -> BOOLEAN
+                    return val == Data.DataType.TypeName.BOOLEAN_VALUE;
+                } else {
+                    // bit(n>1) is not supported
+                    return false;
+                }
             case "smallint":
-                return Data.DataType.TypeName.INT16_VALUE <= val
-                        && val <= Data.DataType.TypeName.INT64_VALUE;
+            case "smallserial":
```
xiangjinwu: `{serial,,big}serial` are not data types. They are just `{small,,big}int` with default value pointing to a sequence generator. Try the following:
```
create table t (v1 smallserial, v2 smallint);
select column_name, data_type from information_schema.columns where table_name = 't' and table_schema = 'public';
```

java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/PostgresValidator.java
```diff
@@ -652,55 +666,144 @@ public void close() throws Exception {
...
+                // MACADDR, MACADDR8, INT4RANGE, INT8RANGE, NUMRANGE, TSRANGE, TSTZRANGE, DATERANGE,
+                // ENUM -> CHARACTER VARYING
                 return val == Data.DataType.TypeName.VARCHAR_VALUE;
+            case "timestamptz":
+            case "timestamp with time zone":
+                // TIMESTAMPTZ, TIMESTAMP WITH TIME ZONE -> TIMESTAMP WITH TIME ZONE
+                return val == Data.DataType.TypeName.TIMESTAMPTZ_VALUE;
+            case "timetz":
+            case "time with time zone":
+            case "time without time zone":
```
xiangjinwu: Why is `time without time zone` here rather than together with its shorter form `time` below?

To be honest, if this `isDataTypeCompatible` is only used to process data from `data_type` column of `information_schema.columns`, it will NEVER see `varchar` and `character varying` at the same time - only a single form would be used by PostgreSQL. The current list is neither precise nor comprehensive.

java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/PostgresValidator.java
```diff
@@ -652,55 +666,144 @@ public void close() throws Exception {
...
             case "date":
+                // DATE -> DATE
                 return val == Data.DataType.TypeName.DATE_VALUE;
             case "time":
+            case "time(1)":
+            case "time(2)":
+            case "time(3)":
+            case "time(4)":
+            case "time(5)":
+            case "time(6)":
```
xiangjinwu: When are these branches hit?

java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/PostgresValidator.java
```diff
@@ -652,55 +666,144 @@ public void close() throws Exception {
...
+            case "money":
+                // MONEY -> NUMERIC
+                return val == Data.DataType.TypeName.DECIMAL_VALUE;
+            case "point":
+                // POINT -> STRUCT<x REAL, y REAL>
+                return val == Data.DataType.TypeName.STRUCT_VALUE;
+            case "ARRAY":
+                // ARRAY -> LIST
+
+                return val == Data.DataType.TypeName.LIST_VALUE;
```
xiangjinwu: Better check element type

java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/PostgresValidator.java
```diff
@@ -652,55 +666,144 @@ public void close() throws Exception {
...
+                            // CITEXT -> CHARACTER VARYING
+                            return val == Data.DataType.TypeName.VARCHAR_VALUE;
+                        case "ltree":
+                            return false;
+                        case "hstore":
+                            return false;
+                        default:
+                            // For other user-defined types, assume they are enum types and pass
+                            // validation
+                            return true;
```
xiangjinwu: If we still have to default to true (for the unlimited user-defined enums), it is better to check `val == Data.DataType.TypeName.VARCHAR_VALUE` and apply this to all other weakly-supported types: xml, inet, cidr, macaddr, etc

wcy-fdu: SGTM

java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/PostgresValidator.java
wcy-fdu: Checking the type of elements in a List is a bit complicated(We may need to 
- Modify SQL queries to obtain array element type information. 
- Extend the ColumnInfo class to include the element type. 
- Implement recursive type compatibility checking.). 

I know it's better to have it, and I think even if we do it, we can complete it in a subsequent independent PR.

java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/PostgresValidator.java
wcy-fdu: Indeed, thanks for pointing out.

e2e_test/source_inline/cdc/postgres/postgres_schema_check.slt
```diff
@@ -0,0 +1,296 @@
...
+        v21 point,
+        v22 citext,
+        v23 inet,
+        v24 int4range,
+        v25 int8range,
+        v26 numrange,
+        v27 tsrange,
+        v28 tstzrange,
+        v29 daterange,
+        v31 date,
```
xiangjinwu: What was v30?

e2e_test/source_inline/cdc/postgres/postgres_schema_check.slt
```diff
@@ -0,0 +1,296 @@
...
+        v33 timestamp(3),
+        v34 numeric(10,2),
+        v35 decimal(8,4),
+        v36 money,
+        v37 cidr,
+        v38 macaddr,
+        v39 macaddr8,
+        v40 int[],
+        v41 bit(1),
+        v43 smallserial,
```
xiangjinwu: and v42?

e2e_test/source_inline/cdc/postgres/postgres_schema_check.slt
```diff
@@ -0,0 +1,296 @@
...
+        'happy'
+    );
+    "
+
+sleep 10s
+
+query TTTT
+select count(*) from t;
+----
+2
```
xiangjinwu, line 226: Will ingested values be checked in a future PR?

e2e_test/source_inline/cdc/postgres/postgres_schema_check.slt
```diff
@@ -0,0 +1,296 @@
...
+        v3 character varying NOT NULL,
+        v4 integer,
+        v5 timestamp(6) without time zone NOT NULL,
+        v6 boolean,
+        v7 smallint,
+        v8 bigint,
+        v9 real,
+        v10 double precision,
+        v11 char(10),
+        v12 varchar(50),
```
xiangjinwu, line 20: If you inspect the `data_type` column of `information_schema.columns`, these are also `character` and `character varying`

e2e_test/source_inline/cdc/postgres/postgres_schema_check.slt
```diff
@@ -0,0 +1,296 @@
...
+        v13 character(20),
+        v14 character varying(100),
+        v15 timestamptz,
+        v16 timetz,
+        v17 interval,
+        v18 bytea,
+        v19 json,
+        v20 xml,
```
xiangjinwu: Add a test for `text`, which in PostgreSQL is different from `varchar`.

e2e_test/source_inline/cdc/postgres/postgres_schema_check.slt
```diff
@@ -0,0 +1,296 @@
...
+        v25 int8range,
+        v26 numrange,
+        v27 tsrange,
+        v28 tstzrange,
+        v29 daterange,
+        v31 date,
+        v32 time(3),
+        v33 timestamp(3),
+        v34 numeric(10,2),
+        v35 decimal(8,4),
```
xiangjinwu, line 43: Both of them are `numeric` but it is okay that we test both of its forms

e2e_test/source_inline/cdc/postgres/postgres_schema_check.slt
wcy-fdu: Previously v42 was bit (8), now bit (> 1) is banned.

e2e_test/source_inline/cdc/postgres/postgres_schema_check.slt
wcy-fdu: time(x) or timestamp(x)

e2e_test/source_inline/cdc/postgres/postgres_schema_check.slt
wcy-fdu: Just want to test whether the data type can pass the verification. In fact, many types will be filled with null in the backfill stage.  see 4 in https://github.com/risingwavelabs/risingwave/issues/22612. It's another job.

e2e_test/source_inline/cdc/postgres/postgres_schema_check.slt
wcy-fdu: Yes, when checking both char(x) and character(x) are character, may be a bit redundant, because some users define data type like this, so I added these two in the test.

xiangjinwu: Oh I meant the java code does not need to handle both forms because there is only going to be 1:
https://github.com/risingwavelabs/risingwave/blob/5221b8587599d35e637cfac48952c2f0bd2ab0d8/java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/PostgresValidator.java#L753-L754

Testing both forms are good: https://github.com/risingwavelabs/risingwave/pull/22200#discussion_r2309676748

Issue #32, opened 2022-01-10
ci: e2e coverage for Rust


skyzh:
Currently, e2e doesn't take Rust codebase into account when calculating coverage. In the future, we should do this.

skyzh:
cargo llvm provides an easy way to run external tests:

https://github.com/taiki-e/cargo-llvm-cov#get-coverage-of-external-tests

We can use this to generate e2e test coverage.

PR #22201, opened 2025-06-11
feat(ci): try try code coverage report for e2e tests

BugenZhao:

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -362,11 +362,13 @@ debug = "line-tables-only"
 split-debuginfo = "off"
 debug-assertions = true
 overflow-checks = true
+rustflags = ["-C", "instrument-coverage"]
 
 [profile.ci-dev]
 inherits = "dev"
 incremental = false
 debug = "line-tables-only"
+rustflags = ["-C", "instrument-coverage"]
 [profile.ci-dev.package."*"] # external dependencies
 opt-level = 1
 [profile.ci-dev.package."tokio"]
@@ -381,6 +383,7 @@ inherits = "dev"
 incremental = false
 debug = "line-tables-only"
 opt-level = 2
+rustflags = ["-C", "instrument-coverage"]
 
 [patch.crates-io]
 # Patch third-party crates for deterministic simulation.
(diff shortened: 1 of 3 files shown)


Makefile.toml
```diff
@@ -755,6 +755,12 @@ if [[ -n ${containers} ]]; then
   docker stop ${containers}
 fi
 
+# Wait for a while for all components to gracefully exit to dump profiling data for coverage.
+if [[ -n $LLVM_PROFILE_FILE ]]; then
+  echo "Waiting for 5 seconds for coverage data to be dumped..."
```
BugenZhao, line 760: We may install a signal handler to call 

```rust
unsafe extern "C" {
    fn __llvm_profile_write_file() -> libc::c_int;
}
```

to make it more robust in the future.

ci/Dockerfile
```diff
@@ -73,7 +73,7 @@ RUN curl -L --proto '=https' --tlsv1.2 -sSf https://raw.githubusercontent.com/ca
 RUN cargo binstall -y --locked cargo-llvm-cov cargo-nextest cargo-sort cargo-cache cargo-machete \
     cargo-make@0.37.24 \
     sqllogictest-bin@0.28.3 \
-    sccache@0.7.4 \
+    sccache@0.10.0 \
```
BugenZhao, line 76: The previous version of sccache does not work well with `RUSTC_WORKSPACE_WRAPPER`.

stdrc:
You are really a legend! Seems this PR finally fixes #32?

stdrc (approved):
LGTM

Makefile.toml
kwannoel: In the meantime after the 5s check, if no coverage data (i.e. `-n $LLVM_PROFILE_FILE`), can we `exit 1` to signal an error?

kwannoel (approved):
Looks like there's some TODO issues, shall we open up an issue to track all these?

Amazing work, thanks for your efforts.

ci/workflows/pull-request.yml
```diff
@@ -762,3 +753,20 @@ steps:
           - BUILDKITE_TOKEN
           - GITHUB_TOKEN
     timeout_in_minutes: 5
+
+  - wait: ~
+
+  - label: "upload coverage reports"
+    command: "ci/scripts/upload-coverage.sh"
+    # TODO: shall we skip this if `run-selected` or `CI_STEPS` is specified.
```
kwannoel, line 761: I think an immediate goal is to let us evaluate new features. So keeping `run-selected` makes sense, since new features may add test only to `main-cron`, rather than to pr, if the test is heavy.

PR #22205, opened 2025-06-11
support creating fragments

kwannoel:

## Checklist

## Documentation

<details>

</details>


kwannoel:
* **#22205** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22205?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22205?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/frontend/src/binder/relation/table_or_source.rs
```diff
@@ -187,8 +206,11 @@ impl Binder {
                                         as_of,
                                         true,
                                     );
-                                } else if let Some(table_catalog) = schema
-                                    .get_created_table_or_any_internal_table_by_name(table_name)
+                                } else if let Some(table_catalog) =
+                                    schema.get_table_by_name(table_name)
+                                    && (bind_creating_relations
+                                        || table_catalog.is_internal_table()
+                                        || table_catalog.is_created())
```
kwannoel: Preserves the original semantics + accepts any table if `bind_creating_relations` is set.

src/frontend/src/binder/relation/table_or_source.rs
```diff
@@ -137,7 +155,8 @@ impl Binder {
                         self.resolve_source_relation(&source_catalog.clone(), as_of, true)?
                     } else if let Ok((table_catalog, schema_name)) = self
                         .catalog
-                        .get_created_table_by_name(db_name, schema_path, table_name)
+                        .get_any_table_by_name(db_name, schema_path, table_name)
+                        && (bind_creating_relations || table_catalog.is_created())
```
kwannoel, line 159: Preserves the original semantics + accepts any table if `bind_creating_relations` is set.

src/frontend/src/binder/relation/table_or_source.rs
```diff
@@ -70,14 +70,32 @@ impl BoundSource {
...
 impl Binder {
+    pub fn bind_catalog_relation_by_object_name(
+        &mut self,
+        object_name: ObjectName,
+        bind_creating_relations: bool,
+    ) -> Result<Relation> {
+        let (schema_name, table_name) =
+            Binder::resolve_schema_qualified_name(&self.db_name, object_name)?;
+        self.bind_catalog_relation_by_name(
+            None,
```
kwannoel, line 81: Before this PR, we called `bind_relation_by_name` to bind the relation. We set the parameter of `allow_cross_db` to false in that function call:
```
        if let Ok(relation) = binder.bind_relation_by_name(object_name.clone(), None, None, false) {
```

If we trace the codepath, this will let the db name resolve to the binder's internal context db_name. So we can just pass in `None` here, since at L78-L79 above, we have already instantiated the binder with the correct db name.

src/frontend/src/binder/relation/table_or_source.rs
```diff
@@ -70,14 +70,32 @@ impl BoundSource {
...
+            &table_name,
+            None,
+            None,
+            bind_creating_relations,
+        )
+    }
+
     /// Binds table or source, or logical view according to what we get from the catalog.
-    pub fn bind_relation_by_name_inner(
+    pub fn bind_catalog_relation_by_name(
```
xxchan, line 91: Why to rename this? (And it becomes inconsistent with others like bind_table)

src/frontend/src/binder/relation/table_or_source.rs
xxchan: nit: `if bind_creating_relations { get_any_table_by_name } else { get_created_table_by_name }` might be slightly better. Or just add `bind_creating_relations` param to `get_table_by_name`

xxchan (approved):

src/frontend/src/binder/relation/table_or_source.rs
kwannoel: The original name `bind_relation_by_name_inner` doesn't indicate what this function does. We have to look into the implementation details, to understand that it actually just binds a catalog relation by name.

A catalog relation can be one of:
- A table catalog (corresponding to mv, table, index)
- A source catalog
- A view catalog
- A system table catalog

> And it becomes inconsistent with others like bind_table

Could you elaborate? Perhaps I can find a better name for it.

xxchan: I also can't get what "catalog relation" mean 🤔 . I mean why `bind_table` is not `bind_catalog_table`?

xxchan: Oh, table is indeed not the same as "catalog table"

xxchan: But why "relation" is not the same as "catalog relation"

kwannoel: `bind_relation_by_name` calls `bind_relation_by_name_inner` (renamed to `bind_catalog_relation_by_name`)

Inside `bind_relation_by_name`, it splits into two branches. The first branch binds a `cte` relation, i.e. a relation that's instantiated within the query itself, rather than corresponding to some existing relation in our catalogs.

In the other branch, we bind a catalog relation or a "base relation", one that existed prior the the query being bound.

For describe fragments, we only care about the latter branch, catalog relation.

src/frontend/src/binder/relation/table_or_source.rs
kwannoel: > if bind_creating_relations { get_any_table_by_name } else { get_created_table_by_name }

This makes the code rather unreadable. We end up with something like:
```
if let Ok(catalog) = { if bind_creating_relations { get_any_table_by_name } else { get_created_table_by_name } } {
 // body of branch
}
```

> Or just add bind_creating_relations param to get_table_by_name

Took a look at the implementation. If we add this to `get_any_table_by_name` as a parameter, it breaks the semantics of the "any" part. We will have to introduce a new function instead. `fn get_table_by_name(..., bind_creating_relations: bool)`. I will add it.

src/frontend/src/binder/relation/table_or_source.rs
kwannoel: Fixed https://github.com/risingwavelabs/risingwave/pull/22205/commits/3881c3f09ebd7c617f35e6081a2aec1e859b98cf

src/frontend/src/binder/relation/table_or_source.rs
kwannoel: Fixed https://github.com/risingwavelabs/risingwave/pull/22205/commits/3881c3f09ebd7c617f35e6081a2aec1e859b98cf

PR #22227, opened 2025-06-13
feat: introduce `as file uri` when ref secret

tabVersion:

in current impl, fill_secret returns the absolute path, but most sdk requires `file://<absolute-path>`

## Checklist

## Documentation

<details>

introduce `AS FILE URI` when referencing a secret, which gives `file://<absolute path>` for a file

</details>

### DIFF
--- a/proto/secret.proto
+++ b/proto/secret.proto
@@ -25,6 +25,8 @@ message SecretRef {
     TEXT = 1;
     // AS FILE
     FILE = 2;
+    // AS FILE with `file://` prefix
+    FILE_URI = 3;
   }
   uint32 secret_id = 1;
   RefAsType ref_as = 2;
--- a/src/common/secret/src/secret_manager.rs
+++ b/src/common/secret/src/secret_manager.rs
@@ -168,6 +168,11 @@ impl LocalSecretManager {
                     self.get_or_init_secret_file(secret_id, secret_value_bytes.clone())?;
                 Ok(path_str)
             }
+            RefAsType::FileUri => {
+                let path_str =
+                    self.get_or_init_secret_file(secret_id, secret_value_bytes.clone())?;
+                Ok(format!("file://{}", path_str))
+            }
             RefAsType::Unspecified => Err(SecretError::UnspecifiedRefType(secret_id)),
         }
     }
--- a/src/common/secret/src/secret_manager.rs
+++ b/src/common/secret/src/secret_manager.rs
@@ -166,7 +166,7 @@ impl LocalSecretManager {
             RefAsType::File => {
                 let path_str =
                     self.get_or_init_secret_file(secret_id, secret_value_bytes.clone())?;
-                Ok(format!("file://{}", path_str))
+                Ok(path_str)
             }
             RefAsType::Unspecified => Err(SecretError::UnspecifiedRefType(secret_id)),
         }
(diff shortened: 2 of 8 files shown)


yuhao-su:
AFAIK there are quite a few properties in rdkafka requires path instead of uri https://docs.confluent.io/platform/current/clients/librdkafka/html/md_CONFIGURATION.html

This can be a breaking change

> most sdk requires file://<absolute-path>

Can you elaborate?

tabVersion:
> AFAIK there are quite a few properties in rdkafka requires path instead of uri [docs.confluent.io/platform/current/clients/librdkafka/html/md_CONFIGURATION.html](https://docs.confluent.io/platform/current/clients/librdkafka/html/md_CONFIGURATION.html)
> 
> This can be a breaking change

make sense, kafka requires an absolute path.
But pulsar oauth requires an uri [ref](https://pulsar.apache.org/docs/2.6.1/security-oauth2/#client-credentials), and serves as an exception. 

Maybe a proper approach here is introducing `as uri` for secret.

hzxa21:
> > AFAIK there are quite a few properties in rdkafka requires path instead of uri [docs.confluent.io/platform/current/clients/librdkafka/html/md_CONFIGURATION.html](https://docs.confluent.io/platform/current/clients/librdkafka/html/md_CONFIGURATION.html)
> > This can be a breaking change
> 
> make sense, kafka requires an absolute path. But pulsar oauth requires an uri [ref](https://pulsar.apache.org/docs/2.6.1/security-oauth2/#client-credentials), and serves as an exception.
> 
> Maybe a proper approach here is introducing `as uri` for secret.

If we know pulsar is an exception, can we add the `file://` prefix to the properties that need file secrets only for pulsar instead of introducing a new syntax. IMO, it may cause confusion for users to choose between `AS FILE` and `AS FILE URI`.

tabVersion:
> > > AFAIK there are quite a few properties in rdkafka requires path instead of uri [docs.confluent.io/platform/current/clients/librdkafka/html/md_CONFIGURATION.html](https://docs.confluent.io/platform/current/clients/librdkafka/html/md_CONFIGURATION.html)
> > > This can be a breaking change
> > 
> > 
> > make sense, kafka requires an absolute path. But pulsar oauth requires an uri [ref](https://pulsar.apache.org/docs/2.6.1/security-oauth2/#client-credentials), and serves as an exception.
> > Maybe a proper approach here is introducing `as uri` for secret.
> 
> If we know pulsar is an exception, can we add the `file://` prefix to the properties that need file secrets only for pulsar instead of introducing a new syntax. IMO, it may cause confusion for users to choose between `AS FILE` and `AS FILE URI`.

after checking the existing codebase, the only one requires uri is MQTT

```
    /// Path to client's certificate file (PEM). Required for client authentication.
    /// Can be a file path under fs:// or a string with the certificate content.
    #[serde(rename = "tls.client_cert")]
    pub client_cert: Option<String>,

    /// Path to client's private key file (PEM). Required for client authentication.
    /// Can be a file path under fs:// or a string with the private key content.
    #[serde(rename = "tls.client_key")]
    pub client_key: Option<String>,
```

and that's because of the community contributor wrote the load cert process himself.

Let's see the pulsar issue as an exception and do hack here

PR #22230, opened 2025-06-14
fix(meta): injected project and merge nodes should have unique operator ids by sink into table

kwannoel:
…r ids

I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## What's changed and what's your intention?

A unique operator id is required for us to accurately track the profiling metrics for newly injected project and merge nodes inside explain analyze.

1. Traverse the union fragment to find the max operator id. Use that +1, +2 to assign operator ids to the injected project and merge nodes.
2. Refactor function name `visit_stream_node` -> `visit_stream_node_body` to reflect the actual usage.

The operator id just needs to be locally unique to the fragment, since it's globally namespaced by the fragment id.

## Checklist

- [ ] I have written necessary rustdoc comments.
- [ ]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  I have checked the [Release Timeline](https://github.com/risingwavelabs/rw-commits-history/blob/main/release_timeline.md) and [Currently Supported Versions](https://docs.risingwave.com/changelog/release-support-policy#support-end-dates-for-recent-releases) to determine which release branches I need to cherry-pick this PR into. 

## Documentation

- [ ]  My PR needs documentation updates. 

<details>
<summary><b>Release note</b></summary>

</details>

PR #22234, opened 2025-06-15
fix(frontend): tolerate duplicate operator ids and missing metrics

kwannoel:

The main goal of this PR is to tolerate project and merge nodes injected by sink into table. See https://github.com/risingwavelabs/risingwave/pull/22230. We ensure metrics collection and graph generation can tolerate this case for backwards compatibility. We also add more test coverage, just to validate that explain analyze shouldn't crash.

### Implementation notes

This PR improves the `EXPLAIN ANALYZE` functionality for stream jobs with the following changes:

1. Added a new e2e test for `EXPLAIN ANALYZE` that tests the command on tables, materialized views, and sinks with different duration settings.

2. Enhanced the stream job metrics collection to be more robust:
   - Made output metrics optional to handle cases where metrics are unavailable
   - Improved error handling when collecting metrics from executors
   - Fixed potential underflow issues when calculating metric deltas

3. Improved the operator graph construction:
   - Used node body discriminants for better operator identification. It avoids manual parsing of operator identity.
   - Fixed dispatcher node creation and connection in the graph
   - Enhanced actor ID collection and representation

4. Added a serial label to visibility mode tests to ensure proper test execution. We want to skip background_ddl tests.

5. Optimize graph traversal. Instead of traversing the stream node graph once for each actor, we can just do it once, for the first actor. This is because all actors in the fragment should have the same stream node graph. This makes our graph construction easier to reason about as well.

These changes make the `EXPLAIN ANALYZE` command more reliable and informative, especially when dealing with background DDL operations or when metrics collection might be incomplete.

## Checklist

## Documentation

### DIFF
--- a/ci/scripts/e2e-test-serial.sh
+++ b/ci/scripts/e2e-test-serial.sh
@@ -108,7 +108,7 @@ risedev slt -p 4566 -d dev './e2e_test/ddl/**/*.slt' --junit "batch-ddl-${profil
 risedev slt -p 4566 -d dev './e2e_test/background_ddl/basic.slt' --junit "batch-ddl-${profile}"
 
 if [[ "$mode" != "single-node" && "$mode" != "standalone" ]]; then
-  risedev slt -p 4566 -d dev './e2e_test/visibility_mode/*.slt' --junit "batch-${profile}"
+  risedev slt -p 4566 -d dev './e2e_test/visibility_mode/*.slt' --junit "batch-${profile}" --label "serial"
 fi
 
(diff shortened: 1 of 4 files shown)


kwannoel:
* **#22234** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22234?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22234?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #22225, opened 2025-06-13
fix(sqlsmith): use `ObjectName` for column name to correctly distinguish qualified name

BugenZhao:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## What's changed and what's your intention?

#22085 is a fix to convert `real_value: String` back to `ast::Ident` properly, by quoting the value with `"` if there's any upper case or special character.

However, in sqlsmith, we previously relied on the incorrect behavior by directly interpolating a `.` into the name of `Column` to represent the qualified name. As a result, the fix failed the fuzzing test in main-cron.

This PR fixed the handling of column name in sqlsmith, by storing an `ObjectName` instead.

```
[2m2025-06-12T16:36:47.783799Z[0m [32m INFO[0m [2mrisingwave_sqlsmith::test_runners::utils[0m[2m:[0m [EXECUTING CREATE MVIEW]: CREATE MATERIALIZED VIEW m1 AS WITH with_0 AS (WITH with_1 AS (SELECT tumble_2.auction AS col_0, tumble_2.date_time AS col_1, ((INT '630') # (SMALLINT '994')) AS col_2, tumble_2.bidder AS col_3 FROM tumble(bid, "bid.date_time", INTERVAL '14') AS tumble_2 WHERE false GROUP BY tumble_2.auction, tumble_2.date_time, tumble_2.bidder HAVING false) SELECT TIME '16:36:47' AS col_0, CAST('cIsP9XEFBc' AS CHARACTER VARYING) AS col_1 FROM with_1) SELECT CAST('kP4CxYgYlr' AS CHARACTER VARYING) AS col_0, (575) AS col_1, ((INTERVAL '-2') > TIME '16:35:16') AS col_2, CAST(NULL AS STRUCT<a TIMESTAMP, b REAL, c DOUBLE>) AS col_3 FROM with_0 WHERE true
[2m2025-06-12T16:36:47.784507Z[0m [32m INFO[0m [2mrisingwave_sqlsmith::test_runners::utils[0m[2m:[0m [UNEXPECTED ERROR]: Error {
    kind: Db,
    cause: Some(
        DbError {
            severity: "ERROR",
            parsed_severity: None,
            code: SqlState(
                EXX000,
            ),
[...]

PR #22235, opened 2025-06-15
refactor(sqlsmith): change config to control weight and features

### LINKED ISSUE #3936: Sqlsmith: Decouple weights from generators

cardigan1008:

Fix #3936 

Previously, we encountered an issue with the inconsistent result of `should_generate()` when checking the agg feature:  `can_agg` is a constant that should be consistent during one single generation of a complete sql statement. #22026 replaces `can_agg` with `should_generate()`. This breaks the consistency as `should_generate()` is called in parts of sql generation. 

This PR reverts the refactor by restoring `can_agg` and using `should_generate()` to initialize it. This preserves the original logic of context and allows us to control the weight as well. 

The inconsistency also brings issues when generating complex features like EOWC. For example, we need to first filter out watermark columns by calling `should_generate()` 1st time, where the result might be false. Then we need to decide whether we use EOWC keyword by calling it 2nd time, where the result is true. The final sql is thus wrong with non-watermark columns using in EOWC.

To address this, we now split configuration into two categories: Syntax and Feature.

- **Syntax items** are lightweight and local to SQL structure (e.g., agg, join, where), controlled by weights and decided once per query generation context.

- **Feature items** are high-level switches that may affect multiple syntax elements or query semantics (e.g., EOWC). These are managed explicitly at the higher level.

Syntax is configured in yaml files by setting its weight. Features is passed with cli options (e.g., `--enable eowc`). 

## Checklist

## Documentation

<details>

</details>


cardigan1008:
* **#22235** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22235?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22235?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#22225** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22225?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

cardigan1008:
Do we need to update doc in `src/tests/sqlsmith/README.md` about the `--enable xx` option?

kwannoel:
> Do we need to update doc in `src/tests/sqlsmith/README.md` about the `--enable xx` option?

We can add it once it's also added to CI and stabilized, I suppose before that the interface might fluctuate, and you will have to keep updating the README.

cardigan1008:
> I'd imagine some features are exclusive to others. For instance joins and eowc. We can either:
> 
> 1. Keep it simple, and just accept a single feature at a time.
> 2. Make it more robust, make sure conflicting features are not enabled at the same time.

Currently, we define joins as a Syntax item and eowc as a Feature item. Syntax items are controlled solely by their weights—disabling them requires setting their weight to 0. Under the first strategy, setting the weight of joins to 0 when eowc is enabled would prevent the generation of joins across all tables, even for those that don't use eowc.

In contrast, the second strategy allows finer control. We can selectively disable joins only when eowc is active—for example, using logic like `if !self.generate(Feature::Eowc) { gen_join() }`. This way, tables not using eowc can still generate joins, avoiding unnecessary limitations.

Issue #22240, opened 2025-06-16
support `SECRET` as function argument


stdrc:

hzxa21:
Assign to @tabVersion based on feature owner.

PR #22242, opened 2025-06-16
feat(expr): add `openai_embedding` function

stdrc:

According to comments of #21929, we decided to implement AI functions as normal expressions with `prebuilt` "context", which do simplifies the code. However, we then have two more things to do to achieve a better UX:

1. #22240
2. #22254

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/expr/impl/src/scalar/ai_model.rs
+++ b/src/expr/impl/src/scalar/ai_model.rs
@@ -1,0 +1,90 @@
+// Copyright 2025 RisingWave Labs
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use async_openai::Client;
+use async_openai::config::OpenAIConfig;
+use async_openai::types::{CreateEmbeddingRequestArgs, EmbeddingInput};
+use risingwave_common::array::{F32Array, ListValue};
+use risingwave_common::types::F32;
+use risingwave_expr::{ExprError, Result, function};
+use thiserror_ext::AsReport;
+
+/// `OpenAI` embedding context that holds the client and model configuration
+#[derive(Debug)]
+pub struct OpenAiEmbeddingContext {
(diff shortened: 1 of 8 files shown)


stdrc:
* **#22257** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22257?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#22242** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22242?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22242?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/expr/impl/src/scalar/ai_model.rs
```diff
@@ -0,0 +1,90 @@
...
+    context: &OpenAiEmbeddingContext,
+) -> Result<Option<ListValue>> {
+    if text.is_empty() {
+        return Ok(None);
+    }
+
+    // Call the OpenAI API for embedding
+    let embedding_request = CreateEmbeddingRequestArgs::default()
+        .model(&context.model)
+        .input(EmbeddingInput::String(text.to_owned()))
```
wenym1: Is there a way to leverage the `EmbeddingInput::StringArray` for us to handle multiple texts in a batch?

src/expr/impl/src/scalar/ai_model.rs
```diff
@@ -0,0 +1,90 @@
...
+/// `OpenAI` embedding function with prebuild optimization for constant `api_key` and `model`
+#[function(
+    "openai_embedding(varchar, varchar, varchar) -> float4[]",
+    prebuild = "OpenAiEmbeddingContext::from_config($0, $1)?"
+)]
+async fn openai_embedding(
+    text: &str,
+    context: &OpenAiEmbeddingContext,
+) -> Result<Option<ListValue>> {
+    if text.is_empty() {
```
wenym1: Will async openai return error when seeing empty text? If not, maybe we can just not treat empty text as a special case, and send a request to get the embedding anyway. Returning None may possibly break the non-null constraint required somewhere downstream.

hzxa21 (approved):

src/expr/impl/src/scalar/ai_model.rs
BugenZhao: +1 for this. As a result we may need to manually impl `Expr`.

src/expr/impl/src/scalar/ai_model.rs
stdrc: OpenAI embedding API doesn't accept empty text.
截屏2025-06-17 14 59 56

Issue #21801, opened 2025-05-09
mysql cdc source: schema change while RW is offline fails the connector permanently


BugenZhao:
### Describe the bug

After creating a MySQL CDC source/table, if we take RW offline, make schema changes on the MySQL side, and then restart RW, we will only receive the latest table schema. This results in a failure to parse the (unconsumed) binlog with the original schema, causing the connector to fail permanently.

### Error message/log

```text
2025-05-09T08:11:29.637686252Z ERROR risingwave_connector_node: Error processing after of row in t because it's different column size with internal schema size 3, but after size 2, restart connector with schema recovery mode. thread="blc-127.0.0.1:8306" class="io.debezium.connector.mysql.MySqlStreamingChangeEventSource"
2025-05-09T08:11:29.639607293Z ERROR risingwave_connector_node: Error during binlog processing. Last offset stored = {transaction_id=file=binlog.000003,pos=236, ts_sec=1746778289, file=binlog.000003, pos=236, server_id=1, event=2}, binlog reader near position = binlog.000003/314 thread="blc-127.0.0.1:8306" class="io.debezium.connector.mysql.MySqlStreamingChangeEventSource"
[...]

PR #22255, opened 2025-06-17
fix(mysql-cdc): introduce `OpendalSchemaHistory` to resolve the schema change loss issue when debezium offline

wcy-fdu:

This PR introduces `OpendalSchemaHistory` to address the issue of Debezium being unable to apply upstream MySQL schema changes when offline.

### Problem Description
The schema history records all DDL operations (such as create/drop table, alter table, etc.) for all tables in **one** upstream MySQL database. Previously, we used `MemorySchemaHistory` to store the MySQL schema history, which meant that once Debezium (and thus RisingWave) went offline, this information would be lost. Therefore, we had to set `recovery.mode` to `recovery`.

About recovery.mode
`no_data`: This means that when no offset is provided, a snapshot of the schema is taken. If an offset is present, it directly restores from the schema history.
`recovery`: This assumes that the schema history is lost or corrupted, and it restores the schema history from the latest state.
`custom`: This is a snapshot mode we customized for compatibility. Simply put,
- if the schema history exists → no schema snapshot is taken, restore directly from the history(That is, no_data).
- If the schema history does not exist → a schema snapshot is taken and the schema is rebuilt(That is, recovery).

Therefore, to resolve this issue: We need
1. a persistent schema history
2. `recovery.mode` set to `custom` .

### Design
Due to Debezium's supported persistent schema history options being limited to File/S3 and Azure Blob, which do not fully meet the needs of RisingWave Cloud, we chose to introduce OpendalSchemaHistory to address this issue. The overall approach is to record the schema history in the object store used by RisingWave's underlying system, but in a separate folder rather than within the folder where the SST files are located.

Additionally, Debezium's implementation of S3 schema history has some potential issues:
1. The schema history file records DDL at the database level. If there are multiple MySQL instances upstream or multiple databases within a single MySQL instance, all these records will be written to the same file, which can lead to write conflicts on S3.
2. In Debezium's implementation, before writing a new record, all historical data is loaded into memory, and then the new record is appended at the end. Since the schema history continuously grows, this memory usage will increase over time, potentially leading to OOM.

To address these two issues, we implemented the following solutions:
1. Use source_id as the directory for the persistent schema history, with each source corresponding to its own directory.
2. Instead of writing to a single file, control the number of records in each file. If the threshold is exceeded, a new file is created. When reading, we do not load all data into memory at once; instead, we read the size of each file and locate the required file based on the index.

### Todo

- [ ] Split `jvm_runtime` into other crates to solve circular dependency issues. (refer to https://github.com/risingwavelabs/risingwave/pull/17552)
- [ ] Add enough tests for this scenario

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/ci/scripts/e2e-source-mysql-offline-schema-change.sh
+++ b/ci/scripts/e2e-source-mysql-offline-schema-change.sh
@@ -135,11 +135,12 @@ else
 fi
 
 echo "--- Verify t has 3 columns"
-COL_COUNT=$(risedev psql -t -c "SELECT COUNT(*) FROM information_schema.columns WHERE table_name = 't';" 2>&1 | xargs)
-if [ "$COL_COUNT" = "3" ]; then
+OUTPUT=$(risedev psql -t -c "SELECT CASE WHEN COUNT(*) = 3 THEN 'OK' ELSE 'FAIL' END FROM information_schema.columns WHERE table_name = 't';" 2>&1)
+if echo "$OUTPUT" | grep -q "OK"; then
     echo "✓ PASS: t has 3 columns after schema change"
 else
-    echo "✗ FAIL: t has $COL_COUNT columns, expected 3"
+    echo "✗ FAIL: t does not have 3 columns"
+    echo "Debug output: $OUTPUT"
     risedev psql -c "SELECT column_name FROM information_schema.columns WHERE table_name = 't';"
     exit 1
 fi
@@ -157,11 +158,12 @@ else
 fi
 
 echo "--- Verify t1 has 3 columns"
-COL_COUNT=$(risedev psql -t -c "SELECT COUNT(*) FROM information_schema.columns WHERE table_name = 't1';" 2>&1 | xargs)
-if [ "$COL_COUNT" = "3" ]; then
+OUTPUT=$(risedev psql -t -c "SELECT CASE WHEN COUNT(*) = 3 THEN 'OK' ELSE 'FAIL' END FROM information_schema.columns WHERE table_name = 't1';" 2>&1)
+if echo "$OUTPUT" | grep -q "OK"; then
     echo "✓ PASS: t1 has 3 columns after schema change"
 else
-    echo "✗ FAIL: t1 has $COL_COUNT columns, expected 3"
+    echo "✗ FAIL: t1 does not have 3 columns"
+    echo "Debug output: $OUTPUT"
     risedev psql -c "SELECT column_name FROM information_schema.columns WHERE table_name = 't1';"
     exit 1
 fi


wcy-fdu:
New changes:
- Use hummock object store
- per source per schema history directory to avoid write conflict
- Set the size of each file (batch strategy) to avoid reading all history records into memory each time.

wcy-fdu:
## Implementation detail && Workflow
### Startup Phase (`doStart`)
**1.File Scanning**

1. Call listObject(objectDir) to get all files in the directory.
2. Filter files matching the pattern schema_history_*.dat.
3. Sort by timestamp (using extractTimestampFromFileName).

**2.  Lazy Load List Initialization**

1. Create an instance of LazyHistoryRecordList.
2. Iterate over each file(only count records, do not load content):
- -  Call getObject(file) to read the file content.
- -  Call toHistoryRecords(data) to deserialize.
- - Count records for each file and update the recordsPerFileIndex array.
3. Assign the lazy load list to this.records.
### Writing Phase (`doStoreRecord`)
1. Find Latest File
- Re-call listObject(objectDir) to get the latest file list.
- Filter and sort to find the file with the largest timestamp.
2. Determine Append or Create New File
3. File Naming Rule: `{source_id}/schema_history_{timestamp}.dat`:
image

###  Reading Phase (`LazyHistoryRecordList.get`)
1.  Locate File

- Find the corresponding file index using the requested index from recordsPerFileIndex array.
- Calculate the relative index within that file.

2. Read from cache or object store


### Stop Phase (`doStop`)
when dropping a source, delete all the schema history file under this source's directory.

java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/DbzConnectorConfig.java
```diff
@@ -149,11 +149,12 @@ public DbzConnectorConfig(
             if (isCdcBackfill) {
                 // disable snapshot locking at all
                 mysqlProps.setProperty("snapshot.locking.mode", "none");
-
+                mysqlProps.setProperty(
+                        "schema.history.internal.source.id", String.valueOf(sourceId));
```
hzxa21: Let's store this in a const and refer to the const everywhere we need it.

java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/DbzConnectorConfig.java
```diff
@@ -149,11 +149,12 @@ public DbzConnectorConfig(
...
                 // If cdc backfill enabled, the source only emit incremental changes, so we must
                 // rewind to the given offset and continue binlog reading from there
                 if (null != startOffset && !startOffset.isBlank()) {
-                    mysqlProps.setProperty("snapshot.mode", "recovery");
+                    mysqlProps.setProperty("snapshot.mode", "no_data");
```
hzxa21: Leave a note here: we don't need to do it in this PR but we may need to investigate all the source that requires schema history topic and see whether we need to change the `snapshot.mode` for them (e.g. sqlserver).

src/jni_core/src/opendal_schema_history.rs
```diff
@@ -0,0 +1,183 @@
...
+    }
+}
+
+async fn get_object_store() -> Arc<ObjectStoreImpl> {
+    if let Some(store) = OBJECT_STORE_INSTANCE.get() {
+        store.clone()
+    } else {
+        let hummock_url = STATE_STORE_URL.get().unwrap();
+        let object_store = build_remote_object_store(
+            hummock_url.strip_prefix("hummock+").unwrap(),
```
hzxa21: This assumes the state store must be hummock, which is not always true, especially in CI and test. Can we fall back to memory schema history if hummock is not used?

src/jni_core/src/opendal_schema_history.rs
```diff
@@ -0,0 +1,183 @@
...
+use risingwave_object_store::object::object_metrics::ObjectStoreMetrics;
+use risingwave_object_store::object::{ObjectStoreImpl, build_remote_object_store};
+
+use crate::{EnvParam, JAVA_BINDING_ASYNC_RUNTIME, execute_and_catch, to_guarded_slice};
+
+static OBJECT_STORE_INSTANCE: OnceLock<Arc<ObjectStoreImpl>> = OnceLock::new();
+
+// schema history is internal state, all data is stored under the DATA_DIRECTORY directory.
+fn prepend_data_directory(path: &str) -> String {
+    let data_dir = DATA_DIRECTORY.get().map(|s| s.as_str()).unwrap_or("");
```
hzxa21: default to `""` looks very dangerous because buckets are shared in cloud environment, which can cause data to be overwritten. We shoulld add a safe guard to make sure data dir is set instead of using default.

src/jni_core/src/opendal_schema_history.rs
```diff
@@ -0,0 +1,183 @@
...
+}
+
+#[unsafe(no_mangle)]
+pub extern "system" fn Java_com_risingwave_java_binding_Binding_putObject(
+    env: EnvParam<'_>,
+    object_name: JString<'_>,
+    data: JByteArray<'_>,
+) {
+    execute_and_catch(env, move |env| {
+        let object_name = env.get_string(&object_name)?;
```
hzxa21: The interface is too generic. Can we add a safe guard to check the object suffix and make sure we are not overwriting hummock data on object store? Maybe we can add a suffix parameter and a white list.

hzxa21 (commented):
Leave some early comments.

java/connector-node/risingwave-source-cdc/src/main/java/com/risingwave/connector/cdc/debezium/internal/OpendalSchemaHistory.java
```diff
@@ -0,0 +1,354 @@
...
+            SchemaHistoryListener listener,
+            boolean useCatalogBeforeSchema) {
+        super.configure(config, comparator, listener, useCatalogBeforeSchema);
+        if (!config.validateAndRecord(ALL_FIELDS, LOGGER::error)) {
+            throw new DebeziumException(
+                    "Error configuring an instance of "
+                            + getClass().getSimpleName()
+                            + "; check the logs for details");
+        }
+        sourceId = config.getString("schema.history.internal.source.id");
```
hzxa21: Use the static const

java/connector-node/risingwave-source-cdc/src/main/java/com/risingwave/connector/cdc/debezium/internal/OpendalSchemaHistory.java
```diff
@@ -0,0 +1,354 @@
...
+        if (sourceId == null || sourceId.isEmpty()) {
```
hzxa21, line 95: In which case will we hit this branch? If there are more than one sources hitting this branch, it will cause path conflict. I think we better throw an error in this case.

java/connector-node/risingwave-source-cdc/src/main/java/com/risingwave/connector/cdc/debezium/internal/OpendalSchemaHistory.java
```diff
@@ -0,0 +1,354 @@
...
+            sourceId = "default_source";
+            config = config.edit().with("source_id", sourceId).build();
+        }
+        objectDir = String.format("mysql-cdc-schema-history-source-%s", sourceId);
```
hzxa21: The implmenetation of this file has nothing to do with `mysql` so let's rename it to `rw-cdc-source-schema-history-...`

java/connector-node/risingwave-source-cdc/src/main/java/com/risingwave/connector/cdc/debezium/internal/OpendalSchemaHistory.java
```diff
@@ -0,0 +1,354 @@
...
+        objectName = String.format("%s/schema_history.dat", objectDir);
+        String maxRecordsStr = config.getString(MAX_RECORDS_PER_FILE_CONFIG);
+        if (maxRecordsStr != null) {
```
hzxa21, line 101: What if `maxRecordsStr == null`?

java/connector-node/risingwave-source-cdc/src/main/java/com/risingwave/connector/cdc/debezium/internal/OpendalSchemaHistory.java
```diff
@@ -0,0 +1,354 @@
...
+            Pattern.compile("schema_history_(\\d+)\\.dat");
+
+    @Override
+    public void configure(
+            Configuration config,
+            HistoryRecordComparator comparator,
+            SchemaHistoryListener listener,
+            boolean useCatalogBeforeSchema) {
+        super.configure(config, comparator, listener, useCatalogBeforeSchema);
+        if (!config.validateAndRecord(ALL_FIELDS, LOGGER::error)) {
```
hzxa21, line 88: Looks like we don't override `ALL_FIELDS`. Will it cause an error given that we add our own field `SOURCE_ID` to the config?

java/connector-node/risingwave-source-cdc/src/main/java/com/risingwave/connector/cdc/debezium/internal/OpendalSchemaHistory.java
```diff
@@ -0,0 +1,354 @@
...
+            return Long.parseLong(m.group(1));
+        }
+        return 0L;
+    }
+
+    /**
+     * A custom List implementation that loads schema history records from Opendal on demand. This
+     * avoids loading all history records into memory at once.
+     */
+    private class LazyHistoryRecordList extends AbstractList<HistoryRecord> {
```
hzxa21: I think extending `AbstractList` is an overkill here because in order to make `OpendalSchemaHistory` work, we only need to implement the [reocverRecords](https://github.com/debezium/debezium/blob/4159397bd40dd5d0c8efe2b97ba61ba8eb220b58/debezium-core/src/main/java/io/debezium/relational/history/AbstractSchemaHistory.java#L162) method. The method provided by [`AbstractList`](https://docs.oracle.com/javase/8/docs/api/java/util/AbstractList.html) is more powerful and will make the implementation less efficient. For example, `add`, `remove`, `set` , `sublist` ... are not needed.

java/connector-node/risingwave-source-cdc/src/main/java/com/risingwave/connector/cdc/debezium/internal/OpendalSchemaHistory.java
```diff
@@ -0,0 +1,354 @@
...
+    @Override
+    protected void doPreStoreRecord(HistoryRecord record) {}
+
+    @Override
+    protected void doStoreRecord(HistoryRecord record) {
+        LOGGER.info("Storing new schema history record.");
+        try {
+            // 1. Find the latest schema_history_*.dat file
+            List<String> files = new ArrayList<>();
+            String[] fileArray = listObject(objectDir);
```
hzxa21: Listing all objects on every `doStoreRecord` is inefficient and unneccessary. We can cache the content and the name of the latest file in memory instead of fetching it everytime.

java/connector-node/risingwave-source-cdc/src/main/java/com/risingwave/connector/cdc/debezium/internal/OpendalSchemaHistory.java
```diff
@@ -0,0 +1,354 @@
...
+        String type = getObjectStoreType();
+        LOGGER.info("Using hummock object store {} to store database history.", type);
+    }
+
+    @Override
+    public String toString() {
+        return getObjectStoreType();
+    }
+
+    private void flushBufferToNewFile() {
```
hzxa21: This method is unused.

java/connector-node/risingwave-source-cdc/src/main/java/com/risingwave/connector/cdc/debezium/internal/OpendalSchemaHistory.java
```diff
@@ -0,0 +1,354 @@
...
+        if (buffer.isEmpty()) {
+            return;
+        }
+        String fileName =
+                String.format("%s/schema_history_%d.dat", objectDir, System.currentTimeMillis());
```
hzxa21: This assumes that there is the minimal gap between two writes >= 1ms. We cannot guarantee that this is always true. Once it is not true, we will overwrite the wrong file. How about using a self maintained sequence number instead of using timestamp.

java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/DbzConnectorConfig.java
wcy-fdu: Agree, let's take eyes on other cdc after this pr merged.

java/connector-node/risingwave-source-cdc/src/main/java/com/risingwave/connector/cdc/debezium/internal/OpendalSchemaHistory.java
wcy-fdu: This is just defensive code; in fact, there will definitely be a source_id. I agree to throw an error.

java/connector-node/risingwave-source-cdc/src/main/java/com/risingwave/connector/cdc/debezium/internal/OpendalSchemaHistory.java
wcy-fdu: It's normal for `maxRecordsStr` to be null, indicating that the user hasn't explicitly configured this optional parameter. In this case, the code uses the default value of 2048, which is reasonable behavior. The if (maxRecordsStr != null) check is necessary to distinguish between "not configured" and "configured but with an invalid value."

java/connector-node/risingwave-source-cdc/src/main/java/com/risingwave/connector/cdc/debezium/internal/OpendalSchemaHistory.java
wcy-fdu: >We cannot guarantee that this is always true.


The gap here actually refers to the interval after finishing writing one file (reaching the record count threshold, which is 2048 by default) before starting to write a new file. A gap of less than 1 ms indicates that more than 2048 records were written within 1 ms, and most of the records here are DDL operations. 
I personally have no issues with maintaining the sequence  number; I'm just indicating that the corner case mentioned above should not occur.

java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/DbzConnectorConfig.java
```diff
@@ -168,7 +173,7 @@ public DbzConnectorConfig(
                     // 'snapshot.mode=recovery' must be configured if binlog offset is
```
hzxa21: documentation needs to be changed as well

java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/DbzConnectorConfig.java
```diff
@@ -324,6 +329,18 @@ public DbzConnectorConfig(
...
+        // Special handling for test environment: override schema history for MySQL tests
+        if (source == SourceTypeE.MYSQL && otherProps.containsKey("schema.history.internal")) {
+            String testSchemaHistory = otherProps.get("schema.history.internal");
+            if (testSchemaHistory.contains("MemorySchemaHistory")) {
+                LOG.info(
+                        "Test environment detected: overriding schema history to {}",
+                        testSchemaHistory);
+                dbzProps.put("schema.history.internal", testSchemaHistory);
+            }
+        }
```
hzxa21: This looks strange to me. Why do we need to do it?

java/connector-node/risingwave-source-cdc/pom.xml
```diff
@@ -49,7 +49,15 @@
...
-
+                <dependency>
+            <groupId>com.risingwave</groupId>
+            <artifactId>java-binding</artifactId>
+        </dependency>
+       <dependency>
+            <groupId>io.debezium</groupId>
+            <artifactId>debezium-storage-s3</artifactId>
+            <version>3.2.0.Alpha1</version>
+        </dependency>
```
hzxa21: Why do we need to depend on `debezum-storage-s3`?

java/connector-node/risingwave-source-cdc-test/src/test/java/com/risingwave/connector/source/SourceTestClient.java
```diff
@@ -166,8 +166,16 @@ protected Iterator<ConnectorServiceProto.GetEventStreamResponse> getEventStreamS
...
-                        .putProperties("server.id", "1") // mysql only
-                        .build();
+                        .putProperties("server.id", "1"); // mysql only
+
+        // Override schema history for MySQL tests to use MemorySchemaHistory
+        if (sourceType == ConnectorServiceProto.SourceType.MYSQL) {
+            reqBuilder.putProperties(
+                    "debezium.schema.history.internal",
+                    "io.debezium.relational.history.MemorySchemaHistory");
+        }
```
hzxa21: ditto: looks strange to me

java/connector-node/risingwave-connector-service/src/main/resources/mysql.properties
```diff
@@ -12,6 +12,10 @@ table.include.list=${database.name}.${table.name:-*}
 # only read table schema of the captured tables in the specified database
 schema.history.internal.store.only.captured.tables.ddl=true
 schema.history.internal.store.only.captured.databases.ddl=true
+
+# For MySQL, use OpendalSchemaHistory to avoid data loss during downtime.
+schema.history.internal=com.risingwave.connector.cdc.debezium.internal.OpendalSchemaHistory
```
hzxa21, line 17: Will this affect the already created mysql source prior to this PR?

java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/DbzConnectorConfig.java
```diff
@@ -168,7 +173,7 @@ public DbzConnectorConfig(
                     // 'snapshot.mode=recovery' must be configured if binlog offset is
                     // specified. It only snapshots the schemas, not the data, and continue binlog
                     // reading from the specified offset
-                    mysqlProps.setProperty("snapshot.mode", "recovery");
+                    mysqlProps.setProperty("snapshot.mode", "no_data");
```
hzxa21: Just realize that the change here will affect mysql source created prior to this PR. Will it be an issue? I think we better test that.

src/compute/src/server.rs
```diff
@@ -206,6 +207,11 @@ pub async fn compute_node_serve(
     };
 
     LicenseManager::get().refresh(system_params.license_key());
+    // Store the state_store_url in a static OnceLock for later use in JNI crate
+    let _ = STATE_STORE_URL.set(state_store_url.to_owned());
+
+    let _ = DATA_DIRECTORY.set(data_directory.to_owned());
```
hzxa21: To be extra safe, let's check the return value andif the variable is set, assert that the value is the same as `state_store_url` and `data_directory`

src/jni_core/src/jvm_runtime.rs
```diff
@@ -25,6 +25,13 @@ use risingwave_common::util::resource_util::memory::system_memory_available_byte
 use thiserror_ext::AsReport;
 use tracing::error;
 
+use crate::opendal_schema_history::{
```
hzxa21: I think you can write the following in lib.rs:
```
use opendal_schema_history::*
```

then you don't need to write the manual use here.

example: https://github.com/risingwavelabs/risingwave/blob/76b83f9253657d17993e4abc1faa66fdcbdb012b/src/jni_core/src/lib.rs#L63

src/jni_core/src/lib.rs
```diff
@@ -19,6 +19,7 @@
 
 pub mod jvm_runtime;
 mod macros;
+pub mod opendal_schema_history;
```
hzxa21: doesn't need to be pub if you write `use opendal_schema_history::*`

src/jni_core/src/opendal_schema_history.rs
```diff
@@ -0,0 +1,270 @@
...
+}
+
+async fn get_object_store() -> Arc<ObjectStoreImpl> {
+    if let Some(store) = OBJECT_STORE_INSTANCE.get() {
+        store.clone()
+    } else {
+        let hummock_url = STATE_STORE_URL.get().unwrap();
+        let object_store = build_remote_object_store(
+            hummock_url.strip_prefix("hummock+").unwrap_or("memory"),
+            Arc::new(ObjectStoreMetrics::unused()),
```
hzxa21: I checked the implementation of `ObjectStoreMetrics::unused()`. It is actually not `unused` but use the `GLOBAL_OBJECT_STORE_METRICS`, like [here](https://github.com/risingwavelabs/risingwave/blob/76b83f9253657d17993e4abc1faa66fdcbdb012b/src/compute/src/server.rs#L193). I suggest we use `Arc::new(GLOBAL_OBJECT_STORE_METRICS.clone()` directly here to make it clearer.

src/jni_core/src/opendal_schema_history.rs
```diff
@@ -0,0 +1,270 @@
...
+            "rw-cdc-schema-history",
+            Arc::new(ObjectStoreConfig::default()),
+        )
+        .await;
+        let arc = Arc::new(object_store);
+        let _ = OBJECT_STORE_INSTANCE.set(arc.clone());
+        arc
+    }
+}
```
hzxa21, line 75: There is a read-write race condition here. You should use `get_or_init` instead of `get` then `set`.

src/jni_core/src/opendal_schema_history.rs
```diff
@@ -0,0 +1,270 @@
...
+
+// schema history is internal state, all data is stored under the DATA_DIRECTORY directory.
+fn prepend_data_directory(path: &str) -> String {
+    let data_dir = DATA_DIRECTORY.get().map(|s| s.as_str()).unwrap_or("");
+    if data_dir.is_empty() {
+        // In cloud environments, using empty data directory can cause data conflicts
+        // between different instances sharing the same bucket. We should ensure
+        // data_directory is properly set to avoid data overwrites.
+        panic!(
+            "DATA_DIRECTORY is not set. This is dangerous in cloud environments as it can cause data conflicts between multiple instances sharing the same bucket. Please ensure data_directory is properly configured."
```
hzxa21: nits: you can just use `expect` in L48:
```
let data_dir = DATA_DIRECTORY.get().map(|s| s.as_str()).expect("DATA_DIRECTORY is not set. This is...");
```

src/jni_core/src/opendal_schema_history.rs
```diff
@@ -0,0 +1,270 @@
...
+        );
+    }
+    if path.starts_with(data_dir) {
+        path.to_owned()
+    } else if data_dir.ends_with('/') || path.starts_with('/') {
+        format!("{}{}", data_dir, path)
+    } else {
+        format!("{}/{}", data_dir, path)
+    }
```
hzxa21: Why are there three cases here? The path is constructed by us and we should make it correct instead of handling multiple cases here. I suggest we always pass relative path to `prepend_data_directory` and construct the absolute path inside this method to make it correct:
```
prepend_data_directory(<path>)

output: <data_dir>/cdc/<path>

```

src/jni_core/src/opendal_schema_history.rs
hzxa21: side note (no need to address in this PR): is it possible to differentiate the object store metrics from opendal scheama history and hummock?

src/jni_core/src/opendal_schema_history.rs
```diff
@@ -0,0 +1,270 @@
...
+        let object_name: Cow<'_, str> = (&object_name).into();
+
+        // Security check: validate file extension before any operation
+        if let Err(error_msg) = validate_dat_file_extension(&object_name) {
+            tracing::error!(
+                "putObject security validation failed, skipping operation: {}",
+                error_msg
+            );
+            return Ok(()); // Skip this operation
+        }
```
hzxa21: You can put this in `prepend_data_directory` to avoid duplicate codes.

src/jni_core/src/opendal_schema_history.rs
```diff
@@ -0,0 +1,270 @@
...
+        let object_name: Cow<'_, str> = (&object_name).into();
+
+        // Security check: validate file extension before any operation
+        if let Err(error_msg) = validate_dat_file_extension(&object_name) {
+            tracing::error!(
+                "getObject security validation failed, returning empty result: {}",
+                error_msg
+            );
+            return Ok(env.byte_array_from_slice(&[])?); // Return empty byte array
+        }
```
hzxa21: ditto

src/jni_core/src/opendal_schema_history.rs
```diff
@@ -0,0 +1,270 @@
...
+        // Note: listObject operates on directories, individual file security is checked in putObject/getObject
+
+        let dir = prepend_data_directory(&dir);
+
+        let files: Vec<String> = JAVA_BINDING_ASYNC_RUNTIME.block_on(async {
+            let object_store = get_object_store().await;
+            let mut file_names = Vec::new();
+            let mut stream = match object_store.list(&dir, None, None).await {
+                Ok(s) => s,
+                Err(_) => return file_names,
```
hzxa21: Better log the errors

src/jni_core/src/opendal_schema_history.rs
```diff
@@ -0,0 +1,270 @@
...
+            );
+            return Ok(env.byte_array_from_slice(&[])?); // Return empty byte array
+        }
+
+        let object_name = prepend_data_directory(&object_name);
+        let result = JAVA_BINDING_ASYNC_RUNTIME.block_on(async {
+            let object_store = get_object_store().await;
+            match object_store.read(&object_name, ..).await {
+                Ok(data) => data,
+                Err(_) => Bytes::new(),
```
hzxa21: Better log the errors

src/jni_core/src/opendal_schema_history.rs
```diff
@@ -0,0 +1,270 @@
...
+        let object_name = prepend_data_directory(&object_name);
+
+        let data_guard = to_guarded_slice(&data, env)?;
+        let data: Vec<u8> = data_guard.slice.to_vec();
+        JAVA_BINDING_ASYNC_RUNTIME
+            .block_on(async {
+                let object_store = get_object_store().await;
+                object_store.upload(&object_name, data.into()).await
+            })
+            .unwrap();
```
hzxa21: It will cause panic whne meeting object store error here. Better fail the call and log the error here.

src/jni_core/src/opendal_schema_history.rs
```diff
@@ -0,0 +1,270 @@
...
+                            relative_path
+                        };
+
+                        if validate_dat_file_extension(relative_path).is_ok() {
+                            file_names.push(obj.key);
+                        } else {
+                            tracing::error!("Filtering out non-.dat file from list: {}", obj.key);
+                        }
+                    }
+                    Err(_) => continue,
```
hzxa21: Why do we continue here? I think it is better to log the errors and return the error.

src/jni_core/src/opendal_schema_history.rs
```diff
@@ -0,0 +1,270 @@
...
+        // Note: deleteObjects operates on directories, individual file security is checked in putObject/getObject
+
+        let dir = prepend_data_directory(&dir);
+
+        JAVA_BINDING_ASYNC_RUNTIME.block_on(async {
+            let object_store = get_object_store().await;
+            let mut keys = Vec::new();
+            let mut stream = match object_store.list(&dir, None, None).await {
+                Ok(s) => s,
+                Err(_) => return,
```
hzxa21: Better log the errors

src/jni_core/src/opendal_schema_history.rs
```diff
@@ -0,0 +1,270 @@
...
+            };
+            use futures::StreamExt;
+            while let Some(obj) = stream.next().await {
+                if let Ok(obj) = obj {
```
hzxa21: Missing error handling

hzxa21 (commented):
Left some comments first.

wcy-fdu:
> Left some comments first.

Sorry this pr is not ready yet, I will request review after CI pass and code cleaning.

src/jni_core/src/macros.rs
```diff
@@ -411,6 +411,16 @@ macro_rules! for_all_plain_native_methods {
 
                 static native boolean iteratorNext(long pointer);
 
+                public static native void putObject(String objectName, byte[] data);
```
zwang28, line 416: Can we just use opendal java instead?

java/connector-node/risingwave-connector-service/src/main/resources/mysql.properties
zwang28: To ensure OpendalSchemaHistory is compatible with existing MySQL sources, what about initializing the object store within `OpendalSchemaHistory::exists` for them so that `exists` always returns true, given that [exists](https://github.com/debezium/debezium/blob/4159397bd40dd5d0c8efe2b97ba61ba8eb220b58/debezium-core/src/main/java/io/debezium/connector/common/BaseSourceTask.java#L104) is called before [recover](https://github.com/debezium/debezium/blob/4159397bd40dd5d0c8efe2b97ba61ba8eb220b58/debezium-core/src/main/java/io/debezium/connector/common/BaseSourceTask.java#L148).
Let me try this.

zwang28: OpendalSchemaHistory's [exists](https://github.com/debezium/debezium/blob/4159397bd40dd5d0c8efe2b97ba61ba8eb220b58/debezium-core/src/main/java/io/debezium/connector/common/BaseSourceTask.java#L104) returns false when there's no entries in object store, which results in the incompatibility.

Let me refactor `OpendalSchemaHistory::exists` to always return true, and handle the no entry case in [recover](https://github.com/debezium/debezium/blob/4159397bd40dd5d0c8efe2b97ba61ba8eb220b58/debezium-core/src/main/java/io/debezium/connector/common/BaseSourceTask.java#L148), i.e. make `OpendalSchemaHistory::initializeStorage` a delegate of `MySqlConnector::initializeStorage` or other base connectors.

java/connector-node/risingwave-source-cdc/src/main/java/com/risingwave/connector/cdc/debezium/internal/OpendalSchemaHistory.java
```diff
@@ -0,0 +1,366 @@
...
+    }
+
+    @Override
+    public boolean storageExists() {
+        // Hummock bucket always exists.
+        return true;
+    }
+
+    @Override
+    public void initializeStorage() {
```
zwang28, line 214: I believe `initializeStorage` should initialize the schema history by query database's schema. Because there's no guaruantee that the log containing the initial schema change event is available.

java/connector-node/risingwave-source-cdc/src/main/java/com/risingwave/connector/cdc/debezium/internal/OpendalSchemaHistory.java
```diff
@@ -0,0 +1,366 @@
...
+                    historyFiles.size(),
+                    sequenceNumber.get());
+
+        } catch (Exception e) {
+            throw new SchemaHistoryException("Failed to initialize schema history", e);
+        }
+    }
+
+    @Override
+    public void doStop() {
```
zwang28: Will doStop be called in other cases, e.g. graceful shutdown of the compute node?

java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/DbzConnectorConfig.java
```diff
@@ -147,14 +148,18 @@ public DbzConnectorConfig(
 
         if (source == SourceTypeE.MYSQL) {
             var mysqlProps = initiateDbConfig(MYSQL_CONFIG_FILE, substitutor);
+
+            // Enable schema history for all MySQL CDC modes to handle schema changes properly
+            mysqlProps.setProperty(OpendalSchemaHistory.SOURCE_ID, String.valueOf(sourceId));
+            mysqlProps.setProperty(OpendalSchemaHistory.MAX_RECORDS_PER_FILE_CONFIG, "2048");
+
             if (isCdcBackfill) {
```
zwang28, line 157: When will isCdcBackfill be false? It's always [set for CDC source](https://github.com/risingwavelabs/risingwave/blob/wcy/opendal_schema_history/src/frontend/src/handler/create_source.rs#L753).

java/connector-node/risingwave-connector-service/src/main/resources/mysql.properties
zwang28: Why `snapshot.mode` was changed from `recovery` to `no-data` in this PR?

`mysqlProps.setProperty("snapshot.mode", "recovery");`

@wcy-fdu I believe we should keep `snapshot.mode=recovery`, so that
- when [offset is null](https://github.com/debezium/debezium/blob/4159397bd40dd5d0c8efe2b97ba61ba8eb220b58/debezium-core/src/main/java/io/debezium/connector/common/BaseSourceTask.java#L80), [no_data](https://github.com/risingwavelabs/risingwave/blob/db620e2070dbe3e00d9a5476c30c791707284e90/java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/DbzConnectorConfig.java#L167) is used, it will initialize the schema via `MySqlDatabaseSchema::initializeStorage` and persist it via `OpendalSchemaHistory::doStoreRecord`. This is how opendal schema history is initialized in this PR.
- when offset is not null and [opendal schema history is empty](https://github.com/debezium/debezium/blob/4159397bd40dd5d0c8efe2b97ba61ba8eb220b58/debezium-core/src/main/java/io/debezium/connector/common/BaseSourceTask.java#L104), it will initialize the schema via `MySqlDatabaseSchema::initializeStorage` and persist it via `OpendalSchemaHistory::doStoreRecord`. This requires `snapshot.mode=recovery`. **It maintains backward compatibility.**
- otherwise, it will [recover](https://github.com/debezium/debezium/blob/4159397bd40dd5d0c8efe2b97ba61ba8eb220b58/debezium-core/src/main/java/io/debezium/connector/common/BaseSourceTask.java#L148C57-L148C64) the schema from OpendalSchemaHistory. This is how opendal schema history is recovered in this PR.

java/connector-node/risingwave-source-cdc/src/main/java/com/risingwave/connector/cdc/debezium/internal/OpendalSchemaHistory.java
```diff
@@ -0,0 +1,366 @@
...
+                    "Error configuring an instance of "
+                            + getClass().getSimpleName()
+                            + "; check the logs for details");
+        }
+        sourceId = config.getString(SOURCE_ID);
+        if (sourceId == null || sourceId.isEmpty()) {
+            throw new DebeziumException(
+                    "Source ID is required for schema history. Please provide a unique source ID to avoid path conflicts between multiple sources.");
+        }
+        objectDir = String.format("rw-cdc-schema-history-source-%s", sourceId);
```
zwang28: To keep th hummock root folder organized, I recommend putting all `rw-cdc-schema-history-source-%s` under a single folder, e.g. `rw-cdc-schema-history`. The object path would be `rw-cdc-schema-history/[source_id]/schema_history.dat`.

java/connector-node/risingwave-connector-service/src/main/resources/mysql.properties
wcy-fdu: had some offline discussion with @zwang28, keep `snapshot.mode=recovery` does not work for offline schema change. We need to continue investigating...

java/connector-node/risingwave-source-cdc/src/main/java/com/risingwave/connector/cdc/debezium/internal/OpendalSchemaHistory.java
wcy-fdu: I agree.

java/connector-node/risingwave-source-cdc/src/main/java/com/risingwave/connector/cdc/debezium/internal/OpendalSchemaHistory.java
```diff
@@ -0,0 +1,366 @@
...
+                sourceId,
+                objectDir);
+        deleteObjects(objectDir);
+    }
+
+    @Override
+    protected void doPreStoreRecord(HistoryRecord record) {}
+
+    @Override
+    protected void doStoreRecord(HistoryRecord record) {
```
zwang28, line 155: Issue: 
It seems that RW doesn't update offset state after handling schema change event.
As a result, when a schema change event is not followed by any data write and RW restarts, RW can reconsume the schema change event, results in duplicate records in opendal schema history.
- It could lead to record sequence like [schema1, schema2, schema1, schema 2].

You can reproduce by:
1. Create a RW MySQL CDC source.
2. `alter table add column` in MySQL.
3. You'll find a new log line `Storing new schema history record`, which indicates the schema change in step 2 has been persisted.
4. Restart RW.
5. You'll find another new log line `Storing new schema history record`. Because the schema change event is reprocessed.
6. You can repeat step 4 and 5.

java/connector-node/risingwave-connector-service/src/main/resources/mysql.properties
zwang28: `RecoverySnapshotter` and `SchemaOnlySnapshotter` only differ in `shouldSnapshotSchema` and `shouldSnapshotOnSchemaError`. Need reviews the usage of these 2 methods.

zwang28: The snapshotter is also used by [executeChangeEventSources](https://github.com/debezium/debezium/blob/v2.6.2.Final/debezium-core/src/main/java/io/debezium/pipeline/ChangeEventSourceCoordinator.java#L189). `RecoverySnapshotter` 's `shouldSnapshotSchema=true` causes the issue here.

Let me see if any other snapshotter can meet our needs, or we can write a new one, i.e. `shouldSnapshotOnSchemaError=true` and `shouldSnapshotSchema=...`.

zwang28: This [customized snapshotter](https://github.com/risingwavelabs/risingwave/commit/ef6cb5c478aba1978f28dabe6a98f224779e38da) can succeed initializing the opendal schema history for a MySQL CDC source created prior to this PR.

However, there's still one corner case it cannot handle, for example:
1. Assume pending binlogs are [data write event 1], [schema change event 1], [data write event 2]
2. When opendal schema history is not initialized yet (e.g. old cluster), the new snapshotter will initialize the schema to latest, i.e. match that of [schema change event 1].
3. The binlog process will fail at [data write event 1], due to schema mismatch:
`risingwave_connector_node: Error processing after of row in pg_my_table because it's different column size with internal schema size 4, but after size 3, restart connector with schema recovery mode. thread="blc-127.0.0.1:4306" class="io.debezium.connector.mysql.MySqlStreamingChangeEventSource"`

zwang28:
BUG: default value is missing for new column during offline schema change. See comment at the end of the script.
[reproduce.sh](https://github.com/user-attachments/files/22509965/reproduce.sh)

This is a bug irrelevant to this PR. #23320

java/connector-node/risingwave-connector-service/src/main/resources/mysql.properties
zwang28: > However, there's still one corner case it cannot handle, for example:

My suggestion is:
- RisingWave kernel won't handle this corner case. Because it's difficult and sometimes impossible to get and use the non-current version schema, e.g. corresponding schema change binlog may have expired.
- The corner case only happen once when upgrading RisingWave to enable opendal schema history.
- The user can avoid this corner case by preventing any schema change during the RisingWave upgrade.

@hzxa21 @wcy-fdu WDYT?

java/connector-node/risingwave-source-cdc/src/main/java/com/risingwave/connector/cdc/debezium/internal/OpendalSchemaHistory.java
zwang28: It's likely that duplicate records in opendal schema history won't cause issue, because schema is [overwritten](https://github.com/debezium/debezium/blob/v2.6.2.Final/debezium-core/src/main/java/io/debezium/relational/history/AbstractSchemaHistory.java#L113) during replay.

zwang28: ~~But I haven't figured out why `HistoryRecord`s with greater position than start offset given by RW are not skipped as expected in my test, i.e. I don't see this [expected log](https://github.com/debezium/debezium/blob/v2.6.2.Final/debezium-core/src/main/java/io/debezium/relational/history/AbstractSchemaHistory.java#L154) here.~~

Because it uses [this comparator](https://github.com/debezium/debezium/blob/v2.6.2.Final/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/strategy/AbstractHistoryRecordComparator.java#L48).

java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/DbzConnectorConfig.java
zwang28: ~~I think only non-shared CDC source will result in `isCdcBackfill=false`. non-shared CDC source has been deprecated.~~
`isCdcBackfill=false` for MongoDB CDC.

java/connector-node/risingwave-source-cdc/src/main/java/com/risingwave/connector/cdc/debezium/internal/OpendalSchemaHistory.java
zwang28: I recommend that the meta node handle the cleanup after it successfully deletes the catalog.

The issue with `doStop` is that the Debezium or RisingWave connector can call it unexpectedly, even in scenarios unrelated to a DROP SOURCE command.

java/connector-node/risingwave-source-cdc/src/main/java/com/risingwave/connector/cdc/debezium/internal/OpendalSchemaHistory.java
zwang28: FYI [this debezium history record comparator](https://github.com/debezium/debezium/blob/v2.6.2.Final/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/strategy/AbstractHistoryRecordComparator.java#L101) doesn't behave correctly here: the record_2 is after record_1, which should be skipped during record replay. However, the comparator won't skip it, because it tries to compare `TIMESTAMP_KEY = "ts_ms"` which is NULL for both.

```
record_1 from start offset: {
  "transaction_id" : "file=binlog.000002,pos=176334",
  "transaction_data_collection_order_cdc.pg_my_table" : 1,
  "file" : "binlog.000002",
  "pos" : 176334,
  "row" : 1,
  "event" : 2
}

record_2 from opendal schema history: {
  "ts_sec" : 1758793154,
  "file" : "binlog.000002",
  "pos" : 183681,
  "server_id" : 1
}; 

```

server_id is not set [here](https://github.com/debezium/debezium/blob/4159397bd40dd5d0c8efe2b97ba61ba8eb220b58/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlOffsetContext.java#L84).

src/jni_core/src/opendal_schema_history.rs
zwang28: Resolved.

src/jni_core/src/opendal_schema_history.rs
```diff
@@ -0,0 +1,279 @@
...
+    Ok(())
+}
+
+// schema history is internal state, all data is stored under the DATA_DIRECTORY directory.
+fn prepend_data_directory(path: &str) -> String {
+    let data_dir = DATA_DIRECTORY.get().map(|s| s.as_str()).expect(
+        "DATA_DIRECTORY is not set. This is dangerous in cloud environments as it can cause data conflicts between multiple instances sharing the same bucket. Please ensure data_directory is properly configured."
+    );
+
+    if path.starts_with(data_dir) {
```
zwang28: Compare `path` and `data_dir` is not safe. It's possible `path` = 'aabc', `data_dir` = 'aa', and the expected value is `aa/aabc`.

zwang28: i.e. need a clear indicator whether should prefix the `path` with `data_dir`.

src/jni_core/src/opendal_schema_history.rs
```diff
@@ -0,0 +1,279 @@
...
+pub extern "system" fn Java_com_risingwave_java_binding_Binding_getObject<'a>(
+    env: EnvParam<'a>,
+    object_name: JString<'a>,
+) -> JByteArray<'a> {
+    execute_and_catch(env, move |env: &mut EnvParam<'_>| {
+        let object_name = env.get_string(&object_name)?;
+        let object_name: Cow<'_, str> = (&object_name).into();
+
+        // Security check: validate file extension before any operation
+        if let Err(error_msg) = validate_dat_file_extension(&object_name) {
```
zwang28: Error should be raised rather than silenced.

src/jni_core/src/opendal_schema_history.rs
```diff
@@ -0,0 +1,279 @@
...
+            return Ok(env.byte_array_from_slice(&[])?); // Return empty byte array
+        }
+
+        let object_name = prepend_data_directory(&object_name);
+        let result = JAVA_BINDING_ASYNC_RUNTIME.block_on(async {
+            let object_store = get_object_store().await;
+            match object_store.read(&object_name, ..).await {
+                Ok(data) => data,
+                Err(e) => {
+                    tracing::error!(
```
zwang28: ditto

src/jni_core/src/opendal_schema_history.rs
```diff
@@ -0,0 +1,279 @@
...
+        // Note: listObject operates on directories, individual file security is checked in putObject/getObject
+
+        let dir = prepend_data_directory(&dir);
+
+        let files: Vec<String> = JAVA_BINDING_ASYNC_RUNTIME.block_on(async {
+            let object_store = get_object_store().await;
+            let mut file_names = Vec::new();
+            let mut stream = match object_store.list(&dir, None, None).await {
+                Ok(s) => s,
+                Err(e) => {
```
zwang28: ditto

src/jni_core/src/opendal_schema_history.rs
```diff
@@ -0,0 +1,279 @@
...
+                        };
+
+                        if validate_dat_file_extension(relative_path).is_ok() {
+                            file_names.push(obj.key);
+                        } else {
+                            tracing::error!("Filtering out non-.dat file from list: {}", obj.key);
+                        }
+                    }
+                    Err(e) => {
+                        tracing::error!("CDC Schema History: Failed to get object from list stream: {}", e.as_report());
```
zwang28: ditto

src/jni_core/src/opendal_schema_history.rs
```diff
@@ -0,0 +1,279 @@
...
+        // Note: deleteObjects operates on directories, individual file security is checked in putObject/getObject
+
+        let dir = prepend_data_directory(&dir);
+
+        JAVA_BINDING_ASYNC_RUNTIME.block_on(async {
+            let object_store = get_object_store().await;
+            let mut keys = Vec::new();
+            let mut stream = match object_store.list(&dir, None, None).await {
+                Ok(s) => s,
+                Err(e) => {
```
zwang28: ditto

src/jni_core/src/opendal_schema_history.rs
```diff
@@ -0,0 +1,279 @@
...
+                            relative_path
+                        };
+
+                        if validate_dat_file_extension(relative_path).is_ok() {
+                            keys.push(obj.key);
+                        } else {
+                            tracing::error!("Skipping deletion of non-.dat file: {}", obj.key);
+                        }
+                    }
+                    Err(e) => {
```
zwang28: ditto

src/jni_core/src/opendal_schema_history.rs
zwang28: https://github.com/risingwavelabs/risingwave/pull/22255/commits/897c7dae262908123f4d0a51f8bff0fa77c046db

java/connector-node/risingwave-source-cdc/src/main/java/com/risingwave/connector/cdc/debezium/internal/OpendalSchemaHistory.java
zwang28: https://github.com/risingwavelabs/risingwave/pull/22255/commits/ed76b30fd8fb590a0fed2c6493c8e87f8405afda

src/jni_core/src/opendal_schema_history.rs
zwang28: https://github.com/risingwavelabs/risingwave/pull/22255/commits/fb7149e7d30d6c6817494fd3d16c2ab9bdd63fc0

src/jni_core/src/opendal_schema_history.rs
zwang28: IIUC, the need comes rom `cachedLatestFile` can be either with or without DATA_DIRECTORY prefix?
- with prefix if its path is restored by listing object.
- without prefix if its path is contructed by creating a new file.

zwang28: I recommend keeping it consistent: `opendal_schema_history.rs` being aware of DATA_DIRECTORY and `OpendalSchemaHistory.java` being unaware of DATA_DIRECTORY.

zwang28: https://github.com/risingwavelabs/risingwave/pull/22255/commits/c34dc1f50e4f0dee1ea5cf80b475082c05afdbcd

src/jni_core/src/opendal_schema_history.rs
zwang28: https://github.com/risingwavelabs/risingwave/commit/fb7149e7d30d6c6817494fd3d16c2ab9bdd63fc0

java/connector-node/risingwave-source-cdc/src/main/java/com/risingwave/connector/cdc/debezium/internal/OpendalSchemaHistory.java
wcy-fdu: ed76b30 LGTM

java/connector-node/risingwave-source-cdc/src/main/java/com/risingwave/connector/cdc/debezium/internal/OpendalSchemaHistory.java
zwang28: https://github.com/risingwavelabs/risingwave/pull/22255/commits/e03cfce8b62b47f510d191f6850cddb0cc1efbdc

java/connector-node/risingwave-source-cdc/src/main/java/com/risingwave/connector/cdc/debezium/internal/OpendalSchemaHistory.java
zwang28: In summary duplicate opendal schema history records don't impact correctness.

src/jni_core/src/macros.rs
zwang28: I'm OK with keeping the JNI version for now.

src/jni_core/src/opendal_schema_history.rs
wcy-fdu: fb7149e LGTM!

wcy-fdu:
The CI test introduced by this PR passed in the latest commit.

image

src/jni_core/src/opendal_schema_history.rs
wcy-fdu: LGTM

java/connector-node/risingwave-connector-service/src/main/resources/mysql.properties
wcy-fdu: I think this "corner" case is actually the problem that this PR wants to solve? This can happen not only during the upgrading phase, but also during the downtime of RisingWave (if there is a schema change upstream during this period). see https://github.com/risingwavelabs/risingwave/issues/21801

wcy-fdu: IIUC, after this pr merged, such case will never occurs because we persist schema history.
So personally, I think that solving compatibility issues is more important than dealing with this corner case, because this corner case will not occur after using the opendal schema history source.

zwang28: After merging this PR, this corner case can still occur when upgrading an old cluster that haven't initialized opendal schema history yet.

hzxa21: How about providing some way to clear the opendal schema history and let dbz read the latest schema? I think this can be useful not only to recover the source when the corner case is met but also useful in many other cases (e.g. there is a bug in opendal schema history).

java/connector-node/risingwave-source-cdc/src/main/java/com/risingwave/connector/cdc/debezium/internal/OpendalSchemaHistory.java
hzxa21: https://github.com/risingwavelabs/risingwave/pull/22255#discussion_r2378568216

will this be an issue or is it irrelevant?

java/connector-node/risingwave-connector-service/src/main/resources/mysql.properties
hzxa21: > clear the opendal schema history and let dbz read the latest schema

This is also useful when there are too many opendal schema history files accumulated in the future and there is a need to clean it up.

java/connector-node/risingwave-source-cdc/src/main/java/com/risingwave/connector/cdc/debezium/internal/OpendalSchemaHistory.java
```diff
@@ -0,0 +1,357 @@
...
+        return result;
+    }
+
+    // Extract sequence number from file name
+    private long extractSequenceFromFileName(String fileName) {
+        Matcher m = HISTORY_FILE_PATTERN.matcher(fileName);
+        if (m.find()) {
+            return Long.parseLong(m.group(1));
+        }
+        return 0L;
```
hzxa21: Is this line reachable? If not, we should throw an exception when there is no match instead of returning 0.

java/connector-node/risingwave-source-cdc/src/main/java/com/risingwave/connector/cdc/debezium/internal/OpendalSchemaHistory.java
```diff
@@ -0,0 +1,357 @@
...
+
+    /**
+     * Initialize sequence number from existing files. This ensures we don't reuse sequence numbers
+     * that have already been used.
+     */
+    private void initializeSequenceNumber(List<String> historyFiles) {
+        long maxSequence = 0;
+
+        // Find the maximum sequence number from existing files
+        for (String fileName : historyFiles) {
```
hzxa21: `historyFiles` is already been sorted by seq number in `listAndSortHistoryFiles` so you can directly get the max seq instead of doing another iteration. If the iteration here is for validation, you should throw an exception when the seq number is not increasing.

java/connector-node/risingwave-source-cdc/src/main/java/com/risingwave/connector/cdc/debezium/internal/OpendalSchemaHistory.java
```diff
@@ -0,0 +1,357 @@
...
+        historyFiles.removeIf(file -> !HISTORY_FILE_PATTERN.matcher(file).find());
+        Collections.sort(historyFiles, Comparator.comparingLong(this::extractSequenceFromFileName));
+        return historyFiles;
+    }
+
+    /**
+     * Initialize cache for the latest file information to optimize future store operations. This
+     * avoids expensive list operations on every doStoreRecord call.
+     */
+    private void initializeLatestFileCache(List<String> historyFiles) {
```
hzxa21: nits: I think we can init `cachedLatestFile` and `cachedFileRecordCount` inside `loadAllHistoryRecords` instead of reading the last file again here.

java/connector-node/risingwave-connector-service/src/main/resources/mysql.properties
wcy-fdu: Do you mean some operation and maintenance measures, such as risectl triggering a full gc of the schema history, or deleting it during runtime in the code?

java/connector-node/risingwave-source-cdc/src/main/java/com/risingwave/connector/cdc/debezium/internal/OpendalSchemaHistory.java
```diff
@@ -0,0 +1,357 @@
...
+            }
+        } catch (Exception e) {
+            throw new SchemaHistoryException("Failed to store schema history record", e);
+        }
+    }
+
+    @Override
+    public boolean storageExists() {
+        // Hummock bucket always exists.
+        return true;
```
hzxa21: We return true no mater whether the schema history is empty or it is initialized. `storageExists` will be called in `historyExists` and we use this interface in the `NoDataRecoverySnapshotter`: https://github.com/risingwavelabs/risingwave/blob/e03cfce8b62b47f510d191f6850cddb0cc1efbdc/java/connector-node/risingwave-source-cdc/src/main/java/com/risingwave/connector/cdc/debezium/internal/NoDataRecoverySnapshotter.java#L45

Will it be an issue?

java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/DbzConnectorConfig.java
wcy-fdu: This is to make the _Connector Node integration Test_ pass. In this test, the previous `MemorySchemaHistory` is used.

java/connector-node/risingwave-source-cdc/src/main/java/com/risingwave/connector/cdc/debezium/internal/OpendalSchemaHistory.java
wcy-fdu: I think no correctness issue, @zwang28 can double confirm.

java/connector-node/risingwave-source-cdc/src/main/java/com/risingwave/connector/cdc/debezium/internal/OpendalSchemaHistory.java
zwang28: This `databaseSchema` object is not an `OpendalSchemaHistory` instance.

zwang28: The `databaseSchema` is an instance of [MySqlDatabaseSchema class](https://github.com/debezium/debezium/blob/v2.6.2.Final/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlDatabaseSchema.java#L62).
The `((HistorizedDatabaseSchema) databaseSchema).historyExists()` calls [HistorizedRelationalDatabaseSchema::historyExists](https://github.com/debezium/debezium/blob/v2.6.2.Final/debezium-core/src/main/java/io/debezium/relational/HistorizedRelationalDatabaseSchema.java#L163), which calls [schemaHistory::exists](https://github.com/debezium/debezium/blob/v2.6.2.Final/debezium-core/src/main/java/io/debezium/relational/HistorizedRelationalDatabaseSchema.java#L40). The [schemaHistory](https://github.com/debezium/debezium/blob/v2.6.2.Final/debezium-core/src/main/java/io/debezium/relational/HistorizedRelationalDatabaseSchema.java#L40) member is an instance of `OpendalSchemaHistory`. So `schemaHistory::exists` actually calls [this](https://github.com/debezium/debezium/blob/v2.6.2.Final/debezium-core/src/main/java/io/debezium/relational/history/AbstractFileBasedSchemaHistory.java#L131).

So the result of `historyExists` is determined by the [records](https://github.com/debezium/debezium/blob/v2.6.2.Final/debezium-core/src/main/java/io/debezium/relational/history/AbstractFileBasedSchemaHistory.java#L37), which should be handled by `OpendalSchemaHistory` correctly.

hzxa21: Thanks for the explanation!

java/connector-node/risingwave-connector-service/src/main/resources/mysql.properties
zwang28: > clear the opendal schema history and let dbz read the latest schema

The corner case is exactly caused by 
- dbz using latest schema due to empty opendal schema history.
- non-latest change event cannot be processed by latest schema.

How would such clearing help this corner case?

hzxa21: IIUC, the corner case can be resolved by retrying the initialization until the latest change events are processed. IIUC, we don't have a way to auto recovery and retry in our new codes. Therefore, if we have a manual way to recover and retry, it will be beneficial.

hzxa21: Chatted offline with @zwang28. My comment [here](https://github.com/risingwavelabs/risingwave/pull/22255#discussion_r2451659133) about having a tool to reset the schema history cannot resolve the corner case, because it is related to the fact the schema history during the upgrade is lost and it has nothing to do with the current PR.

The tool is useful for recovery when bugs happen in opendal schema history and for reclaim spaces from opendal schema history. From this angle, the tool is not a must. We can develop such tool when real issue happens.

java/connector-node/risingwave-source-cdc/src/main/java/com/risingwave/connector/cdc/debezium/internal/OpendalSchemaHistory.java
zwang28: As explained above, it won't cause issue because schema is overwritten when processing each history record, so eventually it's correct after applying the lastest one. 

`class TableChange` (from history record) contains [complete table metadata](https://github.com/debezium/debezium/blob/v3.2.4.Final/debezium-core/src/main/java/io/debezium/relational/history/JsonTableChangeSerializer.java#L236), which is used to overwrite.

zwang28: A sample history record generated by `alter add column`. It contains complete columns.

[alter_add_col.json](https://github.com/user-attachments/files/23181680/alter_add_col.json)

zwang28 (approved):
Thanks.

PR #22271, opened 2025-06-18
feat(ci): support fast docker workflow in github workflow

kwannoel:

We support fast docker workflow in parallel to the current docker workflow. Just so users can quickly try out a patch. Concurrently we will build the normal image, in case there are any issues, users can switch over.

Note that this workflow only supports fast builds for `ARM64`, because we only have that architecture stack provisioned in our CI. That's mostly fine because our production deployments are typically on that arch.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/.github/workflows/build-docker-image.yml
+++ b/.github/workflows/build-docker-image.yml
@@ -8,6 +8,11 @@ on:
         required: true
         type: string
         default: 'unlabeled'
+      fast:
+        description: 'It will build an Arm64 image quickly, in parallel with the normal build'
+        required: false
+        type: boolean
+        default: false
 
 jobs:
   build_image:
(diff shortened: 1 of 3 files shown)


kwannoel:
I went to look at our fast docker workflow again, the reason why our docker workflow is much faster for ARM than for x86 is because we provision a bigger machine for it 😆  So x86 is the actual bottleneck.
Maybe we should just do that for x86 too, and let our default docker workflow be faster. cc @cyliu0 what do you think, is it possible to provision a larger instance for x86, just for image building

tabVersion:
I might lack some context, do we sacrifice something to get it faster?

kwannoel:
> I might lack some context, do we sacrifice something to get it faster?

Yes, we only build arm image. Because no fast pipeline exists for x86 arch.

kwannoel:
https://github.com/risingwavelabs/risingwave/pull/23402

PR #22289, opened 2025-06-19
fix(meta): unify validation logic for system parameters

Chiro11:

This PR enhances system parameter validation by introducing checks for initial parameter values at startup.

The validation logic is now shared between parameter initialization and runtime updates, ensuring consistent rule enforcement for all parameters, regardless of their mutability.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/common/src/system_param/mod.rs
+++ b/src/common/src/system_param/mod.rs
@@ -249,17 +249,15 @@ macro_rules! impl_system_params_from_kv {
 /// Define check rules when a field is changed.
 /// If you want custom rules, please override the default implementation in
 /// `OverrideValidateOnSet` below.
-macro_rules! impl_default_validation_on_set {
+macro_rules! impl_default_validation {
     ($({ $field:ident, $type:ty, $default:expr, $is_mutable:expr, $($rest:tt)* },)*) => {
         #[allow(clippy::ptr_arg)]
-        trait ValidateOnSet {
+        trait Validate {
             $(
+                /// Default implementation does nothing.
+                /// Specific checks are implemented in `OverrideValidate`.
                 fn $field(_v: &$type) -> Result<()> {
-                    if !$is_mutable {
-                        Err(format!("{:?} is immutable", key_of!($field)))
-                    } else {
-                        Ok(())
-                    }
+                    Ok(())
                 }
             )*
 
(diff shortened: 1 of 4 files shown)


src/common/src/system_param/mod.rs
```diff
@@ -387,23 +389,50 @@ macro_rules! impl_system_params_for_test {
             ret.backup_storage_url = Some("memory-isolated-for-test".into());
             ret.backup_storage_directory = Some("backup".into());
             ret.use_new_object_prefix_strategy = Some(false);
-            ret.time_travel_retention_ms = Some(0);
```
yezizp2012: I understand that removing this line is necessary to pass the config validation. But seems like 0 is a legal value, which might be used for testing purpose? https://github.com/risingwavelabs/risingwave/blob/e1ae4d45f1e719277af80ce489003c9c49f74367/src/meta/src/hummock/manager/time_travel.rs#L528-L536 should we change the validation logic for `time_travel_retention_ms`? 
https://github.com/risingwavelabs/risingwave/blob/e1ae4d45f1e719277af80ce489003c9c49f74367/src/common/src/system_param/mod.rs#L429-L438
Need confirm from @zwang28 .

src/config/ci-backfill.toml
```diff
@@ -11,5 +11,5 @@ streaming_upload_attempt_timeout_ms = 20000
 # NOTE(kwannoel): If can't reproduce it, set to a lower number.
 # This will throttle snapshot read.
 # barrier_interval_ms = 1
-barrier_interval_ms = 50
```
yezizp2012: Ditto. I think we should change the validation logic for `barrier_interval_ms`. Increasing the configuration here will lead to a longer ci test time.

src/tests/simulation/src/backfill.toml
```diff
@@ -9,5 +9,5 @@ stream_chunk_size = 1
 # NOTE(kwannoel): If can't reproduce it, set to a lower number.
 # This will throttle snapshot read.
 # barrier_interval_ms = 1
-barrier_interval_ms = 50
```
yezizp2012: Ditto.

yezizp2012 (commented):
Rest LGTM.

src/config/ci-backfill.toml
Chiro11: OK, I changed the validation value.

src/common/src/system_param/mod.rs
zwang28: 0 is used to disable time travel.

Please help modfiy the validation rule to permit 0.

please don't remove the line in the test.

yezizp2012 (approved):
LGTM

PR #22295, opened 2025-06-19
simplify filter predicate before converting row_number + filter to topn

stdrc:

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/planner_test/tests/testdata/output/over_window_function.yaml
+++ b/src/frontend/planner_test/tests/testdata/output/over_window_function.yaml
@@ -855,6 +855,110 @@
           └─StreamOverWindow { window_functions: [row_number() OVER(PARTITION BY t.x ORDER BY t.y ASC ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW)] }
             └─StreamExchange { dist: HashShard(t.x) }
               └─StreamTableScan { table: t, columns: [t.x, t.y, t.z, t._row_id], stream_scan_type: ArrangementBackfill, stream_key: [t._row_id], pk: [_row_id], dist: UpstreamHashShard(t._row_id) }
+- name: test ibis window function optimization
+  sql: |
+    CREATE TABLE t (a INT, b INT, c INT);
+    SELECT a, b, c
+    FROM (
+      SELECT
+        a, b, c,
+        ROW_NUMBER() OVER (PARTITION BY a ORDER BY b) - 1 AS rn
+      FROM t
+    )
+    WHERE rn < 10;
+  logical_plan: |-
+    LogicalProject { exprs: [t.a, t.b, t.c] }
+    └─LogicalFilter { predicate: ($expr1 < 10:Int32) }
+      └─LogicalProject { exprs: [t.a, t.b, t.c, (row_number - 1:Int32) as $expr1] }
+        └─LogicalOverWindow { window_functions: [row_number() OVER(PARTITION BY t.a ORDER BY t.b ASC ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW)] }
+          └─LogicalProject { exprs: [t.a, t.b, t.c, t._row_id, t._rw_timestamp] }
+            └─LogicalScan { table: t, columns: [t.a, t.b, t.c, t._row_id, t._rw_timestamp] }
+  optimized_logical_plan_for_batch: |-
+    LogicalTopN { order: [t.b ASC], limit: 10, offset: 0, group_key: [t.a] }
+    └─LogicalScan { table: t, columns: [t.a, t.b, t.c] }
+  optimized_logical_plan_for_stream: |-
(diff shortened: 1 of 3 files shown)


stdrc:
* **#22295** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22295?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22295?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/frontend/src/optimizer/rule/over_window_to_topn_rule.rs
```diff
@@ -137,6 +150,131 @@ impl Rule for OverWindowToTopNRule {
...
+            Add | Subtract => {
+                let inputs = arithmetic_func.inputs();
+                if inputs.len() == 2 {
+                    // Find column reference and constant
+                    let (column_ref, arith_const, reverse_op) = if inputs[1].is_const() {
+                        // col op const
+                        let reverse_op = match arithmetic_func.func_type() {
+                            Add => Subtract,
+                            Subtract => Add,
+                            _ => return None,
```
xxchan:
```suggestion
                            _ => unreachable!(),
```

src/frontend/src/optimizer/rule/over_window_to_topn_rule.rs
```diff
@@ -137,6 +150,131 @@ impl Rule for OverWindowToTopNRule {
     }
 }
 
+impl OverWindowToTopNRule {
+    /// Simplify arithmetic expressions in filter conditions before TopN optimization
+    /// For example: (`row_number` - 1) = 0 -> `row_number` = 1
+    fn simplify_filter_arithmetic(&self, filter: &LogicalFilter) -> Option<LogicalFilter> {
```
xxchan, line 156: Maybe `simplify_comparison_arithmetic`?

src/frontend/planner_test/tests/testdata/output/over_window_function.yaml
```diff
@@ -855,6 +855,110 @@
...
+        a, b, c,
+        ROW_NUMBER() OVER (PARTITION BY a ORDER BY b) + 1 AS rn_plus_one
+      FROM t
+    )
+    WHERE rn_plus_one = 2;
+  logical_plan: |-
+    LogicalProject { exprs: [t.a, t.b, t.c, $expr1] }
+    └─LogicalFilter { predicate: ($expr1 = 2:Int32) }
+      └─LogicalProject { exprs: [t.a, t.b, t.c, (row_number + 1:Int32) as $expr1] }
+        └─LogicalOverWindow { window_functions: [row_number() OVER(PARTITION BY t.a ORDER BY t.b ASC ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW)] }
```
kwannoel, line 921: Potential for future optimization? Seems like `rn_plus_one` is always `2`. I guess this can just be a project

src/frontend/src/optimizer/rule/over_window_to_topn_rule.rs
```diff
@@ -137,6 +150,131 @@ impl Rule for OverWindowToTopNRule {
...
+
+        if new_expr != expr {
+            Some(Condition::with_expr(new_expr))
+        } else {
+            None
+        }
+    }
+}
+
+/// Filter arithmetic simplification rewriter: simplifies (col op const) = const2 to col = (const2 `reverse_op` const)
```
kwannoel: This optimizer rule seems kind of adhoc. If we want to add it, we'd better clarify in the docstring that this is specifically used for OverWindow, to simplify the filter predicate and allow it to be rewritten to TopN.

Further, we should also state that it only supports `-/+` for `op/reverse_op`.
```
(col op const) = const2 to col = (const2 `reverse_op` const)
```

kwannoel: > This optimizer rule seems kind of adhoc. If we want to add it, we'd better clarify in the docstring that this is specifically used for OverWindow, to simplify the filter predicate and allow it to be rewritten to TopN.

Oh nevermind it's already part of the `over_window_to_topn_rule` module, ignore this comment.

kwannoel (approved):

src/frontend/src/optimizer/rule/over_window_to_topn_rule.rs
stdrc: `simplify_filter_arithmetic` is called for `Filter` node with no brain, I guess the name works.

src/frontend/planner_test/tests/testdata/output/over_window_function.yaml
stdrc: Indeed, but I guess if the condition is like this one, the user won't want to keep the row number in the final select list.

PR #22301, opened 2025-06-20
fix(parquet-source): recursively check the schema of nested data types

wcy-fdu:

The previous implementation only performed schema checks on the outermost layer of nested data types, which could lead to mismatches in the inner schema. 
e.g  `struct<f1: Double, f2: Utf8>` should not match `rw_struct<f1: decimal, f2: Utf8>`

This PR adds recursive validation for nested types and includes unit tests.

## Checklist

## Documentation

<details>

</details>


wcy-fdu:
This PR only adds schema checks. For user-defined columns, if they do not match the Parquet schema 
1. The column name cannot be found in the Parquet file; 
2. The column name matches but the data type is inconsistent), 

a null value will currently be filled in. In the future, we may read the metadata of a Parquet file when creating a source/table to inform the user in advance about any schema mismatches.

src/common/src/array/arrow/arrow_impl.rs
```diff
@@ -1442,82 +1443,199 @@ impl From<&arrow_array::Decimal256Array> for Int256Array {
...
+        | (ArrowType::Binary | ArrowType::LargeBinary, RwType::Bytea) => true,
+
+        // Struct type recursive matching
+        // Arrow's Struct matches RisingWave's Struct if all field names and types match recursively
+        (ArrowType::Struct(arrow_fields), RwType::Struct(rw_struct)) => {
+            if arrow_fields.len() != rw_struct.len() {
+                return false;
+            }
+            for (arrow_field, (rw_name, rw_ty)) in
+                arrow_fields.iter().zip_eq_debug(rw_struct.iter())
```
xiangjinwu:
```suggestion
                arrow_fields.iter().zip_eq_fast(rw_struct.iter())
```

src/common/src/array/arrow/arrow_impl.rs
```diff
@@ -48,6 +48,7 @@ use arrow_buffer::OffsetBuffer;
 use arrow_schema::TimeUnit;
 use chrono::{DateTime, NaiveDateTime, NaiveTime};
 use itertools::Itertools;
+use rw_iter_util::ZipEqDebug;
```
xiangjinwu: unused

PR #22309, opened 2025-06-20
support adding latency to pg

kwannoel:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## Checklist

- [ ] I have written necessary rustdoc comments.
- [ ]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  I have checked the [Release Timeline](https://github.com/risingwavelabs/rw-commits-history/blob/main/release_timeline.md) and [Currently Supported Versions](https://docs.risingwave.com/changelog/release-support-policy#support-end-dates-for-recent-releases) to determine which release branches I need to cherry-pick this PR into. 

## Documentation

- [ ]  My PR needs documentation updates. 

<details>
<summary><b>Release note</b></summary>

</details>

PR #22310, opened 2025-06-20
perf(connector): optimize pg native sink writer with pipelining

kwannoel:

[
We add pipelining in this PR](https://www.postgresql.org/docs/current/libpq-pipeline-mode.html), which can be simply done by concurrently executing a series of futures.

This also removes complicated logic around parameter batching, just do query pipelining.

These fixes are done for PG rust sink.

## Checklist

### DIFF
--- a/src/connector/src/sink/postgres.rs
+++ b/src/connector/src/sink/postgres.rs
@@ -19,9 +19,8 @@ use async_trait::async_trait;
 use itertools::Itertools;
 use phf::phf_set;
 use risingwave_common::array::{Op, StreamChunk};
-use risingwave_common::bail;
 use risingwave_common::catalog::Schema;
-use risingwave_common::row::{Row, RowExt};
+use risingwave_common::row::Row;
 use serde_derive::Deserialize;
 use serde_with::{DisplayFromStr, serde_as};
 use simd_json::prelude::ArrayTrait;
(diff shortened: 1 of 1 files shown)


kwannoel:
* **#22310** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22310?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22310?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#22309** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22309?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

BugenZhao:
> 1. We maintain the order of statement execution using `FuturesOrdered`.

Is this true? I guess it only ensures the order of returning the future outputs, not the actual execution order.

src/connector/src/sink/postgres.rs
```diff
@@ -391,15 +328,41 @@ impl PostgresSinkWriter {
...
+            pk_indices,
             pk_types,
             schema_types,
-            schema,
+            raw_insert_sql: Arc::new(raw_insert_sql),
+            raw_upsert_sql: Arc::new(raw_upsert_sql),
+            raw_delete_sql: Arc::new(raw_delete_sql),
+            insert_sql: Arc::new(insert_sql),
+            delete_sql: Arc::new(delete_sql),
+            upsert_sql: Arc::new(upsert_sql),
```
stdrc, line 365: What's the benefits of using `Arc` here for these fields?

src/connector/src/sink/postgres.rs
```diff
@@ -256,83 +258,18 @@ impl Sink for PostgresSink {
...
-    pk_indices_lookup: HashSet<usize>,
     is_append_only: bool,
     client: tokio_postgres::Client,
+    pk_indices: Vec<usize>,
     pk_types: Vec<PgType>,
     schema_types: Vec<PgType>,
-    schema: Schema,
+    raw_insert_sql: Arc<String>,
+    raw_upsert_sql: Arc<String>,
+    raw_delete_sql: Arc<String>,
```
stdrc, line 269: Why do we need to have raw SQL here? Is it possible to reconstruct (`Display` or something) from the PG `Statement`?

src/connector/src/sink/postgres.rs
```diff
@@ -415,219 +378,99 @@ impl PostgresSinkWriter {
...
-            &self.pk_indices,
-            &self.pk_indices_lookup,
-            parameters,
-            remaining,
-            true,
-        )
-        .await?;
-        transaction.commit().await?;
+
+        while let Some(result) = insert_futures.next().await {
```
stdrc, line 411: Seems like we are using `FuturesOrdered` as a `Vec` here, but are you sure `.next` won't spawn all the futures to run concurrently which will then cause the statements to be executed in a different order from the correct one? I suppose we only use `FuturesOrdered` to get results in the same order as insertion, but the execution still expected to be concurrent and out-of-order.

src/connector/src/sink/postgres.rs
stdrc: Nvm I see we need to clone this below

src/connector/src/sink/postgres.rs
kwannoel: No https://docs.rs/tokio-postgres/latest/tokio_postgres/struct.Statement.html

kwannoel:
@BugenZhao @stdrc I reverted the interleave commit. We will now run all deletes, followed by all inserts. This matches behaviour of JDBC. For our PG sink, we will have at most one entry per key, since we will run compaction on the input chunk.

* There are some caveats to this, when recovery happens. But that bug is present in both JDBC and PG sink, we can fix it later.

src/connector/src/sink/postgres.rs
kwannoel: resolved by https://github.com/risingwavelabs/risingwave/pull/22310/commits/98b3f8088ef208171cb4967477f3454b76048693

src/connector/src/sink/postgres.rs
```diff
@@ -415,219 +378,103 @@ impl PostgresSinkWriter {
...
-            //
-            // We may opt for an optimization flag to toggle this behaviour in the future.
-            chunk.data_types().len(),
-        );
-        let mut delete_parameter_buffer =
-            ParameterBuffer::new(&self.pk_types, chunk.cardinality() * self.pk_indices.len());
-        // 1d flattened array of parameters to be deleted.
+        let transaction = Arc::new(self.client.transaction().await?);
+        let mut delete_futures = FuturesOrdered::new();
+        let mut upsert_futures = FuturesOrdered::new();
```
chenzl25: It seems that even if we use `FuturesOrdered`, the actual statement execution order can't be guaranteed. If we have multiple statement upserts of the same PK, then it appears that the stale value could overwrite the latest value.

chenzl25: Just know that we have compacted the chunk for postgres sink already (except the reading from logstore), it seems we can use FuturesUnordered instead.

Issue #21797, opened 2025-05-09
Disable stream creation when creating a NATS source


benjamin-awd:
### Is your feature request related to a problem? Please describe.

Creating a NATS source where consumer replicas != stream replicas will result in a NATS error:
```
Nats error: JetStream error: 400 (code consumer config replica count exceeds parent stream, error code 10126
which will then break the cluster during sink creation:
  Caused by these errors (recent errors listed first):
    1: gRPC request to meta service (call `/ddl_service.DdlService/CreateSink`) failed: Internal error
    2: failed to collect barrier
  
  Caused by these errors (recent errors listed first):
    1: gRPC request to meta service (call `/ddl_service.DdlService/CreateSink`) failed: The service is currently unavailable
    2: Service unavailable: database 6 unavailable cluster is under recovering
```
CPU usage then spikes to 100, meaning that in the worst case it is impossible to drop the source and stops ingestion for all other sources running on the same database

### Describe the solution you'd like

Option 1: add a config flag to disable stream creation when creating a NATS source

Option 2: change the default to `get_stream` instead of `get_or_create_stream`, so that source creation fails if the stream doesn't already exist i.e do not create stream unless a source setting like `create_stream_if_not_exists=true` is specified

FWIW I prefer option 2, since high volume streams require careful configuration e.g. choosing between file or memory storage, number of replicas, max bytes per stream, discard policy etc

tabVersion:
I am not sure if getting this correct

https://github.com/risingwavelabs/risingwave/blob/14d4c71e81fbaa57e48e4ce79dc1faa13c82cd95/src/connector/src/connector_common/common.rs#L943-L950

This logic load stream directly if the stream exists, so it should not create a new one if predefined.

benjamin-awd:
@tabVersion 

I think there's another line here that creates the stream if it doesn't yet exist:
https://github.com/risingwavelabs/risingwave/blob/14d4c71e81fbaa57e48e4ce79dc1faa13c82cd95/src/connector/src/connector_common/common.rs#L978

tabVersion:
make sense, will fix later.

PR #22315, opened 2025-06-21
feat: add `allow_create_stream` to avoid create nats stream by mistake

tabVersion:

resolve https://github.com/risingwavelabs/risingwave/issues/21797

per request by poc user

### Problem

The NATS connector currently creates streams automatically when they don't exist, which can lead to unintended stream creation in production environments. This poses security and operational risks:
Streams may be created with default configurations that don't match production requirements
Typos in stream names could result in unwanted streams being created
No explicit control over when RisingWave should have stream creation permissions

### Solution

This PR introduces a new boolean configuration parameter allow_create_stream that provides explicit control over stream creation behavior:
Default: allow_create_stream = false - RisingWave will NOT create streams automatically
Explicit permission: Users must set allow_create_stream = true to enable stream creation
Clear error messaging: When a stream doesn't exist and creation is disabled, users get a helpful error message

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/src/connector_common/common.rs
+++ b/src/connector/src/connector_common/common.rs
@@ -833,6 +833,9 @@ pub struct NatsCommon {
     #[serde(rename = "max_message_size")]
     #[serde_as(as = "Option<DisplayFromStr>")]
     pub max_message_size: Option<i32>,
+    #[serde(rename = "allow_create_stream", default)]
+    #[serde_as(as = "DisplayFromStr")]
+    pub allow_create_stream: bool,
 }
 
 impl EnforceSecret for NatsCommon {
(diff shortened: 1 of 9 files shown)


src/connector/src/connector_common/common.rs
```diff
@@ -833,6 +833,9 @@ pub struct NatsCommon {
     #[serde(rename = "max_message_size")]
     #[serde_as(as = "Option<DisplayFromStr>")]
     pub max_message_size: Option<i32>,
+    #[serde(rename = "allow_create_stream", default)]
+    #[serde_as(as = "DisplayFromStr")]
+    pub allow_create_stream: bool,
```
xiangjinwu, line 838: Will existing sources created before this PR also default to false, leading to them unable to create the stream automatically?

Similar issue: #22206

tabVersion: The issue happens with little possibility. 

For most existing streaming job, once they start the source, there must be a stream running. If the cluster experience a recovery, it will not hit the check here, it gets the stream directly without creating one.

One exception is, drop the stream and rely on RisingWave to recreate it. I think it is a wrong usage and we can ignore the case. Because creating a stream materializes some data on Nats server, this should be handled with caution.

xiangjinwu (approved):

PR #22320, opened 2025-06-23
fix(stream): truncate flushed chunks which were read

kwannoel:

### Bug fix

Chunks residing in the logstore buffer may also reside in storage if they were flushed. When reading these chunks we should truncate the persisted chunk. This PR makes that improvement.

### Test fix

Separately, we also improve the reliability of the test. Before this PR we would use `row count` to tell when the MV has caught up to upstream. However, this is not a reliable or general mechanism. It's not general because it assumes the MV's workload is strictly increasing. It's also not reliable, because of the same reason. In our tests, we cycle through deletes and inserts, so there's a very slight chance that we hit the row count even though we haven't fully caught up with upstream:
```
| insert | delete | insert | delete | insert | delete
                          ^--------------- if logstore is processed till here,
                                           it will be seen as "caught up"
```

After we introduce offline realignment of unaligned join in the previous PR, we use it now in this PR, and will deterministically enforce realignment.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/tests/simulation/tests/integration_tests/log_store/utils.rs
+++ b/src/tests/simulation/tests/integration_tests/log_store/utils.rs
@@ -136,7 +136,7 @@ pub(crate) async fn assert_lag_in_log_store(
     bail!("there was no lag in the logstore")
 }
 
-pub(crate) async fn assert_no_lag_in_log_store(cluster: &mut Cluster, name: &str) -> Result<()> {
+pub(crate) async fn assert_no_lag_in_log_store(cluster: &mut Cluster, name: &str, result_count: usize) -> Result<()> {
     let mut session = cluster.start_session();
     let query = format!("SELECT COUNT(*) FROM {name}");
     let result = session.run(query).await?;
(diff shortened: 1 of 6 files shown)


kwannoel:
* **#22364** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22364?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#22320** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22320?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22320?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#22331** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22331?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

wenym1 (approved):
LGTM

Seems the fix indicated by the PR title is already included in the previous PR. May update this PR title.

kwannoel:
> LGTM
> 
> Seems the fix indicated by the PR title is already included in the previous PR. May update this PR title.

Looks like I mistakenly include the fix in the previous PR. Updated.

src/tests/simulation/tests/integration_tests/log_store/utils.rs
```diff
@@ -136,34 +136,34 @@ pub(crate) async fn assert_lag_in_log_store(
...
+
+    cluster.simple_kill_nodes(&nodes).await;
+    cluster.simple_restart_nodes(&nodes).await;
+
+    // wait for recovery
+    cluster.wait_for_recovery().await?;
+
+    let mut session = cluster.start_session();
+    assert_parallelism_eq(&mut session, 10).await;
+
```
kwannoel:
```suggestion
```

PR #22326, opened 2025-06-23
feat: Use ActorAlignmentMapping instead of WorkerSlotMapping in the scheduler

shanicky:

This PR attempts to use a new `ActorAlignmentId`, which is `(worker_id, actor_idx)`, instead of the original `WorkerSlotId` `(worker_id, slot_idx)` in the scheduler. This way, an `ActorAlignmentMapping` can accommodate the mapping relationships for Worker, Actor, and Vnode all at once, enabling more precise control and laying the groundwork for future PRs.

In future plans, WorkerSlotMapping will be gradually phased out.

## Checklist

### DIFF
--- a/src/common/src/hash/consistent_hash/mapping.rs
+++ b/src/common/src/hash/consistent_hash/mapping.rs
@@ -400,10 +400,6 @@ impl ActorMapping {
     ) -> ActorAlignmentMapping {
         let mut worker_actors = HashMap::new();
 
-        let challenge = self.iter_unique().collect_vec();
-
-        assert!(challenge.is_sorted());
-
         for (idx, actor_id) in self.iter_unique().enumerate() {
             let worker_id = actor_to_worker
                 .get(&actor_id)


src/common/src/hash/consistent_hash/mapping.rs
```diff
@@ -31,6 +31,51 @@ use crate::util::iter_util::ZipEqDebug;
...
+#[derive(Clone, Copy, PartialEq, Eq, Hash, Ord, PartialOrd)]
+pub struct ActorAlignmentId(u64);
+
+impl ActorAlignmentId {
+    pub fn worker_id(&self) -> u32 {
+        (self.0 >> 32) as u32
+    }
+
+    pub fn actor_idx(&self) -> u32 {
+        self.0 as u32
```
yezizp2012, line 43: Please add some comments, as the `as` operation here is somewhat implicit.

yezizp2012 (approved):
lgtm

PR #22327, opened 2025-06-23
feat: introduce rwadmin user for cloud control plane

yezizp2012:

Related #22260 . In this PR, we introduced a new initial super user `rwadmin`, dedicated solely to the cloud control plane. This user cannot be altered or dropped, except for password changes made by the user itself. Note that:
1. For newly initialized clusters, `rwadmin` will be created automatically together with default superusers `root` and `postgres`. 
2. For existing clusters, to ensure security (mainly for non-cloud clusters), `rwadmin` is not automatically created, but we can still use other superusers to create it. Following that, the cloud control plane can utilize it for login and periodically update its password. Cc @wjf3121 

Additionally, I think that in the official quick start documentation, a security reminder should be added that when using RisingWave in a prod environment, the default superusers' password should be changed. @hengm3467 

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/meta/model/migration/src/m20230908_072257_init.rs
+++ b/src/meta/model/migration/src/m20230908_072257_init.rs
@@ -931,6 +931,14 @@ impl MigrationTrait for Migration {
                 true.into(),
                 true.into(),
             ])
+            .values_panic([
+                3.into(),
+                "rwadmin".into(),
+                true.into(),
+                true.into(),
+                true.into(),
+                true.into(),
+            ])
             .to_owned();
 
         // Since User table is newly created, we assume that the initial user id of `root` is 1 and `postgres` is 2.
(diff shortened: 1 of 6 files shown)


src/meta/model/migration/src/m20230908_072257_init.rs
```diff
@@ -931,6 +931,14 @@ impl MigrationTrait for Migration {
...
+                3.into(),
+                "rwadmin".into(),
+                true.into(),
+                true.into(),
+                true.into(),
+                true.into(),
+            ])
             .to_owned();
 
         // Since User table is newly created, we assume that the initial user id of `root` is 1 and `postgres` is 2.
```
BugenZhao: Shall we update the comment as well?

PR #22331, opened 2025-06-23
feat(meta): support offline realignment of unaligned join

kwannoel:

Only used for testing: https://github.com/risingwavelabs/risingwave/pull/22320.

I don't plan to make it user-facing (yet).

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/meta/service/src/stream_service.rs
+++ b/src/meta/service/src/stream_service.rs
@@ -36,6 +36,7 @@ use risingwave_pb::meta::table_fragments::PbState;
 use risingwave_pb::meta::table_fragments::actor_status::PbActorState;
 use risingwave_pb::meta::table_fragments::fragment::PbFragmentDistributionType;
 use risingwave_pb::meta::*;
+use risingwave_pb::stream_plan::stream_node::NodeBody;
 use tonic::{Request, Response, Status};
 
 use crate::barrier::{BarrierScheduler, Command};
@@ -538,6 +539,35 @@ impl StreamManagerService for StreamServiceImpl {
             .await?;
 
         Ok(Response::new(AlterConnectorPropsResponse {}))
+    }
+
+    async fn set_sync_log_store_aligned(
+        &self,
+        request: Request<SetSyncLogStoreAlignedRequest>,
+    ) -> Result<Response<SetSyncLogStoreAlignedResponse>, Status> {
+        let req = request.into_inner();
+        let job_id = req.job_id;
+        let aligned = req.aligned;
+
+        self.metadata_manager
+            .catalog_controller
+            .mutate_fragments_by_job_id(
+                job_id as _,
+                |_mask, stream_node| {
+                    if let Some(body) = &mut stream_node.node_body
+                        && let NodeBody::SyncLogStore(sync_log_store) = body
+                    {
+                        sync_log_store.aligned = aligned;
+                        true
+                    } else {
+                        false
+                    }
+                },
+                "no fragments found with synced log store",
+            )
+            .await?;
+
+        Ok(Response::new(SetSyncLogStoreAlignedResponse {}))
     }
 }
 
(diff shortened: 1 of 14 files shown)


kwannoel:
* **#22364** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22364?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#22320** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22320?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#22331** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22331?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22331?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/stream/src/executor/sync_kv_log_store.rs
```diff
@@ -550,6 +554,66 @@ impl<S: StateStore> SyncedKvLogStoreExecutor<S> {
...
+            #[for_await]
+            for message in input {
+                match message? {
+                    Message::Barrier(barrier) => {
+                        // Truncate the logstore
+                        let post_seal = initial_write_state
+                            .seal_current_epoch(barrier.epoch.curr, progress.take());
+                        let update_vnode_bitmap =
+                            barrier.as_update_vnode_bitmap(self.actor_context.id);
+                        post_seal.post_yield_barrier(update_vnode_bitmap).await?;
```
wenym1: Call after `yield barrier`.

May also update the barrier progress to make the persisted progress consistent with the actual progress.

src/stream/src/executor/sync_kv_log_store.rs
```diff
@@ -550,6 +554,66 @@ impl<S: StateStore> SyncedKvLogStoreExecutor<S> {
         let mut pause_stream = first_barrier.is_pause_on_startup();
         let mut initial_write_epoch = first_write_epoch;
 
+        if self.aligned {
```
wenym1, line 557: When the lag is quite large, it's likely to be stuck on consuming all previous rows. Shall we at least wrap this branch with `instrument_await` so that we are aware of being stuck here.

We shall also add some metrics to monitor the lag.

src/meta/service/src/stream_service.rs
```diff
@@ -539,6 +541,35 @@ impl StreamManagerService for StreamServiceImpl {
...
+
+    async fn set_sync_log_store_aligned(
+        &self,
+        request: Request<SetSyncLogStoreAlignedRequest>,
+    ) -> Result<Response<SetSyncLogStoreAlignedResponse>, Status> {
+        let req = request.into_inner();
+        let job_id = req.job_id;
+        let aligned = req.aligned;
+
+        self.metadata_manager
```
wenym1, line 553: How do we ensure that the mutation takes effect? Or this ddl is expected to be offline and its user is aware of this?

wenym1:
> LGTM
> 
> > I don't plan to make it user-facing (yet).
> 
> If so, should we only enable the related logic when `cfg!(debug_assertions)`?

Or even remove the frontend related logic, and only expose it as a meta node rpc? We can trigger it via risectl.

src/stream/src/executor/sync_kv_log_store.rs
kwannoel: no need, only enable in debug mode.

src/meta/service/src/stream_service.rs
kwannoel: Entirely offline.

kwannoel:
> LGTM
> 
> > I don't plan to make it user-facing (yet).
> 
> If so, should we only enable the related logic when `cfg!(debug_assertions)`?

We only have `debug_assertions` enabled for development builds. I will use this method to gate this feature.

yezizp2012 (approved):

PR #22342, opened 2025-06-24
feat(storage): batch add flushed imm to uploader

wenym1:

Previously in `LocalHummockStorage`, we add the newly generated imm to uploader whenever on `flush`. This works for normal state table, where we only flush once in each epoch. However, for state of kv log store, we may flush multiple times on multiple chunks, which generates a large number of imm and meanwhile many `HummockEvent::ImmToUploader`, and causes too much pending events and unable to handle hummock event timely.

In this PR, we add a new option `upload_on_flush` in `NewLocalOptions`. When it's false, we won't always add the imm generated in `flush` to uploader. Instead, we will add them in a batch when their total size reaches `mem_table_spill_threshold`. We will also add the pending imms to uploader on `seal_current_epoch`.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/storage/src/hummock/store/version.rs
+++ b/src/storage/src/hummock/store/version.rs
@@ -141,11 +141,18 @@ pub enum VersionUpdate {
 #[derive(Clone)]
 pub struct StagingVersion {
     pending_imm_size: usize,
-    // newer data comes last
+    /// It contains the imms added but not sent to the uploader of hummock event handler.
+    /// It is non-empty only when `upload_on_flush` is false.
+    ///
+    /// It will be sent to the uploader when `pending_imm_size` exceed threshold or on `seal_current_epoch`.
+    ///
+    /// newer data comes last
     pub pending_imms: Vec<ImmutableMemtable>,
-    // newer data comes first
-    // Note: Currently, building imm and writing to staging version is not atomic, and therefore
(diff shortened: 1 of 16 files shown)


wenym1:
* **#22342** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22342?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22342?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/storage/src/hummock/store/version.rs
```diff
@@ -147,10 +140,13 @@ pub enum VersionUpdate {
 
 #[derive(Clone)]
 pub struct StagingVersion {
+    pending_imm_size: usize,
+    // newer data comes last
+    pub pending_imms: Vec<ImmutableMemtable>,
```
hzxa21, line 150: nits: add some documentations to explain the difference between `uploading_imms` and `pending_imms` for someone not super familier with how event handler and uploader works.

hzxa21 (approved):
LGTM. Thanks for the quick fix.

Li0k (approved):
LGTM, thanks for the fix

PR #22344, opened 2025-06-24
Refactor sqlparser operator ast

xiangjinwu:

Follow-up of #22333. Support (dispatch) most custom operators in binder, without tokenizer/parser being aware of the individual operator. This cleans up a lot of unnecessary token/operator enum variants.

A few more tokens are unnecessary but not removed in this PR yet. This is to make the review easier and focused.
* Arrow `->`
* Concat `||`
* ArrowAt `<@` and AtArrow `@>`
* ShiftLeft `<<` and ShiftRight `>>`

## Checklist

## Documentation

<details>

</details>


src/frontend/src/binder/expr/mod.rs
```diff
@@ -504,10 +504,10 @@ impl Binder {
...
             // Use the `bind_binary_op` path to handle the ALL|ANY pattern.
             let op = match (expr_type, negated) {
-                (ExprType::Like, false) => BinaryOperator::PGLikeMatch,
-                (ExprType::Like, true) => BinaryOperator::PGNotLikeMatch,
-                (ExprType::ILike, false) => BinaryOperator::PGILikeMatch,
-                (ExprType::ILike, true) => BinaryOperator::PGNotILikeMatch,
+                (ExprType::Like, false) => BinaryOperator::Custom("~~".to_owned()),
+                (ExprType::Like, true) => BinaryOperator::Custom("!~~".to_owned()),
+                (ExprType::ILike, false) => BinaryOperator::Custom("~~*".to_owned()),
+                (ExprType::ILike, true) => BinaryOperator::Custom("!~~*".to_owned()),
```
BugenZhao, line 510: Shall we consider finding some way to reuse these literals, or refactoring the code here to avoid converting `ExprType` back to AST?

xiangjinwu: This followed the PostgreSQL practice, where LIKE-without-ESCAPE-clause are treated as operator internally and allowed to work together with SOME/ALL. Notably `SIMILAR TO` is syntactically the same as `LIKE` but it cannot work with SOME/ALL in PostgreSQL.

https://www.postgresql.org/docs/17/functions-matching.html#FUNCTIONS-LIKE:~:text=the%20parser%20actually,be%20included%20there.

The proper fix for us would be to update the sqlparser, so that LIKE-without-ESCAPE-clause are already `BinaryOp::Custom` rather than `Expr::Like` in AST. Then we no longer need to convert `ExprType::Like` back to AST in binder. Let me do that in a separate PR.

PR #22352, opened 2025-06-25
refactor(jni_core): move JVM runtime to common crate

wcy-fdu:

Refactored the global JVM runtime logic from `jni_core` to common to avoid a circular dependency:
Previously, `HDFS` in object_store depended on jni_core for JVM initialization, but `jni_core` now needs to access object store for `OpendalSchemaHistory`(refer to https://github.com/risingwavelabs/risingwave/pull/22255). 
Moving the global JVM to common allows both crates to use it independently and resolves the circular dependency.

## Checklist

## Documentation

<details>

</details>


BugenZhao:
IIUC, the circular dependency is because we are to introduce dependency on `risingwave_object_store` in `risingwave_jni_core` for object store method implementation.

However, is it possible to split the declaration and the implementation of these methods? So that we can leave the implementation inside the `object_store` crate, which removes this dependency.

src/jni_core/src/lib.rs
```diff
@@ -64,6 +65,14 @@ use tracing_slf4j::*;
 pub static JAVA_BINDING_ASYNC_RUNTIME: LazyLock<Runtime> =
     LazyLock::new(|| tokio::runtime::Runtime::new().unwrap());
 
+// Initialize JVM and register native methods when the module is loaded
```
wenym1: Where do we call this method? Seems dangerous to call it for more than once, to register java binding methods again.

src/common/Cargo.toml
```diff
@@ -51,6 +52,7 @@ humantime = "2.2"
 itertools = { workspace = true }
 itoa = "1.0"
 jiff = "0.1.15"
+jni = { version = "0.21.1", features = ["invocation"] }
```
wenym1: Can we define the dependency in the workspace `Cargo.toml`?

wcy-fdu:
> IIUC, the circular dependency is because we are to introduce dependency on `risingwave_object_store` in `risingwave_jni_core` for object store method implementation.
> 
> However, is it possible to split the declaration and the implementation of these methods? So that we can leave the implementation inside the `object_store` crate, which removes this dependency.

Actually we are going to use `ObjectStoreImpl` in jni_core, because we want to write/read mysql schema history into the object store which RW used, so I think it's better to build a `ObjectStoreImpl` in jni_core, rather than split some method. 
This pr is preliminary preparations for https://github.com/risingwavelabs/risingwave/pull/22255, and you can refer to [opendal_schema_history.rs](https://github.com/risingwavelabs/risingwave/blob/ff10ff84c3d1332dce76ff9ff48263f1fca3463d/src/jni_core/src/opendal_schema_history.rs) for detail. It's still draft, but the overall logic can explain this PR.

BugenZhao:
> > IIUC, the circular dependency is because we are to introduce dependency on `risingwave_object_store` in `risingwave_jni_core` for object store method implementation.
> > However, is it possible to split the declaration and the implementation of these methods? So that we can leave the implementation inside the `object_store` crate, which removes this dependency.
> 
> Actually we are going to use `ObjectStoreImpl` in jni_core, because we want to write/read mysql schema history into the object store which RW used, so I think it's better to build a `ObjectStoreImpl` in jni_core, rather than split some method. This pr is preliminary preparations for #22255, and you can refer to [opendal_schema_history.rs](https://github.com/risingwavelabs/risingwave/blob/ff10ff84c3d1332dce76ff9ff48263f1fca3463d/src/jni_core/src/opendal_schema_history.rs) for detail. It's still draft, but the overall logic can explain this PR.

Yes, I've checked that linked PR. 

> Actually we are going to use `ObjectStoreImpl` in jni_core

I'm just a bit concerned that this might make the crate too heavy, especially since it's named `core` 🤣. I thought we should only include code that's very general for Java and Rust interoperability.

wcy-fdu:
> I'm just a bit concerned that this might make the crate too heavy, especially since it's named `core` 🤣. I thought we should only include code that's very general for Java and Rust interoperability.

I got your point. Previously I was also considering only expose some public api(like `put_object`, `get_object`) in jni_core rather than the `ObjectStoreImpl`, but think twice, we need to not only read/write into S3, but gcs/azure based on the platform. So I guess it's better to use `state_store_url` to construct a general object store. WDYT @hzxa21 

But anyway, to read and write the object store, we need to rely on the `object_store` crate in `jni_core`. We can discuss the implementation detail in follow up pr, this pr only removed the dependencies.

src/connector/src/sink/remote.rs
```diff
@@ -708,9 +709,7 @@ struct EmbeddedConnectorClient {
 
 impl EmbeddedConnectorClient {
     fn new() -> Result<Self> {
-        let jvm = JVM
-            .get_or_init()
-            .context("failed to create EmbeddedConnectorClient")?;
+        let jvm = JVM.get_or_init()?;
```
wenym1: May keep the original `.context` to provide the same error context?

src/common/src/global_jvm.rs
```diff
@@ -0,0 +1,44 @@
...
+
+pub static JVM: JavaVmWrapper = JavaVmWrapper;
+static INSTANCE: OnceLock<JavaVM> = OnceLock::new();
+static JVM_BUILDER: Mutex<Option<Box<dyn FnOnce() -> JavaVM + Send>>> = Mutex::new(None);
+
+pub struct JavaVmWrapper;
+
+impl JavaVmWrapper {
+    pub fn register_jvm_builder(&self, builder: Box<dyn FnOnce() -> JavaVM + Send>) {
+        *JVM_BUILDER.lock().unwrap() = Some(builder);
```
wenym1: May check the `Option` to assert that the builder is registered for at most once, or we add a warning log when the builder is already registered.

Multiple calls on `register_jvm_builder` can happen when running in standalone mode and multiple components run in a single process.

src/common/src/global_jvm.rs
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    /// Get the global singleton JVM instance, initializing it with the registered closure if not already initialized.
+    pub fn get_or_init(&self) -> anyhow::Result<&'static JavaVM> {
+        INSTANCE.get_or_try_init(|| {
+            let builder = JVM_BUILDER.lock().unwrap().take().ok_or_else(|| {
```
wenym1: When `get_or_try_init` returns `Err`, the `OnceLock` value won't be initialized, and later call on `get_or_try_init` will call the builder again, but then the `JVM_BUILDER` has been taken out.

 I am thinking that, maybe we should turn the type of builder from `FnOnce()` to `Fn()`, so that when calling the builder method, we don't need to `take` it, and then after the initial `Err` we can still call the registered builder method later.

PR #22363, opened 2025-06-25
feat: use proc_macro to generate allow_alter props

tabVersion:

so https://github.com/risingwavelabs/risingwave/pull/20780 can use the generate whitelist

Here’s a PR description based on your diff:

---

#### Overview

This PR introduces a new auto-generated Rust module, `allow_alter_on_fly_fields.rs`, to the RisingWave connector framework. This module provides a single source of truth for which connector configuration fields can be altered on the fly (without requiring a restart or redeployment).

#### Key Changes

- **New Auto-Generated File:**  
  - `src/connector/src/allow_alter_on_fly_fields.rs` is now generated via `./risedev generate-with-options`.
  - This file contains:
    - Static maps of source and sink connector type names to their `allow_alter_on_fly` fields.
    - Utility functions to query and validate which fields are alterable for each connector.
    - The file is marked as auto-generated and should not be edited manually.

- **Connector Field Annotation:**  
  - The `#[with_option(allow_alter_on_fly)]` attribute is added to relevant fields (e.g., `KafkaCommon::sync_call_timeout`), marking them as alterable on the fly.

- **Test Coverage:**  
  - Adds a test (`test_allow_alter_on_fly_fields_rust_up_to_date`) to ensure the generated Rust file is up-to-date with the codebase.

- **Sink/Source Config Refactoring:**  
  - Introduces wrapper config structs (e.g., `ElasticSearchConfig`, `OpenSearchConfig`, `SnowflakeConfig`) to support the new code generation and maintain compatibility.

- **Build/Dev Workflow:**  
  - Updates the `Makefile.toml` to include the new generation and test steps for the `allow_alter_on_fly` fields.

#### Motivation

- **Consistency:**  
  Centralizes the definition of alterable fields, reducing the risk of drift between documentation, code, and runtime behavior.

- **Automation:**  
  Ensures that changes to connector options are automatically reflected in the generated code and tests, improving maintainability.

- **Extensibility:**  
  Makes it easier to add or modify alterable fields for connectors in the future.

#### How to Update

- Run `./risedev generate-with-options` to regenerate the file after changing connector field annotations.

#### Backward Compatibility

- No breaking changes to existing connector APIs.
- The new module is additive and only affects dynamic configuration updates.

---

**Note:**  
Do not manually edit `allow_alter_on_fly_fields.rs`. Always use the code generation workflow.

---

Let me know if you want a more concise or detailed version, or if you want to highlight specific connectors or fields!

## Checklist

## Documentation

<details>

</details>


src/connector/src/changeable_fields.rs
```diff
@@ -0,0 +1,56 @@
...
+
+use std::collections::{HashMap, HashSet};
+use std::sync::LazyLock;
+
+/// Map of source connector names to their changeable field names
+pub static SOURCE_CHANGEABLE_FIELDS: LazyLock<HashMap<String, HashSet<String>>> = LazyLock::new(|| {
+    let mut map = HashMap::new();
+    // KafkaProperties
+    map.insert(
+        "KafkaProperties".to_owned(),
```
wenym1: The key of the generated map is the struct type name. In actual usage, it seems hard to access the map from a source (for example, when checking kafka source, we need to access its entry by `"KafkaProperties"`). 

Therefore, shall we change to use the canonical type (obtained by std::any::type_name<T>) of property struct as the key? And then we can use the `type_name` of the corresponding `SourceProperties` of the source to access the map.

src/connector/src/lib.rs
```diff
@@ -43,6 +43,10 @@ use duration_str::parse_std;
 use serde::de;
 
 pub mod aws_utils;
+
+#[rustfmt::skip]
+pub mod changeable_fields;
```
wenym1: In rust, it seems better to name with `mutable`.

xxchan: +1 for `mutable` (or maybe `alterable`)

However, there's still an ambiguity: alterable on the fly vs alterable via rebuild

src/connector/src/changeable_fields.rs
xxchan: I'm not sure about the final form, but better to restack #20780 on this PR to demonstrate how this is used

src/connector/src/lib.rs
tabVersion: prefer `allow_alter_on_fly`

src/connector/src/changeable_fields.rs
tabVersion: > Therefore, shall we change to use the canonical type (obtained by std::any::type_name) of property struct as the key? And then we can use the type_name of the corresponding SourceProperties of the source to access the map.

I'm afraid it is not that easy.  

Originally, the proc_macro writes every possible fields of the top level Props for each connector to the yaml file. 
This approach uses two steps:
1) add more marks to the fields in the yaml
2) loads the yaml and pick the changeable fields

So at the time of code_gen, it does know the canonical type in the definition code.

I think the `XXXProperty` struct names are sufficient, we can check whether the keys(struct name) contain the desired connector name.

tabVersion: > I'm not sure about the final form, but better to restack https://github.com/risingwavelabs/risingwave/pull/20780 on this PR to demonstrate how this is used

There is already a example field `sync_call_timeout`

you can see: 1) the changeable mark in the yaml (both source and sink);. 2) the field appears in the generated code, under the keys `KafkaProperties` for source and `KafkaConfig` for sink. 

So for usage, we check the keys contains the desired connector name, or introduce a macro to do the mapping.

wenym1: > So at the time of code_gen, it does know the canonical type in the definition code.

Just had a try for sources, and it worked. Pushed a commit [c582bf8](https://github.com/risingwavelabs/risingwave/pull/22363/commits/c582bf87ef95adc2ead982769e8f628023a1676f)

The basic idea is to write `std::any::type_name<$struct_name>` as the key of the map, so that during code-gen we don't need to know the canonical type name.

At the moment, the idea only works for source, where we include the type name of source prop in the `for_all_sources` macro. To port the similar idea to sink, we may also need some way to define something like a config type for all sinks.

wenym1:
In [91b5211](https://github.com/risingwavelabs/risingwave/pull/22363/commits/91b52110867996684dbea56a94699ce7a645026a), we may need to update the code of code-gen from changeable to allow_alter_on_the_fly.

src/connector/src/sink/mod.rs
```diff
@@ -114,50 +114,77 @@ macro_rules! for_all_sinks {
...
+    };
+}
+
+// Convenience macro that uses for_all_sinks
+#[macro_export]
+macro_rules! use_all_sink_configs {
+    () => {
+        $crate::for_all_sinks! { $crate::generate_config_use_clauses }
+    };
+}
```
tabVersion, line 183: @wenym1 can you help on it?

wenym1 (approved):
LGTM!

PR #22375, opened 2025-06-26
test(e2e): Add Kafka Startup Timestamp Test

tabVersion:

This PR introduces a new end-to-end test to verify the Kafka source's startup timestamp functionality in RisingWave. The test consists of:
A Python script (kafka_startup_millis.py) that:
Produces initial records to a Kafka topic.
Waits for a short interval, then records the current timestamp in milliseconds.
Produces additional records after the timestamp.
Creates a RisingWave table with scan.startup.mode = 'timestamp' and scan.startup.timestamp_millis set to the recorded value.
Verifies that only records produced after the specified timestamp are ingested.
Cleans up by dropping the test table.
An SLT test file (kafka_startup_millis.slt) that:
Creates and deletes the Kafka topic.
Runs the Python script as part of the test flow.
This test ensures that RisingWave's Kafka connector correctly honors the scan.startup.timestamp_millis property, only ingesting records with timestamps greater than or equal to the specified value.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/e2e_test/source_inline/kafka/kafka_startup_millis.py
+++ b/e2e_test/source_inline/kafka/kafka_startup_millis.py
@@ -111,10 +111,7 @@ def main():
 
             # Expected: Should only see records with timestamps >= current_timestamp_millis
             expected_count = len([r for r in additional_records])
-            if count_result == expected_count:
-                print(f"✅ Test PASSED: Got expected count {expected_count}")
-            else:
-                print(f"❌ Test FAILED: Expected count {expected_count}, got {count_result}")
+            assert count_result == expected_count, f"❌ Test FAILED: Expected count {expected_count}, got {count_result}"
 
         finally:
             # Step 7: Drop the table


e2e_test/requirements.txt
```diff
@@ -1,7 +1,7 @@
 -r source_legacy/requirements.txt
 -r source_inline/requirements.txt
 requests
-confluent_kafka==2.6.2
+confluent_kafka==2.10.1
```
tabVersion, line 5: bump the version here because v2.6.2 is yanked

https://pypi.org/project/confluent-kafka/2.10.1/#history

e2e_test/source_inline/kafka/kafka_startup_millis.py
```diff
@@ -0,0 +1,136 @@
...
+            print(f"Records in table:")
+            for record in records:
+                print(f"  {record}")
+
+            # Expected: Should only see records with timestamps >= current_timestamp_millis
+            expected_count = len([r for r in additional_records])
+            if count_result == expected_count:
+                print(f"✅ Test PASSED: Got expected count {expected_count}")
+            else:
+                print(f"❌ Test FAILED: Expected count {expected_count}, got {count_result}")
```
xxchan: You don't have assertion here. The test will always pass

Issue #22328, opened 2025-06-23
TranslateApplyRule produced bloated plan when there are multiple subqueries in a select


xxchan:
### Describe the bug
```
CREATE TABLE table_a (id INT, data_a TEXT);
CREATE TABLE table_b (id INT, data_b TEXT);

explain WITH input AS (SELECT 1 as test_id, 'test' as item)
SELECT 
    test_id,
    item,
    ARRAY(
        SELECT a.data_a 
        FROM table_a a 
        WHERE a.id = input.test_id
        ORDER BY a.data_a
    ) AS array_from_a,
    ARRAY(
        SELECT b.data_b
        FROM table_b b 
        WHERE b.id = input.test_id
        ORDER BY b.data_b
    ) AS array_from_b
FROM input;
```

In the plan, we can see we have a duplicate `BatchScan(table_a)`, which is not ideal.

```
 BatchExchange { order: [], dist: Single }
 └─BatchHashJoin { type: LeftOuter, predicate: 1:Int32 IS NOT DISTINCT FROM 1:Int32 }
   ├─BatchHashJoin { type: LeftOuter, predicate: 1:Int32 IS NOT DISTINCT FROM 1:Int32 }
   │ ├─BatchExchange { order: [], dist: HashShard(1:Int32) }
   │ │ └─BatchValues { rows: [[1:Int32, 'test':Varchar]] }
   │ └─BatchProject { exprs: [1:Int32, Coalesce(array_agg(table_a.data_a order_by(table_a.data_a ASC)) filter(IsNotNull(1:Int32)), ARRAY[]:List(Varchar)) as $expr1] }
   │   └─BatchHashAgg { group_key: [1:Int32], aggs: [array_agg(table_a.data_a order_by(table_a.data_a ASC)) filter(IsNotNull(1:Int32))] }
   │     └─BatchHashJoin { type: LeftOuter, predicate: 1:Int32 IS NOT DISTINCT FROM table_a.id }
   │       ├─BatchHashAgg { group_key: [1:Int32], aggs: [] }
   │       │ └─BatchExchange { order: [], dist: HashShard(1:Int32) }
   │       │   └─BatchValues { rows: [[1:Int32]] }
   │       └─BatchExchange { order: [], dist: HashShard(table_a.id) }
   │         └─BatchProject { exprs: [table_a.id, table_a.data_a, table_a.data_a, 1:Int32] }
   │           └─BatchFilter { predicate: IsNotNull(table_a.id) }
[...]

PR #22386, opened 2025-06-26
feat(optimizer): optimize select clause's subquery unnesting

chenzl25:

- Resolve https://github.com/risingwavelabs/risingwave/issues/22328
- Optimize subquery unnesting for case when expressions. Users would likely use multiple subqueries in select clauses when case-when expressions are used, because each arm of the case-when expression could have its subqueries. Previously, we built an apply operator for each of these subqueries, which would increase the subquery unnesting algorithm complexity quite a lot. However, if we consider more carefully, we would find that all of them could share the same apply operator! Because these subqueries are in the same select clause and they are scalar subqueries. We can use a cross join to join these scalar subqueries without violating the max-one-row restriction. After these optimizations, we can significantly speed up the subquery unnesting.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/optimizer/plan_visitor/cardinality_visitor.rs
+++ b/src/frontend/src/optimizer/plan_visitor/cardinality_visitor.rs
@@ -178,7 +178,7 @@ impl PlanVisitor for CardinalityVisitor {
             // TODO: refine the cardinality of full outer join
             JoinType::FullOuter => {
                 if left.is_at_most(1) && right.is_at_most(1) {
-                    left.mul(right)
+                    Cardinality::new(0, 1)
                 } else {
                     Cardinality::unknown()
                 }


BugenZhao (approved):
LGTM!

src/frontend/src/planner/select.rs
```diff
@@ -413,6 +428,139 @@ impl Planner {
...
+    ///
+    /// The cross join way only meaningful when there are multiple subqueries
+    ///
+    ///            Apply
+    ///           /    \
+    ///         left   CrossJoin
+    ///                   /   \
+    ///           Subquery1   CrossJoin
+    ///                         /   \
+    ///                 Subquery2   Subquery3
```
xiangjinwu, line 452:
```
    ///            Apply
    ///           /    \
    ///         left   CrossJoin
    ///                   /   \
    ///           CrossJoin  Subquery3
    ///              /   \
    ///    Subquery1   Subquery2
```
Looks like the CrossJoin part is left-deep, according to the implementation below
https://github.com/risingwavelabs/risingwave/blob/1d461bd1a039994836d7c944a84ab0e4a77b13d1/src/frontend/src/planner/select.rs#L529-L538

src/frontend/src/planner/select.rs
```diff
@@ -413,6 +428,139 @@ impl Planner {
...
+                SubqueryKind::Array => subroot.into_array_agg()?,
+                _ => bail_not_implemented!(issue = 1343, "{:?}", subquery.kind),
+            };
+            if right.is_none() {
+                right = Some(subplan);
+            } else {
+                right = Some(LogicalJoin::create(
+                    right.clone().unwrap(),
+                    subplan,
+                    JoinType::Inner,
```
xiangjinwu: For scalar subquery, the result may be 0 rows and should be replaced by 1 row of null. This can be problematic in such inner cross join:
```
create table a (a1 int);
create table b (b1 int);
insert into b values (2);
select 1+2, (select a1 from a);
select 1+2, (select b1 from b);
select 1+2, (select a1 from a), (select b1 from b); -- expect (3,null,2) actual (3,null,null)
```

src/frontend/src/planner/select.rs
```diff
@@ -413,6 +430,140 @@ impl Planner {
...
+    /// ```text
+    ///            Apply
+    ///           /    \
+    ///         left   CrossJoin
+    ///                   /   \
+    ///           Subquery1   CrossJoin
+    ///                         /   \
+    ///                 Subquery2   Subquery3
+    /// ```
+    /// Typically, this is used for scalar subqueries in select clauses, because users might write subqueries in a case-when exprs multiple times.
```
xxchan: nit about the comment: The case doesn't need to involve case-when, e.g., `select subquery1, subquery2`. 

(I thought this is tricky, but in PG we cannot refer to the alias in the same SELECT, so seems not as hard as I imagined..)

src/frontend/src/planner/select.rs
chenzl25: Yes. The cross join part is left-deep. The left-deep is more about the apply operator in terms of subquery unnesting

xxchan (commented):
L🐮TM

src/frontend/src/planner/select.rs
chenzl25: So it should be a full outer join🤔

src/frontend/src/optimizer/plan_visitor/cardinality_visitor.rs
```diff
@@ -176,7 +176,11 @@ impl PlanVisitor for CardinalityVisitor {
             JoinType::RightSemi | JoinType::RightAnti => right.min(0..),
 
             // TODO: refine the cardinality of full outer join
-            JoinType::FullOuter => Cardinality::unknown(),
+            JoinType::FullOuter => if left.is_at_most(1) && right.is_at_most(1) {
+                left.mul(right)
+            } else {
+                Cardinality::unknown()
+            },
```
chenzl25: cc @BugenZhao Improve it a bit. PTAL

src/frontend/src/planner/select.rs
```diff
@@ -413,6 +430,140 @@ impl Planner {
...
+            };
+            if right.is_none() {
+                right = Some(subplan);
+            } else {
+                right = Some(LogicalJoin::create(
+                    right.clone().unwrap(),
+                    subplan,
+                    JoinType::FullOuter,
+                    ExprImpl::literal_bool(true),
+                ));
```
chenzl25, line 537: Use a full outer join to fix the null issue. cc @xiangjinwu

src/frontend/src/optimizer/plan_visitor/cardinality_visitor.rs
xiangjinwu: Same error: When left is 0..=0 and right is 1..=1, their full outer cardinality would be 1, which is not in the range of 0..=0

chenzl25: I think we can hard-code it to 0..1🤔

src/frontend/src/optimizer/plan_visitor/cardinality_visitor.rs
```diff
@@ -176,7 +176,13 @@ impl PlanVisitor for CardinalityVisitor {
...
 
             // TODO: refine the cardinality of full outer join
-            JoinType::FullOuter => Cardinality::unknown(),
+            JoinType::FullOuter => {
+                if left.is_at_most(1) && right.is_at_most(1) {
+                    Cardinality::new(0, 1)
+                } else {
+                    Cardinality::unknown()
+                }
+            }
```
chenzl25, line 186: Hardcode it to 0..1

PR #22388, opened 2025-06-26
fix(connector): correctly handle the high watermark offset in rdkafka

### LINKED ISSUE #22387: correctly handle RD_KAFKA_OFFSET_END
### Describe the bug

The behavior of rdkafka is as follows: if a specific time is provided but no messages have arrived by that time, rdkafka  `offsets_for_times` will return an offset of -1, which is a constant representing the latest offset, i.e., latest.

```rust
pub const RD_KAFKA_OFFSET_END: i32 = -1;
```

However, when we use this offset via the `elem.offset()` method, this method converts the offset into an enum, resulting in `Offset::End` instead of `Offset::Offset(xxxx)`. 

```
   pub fn from_raw(raw_offset: i64) -> Offset {
        match raw_offset {
            OFFSET_BEGINNING => Offset::Beginning,
            OFFSET_END => Offset::End, // -1 => Offset::End
            OFFSET_STORED => Offset::Stored,
            OFFSET_INVALID => Offset::Invalid,
            n if n <= OFFSET_TAIL_BASE => Offset::OffsetTail(-(n - OFFSET_TAIL_BASE)),
            n => Offset::Offset(n),
        }
    }
```

This causes the match statement to fall into the default branch. 

```    match elem.offset() {
                Offset::Offset(offset) => {
                    // XXX(rc): currently in RW source, `offset` means the last consumed offset, so we need to subtract 1
                    result.insert(elem.partition(), Some(offset - 1));
                }
                _ => {
                    let (_, high_watermark) = self
                        .client
                        .fetch_watermarks(
                            self.topic.as_str(),
                            elem.partition(),
                            self.sync_call_timeout,
                        )
                        .await?;
                    result.insert(elem.partition(), Some(high_watermark));
                }
            }
```
Then, when calling `high_watermark`, we get the high watermark offset. During processing, we add 1 to this offset, which results in assigning an offset that doesn't exist. Consequently, the consumer resets back to the earliest offset.

```
025-06-26T23:54:52.131117+08:00 DEBUG           rw-streaming actor{otel.name="Actor 26" actor_id=26 prev_epoch=8763348882882560 curr_epoch=8763348882948096}:executor{otel.name="Source 1A00002716"}: risingwave_stream::executor::source::reader_stream: build stream source reader with state: Some([Kafka(KafkaSplit { topic: "t", partition: 0, start_offset: Some(2), stop_offset: None })])
2025-06-26T23:54:52.132267+08:00  WARN ThreadId(115) librdkafka: librdkafka: OFFSET [thrd:main]: t [0]: offset reset (at offset 3 (leader epoch -1), broker 0) to offset BEGINNING (leader epoch -1): fetch failed due to requested offset not available on the broker: Broker: Offset out of range
```

### Error message/log

```text

```

### To Reproduce

_No response_

### Expected behavior

_No response_

_No response_

### The version of RisingWave

_No response_

### Additional context

_No response_

shanicky:

close #22387 

## Checklist

### DIFF
--- a/src/connector/src/source/kafka/enumerator.rs
+++ b/src/connector/src/source/kafka/enumerator.rs
@@ -454,7 +454,7 @@ impl KafkaSplitEnumerator {
                             self.sync_call_timeout,
                         )
                         .await?;
-                    result.insert(elem.partition(), Some(high_watermark));
+                    result.insert(elem.partition(), Some(high_watermark - 1));
                 }
             }
         }


tabVersion:
close for fix via https://github.com/risingwavelabs/risingwave/pull/22413

PR #22414, opened 2025-06-27
feat(streaming): add chunk compaction post-reader, over logstore stream

kwannoel:

JDBC sinks rely on the invariant that chunk boundaries are respected, and incoming chunks only have a single insert / delete operation on a key.

This is because they do batch processing, where they sequence deletes before inserts.

If we have the following sequence of events in a batch:
```
- 1 1 3
+ 1 1 4
- 1 1 4
```

Where `1` is the pk, it will get re-batched into:
```
deletes
- 1 1 3
- 1 1 4

inserts
+ 1 1 4
```

And we end up with an extra `+ 1 1 4` record.

This can happen when there's a logstore lag, and recovery happens. Then when we read from the logstore stream, we don't respect the chunk boundaries, so previously separate chunks can end up in the same chunk, if they are smaller than the chunk max batch size.

This PR fixes it by running another compact after the sink output.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/common/log_store_impl/kv_log_store/reader.rs
+++ b/src/stream/src/common/log_store_impl/kv_log_store/reader.rs
@@ -45,7 +45,6 @@ use tokio::sync::{oneshot, watch};
 use tokio::time::sleep;
 use tokio_stream::StreamExt;
 
-use crate::common::compact_chunk::merge_chunk_row;
 use crate::common::log_store_impl::kv_log_store::buffer::{
     LogStoreBufferItem, LogStoreBufferReceiver,
 };
@@ -191,12 +190,9 @@ pub struct KvLogStoreReader<S: StateStoreRead> {
     identity: String,
 
     rewind_delay: RewindDelay,
-
-    downstream_pk_indices: Vec<usize>,
 }
 
 impl<S: StateStoreRead> KvLogStoreReader<S> {
-    #[expect(clippy::too_many_arguments)]
     pub(crate) fn new(
         state: LogStoreReadState<S>,
         rx: LogStoreBufferReceiver,
@@ -205,7 +201,6 @@ impl<S: StateStoreRead> KvLogStoreReader<S> {
         metrics: KvLogStoreMetrics,
         is_paused: watch::Receiver<bool>,
         identity: String,
-        downstream_pk_indices: Vec<usize>,
     ) -> Self {
         let rewind_delay = RewindDelay::new(&metrics);
         Self {
@@ -221,7 +216,6 @@ impl<S: StateStoreRead> KvLogStoreReader<S> {
             is_paused,
             identity,
             rewind_delay,
-            downstream_pk_indices,
         }
     }
 }
@@ -376,8 +370,82 @@ impl<S: StateStoreRead> KvLogStoreReader<S> {
             range_start,
         )
     }
+}
 
-    async fn next_item_inner(&mut self) -> LogStoreResult<(u64, LogStoreReadItem)> {
+impl<S: StateStoreRead> LogReader for KvLogStoreReader<S> {
+    async fn start_from(&mut self, start_offset: Option<u64>) -> LogStoreResult<()> {
+        // init or rewind must be executed before start_from.
+        let aligned_range_start =
+            if let KvLogStoreReaderFutureState::Reset(aligned_range_start) = &self.future_state {
+                aligned_range_start
+            } else {
+                panic!("future state is not Reset");
+            };
+
+        // Construct the log reader's read stream based on start_offset, aligned_range_start or persisted_epoch.
+        let range_start = match (start_offset, aligned_range_start) {
+            (Some(rewind_start_offset), _) => {
+                tracing::info!(
+                    "Sink error occurred. Rebuild the log reader stream from the rewind start offset returned by the coordinator."
+                );
+                LogStoreReadStateStreamRangeStart::LastPersistedEpoch(rewind_start_offset)
+            }
+            (None, Some(aligned_range_start)) => aligned_range_start.clone(),
+            (None, None) => {
+                // still consuming persisted state store data
+                let persisted_epoch =
+                    self.truncate_offset
+                        .map(|truncate_offset| match truncate_offset {
+                            TruncateOffset::Chunk { epoch, .. } => epoch.prev_epoch(),
+                            TruncateOffset::Barrier { epoch } => epoch,
+                        });
+
+                match persisted_epoch {
+                    Some(last_persisted_epoch) => {
+                        LogStoreReadStateStreamRangeStart::LastPersistedEpoch(last_persisted_epoch)
+                    }
+                    None => LogStoreReadStateStreamRangeStart::Unbounded,
+                }
+            }
+        };
+        self.future_state = KvLogStoreReaderFutureState::ReadStateStoreStream(
+            self.read_persisted_log_store(range_start).await?,
+        );
+        self.rx.rewind(start_offset);
+        Ok(())
+    }
+
+    async fn init(&mut self) -> LogStoreResult<()> {
+        if let Some(init_epoch_rx) = self.init_epoch_rx.take() {
+            let init_epoch = init_epoch_rx
+                .await
+                .map_err(|_| anyhow!("should get the first epoch"))?;
+            let first_write_epoch = init_epoch.curr;
+
+            assert_eq!(
+                self.first_write_epoch.replace(first_write_epoch),
+                None,
+                "should not init twice"
+            );
+        } else {
+            let (new_vnode_bitmap, write_epoch) = self
+                .update_vnode_bitmap_rx
+                .recv()
+                .await
+                .ok_or_else(|| anyhow!("failed to receive update vnode"))?;
+            self.state.serde.update_vnode_bitmap(new_vnode_bitmap);
+            self.first_write_epoch = Some(write_epoch);
+        };
+
+        self.future_state = KvLogStoreReaderFutureState::Reset(None);
+        self.latest_offset = None;
+        self.truncate_offset = None;
+        self.rewind_delay = RewindDelay::new(&self.metrics);
+
+        Ok(())
+    }
+
+    async fn next_item(&mut self) -> LogStoreResult<(u64, LogStoreReadItem)> {
         while *self.is_paused.borrow_and_update() {
             info!("next_item of {} get blocked by is_pause", self.identity);
             self.is_paused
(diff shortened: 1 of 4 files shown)


kwannoel:
* **#22414** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22414?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22414?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

wenym1:
Seems that this fix is only for correctness of sink. Adding this logic in the kv log store may increase unnecessary overhead on log store executor.

Previously we do a merge chunk on the sink writer side. 
https://github.com/risingwavelabs/risingwave/blob/68f698d3853addd38843833ae97be35f2224475b/src/stream/src/executor/sink.rs#L517
But actually the requirement on a chunk being merged is on the reader side, more specifically, required by sinks. Therefore, shall we just move the merge chunk logic from sink writer side to the log reader side? We can conditionally do merge chunk on the following line.
https://github.com/risingwavelabs/risingwave/blob/68f698d3853addd38843833ae97be35f2224475b/src/stream/src/executor/sink.rs#L576

kwannoel:
> Seems that this fix is only for correctness of sink. Adding this logic in the kv log store may increase unnecessary overhead on log store executor.
> 
> Previously we do a merge chunk on the sink writer side. 
> https://github.com/risingwavelabs/risingwave/blob/68f698d3853addd38843833ae97be35f2224475b/src/stream/src/executor/sink.rs#L517
> But actually the requirement on a chunk being merged is on the reader side, more specifically, required by sinks. Therefore, shall we just move the merge chunk logic from sink writer side to the log reader side? We can conditionally do merge chunk on the following line.
> https://github.com/risingwavelabs/risingwave/blob/68f698d3853addd38843833ae97be35f2224475b/src/stream/src/executor/sink.rs#L576

Synced Logstore executor doesn't use this path for its reads. It's only used by sink executor.

If we remove writer compaction, it can degrade the performance under high load. Because we have a much larger amount of data being written. I think the overall time complexity is acceptable, it only hits this path after recovery. Normal logstore reads go through flushed chunk / buffer, which should preserve the chunk boundaries.

(I need to verify again if a merged flushed chunk is compacted in the write path)

kwannoel:
> > Seems that this fix is only for correctness of sink. Adding this logic in the kv log store may increase unnecessary overhead on log store executor.
> > Previously we do a merge chunk on the sink writer side.
> > https://github.com/risingwavelabs/risingwave/blob/68f698d3853addd38843833ae97be35f2224475b/src/stream/src/executor/sink.rs#L517
> > 
> > But actually the requirement on a chunk being merged is on the reader side, more specifically, required by sinks. Therefore, shall we just move the merge chunk logic from sink writer side to the log reader side? We can conditionally do merge chunk on the following line.
> > https://github.com/risingwavelabs/risingwave/blob/68f698d3853addd38843833ae97be35f2224475b/src/stream/src/executor/sink.rs#L576
> 
> Synced Logstore executor doesn't use this path for its reads. It's only used by sink executor.
> 
> If we remove writer compaction, it can degrade the performance under high load. Because we have a much larger amount of data being written. I think the overall time complexity is acceptable, it only hits this path after recovery. Normal logstore reads go through flushed chunk / buffer, which should preserve the chunk boundaries.
> 
> (I need to verify again if a merged flushed chunk is compacted in the write path)

> If we remove writer compaction, it can degrade the performance under high load. Because we have a much larger amount of data being written.

Retracting part of my comment as I misunderstood. I was thinking you meant to migrate all the logic in `process_msg` from the reader to the writer. But I guess you mean just the `merge_chunk` portion for upsert sink (the following logic is actually abstracted by `merge_chunk`):

https://github.com/risingwavelabs/risingwave/blob/68f698d3853addd38843833ae97be35f2224475b/src/stream/src/executor/sink.rs#L513-L522

In that case, I think Writer IO is unlikely to be bottleneck, we should not have many overlapping keys I suppose, since there's no PK mismatch. I do have other things to clarify though:

If we migrate all the compaction logic to the reader, do you mean to wrap all the output from the reader?

Compact chunks from all of the following:
- Flushed chunk reads
- Log store reads
- Buffer reads

I'm slightly concerned there will be some performance impact. We need to do some testing to evaluate. In the case of JDBC sinks, the downstream system should be the bottleneck, rather than the chunk processing part in RW. So I think any performance impact should be minimal.

> We can conditionally do merge chunk on the following line.

I suppose this means:
1. Check for sink type == JDBC || pg native sink.
2. If JDBC: Call merge_chunk

Further, we also have to make sure that no other sinks beside JDBC sinks rely on the property of only 1 op / key.

If it's as what I mentioned above, your idea generally LGTM. We probably just need to do some benchmarking to evaluate it.

chenzl25:
> Synced Logstore executor doesn't use this path for its reads. It's only used by sink executor.

IIRC, SyncedLogstore is only used by UnalignedJoin and we don't need to compact the chunk for SyncedLogstore, so put this logic into the sink executor seems reasonable.

src/stream/src/common/log_store_impl/kv_log_store/reader.rs
```diff
@@ -602,6 +534,106 @@ impl<S: StateStoreRead> LogReader for KvLogStoreReader<S> {
...
+        // as such we must do a compaction over the output,
+        // otherwise there's a risk that adjacent chunks with DELETEs on the same PK
+        // will have double DELETEs followed by unspecified sequence of INSERTs,
+        // and lead to inconsistent data downstream.
+        self.next_item_inner().await.map(|(epoch, item)| {
+            let item = match item {
+                LogStoreReadItem::StreamChunk { chunk, chunk_id }
+                    // downstream_pk_indices not empty
+                    // implies that the sink is a upsert sink
+                    if !self.downstream_pk_indices.is_empty() =>
```
kwannoel: In the writer branch, we previously had to hit the following conditions before compaction is triggered over upsert:
```
1. !(need_advance_delete || re_construct_with_sink_pk)
2. SinkType::Upsert => ...
```

As we can see, after moving to the `reader`, we only check:
```
                    if !self.downstream_pk_indices.is_empty() =>
```

This means in the reader we compact chunks from the following branches:
```
need_advance_delete || re_construct_with_sink_pk -- compacted in both writer and reader.
!(need_advance_delete || re_construct_with_sink_pk) -- previously compacted in writer, now reader.
upsert (!self.downstream_pk_indices.is_empty()) -- previously compacted in writer, now reader.
```

It's necessary to re-compact chunks from `need_advance_delete || re_construct_with_sink_pk` because the chunk boundaries are not respected when reading from logstore, and we encounter the same problems.

It's alright to have double compaction work as well. Because when upstream_pk != downstream_pk, the number of records with the same downstream_pk and different upstream_pk could be high. So this will greatly reduce the write cardinality if we compact in the writer.
Then, when we read in the reader, we resolve the separate problem of duplicate operations from different chunks merged into the same chunk.
This approach means we don't have too much logic change, and should have minimal performance impact as well.

src/stream/src/common/log_store_impl/kv_log_store/reader.rs
```diff
@@ -602,6 +534,106 @@ impl<S: StateStoreRead> LogReader for KvLogStoreReader<S> {
...
+        };
+
+        self.future_state = KvLogStoreReaderFutureState::Reset(None);
+        self.latest_offset = None;
+        self.truncate_offset = None;
+        self.rewind_delay = RewindDelay::new(&self.metrics);
+
+        Ok(())
+    }
+
```
kwannoel: Can ignore this diff, it's not actually newly added.

src/stream/src/common/log_store_impl/kv_log_store/reader.rs
```diff
@@ -370,82 +376,8 @@ impl<S: StateStoreRead> KvLogStoreReader<S> {
...
-            self.first_write_epoch = Some(write_epoch);
-        };
-
-        self.future_state = KvLogStoreReaderFutureState::Reset(None);
-        self.latest_offset = None;
-        self.truncate_offset = None;
-        self.rewind_delay = RewindDelay::new(&self.metrics);
-
-        Ok(())
-    }
```
kwannoel: Can ignore this diff

src/stream/src/common/log_store_impl/kv_log_store/reader.rs
```diff
@@ -602,6 +534,106 @@ impl<S: StateStoreRead> LogReader for KvLogStoreReader<S> {
...
+
+        self.future_state = KvLogStoreReaderFutureState::Reset(None);
+        self.latest_offset = None;
+        self.truncate_offset = None;
+        self.rewind_delay = RewindDelay::new(&self.metrics);
+
+        Ok(())
+    }
+
+    async fn next_item(&mut self) -> LogStoreResult<(u64, LogStoreReadItem)> {
```
kwannoel: We wrap the old implementation above, and compact the output chunks.

src/stream/src/common/log_store_impl/kv_log_store/reader.rs
```diff
@@ -370,82 +376,8 @@ impl<S: StateStoreRead> KvLogStoreReader<S> {
...
-        self.future_state = KvLogStoreReaderFutureState::Reset(None);
-        self.latest_offset = None;
-        self.truncate_offset = None;
-        self.rewind_delay = RewindDelay::new(&self.metrics);
-
-        Ok(())
-    }
-
-    async fn next_item(&mut self) -> LogStoreResult<(u64, LogStoreReadItem)> {
+    async fn next_item_inner(&mut self) -> LogStoreResult<(u64, LogStoreReadItem)> {
```
kwannoel: We move `next_item_inner` to the struct level impl of `KvLogStoreReader`, so we can wrap it and compact the chunks below. https://github.com/risingwavelabs/risingwave/pull/22414/files#r2174258503

src/stream/src/executor/sink.rs
```diff
@@ -572,8 +563,23 @@ impl<F: LogStoreFactory> SinkExecutor<F> {
...
         let mut log_reader = log_reader
             .transform_chunk(move |chunk| {
+                // NOTE(kwannoel):
+                // After the optimization in https://github.com/risingwavelabs/risingwave/pull/12250,
+                // `DELETE`s will be sequenced before `INSERT`s in JDBC sinks and PG rust sink.
+                // This is because there's a risk that adjacent chunks with `DELETE`s on the same PK
+                // will get merged into a single chunk, since the logstore format doesn't preserve
+                // chunk boundaries.
+                // Then we will have double `DELETE`s followed by unspecified sequence of `INSERT`s,
+                // and lead to inconsistent data downstream.
```
wenym1: Previously the condition to `merge_chunk_row` is `(!(need_advance_delete || re_construct_with_sink_pk)) && sink_type == SinkType::Upsert`, which is equivalent to  `!need_advance_delete && !re_construct_with_sink_pk && sink_type == SinkType::Upsert`, where 
```
let stream_key_sink_pk_mismatch = {
    stream_key
        .iter()
        .any(|i| !sink_param.downstream_pk.contains(i))
};
let need_advance_delete =
    stream_key_sink_pk_mismatch && sink_param.sink_type != SinkType::AppendOnly;
// NOTE(st1page): reconstruct with sink pk need extra cost to buffer a barrier's data, so currently we bind it with mismatch case.
let re_construct_with_sink_pk = need_advance_delete
    && sink_param.sink_type == SinkType::Upsert
    && !sink_param.downstream_pk.is_empty();
```

Since `let re_construct_with_sink_pk = need_advance_delete && (...)`, `!need_advance_delete` implies `!re_construct_with_sink_pk`.

`!need_advance_delete` is equivalent to `((stream_key matches downstream_key) or (sink_type == AppendOnly))`. `!need_advance_delete && sink_type == SinkType::Upsert` is equivalent to `(stream_key matches downstream_key) && SinkType::Upsert`  and I think this is the condition to `merge_chunk_row` here. This is reasonable, because when sink type is not upsert, we have only insert, and there is no delete and update to cause inconsistency, and when stream key does not match with downstream pk, on writer side we already have a barrier level compaction on downstream pk, and therefore no need to do extra compact here.

kwannoel: Think this is hard to reason about. In the future if anyone has to read this section of code they have to understand:
1. Why the condition is: `(stream_key matches downstream_key) && SinkType::Upsert`
2. Why this is necessary for downstream pg sinks / JDBC sinks (unavoidable).

I will change it and also add more docs to state (1).

kwannoel: The main upside is that we avoid the extra compact, when there's a pk mismatch / for some reason the sink is append only and has pk.

kwannoel: > The main upside is that we avoid the extra compact, when there's a pk mismatch / for some reason the sink is append only and has pk.

Not sure if it's worth it. I guess the other more important reason is we preserve the existing logic.

wenym1 (approved):
LGTM

Issue #22145, opened 2025-06-08
a session variable to control the encoding in join


lmatz:
### Is your feature request related to a problem? Please describe.

https://github.com/risingwavelabs/risingwave/issues/20017

Sometimes, the join does not have a large state, and the user has abundant memory to cache states, so we can minimize CPU consumption at the cost of more memory consumption.

When the join has large states, we can choose to minimize memory consumption.

The significance of CPU consumption of the current encoding is demonstrated in the slack link of the referred issue

PR #22416, opened 2025-06-28
feat: add a session variable to switch join encoding type

lmatz:

#22145 

`EncodedJoinRow` stays the same. 

We add a new `Unencoded`, aka `JoinRow<OwnedRow>`, which is a row that does not go through serialization and deserialization when putting into the operator cache and getting out of the operator cache. This is the same row we use in all kinds of operations in the Join operator.

~Use an enum to unify both types. Made a comment regarding the size of them on the stack.~

Add a session variable to switch between these two: 1. memory optimized (the current one), 2. CPU optimized (the new one).

## Checklist

## Documentation

<details>

</details>

This PR introduces a new session variable called `streaming_join_encoding`.
It has two options: `memory_optimized` and `cpu_optimized` to control which encoding of rows is used in the join operator.

The memory optimized encoding consumes more CPU but the size of cached rows used in the join operator is largely reduced, which means we can cache much more rows with the same amount of memory. We recommend users to choose this encoding when the state is much larger than the memory capacity and disk cache capacity. This is generally also when you observe high cache misses or high remote I/O on Grafana.

The cpu optimized encoding consumes much less CPU but the size of cached rows used in the join operator is larger. We recommend users to choose this encoding when the state is smaller than or roughly the same as the memory capacity and disk cache capacity. Depending on the complexity and structure of the streaming query, the improvement can be as large as 50%.


src/stream/src/executor/join/row.rs
```diff
@@ -80,3 +127,19 @@ impl EncodedJoinRow {
...
+
+    #[test]
+    fn test_cached_join_row_sizes() {
+        let enum_size = size_of::<CachedJoinRow>();
+        let encoded_size = size_of::<EncodedJoinRow>();
+        let unencoded_size = size_of::<JoinRow<OwnedRow>>();
+
+        assert_eq!(enum_size, 40);
+        assert_eq!(encoded_size, 40);
+        assert_eq!(unencoded_size, 24);
```
lmatz, line 155: 16 bytes extra overhead, I think it is acceptable. After all, this memory/cpu encoding tradeoff makes sense only when the row size is relatively large. They are similar when the row size is small.

src/stream/src/executor/join/row.rs
```diff
@@ -61,6 +68,44 @@ fn build_degree_row(order_key: impl Row, degree: DegreeType) -> impl Row {
     order_key.chain(row::once(Some(ScalarImpl::Int64(degree as i64))))
 }
 
+#[derive(Clone, Debug)]
+pub enum CachedJoinRow {
+    Encoded(EncodedJoinRow),
+    Unencoded(JoinRow<OwnedRow>),
```
lmatz: Just realized one possible further improvement is to make the clone of `JoinRow<OwnedRow>` a shallow copy instead of a deep copy, as `OwnedRow` is `Box<[Datum]>`. 

Is `Arc<Some_Lock<>>` the way to go?

BugenZhao: There's no need for `SomeLock` as row is designed to be immutable.

However, I'm still dubious if it's a good idea to switch `Box` into `Arc` for `OwnedRow`. In most cases we should avoid cloning a row as much as possible, instead of optimizing the cloning, considering that it introduces atomic operations every time a row is created or dropped.

For this specific case, instead of changing the internal of `OwnedRow`, I think you can try directly using `Arc<[Datum]>` here (by implementing `Row` trait for it). `Row` does not have to be an `OwnedRow`. 😄

e2e_test/streaming/join.slt
```diff
@@ -23,6 +23,24 @@ select v1, v2, v3, v10, v20, v30 from mv1;
 1 4 2 1 3 5
 2 3 3 2 4 6
 
+statement ok
+set streaming_join_encoding_type to memoryoptimized;
```
BugenZhao: Shall we name it

```suggestion
set streaming_join_encoding_optimized_for to memory;
```

src/stream/src/executor/join/row.rs
```diff
@@ -47,10 +48,16 @@ impl<R: Row> JoinRow<R> {
         (&self.row, degree)
     }
 
-    pub fn encode(&self) -> EncodedJoinRow {
-        EncodedJoinRow {
-            compacted_row: (&self.row).into(),
-            degree: self.degree,
+    pub fn encode<const N: JoinEncodingPrimitive>(&self) -> CachedJoinRow {
```
BugenZhao: By making `JoinEncodingPrimitive` a const generic we doubled the code to be compiled after monomorphization, but here we fit different variants back to a single `enum` type and use runtime dispatching when calling method on it. This doesn't seem to be a good deal. Could you try either...

- use generic `CachedJoinRow` here (like by associating a type to the `JoinEncoding` parameter)
- or simply use a boolean/enum field to indicate the encoding type?

src/stream/src/from_proto/hash_join.rs
```diff
@@ -222,18 +229,30 @@ impl<S: StateStore> HashKeyDispatcher for HashJoinExecutorDispatcherArgs<S> {
...
+                    | (JoinTypeProto::Unspecified, _)
+                    | (_, JoinEncodingTypeProto::Unspecified ) => unreachable!(),
+                    $(
+                        (JoinTypeProto::$proto, JoinEncodingTypeProto::MemoryOptimized) => build!($join_type, MemoryOptimized),
+                        (JoinTypeProto::$proto, JoinEncodingTypeProto::CpuOptimized) => build!($join_type, CPUOptimized),
+                    )*
+                }
+            };
+        }
+        build_match! {
```
BugenZhao, line 244: Seems that `$proto` is always the same as `$join_type`?

e2e_test/streaming/join.slt
lmatz: Sounds good

I struggled with the naming. I don't know if we will come up with any new encoding algorithm
I guess at most 3 makes sense? cpu optimized, memory optimized, and a balanced one, WDYT?

`set streaming_join_encoding_optimized_for to` `memory`, `cpu`, or `balanced`

src/stream/src/from_proto/hash_join.rs
lmatz: True

src/stream/src/executor/join/row.rs
lmatz: Got it, let's keep `Box` at the moment. Let me try to reduce the cloning in a separate PR.

src/stream/src/executor/join/row.rs
lmatz: updated, I thought the compiler was going to do some magic about the `const N` that only one code path will be kept for each instantiation of the function with respect to each choice of `N`

Now it is dispatched during compilation, 
But not sure if it is exactly what `use generic CachedJoinRow here (like by associating a type to the JoinEncoding parameter)` suggests

BugenZhao: Oh I don't mean this. I mean we should avoid introducing an enum to erase the type under different `JoinEncoding`. Instead, we should use different types to avoid any runtime dispatching. The trait can be like this:


```rust
pub trait JoinEncoding: 'static + Send + Sync {
    type EncodedRow;

    fn encode<R: Row>(row: &JoinRow<R>) -> Self::EncodeRow;
}
```

lmatz: updated

e2e_test/streaming/asof_join.slt
```diff
@@ -152,6 +152,34 @@ select * from mv1;
...
+create materialized view mv_cpu as SELECT t1.v1 t1_v1, t1.v2 t1_v2, t1.v3 t1_v3, t2.v1 t2_v1, t2.v2 t2_v2, t2.v3 t2_v3 FROM t1 ASOF LEFT JOIN t2 ON t1.v1 = t2.v1 and t1.v2 > t2.v2;
+
+query III
+select * from mv_cpu;
+----
+2 3 4 NULL NULL NULL
+
+statement ok
+drop materialized view mv_cpu;
+
```
stdrc: Shall we just run all existing join tests with the new mode? Like in https://github.com/risingwavelabs/risingwave/blob/5417d18ee47628df4c668e23966604c7fb9c5a86/e2e_test/streaming/over_window/main.slt.

lmatz: good idea

kwannoel:
Any benchmark?

lmatz:
https://github.com/risingwavelabs/risingwave/issues/20017

On slack

https://risingwave-labs.slack.com/archives/C034TRPKN1F/p1735893228781809?thread_ts=1733464753.437989&cid=C034TRPKN1F

> still using the name of the “decode” function but actually it is just copying the OwnedRow once
“decode” CPU percentage: 50% -> 14%

> MV creation: 3min -> 2min

> dev=> CREATE MATERIALIZED VIEW IF NOT EXISTS mv AS
    SELECT a.v1 as av1, b.v1 as bv1, a.v2 as av2, b.v2 as bv2, a.v3 as av3, b.v3 as bv3, a.v4 as av4, b.v4 as bv4, a.v5 as bv5
    FROM a, b
    WHERE a.v1 = b.v1 AND b.v2 = ANY(a.v2) AND a.v3 < b.v3 AND b.v3 < a.v4 AND a.v5 < b.v4;
CREATE_MATERIALIZED_VIEW
Time: 117322.207 ms (01:57.322)

ci/workflows/pull-request.yml
```diff
@@ -198,7 +198,7 @@ steps:
           run: source-test-env
           upload-container-logs: always
       - ./ci/plugins/upload-failure-logs
-    timeout_in_minutes: 20
+    timeout_in_minutes: 25
```
kwannoel, line 201: Can you trigger main-cron to run also for this workflow. The timeout for main-cron may need to be adjusted as well.

e2e_test/streaming/join/main.slt
```diff
@@ -0,0 +1,16 @@
+statement ok
+set streaming_join_encoding_optimized_for to cpu;
```
kwannoel:
```suggestion
set streaming_join_encoding to cpu_optimized;
```

Nit: Seems more readable.

src/stream/src/from_proto/hash_join.rs
```diff
@@ -197,43 +199,57 @@ impl<S: StateStore> HashKeyDispatcher for HashJoinExecutorDispatcherArgs<S> {
...
+            (Inner),
+            (LeftOuter),
+            (RightOuter),
+            (FullOuter),
+            (LeftSemi),
+            (LeftAnti),
+            (RightSemi),
+            (RightAnti)
```
kwannoel: Nit: Why do we need parenthesis here. Seems like this can be removed from the macro instead of `(Inner)`, can just be `Inner`.

kwannoel (approved):
Thanks, this looks good. @yuhao-su could you also leave a review?

kwannoel:
Could you also write in the release-note, when users may expect to use this, and what's the caveats.

PR #22419, opened 2025-06-30
refactor(meta): wrap fragment type mask with struct

wenym1:

Previously when declaring the `FragmentTypeFlag`, we need to carefully set the raw int value of the enum variant to be a well-formed power of 2 value. In this PR, we will change to use a macro to generate the enum, so that we only need to declare a list of flags, and then the value can be correctly generated.

Besides, previously we operate with `fragment_type_mask` as raw int value, where we have to do raw bit operation on int value, which is unreadable, and can easily make mistake. In this PR, we will wrap the raw int value with a struct `FragmentTypeMask`, and provide method like `add`, `contains` to encapsulate the logic of `|=` and `(mask | flag) != 0`.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/proto/stream_plan.proto
+++ b/proto/stream_plan.proto
@@ -1050,26 +1050,6 @@ message StreamActor {
   plan_common.ExprContext expr_context = 10;
 }
 
-// Indicates whether the fragment contains some special kind of nodes.
-enum FragmentTypeFlag {
-  FRAGMENT_TYPE_FLAG_FRAGMENT_UNSPECIFIED = 0;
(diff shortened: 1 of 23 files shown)


wenym1:
* **#22419** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22419?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22419?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #22421, opened 2025-06-30
refactor(frontend): static typed plan phase

wenym1:

Previously in frontend optimizer, we have `PlanPhase` enum to determine the current phase of the plan. We have assertion on the phase to ensure that plan of invalid phase is not passed by mistake.

In this PR, we will change from dynamic enum type to static generic type, so that we can detect the incorrect usage at compile time.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/optimizer/mod.rs
+++ b/src/frontend/src/optimizer/mod.rs
@@ -46,6 +46,7 @@ use std::assert_matches::assert_matches;
 use std::collections::{BTreeMap, HashMap};
 use std::marker::PhantomData;
 
+use educe::Educe;
 use fixedbitset::FixedBitSet;
 use itertools::Itertools as _;
 pub use logical_optimization::*;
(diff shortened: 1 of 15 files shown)


wenym1:
* **#22421** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22421?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22421?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/frontend/src/optimizer/mod.rs
```diff
@@ -97,11 +98,11 @@ use crate::utils::{ColIndexMappingRewriteExt, WithOptionsSecResolved};
 /// column is required in optimization, but the final generated plan will remove the unnecessary
 /// column in the result.
 #[derive(Debug, Clone)]
-pub struct PlanRoot {
+pub struct PlanRoot<P> {
     // The current plan node.
-    plan: PlanRef,
+    pub plan: PlanRef,
```
BugenZhao, line 105: Actually we also have some type-safer versions of `PlanRef` like `StreamPlanRef` or `BatchPlanRef`. Wondering if we can use them to achieve better safety further.

src/frontend/src/optimizer/mod.rs
```diff
@@ -97,11 +98,11 @@ use crate::utils::{ColIndexMappingRewriteExt, WithOptionsSecResolved};
 /// column is required in optimization, but the final generated plan will remove the unnecessary
 /// column in the result.
 #[derive(Debug, Clone)]
-pub struct PlanRoot {
+pub struct PlanRoot<P> {
```
BugenZhao: I guess you can use `educe` to derive implementation without bounds on type parameters.

chenzl25 (approved):
LGTM! Master of macro!

src/frontend/src/optimizer/mod.rs
wenym1: I am having a try, but it's quite difficult. At the moment my branch already has 120 changed files and 900+ LOC.

PR #22423, opened 2025-06-30
feat(storage): bump OpenDAL to v0.53

wcy-fdu:

As title. 

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/src/source/filesystem/opendal_source/opendal_enumerator.rs
+++ b/src/connector/src/source/filesystem/opendal_source/opendal_enumerator.rs
@@ -19,7 +19,7 @@ use async_trait::async_trait;
 use chrono::{DateTime, Utc};
 use futures::stream::{self, BoxStream};
 use futures::{StreamExt, TryStreamExt};
-use opendal::{Metakey, Operator};
+use opendal::Operator;
 use risingwave_common::types::Timestamptz;
 
 use super::OpendalSource;
@@ -74,12 +74,7 @@ impl<Src: OpendalSource> OpendalEnumerator<Src> {
             true => prefix,
             false => "/",
         };
-        let object_lister = self
-            .op
-            .lister_with(list_prefix)
-            .recursive(true)
-            .metakey(Metakey::ContentLength | Metakey::LastModified)
-            .await?;
+        let object_lister = self.op.lister_with(list_prefix).recursive(true).await?;
         let stream = stream::unfold(object_lister, |mut object_lister| async move {
             match object_lister.next().await {
                 Some(Ok(object)) => {
--- a/src/object_store/src/object/opendal_engine/opendal_object_store.rs
+++ b/src/object_store/src/object/opendal_engine/opendal_object_store.rs
@@ -22,7 +22,7 @@ use futures::{StreamExt, stream};
 use opendal::layers::{RetryLayer, TimeoutLayer};
 use opendal::raw::BoxedStaticFuture;
 use opendal::services::Memory;
-use opendal::{Execute, Executor, Metakey, Operator, Writer};
+use opendal::{Execute, Executor, Operator, Writer};
 use risingwave_common::config::ObjectStoreConfig;
 use risingwave_common::range::RangeBoundsExt;
 use thiserror_ext::AsReport;
@@ -227,7 +227,7 @@ impl ObjectStore for OpendalObjectStore {
     /// Deletes the objects with the given paths permanently from the storage. If an object
     /// specified in the request is not found, it will be considered as successfully deleted.
     async fn delete_objects(&self, paths: &[String]) -> ObjectResult<()> {
-        self.op.remove(paths.to_vec()).await?;
+        self.op.delete_iter(paths.to_vec()).await?;
         Ok(())
     }
 
@@ -237,11 +237,7 @@ impl ObjectStore for OpendalObjectStore {
         start_after: Option<String>,
         limit: Option<usize>,
     ) -> ObjectResult<ObjectMetadataIter> {
-        let mut object_lister = self
-            .op
-            .lister_with(prefix)
-            .recursive(true)
-            .metakey(Metakey::ContentLength);
+        let mut object_lister = self.op.lister_with(prefix).recursive(true);
         if let Some(start_after) = start_after {
             object_lister = object_lister.start_after(&start_after);
         }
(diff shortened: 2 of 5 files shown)


wcy-fdu:
longevity test passed on [s3](https://buildkite.com/risingwave-test/longevity-test/builds/2593) and [gcs](https://github.com/risingwavelabs/risingwave/issues/22489)

wcy-fdu:
[Perf test result](http://metabase.risingwave-cloud.xyz/dashboard/1405-nexmark-8c32g-1cn-rw-source-throughput?start_date=2025-07-01)

wcy-fdu:
Reran several queries that had previously experienced performance regression and found that there should not be much difference in performance. refer to the newest [perf test result](http://metabase.risingwave-cloud.xyz/dashboard/1405-nexmark-8c32g-1cn-rw-source-throughput?start_date=2025-07-30).

wcy-fdu:
CI failed due to some behavior change of lister, more detail refer to https://github.com/apache/opendal/discussions/6481

e2e_test/source_inline/fs/posix_fs.slt
```diff
@@ -62,7 +62,7 @@ diamonds_mv	FS_FETCH	{MVIEW,FS_FETCH}	0
 diamonds_mv	SOURCE	{SOURCE}	0
 
 
-sleep 3s
```
Li0k: Why need to increase the sleep interval?

src/connector/src/source/filesystem/opendal_source/azblob_source.rs
```diff
@@ -54,6 +54,11 @@ impl<Src: OpendalSource> OpendalEnumerator<Src> {
             .layer(RetryLayer::default())
             .finish();
 
+        // Check if we need to call stat() to get complete metadata
```
Li0k: nits:  The metadata that **need_stat_metadata** requires can be obtained from op, so we don’t need to pass it in explicitly, it can be constructed on the fly via op. Of course, that’s up to you.

zwang28: Do you mean metadata_capability? +1.

src/object_store/src/object/opendal_engine/opendal_object_store.rs
```diff
@@ -237,35 +245,44 @@ impl ObjectStore for OpendalObjectStore {
...
+            let op = op.clone();
+
+            async move {
+                match object_lister.next().await {
+                    Some(Ok(object)) => {
+                        let key = object.path().to_owned();
+                        let mut meta = object.metadata().clone();
+
+                        // If we need to call stat() to get complete metadata
+                        if need_stat_metadata {
```
zwang28: I think we should check metadata_capability for required metadata, and only call explictly stat when needed. e.g. for S3 object.metadata() may already contains all required metadata, so no need for stat.

All need_stat_metadata can be removed.

src/connector/src/source/filesystem/opendal_source/opendal_enumerator.rs
```diff
@@ -85,36 +87,44 @@ impl<Src: OpendalSource> OpendalEnumerator<Src> {
...
+            let op = op.clone();
+
+            async move {
+                match object_lister.next().await {
+                    Some(Ok(object)) => {
+                        let name = object.path().to_owned();
+                        let mut meta = object.metadata().clone();
+
+                        // If we need to call stat() to get complete metadata
+                        if need_stat_metadata {
```
zwang28: ditto

src/object_store/src/object/opendal_engine/opendal_object_store.rs
zwang28: https://docs.rs/opendal/0.53.3/opendal/docs/rfcs/rfc_5314_remove_metakey/index.html#guide-level-explanation

wcy-fdu: Actually I did't find metadata_capability() in `Operator`, I guess the doc is wrong. refer to [operator.rs](https://github.com/apache/opendal/blob/v0.53.3/core/src/types/operator/operator.rs)

wcy-fdu: You can only see metadata_capability() in the whole repo once, in a rfc design, not in codebase. https://github.com/search?q=repo%3Aapache%2Fopendal%20metadata_capability&type=code

src/connector/src/source/filesystem/opendal_source/azblob_source.rs
wcy-fdu: However, when taking op.info().full_capability() from the op in the list, the action will be done per list. The current implementation is to do it once when the op is built.

zwang28: > we don’t need to pass it in explicitly, it can be constructed on the fly via op

+1.

> when taking op.info().full_capability() from the op in the list, the action will be done per list

I think this op is cheap, i.e. reading some in-mem value. Can you confirm?

src/connector/src/source/filesystem/opendal_source/opendal_enumerator.rs
```diff
@@ -85,36 +85,47 @@ impl<Src: OpendalSource> OpendalEnumerator<Src> {
...
+        let op = self.op.clone();
+        let full_capability = op.info().full_capability();
+        let stream = stream::unfold(object_lister, move |mut object_lister| {
+            let op = op.clone();
+
+            async move {
+                match object_lister.next().await {
+                    Some(Ok(object)) => {
+                        let name = object.path().to_owned();
+                        let mut meta = object.metadata().clone();
```
zwang28: the metadata().clone() can be avoided.
e.g.
let (last_modified, content_length) = if !full_capability.list_has_content_length
                            || !full_capability.list_has_last_modified {
...
} else {
...
}

src/object_store/src/object/opendal_engine/opendal_object_store.rs
```diff
@@ -237,35 +237,47 @@ impl ObjectStore for OpendalObjectStore {
...
+        let op = self.op.clone();
+        let full_capability = op.info().full_capability();
+        let stream = stream::unfold(object_lister, move |mut object_lister| {
+            let op = op.clone();
+
+            async move {
+                match object_lister.next().await {
+                    Some(Ok(object)) => {
+                        let key = object.path().to_owned();
+                        let mut meta = object.metadata().clone();
```
zwang28: ditto

zwang28 (changes requested):

hzxa21 (approved):
Cargo.lock LGTM

Issue #21433, opened 2025-04-16
Support viewing background sink progress


kwannoel:
To mitigate this, we can support `show jobs <rel_name>` which directly queries the internal backfill tables of some relation created with background_ddl.

However, it will not show up in `show jobs`, as we only filter for created jobs. Eventually we still need to support it.
So we need to refactor the path of `create sink` to follow `create materialized view`. Such that `create sink` will only be marked as created once backfill is complete.

For sink into table, we have to pay extra attention to it.

Issue #16612, opened 2024-05-07
Tracking: Stabilize `BACKGROUND_DDL`


kwannoel:
Currently background ddl has been a beta feature for a long time. It's time to make its implementation more polished.

- [ ] https://github.com/risingwavelabs/risingwave/issues/15603
- [ ] Refactor sink: Instead of marking it as "finished", it should also follow the same background creation path as MV.

kwannoel:
Not urgent, wait for meta sql backend stabilized.

kwannoel:
Can be worked on now. Since meta sql backend has been stabilized.

Issue #12771, opened 2023-10-11
discussion: allow stream query on creating mv? 


st1page:
offline discussed with @wyhyhyhyh
Currently, the creating MV is invisible and the user can not do batch or streaming query on it.
```SQL
dev=> SET BACKGROUND_DDL=true;
SET_VARIABLE
dev=> CREATE TABLE t (v1 int);
CREATE_TABLE
dev=> INSERT INTO t select * from generate_series(1, 300000);
INSERT 0 300000
dev=> FLUSH;
FLUSH
dev=> CREATE MATERIALIZED VIEW m1 as SELECT * FROM t;
CREATE_MATERIALIZED_VIEW
dev=> CREATE MATERIALIZED VIEW m2 as SELECT * FROM m1;
ERROR:  QueryError: Catalog error: table or source not found: m1
```
Complex data processing pipelines usually are layered which include many materialized views depending on each other. While creating a stream query on the existing MV, RW will backfill all the historical data in the upstream MV and union it with the coming changes. 
Considering there are mvA, a mvB on mvA, and a mvC on mvB(S->MvA->MvB->MvC). Under the current design, the user must create the materialized views one by one. He should create the MvA and wait for all historical data to be backfilled in MvB, and only after that, he can create the mvB. But if the pipeline can be 

1. allow users to create multiple materialized views using syntax like "transactional DDL"
2. allow users to create stream query on the creating mv, so that the downstream mv will consume the historical data faster(because there are less data than the base table) and catch up with the upstream.

fuyufjh:
How about letting `create table as <query>` always return immediately and backfill in the background?

st1page:
Also for user experience, I think we should allow stream query on creating mv, or he must wait for the mv backfill all the historical data.

chenzl25:
> allow users to create stream query on the creating mv, so that the downstream mv will consume the historical data faster(because there are less data than the base table) and catch up with the upstream.

Personally I prefer the second one, because it is more practical and the concept of transactional ddl is too big? BTW, recoverable backfill is also needed in this case. We can return the ddl immediately to make creating mv visible to streaming query and batch query. But for batch query consistency, we need to block it until all its upstream backfilling mvs are finished.

kwannoel:
> > allow users to create stream query on the creating mv, so that the downstream mv will consume the historical data faster(because there are less data than the base table) and catch up with the upstream.
> 
[...]

PR #22435, opened 2025-06-30
feat: support viewing background sink progress and some refactoring

yezizp2012:

Part of #16612 ,  #21433 .

Still not enabled for sink into table. In local testing, I identified the need for further refactoring to stabilize and enhance the recoverability of the background sinking into the table with progress tracking. I'll create a separate PR for this issue, keeping this one straightforward.

## Checklist

## Documentation

<details>

</details>


e2e_test/background_ddl/sink.slt
```diff
@@ -0,0 +1,66 @@
...
+onlyif can-use-recover
+statement count 1
+show jobs;
+
+statement ok
+alter sink sink1 set SINK_RATE_LIMIT to 10000;
+
+sleep 5s
+
+statement count 0
```
kwannoel:
```suggestion
statement count 0 retry 3 backoff 2s
```

e2e_test/background_ddl/sink.slt
```diff
@@ -0,0 +1,66 @@
...
+sleep 5s
+
+onlyif can-use-recover
+statement count 1
+show jobs;
+
+statement ok
+alter sink sink1 set SINK_RATE_LIMIT to 10000;
+
+sleep 5s
```
kwannoel:
```suggestion
```

src/frontend/src/catalog/root_catalog.rs
```diff
@@ -1011,8 +1030,12 @@ impl Catalog {
...
-        } else if schema.get_sink_by_name(relation_name).is_some() {
-            Err(CatalogError::duplicated("sink", relation_name.to_owned()))
+        } else if let Some(sink) = schema.get_any_sink_by_name(relation_name) {
+            Err(CatalogError::Duplicated(
+                "table",
+                relation_name.to_owned(),
+                !sink.is_created(),
+            ))
         } else if schema.get_view_by_name(relation_name).is_some() {
             Err(CatalogError::duplicated("view", relation_name.to_owned()))
```
kwannoel, line 1040: Maybe we need to apply this to MVs later.. In a separate PR.

e2e_test/.coverageignore
```diff
@@ -15,3 +15,6 @@ e2e_test/source_legacy/pulsar/**/*.slt*
 
 # Manually executed in `e2e_test/subscription/main.py`
 e2e_test/subscription/**/*.slt*
+
+# Manually executed in `run-e2e-tests.sh`
+e2e_test/background_ddl/**/*.slt*
```
kwannoel: I think it's not executed at all

kwannoel: Could you add a line to execute it inside `e2e-test-serial.sh`?

You can ctrl-F `background_ddl`. There's a line like:
```
risedev slt -p 4566 -d dev './e2e_test/background_ddl/basic.slt' --junit "batch-ddl-${profile}"
```

kwannoel (commented):
Thanks for the PR! Will take a look again after the tests run.

src/frontend/src/catalog/schema_catalog.rs
```diff
@@ -776,8 +782,22 @@ impl SchemaCatalog {
...
+        &self,
+        sink_name: &str,
+        bind_creating: bool,
+    ) -> Option<&Arc<SinkCatalog>> {
+        self.sink_by_name
+            .get(sink_name)
+            .filter(|s| bind_creating || s.is_created())
+    }
+
+    pub fn get_any_sink_by_name(&self, sink_name: &str) -> Option<&Arc<SinkCatalog>> {
```
wenym1, line 795: The method name `get_any_sink` indicates that it includes both creating and created sinks, but it actually returns only created sinks. Is this expected?

src/frontend/src/handler/drop_sink.rs
```diff
@@ -62,55 +64,69 @@ pub async fn handle_drop_sink(
...
-
-        let columns_without_rw_timestamp = table_catalog.columns_without_rw_timestamp();
-        for sink in fetch_incoming_sinks(&session, &incoming_sink_ids)? {
-            hijack_merger_for_target_table(
-                &mut graph,
-                &columns_without_rw_timestamp,
-                &sink,
-                Some(&sink.unique_identity()),
-            )?;
+    match sink.stream_job_status {
```
wenym1, line 67: Can we possibly drop a creating sink in frontend? Do we have similar logic in dropping mv?

src/meta/src/controller/streaming_job.rs
```diff
@@ -487,7 +503,7 @@ impl CatalogController {
         txn.commit().await?;
 
         if !objects.is_empty() {
-            assert!(is_materialized_view);
+            assert!(need_notify);
```
wenym1, line 501: Curious what was the original purpose to notify frontend specially for mv?

e2e_test/.coverageignore
yezizp2012: Ah, it appears I searched in an older branch somehow and found it in a pre-existing file `run-e2e-tests.sh`. I'll add it.

src/frontend/src/catalog/root_catalog.rs
yezizp2012: It already exists in L1011.

src/frontend/src/handler/drop_sink.rs
yezizp2012: Yes, we can. We also have such similar logic in dropping mv.

src/meta/src/controller/streaming_job.rs
yezizp2012: This is to make the mv in the creating state searchable in the FE. This's the related issue: #12771 .

kwannoel (approved):
Thanks!

src/meta/src/controller/streaming_job.rs
kwannoel: > Curious what was the original purpose to notify frontend specially for mv?

I supported it, so the user can just run `drop materialized view <mv_name>` to cancel their background / foreground streaming job, rather than having to go through the whole `show jobs`, `cancel jobs` etc...

PR #22447, opened 2025-07-01
refactor(cdc): ban create table with connector info for cdc[low priority]

wcy-fdu:

As title, `create table with` should be deprecated, the recommend syntax is `create table from source table '<database_name>.<table_name>';`

This pr ban the deprecated syntax in frontend, **only for cdc**.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/handler/create_table.rs
+++ b/src/frontend/src/handler/create_table.rs
@@ -1349,6 +1349,29 @@ pub async fn handle_create_table(
 ) -> Result<RwPgResponse> {
     let session = handler_args.session.clone();
 
+    // CDC connector ban check
+    if let Some(connector) = handler_args.with_options.get_connector() {
+        use risingwave_connector::source::cdc::{
+            MONGODB_CDC_CONNECTOR, MYSQL_CDC_CONNECTOR, POSTGRES_CDC_CONNECTOR,
+            SQL_SERVER_CDC_CONNECTOR,
+        };
+        if [
+            MYSQL_CDC_CONNECTOR,
+            POSTGRES_CDC_CONNECTOR,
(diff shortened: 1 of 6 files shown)


src/frontend/src/handler/create_table.rs
```diff
@@ -1349,6 +1349,29 @@ pub async fn handle_create_table(
 ) -> Result<RwPgResponse> {
     let session = handler_args.session.clone();
 
+    // CDC connector ban check
```
hzxa21, line 1352: I suggest we put the checking logic in [check_create_table_with_source](https://github.com/risingwavelabs/risingwave/blob/b6524f9d6ad9156d54afb47194bfd7d59353e153/src/frontend/src/handler/create_table.rs#L1026) so that `EXPLAIN` command can also work.

src/frontend/src/handler/create_table.rs
```diff
@@ -1349,6 +1349,29 @@ pub async fn handle_create_table(
...
+    if let Some(connector) = handler_args.with_options.get_connector() {
```
hzxa21, line 1353: Also, how about implementing a function in the `WithPropertiesExt` to check for the non shared source cdc connector usage? https://github.com/risingwavelabs/risingwave/blob/b6524f9d6ad9156d54afb47194bfd7d59353e153/src/connector/src/with_options.rs#L152

hzxa21: Looks like `is_shareable_cdc_connector` is what you are looking for. You don't need a new method.

wcy-fdu:
low priority

wcy-fdu:
Reopen when needed.

Issue #22425, opened 2025-06-30
Setting Sink rate limit to 0 will cause a freeze.


xxhZs:
### Describe the bug

```
CREATE TABLE t1(v1 int, v2 int);
CREATE SINK s1 AS select * from t1 WITH (
    connector = 'blackhole'
);
ALTER SINK s1 SET SINK_RATE_LIMIT = 0;
 ALTER SINK s1 SET SINK_RATE_LIMIT = default;
```

https://github.com/risingwavelabs/risingwave/pull/21970

### Error message/log

```text

```

PR #22458, opened 2025-07-01
fix(sink): fix mem log store was blocked with rate limit = 0

xxhZs:

https://github.com/risingwavelabs/risingwave/issues/22425

Before, we could also enable the rate limit for mem log store. Therefore, when it was 0, the reader would not consume the barrier, causing the writer to not pull the upstream barrier, thus leading to a deadlock.
In this PR, we can only set the limit when it comes to KV log store .

## Checklist

## Documentation

<details>

</details>


src/frontend/src/handler/alter_streaming_rate_limit.rs
```diff
@@ -97,14 +101,46 @@ pub async fn handle_alter_streaming_rate_limit(
...
+            };
+            let selection = Some(Expr::BinaryOp {
+                left: Box::new(Expr::Identifier(SINK_ID.into())),
+                op: BinaryOperator::Eq,
+                right: Box::new(Expr::Value(Value::Number(sink_id.to_string()))),
+            });
+            let query = gen_query_from_table_name_with_selection(
+                ObjectName::from(vec![
+                    Ident::from("rw_catalog"),
+                    Ident::from("rw_sink_decouple"),
```
hzxa21: Checking the system table here seems a bit strange. I think it is better to do the check in meta [update_sink_rate_limit_by_job_id](https://github.com/risingwavelabs/risingwave/blob/5a972a6704769bc819744b0a91c92613dba9ef5a/src/meta/src/controller/streaming_job.rs#L1666). You can check the `SinkLogStoreType` in the sink plan node there.

tabVersion:
A little strange to me.

It is true that rate limiter affects consuming barrier. But solving it by limiting alter sink rate_limit command can only affect the decoupled sink, where barrier is already collected, is really confusing.
It can affect some normal usage, eg. throttle the sink to prevent overload the downstream. It makes sense for both decouple sink and un-decouple sink.

I think a better solution is prevent the rate limit set to 0, it should at least consume 1 msg/s to keep the graph moving.

xxhZs:
> A little strange to me.
> 
> It is true that rate limiter affects consuming barrier. But solving it by limiting alter sink rate_limit command can only affect the decoupled sink, where barrier is already collected, is really confusing. It can affect some normal usage, eg. throttle the sink to prevent overload the downstream. It makes sense for both decouple sink and un-decouple sink.
> 
> I think a better solution is prevent the rate limit set to 0, it should at least consume 1 msg/s to keep the graph moving.

We only imposed a sink limit on memory because our sink limit relies on the KV log store for storage, which creates a discrepancy between the sink rate and stream rate. When using memory, the sink limit and source limit are effectively the same.

tabVersion:
@xxhZs any update on this?

hzxa21:
@tabVersion some backgrounds I would like to add:
sink rate limit is originally introduced under the assumption that the sink is decoupled (this is the default behavior) because if the sink is coupled, rate limiting the sink will affect the checkpoint of the overall streaming graph, even for the jobs that have nothing to do with the sink, which is more destructive than the sink rate limit itself. In other words, I think sink throttling is only valid with strong isolation with sink decoupling. AFAIK, we don't have a use case that requires sink rate limit for non decoupled sink. We can reconsider that if we do such cases.

tabVersion (approved):
the approach lgtm

hzxa21:
```
<html>
<body>
<!--StartFragment-->
e2e_test/background_ddl/sink.slt                             .. [FAILED] after 5349 ms
--
  |  
  | failed to run `e2e_test/background_ddl/sink.slt`
  |  
  | Caused by:
  | statement failed: db error: ERROR: Failed to run the query
  |  
  | Caused by:
  | Invalid Parameter Value: Invalid parameter: sink rate limit is only supported for kv log store, please SET sink_decouple = TRUE before CREATE SINK
  |  
  | [SQL] alter sink sink1 set SINK_RATE_LIMIT to 10000;
  | at e2e_test/background_ddl/sink.slt:35

<!--EndFragment-->
</body>
</html>
```

PR #22432, opened 2025-06-30
feat(frontend): support create vector index on frontend

wenym1:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## What's changed and what's your intention?

As titled. Example:
```
create table items (id int primary key, name string, embedding vector(3)) append only;
create index i on items using flat (embedding) include (name) with (distance_type = 'inner_product');
```

## Checklist

- [ ] I have written necessary rustdoc comments.
- [ ]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  I have checked the [Release Timeline](https://github.com/risingwavelabs/rw-commits-history/blob/main/release_timeline.md) and [Currently Supported Versions](https://docs.risingwave.com/changelog/release-support-policy#support-end-dates-for-recent-releases) to determine which release branches I need to cherry-pick this PR into. 

## Documentation

- [ ]  My PR needs documentation updates. 

<details>
<summary><b>Release note</b></summary>

</details>

PR #22348, opened 2025-06-24
feat(batch): support vector index nearest batch executor

wenym1:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## What's changed and what's your intention?

As titled. The vector index nearest batch executor will take an input, and use the embedding columns in the input, to query the topn nearest rows in vector index, and include the results in an extra columns, which is a list of rows of struct type.

In future PR, for the following query
```
SELECT * FROM items ORDER BY embedding <-> '[3,1,2]' LIMIT 5;
 id | embedding
----+-----------
  1 | [1,2,3]
  2 | [4,5,6]
```

the plan will be like
```
BatchValues(vector [3,1,2]) => 
BatchVectorIndexNearest => 
	query_vector | vector_info(List<struct<output>>)
	--------------------------
	   [3,1,2]   | array(row(1, [1,2,3]), row(2, [4,5,6]))
BatchProjectSet(unnest(vector_info)) =>
	vector_info(struct<output>)
	--------------------------
	row(1, [1,2,3])
	row(2, [4,5,6])
BatchProject => output
	 id | embedding
	----+-----------
	  1 | [1,2,3]
	  2 | [4,5,6]
```

## Checklist

- [ ] I have written necessary rustdoc comments.
- [ ]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  I have checked the [Release Timeline](https://github.com/risingwavelabs/rw-commits-history/blob/main/release_timeline.md) and [Currently Supported Versions](https://docs.risingwave.com/changelog/release-support-policy#support-end-dates-for-recent-releases) to determine which release branches I need to cherry-pick this PR into. 

## Documentation

- [ ]  My PR needs documentation updates. 

<details>
<summary><b>Release note</b></summary>

</details>

PR #22241, opened 2025-06-16
feat(streaming): implement vector index write executor

wenym1:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## Checklist

- [ ] I have written necessary rustdoc comments.
- [ ]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  I have checked the [Release Timeline](https://github.com/risingwavelabs/rw-commits-history/blob/main/release_timeline.md) and [Currently Supported Versions](https://docs.risingwave.com/changelog/release-support-policy#support-end-dates-for-recent-releases) to determine which release branches I need to cherry-pick this PR into. 

## Documentation

- [ ]  My PR needs documentation updates. 

<details>
<summary><b>Release note</b></summary>

</details>

PR #22467, opened 2025-07-01
feat(frontend): optimizer rule to convert vector top n to vector index read

wenym1:

Introduce a rule to convert the following TopN pattern to `LogicalVectorSearch`
```text
     LogicalTopN { order: [$expr1 ASC], limit: TOP_N, offset: 0 }
     └─LogicalProject { exprs: [VectorDistanceFunc(vector_expr1, vector_expr2) as $expr1, other_exprs...] }
```
to
```text
    LogicalProject { exprs: [other_exprs... + distance_column] }
      └─LogicalVectorSearch { distance_type: `PbDistanceType`, top_n: TOP_N, left: vector_expr1, right: vector_expr2, output_columns: [...] }
```

## Checklist

## Documentation

<details>

</details>


wenym1:
* **#22808** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22808?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#22773** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22773?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#22545** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22545?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#22467** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22467?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22467?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#22348** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22348?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#22432** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22432?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#22241** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22241?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/frontend/src/optimizer/plan_node/logical_vector_search.rs
```diff
@@ -0,0 +1,338 @@
...
+    core: VectorSearchCore,
+}
+
+impl LogicalVectorSearch {
+    pub(crate) fn new(
+        top_n: u64,
+        distance_type: PbDistanceType,
+        left: ExprImpl,
+        right: ExprImpl,
+        output_input_idx: Vec<usize>,
```
chenzl25: Why do we need this field? It seems we can use a project operator outside instead.

wenym1: Indeed. The logic was previously for optionally excluding the vector and distance column in the output, but now this logic has been removed, and we always yield the distance column and exclude the vector column.

I will remove this field and its usage.

wenym1: Oh wait, this is not true. 

The input of `LogicalVectorSearch` will be the `LogicalScan`. We need to store which columns of `LogicalScan` should be included in the output. The `output_col_idx` in `TableScan` is not sufficient. When viable vector index does not exists, `LogicalVectorSearch` will be converted back into a TopN, and the TopN takes the vector column as the input to calculate the distance, which means the `output_col_idx` in `TableScan` should always include the vector columns. However, when viable vector index exists, we don't have to include the vector column, unless the query output requires it. Therefore, we store an extra `output_col_idx` here to indicate whether the vector column is in the input. When the query output requires the vector column, we will treat the vector index as non-covering index, and should later unnecessarily look up the base table for the vector columns.

chenzl25: I think the schema of LogicalVectorSearch could be simplified as input_schema + the distance. If the query output don't need this column, it can prune the column with a project operator. We can always keep the vector column as input and I think it should be since the `left` and `right` would refer to it.

src/frontend/src/optimizer/rule/top_n_to_vector_search_rule.rs
```diff
@@ -0,0 +1,126 @@
...
+
+pub struct TopNToVectorSearchRule;
+
+impl TopNToVectorSearchRule {
+    pub fn create() -> BoxedRule<Logical> {
+        Box::new(TopNToVectorSearchRule)
+    }
+}
+
+impl Rule<Logical> for TopNToVectorSearchRule {
```
chenzl25: I think you can draw a graph to represent how the plan is transformed before and after this rule. AI is good at this stuff.

src/frontend/src/optimizer/plan_node/logical_vector_search.rs
```diff
@@ -0,0 +1,338 @@
...
+use crate::optimizer::property::{FunctionalDependencySet, Order};
+use crate::utils::{ColIndexMappingRewriteExt, Condition};
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+struct VectorSearchCore {
+    top_n: u64,
+    distance_type: PbDistanceType,
+    left: ExprImpl,
+    right: ExprImpl,
+    output_input_idx: Vec<usize>,
```
chenzl25: It seems `output_indices` is a better name.

src/frontend/src/optimizer/plan_node/logical_vector_search.rs
```diff
@@ -0,0 +1,338 @@
...
+    }
+}
+
+impl PredicatePushdown for LogicalVectorSearch {
+    fn predicate_pushdown(
+        &self,
+        predicate: Condition,
+        ctx: &mut PredicatePushdownContext,
+    ) -> PlanRef {
+        gen_filter_and_pushdown(self, predicate, Condition::true_cond(), ctx)
```
chenzl25, line 260: ~~Considering we have an `output_indices`, we need to apply the o2i_mapping to the predicate before pushing the predicate to the input.~~

chenzl25 (approved):
Rest LGTM

PR #22487, opened 2025-07-02
feat: decouple streaming now timestamp and epoch

hzxa21:

## TL;DR
add a `stream_now_progress_ratio` to decouple now timestamp from the epoch timestamp to avoid introducing excessive changes within one epoch from dynamic filter
``` rust
// stream_now_progress_ratio must be set > 1.0, otherwise it will be ignore
let adjusted_timestamp = prev_now_timestamp + barrier_interval_ms * stream_now_progress_ratio;
let now_timestamp = min(adjusted_timestamp, epoch_timestamp);
```

## Summary
This pull request introduces enhancements to the `NowExecutor` by adding support for a configurable `progress_ratio` to adjust timestamp progression dynamically. It also includes changes to the configuration system to support this new feature. Below are the key changes grouped by theme:

### Configuration Updates
* Added a new `now_progress_ratio` field to `StreamingDeveloperConfig` to allow optional configuration of the progress ratio for the `NowExecutor`. A default function `streaming_now_progress_ratio` returning `None` was also added. (`src/common/src/config.rs`, [[1]](diffhunk://#diff-c81fa6ff8ebe6cc6822331e0f5ae282664862104a8b9189edce5e9336685690fR1300-R1302) [[2]](diffhunk://#diff-c81fa6ff8ebe6cc6822331e0f5ae282664862104a8b9189edce5e9336685690fR2417-R2420)

### `NowExecutor` Enhancements
* Introduced `progress_ratio` and `barrier_interval_ms` as new fields in the `NowExecutor` to enable dynamic timestamp adjustment during execution. These fields are initialized in the constructor and used in the `execute` method to calculate adjusted timestamps. (`src/stream/src/executor/now.rs`, [[1]](diffhunk://#diff-02242739957400ef6800c5c4fcb2889b19700416cdb8035cefc29c4815896069R41-R44) [[2]](diffhunk://#diff-02242739957400ef6800c5c4fcb2889b19700416cdb8035cefc29c4815896069R73-R83) [[3]](diffhunk://#diff-02242739957400ef6800c5c4fcb2889b19700416cdb8035cefc29c4815896069R95-R96) [[4]](diffhunk://#diff-02242739957400ef6800c5c4fcb2889b19700416cdb8035cefc29c4815896069L155-R195)
* Modified the timestamp progression logic in the `execute` method to account for the `progress_ratio`. If the ratio is greater than `1.0`, the timestamp is adjusted to avoid large gaps, ensuring smoother downstream operations. (`src/stream/src/executor/now.rs`, [src/stream/src/executor/now.rsL155-R195](diffhunk://#diff-02242739957400ef6800c5c4fcb2889b19700416cdb8035cefc29c4815896069L155-R195))

Verify it works via the following config locally:
```
[streaming.developer]
stream_now_progress_ratio = 2.0

```

TODO:
- UT

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/executor/now.rs
+++ b/src/stream/src/executor/now.rs
@@ -177,9 +177,10 @@ impl<S: StateStore> NowExecutor<S> {
                         < new_timestamp.timestamp_millis()
                     {
                         info!(
-                            "adjusting progress timestamp from {} to {}. barrier_interval_ms: {}, progress_ratio: {}",
+                            "adjusted next now timestamp from {} to {}. curr_epoch: {}, barrier_interval_ms: {}, progress_ratio: {}",
                             new_timestamp.timestamp_millis(),
                             progress_timestamp,
+                            barrier.get_curr_epoch(),
                             barrier_interval_ms,
                             progress_ratio
                         );
--- a/src/stream/src/executor/now.rs
+++ b/src/stream/src/executor/now.rs
@@ -38,6 +38,10 @@ pub struct NowExecutor<S: StateStore> {
     barrier_receiver: UnboundedReceiver<Barrier>,
 
     state_table: StateTable<S>,
+
+    progress_ratio: Option<f32>,
+
+    barrier_interval_ms: u32,
 }
 
 pub enum NowMode {
@@ -66,6 +70,8 @@ impl<S: StateStore> NowExecutor<S> {
         eval_error_report: ActorEvalErrorReport,
         barrier_receiver: UnboundedReceiver<Barrier>,
         state_table: StateTable<S>,
+        progress_ratio: Option<f32>,
+        barrier_interval_ms: u32,
     ) -> Self {
         Self {
             data_types,
@@ -73,6 +79,8 @@ impl<S: StateStore> NowExecutor<S> {
             eval_error_report,
             barrier_receiver,
             state_table,
+            progress_ratio,
+            barrier_interval_ms,
         }
     }
 
@@ -84,6 +92,8 @@ impl<S: StateStore> NowExecutor<S> {
             eval_error_report,
             barrier_receiver,
             mut state_table,
+            progress_ratio,
+            barrier_interval_ms,
         } = self;
 
         let max_chunk_size = crate::config::chunk_size();
@@ -116,7 +126,7 @@ impl<S: StateStore> NowExecutor<S> {
         for barriers in
             UnboundedReceiverStream::new(barrier_receiver).ready_chunks(MAX_MERGE_BARRIER_SIZE)
         {
-            let mut curr_timestamp = None;
+            let mut curr_timestamp: Option<ScalarImpl> = None;
             if barriers.len() > 1 {
                 warn!(
                     "handle multiple barriers at once in now executor: {}",
@@ -124,7 +134,7 @@ impl<S: StateStore> NowExecutor<S> {
                 );
             }
             for barrier in barriers {
-                let new_timestamp = Some(barrier.get_curr_epoch().as_scalar());
+                let new_timestamp = barrier.get_curr_epoch().as_timestamptz();
                 let pause_mutation =
                     barrier
                         .mutation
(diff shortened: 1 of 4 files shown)


lmatz:
Does it mean that it will be temporarily inconsistent when the now executor is still catching up with the current epoch?

hzxa21:
> Does it mean that it will be temporarily inconsistent when the now executor is still catching up with the current epoch?

I think consistency relies on all NOW emitting the same timestamp rather than emitting the latest timestamp so consistency is still there.

By default we will still stick with the original behaviour and when we saw big epoch caused by temporal filter, we can turn on this new behavior as a mitigation via developer config.

hzxa21:
@wenym1 @chenzl25 @lmatz I think we can release this in 2.6. PTAL.

src/stream/src/executor/now.rs
```diff
@@ -38,6 +38,10 @@ pub struct NowExecutor<S: StateStore> {
     barrier_receiver: UnboundedReceiver<Barrier>,
 
     state_table: StateTable<S>,
+
+    progress_ratio: Option<f32>,
+
+    barrier_interval_ms: u32,
```
wenym1, line 44: Seems wired to directly use `barrier_interval_ms`, because we support alter the `barrier_interval_ms` on meta node? Can we try instead dynamically use the time gap between we receive two batches of barriers as a replacement of `barrier_interval_ms`? We may store the last timestamp when we received the previous batch of barriers, and when we receive the next barrier, we minus the current timestamp by previous timestamp to be the gap. For the first barrier that an executor receives, it can use a default value, for example, 1s or the initial value of barrier_interval_ms read from param reader.

src/stream/src/executor/now.rs
```diff
@@ -116,15 +126,16 @@ impl<S: StateStore> NowExecutor<S> {
         for barriers in
             UnboundedReceiverStream::new(barrier_receiver).ready_chunks(MAX_MERGE_BARRIER_SIZE)
         {
-            let mut curr_timestamp = None;
+            let mut curr_timestamp: Option<ScalarImpl> = None;
```
wenym1: Seems more straightforward to directly use the inner timestamp type of `ScalarImpl::Timestamptz(timestamp type)`

wenym1: So is the `last_timestamp` above.

src/stream/src/executor/now.rs
```diff
@@ -152,7 +163,37 @@ impl<S: StateStore> NowExecutor<S> {
...
+                    && progress_ratio > 1.0
+                {
+                    // curr_timestamp = min(last_timestamp + barrier_interval * progress_ratio, timestamp from epoch)
+                    // to avoid having a big gap between the last timestamp and the current timestamp,
+                    // which may cause excessive changes in downstream dynamic filter
+                    let progress_timestamp = timestamp
+                        .timestamp_millis()
+                        .checked_add((barrier_interval_ms as f32 * progress_ratio).ceil() as i64)
+                        .expect("progress_timestamp is out of i64 range");
+                    let adjusted_timestamp = if progress_timestamp
```
wenym1, line 183: If the cluster is lagged for a long time, before the curr_timestamp catch up, we will keep yielding this log for every barrier, and may be flooding the log. Can we change to periodically yielding the log, like once per minute? Besides, we may also include the state table id in the log in case we may have multiple `now()` running in the cluster.

Actually, without this log, we can still monitor the lag by inspecting the state table of `now()`.

src/stream/src/executor/now.rs
hzxa21: Thought about that but I prefer not not introduce the dynamic time gap mechanism for the below reasons:
1. `progress_ratio` is disabled by default and will only be enabled via a developer config. That means it is served as a mitigation step instead of an always-on feature. I prefer keeping it simpler.
2. The `progress_ratio` here is used to reduce the amount of **historical** states scanned by the downstream so what matters here is the barrier interval in the past, not the currernt barrier interval.
3. The time gap calculation based on two batches of barriers can be tricky. For example
    a. If we use the epoch as the timestamp and the cluster is in congestion, the epoch between two batches of barriers can be large, which will result in a large progression on now and cause the same issue prior to this PR.
    b. If we use system timestamp, different now executor can get different timestamp, which breaks the now consistency.
4. Even when `barrier_interval_ms` changes in runtime, a recovery can bring things back to normal.


All in all, I prefer keeping things simpler because we can reason about the behavior (not neccessarily optimal but more deterministic) by only tuning a single knob `stream_now_progress_ratio` under `[streaming.developer]` when it comes to issue mitigation.

src/stream/src/executor/now.rs
hzxa21: Good point. I changed it to debug log and add an info log to print out the configs when now starts.

chenzl25 (approved):
LGTM!

src/stream/src/executor/now.rs
hzxa21: It will benefit the stream chunk generation and writing to state table by using `Datum` directly. I have renamed the variables to make it clearer and also uses `as_timestamptz` [here](https://github.com/risingwavelabs/risingwave/blob/28fafe88e13b7e516461639d235e7182784b88de/src/stream/src/executor/now.rs#L175) to make sure the datum type must be timestamptz.

PR #22503, opened 2025-07-04
fix(cdc): use `lsn_commit` and `lsn_proc` for pg cdc offset

### LINKED ISSUE #22369: Use lsn_commit and lsn_proc as offset for Postgres CDC
Currently we are using `lsn` for Postgres CDC offset, instead we should use `lsn_commit` and `lsn_proc` to get a total order between messages in interleaving transactions.

yuhao-su:

fix #22369

This pull request will use `lsn_commit` and `lsn_proc` to give a total order to debezium change event.

Previously the `lsn` we used only reflect the order of the event inside a transaction. 

When there are interleaving TXs, consider the following case:
```
LSN 100  T1: INSERT …
LSN 108  T2: INSERT …
LSN 112  T1: INSERT …
LSN 118  T2: COMMIT        -- T2 commits first
LSN 124  T1: INSERT …
LSN 130  T1: COMMIT        -- T1 commits later
```
The change events from debezium will be grouped together by TXs. So they will arrive as

```
LSN 108  T2: INSERT …
LSN 118  T2: COMMIT        -- T2 commits first
LSN 100  T1: INSERT …
LSN 112  T1: INSERT …
LSN 124  T1: INSERT …
LSN 130  T1: COMMIT        -- T1 commits later
```

When T2 and T1 were separated into 2 chunks, the `LSN 110 112` in T1 could be filtered by `mark_cdc_chunk` when backfilling, resulting in missing data.

### Test coverage improvements:

* **Postgres Offset Tests (`src/connector/src/source/cdc/external/postgres.rs`)**: Added tests for `PartialOrd` behavior with new fields and for Debezium offset parsing to validate field presence and error handling.

### Code cleanup:

* **Snapshot Reading Logic (`src/stream/src/executor/backfill/cdc/upstream_table/snapshot.rs`)**: Removed unreachable code in the snapshot reading loop for improved clarity.

## Checklist

## Documentation

<details>

</details>


yuhao-su:
> However, I think it is better to have a test coverage on the issue. 

It's very hard to do the test because it happens when backfilling. I did modify some code locally to reproduce it more easily.

src/connector/src/source/cdc/external/postgres.rs
```diff
@@ -40,12 +40,35 @@ pub struct PostgresOffset {
...
+        match (
+            self.lsn_commit,
+            self.lsn_proc,
+            other.lsn_commit,
+            other.lsn_proc,
+        ) {
+            (_, Some(self_proc), _, Some(other_proc)) => {
+                // if both have `lsn_commit` and `lsn_proc`, compare `lsn_commit` first, then `lsn_proc`
+                // if `lsn_commit` is None, fall back to `lsn_proc`
+                match self.lsn_commit.partial_cmp(&other.lsn_commit) {
```
BugenZhao: This looks confusing to me as it's mixing `Option` from field and `Option` from `partial_ord`. There's seems no need to implement `PartialOrd`. We may consider implement `Ord` instead and forward `PartialOrd` impl to `Some(self.cmp(other))`.

yuhao-su: > mixing Option from field and Option from partial_ord

I not sure if I got your point. But the returning value of the implemented partialord are all from `partial_cmp` but not directly related to the vlaue of option of `lsn_proc`

wcy-fdu (approved):
LGTM! Good Catch!

src/connector/src/source/cdc/external/postgres.rs
BugenZhao: All fields here are primitive and implement `Ord`. I mean shall we also directly implement `Ord` here for clarity?

yuhao-su: It is indeed a total order. I'll try do that

BugenZhao (approved):

src/connector/src/source/cdc/external/postgres.rs
```diff
@@ -56,15 +102,27 @@ impl PostgresOffset {
...
+
+        // `lsn_commit` may not be present in the offset for the first tx.
+        let lsn_commit = dbz_offset.source_offset.lsn_commit;
+
+        let lsn_proc = dbz_offset
+            .source_offset
+            .lsn_proc
+            .context("invalid postgres lsn_proc")?;
+
         Ok(Self {
```
zwang28, line 130: `lsn_commit` and `lsn_proc` should be set in `PostgresExternalTableReader::current_cdc_offset` as well?
Although I guess it's difficult / impossible to obtain them?

zwang28: It leads to the branch? i.e. `lsn_commit and lsn_proc being both None` causes the fix in this PR being skipped, which means the `filtering by offset logic` is still not 100% safe.
```
// Fall back to lsn comparison when either lsn_commit or lsn_proc is missing
self.lsn.cmp(&other.lsn)
```

yuhao-su: lsn_commit is maintained by dbz not pg. And there is no way to have that before the last tx commit.

> which means the filtering by offset logic is still not 100% safe.

You are probably right. The first `current_cdc_offset` can still cause unsafe filter.

zwang28: Can we initialize the `last_binlog_offset` on the first `get_cdc_chunk_last_offset` rather than by `PostgresExternalTableReader::current_cdc_offset` in `CdcBackfillExecutor`?
Note that if applying this change, do remember to change [this](https://github.com/risingwavelabs/risingwave/blob/main/src/stream/src/executor/backfill/cdc/state.rs#L71) because it can be None now.

zwang28: > // Fall back to lsn comparison when either lsn_commit or lsn_proc is missing
> self.lsn.cmp(&other.lsn)

We can add a warning here, and merge this PR. Since this PR has resolved the major issue.

zwang28 (approved):

hzxa21:
CI fails at `different_pk_and_dist_key`. Confirmed with @wenym1. It is a flaky test caused by #23071. Retry can help.

PR #22527, opened 2025-07-07
feat(iceberg): introduce small file compaction

Li0k:

This PR introduces Small Data File Compaction  and modifies the default behavior of the append-only Iceberg sink. It switches the default from Full Compaction to Small Data File Compaction to reduce write amplification and space amplification.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/handler/create_table.rs
+++ b/src/frontend/src/handler/create_table.rs
@@ -1696,11 +1696,10 @@ pub async fn create_iceberg_engine_table(
 
     let mut sink_with = with_common.clone();
 
-    sink_with.insert("primary_key".to_owned(), pks.join(","));
-
     if table.append_only {
         sink_with.insert("type".to_owned(), "append-only".to_owned());
     } else {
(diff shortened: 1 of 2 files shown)


src/frontend/src/handler/create_table.rs
```diff
@@ -1693,7 +1693,12 @@ pub async fn create_iceberg_engine_table(
     let mut sink_with = with_common.clone();
 
     sink_with.insert("primary_key".to_owned(), pks.join(","));
-    sink_with.insert("type".to_owned(), "upsert".to_owned());
+
+    if table.append_only {
+        sink_with.insert("type".to_owned(), "append-only".to_owned());
+    } else {
+        sink_with.insert("type".to_owned(), "upsert".to_owned());
+    }
```
chenzl25: For append_only iceberg table, we should remove the `sink_with.insert("primary_key".to_owned(), pks.join(","));`

PR #22529, opened 2025-07-07
fix: construct user info after secret and conn privilege granted or revoked

yezizp2012:

Left by #20755 .

## Checklist

## Documentation

<details>

</details>


src/frontend/src/catalog/system_catalog/rw_catalog/rw_secrets.rs
```diff
@@ -32,6 +33,9 @@ struct RwSecret {
 fn read_rw_view_info(reader: &SysCatalogReaderImpl) -> Result<Vec<RwSecret>> {
```
shanicky: view?

PR #22530, opened 2025-07-07
fix(iceberg): spwan blocking task for dropping iceberg table

chenzl25:

- Dropping an iceberg table in JNI could be blocking for a long time, since it would purge the data from object store by default, so we should spwan blocking this task .

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/src/connector_common/iceberg/jni_catalog.rs
+++ b/src/connector/src/connector_common/iceberg/jni_catalog.rs
@@ -364,28 +364,42 @@ impl Catalog for JniCatalog {
 
     /// Drop a table from the catalog.
     async fn drop_table(&self, table: &TableIdent) -> iceberg::Result<()> {
-        execute_with_jni_env(self.jvm, |env| {
-            let table_name_str = format!(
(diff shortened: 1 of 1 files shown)


src/connector/src/connector_common/iceberg/jni_catalog.rs
```diff
@@ -364,28 +364,42 @@ impl Catalog for JniCatalog {
 
     /// Drop a table from the catalog.
     async fn drop_table(&self, table: &TableIdent) -> iceberg::Result<()> {
-        execute_with_jni_env(self.jvm, |env| {
```
BugenZhao, line 367: Oh, does it mean that the closure is actually invoked in the current thread, that is, every system thread can be a Java thread? I used to think it will send the closure to a "Java thread pool" or something else..

chenzl25: Yes. We call `jvm.attach_current_thread` inside execute_with_jni_env, which attaches the current thread to the Java VM.

Issue #22474, opened 2025-07-02
Scale down CDC executors to single paralellelism after initial snapshot is completed


zwang28:

zwang28:
We don't scale down for now, since it requires a command barrier to rebuild actor.
We've all relevant function ready so it's trivial to add this when necessary.

PR #22536, opened 2025-07-08
feat(cdc): add support for parallelized cdc backfill

### LINKED ISSUE #22472: Determine CDC snapshot splits when creating table with shared CDC source

### LINKED ISSUE #22473: Add support for parallelized CdcBackfillExecutor

zwang28:

This PR support parallelized CDC backfill for Postgres CDC. Close #22472. Close #22473. Part of #22474.
- The parallelized CDC backfill is disabled by default, so the original CDC backfill strategy is used. 
  - It means this PR won't impact CDC tables that have been using the original CDC backfill strategy.
  - Set the new CDC table option `backfill_num_rows_per_split` > 0 when creating a new CDC table to enable parallelized CDC backfill for this table.
- When a new CDC table is created, meta node will calculate and persist its snapshot splits.
  - The snapshot splits are generated based on the `backfill_num_rows_per_split`.
  - The snapshot splits are assigned to actors during recovery / reschedule.
- A new `ParallelizedCdcBackfillExecutor` is implemented. It uses these snapshot splits during backfill.
  - The parallelism of this fragment is determined by the new CDC table option `backfill_parallelism`. It can be modified later via risectl.
  - The actor backfills the snapshot splits sequentially.
  - The actor filters and processes WAL according to the the merged range of snapshot splits assigned to it.
- A new ctl command is added to allow modifying CDC table backfill parallelism.
  - It can be used either during or after CDC table backfill is finished.
  - Maximum parallelism is hard-coded to 256.

## Checklist

## Documentation

<details>

</details>


src/connector/src/source/cdc/external/mod.rs
```diff
@@ -87,17 +88,20 @@ impl CdcTableType {
...
             )),
             Self::Postgres => Ok(ExternalTableReaderImpl::Postgres(
-                PostgresExternalTableReader::new(config, schema, pk_indices).await?,
+                PostgresExternalTableReader::new(config, schema, pk_indices, schema_table_name)
+                    .await?,
             )),
             Self::SqlServer => Ok(ExternalTableReaderImpl::SqlServer(
                 SqlServerExternalTableReader::new(config, schema, pk_indices).await?,
             )),
+            // zw: why no matching arm for Citus?
```
hzxa21: Because citus is never supported for cdc backfill (create source + create table).

src/connector/src/source/cdc/external/postgres.rs
```diff
@@ -104,13 +107,59 @@ impl ExternalTableReader for PostgresExternalTableReader {
     ) -> BoxStream<'_, ConnectorResult<OwnedRow>> {
         self.snapshot_read_inner(table_name, start_pk, primary_keys, limit)
```
hzxa21, line 109: nits: check whether `table_name == self.schema_table_name`?

src/frontend/src/optimizer/plan_node/stream_cdc_table_scan.rs
```diff
@@ -64,37 +64,51 @@ impl StreamCdcTableScan {
 
     /// Build catalog for cdc backfill state
     /// Right now we only persist whether the backfill is finished and the corresponding cdc offset
-    /// schema: | `split_id` | `pk...` | `backfill_finished` | `row_count` | `cdc_offset` |
+    /// schema: | `split_id` | `backfill_finished` | `row_count` | `cdc_offset` |
```
hzxa21: I think `pk` is in the schema for non-parallelized backfill?

src/meta/src/barrier/checkpoint/creating_job/mod.rs
```diff
@@ -83,10 +83,6 @@ impl CreatingStreamingJobControl {
             "new creating job"
         );
         let snapshot_backfill_actors = info.stream_job_fragments.snapshot_backfill_actor_ids();
-        let backfill_nodes_to_pause =
```
hzxa21: Why do we remove this?

src/meta/src/stream/cdc.rs
```diff
@@ -0,0 +1,261 @@
...
+    txn.commit().await?;
+    // TODO(zw): handle metadata backup&restore
+    Ok(())
+}
+
+/// Returns true if the fragment is CDC scan and has parallelized backfill enabled.
+fn is_parallelized_backfill_enabled_cdc_scan_fragment(fragment: &Fragment) -> bool {
+    match &fragment.nodes.node_body {
+        Some(NodeBody::StreamCdcScan(node)) => is_parallelized_backfill_enabled(node),
+        Some(NodeBody::Project(_)) => {
```
hzxa21: Is it safe to only check this two node body in a flat manner? Shall we use visitor to find out whether it is parallel backfilling?

src/meta/src/rpc/ddl_controller.rs
```diff
@@ -824,6 +864,20 @@ impl DdlController {
                 .create_split_enumerator(SourceEnumeratorContext::dummy().into())
                 .await?;
 
+            if is_parallelized_backfill_enabled(stream_cdc_scan) {
+                // Create parallel splits for a CDC table. The resulted split assignments are persisted and immutable.
+                try_init_parallel_cdc_table_snapshot_splits(
```
hzxa21, line 873: It looks a bit counter-intuitive to init the splits in the `validate` call. What happen if `validate` completes succesfully and the cdc table creation fails? I don't find a place where we delete the rows in `cdc_table_snapshot_split` meta store table.

src/connector/src/source/cdc/external/postgres.rs
```diff
@@ -254,6 +351,357 @@ impl PostgresExternalTableReader {
...
+    }
+
+    async fn next_split_right_bound_exclusive(
+        &self,
+        left_value: &ScalarImpl,
+        max_value: &ScalarImpl,
+        max_split_size: u64,
+        split_column: &Field,
+    ) -> ConnectorResult<Option<Datum>> {
+        let sql = format!(
```
hzxa21, line 391: This is for performance and not for correctness. I am thinking which one is better: 
1. running the current SQL per split sequentially to fetch the split boundary
2. running a SQL with offset per split sequentially to fetch the split boundary. Example:
```sql
SELECT id FROM t
WHERE id >= left_value
ORDER BY id_column ASC
LIMIT 1
OFFSET max_rows_in_split;
```
3. running a single SQL to fetch all the split boundaries at once. Example (suggested by Gemini):
```sql
WITH RankedData AS (
    -- Assign a row number to each record, ordered by the id_column.
    -- This helps in distributing rows evenly across chunks.
    SELECT
        id_column,
        ROW_NUMBER() OVER (ORDER BY id_column) AS rn,
        COUNT(*) OVER () AS total_rows -- Get the total number of rows in the table
    FROM
        your_table_name
),
ChunkedData AS (
    -- Divide the ranked data into the specified number of chunks.
    -- NTILE(N) distributes rows into N groups as equally as possible.
    SELECT
        id_column,
        rn,
        total_rows,
        NTILE(CAST(total_rows / 100000.0 + 0.99999 AS INTEGER)) OVER (ORDER BY rn) AS chunk_id
        -- The NTILE argument (number_of_chunks) is dynamically calculated here.
        -- It aims to create chunks of approximately 100,000 rows each.
        -- Adjust '100000.0' to your desired chunk size.
        -- We add 0.99999 and cast to INTEGER to ensure we get enough chunks
        -- even if total_rows is not perfectly divisible.
)
-- Select the minimum and maximum id_column for each generated chunk.
SELECT
    chunk_id,
    MIN(id_column) AS min_id,
    MAX(id_column) AS max_id,
    COUNT(id_column) AS rows_in_chunk
FROM
    ChunkedData
GROUP BY
    chunk_id
ORDER BY
    chunk_id;
```

I haven't verfied this but I think in all cases, PG will perform a full scan.

hzxa21 (commented):
Post some comments first. I am still reading the executor parts.

proto/source.proto
```diff
@@ -18,3 +18,13 @@ message SourceActorInfo {
...
   ConnectorSplits splits = 2;
 }
+
+message CdcTableSnapshotSplits {
+  repeated CdcTableSnapshotSplit splits = 1;
+}
+
+message CdcTableSnapshotSplit {
+  int64 split_id = 1;
+  bytes left = 2;
```
hzxa21: nits: how about including the `inclusive/exclusive` in the naming of `left/right`?

hzxa21:
@wenym1 can you take a second look on the changes under `meta/src/barrier`

src/meta/src/stream/cdc.rs
```diff
@@ -0,0 +1,261 @@
...
+            right: Set(split.right_bound_exclusive.value_serialize()),
+        });
+        if insert_batch.len() >= insert_batch_size as usize {
+            cdc_table_snapshot_split::Entity::insert_many(std::mem::take(&mut insert_batch))
+                .exec(&txn)
+                .await?;
+        }
+        if splits_num % sleep_split_interval == 0 {
+            tokio::time::sleep(Duration::from_millis(sleep_duration_millis)).await;
+        }
```
hzxa21, line 119: Why do we need the sleep here? To reduce the load to upstream cdc source db?

src/meta/src/stream/scale.rs
```diff
@@ -2037,7 +2093,10 @@ impl ScaleController {
             ?target_plan,
             "generate_table_resize_plan finished target_plan"
         );
-
+        if generate_plan_for_cdc_table_backfill {
+            job_reschedule_post_updates.resource_group_updates = HashMap::default();
+            job_reschedule_post_updates.parallelism_updates = HashMap::default();
```
hzxa21, line 2098: IIUC, if `generate_plan_for_cdc_table_backfill` is true, `resource_group_updates` and `parallelism_updates` must be empty. Should we do an assertion here instead of overwriting them.

@shanicky please help to review the changes related to scaling in this PR as well.

src/meta/src/rpc/ddl_controller.rs
hzxa21: I saw fk and on delete cascade specified [here](https://github.com/risingwavelabs/risingwave/blob/51797989257e2f7e9a31d56561cebb79dba326f1/src/meta/model/migration/src/m20250702_062029_cdc_table_snapshot_splits.rs#L49). Does it mean that the state will be auto deleted when the table is dropped or failed in creation?

src/connector/src/source/cdc/external/postgres.rs
```diff
@@ -244,6 +294,53 @@ impl PostgresExternalTableReader {
         format!("({}) > ({})", col_expr, arg_expr)
     }
 
+    // row filter expression: (v1, v2, v3) >= ($1, $2, $3) AND (v1, v2, v3) < ($1, $2, $3)
+    fn split_filter_expression(
+        columns: &[String],
```
hzxa21, line 301: Just to confirm, `columns` will also have length == 1 in the current implementation, right?

src/connector/src/source/cdc/external/postgres.rs
```diff
@@ -254,6 +351,357 @@ impl PostgresExternalTableReader {
...
+    #[try_stream(boxed, ok = OwnedRow, error = ConnectorError)]
+    async fn split_snapshot_read_inner(
+        &self,
+        table_name: SchemaTableName,
+        left: OwnedRow,
+        right: OwnedRow,
+        split_columns: Vec<Field>,
+    ) {
+        assert_eq!(left.len(), 1, "multiple split columns is not supported yet");
+        assert_eq!(
```
hzxa21, line 490: should we assert the length of `split_columns` as well?

src/stream/src/executor/mod.rs
```diff
@@ -302,6 +309,7 @@ pub struct AddMutation {
     pub subscriptions_to_add: Vec<(TableId, u32)>,
     /// nodes which should start backfill
     pub backfill_nodes_to_pause: HashSet<FragmentId>,
+    pub actor_cdc_table_snapshot_splits: CdcTableSnapshotSplitAssignment,
```
hzxa21, line 324: nits: some documentation will help because no everyone knows about the cdc parallel backfilling design

src/stream/src/executor/backfill/cdc/cdc_backill_v2.rs
```diff
@@ -0,0 +1,664 @@
...
+            let all_snapshot_splits = match reset_barrier.mutation.as_deref() {
+                Some(Mutation::Add(add)) => &add.actor_cdc_table_snapshot_splits,
+                Some(Mutation::Update(update)) => &update.actor_cdc_table_snapshot_splits,
+                _ => {
+                    return Err(anyhow::anyhow!("ParallelizedCdcBackfillExecutor expects either Mutation::Add or Mutation::Update to initialize CDC table snapshot splits.").into());
+                }
+            };
+            let mut actor_snapshot_splits = vec![];
+            // TODO(zw): optimization: remove consumed splits to reduce barrier size for downstream.
+            if let Some(splits) = all_snapshot_splits.get(&self.actor_ctx.id) {
```
hzxa21, line 143: Should we assert or print a log when `all_snapshot_splits` doesn't contain any split for the current actor id?

src/stream/src/executor/backfill/cdc/cdc_backill_v2.rs
```diff
@@ -0,0 +1,664 @@
...
+            }
+            tracing::debug!(?actor_snapshot_splits, "actor splits");
+
+            let mut is_snapshot_paused = reset_barrier.is_pause_on_startup();
+            let barrier_epoch = reset_barrier.epoch;
+            yield Message::Barrier(reset_barrier);
+            if !is_reset {
+                state_impl.init_epoch(barrier_epoch).await?;
+                is_reset = true;
+                tracing::info!("Initialize executor.");
```
hzxa21: nits: include table_id in the logs?

src/stream/src/executor/backfill/cdc/cdc_backill_v2.rs
```diff
@@ -0,0 +1,664 @@
...
+            } else {
+                tracing::info!("Reset executor.");
```
hzxa21: ditto

src/stream/src/executor/backfill/cdc/cdc_backill_v2.rs
```diff
@@ -0,0 +1,664 @@
...
+                let state = state_impl.restore_state(split.split_id).await?;
+                if !state.is_finished {
+                    next_split_idx = idx;
+                    break;
+                }
+                extends_current_actor_bound(&mut current_actor_bounds, split);
+            }
+            for split in actor_snapshot_splits.iter().skip(next_split_idx) {
+                // Initialize state so that overall progress can be measured.
+                state_impl.mutate_state(split.split_id, false, 0).await?;
```
hzxa21, line 192: This means that we will redo the backfill for the unfinished split from scratch on recovery, right?

src/stream/src/executor/backfill/cdc/cdc_backill_v2.rs
```diff
@@ -0,0 +1,664 @@
...
+            let mut current_actor_bounds = None;
+            // Find next split that need backfill.
+            let mut next_split_idx = actor_snapshot_splits.len();
+            for (idx, split) in actor_snapshot_splits.iter().enumerate() {
+                let state = state_impl.restore_state(split.split_id).await?;
+                if !state.is_finished {
+                    next_split_idx = idx;
+                    break;
+                }
+                extends_current_actor_bound(&mut current_actor_bounds, split);
```
hzxa21, line 188: We rely on the assumption that all the assigned splits are consecutive in the split key. Can we have a check in this new executor to ensure that?

src/frontend/src/optimizer/plan_node/stream_cdc_table_scan.rs
zwang28: Will fix the comment.

src/meta/src/barrier/checkpoint/creating_job/mod.rs
zwang28: unexpected removal. Will fix.

src/meta/src/rpc/ddl_controller.rs
zwang28: > Does it mean that the state will be auto deleted when the table is dropped or failed in creation?

Yes.

src/meta/src/stream/cdc.rs
zwang28: > To reduce the load to upstream cdc source db

Yes. It's configurable so it can be turned off whenever necessary.

src/connector/src/source/cdc/external/postgres.rs
zwang28: Yes.

src/stream/src/executor/backfill/cdc/cdc_backill_v2.rs
zwang28: > all_snapshot_splits doesn't contain any split for the current actor id

It's an expected case. For example, backfill parallelism > number of snapshot splits.

src/stream/src/executor/backfill/cdc/cdc_backill_v2.rs
zwang28: Yes, it's by design.

Actually if will even redo finished split after reschedule:
1. In total 2 backfill parallelism with assigned splits [1,2,3] and [4,5,6] respectively.
2. splits [1,2] and [4,5] have been finished.
3. Scale down to 1 backfill parallelism in total, with assigned splits [1,2,3,4,5,6].
4. This new backfill paralellism will treat [1,2] as finished and [3,4,5,6] as unfinished.

Reschedle is rare and this design keep simplicity, i.e. avoiding tracking progress of all splits to filter WAL.

src/meta/src/stream/cdc.rs
zwang28: It's the same methods used by original `validate_cdc_table`.
I'll refactor it.

src/connector/src/source/cdc/external/postgres.rs
zwang28: Method 3 appears to be more efficient. But it may result in overlapped split ranges.

~~Let me try whether RANK works.~~ Not work.

src/meta/src/stream/scale.rs
zwang28: > IIUC, if generate_plan_for_cdc_table_backfill is true, resource_group_updates and parallelism_updates must be empty

Since I keep the original code snippet unchanged, `parallelism_updates` won't be empty.
```
job_reschedule_post_updates
    .parallelism_updates
    .insert(TableId::from(*job_id), parallelism);
```

Either adding a `if !generate_plan_for_cdc_table_backfill {}` for the code snippet above, or overwriting them in the end like what I'm doing.

hzxa21 (approved):
LGTM. Thanks for the efforts!

Given that it is a large PR, please wait for @wenym1's review on the barrier manager parts and @shanicky's review on the scaling parts before merging to main.

src/connector/src/source/cdc/external/postgres.rs
zwang28: The cost of 1&2 are close. Scan is the dominant cost.

```
risingwave=> explain WITH t as (SELECT id3 FROM pg_cdc_table WHERE id3 >= 100 ORDER BY id3 ASC LIMIT 2) SELECT CASE WHEN MAX(id3) < 222 THEN MAX(id3) ELSE NULL END FROM t;
                                                      QUERY PLAN
----------------------------------------------------------------------------------------------------------------------
 Aggregate  (cost=1.24..1.25 rows=1 width=8)
   ->  Limit  (cost=0.57..1.23 rows=2 width=8)
         ->  Index Only Scan using pg_cdc_table_pkey on pg_cdc_table  (cost=0.57..38304333.82 rows=115582238 width=8)
               Index Cond: (id3 >= 100)
(4 rows)

risingwave=> explain SELECT CASE WHEN id3 < 222 THEN id3 ELSE NULL END FROM pg_cdc_table WHERE id3 >= 100 ORDER BY id3 ASC LIMIT 1 OFFSET 2;
;
                                                   QUERY PLAN
-----------------------------------------------------------------------------------------------------------------
 Limit  (cost=1.24..1.57 rows=1 width=16)
   ->  Index Only Scan using pg_cdc_table_pkey on pg_cdc_table  (cost=0.57..38593289.42 rows=115582238 width=16)
         Index Cond: (id3 >= 100)
(3 rows)
```

hzxa21:
@zwang28 Can you help complete the release note section given that this is a user facing change? I missed that during my review.

PR #22543, opened 2025-07-08
fix: add join miss and independent panel

lmatz:

## Checklist

## Documentation

<details>

</details>


lmatz:
Actually, the existing metrics do not reveal what we really care about. 

There are three cases:
1. in the operator cache -> cache hit
2. not in the operator cache and not in the storage -> no existence at all
3. not in the operator cache but in the storage -> cache miss

1 and 2 are super fast as 2's meta file is in the memory

What we care about is (3) only because reading from disk cache or remote storage both incurr non-negligible latency.

The exisitng lookup miss count include both (2) and (3).

PR #22556, opened 2025-07-10
fix(cdc): add text type check for auto schema change

wcy-fdu:

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/common/src/types/mod.rs
+++ b/src/common/src/types/mod.rs
@@ -143,7 +143,7 @@ pub enum DataType {
     #[from_str(regex = "(?i)^date$")]
     Date,
     #[display("character varying")]
-    #[from_str(regex = "(?i)^character varying$|^varchar$")]
+    #[from_str(regex = "(?i)^character varying$|^varchar$|^text$")]
     Varchar,
     #[display("time without time zone")]
     #[from_str(regex = "(?i)^time$|^time without time zone$")]
@@ -1480,6 +1480,8 @@ mod tests {
 
         assert_eq!(DataType::from_str("varchar").unwrap(), DataType::Varchar);
         assert_eq!(DataType::from_str("VARCHAR").unwrap(), DataType::Varchar);
+        assert_eq!(DataType::from_str("text").unwrap(), DataType::Varchar);
+        assert_eq!(DataType::from_str("TEXT").unwrap(), DataType::Varchar);
 
         assert_eq!(DataType::from_str("time").unwrap(), DataType::Time);
         assert_eq!(
@@ -1555,6 +1557,10 @@ mod tests {
         );
         assert_eq!(
             DataType::from_str("varchar[]").unwrap(),
+            DataType::List(Box::new(DataType::Varchar))
+        );
+        assert_eq!(
+            DataType::from_str("text[]").unwrap(),
             DataType::List(Box::new(DataType::Varchar))
         );
         assert_eq!(


src/common/src/types/mod.rs
```diff
@@ -143,7 +143,7 @@ pub enum DataType {
     #[from_str(regex = "(?i)^date$")]
     Date,
     #[display("character varying")]
-    #[from_str(regex = "(?i)^character varying$|^varchar$")]
+    #[from_str(regex = "(?i)^character varying$|^varchar$|^text$")]
```
xiangjinwu: I thought this `from_str` was only used by unit tests. Could you point out how this is used by PG CDC?

wcy-fdu: I think we should add two method: `type_name_to_pg_type` and `pg_type_to_rw_type`, just like how mysql did. Here we should test as many data types as possible, I will complete it in the following PR (if there are any omissions).

wcy-fdu:
I suggest that this PR that is only for `text` be merged first, because there is a potential user who is about to start using pg-cdc with auto schema change, and their schema has a large number of text types. We will conduct a series of stability tests (data_type, auto schema change, e2e schema change, parallel backfill) in the near future, but due to the lack of support for `text` type, the first step of the test is blocked.

wcy-fdu:
cc @xiangjinwu PTAL😁

src/connector/src/source/cdc/external/postgres.rs
```diff
@@ -256,6 +257,56 @@ impl PostgresExternalTableReader {
...
+    match ty_name.to_lowercase().as_str() {
+        "smallint" | "int2" => Some(PgType::INT2),
+        "integer" | "int" | "int4" => Some(PgType::INT4),
+        "bigint" | "int8" => Some(PgType::INT8),
+        "real" | "float4" => Some(PgType::FLOAT4),
+        "double precision" | "float8" => Some(PgType::FLOAT8),
+        "numeric" | "decimal" => Some(PgType::NUMERIC),
+        "boolean" | "bool" => Some(PgType::BOOL),
+        "varchar" | "character varying" => Some(PgType::VARCHAR),
+        "char" | "character" | "bpchar" => Some(PgType::BPCHAR),
```
xiangjinwu, line 270: This is the ambiguous part and why we should know whether upstream is sending `smallint` or `int2` - it cannot be both.
|oid|display|typname|
|-|-|-|
|21|smallint|int2|
|1042|char|bpchar|
|18|"char"|char|

Without knowing whether we are parsing display name or internal typname, we do not know whether `char` means 1042 or 18, although both would eventually be mapped to RisingWave's varchar.

src/connector/src/source/cdc/external/postgres.rs
```diff
@@ -256,6 +257,56 @@ impl PostgresExternalTableReader {
...
+        "date" => Some(PgType::DATE),
+        "time" | "time without time zone" => Some(PgType::TIME),
+        "timestamp" | "timestamp without time zone" => Some(PgType::TIMESTAMP),
+        "timestamptz" | "timestamp with time zone" => Some(PgType::TIMESTAMPTZ),
+        "interval" => Some(PgType::INTERVAL),
+        "json" => Some(PgType::JSON),
+        "jsonb" => Some(PgType::JSONB),
+        "uuid" => Some(PgType::UUID),
+        "point" => Some(PgType::POINT),
+        _ => None,
```
xiangjinwu, line 282: What about array types? For example `int[]` has an internal name `_int4`

src/connector/src/source/cdc/external/postgres.rs
```diff
@@ -256,6 +257,56 @@ impl PostgresExternalTableReader {
...
+        PgType::FLOAT8 => DataType::Float64,
+        PgType::NUMERIC => DataType::Decimal,
+        PgType::DATE => DataType::Date,
+        PgType::TIME => DataType::Time,
+        PgType::TIMESTAMP => DataType::Timestamp,
+        PgType::TIMESTAMPTZ => DataType::Timestamptz,
+        PgType::INTERVAL => DataType::Interval,
+        PgType::VARCHAR | PgType::TEXT | PgType::BPCHAR => DataType::Varchar,
+        PgType::BYTEA => DataType::Bytea,
+        PgType::JSON | PgType::JSONB => DataType::Jsonb,
```
xiangjinwu, line 302: Do we already have similar logic in the code base? Let's aim to unify them eventually. That could help us avoid problems like forgetting to map UUID to Varchar or handle any array types here.

src/connector/src/parser/unified/debezium.rs
```diff
@@ -204,12 +205,18 @@ pub fn parse_schema_change(
...
+                            let ty = type_name_to_pg_type(type_name.as_str());
+                            match ty {
+                                Some(ty) => pg_type_to_rw_type(&ty).map_err(|err| {
+                                    tracing::warn!(error=%err.as_report(), "unsupported postgres type in schema change message");
+                                    AccessError::UnsupportedType {
+                                        ty: type_name.clone(),
+                                    }
+                                })?,
+                                None => {
+                                    Err(AccessError::UnsupportedType { ty: type_name })?
```
xiangjinwu, line 217: Did you intend to split these 2 different error cases? Why only logging in the first case?

xiangjinwu (commented):
Waiting to see the effectiveness of new test - it should fail without the fix:
https://buildkite.com/risingwavelabs/pull-request/builds/78513

wcy-fdu:
This pr can be included in https://github.com/risingwavelabs/risingwave/pull/22610.

Don't close the current PR for now, because we are not sure how long it will take to merge the complete fix PR(https://github.com/risingwavelabs/risingwave/pull/22610), and this PR may block some users. Set it to draft first, and merge it in advance if necessary.

wcy-fdu:
Cherry picked into specific branch.

PR #22557, opened 2025-07-10
feat: Add CASCADE support for DROP CONNECTION

tabVersion:

**ALL CODE IS GENERATED FROM CLAUDE CODE**

### Summary
This PR implements CASCADE functionality for the `DROP CONNECTION` statement, allowing users to automatically drop dependent objects when dropping a connection.

#### New Syntax Support
- **`DROP CONNECTION connection_name CASCADE`** - Drops the connection and all dependent objects (tables, sinks, etc.)
- **`DROP CONNECTION IF EXISTS connection_name CASCADE`** - Same as above, but doesn't error if the connection doesn't exist
- **`DROP CONNECTION connection_name RESTRICT`** - Explicitly prevents dropping if dependencies exist (existing behavior)

#### Behavior Changes
- **Before**: `DROP CONNECTION` would fail with "connection used by X other objects" if any tables, sinks, or other objects depended on the connection
- **After**: 
  - `DROP CONNECTION` (without CASCADE) - Same behavior as before (fails if dependencies exist)
  - `DROP CONNECTION CASCADE` - Automatically drops all dependent objects and then drops the connection
  - `DROP CONNECTION RESTRICT` - Explicitly fails if dependencies exist (same as default behavior)

#### Example Usage
```sql
-- Create connection and dependent objects
CREATE CONNECTION my_conn WITH (type = 'kafka', ...);
CREATE TABLE my_table (...) WITH (connector = 'kafka', connection = my_conn, ...);
CREATE SINK my_sink FROM data WITH (connector = 'kafka', connection = my_conn, ...);

-- This would fail before this PR
DROP CONNECTION my_conn;  -- Error: connection used by 2 other objects

-- New functionality - automatically drops my_table and my_sink first
DROP CONNECTION my_conn CASCADE;  -- ✅ Success
```

### Implementation Details
- Extended the DDL service protocol to support cascade mode
- Updated the frontend handler to process the CASCADE keyword
- Modified the meta controller to handle cascading drops for connections
- Added comprehensive end-to-end tests covering various scenarios

### Testing
- Added comprehensive SQLLogicTest suite in `e2e_test/source_inline/connection/drop_cascade.slt`
- Tests cover CASCADE with dependencies, without dependencies, IF EXISTS variants, and error cases
- Updated parser tests to validate new syntax

This change brings `DROP CONNECTION` in line with other RisingWave objects that already support CASCADE functionality.
## Checklist

## Documentation

<details>

refer to the user facing section above

</details>

### DIFF
--- a/src/frontend/src/handler/drop_connection.rs
+++ b/src/frontend/src/handler/drop_connection.rs
@@ -25,6 +25,7 @@ pub async fn handle_drop_connection(
     handler_args: HandlerArgs,
     connection_name: ObjectName,
     if_exists: bool,
+    cascade: bool,
 ) -> Result<RwPgResponse> {
     let session = handler_args.session;
     let db_name = &session.database();
@@ -59,7 +60,9 @@ pub async fn handle_drop_connection(
     };
 
     let catalog_writer = session.catalog_writer()?;
-    catalog_writer.drop_connection(connection_id).await?;
+    catalog_writer
+        .drop_connection(connection_id, cascade)
+        .await?;
 
     Ok(PgResponse::empty_result(StatementType::DROP_CONNECTION))
 }
--- a/e2e_test/source_inline/connection/drop_cascade.slt
+++ b/e2e_test/source_inline/connection/drop_cascade.slt
@@ -1,0 +1,139 @@
+control substitution on
+
+# Test DROP CONNECTION CASCADE functionality
+
+# for non-shared source
+statement ok
+set streaming_use_shared_source to false;
+
+# Create secret for connection
+statement ok
+create secret sec_cascade with (backend = 'meta') as '${RISEDEV_KAFKA_BOOTSTRAP_SERVERS}';
+
+# Create connection that will have dependencies
+statement ok
+create connection conn_cascade with (type = 'kafka', properties.bootstrap.server = secret sec_cascade, properties.security.protocol = 'plaintext');
+
+# Create Kafka topic for testing
+system ok
+rpk topic create connection_cascade_test -p 1
+
+# Create table that depends on the connection
+statement ok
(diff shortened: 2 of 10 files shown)


chenzl25 (commented):
I think we should ban dropping a connector with iceberg type, since iceberg engine relies on it.

tabVersion:
> I think we should ban dropping a connector with iceberg type, since iceberg engine relies on it.

Can you elaborate on how IcebergConnection serves differently?

chenzl25:
> > I think we should ban dropping a connector with iceberg type, since iceberg engine relies on it.
> 
> Can you elaborate on how IcebergConnection serves differently?

Here is the doc https://docs.risingwave.com/store/iceberg-table-engine

Iceberg engine relies on iceberg connection to setup the configuration, so if we drop cascade the connection, it would drop all the metadata in risingwave, but it would remain the iceberg data without being dropped. I think this PR doesn't handle it well, so let's ban this feature for iceberg connection first.

src/frontend/src/handler/drop_connection.rs
```diff
@@ -55,11 +58,30 @@ pub async fn handle_drop_connection(
...
     };
 
+    // Ban DROP CASCADE for Iceberg connections
+    if is_iceberg_connection && cascade {
+        return Err(ErrorCode::NotSupported(
+            "DROP CONNECTION CASCADE".to_owned(),
+            "Please drop dependent objects manually before dropping the Iceberg connection, or use DROP CONNECTION without CASCADE".to_owned(),
+        )
+        .into());
+    }
```
tabVersion, line 79: @chenzl25 here we ban drop connection cascade for Iceberg Connection

PR #22558, opened 2025-07-10
fix: Add Pulsar startup with timestamp 

tabVersion:

### Summary
Add e2e test case for Pulsar `scan.startup.timestamp.millis` functionality, similar to the existing Kafka startup timestamp test.

### Changes
- **New test files:**
  - `e2e_test/source_inline/pulsar/pulsar_startup_millis.slt` - Test control file
  - `e2e_test/source_inline/pulsar/pulsar_startup_millis.py` - Python test script
- **Dependencies:**
  - Added `pulsar-client` to `e2e_test/source_inline/requirements.txt`

### Test Logic
The test verifies that when creating a Pulsar source table with `scan.startup.timestamp.millis` property:
1. Produces initial records to the topic
2. Records a timestamp after a 1-second delay
3. Produces additional records
4. Creates a RisingWave table with the recorded timestamp as `scan.startup.timestamp.millis`
5. Verifies only records with timestamps >= the startup timestamp are consumed

### Technical Details
- Uses partitioned topic (1 partition) for more realistic testing
- Leverages existing `pulsar_util.py` for topic management
- Tests the `PulsarEnumeratorOffset::Timestamp` functionality and `consumer.seek()` implementation
- Follows the same pattern as the Kafka startup timestamp test for consistency

### Testing
This test ensures the Pulsar timestamp-based offset functionality works correctly, which is important for scenarios where users need to start consuming from a specific point in time rather than earliest/latest offsets.

## Checklist

## Documentation

<details>

</details>


e2e_test/source_inline/pulsar/pulsar_startup_millis.slt
xiangjinwu, line 1: This is not an inline-slt. Almost all testing is in the python script. We shall consider rewrite it with `$__NOW__`

xiangjinwu (commented):
Looks like the added tests can also pass on main without code changes in this PR:
https://buildkite.com/risingwavelabs/pull-request/builds/82408

What is the fix about? The PR description is off

PR #22559, opened 2025-07-10
fix(cdc): limit cdc offset fetch concurrency globally

chenzl25:

-  If a user has hundreds of CDC tables, and once recovery occurs, the CDC backfill will create a connection to the upstream database. Currently, every parallelism would create a connection to the upstream database even though the databases are the same. In this scenario, the cluster would encounter a `too many connections` issue, and a recovery loop happens.
- This PR limit the CDC offset fetch concurrence to 10 globally to workaround this issue. In the future, we can use a global database pool to mitigate this issue.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/executor/backfill/cdc/cdc_backfill.rs
+++ b/src/stream/src/executor/backfill/cdc/cdc_backfill.rs
@@ -246,9 +246,17 @@ impl<S: StateStore> CdcBackfillExecutor<S> {
         );
 
         let mut upstream = upstream.peekable();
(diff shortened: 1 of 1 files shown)


wcy-fdu (approved):
Thanks for the hot fix!

wcy-fdu:
I am thinking that if how can we limit the number of upstream connections in the future:
- for MySQL, since I have add connection pool for single table, I think we can use the global connection pool to manage it.
- for other cdc, haven't dive deep yet.

cc @hzxa21 for awareness

PR #22561, opened 2025-07-10
fix(frontend): fix topn on index optimization rule bug

chagelo:

Previously, the component of a composite index present in the equality constant expression of the WHERE clause was extracted, and the longest common index prefix was constructed in ascending order. However, some indexes might be in descending order.

Previous behavior:

```sql
dev=> create table t(a int, b int, c int);
CREATE_TABLE
dev=> create index t_index_ab on t(a desc, b asc);
CREATE_INDEX
dev=> explain select * from t where a=10 order by b limit 1;
                                        QUERY PLAN                                        
------------------------------------------------------------------------------------------
 BatchTopN { order: [t_index_ab.b ASC], limit: 1, offset: 0 }
 └─BatchExchange { order: [], dist: Single }
   └─BatchTopN { order: [t_index_ab.b ASC], limit: 1, offset: 0 }
     └─BatchScan { table: t_index_ab, columns: [a, b, c], scan_ranges: [a = Int32(10)] }
(4 rows)
```

Now:

```sql
BatchSort { order: [t_index_ab.b ASC] }
 └─BatchTopN { order: [t_index_ab.a DESC, t_index_ab.b ASC], limit: 1, offset: 0 }
   └─BatchExchange { order: [], dist: Single }
     └─BatchLimit { limit: 1, offset: 0 }
       └─BatchScan { table: t_index_ab, columns: [a, b, c], scan_ranges: [a = Int32(10)], limit: 1 }
```

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/optimizer/rule/top_n_on_index_rule.rs
+++ b/src/frontend/src/optimizer/rule/top_n_on_index_rule.rs
@@ -60,14 +60,22 @@ impl TopNOnIndexRule {
         let prefix = input_refs
             .into_iter()
             .flat_map(|input_ref| {
-                vec![
+                [
                     ColumnOrder {
                         column_index: input_ref.index,
-                        order_type: OrderType::ascending(),
+                        order_type: OrderType::ascending_nulls_first(),
                     },
                     ColumnOrder {
                         column_index: input_ref.index,
-                        order_type: OrderType::descending(),
+                        order_type: OrderType::ascending_nulls_last(),
+                    },
+                    ColumnOrder {
+                        column_index: input_ref.index,
+                        order_type: OrderType::descending_nulls_first(),
+                    },
+                    ColumnOrder {
+                        column_index: input_ref.index,
+                        order_type: OrderType::descending_nulls_last(),
                     },
                 ]
             })
--- a/src/frontend/planner_test/tests/testdata/input/index_selection.yaml
+++ b/src/frontend/planner_test/tests/testdata/input/index_selection.yaml
@@ -505,10 +505,19 @@
     select * from t1 where c = 1 and a = 2 and d = 3 order by b desc limit 10
   expected_outputs:
     - batch_plan
-- name: topn index selection + limit pushdown with point get
+- name: topn index selection + limit pushdown with point get + descending order point get
   sql: |
     create table t1 (a int, b int, c int);
     create index idx1 on t1(a asc, b);
     select * from t1 where a = 1 order by b limit 10;
+  expected_outputs:
+    - batch_plan
+- name: topn index selection + limit pushdown with multi point get + various order combination point get
+  sql: |
+    create table t1 (a int, b int, c int);
+    create index idx1 on t1(a desc nulls first, c desc nulls last);
+    create index idx2 on t1(a desc nulls first, c asc nulls last, b desc nulls first);
+    create index idx3 on t1(a desc nulls last, c asc nulls first, b desc nulls last);
+    select * from t1 where c = 2 and a = 2 order by b desc nulls last limit 10;
   expected_outputs:
     - batch_plan
--- a/src/frontend/planner_test/tests/testdata/output/index_selection.yaml
+++ b/src/frontend/planner_test/tests/testdata/output/index_selection.yaml
@@ -845,7 +845,7 @@
       └─BatchExchange { order: [], dist: Single }
         └─BatchLimit { limit: 10, offset: 0 }
           └─BatchScan { table: idx3, columns: [idx3.a, idx3.b, idx3.c, idx3.d], scan_ranges: [idx3.a = Int32(2) AND idx3.c = Int32(1) AND idx3.d = Int32(3)], limit: 10, distribution: UpstreamHashShard(idx3.a) }
-- name: topn index selection + limit pushdown with point get
+- name: topn index selection + limit pushdown with point get + descending order point get
   sql: |
     create table t1 (a int, b int, c int);
     create index idx1 on t1(a asc, b);
@@ -856,3 +856,16 @@
       └─BatchExchange { order: [], dist: Single }
         └─BatchLimit { limit: 10, offset: 0 }
           └─BatchScan { table: idx1, columns: [idx1.a, idx1.b, idx1.c], scan_ranges: [idx1.a = Int32(1)], limit: 10, distribution: UpstreamHashShard(idx1.a) }
+- name: topn index selection + limit pushdown with multi point get + various order combination point get
+  sql: |
+    create table t1 (a int, b int, c int);
+    create index idx1 on t1(a desc nulls first, c desc nulls last);
+    create index idx2 on t1(a desc nulls first, c asc nulls last, b desc nulls first);
+    create index idx3 on t1(a desc nulls last, c asc nulls first, b desc nulls last);
+    select * from t1 where c = 2 and a = 2 order by b desc nulls last limit 10;
+  batch_plan: |-
+    BatchSort { order: [idx3.b DESC NULLS LAST] }
+    └─BatchTopN { order: [idx3.a DESC NULLS LAST, idx3.c ASC NULLS FIRST, idx3.b DESC NULLS LAST], limit: 10, offset: 0 }
+      └─BatchExchange { order: [], dist: Single }
+        └─BatchLimit { limit: 10, offset: 0 }
+          └─BatchScan { table: idx3, columns: [idx3.a, idx3.b, idx3.c], scan_ranges: [idx3.a = Int32(2) AND idx3.c = Int32(2)], limit: 10, distribution: UpstreamHashShard(idx3.a) }


wenym1:
Thanks for your contribution.

I am quite confused about the motivation and the query plan described in the PR description. The leaf `BatchScan` has a `scan_ranges: [a = Int32(10)]` to limit that all output rows have the same value `a`, and therefore adding `t_index_ab.a DESC` in the `order` of subsequent `BatchTopN` seems unnecessary to me. 

Could you provide a concrete example on the incorrectness? The example may create a table and an index on it, insert some data, and run an actual query, and then present the actual and expected results.

chagelo:
@kwannoel PTAL. 

@wenym1 The key is the bottom `BatchScan`, there is a `limit: 1` so that the entire SQL only needs to scan one row.

wenym1:
> @kwannoel PTAL.
> 
> @wenym1 The key is the bottom `BatchScan`, there is a `limit: 1` so that the entire SQL only needs to scan one row.

Oh, I got your point. Previously the `TopNOnIndexRule` does not apply due to the issue in the PR description, and the `BatchScan` on index is generated later in the simple `IndexSelectionRule`. The difference is that the `BatchTopN` generated by `TopNOnIndexRule` has a prefix that , so that its limit can be later pushed down to the `BatchScan`.

Theoretical, if we can have a new optimizer rule, we can push down the limit in `BatchTopN` to the `BatchScan`, when seeing that (eq scan range in `BatchScan` + order of `BatchTopN`) is the prefix of (pk of the table of `BatchScan`), such as for the plan below
```
...
   └─BatchTopN { order: [t_index_ab.b ASC], limit: 1, offset: 0 }
     └─BatchScan { table: t_index_ab, columns: [a, b, c], scan_ranges: [a = Int32(10)] }
```

src/frontend/src/optimizer/rule/top_n_on_index_rule.rs
```diff
@@ -59,9 +59,17 @@ impl TopNOnIndexRule {
...
+                vec![
+                    ColumnOrder {
+                        column_index: input_ref.index,
+                        order_type: OrderType::ascending(),
+                    },
+                    ColumnOrder {
+                        column_index: input_ref.index,
+                        order_type: OrderType::descending(),
+                    },
+                ]
```
wenym1:
```suggestion
            .flat_map(|input_ref| {
                [
                    ColumnOrder {
                        column_index: input_ref.index,
                        order_type: OrderType::ascending(),
                    },
                    ColumnOrder {
                        column_index: input_ref.index,
                        order_type: OrderType::descending(),
                    },
                ]
```

Besides, `OrderType` has 4 variants, (ASC + DESC) * (nulls smallest + largest). For correctness, we may have to include all the 4 variants.

wenym1 (approved):
Rest LGTM. Thanks for the fix!

chagelo:
> > @kwannoel PTAL.
> > @wenym1 The key is the bottom `BatchScan`, there is a `limit: 1` so that the entire SQL only needs to scan one row.
> 
> Oh, I got your point. Previously the `TopNOnIndexRule` does not apply due to the issue in the PR description, and the `BatchScan` on index is generated later in the simple `IndexSelectionRule`. The difference is that the `BatchTopN` generated by `TopNOnIndexRule` has a prefix that , so that its limit can be later pushed down to the `BatchScan`.
> 
> Theoretical, if we can have a new optimizer rule, we can push down the limit in `BatchTopN` to the `BatchScan`, when seeing that (eq scan range in `BatchScan` + order of `BatchTopN`) is the prefix of (pk of the table of `BatchScan`), such as for the plan below
> 
> ```
> ...
>    └─BatchTopN { order: [t_index_ab.b ASC], limit: 1, offset: 0 }
>      └─BatchScan { table: t_index_ab, columns: [a, b, c], scan_ranges: [a = Int32(10)] }
> ```

It is exactly what `TopNOnIndexRule` has done.

chagelo:
> > @kwannoel PTAL.
> > @wenym1 The key is the bottom `BatchScan`, there is a `limit: 1` so that the entire SQL only needs to scan one row.
> 
> Oh, I got your point. Previously the `TopNOnIndexRule` does not apply due to the issue in the PR description, and the `BatchScan` on index is generated later in the simple `IndexSelectionRule`. The difference is that the `BatchTopN` generated by `TopNOnIndexRule` has a prefix that , so that its limit can be later pushed down to the `BatchScan`.
> 
> Theoretical, if we can have a new optimizer rule, we can push down the limit in `BatchTopN` to the `BatchScan`, when seeing that (eq scan range in `BatchScan` + order of `BatchTopN`) is the prefix of (pk of the table of `BatchScan`), such as for the plan below
> 
> ```
> ...
>    └─BatchTopN { order: [t_index_ab.b ASC], limit: 1, offset: 0 }
>      └─BatchScan { table: t_index_ab, columns: [a, b, c], scan_ranges: [a = Int32(10)] }
> ```

It is exactly what TopNOnIndexRule has done.

wenym1:
> > > @kwannoel PTAL.
> > > @wenym1 The key is the bottom `BatchScan`, there is a `limit: 1` so that the entire SQL only needs to scan one row.
> > 
> > 
> > Oh, I got your point. Previously the `TopNOnIndexRule` does not apply due to the issue in the PR description, and the `BatchScan` on index is generated later in the simple `IndexSelectionRule`. The difference is that the `BatchTopN` generated by `TopNOnIndexRule` has a prefix that , so that its limit can be later pushed down to the `BatchScan`.
> > Theoretical, if we can have a new optimizer rule, we can push down the limit in `BatchTopN` to the `BatchScan`, when seeing that (eq scan range in `BatchScan` + order of `BatchTopN`) is the prefix of (pk of the table of `BatchScan`), such as for the plan below
> > ```
> > ...
> >    └─BatchTopN { order: [t_index_ab.b ASC], limit: 1, offset: 0 }
> >      └─BatchScan { table: t_index_ab, columns: [a, b, c], scan_ranges: [a = Int32(10)] }
> > ```
> 
> It is exactly what TopNOnIndexRule has done.

I see. But I think the optimizer is not generating the best query plan. Instead of 
```
BatchSort { order: [t_index_ab.b ASC] }
 └─BatchTopN { order: [t_index_ab.a DESC, t_index_ab.b ASC], limit: 1, offset: 0 }
   └─BatchExchange { order: [], dist: Single }
     └─BatchLimit { limit: 1, offset: 0 }
       └─BatchScan { table: t_index_ab, columns: [a, b, c], scan_ranges: [a = Int32(10)], limit: 1 }
```
a better one would be like
```
BatchProject { exprs: [Int32(10) as a, t_index_ab.b, t_index_ab.c] }
└─BatchLimit { limit: 1, offset: 0 }
   └─BatchExchange { order: [t_index_ab.b], dist: Single }
     └─BatchScan { table: t_index_ab, columns: [b, c], scan_ranges: [a = Int32(10)], limit: 1 }
```

The optimization trace (enabled by running `explain(trace) <QUERY>`) shall be like
0. starting from
```
LogicalTopN { order: [t.c ASC], limit: 1, offset: 0 }
 └─LogicalScan { table: t, columns: [t.a, t.b, t.c], predicate: (t.a = 10:Int32) }
```
1. apply TopNOnIndexRule
```
LogicalProject { exprs: [10:Int32 as a, t_index_ab.b, t_index_ab.c] }
  └─LogicalTopN { order: [t_index_ab.b ASC], limit: 1, offset: 0 }
     └─LogicalScan { table: t_index_ab, columns: [t_index_ab.b, t_index_ab.c], predicate: (t_index_ab.a = 10:Int32) }
```
the key difference is that
* we don't include the `a` from the output of `LogicalScan`, and instead we set `a` as a const value via a later `LogicalProject`.
* the `order` of `LogicalTopN` includes only `b`, instead of `[a, b]`. The current `[a, b]` has prevented us from doing many later optimization.
2. turn logical plan into a same batch plan
```
BatchProject { exprs: [10:Int32 as a, t_index_ab.b, t_index_ab.c] }
  └─BatchTopN { order: [t_index_ab.b ASC], limit: 1, offset: 0 }
     └─BatchScan { table: t_index_ab, columns: [t_index_ab.b, t_index_ab.c], predicate: (t_index_ab.a = 10:Int32) }
```
the upper most `BatchSort` is not needed, because the `BatchTopN` is already ordered by `b`.
3. turn into physical plan for each plan node
```
BatchProject { exprs: [Int32(10) as a, t_index_ab.b, t_index_ab.c] }
└─BatchLimit { limit: 1, offset: 0 }
   └─BatchExchange { order: [t_index_ab.b], dist: Single }
     └─BatchLimit { limit: 1, offset: 0 }
       └─BatchScan { table: t_index_ab, columns: [b, c], scan_ranges: [a = Int32(10)]}
```
the key point here is that
* for `BatchScan`, when some of the eq predicates is the prefix (a = Int32(10)) of table pk (a, b), and the prefix of output columns matches the subsequent columns (b) after the eq predicates in pk, the `BatchScan` should have order as the prefix of output column (b)
* when turning `BatchTopN` into two phase physical plan, when seeing the order of input matches the order of itself, not only turning the first phase as `BatchLimit`, but also the second phase.
4. push down limit to `BatchScan`
```
BatchProject { exprs: [Int32(10) as a, t_index_ab.b, t_index_ab.c] }
└─BatchLimit { limit: 1, offset: 0 }
   └─BatchExchange { order: [t_index_ab.b], dist: Single }
     └─BatchScan { table: t_index_ab, columns: [b, c], scan_ranges: [a = Int32(10)], limit: 1 }
```

I will merge this PR first. This refinement can be done in a separate PR. I will create an issue on it.

PR #22563, opened 2025-07-10
feat: support ALTER CONNECTION CONNECTOR WITH

tabVersion:

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/meta/service/src/stream_service.rs
+++ b/src/meta/service/src/stream_service.rs
@@ -26,7 +26,9 @@ use risingwave_meta::manager::MetadataManager;
 use risingwave_meta::model::ActorId;
 use risingwave_meta::stream::{SourceManagerRunningInfo, ThrottleConfig};
 use risingwave_meta::{MetaError, model};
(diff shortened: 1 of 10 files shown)


src/meta/service/src/stream_service.rs
```diff
@@ -585,18 +587,68 @@ impl StreamManagerService for StreamServiceImpl {
...
+                let (options, secret_refs) = connection_options_with_secret.into_parts();
+                let new_props_plaintext = secret_manager
+                    .fill_secrets(options, secret_refs)
+                    .map_err(MetaError::from)?
+                    .into_iter()
+                    .collect::<HashMap<String, String>>();
+
+                // Prepare runtime mutation for all dependent sources and sinks with their complete properties
+                let mut dependent_mutation = HashMap::default();
+                for (source_id, complete_source_props) in updated_sources_with_props {
```
zwang28, line 718: Is TABLE using connection included here?

zwang28 (approved):

src/meta/service/src/stream_service.rs
tabVersion: Yes, table with connector will create a source accordingly.

PR #22571, opened 2025-07-11
feat(iceberg): support manual compaction of iceberg engine table

chenzl25:

- Support manual compaction of the iceberg engine table. Reusing the syntax `VECUUM`.
- Currently, our iceberg compaction task has no ack response, so we will loop the iceberg metadata to wait for a compaction snapshot being produced.

## Checklist

## Documentation

<details>

</details>


src/frontend/src/handler/vacuum.rs
```diff
@@ -0,0 +1,64 @@
...
+use crate::error::{ErrorCode, Result, RwError};
+use crate::handler::{HandlerArgs, RwPgResponse};
+
+pub async fn handle_vacuum(
+    handler_args: HandlerArgs,
+    table_name: ObjectName,
+) -> Result<RwPgResponse> {
+    let session = &handler_args.session;
+    let db_name = &session.database();
+    let (schema_name, table_name) = Binder::resolve_schema_qualified_name(db_name, table_name)?;
```
xiangjinwu: nit: can move this line inside block below to hide the unresolved `schema_name`.

src/frontend/src/handler/vacuum.rs
```diff
@@ -0,0 +1,64 @@
...
+        if table.engine() != Engine::Iceberg {
+            return Err(ErrorCode::InvalidInputSyntax(format!(
+                "VACUUM can only be used on Iceberg engine tables, but table '{}' uses {:?} engine",
+                table_name,
+                table.engine()
+            ))
+            .into());
+        }
+
+        TableId::new(table.id().table_id)
```
xiangjinwu: What's the point of this? Why not just `table.id()`?

src/frontend/src/handler/vacuum.rs
```diff
@@ -0,0 +1,64 @@
...
+    };
+
+    session.env().meta_client().compact_table(table_id).await?;
+
+    Ok(PgResponse::builder(StatementType::VACUUM).row_cnt(1).into())
```
xiangjinwu: Why 1? Looks like [PostgreSQL](https://github.com/postgres/postgres/blob/REL_17_5/src/include/tcop/cmdtaglist.h#L219) does not include rowcount for `VACUUM`.

xiangjinwu (approved):
The sqlparser and frontend part generally looks good me.

As suggested by the automated message, let's include a sqlparser test case in the yaml file for this new statement. Although it is pretty simple right now, the test case can enforce future editors to update it according and capture potential bugs when it gets complicated.

src/frontend/src/handler/vacuum.rs
chenzl25: Good point. I didn't notice it before.

PR #22573, opened 2025-07-11
fix: inaccurate join barrier processing time and description

lmatz:

New one
SCR-20250711-mmp

It has nothing to do with "per second".
And it used to be a percentage with %, which makes no sense 

## Checklist

## Documentation

<details>

</details>


grafana/dashboard/dev/streaming_actors.py
```diff
@@ -246,8 +246,8 @@ def _(outer_panels: Panels):
                         ),
                     ],
                 ),
-                panels.timeseries_percentage(
-                    "Join Actor Match Duration Per Second",
+                panels.timeseries_actor_latency(
+                    "Join Actor Match Duration Per Chunk/Barrier",
```
kwannoel, line 258: Why per chunk/barrier. Shouldn't it just be per barrier? 🤔

lmatz: There are two types of metrics in this panel:
1. how much time does it take to process a chunk
2. how much time does it take to process a barrier

PR #22578, opened 2025-07-11
feat(sink): support snowflake sink upsert

xxhZs:

We support snowflake sink
We will create a task to run merge into in snowflake with jdbc
And we will drop it when drop sink.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/java/connector-node/risingwave-jdbc-runner/pom.xml
+++ b/java/connector-node/risingwave-jdbc-runner/pom.xml
@@ -31,16 +31,18 @@
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-text</artifactId>
         </dependency>
-        <dependency>
-            <groupId>org.apache.logging.log4j</groupId>
-            <artifactId>log4j-slf4j2-impl</artifactId>
-        </dependency>
 
         <!-- https://mvnrepository.com/artifact/net.snowflake/snowflake-jdbc -->
         <dependency>
             <groupId>net.snowflake</groupId>
             <artifactId>snowflake-jdbc</artifactId>
             <version>3.23.0</version>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-slf4j2-impl</artifactId>
+            <scope>test</scope>
         </dependency>
     </dependencies>
 </project>


src/connector/src/sink/file_sink/snowflake.rs
hzxa21, line 1: empty file?

src/connector/src/sink/snowflake/mod.rs
```diff
@@ -0,0 +1,526 @@
...
+pub const DEFAULT_SCHEDULE: &str = "1 HOUR";
+
+#[serde_as]
+#[derive(Debug, Clone, Deserialize, WithOptions)]
+pub struct SnowflakeConfig {
+    #[serde(flatten)]
+    pub s3_inner: S3Config,
+
+    #[serde(rename = "snowflake.cdc_table_name")]
+    pub snowflake_cdc_table_name: Option<String>,
```
hzxa21, line 68: This is an internal table user doesn't need to be aware. I think we don't need to expose this option to user.

src/connector/src/sink/snowflake/snowflake_jni_client.rs
```diff
@@ -0,0 +1,305 @@
...
+
+fn build_start_task_sql(snowflake_task_context: &SnowflakeTaskContext) -> String {
+    let SnowflakeTaskContext {
+        task_name,
+        database,
+        schema,
+        ..
+    } = snowflake_task_context;
+    let full_task_name = format!("{}.{}.{}", database, schema, task_name);
+    format!("ALTER TASK {} RESUME", full_task_name)
```
hzxa21, line 117: Why do we need this?

src/meta/src/manager/sink_coordination/handle.rs
```diff
@@ -132,9 +132,9 @@ impl SinkWriterCoordinationHandle {
                             prev_epoch
                         )));
                     }
-                    if request.metadata.is_none() {
-                        return Poll::Ready(Err(anyhow!("empty commit metadata")));
-                    };
+                    // if request.metadata.is_none() {
+                    //     return Poll::Ready(Err(anyhow!("empty commit metadata")));
+                    // };
```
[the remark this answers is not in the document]
hzxa21: Why do we need this?

hzxa21 (commented):
Is it possible for us to create the two tables (intermediate changelog table and target table)?

src/connector/src/sink/snowflake/mod.rs
xxhZs: We need to expose this because the management of permissions for this table involves S3. Therefore, for the time being, users will have to create the table themselves and then send it to us.

src/connector/src/sink/snowflake/snowflake_jni_client.rs
xxhZs: The test results indicate that both the database  and schema parameters are required; otherwise, Snowflake will fail to recognize the object.

hzxa21: I mean why do we need to do `alter task resume`

Issue #22570, opened 2025-07-11
improving optimizer TopNOnIndexRule


wenym1:
> > > @wenym1 The key is the bottom `BatchScan`, there is a `limit: 1` so that the entire SQL only needs to scan one row.
> > 
> > 
> > Oh, I got your point. Previously the `TopNOnIndexRule` does not apply due to the issue in the PR description, and the `BatchScan` on index is generated later in the simple `IndexSelectionRule`. The difference is that the `BatchTopN` generated by `TopNOnIndexRule` has a prefix that , so that its limit can be later pushed down to the `BatchScan`.
> > Theoretical, if we can have a new optimizer rule, we can push down the limit in `BatchTopN` to the `BatchScan`, when seeing that (eq scan range in `BatchScan` + order of `BatchTopN`) is the prefix of (pk of the table of `BatchScan`), such as for the plan below
> > ```
> > ...
> >    └─BatchTopN { order: [t_index_ab.b ASC], limit: 1, offset: 0 }
> >      └─BatchScan { table: t_index_ab, columns: [a, b, c], scan_ranges: [a = Int32(10)] }
> > ```
> 
> It is exactly what TopNOnIndexRule has done.

I see. But I think the optimizer is not generating the best query plan. Instead of 
```
BatchSort { order: [t_index_ab.b ASC] }
 └─BatchTopN { order: [t_index_ab.a DESC, t_index_ab.b ASC], limit: 1, offset: 0 }
   └─BatchExchange { order: [], dist: Single }
     └─BatchLimit { limit: 1, offset: 0 }
       └─BatchScan { table: t_index_ab, columns: [a, b, c], scan_ranges: [a = Int32(10)], limit: 1 }
```
a better one would be like
```
BatchProject { exprs: [Int32(10) as a, t_index_ab.b, t_index_ab.c] }
└─BatchLimit { limit: 1, offset: 0 }
   └─BatchExchange { order: [t_index_ab.b], dist: Single }
     └─BatchScan { table: t_index_ab, columns: [b, c], scan_ranges: [a = Int32(10)], limit: 1 }
```

The optimization trace (enabled by running `explain(trace) <QUERY>`) shall be like
0. starting from
```
LogicalTopN { order: [t.c ASC], limit: 1, offset: 0 }
 └─LogicalScan { table: t, columns: [t.a, t.b, t.c], predicate: (t.a = 10:Int32) }
```
1. apply TopNOnIndexRule
```
[...]

PR #22581, opened 2025-07-13
fix(optimization): simplify topn order after TopNOnIndexRule optimization

chagelo:

close https://github.com/risingwavelabs/risingwave/issues/22570
## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/planner_test/tests/testdata/output/index_selection.yaml
+++ b/src/frontend/planner_test/tests/testdata/output/index_selection.yaml
@@ -803,34 +803,31 @@
     create index idx1 on t1(a, b desc);
     select * from t1 where a = 1 order by b desc limit 10
   batch_plan: |-
-    BatchSort { order: [idx1.b DESC] }
-    └─BatchTopN { order: [idx1.a ASC, idx1.b DESC], limit: 10, offset: 0 }
-      └─BatchExchange { order: [], dist: Single }
-        └─BatchLimit { limit: 10, offset: 0 }
-          └─BatchScan { table: idx1, columns: [idx1.a, idx1.b], scan_ranges: [idx1.a = Int32(1)], limit: 10, distribution: UpstreamHashShard(idx1.a) }
+    BatchTopN { order: [idx1.b DESC], limit: 10, offset: 0 }
+    └─BatchExchange { order: [], dist: Single }
+      └─BatchLimit { limit: 10, offset: 0 }
+        └─BatchScan { table: idx1, columns: [idx1.a, idx1.b], scan_ranges: [idx1.a = Int32(1)], limit: 10, distribution: UpstreamHashShard(idx1.a) }
 - name: topn index selection + limit pushdown with multi point get
   sql: |
     create table t1 (a int, b int, c int);
     create index idx1 on t1(a, c, b desc);
     select * from t1 where c = 1 and a = 2 order by b desc limit 10
   batch_plan: |-
-    BatchSort { order: [idx1.b DESC] }
-    └─BatchTopN { order: [idx1.a ASC, idx1.c ASC, idx1.b DESC], limit: 10, offset: 0 }
-      └─BatchExchange { order: [], dist: Single }
-        └─BatchLimit { limit: 10, offset: 0 }
-          └─BatchScan { table: idx1, columns: [idx1.a, idx1.b, idx1.c], scan_ranges: [idx1.a = Int32(2) AND idx1.c = Int32(1)], limit: 10, distribution: UpstreamHashShard(idx1.a) }
+    BatchTopN { order: [idx1.b DESC], limit: 10, offset: 0 }
+    └─BatchExchange { order: [], dist: Single }
+      └─BatchLimit { limit: 10, offset: 0 }
+        └─BatchScan { table: idx1, columns: [idx1.a, idx1.b, idx1.c], scan_ranges: [idx1.a = Int32(2) AND idx1.c = Int32(1)], limit: 10, distribution: UpstreamHashShard(idx1.a) }
 - name: topn index selection + limit pushdown with multi point get + unindexed point get
   sql: |
     create table t1 (a int, b int, c int, d int);
     create index idx1 on t1(a, c, b desc);
     select * from t1 where c = 1 and a = 2 and d = 3 order by b desc limit 10
   batch_plan: |-
-    BatchSort { order: [idx1.b DESC] }
-    └─BatchTopN { order: [idx1.a ASC, idx1.c ASC, idx1.b DESC], limit: 10, offset: 0 }
-      └─BatchExchange { order: [], dist: Single }
-        └─BatchLimit { limit: 10, offset: 0 }
-          └─BatchFilter { predicate: (idx1.d = 3:Int32) }
-            └─BatchScan { table: idx1, columns: [idx1.a, idx1.b, idx1.c, idx1.d], scan_ranges: [idx1.a = Int32(2) AND idx1.c = Int32(1)], distribution: UpstreamHashShard(idx1.a) }
+    BatchTopN { order: [idx1.b DESC], limit: 10, offset: 0 }
+    └─BatchExchange { order: [], dist: Single }
+      └─BatchLimit { limit: 10, offset: 0 }
+        └─BatchFilter { predicate: (idx1.d = 3:Int32) }
+          └─BatchScan { table: idx1, columns: [idx1.a, idx1.b, idx1.c, idx1.d], scan_ranges: [idx1.a = Int32(2) AND idx1.c = Int32(1)], distribution: UpstreamHashShard(idx1.a) }
 - name: topn index selection + limit pushdown with multi index
   sql: |
     create table t1 (a int, b int, c int, d int);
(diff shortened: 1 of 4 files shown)


chagelo:
@wenym1 PTAL

src/frontend/src/optimizer/plan_node/generic/top_n.rs
```diff
@@ -32,6 +32,7 @@ pub struct TopN<PlanRef> {
     pub limit_attr: TopNLimit,
     pub offset: u64,
     pub order: Order,
+    pub full_order: Order,
```
wenym1, line 35: Doesn't look like a good idea to introduce another concept `full_order` simply for this issue. Besides, `order` and `full_order` does not cover all possible orders when we have more than 2 index columns. For example, when we create index on `(a, b, c)`, and the query have fixed prefix `(a=xxx, b=xxx)`, the order can be any of `[a, c]`, `[b, c]`, `[a, b, c]` and `[c]`.

IIUC, the reason for introducing it is to satisfy `input.order().satisfies(&self.core.order)`. Instead of applying a different order to satisfy, it's more straight forward to make `input.order()` to be more precise. The `input` in the current issue is `LogicalScan { table: t_index_ab, columns: [t_index_ab.a, t_index_ab.b, t_index_ab.c], predicate: (t_index_ab.a = 10:Int32) }`, and its current derived order is `[t_index_ab.a, t_index_ab.b]`. Our desired order is `[t_index_ab.b]`, and it's actually satisfied, because the prefix `t_index_ab.a` is a const value. However, as described above, we may have combinational number of orders in the plan node, and it can be ambiguous to choose any of the orders as the derived order. To solve this ambiguity, As mentioned in the issue, 
> for BatchScan, when some of the eq predicates is the prefix (a = Int32(10)) of table pk (a, b), and the prefix of output columns matches the subsequent columns (b) after the eq predicates in pk, the BatchScan should have order as the prefix of output column (b)

we can simply not include the const column in the output of `LogicalScan`, and then the `LogicalScan` can have a unique order. The const column can be added later via a `LogicalProject`. Pulling up const column from `LogicalScan` can happen either in the `TopNIndexRule`, or even in the earlier `PredicatePushdownRule`.

src/frontend/planner_test/tests/testdata/output/index_selection.yaml
```diff
@@ -840,22 +837,20 @@
...
     create table t1 (a int, b int, c int);
     create index idx1 on t1(a asc, b);
     select * from t1 where a = 1 order by b limit 10;
   batch_plan: |-
-    BatchSort { order: [idx1.b ASC] }
-    └─BatchTopN { order: [idx1.a ASC, idx1.b ASC], limit: 10, offset: 0 }
-      └─BatchExchange { order: [], dist: Single }
-        └─BatchLimit { limit: 10, offset: 0 }
-          └─BatchScan { table: idx1, columns: [idx1.a, idx1.b, idx1.c], scan_ranges: [idx1.a = Int32(1)], limit: 10, distribution: UpstreamHashShard(idx1.a) }
+    BatchTopN { order: [idx1.b ASC], limit: 10, offset: 0 }
```
wenym1, line 850: The input from `BatchScan` is already sorted in the required order, and the `BatchTopN` is yet another unnecessary `BatchSort`. We can actually follow what mentioned in the issue to replace the `BatchTopN` here with `BatchLimit`.
> when turning BatchTopN into two phase physical plan, when seeing the order of input matches the order of itself, not only turning the first phase as BatchLimit, but also the second phase.

In `impl ToDistributedBatch for BatchTopN` and `impl ToLocalBatch for BatchTopN`, we didn't implement the method `to_xxx_with_order_required` specially. We can instead change to implement the `to_xxx_with_order_required`, and when seeing the required order is the same as the order of `BatchTopN`, we can turn the second phase into `BatchLimit`.

src/frontend/src/optimizer/plan_node/generic/top_n.rs
chagelo: > Doesn't look like a good idea to introduce another concept `full_order` simply for this issue. Besides, `order` and `full_order` does not cover all possible orders when we have more than 2 index columns. For example, when we create index on `(a, b, c)`, and the query have fixed prefix `(a=xxx, b=xxx)`, the order can be any of `[a, c]`, `[b, c]`, `[a, b, c]` and `[c]`.

@wenym1 

1. Emm, in previous implementation, for `where a = 10 and b = 10 order by [...,]c`, `Explain(trace)` shows that the field `order` of TopN is always `[c]` whatever `order by` is, after `TopNOnIndexRule`, this field order of TopN(also `self.core.order` in `input.order().satisfies(&self.core.order)`) becomes `[b, c]` from `[c]` as [LogicalTopN.clone_with_input_and_prefix](https://github.com/risingwavelabs/risingwave/blob/be92d020c837f0b71ed79ffb8c7f8fe833e1c08e/src/frontend/src/optimizer/plan_node/logical_topn.rs#L179) shows, so that final BatchTopN has order `order: [t_index_bc.b ASC, t_index_bc.c ASC]`.  So previous I want to simplify this `core.order`.

>However, as described above, we may have combinational number of orders in the plan node, and it can be ambiguous to choose any of the orders as the derived order.

Here you mentioned combinational number of orders, I don't understand this.

2. The `input.order()` is obtained from `TableScan.pk` as the [BatchSeqScan::new_inner](https://github.com/risingwavelabs/risingwave/blob/51bcef0563a841993d17ed8affaac4c3ec4ad8e2/src/frontend/src/optimizer/plan_node/batch_seq_scan.rs#L52) shows.  When I printed it, I found the pk was `[$0 ASC, $1 ASC, $3 ASC]`. This seems to correspond to columns `b, c`, and `_row_id` rather than a single column `_row_id`. This pk confusing me.
3. Now I am trying "we can simply not include the const column in the output of LogicalScan" you mentioned above, but seems to have a bug.

image

BTW, how to debug this generally? 😂

wenym1: > Here you mentioned combinational number of orders, I don't understand this.

For example, when we have a table `t (a int, b int, c int)` with pk as `[a, b, c]`, and we have a `LogicalScan` on it as `LogicalScan { table: t, columns: [t.a, t.b, t.c], predicate: (t.a = 10:Int32, t.b = 10:Int32) }`. For such a plan node, it can satisfies at least 4 orders, `[a, b, c]`, `[b, c]`, `[a, c]`, `[c]`, because the pk prefixes `(a=10, b=10)` are const. 4, which is 2*2, is the number of different combinations of whether or not `a` or `b` is present in the order, and therefore I called it combinational number. 

However, we can only choose one of them as the derived order to return on `fn order(&self) -> Order`. Currently, we seem to choose the full pk `[a, b, c]` as the derived order. And then when its parent plan node (can be any plan node other than LogicalTopN), which takes the `LogicalScan` as input, is checking `input.order().satisfy([a, c])`, it's actually checking `[a, b, c].satisfy([a, c])`, and the check will return false. But the `LogicalScan` can actually satisfy the order of `[a, c]`, which brings in some limitation in the optimizer. 

To summarize, the `LogicalScan` satisfies many different orders, but we only choose one of them as the derived order to return, and then the optimizer may fail to acknowledge that it satisfies an order, and then fail to generate the best plan.

wenym1: > BTW, how to debug this generally? 😂

How did you launch RW? Via `risedev`?

Usually you can find the stack trace of the panic in the log of frontend node. If you launched with `risedev`, the log is in `.risingwave/log/frontend.log`.

If your IDE is powerful enough, you can launch RW from your IDE by running the main function in `src/cmd_all/src/bin/risingwave.rs`, and then the debugger of your IDE can attach breakpoint and hook to help you debug more efficiently.

wenym1: > BTW, how to debug this generally? 😂

How did you launch RW? Via `risedev`?

Usually you can find the stack trace of the panic in the log of frontend node. If you launched with `risedev`, the log is in `.risingwave/log/frontend.log`.

If your IDE is powerful enough, you can launch RW from your IDE by running the main function in `src/cmd_all/src/bin/risingwave.rs`, and then the debugger of your IDE can attach breakpoint and hook to help you debug more efficiently.

wenym1: > The input.order() is obtained from TableScan.pk as the [BatchSeqScan::new_inner](https://github.com/risingwavelabs/risingwave/blob/51bcef0563a841993d17ed8affaac4c3ec4ad8e2/src/frontend/src/optimizer/plan_node/batch_seq_scan.rs#L52) shows. When I printed it, I found the pk was [$0 ASC, $1 ASC, $3 ASC]. This seems to correspond to columns b, c, and _row_id rather than a single column _row_id. This pk confusing me.

`[b, c]` is the index column, and `_row_id` is the pk of the original table. The pk of the index table is `[b, c, _row_id]`, so that we can find the `_row_id` when providing with `[b, c]`. With the pk `_row_id` of the original table, we can further get other columns of the row from the original table by looking up the pk, if the other columns are not included in the index.

chagelo: Ok, thanks sincerely!

wenym1:
@chagelo Kindly remind, is there any update on this PR?

chagelo:
> @chagelo Kindly remind, is there any update on this PR?

Sorry, no. I tried but failed and didn't know how to finish this.

I wiil follow up this and see how other contributors implement it.

PR #22585, opened 2025-07-14
fix(cdc): fix the handling of default values and enum type during auto schema change

wcy-fdu:

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/java/connector-node/risingwave-source-cdc/src/main/java/io/debezium/connector/postgresql/connection/pgoutput/PgOutputMessageDecoder.java
+++ b/java/connector-node/risingwave-source-cdc/src/main/java/io/debezium/connector/postgresql/connection/pgoutput/PgOutputMessageDecoder.java
@@ -363,6 +363,10 @@ public class PgOutputMessageDecoder extends AbstractMessageDecoder {
         columnDefaults =
                 readColumns.stream()
                         .filter(io.debezium.relational.Column::hasDefaultValue)
+                        .filter(
+                                column ->
+                                        isConstantDefaultValue(
+                                                column.defaultValueExpression().orElse(null)))
                         .collect(
                                 toMap(
                                         io.debezium.relational.Column::name,
@@ -1046,5 +1050,97 @@ public class PgOutputMessageDecoder extends AbstractMessageDecoder {
         if (connection != null) {
             connection.close();
         }
+    }
+
+    /**
+     * Checks if a given default value expression is a constant value. Constant values include:
+     * string literals, numeric literals, boolean literals, NULL, and type-cast literals. Function
+     * expressions (e.g., nextval(), now(), current_timestamp) are not considered constants.
+     *
+     * @param defaultValueExpression The default value expression to check
+     * @return true if the expression is a constant value, false otherwise
+     */
+    private boolean isConstantDefaultValue(String defaultValueExpression) {
+        if (defaultValueExpression == null || defaultValueExpression.trim().isEmpty()) {
+            return false;
+        }
+
+        String expr = defaultValueExpression.trim();
+
+        // Common function patterns that should be excluded
+        if (containsFunctionCall(expr)) {
+            LOGGER.trace("Skipping function expression: {}", expr);
+            return false;
+        }
+
+        // Check for constant patterns
+        return isLiteralConstant(expr);
+    }
+
+    /** Checks if the expression contains function call patterns. */
+    private boolean containsFunctionCall(String expr) {
+        // Common PostgreSQL function patterns
+        String[] functionPatterns = {
+            "nextval(",
+            "currval(",
+            "setval(",
+            "now()",
+            "current_timestamp",
+            "current_date",
+            "current_time",
+            "localtime",
+            "localtimestamp",
+            "gen_random_uuid()",
+            "random()",
+            "clock_timestamp()",
+            "statement_timestamp()",
+            "transaction_timestamp()",
+            "uuid_generate",
+            "extract(",
+            "date_part(",
+            "age(",
+            "justify_"
+        };
+
+        String lowerExpr = expr.toLowerCase();
+        for (String pattern : functionPatterns) {
+            if (lowerExpr.contains(pattern.toLowerCase())) {
+                return true;
+            }
+        }
+
+        // General pattern: word followed by parentheses (function call)
+        return expr.matches(".*\\w+\\s*\\(.*\\).*");
+    }
+
+    /** Checks if the expression is a literal constant. */
+    private boolean isLiteralConstant(String expr) {
+        // String literal: 'value' or 'value'::type
+        if (expr.matches("'[^']*'(::.*)?")) {
+            return true;
+        }
+
+        // Numeric literal: 123, 3.14, -42, 1.23e-4
+        if (expr.matches("-?\\d+(\\.\\d+)?([eE][+-]?\\d+)?")) {
+            return true;
+        }
+
+        // Boolean literal
+        if (expr.toLowerCase().matches("(true|false)(::.*)?")) {
+            return true;
+        }
+
+        // NULL literal
+        if (expr.toLowerCase().matches("null(::.*)?")) {
+            return true;
+        }
+
+        // Array literal: '{1,2,3}' or ARRAY['a','b','c']
+        if (expr.matches("\\{.*\\}(::.*)?")
+                || expr.toLowerCase().matches("array\\s*\\[.*\\](::.*)?")) {
+            return true;
+        }
+
+        return false;
     }
 }
--- a/java/connector-node/risingwave-source-cdc/src/main/java/io/debezium/connector/postgresql/connection/pgoutput/PgOutputMessageDecoder.java
+++ b/java/connector-node/risingwave-source-cdc/src/main/java/io/debezium/connector/postgresql/connection/pgoutput/PgOutputMessageDecoder.java
@@ -808,6 +808,28 @@ public class PgOutputMessageDecoder extends AbstractMessageDecoder {
                 editor.defaultValueExpression(columnMetadata.getDefaultValueExpression());
             }
 
+            // Check if this column is an enum type and set enum values
+            if (isEnumType(columnMetadata.getPostgresType())) {
+                try {
+                    List<String> enumValues =
+                            queryEnumValues(columnMetadata.getPostgresType().getName());
+                    if (!enumValues.isEmpty()) {
+                        editor.enumValues(enumValues);
+                        LOGGER.trace(
+                                "Found enum values for column '{}' of type '{}': {}",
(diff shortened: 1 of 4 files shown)


java/connector-node/risingwave-source-cdc/src/main/java/io/debezium/connector/postgresql/connection/pgoutput/PgOutputMessageDecoder.java
```diff
@@ -363,6 +363,10 @@ private Table handleRelationMessage(ByteBuffer buffer, TypeRegistry typeRegistry
         columnDefaults =
                 readColumns.stream()
                         .filter(io.debezium.relational.Column::hasDefaultValue)
+                        .filter(
```
hzxa21: Please wrap you changes in `/* patch code */`. Same for other changes in this files.

java/connector-node/risingwave-source-cdc/src/main/java/io/debezium/connector/postgresql/connection/pgoutput/PgOutputMessageDecoder.java
```diff
@@ -808,6 +812,28 @@ private Table resolveRelationFromMetadata(PgOutputRelationMetaData metadata) {
...
+                                enumValues);
+                    }
+                } catch (SQLException e) {
+                    LOGGER.warn(
+                            "Failed to query enum values for column '{}' of type '{}': {}",
+                            columnMetadata.getColumnName(),
+                            columnMetadata.getPostgresType().getName(),
+                            e.getMessage());
+                }
+            }
```
hzxa21: We don't need to implement our own logic to fetch the enum values from pg because the `PostgresType` itself already do that for us during construction. You can directly call [getEnumValues](https://github.com/debezium/debezium/blob/4159397bd40dd5d0c8efe2b97ba61ba8eb220b58/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresType.java#L127) in `PostgresType`.

More details on how the enum values are fetched:
1. [typeRegistry.get](https://github.com/debezium/debezium/blob/4159397bd40dd5d0c8efe2b97ba61ba8eb220b58/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/pgoutput/PgOutputMessageDecoder.java#L327) is called to construct `PostgresType`.
2. [`resolveUnkonwnType`](https://github.com/debezium/debezium/blob/4159397bd40dd5d0c8efe2b97ba61ba8eb220b58/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/TypeRegistry.java#L179) is called if the type oid is not found.
3. A [SQL](https://github.com/debezium/debezium/blob/4159397bd40dd5d0c8efe2b97ba61ba8eb220b58/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/TypeRegistry.java#L77) (similar to what you have written in this PR) will be issued to pg to fetch the enum values.

java/connector-node/risingwave-source-cdc/src/main/java/io/debezium/connector/postgresql/connection/pgoutput/PgOutputMessageDecoder.java
```diff
@@ -363,6 +363,10 @@ private Table handleRelationMessage(ByteBuffer buffer, TypeRegistry typeRegistry
         columnDefaults =
                 readColumns.stream()
                         .filter(io.debezium.relational.Column::hasDefaultValue)
+                        .filter(
+                                column ->
+                                        isConstantDefaultValue(
```
hzxa21: Filtering out non-const default here can be dangerous because it also affects the debezium internal [decoder state](https://github.com/debezium/debezium/blob/4159397bd40dd5d0c8efe2b97ba61ba8eb220b58/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/pgoutput/PgOutputMessageDecoder.java#L363). Given that it is RW that doesn't care about the non-const default value rater than debezium, it is better we do it in the RW's code instead of inside debezium.

hzxa21: I think it is better to handle it here: https://github.com/risingwavelabs/risingwave/blob/22fc3da474de324e5a1f25e86dfd839426e23f14/src/connector/src/parser/unified/debezium.rs#L250

java/connector-node/risingwave-source-cdc/src/main/java/io/debezium/connector/postgresql/connection/pgoutput/PgOutputMessageDecoder.java
```diff
@@ -808,6 +808,21 @@ private Table resolveRelationFromMetadata(PgOutputRelationMetaData metadata) {
                 editor.defaultValueExpression(columnMetadata.getDefaultValueExpression());
             }
 
+            /* patch code */
+            // Check if this column is an enum type and set enum values
+            if (columnMetadata.getPostgresType().getEnumValues() != null) {
+                List<String> enumValues = columnMetadata.getPostgresType().getEnumValues();
+                if (!enumValues.isEmpty()) {
+                    editor.enumValues(enumValues);
```
xiangjinwu, line 824: Caveat:
* The actual list of values seem currently unused. Only whether the field is set to non-null value.
* If some logic uses these values in the future, it can be inaccurate - when upstream PostgreSQL uses `ALTER TYPE` that adds a new variant to this enum, the list here won't be updated to keep in sync

Maybe we shall add this comment to the code here or the usage below

src/connector/src/parser/unified/debezium.rs
```diff
@@ -32,8 +32,80 @@ use crate::source::cdc::build_cdc_table_id;
...
+        return true;
+    }
+
+    // 3. Check if it contains operators
+    if regex::Regex::new(r"[+\-*/<>=!&|]").unwrap().is_match(expr) {
+        return false;
+    }
+
+    // 4. Other cases might be constants
+    true
```
xiangjinwu: This sounds as bad as `isDataTypeCompatible` defaulting to `true`

src/connector/src/parser/unified/debezium.rs
```diff
@@ -32,8 +32,80 @@ use crate::source::cdc::build_cdc_table_id;
...
+        return false;
+    }
+
+    // 2. Check if it's a literal constant
+    if is_literal_constant(expr) {
+        return true;
+    }
+
+    // 3. Check if it contains operators
+    if regex::Regex::new(r"[+\-*/<>=!&|]").unwrap().is_match(expr) {
```
xiangjinwu: Why bother implementing a problematic parsing again? For example `2 ^ 3` is not covered.

src/connector/src/parser/unified/debezium.rs
```diff
@@ -32,8 +32,80 @@ use crate::source::cdc::build_cdc_table_id;
...
+
+/// Checks if the expression is a literal constant.
+fn is_literal_constant(expr: &str) -> bool {
+    // String literal: 'value' or 'value'::type
+    if regex::Regex::new(r"'[^']*'(::.*)?").unwrap().is_match(expr) {
+        return true;
+    }
+
+    // Numeric literal: 123, 3.14, -42, 1.23e-4
+    if regex::Regex::new(r"-?\d+(\.\d+)?([eE][+-]?\d+)?")
```
xiangjinwu: Avoid re-implementing these if possible. For example PostgreSQL also accepts `0x1a` or `0o23` (although they are stored in `information_schema.columns` as decimal)

src/connector/src/parser/unified/debezium.rs
```diff
@@ -32,8 +32,80 @@ use crate::source::cdc::build_cdc_table_id;
...
+    if regex::Regex::new(r"null(::.*)?")
+        .unwrap()
+        .is_match(&expr.to_lowercase())
+    {
+        return true;
+    }
+
+    // Array literal: '{1,2,3}' or ARRAY['a','b','c']
+    if regex::Regex::new(r"\{.*\}(::.*)?").unwrap().is_match(expr)
+        || regex::Regex::new(r"array\s*\[.*\](::.*)?")
```
xiangjinwu: `array[2, 3 + 4]` is not a constant. Let's keep it simple and keep supporting only `'xxx'::ty`

src/connector/src/parser/unified/debezium.rs
```diff
@@ -32,8 +32,39 @@ use crate::source::cdc::build_cdc_table_id;
...
+
+    let expr = default_value_expression.trim();
+
+    // Special handling for nextval() function - skip parsing for existing columns
+    if expr.starts_with("nextval(") {
+        return false;
+    }
+
+    // Otherwise, assume it's a constant value and continue to handle the default value as before.
+    true
```
xiangjinwu: Given the updated function name, we can just return `expr.starts_with("nextval(")` at the end, right?

src/connector/src/parser/unified/debezium.rs
```diff
@@ -232,72 +276,79 @@ pub fn parse_schema_change(
...
-                                        .split("::")
-                                        .map(|s| s.trim_matches('\''))
-                                        .next()
-                                    {
-                                        None => {
-                                            value_text = None;
+                            // Only process constant default values
+                            if !is_bigserial_default(default_val_expr_str) {
+                                tracing::warn!(target: "auto_schema_change",
+                                    "Skip unsupported expression: {}", default_val_expr_str);
```
xiangjinwu: We can be more specific here, like "ignoring unsupported bigserial default"

src/connector/src/source/cdc/external/postgres.rs
```diff
@@ -711,6 +711,56 @@ fn is_supported_even_split_data_type(data_type: &DataType) -> bool {
...
+        PgType::NUMERIC => DataType::Decimal,
+        PgType::DATE => DataType::Date,
+        PgType::TIME => DataType::Time,
+        PgType::TIMESTAMP => DataType::Timestamp,
+        PgType::TIMESTAMPTZ => DataType::Timestamptz,
+        PgType::INTERVAL => DataType::Interval,
+        PgType::VARCHAR | PgType::TEXT | PgType::BPCHAR => DataType::Varchar,
+        PgType::BYTEA => DataType::Bytea,
+        PgType::JSON | PgType::JSONB => DataType::Jsonb,
+        _ => {
```
xiangjinwu, line 757: UUID / POINT / more from #22200?

wcy-fdu: should be included in https://github.com/risingwavelabs/risingwave/pull/22610 if needed, this PR only solves `enum` and `bigserial`

PR #22595, opened 2025-07-14
refactor(compute): Combined the same logic of Merge and Union

Chiro11:

Previously, the `Merge` and `Union` operators both contained the same logic: collecting barriers and watermarks from multiple upstreams, and aligning them.
Now, we extract this logic and implement it as a generic class, so that different logics can be used. In the future, the `UnionUpstreamSink` operator will also use this class.

## Checklist

## Documentation

<details>

</details>


Chiro11:
* **#22614** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22614?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#22595** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22595?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22595?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/stream/src/executor/exchange/input.rs
```diff
@@ -26,34 +26,51 @@ use super::permit::Receiver;
...
+    DispatcherMessageStreamItem, MessageStreamItemInner,
 };
-use crate::task::{FragmentId, LocalBarrierManager, UpDownActorIds, UpDownFragmentIds};
+use crate::task::{FragmentId, InputId, LocalBarrierManager, UpDownActorIds, UpDownFragmentIds};
 
-/// `Input` provides an interface for [`MergeExecutor`](crate::executor::MergeExecutor) and
+/// `Input` is a more abstract upstream input type, used for `DynamicReceivers` type
+/// handling of multiple upstream inputs
+pub trait Input: Stream + Send {
+    /// The upstream input id.
```
wenym1, line 37: Instead of a unified `InputId` type, we can have a associated type named like `InputIdType` in the trait definition to support different id type in different implementations.

src/stream/src/executor/exchange/input.rs
```diff
@@ -26,34 +26,51 @@ use super::permit::Receiver;
...
+/// `ActorInput` provides an interface for [`MergeExecutor`](crate::executor::MergeExecutor) and
 /// [`ReceiverExecutor`](crate::executor::ReceiverExecutor) to receive data from upstream actors.
-pub trait Input: DispatcherMessageStream {
+/// Only used for actor inputs.
+pub trait ActorInput: Input<Item = DispatcherMessageStreamItem> {
     /// The upstream actor id.
     fn actor_id(&self) -> ActorId;
 
-    fn boxed_input(self) -> BoxedInput
+    fn boxed_input(self) -> BoxedActorInput
```
wenym1: This method seems to be a general method of `Input`, and can be moved there.

src/stream/src/executor/exchange/input.rs
```diff
@@ -26,34 +26,51 @@ use super::permit::Receiver;
...
+pub type BoxedInput<Message> = Pin<Box<dyn Input<Item = Message>>>;
+pub type BoxedMessageInput<M> = BoxedInput<MessageStreamItemInner<M>>;
+
+/// `ActorInput` provides an interface for [`MergeExecutor`](crate::executor::MergeExecutor) and
 /// [`ReceiverExecutor`](crate::executor::ReceiverExecutor) to receive data from upstream actors.
-pub trait Input: DispatcherMessageStream {
+/// Only used for actor inputs.
+pub trait ActorInput: Input<Item = DispatcherMessageStreamItem> {
     /// The upstream actor id.
     fn actor_id(&self) -> ActorId;
```
wenym1: Can we also refactor to use `Input::id()` instead of having a separate `actor_id()`?

src/stream/src/executor/merge.rs
```diff
@@ -121,6 +119,9 @@ pub struct MergeExecutor {
     /// Upstream channels.
     upstreams: SelectReceivers,
 
+    /// Upstream actor ids.
```
wenym1: This field is stored specially to replace original call on `select_all.upstream_actor_ids()`. Can we instead provide a similar method `upstream_ids()` in `DynamicReceivers` so that we don't have to maintain a separate set of `upstream_actor_ids`?

src/stream/src/executor/exchange/input.rs
wenym1: Or we can actually unify `ActorInput` and `Input`. 
```
trait ActorInput = Input<InputId = ActorId, Item = DispatcherMessageStreamItem>;
```

src/stream/src/executor/merge.rs
```diff
@@ -331,29 +341,23 @@ impl MergeExecutor {
                             // Add the new upstreams to select.
                             select_all.add_upstreams_from(select_new);
 
-                            // Add buffers to the buffered watermarks for all cols
-                            select_all
```
wenym1, line 335: Do we forget to update the `buffered_watermarks` in the refactored code?

src/stream/src/executor/merge.rs
```diff
@@ -385,201 +387,6 @@ impl Execute for MergeExecutor {
     }
 }
 
-/// A stream for merging messages from multiple upstreams.
```
wenym1, line 388: Could you keep the code in the original place for easier review? You can reorganize the code after approval and before merge this PR.

src/stream/src/executor/merge.rs
Chiro11: This is unified into `add_upstreams_from` function.

src/stream/src/executor/merge.rs
Chiro11: Ok.

src/stream/src/executor/merge.rs
```diff
@@ -385,29 +353,37 @@ impl Execute for MergeExecutor {
...
-pub struct SelectReceivers {
+/// Can dynamically add and delete upstream streams.
+/// For the meaning of the generic parameter `M` used, refer to `BarrierInner<M>`.
+pub struct DynamicReceivers<InputId, M> {
     /// The barrier we're aligning to. If this is `None`, then `blocked_upstreams` is empty.
-    barrier: Option<DispatcherBarrier>,
+    barrier: Option<BarrierInner<M>>,
     /// The upstreams that're blocked by the `barrier`.
-    blocked: Vec<BoxedInput>,
+    blocked: Vec<BoxedMessageInput<InputId, M>>,
```
wenym1: Can try making `upstream_input_ids` as `fn upstream_input_ids(&self) -> impl Iterator<Item = InputId> + '_`? It can be implemented as `blocked.iter().map(...).chain(active.iter().map(...))`, and then we don't have to maintain a separate `upstream_input_ids`.

src/stream/src/executor/union.rs
```diff
@@ -53,125 +50,53 @@ impl UnionExecutor {
...
+        union_receivers.boxed()
     }
 }
 
 #[pin_project]
-struct Input {
+struct UnionExecutorInput {
+    id: usize,
     #[pin]
     inner: BoxedMessageStream,
```
wenym1, line 80: `BoxedMessageStream` is already `Unpin`. No need as extra `pin_project`.

PR #22609, opened 2025-07-15
feat(sqlsmith): integrate reducer into sqlsmith

cardigan1008:

This PR integrates a **fine-grained** reducer into SQLSmith's error triaging pipeline. It complements the existing statement-level reducer by further minimizing the last (failing) SQL statement in the input sequence while preserving its erroneous behavior (e.g., catalog errors, panics, execution crashes).

### Key changes:
- After the statement-level reducer simplifies the input, the last SQL statement is passed to a new AST-based reducer.
- This reducer applies a sequence of transformation passes (e.g., removing clauses, pulling up expressions) to shrink the query.
- The output combines:
  - setup statements (unchanged),
  - and the final reduced query (minimized but still triggers the same failure).

### Example
Original input:

```sql
-- ERROR: Failed to run the query
-- Caused by: table or source not found: t1

SELECT 1, 2 FROM alltypes1 
UNION 
SELECT 2000, CASE WHEN true THEN 'ghijkl' ELSE 'mnopqr' END FROM t1 WHERE 1 = 1 GROUP BY 1, 2 ORDER BY 2 ASC;
```

Reduced output:

```sql
SELECT NULL FROM t1;
```

### Usage

```sh
cargo build
./riscdev d
./target/debug/sqlsmith-reducer -i input.sql -o reduced.sql
```

## Checklist

## Documentation

<details>

</details>


src/tests/sqlsmith/src/sqlreduce/reducer.rs
```diff
@@ -0,0 +1,199 @@
...
+    ///
+    /// The process continues until a global fixed point is reached (i.e., no transformation
+    /// makes progress on any part of the SQL).
+    ///
+    /// # Arguments
+    /// - `sql`: The SQL string (usually the failing query) to reduce.
+    ///
+    /// # Returns
+    /// - A reduced SQL string (still failing) that is minimized w.r.t the current passes.
+    pub async fn reduce_until_fixed_point(&self, sql: &str) -> String {
```
kwannoel: To my understanding, rather than exploring the entire shrinkable search space like [`quickcheck::shrink`](https://hackage-content.haskell.org/package/QuickCheck-2.16.0.0/docs/Test-QuickCheck-Arbitrary.html#v:shrink), we choose a greedy approach here, and assume we will come up with a good enough reduction path.

Assumption looks good to me, we can try it out in practice and see if it holds. If not we can change the interface to produce a lazy-stream of shrinked SQLs to try.

kwannoel: Nit: Does this function need to be `pub`? It's only called by `reduce` in this module.

cardigan1008: > To my understanding, rather than exploring the entire shrinkable search space like [`quickcheck::shrink`](https://hackage-content.haskell.org/package/QuickCheck-2.16.0.0/docs/Test-QuickCheck-Arbitrary.html#v:shrink), we choose a greedy approach here, and assume we will come up with a good enough reduction path.

Adopting the QuickCheck-style exhaustive enumeration is also a viable option. While it may introduce an O(N²) complexity, it's likely acceptable in practice, especially considering that the pull-up step can significantly reduce the overall reduction time. Initially, I didn’t see a major difference between the two strategies — but I agree it might be worth revisiting.

src/tests/sqlsmith/src/sqlreduce/checker.rs
```diff
@@ -0,0 +1,40 @@
...
+
+        old_result.0 == new_result.0 && old_result.1 == new_result.1
+    }
+}
+
+pub async fn run_query(client: &Client, query: &str) -> (bool, String) {
+    let query_task = client.simple_query(query);
+    match query_task.await {
+        Ok(_) => (true, String::new()),
+        Err(e) => (false, e.to_string()),
```
kwannoel: Wonder if these can be simplified into:
```
        let old_result = run_query(self.client, old).await.map_err(|e| e.to_string());
        let new_result = run_query(self.client, new).await.map_err(|e| e.to_string());
        old_result == new_result
```

src/tests/sqlsmith/src/sqlreduce/checker.rs
```diff
@@ -0,0 +1,40 @@
...
+        let new_result = run_query(self.client, new).await;
+
+        old_result.0 == new_result.0 && old_result.1 == new_result.1
+    }
+}
+
+pub async fn run_query(client: &Client, query: &str) -> (bool, String) {
+    let query_task = client.simple_query(query);
+    match query_task.await {
+        Ok(_) => (true, String::new()),
```
kwannoel, line 102: What if the query that suceeds changes the state of the database? For instance if it's a streaming query, it will be persisted. Do we have any cleanup step for shrinking validation?

src/tests/sqlsmith/src/bin/reducer.rs
```diff
@@ -26,9 +39,50 @@ struct Args {
...
+    #[arg(short, long, default_value = "single")]
+    strategy: ReductionStrategy,
+
+    /// For consecutive strategy, number of elements to reduce at once (used only when strategy = consecutive)
+    #[arg(short, long, default_value_t = 2)]
+    consecutive_k: usize,
 }
 
-fn main() {
+#[tokio::main(flavor = "multi_thread", worker_threads = 5)]
```
kwannoel, line 52: Something to note is that if the system _panics_ due to a query, we may need someway to restart it. I think we may encounter this when integrating the reducer into CI, and may need some mechanism to restart RisingWave from a clean-state if this happens.

Can address in separate PR when integrating into CI.

src/tests/sqlsmith/src/sqlreduce/passes/mod.rs
```diff
@@ -0,0 +1,79 @@
...
+pub enum Strategy {
+    Single,             // one-at-a-time
+    Aggressive,         // all combinations
+    Consecutive(usize), // delete k consecutive elements
+}
+
+pub trait Transform: Send + Sync {
+    fn name(&self) -> String;
+
+    fn get_reduction_points(&self, ast: Ast) -> Vec<usize>;
```
kwannoel, line 64: Let's add some docs for reduction points, like some examples. To my understanding it's a way to _identify_ potential reductions, e.g. some select item, an operator, etc...

src/tests/sqlsmith/src/sqlreduce/checker.rs
cardigan1008: You're right — that's indeed a concern. Note that this issue also applies to the preceding_stmts in `reduce()` we execute (e.g., CREATE TABLE), which already modify the database state before the final query is run. Do you have any idea on how to avoid this? Maybe follow what simulation has done?

src/tests/sqlsmith/src/sqlreduce/passes/pullup.rs
```diff
@@ -0,0 +1,293 @@
...
+                }
+            }
+        }
+        ast.clone()
+    }
+}
+
+/// Simplify array constructors in SELECT projections by pulling up the first element.
+///
+/// This transformation replaces an `ARRAY[...]` expression with its first element.
```
kwannoel, line 187: Not sure how effective this will be, since it changes the type of the select item, invalidating references which depend on the array type. Do you mean to do `SELECT ARRAY[1]` instead in the example?

cardigan1008: Yes, a mistake here. It should be 

```rust
if let Expr::Array { elem, named } = expr {
    if let Some(first_elem) = elem.first() {
        *expr = Expr::Array {
            elem: vec![first_elem.clone()],
            named: *named,
        };
    }
}
```

src/tests/sqlsmith/src/sqlreduce/checker.rs
cardigan1008: Fixed by using schema as suggested.

cardigan1008:
Some update: 
1. In remove pass, we get the reversed reduction points now. For example, for `GROUP BY a, b, c`, the original result is `[0, 1, 2]`. This caused a panic in single mode when removing `a` and `b` first, followed by `rp.remove(2)`, which no longer existed. Reversing the order prevents this. 
2. Fix some errors like keeping the types consistent in row pullup. 
3. Add unit tests for each pass. Exclude the test for whole reducer for now.

kwannoel (approved):
LGTM, great work and efforts!

hzxa21 (approved):
Cargo.lock LGTM

PR #22610, opened 2025-07-15
fix(pg-cdc): completely fix the schema conversion problem during auto schema change

wcy-fdu:

Still draft, The behavior of some data types remains to be confirmed.

## Checklist

## Documentation

<details>

</details>


src/connector/src/source/cdc/external/postgres.rs
```diff
@@ -256,6 +257,126 @@ impl PostgresExternalTableReader {
...
+        PgType::JSONB_ARRAY => DataType::List(Box::new(DataType::Jsonb)),
+        PgType::UUID_ARRAY => DataType::List(Box::new(DataType::Varchar)),
+        PgType::OID => DataType::Int64,
+        PgType::OID_ARRAY => DataType::List(Box::new(DataType::Int64)),
+        PgType::POINT_ARRAY => DataType::List(Box::new(DataType::Struct(
+            risingwave_common::types::StructType::new(vec![
+                ("x", DataType::Float32),
+                ("y", DataType::Float32),
+            ]),
+        ))),
```
tabVersion: is it possible to expand with macro?

src/connector/src/source/cdc/external/postgres.rs
```diff
@@ -256,6 +257,126 @@ impl PostgresExternalTableReader {
     }
 }
 
+pub fn type_name_to_pg_type(ty_name: &str) -> Option<PgType> {
+    let ty_name_lower = ty_name.to_lowercase();
+
+    // Handle array types (prefixed with _)
+    if let Some(base_type) = ty_name_lower.strip_prefix('_') {
```
tabVersion, line 717: in which case will it enter the `else` branch? can you share a link

tabVersion (approved):
other parts LGTM

src/connector/src/source/cdc/external/postgres.rs
wcy-fdu: In the new added test, `int[]` type. All array type will start with `_`, like `float[]->_float`, `int[]->_int4`

src/connector/src/source/cdc/external/postgres.rs
xiangjinwu: https://github.com/risingwavelabs/risingwave/blob/main/src/common/src/types/postgres_type.rs
This file can be used as a reference. `for_all_base_types` provides the info and each calling site project it to different shapes:
* `type_len`: `impl_type_len`
* `from_oid`: `impl_from_oid`
* `to_oid`: `impl_to_oid`
* `pg_name`: `impl_pg_name`

wcy-fdu: I think using macros here is a bit of an over-optimization, and it doesn't seem to save much code.🤔

wcy-fdu: Merge it first, and then optimize the code if needed.

PR #22611, opened 2025-07-15
feat(sink): support SQL syntax to create sink with auto schema refresh from upstream table

wenym1:

new with options in create sink statement to specify auto schema change
```
create sink s from t with (connector = 'blackhole', auto.schema.change = 'true');
```

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/e2e_test/sink/auto_schema_change.slt
+++ b/e2e_test/sink/auto_schema_change.slt
@@ -1,3 +1,5 @@
+# test auto schema change on table without specified pk
+
 statement ok
 create table t(id int, name string);
 
@@ -10,7 +12,8 @@ create sink s from t with (connector = 'kafka', auto.schema.change = 'true');
 statement error auto schema change not supported for CREATE SINK AS QUERY
 create sink s as select * from t with (connector = 'blackhole', auto.schema.change = 'true');
 
-# error on auto schema change on MV
+## test error on auto schema change on MV
+
 statement ok
 create materialized view m1 as select * from t;
 
@@ -19,6 +22,10 @@ create sink s from m1 with (connector = 'blackhole', auto.schema.change = 'true'
 
 statement ok
 drop materialized view m1;
+
+## end
+
+## test error on generated columns and sink-into-table
 
 statement ok
 create table t1 (id int primary key, name string, g int  as id + 1);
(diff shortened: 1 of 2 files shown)


wenym1:
* **#22661** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22661?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#22660** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22660?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#22611** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22611?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22611?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

wenym1:
> In terms of the SQL syntax, there are two options:
> 
> 1. Use keyword. Example: `CREATE SINK ... FROM t AUTO REFRESH SCHEMA with (...)`
> 
> * Pros:
>   
>   * The semantic is clearer and more rigid
> * Cons:
>   
>   * Need to maintain additional field in proto and meta store table
>   * The behavior cannot change once the sink is created
> 
> 2. Use with option. Example: `CREATE SINK ... FROM t with (auto_refresh_schema = true, ...)`
> 
> * Pros:
>   
>   * More flexible
>   * Requires no changes in proto and meta store schema
> * Cons:
>   
>   * The sematic is implicit
> 
> I am more leaning towards option 2. Hope to hear @xiangjinwu's opinion as well because we once discussed the criteria of adding new SQL syntax.

Have changed to use syntax `CREATE SINK ... FROM t with (auto.refresh.schema = true, ...)`.

Besides, we have to add a new field in meta store table anyway for both syntaxes, because the with options only contains a boolean value, but we should store `Option<u32>`, where `u32` is the schema of the upstream table id.

src/frontend/src/handler/create_sink.rs
```diff
@@ -179,13 +203,48 @@ pub async fn gen_sink_plan(
...
 
-    let (dependent_relations, dependent_udfs, bound) = {
+    let (dependent_relations, dependent_udfs, bound, auto_refresh_schema_from_table) = {
         let mut binder = Binder::new_for_stream(session);
+        let auto_refresh_schema_from_table = if let Some((from_name, true)) = &direct_sink_from_name
+        {
+            let from_relation =
+                binder.bind_relation_by_name(from_name.clone(), None, None, true)?;
+            if let Relation::BaseTable(table) = from_relation {
+                if table.table_catalog.table_type != TableType::Table {
```
hzxa21, line 218: notes for future work: it should be fairly easy to support source as well, right?

PR #22614, opened 2025-07-15
feat(compute): Implement UnionUpstreamSink operator

Chiro11:

Currently, we have only implemented the `UpstreamSinkUnion` operator. In `create_nodes_inner` function, we replaced the previous `Merge` + `Project` operator in a way that is not completely reliable. Soon, once the meta's modification is completed, the hack here will no longer be needed.
In addition, the operator does not currently support dynamic increase and decrease of upstream, but will support it in the future.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/meta/src/rpc/ddl_controller.rs
+++ b/src/meta/src/rpc/ddl_controller.rs
@@ -1025,7 +1025,7 @@ impl DdlController {
                                             upstream_actor_id: vec![],
                                             upstream_fragment_id,
                                             upstream_dispatcher_type: PbDispatcherType::Hash as _,
-                                            fields: sink_fields.to_vec(),
+                                            fields: vec![],
                                         }
                                     };
 
--- a/src/meta/src/rpc/ddl_controller.rs
+++ b/src/meta/src/rpc/ddl_controller.rs
@@ -1025,7 +1025,7 @@ impl DdlController {
                                             upstream_actor_id: vec![],
                                             upstream_fragment_id,
                                             upstream_dispatcher_type: PbDispatcherType::Hash as _,
-                                            fields: vec![],
+                                            fields: sink_fields.to_vec(),
                                         }
                                     };
 
--- a/src/stream/src/executor/merge.rs
+++ b/src/stream/src/executor/merge.rs
@@ -41,7 +41,7 @@ pub(crate) enum MergeExecutorUpstream {
 pub(crate) struct MergeExecutorInput {
     upstream: MergeExecutorUpstream,
     actor_context: ActorContextRef,
-    pub(crate) upstream_fragment_id: UpstreamFragmentId,
+    upstream_fragment_id: UpstreamFragmentId,
     local_barrier_manager: LocalBarrierManager,
     executor_stats: Arc<StreamingMetrics>,
     pub(crate) info: ExecutorInfo,
@@ -226,7 +226,7 @@ impl MergeExecutor {
     }
 
     #[try_stream(ok = Message, error = StreamExecutorError)]
-    async fn execute_inner(mut self: Box<Self>) {
+    pub(crate) async fn execute_inner(mut self: Box<Self>) {
         let select_all = self.upstreams;
         let select_all = BufferChunks::new(select_all, self.chunk_size, self.schema);
         let actor_id = self.actor_context.id;
--- a/src/stream/src/executor/merge.rs
+++ b/src/stream/src/executor/merge.rs
@@ -41,7 +41,7 @@ pub(crate) enum MergeExecutorUpstream {
 pub(crate) struct MergeExecutorInput {
     upstream: MergeExecutorUpstream,
     actor_context: ActorContextRef,
-    upstream_fragment_id: UpstreamFragmentId,
+    pub(crate) upstream_fragment_id: UpstreamFragmentId,
     local_barrier_manager: LocalBarrierManager,
     executor_stats: Arc<StreamingMetrics>,
     pub(crate) info: ExecutorInfo,
@@ -171,6 +171,7 @@ impl MergeExecutor {
         inputs: Vec<super::exchange::permit::Receiver>,
         local_barrier_manager: crate::task::LocalBarrierManager,
         schema: Schema,
+        chunk_size: usize,
     ) -> Self {
         use super::exchange::input::LocalInput;
         use crate::executor::exchange::input::ActorInput;
@@ -197,7 +198,7 @@ impl MergeExecutor {
             local_barrier_manager,
             metrics.into(),
             barrier_rx,
-            100,
+            chunk_size,
             schema,
         )
     }
@@ -638,6 +639,7 @@ mod tests {
             rxs,
             barrier_test_env.local_barrier_manager.clone(),
             Schema::new(vec![]),
+            100,
         );
         let mut merger = merger.boxed().execute();
         for (idx, epoch) in epochs {
--- a/src/stream/src/executor/upstream_sink_union.rs
+++ b/src/stream/src/executor/upstream_sink_union.rs
@@ -27,6 +27,8 @@ use crate::task::{FragmentId, LocalBarrierManager};
 
 type ProcessedMessageStream = impl Stream<Item = MessageStreamItem>;
 
+/// A wrapper that merges data from a single upstream fragment and applies projection expressions.
+/// Each `SinkHandlerInput` represents one upstream fragment with its own merge executor and projection logic.
 #[pin_project]
 pub struct SinkHandlerInput {
     /// The ID of the upstream fragment that this input is associated with.
@@ -97,6 +99,7 @@ impl Stream for SinkHandlerInput {
     }
 }
 
+/// Information about an upstream fragment including its schema and projection expressions.
 #[derive(Debug)]
 pub struct UpstreamInfo {
     pub upstream_fragment_id: FragmentId,
@@ -106,6 +109,17 @@ pub struct UpstreamInfo {
 
 type BoxedSinkInput = BoxedMessageInput<FragmentId, BarrierMutationType>;
 
+/// `UpstreamSinkUnionExecutor` merges data from multiple upstream fragments, where each fragment
+/// has its own merge logic and projection expressions. This executor is specifically designed for
+/// sink operations that need to union data from different upstream sources.
+///
+/// Unlike a simple union that just merges streams, this executor:
+/// 1. Creates a separate `MergeExecutor` for each upstream fragment
+/// 2. Applies fragment-specific projection expressions to each stream
+/// 3. Unions all the processed streams into a single output stream
+///
+/// This is useful for sink operators that need to collect data from multiple upstream fragments
+/// with potentially different schemas or processing requirements.
 pub struct UpstreamSinkUnionExecutor {
     /// The context of the actor.
     actor_context: ActorContextRef,
--- a/src/stream/src/executor/upstream_sink_union.rs
+++ b/src/stream/src/executor/upstream_sink_union.rs
@@ -25,7 +25,6 @@ use crate::executor::project::apply_project_exprs;
 use crate::executor::{BarrierMutationType, BoxedMessageInput, DynamicReceivers, MergeExecutor};
 use crate::task::{FragmentId, LocalBarrierManager};
 
-type MergeStream = impl Stream<Item = MessageStreamItem>;
 type ProcessedMessageStream = impl Stream<Item = MessageStreamItem>;
 
 #[pin_project]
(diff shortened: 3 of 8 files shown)


Chiro11:
* **#22614** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22614?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22614?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#22595** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22595?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/stream/src/executor/upstream_sink_union.rs
```diff
@@ -0,0 +1,455 @@
...
+/// `MergeProjectExecutor` applies a projection to the output of a merge executor.
+struct MergeProjectExecutor {
+    /// Expressions of the current projection.
+    project_exprs: Vec<NonStrictExpression>,
+
+    /// The stream of messages after handled by merge executor.
+    merge: BoxedMessageStream,
+}
+
+impl MergeProjectExecutor {
```
wenym1: Seems unnecessary to introduce a `MergeProjectExecutor` type, since we always create it and then call `executor_inner` immediately. Can be simplified into a single function like
```
#[try_stream(ok = Message, error = StreamExecutorError)]
async fn apply_project_exprs_stream(stream: BoxedMessageStream, project_exprs: Vec<NonStrictExpression>,) {
    ...
}
```

src/stream/src/executor/upstream_sink_union.rs
```diff
@@ -0,0 +1,455 @@
...
+        }
+    }
+}
+
+pub struct SinkHandlerInput {
+    /// The ID of the upstream fragment that this input is associated with.
+    upstream_fragment_id: FragmentId,
+
+    /// The stream of messages from the upstream fragment.
+    executor_stream: BoxedMessageStream,
```
wenym1: The type of `executor_stream` can be static instead of dynamic boxed stream.

src/stream/src/executor/merge.rs
```diff
@@ -41,7 +41,7 @@ pub(crate) enum MergeExecutorUpstream {
 pub(crate) struct MergeExecutorInput {
     upstream: MergeExecutorUpstream,
     actor_context: ActorContextRef,
-    upstream_fragment_id: UpstreamFragmentId,
+    pub(crate) upstream_fragment_id: UpstreamFragmentId,
```
wenym1: Seems unnecessary to make it pub?

src/stream/src/task/actor_manager.rs
```diff
@@ -254,21 +256,59 @@ impl StreamActorManager {
...
-                    store.clone(),
-                    actor_context,
-                    vnode_bitmap.clone(),
-                    has_stateful || is_stateful,
-                    subtasks,
-                    local_barrier_manager,
-                )
-                .await?,
-            );
+            // Check if the input stream node is for sink-into operation.
```
wenym1: I suppose the starting point of pattern matching should be `NodeBody::Union` instead of `NodeBody::Project`? Also the check logic seems better to move to the beginning of `create_node_inner` just like the case of `SnapshotBackfill`, instead of hacking deep into the implementation of it.

src/stream/src/executor/upstream_sink_union.rs
```diff
@@ -0,0 +1,455 @@
...
+use crate::executor::{BarrierMutationType, BoxedMessageInput, DynamicReceivers, MergeExecutor};
+use crate::task::{FragmentId, LocalBarrierManager};
+
+/// `MergeProjectExecutor` applies a projection to the output of a merge executor.
+struct MergeProjectExecutor {
+    /// Expressions of the current projection.
+    project_exprs: Vec<NonStrictExpression>,
+
+    /// The stream of messages after handled by merge executor.
+    merge: BoxedMessageStream,
```
wenym1: The type of `merge` stream can also be static. It's created by `merge.execute()`, but actually `merge.execute()` is `merge.execute_inner().boxed()`, and `merge.execute_inner()` returns a stream with static type.

yezizp2012:
I will take a look tmr. 🫡

src/stream/src/task/actor_manager.rs
```diff
@@ -234,6 +378,22 @@ impl StreamActorManager {
             });
         }
 
+        if let NodeBody::Union(_) = node.get_node_body().unwrap() {
```
wenym1: I never meant doing the pattern matching only on `NodeBody::Union` and always hijack all the node creation logic under it. 😂

In general, I meant
```
async fn create_node_inner() {
    ...
    if let Some((dml_merge, project_merges, union)) = as_sink_into_table_union(node.get_node_body().unwrap()) {
        return self.create_sink_into_table_union(dml_merge, project_merges, union);
    }
    ...
}

```

where `as_sink_into_table_union` looks like
```
fn as_sink_into_table_union(node: &StreamNode) -> Option<
    (
        &MergeNode,
        Vec<(&ProjectNode, &MergeNode)>,
        &UnionNode
    )
> {
    if let NodeBody::Union(union) = node.get_node_body().unwrap() {
        for input in &node.inputs {
            // match the pattern of dml MergeNode or ProjectNode+MergeNode
            // if not matched in any way, simply return None so that the outer 
            // logic can fall back to normal logic of executor creation.
        }
        Some((dml_merge, project_merges, union))
    } else {
        None
    }
}
```

If we can assume that the dml MergeNode is always the first item in `inputs`, the pattern matching can be easier. We can check whether `inputs[0]` is `MergeNode` and `inputs[1..]` are all `ProjectNode`+`MergeNode` May need to check the code of sink-into-table to see how the input of Union is composed.

The `create_sink_into_table_union` would be as simple as like the following, and no need to call `create_node_inner` recursively again.
```
fn create_sink_into_table_union(
    &self,
    dml_merge: &MergeNode,
    project_merges: Vec<(&ProjectNode, &MergeNode)>,
    union: &UnionNode,
) -> Executor {
    UnionExecutor::new(
        MergeExecutor::new(dml_merge),
        UpstreamSinkExecutor::new(project_merges),
    )
}
```

src/stream/src/executor/upstream_sink_union.rs
```diff
@@ -0,0 +1,464 @@
...
+    ) -> Self {
+        let merge_stream = Self::generate_stream_from_merge(merge);
+        let processed_stream = Self::apply_project_exprs_stream(merge_stream, project_exprs);
+        Self {
+            upstream_fragment_id,
+            processed_stream,
+        }
+    }
+
+    #[define_opaque(MergeStream)]
```
wenym1: Seems unnecessary to define the `MergeStream` opaque type. The `apply_project_exprs_stream` is called immediately after `generate_stream_from_merge`, so we can replace the parameter `merge_stream: MergeStream` with `merge: Box<MergeExecutor>`, and do `merge.execute_inner()` in `apply_project_exprs_stream_impl`.

src/stream/src/task/actor_manager.rs
Chiro11: OK. 😂 I thought dml-merge was just an example. In fact, I thought tables may have more complicated cases of graph building or other fragment modification, so I tried to write the hijack here more generally. If dml-merge is the only case and always as the first input, then you're right, I will write it more specifically.

wenym1: > OK. 😂 I thought dml-merge was just an example. In fact, I thought tables may have more complicated cases of graph building or other fragment modification, so I tried to write the hijack here more generally. If dml-merge is the only case and always as the first input, then you're right, I will write it more specifically.

You can check the code of how the fragment stream node is composed when creating/dropping sink-into-table. The strategy should be stable, and fragment pattern shouldn't be complicated. 

Besides, in the future after we support the meta and frontend part of this feature, we will need to overwrite the fragments of existing tables for backward compatibility. We may match the same fragment pattern and do the replacement.

yezizp2012: @shanicky Can you help to double check the compose strategy for sink-into-table?

src/stream/src/executor/upstream_sink_union.rs
```diff
@@ -0,0 +1,464 @@
...
+#[derive(Debug)]
+pub struct UpstreamInfo {
+    pub upstream_fragment_id: FragmentId,
+    pub merge_schema: Schema,
+    pub project_exprs: Vec<NonStrictExpression>,
+}
+
+type BoxedSinkInput = BoxedMessageInput<FragmentId, BarrierMutationType>;
+
+pub struct UpstreamSinkUnionExecutor {
```
yezizp2012, line 123: Please add a doc comment like other executors for `UpstreamSinkUnionExecutor` to improve maintenance.

src/stream/src/task/actor_manager.rs
```diff
@@ -202,6 +210,164 @@ impl StreamActorManager {
...
+        let NodeBody::Union(_) = node.get_node_body().unwrap() else {
+            return None;
+        };
+
+        let mut merge_projects = Vec::new();
+
+        let mut rev_iter = node.get_input().iter().rev();
+        for union_input in rev_iter.by_ref() {
+            let NodeBody::Project(project) = union_input.get_node_body().unwrap() else {
+                break;
```
wenym1: When we break here, the current `union_input` has been consumed in the iterator. After `break`, it won't be in either `rev_iter` or `merge_projects`. Example code to reproduce.

```
let vec = vec![1, 2, 3, 4, 5];
let mut iter = vec.iter();
let mut first = vec![];
for i in iter.by_ref() {
    if *i == 3 {
        break;
    }
    first.push(*i);
}
let second: Vec<_> = iter.cloned().collect();
assert_eq!(vec![1, 2], first);
assert_eq!(vec![4, 5], second);
```

src/stream/src/task/actor_manager.rs
```diff
@@ -202,6 +210,164 @@ impl StreamActorManager {
...
+            if !project_check {
+                break;
+            }
+
+            let NodeBody::Merge(merge) = project_input.get_node_body().unwrap() else {
+                break;
+            };
+
+            #[allow(deprecated)]
+            let merge_check = merge.get_upstream_actor_id().is_empty()
```
wenym1: No need to check the deprecated field.

src/stream/src/executor/upstream_sink_union.rs
```diff
@@ -0,0 +1,473 @@
...
+            .with_guarded_label_values(&[
+                actor_id.to_string().as_str(),
+                fragment_id.to_string().as_str(),
+                "",
+                "UpstreamSinkUnion",
+            ]);
+
+        let upstreams = DynamicReceivers::new(inputs, Some(barrier_align), None);
+        pin_mut!(upstreams);
+
```
wenym1, line 316: Remove the comment?

src/stream/src/task/actor_manager.rs
Chiro11: This made me realize that this check did not work, and fixed some bugs.

src/meta/src/rpc/ddl_controller.rs
```diff
@@ -1021,7 +1021,7 @@ impl DdlController {
                                             upstream_actor_id: vec![],
                                             upstream_fragment_id,
                                             upstream_dispatcher_type: PbDispatcherType::Hash as _,
-                                            fields: sink_fields.to_vec(),
+                                            fields: vec![],
```
wenym1: Discussed offline. @Chiro11 mentions that the `fields` won't be used. Will have a PR later to mark this field as deprecated.

wenym1 (approved):
LGTM

src/stream/src/task/actor_manager.rs
```diff
@@ -202,6 +210,157 @@ impl StreamActorManager {
...
+                    vnode_bitmap.clone(),
+                    has_stateful,
+                    subtasks,
+                    local_barrier_manager,
+                )
+                .await?,
+            );
+        }
+
+        // Use the first MergeNode to fill in the info of the new node.
```
kwannoel, line 247: Why do we use `MergeNode` to fill in the info of the `UpstreamSinkUnion` node? Looks really weird to me.

kwannoel: Shouldn't we use the union node instead?

src/stream/src/executor/upstream_sink_union.rs
```diff
@@ -0,0 +1,471 @@
...
+            actor_context: actor_ctx,
+            local_barrier_manager,
+            executor_stats: metrics.into(),
+            chunk_size,
+            upstream_infos: vec![],
+        }
+    }
+
+    #[allow(dead_code)]
+    async fn new_sink_input(
```
kwannoel, line 198: How do we plan to use this?

src/stream/src/executor/upstream_sink_union.rs
```diff
@@ -0,0 +1,471 @@
...
+    }
+}
+
+impl UpstreamSinkUnionExecutor {
+    pub fn new(
+        ctx: ActorContextRef,
+        local_barrier_manager: LocalBarrierManager,
+        executor_stats: Arc<StreamingMetrics>,
+        chunk_size: usize,
+        upstream_infos: Vec<(FragmentId, Schema, Vec<NonStrictExpression>)>,
```
kwannoel, line 160: Why do we initialize inside the executor? The typical pattern is to initialize actors externally, then chain the executors together. So rather than `upstream_infos`, it should be `upstream_executors`.

Chiro11: In order to dynamically increase or decrease upstream, based on the `Update` info in Barrier. This part is still being implemented.

Issue #22491, opened 2025-07-03
sink-into-table refinement


wenym1:
### Background

Currently, the process of create/drop sink-into-table works as something like alter streaming job, where we have to  alter the fragment stream node of downstream table. In other words, the fragment metadata of the sink-into-table job and the downstream table are strongly coupled, and this makes the process create/drop sink-into-table complicated.

<img width="774" alt="Image" src="https://github.com/user-attachments/assets/b179cbdb-dbf5-4212-9625-37fd8b918849" />

Ideally, the metadata operation of create/drop sink-into-table should be as simple as creating/dropping a streaming job, and no metadata of other streaming jobs should be affected.

### Design

Current the streaming graph of sink-into-table looks like the following, where we explicitly store the sink upstreams in the persisted fragment stream node of the table.
![Image](https://github.com/user-attachments/assets/1bb2e1f7-d046-43ee-b864-4799b7993808)

In the proposed design, the streaming graph will look like the following
![Image](https://github.com/user-attachments/assets/dac1926b-1a5f-4288-8f3e-8cc8067e65ed)

No matter how many upstream sink-into-table are created and dropped, the persisted fragment stream node of the table is fixed. We will have a new stream node type named like `UpstreamSinkUnion`. It differs to the current simple `Union` in that
* it doesn't store its upstream in its persisted fragment stream node, and instead the upstreams are filled at runtime by global barrier worker during recovery.
* its executor supports dynamically adding and removing upstream inputs.

### composing upstreams of `UpstreamSinkUnion` during recovery

In global barrier worker, during recovery, we can restore the job and fragment information of all tables and sink-into-table sinks, and we are able to resolve the target table of each sink-into-table sink. Based on these information, we are able to compose the connections between sink-into-table sinks and downstream tables, and correctly fill in the upstream of `UpstreamSinkUnion`.

### create/drop sink-into-table on the flight

When handling create/drop sink-into-table ddl, on global barrier worker, we will generate a barrier with a new mutation. When the streaming executor of `UpstreamSinkUnion` receives this mutation, it will dynamically add a new upstream to streaming job, or remove an upstream.

### backward-compatibility to existing sink-into-table

on meta node bootstrap, we can modify persisted fragment stream node of the table of sink-into-table, from merge-project-union, to this `UpstreamSinkUnion`, before launching the global barrier worker.

BugenZhao:
[...]

PR #22616, opened 2025-07-15
feat: support recoverable background sink into table with progress tracking

yezizp2012:

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/e2e_test/background_ddl/sink.slt
+++ b/e2e_test/background_ddl/sink.slt
@@ -26,10 +26,7 @@ statement ok
 recover;
 
 onlyif can-use-recover
-sleep 5s
-
-onlyif can-use-recover
-statement count 1
+statement count 1 retry 3 backoff 2s
 show jobs;
 
 statement ok
@@ -52,7 +49,31 @@ drop sink sink2;
 statement count 0
 show jobs;
 
-# TODO: support backfill progress for background sink into table
+statement ok
+create table t1(v1 int primary key);
+
+statement ok
+create sink sink2 into t1 from t;
+
+statement count 1
+show jobs;
+
+onlyif can-use-recover
+statement ok
+recover;
+
+onlyif can-use-recover
+statement count 1 retry 3 backoff 2s
+show jobs;
+
+statement ok
+drop sink sink2;
+
+statement count 0
+show jobs;
+
+statement ok
+drop table t1;
 
 statement ok
 drop table t cascade;
--- a/src/frontend/src/handler/cancel_job.rs
+++ b/src/frontend/src/handler/cancel_job.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use pgwire::pg_response::{PgResponse, StatementType};
+use pgwire::pg_response::StatementType;
 use risingwave_common::types::Fields;
 use risingwave_pb::meta::cancel_creating_jobs_request::{CreatingJobIds, PbJobs};
 use risingwave_sqlparser::ast::JobIdents;
@@ -26,18 +26,51 @@ pub(super) async fn handle_cancel(
     jobs: JobIdents,
 ) -> Result<RwPgResponse> {
     let session = handler_args.session;
+    let job_ids = jobs.0;
+    let mut filtered_job_ids = vec![];
+    let mut notices = vec![];
+    {
+        let catalog_reader = session.env().catalog_reader().read_guard();
+        for job_id in job_ids {
+            let database_catalog = catalog_reader.get_database_by_name(&session.database())?;
+            let sink_catalog = database_catalog
+                .iter_schemas()
+                .find_map(|schema| schema.get_sink_by_id(&job_id));
+            if let Some(sink_catalog) = sink_catalog {
+                if sink_catalog.is_created() {
+                    continue; // Skip already created sinks
+                } else if sink_catalog.target_table.is_some() {
+                    notices.push(format!(
+                        "Please use `DROP SINK {}` to cancel sink into table job.",
+                        sink_catalog.name
+                    ));
+                    continue;
+                }
+            }
+            filtered_job_ids.push(job_id);
+        }
+    }
 
-    let canceled_jobs = session
-        .env()
-        .meta_client()
-        .cancel_creating_jobs(PbJobs::Ids(CreatingJobIds { job_ids: jobs.0 }))
-        .await?;
+    let mut response_builder = RwPgResponse::builder(StatementType::CANCEL_COMMAND);
+    for notice in notices {
+        response_builder = response_builder.notice(notice);
+    }
+
+    let canceled_jobs = if !filtered_job_ids.is_empty() {
+        session
+            .env()
+            .meta_client()
+            .cancel_creating_jobs(PbJobs::Ids(CreatingJobIds {
+                job_ids: filtered_job_ids,
+            }))
+            .await?
+    } else {
+        vec![]
+    };
     let rows = canceled_jobs
         .into_iter()
         .map(|id| CancelRow { id: id.to_string() });
-    Ok(PgResponse::builder(StatementType::CANCEL_COMMAND)
-        .rows(rows)
-        .into())
+    Ok(response_builder.rows(rows).into())
 }
 
 #[derive(Fields)]
--- a/src/frontend/src/handler/drop_sink.rs
+++ b/src/frontend/src/handler/drop_sink.rs
@@ -15,7 +15,6 @@
 use std::collections::HashSet;
 
 use pgwire::pg_response::{PgResponse, StatementType};
-use risingwave_common::catalog::StreamJobStatus;
 use risingwave_pb::ddl_service::{ReplaceJobPlan, replace_job_plan};
 use risingwave_pb::meta::cancel_creating_jobs_request::{CreatingJobIds, PbJobs};
 use risingwave_sqlparser::ast::ObjectName;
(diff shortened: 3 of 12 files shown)


src/meta/src/controller/streaming_job.rs
```diff
@@ -1238,10 +1224,11 @@ impl CatalogController {
                 }
 
                 if let Some(sink_id) = dropping_sink_id {
-                    let drained = incoming_sinks
+                    let _drained = incoming_sinks
                         .extract_if(.., |id| *id == sink_id)
                         .collect_vec();
-                    debug_assert_eq!(drained, vec![sink_id]);
+                    // TODO(august): re-enable this assertion after refactoring sink into table
+                    // debug_assert_eq!(drained, vec![sink_id]);
```
Chiro11, line 1231: Why can't we assert this?

yezizp2012: When the cancel command reaches the meta, it cannot determine whether FE has received the notify update of the table's `incoming_sink` field. To cancel SITT, we must replace the downstream table's streaming graph. The sole method involves utilizing the drop interface to abort the creation, necessitating some workaround solutions. Add these workarounds will be revert after you refactoring.

src/frontend/src/handler/cancel_job.rs
```diff
@@ -26,18 +26,51 @@ pub(super) async fn handle_cancel(
...
+            let database_catalog = catalog_reader.get_database_by_name(&session.database())?;
+            let sink_catalog = database_catalog
+                .iter_schemas()
+                .find_map(|schema| schema.get_sink_by_id(&job_id));
+            if let Some(sink_catalog) = sink_catalog {
+                if sink_catalog.is_created() {
+                    continue; // Skip already created sinks
+                } else if sink_catalog.target_table.is_some() {
+                    notices.push(format!(
+                        "Please use `DROP SINK {}` to cancel sink into table job.",
```
kwannoel, line 44: Will we continue to drop / cancel the sink job? If not, perhaps we can just return an error here.

src/meta/src/controller/catalog/drop_op.rs
```diff
@@ -198,9 +198,13 @@ impl CatalogController {
                 .count(&txn)
                 .await?;
             if creating != 0 {
-                return Err(MetaError::permission_denied(format!(
-                    "can not drop {creating} creating streaming job, please cancel them firstly"
-                )));
+                if creating == 1 && object_type == ObjectType::Sink {
```
kwannoel, line 201: Why we can only drop one sink job at a time?

kwannoel (approved):
Rest LGTM. @Chiro11 should also take a look, in case there's conflicts with the work of https://github.com/risingwavelabs/risingwave/issues/22491. Thanks @yezizp2012 for your efforts.

src/frontend/src/handler/cancel_job.rs
yezizp2012: Yes, the behavior of the `cancel` command is to accept multiple job parameters, and only the IDs that are successfully canceled will be returned. Therefore, I prefer not to return an error and instead add a notice, maintaining the behavior consistent with before.

e2e_test/batch/catalog/rw_depend.slt.part
```diff
@@ -27,7 +27,7 @@ create sink sink2 into t2 as select a from t1;
 
 # equivalent to:
 # select objid::regclass, refobjid::regclass from rw_depend;
-query TT rowsort
+query TT rowsort retry 3 backoff 1s
```
wenym1, line 30: I think any failure is not expected here, since the create sink into table statement has returned successfully?

e2e_test/background_ddl/sink.slt
```diff
@@ -55,7 +52,31 @@ drop sink sink2;
...
+
+statement count 1
+show jobs;
+
+onlyif can-use-recover
+statement ok
+recover;
+
+onlyif can-use-recover
+statement count 1 retry 5 backoff 2s
```
wenym1: The recovery command should wait for the finish of recovery. So I think we can expect that the `show jobs` should succeed immediately? ditto for the retry added in lines above.

src/meta/src/controller/catalog/drop_op.rs
yezizp2012: This is a workaround to allow dropping creating sink using `DROP SINK` command. I just want to limit the functionality to "DROP creating sink into table". Currently since "sink into table" requires replacing the downstream table, it's safer to reject "drop cascade" when there are jobs still being created.

src/frontend/src/handler/cancel_job.rs
```diff
@@ -26,18 +26,51 @@ pub(super) async fn handle_cancel(
...
+        for job_id in job_ids {
+            let database_catalog = catalog_reader.get_database_by_name(&session.database())?;
+            let sink_catalog = database_catalog
+                .iter_schemas()
+                .find_map(|schema| schema.get_sink_by_id(&job_id));
+            if let Some(sink_catalog) = sink_catalog {
+                if sink_catalog.is_created() {
+                    continue; // Skip already created sinks
+                } else if sink_catalog.target_table.is_some() {
+                    notices.push(format!(
```
wenym1, line 43: Currently, what's the difference between dropping and cancelling except for checking whether the job is already created? Do we have a plan to unify the handling logic of them in the future after the sink into table refactor?

src/meta/src/rpc/ddl_controller.rs
```diff
@@ -1360,10 +1360,7 @@ impl DdlController {
         // create streaming jobs.
         let stream_job_id = streaming_job.id();
         match (streaming_job.create_type(), &streaming_job) {
-            // TODO(August): Unify background sink into table's creation path with MV below.
```
wenym1, line 1363: We can avoid include `&streaming_job` in the match

src/meta/src/stream/stream_manager.rs
```diff
@@ -642,6 +647,23 @@ impl GlobalStreamManager {
...
+            if self
+                .creating_job_info
+                .check_job_exists(TableId::new(job_id as _))
+                .await
+            {
+                tracing::info!(
+                    ?job_id,
+                    "streaming job is creating, cancel it with drop directly"
+                );
+                self.metadata_manager
```
wenym1, line 661: Should we also remove the entry in `self.creating_job_info`?

src/meta/src/controller/catalog/mod.rs
```diff
@@ -1092,6 +1089,26 @@ impl CatalogControllerInner {
         }
     }
 
+    pub(crate) fn notify_cancelled(&mut self, database_id: Option<DatabaseId>, id: ObjectId) {
```
wenym1: Do we have any call with `database_id` as `None`? If not, we can change to always pass `database_id: DatabaseId`.

src/frontend/src/handler/drop_sink.rs
```diff
@@ -64,67 +64,60 @@ pub async fn handle_drop_sink(
...
-                        },
-                    )),
-                    fragment_graph: Some(graph),
-                });
+            let (mut graph, mut table, source, target_job_type) =
+                reparse_table_for_sink(&session, &table_catalog).await?;
+
+            let mut incoming_sink_ids: HashSet<_> =
+                table_catalog.incoming_sinks.iter().copied().collect();
+            incoming_sink_ids.remove(&sink_id.sink_id);
```
wenym1, line 92: Should we keep the assertion?

wenym1 (commented):
Rest LGTM

src/frontend/src/handler/drop_sink.rs
yezizp2012: Nope, the update of the table catalog and the synchronization of the sink catalog are not concurrent. The table catalog update occurs after the barrier collection. Therefore, when dropping a creating sink, the table's `incoming_sinks` may not yet be updated. Will add it back when the refactoring is done.

e2e_test/batch/catalog/rw_depend.slt.part
yezizp2012: background ddl tests will enable background_ddl randomly. This makes it uncertain when `sink2` will be created and visible in the system catalog.

src/frontend/src/handler/cancel_job.rs
yezizp2012: Yes, after the refactoring into a table, they will naturally be unified.

src/meta/src/stream/stream_manager.rs
yezizp2012: The entry will be deleted by the create handler upon receiving this error.

src/frontend/src/handler/drop_sink.rs
wenym1: Seems dangerous to have inconsistent metadata on frontend side. Do we plan to address the inconsistency in near future via notification with more atomicity?

e2e_test/batch/catalog/rw_depend.slt.part
wenym1: Do we have plan to make the background ddl visible in frontend before it returns? Currently for background ddl, we return `IGNORED_NOTIFICATION_VERSION` to frontend. But actually on meta side, the background ddl returns after collecting the first barrier, and the creating job has been marked as `Creating` already. If we return a meaningful notification version for frontend to wait for, the background ddl job can be visible immediately in the same session. 

Not related to this PR. We can address in later PR.

PR #22627, opened 2025-07-16
feat: hashicorp_vault as secret backend

tabVersion:

## Checklist

## Documentation

<details>

RisingWave now supports **HashiCorp Vault** as a secret backend, enabling secure management of sensitive credentials used in data sources and sinks. This integration allows you to store secrets externally in Vault and reference them securely within RisingWave configurations.

#### **🔐 Multiple Authentication Methods**
- **Token Authentication**: Use Vault tokens for direct access
- **AppRole Authentication**: Use role-based authentication for enhanced security

#### **🎯 Flexible Secret Access**
- Access specific fields within Vault secrets
- Support for **Vault KV v2 secrets engine**
- Configurable TLS verification settings

#### **🔗 Seamless Integration** 
- Reference Vault secrets directly in data source configurations
- Works with Kafka, database connections, and other connectors requiring credentials
- Transparent secret resolution at runtime

### 🚀 Benefits

- **Enhanced Security**: Keep sensitive credentials out of RisingWave configurations
- **Centralized Secret Management**: Leverage existing Vault infrastructure
- **Audit and Compliance**: Take advantage of Vault's audit logging and access controls
- **Secret Rotation**: Automatically pick up updated secrets without configuration changes

### 📋 Supported Configuration Options

| Parameter | Description | Required | Default |
|-----------|-------------|----------|---------|
| `backend` | Must be `'hashicorp_vault'` | ✅ | - |
| `addr` | Vault server address (e.g., `http://vault:8200`) | ✅ | - |
| `path` | Path to the secret in Vault | ✅ | - |
| `field` | Specific field within the secret to retrieve | ❌ | - |
| `auth_method` | Authentication method (`token` or `approle`) | ✅ | - |
| `auth_token` | Vault token (for token auth) | ⚡ | - |
| `auth_role_id` | Role ID (for AppRole auth) | ⚡ | - |
| `auth_secret_id` | Secret ID (for AppRole auth) | ⚡ | - |
| `tls_skip_verify` | Skip TLS certificate verification | ❌ | `false` |

⚡ *Required based on chosen authentication method*

</details>


tabVersion:
We can ignore the protobuf breaking change.

In prev impl, we ban creating hashicrop_vault backend secret so no historical data exist.

MrCroxx (approved):
LGTM on behalf of cargo-lock

ci/scripts/setup-vault.sh
```diff
@@ -0,0 +1,102 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+echo "Setting up HashiCorp Vault test data..."
+
```
kwannoel, line 6: Are there some docs from hashicorp for the setup steps? We can link them.

src/common/secret/src/vault_client.rs
```diff
@@ -0,0 +1,423 @@
...
+
+        Ok(Self { client, config })
+    }
+
+    pub async fn get_secret(&self) -> Result<Vec<u8>> {
+        // Try to get secret, with retry logic for token invalidation
+        let mut force_refresh_token = false;
+
+        // Retry loop for handling token invalidation
+        for retry_count in 0..1 {
```
kwannoel: This looks a little odd `0..1` means we only run this loop once, at `0`. `1` is excluded from the iteration.
This means the loop is not necessary.

src/common/secret/src/vault_client.rs
```diff
@@ -0,0 +1,423 @@
...
+            // Success case - process the response and break out of retry loop
+            return self.process_secret_response(response).await;
+        }
+
+        // Should never reach here due to early returns, but adding unreachable for safety
+        unreachable!("")
+    }
+
+    async fn process_secret_response(&self, response: reqwest::Response) -> Result<Vec<u8>> {
+        // a demo response:
```
kwannoel, line 231: Any api reference?

src/common/secret/src/vault_client.rs
tabVersion: Good catch, the loop here serves as a retry. If cached approle auth token expires, we can re-request one and update cache.

src/common/secret/src/vault_client.rs
tabVersion: updated in code comment.

ci/scripts/setup-vault.sh
tabVersion: updated in code comment.

src/common/secret/src/vault_client.rs
```diff
@@ -0,0 +1,424 @@
...
+
+        // Retry loop for handling token invalidation
+        for retry_count in 0..2 {
+            // Get token (either directly or via app role)
+            let token = self.get_token_internal(force_refresh_token).await?;
+
+            // Fetch secret from Vault
+            let url = format!(
+                "{}/v1/{}",
+                self.config.addr.trim_end_matches('/'),
```
kwannoel: Why don't we parse it using `url` instead, so we can verify it's a well-formed url.

kwannoel: https://docs.rs/url/latest/url/

src/common/secret/src/vault_client.rs
```diff
@@ -0,0 +1,424 @@
...
+                    response.text().await.unwrap_or_default()
+                ));
+            }
+
+            // Success case - process the response and break out of retry loop
+            return self.process_secret_response(response).await;
+        }
+
+        // Should never reach here due to early returns, but adding unreachable for safety
+        unreachable!("")
```
kwannoel: Prefer to return an internal error instead. Especially since it will run on the meta-node. At least the caller can handle the error.

src/common/secret/src/vault_client.rs
```diff
@@ -0,0 +1,424 @@
...
+                    response.status(),
+                    response.text().await.unwrap_or_default()
+                ));
+            }
+
+            // Success case - process the response and break out of retry loop
+            return self.process_secret_response(response).await;
+        }
+
+        // Should never reach here due to early returns, but adding unreachable for safety
```
kwannoel: Don't think this is true. We have this code block:
```rs
            if (response.status() == 401 || response.status() == 403)
                && retry_count == 0
                && matches!(self.config.auth, HashiCorpVaultAuth::AppRole { .. })
            {
                // this case means the token changed during cache, need to trigger a refresh
                force_refresh_token = true;
                continue;
            }
```

If it runs at the second iteration of the for loop, it will just break the loop and hit the `unreachable!` line below.

tabVersion: Hummmm, make sense. I assume the approle is alway valid here because we have done validation upon creation. But yes, there should be a catch-all error.

src/common/secret/src/vault_client.rs
```diff
@@ -0,0 +1,424 @@
...
+    async fn get_token_internal(&self, force_refresh: bool) -> Result<String> {
+        match &self.config.auth {
+            HashiCorpVaultAuth::Token { auth_token } => Ok(auth_token.clone()),
+            HashiCorpVaultAuth::AppRole {
+                auth_role_id,
+                auth_secret_id,
+            } => {
+                // Create cache key with vault base URL and role_id
+                let cache_key = TokenCacheKey {
+                    vault_base_url: self.config.addr.trim_end_matches('/').to_owned(),
```
kwannoel, line 285: As per above, we should use a url parsing library.

e2e_test/source_inline/vault/vault_secret_ddl.slt
```diff
@@ -0,0 +1,60 @@
...
+# Test error case - invalid field
+statement error
+CREATE SECRET vault_invalid_field
+WITH (
+  backend = 'hashicorp_vault',
+  addr = 'http://vault-server:8200',
+  path = 'secret/data/myapp/db',
+  field = 'nonexistent_field',
+  auth_method = 'token',
+  auth_token = 'root-token'
```
kwannoel, line 33: Can we use `secret` for `auth_token` here 🤔 . I suppose we can bootstrap by storing the `auth_token` secret inside RW.

src/frontend/src/handler/alter_secret.rs
```diff
@@ -67,13 +66,16 @@ pub async fn handle_alter_secret(
                     };
                     secret_payload.encode_to_vec()
                 }
-                secret::SecretBackend::HashicorpVault(_) => {
-                    bail_not_implemented!("hashicorp_vault backend is not implemented yet")
+                secret::SecretBackend::HashicorpVault(_vault_backend) => {
+                    return Err(crate::error::ErrorCode::InvalidParameterValue(
+                        "credential must be null when altering hashicorp_vault backend".to_owned(),
```
kwannoel: How come?

src/frontend/src/handler/create_secret.rs
```diff
@@ -51,7 +52,15 @@ pub async fn handle_create_secret(
     }
     let with_options = WithOptions::try_from(stmt.with_properties.0.as_ref() as &[SqlOption])?;
 
-    let secret_payload = get_secret_payload(stmt.credential, with_options)?;
+    // Check for secret references in WITH options (forbid them during secret creation)
+    if !with_options.secret_ref().is_empty() {
+        return Err(ErrorCode::InvalidParameterValue(
+            "Secret references are not allowed when creating a secret".to_owned(),
```
kwannoel, line 56: Actually why do we reject it 🤔 Don't think there's a risk of recursion.

src/frontend/src/handler/alter_secret.rs
tabVersion: you can skip the ALTER  part, I doubt in which case users will alter secrets in risingwave, rather than changing directly in Vault.

e2e_test/source_inline/vault/vault_secret_ddl.slt
tabVersion: The problem lays on recursion secret. Can place it on roadmap.

src/frontend/src/handler/create_secret.rs
tabVersion: The fill_secrets interface is designed for resolving secrets at the same layer. I prefer to ban first the impl in following pr.

kwannoel (commented):
On recovery, how do we get secrets from Hashicorp vault?

kwannoel (approved):

kwannoel:
LGTM, thanks for the efforts

src/expr/impl/tests/sig.rs
```diff
@@ -45,7 +45,10 @@ fn test_func_sig_map() {
                         "{}({}) -> {}",
                         v[0].name.as_str_name().to_ascii_lowercase(),
                         v[0].inputs_type.iter().format(", "),
-                        v.iter().map(|sig| &sig.ret_type).format("/")
+                        v.iter()
+                            .map(|sig| sig.ret_type.to_string())
+                            .sorted() // sort the return types to make the test stable
+                            .format("/")
```
tabVersion, line 51: credits to @xiangjinwu for stabilizing the test

PR #22631, opened 2025-07-16
feat: parallelize apply_chunk in agg

lmatz:

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/executor/aggregate/agg_group.rs
+++ b/src/stream/src/executor/aggregate/agg_group.rs
@@ -16,6 +16,7 @@ use std::fmt::Debug;
 use std::marker::PhantomData;
 use std::sync::Arc;
 
+use futures::future::try_join_all;
 use risingwave_common::array::StreamChunk;
 use risingwave_common::array::stream_record::{Record, RecordType};
 use risingwave_common::bitmap::Bitmap;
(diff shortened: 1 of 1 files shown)


src/stream/src/executor/aggregate/agg_group.rs
```diff
@@ -383,12 +384,24 @@ impl<S: StateStore, Strtg: Strategy> AggGroup<S, Strtg> {
...
             tracing::trace!(group = ?self.ctx.group_key_row(), "first time see this group");
         }
-        for (((state, call), func), visibility) in (self.states.iter_mut())
-            .zip_eq_fast(calls)
-            .zip_eq_fast(funcs)
-            .zip_eq_fast(visibilities)
-        {
-            state.apply_chunk(chunk, call, func, visibility).await?;
+
+        let concurrency = 10;
```
lmatz, line 388: or try_join_all at once,

I saw a few other places that just try all of them in agg

chenzl25: 10 looks good to me

lmatz:
```
pub async fn apply_chunk(
        &mut self,
        chunk: &StreamChunk,
        call: &AggCall,
        func: &BoxedAggregateFunction,
        visibility: Bitmap,
    ) -> StreamExecutorResult<()> {
        match self {
            Self::Value(state) => {
                let chunk = chunk.project_with_vis(call.args.val_indices(), visibility);
                func.update(state, &chunk).await?;
                Ok(())
            }
            Self::MaterializedInput(state) => {
                // the input chunk for minput is unprojected
                let chunk = chunk.clone_with_vis(visibility);
                state.apply_chunk(&chunk)
            }
        }
    }
```

Actually most of the time `func.update` is a lightweight `sync` function.

PR #22642, opened 2025-07-17
fix(file source): Optimize list prefix handling to avoid root directory scans

wcy-fdu:

As title, previously, to support the match pattern set as a single file, we changed the prefixes without a trailing slash to the list root directory, which resulted in too many files being listed. This PR addresses that issue.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/src/source/filesystem/opendal_source/opendal_enumerator.rs
+++ b/src/connector/src/source/filesystem/opendal_source/opendal_enumerator.rs
@@ -53,8 +53,15 @@ impl<Src: OpendalSource> SplitEnumerator for OpendalEnumerator<Src> {
     async fn list_splits(&mut self) -> ConnectorResult<Vec<OpendalFsSplit<Src>>> {
         let empty_split: OpendalFsSplit<Src> = OpendalFsSplit::empty_split();
         let prefix = self.prefix.as_deref().unwrap_or("/");
+        let list_prefix = if prefix.ends_with("/") {
+            prefix.to_owned()
+        } else if let Some(parent_pos) = prefix.rfind('/') {
+            prefix[..=parent_pos].to_owned()
+        } else {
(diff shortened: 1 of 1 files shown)


src/connector/src/source/filesystem/opendal_source/opendal_enumerator.rs
```diff
@@ -120,3 +131,55 @@ impl<Src: OpendalSource> OpendalEnumerator<Src> {
...
+    #[test]
+    fn test_prefix_logic() {
+        let test_cases = vec![
+            ("a/b/c/hello*/*.json", "a/b/c/hello", "a/b/c/"),
+            ("a/b/c.json", "a/b/c.json", "a/b/"),
+            ("a/b/c/", "a/b/c/", "a/b/c/"),
+            ("a/b/c", "a/b/c", "a/b/"),
+            ("file.json", "file.json", "/"),
+            ("*.json", "", "/"),
+            ("a/b/c/[h]ello*/*.json", "a/b/c/", "a/b/c/"),
```
tabVersion, line 154: according to the logic `prefix.rfind('/')`, why this case expects `a/b/c/` rather than `a/b/c/[h]ello/`

wcy-fdu: `get_prefix()` is designed to [extract "safe non-glob parts".](https://github.com/risingwavelabs/risingwave/blob/9e06d9908de6701a3df6598f6f4500dd9aa02b20/src/connector/src/source/filesystem/s3/enumerator.rs#L16) It stops when it encounters any glob special characters (*, [, {), so `get_prefix(""a/b/c/[h]ello*/*.json")` return `a/b/c/`.

tabVersion (approved):
thanks for the fix

PR #5246, opened 2022-09-09
feat(risedev): introduce DocSlt tool to extract SQL examples written in SLT syntax to SLT files

richardchien:
I hereby agree to the terms of the [Singularity Data, Inc. Contributor License Agreement](https://gist.github.com/skyzh/0663682a70b0edde7ae991492f2314cb#file-s9y_cla).

## What's changed and what's your intention?

Implementation for #5117.

## Checklist

- [x] I have written necessary rustdoc comments
- [x] I have added necessary unit tests and integration tests
- [x] All checks passed in `./risedev check` (or alias, `./risedev c`)

## Refer to a related PR or issue link (optional)

#5117

Issue #5117, opened 2022-09-05
RFC: DocSlt, documentation examples as sqllogictest tests


richardchien:
### Background

Currently we write e2e tests in the form of sqllogictest scripts, and we have a dedicated `e2e_test` folder in the source tree for this. However, sometimes we may want to write some SQL examples directly in comments near the code to document the behavior of code, so that further modification won't break the tests easily.

Such cases include:
 https://github.com/risingwavelabs/risingwave/blob/14de3bff08d0738b3badd2acc94d8155980a6298/src/expr/src/expr/expr_array_concat.rs#L76-L83

https://github.com/risingwavelabs/risingwave/blob/5f8fd83064691677ea91d18c1a960d920c2f3a07/src/expr/src/vector_op/agg/functions.rs#L128-L133

https://github.com/risingwavelabs/risingwave/blob/5f8fd83064691677ea91d18c1a960d920c2f3a07/src/frontend/src/binder/query.rs#L48-L78

### Design

Like **rustdoc**'s [documentation tests](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html), this RFC introduces **DocSlt**, to allow running slt tests in doc comments.

Developers can write examples in slt syntax like below:

````rust
/// Concatenates two arrays with same data type.
/// The behavior is the same as PG.
///
/// Examples:
///
/// ```slt
/// query T
/// select array_cat(array[66], array[123]);
/// ----
/// {66,123}
///
/// query T
/// select array_cat(array[66], null::int[]);
/// ----
/// {66}
///
/// query T
/// select array_cat(null::int[], array[123]);
/// ----
/// {123}
/// ```
````

Then use the following commands to generate slt files from such documentation and run those slt tests:

```sh
./risedev docslt -p risingwave_expr # extract SQL examples in slt syntax from `risingwave_expr`
./risedev slt-generated -p 4566 -d dev
# or ./risedev slt -p 4566 -d dev 'e2e_test/generated/**/*.slt'
```

The generated slt files should be added into source version control, just like the yaml files of planner tests.

The generation is in granularity of file, which means all the docslt blocks in one source file are placed in one slt file.

You can use `slt-setup` and `slt-teardown` blocks to control the setup and teardown process, like this:

````rust
//! To setup and teardown docslt environment, you can use `slt-setup`
//! and `slt-teardown` code blocks:
//!
//! ```slt-setup
//! statement ok
//! create table t (a int);
//! ```
//!
[...]

PR #22647, opened 2025-07-17
feat(expr): more `vector` distance functions

xiangjinwu:

In addition to `l2_distance`/`<->`, this PR adds 3 more distance functions found in pgvector:
* `cosine_distance`/`<=>`
* `l1_distance`/`<+>`
* `inner_product` (and `<#>` being negative inner product)

The docslt test cases are taken from pgvector regress tests:
https://github.com/pgvector/pgvector/blob/v0.8.0/test/expected/vector_type.out#L373-L535

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/expr/impl/src/scalar/vector_dist.rs
+++ b/src/expr/impl/src/scalar/vector_dist.rs
@@ -16,13 +16,10 @@ use risingwave_common::types::{F64, VectorRef};
 use risingwave_common::util::iter_util::ZipEqFast;
 use risingwave_expr::{ExprError, Result, function};
 
-#[function("l2_distance(vector, vector) -> float8"/*, type_infer = "unreachable"*/)]
-fn l2_distance(lhs: VectorRef<'_>, rhs: VectorRef<'_>) -> Result<F64> {
-    let lhs = lhs.into_slice();
-    let rhs = rhs.into_slice();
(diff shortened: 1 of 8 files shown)


src/expr/impl/src/scalar/vector_dist.rs
```diff
@@ -31,10 +28,210 @@ fn l2_distance(lhs: VectorRef<'_>, rhs: VectorRef<'_>) -> Result<F64> {
...
+
     let mut sum = 0.0f32;
     for (l, r) in lhs.iter().zip_eq_fast(rhs.iter()) {
         let diff = l - r;
         sum += diff * diff;
     }
     Ok((sum as f64).sqrt().into())
 }
+
+/// ```slt
```
wenym1, line 77: Curious how do we make rust doc test to run the slt test internal?

wenym1 (approved):
LGTM

src/expr/impl/src/scalar/vector_dist.rs
xiangjinwu: #5117 / #5246

They are not executed by rust doc test. They just follow similar syntax but extracted into standalone slt files in `e2e_test/generated/docslt` by risedev, and then executed together as other slt files.

PR #22660, opened 2025-07-18
initial impl

wenym1:

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/meta/src/rpc/ddl_controller.rs
+++ b/src/meta/src/rpc/ddl_controller.rs
@@ -23,9 +23,7 @@ use anyhow::{Context, anyhow};
 use await_tree::{InstrumentAwait, span};
 use either::Either;
 use itertools::Itertools;
-use risingwave_common::catalog::{
-    AlterDatabaseParam, ColumnCatalog, ColumnId, Field, FragmentTypeFlag,
-};
+use risingwave_common::catalog::{AlterDatabaseParam, ColumnCatalog, ColumnId, FragmentTypeFlag};
 use risingwave_common::config::DefaultParallelism;
 use risingwave_common::hash::VnodeCountCompat;
 use risingwave_common::secret::{LocalSecretManager, SecretEncryption};
@@ -36,7 +34,6 @@ use risingwave_common::util::stream_graph_visitor::{
 use risingwave_common::{bail, bail_not_implemented, must_match};
 use risingwave_connector::WithOptionsSecResolved;
 use risingwave_connector::connector_common::validate_connection;
-use risingwave_connector::sink::catalog::SinkType;
 use risingwave_connector::source::{
     ConnectorProperties, SourceEnumeratorContext, UPSTREAM_SOURCE_KEY,
 };
@@ -58,10 +55,9 @@ use risingwave_pb::ddl_service::{
 };
 use risingwave_pb::meta::table_fragments::PbActorStatus;
 use risingwave_pb::meta::table_fragments::actor_status::PbActorState;
-use risingwave_pb::plan_common::PbColumnCatalog;
-use risingwave_pb::stream_plan::stream_node::{NodeBody, PbNodeBody};
+use risingwave_pb::stream_plan::stream_node::NodeBody;
 use risingwave_pb::stream_plan::{
-    MergeNode, PbDispatchOutputMapping, PbDispatcherType, PbStreamFragmentGraph, PbStreamScanType,
+    MergeNode, PbDispatchOutputMapping, PbDispatcherType, PbStreamFragmentGraph,
     StreamFragmentGraph as StreamFragmentGraphProto,
 };
 use risingwave_pb::telemetry::{PbTelemetryDatabaseObject, PbTelemetryEventStage};
@@ -89,8 +85,8 @@ use crate::stream::{
     ActorGraphBuildResult, ActorGraphBuilder, AutoRefreshSchemaSinkContext,
     CompleteStreamFragmentGraph, CreateStreamingJobContext, CreateStreamingJobOption,
     GlobalStreamManagerRef, JobRescheduleTarget, ReplaceStreamJobContext, SourceChange,
-    SourceManagerRef, StreamFragmentGraph, clone_fragment, create_source_worker, state_match,
-    validate_sink,
+    SourceManagerRef, StreamFragmentGraph, create_source_worker,
+    rewrite_refresh_schema_sink_fragment, state_match, validate_sink,
 };
 use crate::telemetry::report_event;
 use crate::{MetaError, MetaResult};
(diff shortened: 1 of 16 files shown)


wenym1:
* **#22661** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22661?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#22660** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22660?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22660?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#22611** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22611?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/meta/src/stream/stream_graph/fragment.rs
```diff
@@ -331,6 +336,262 @@ impl StreamFragmentEdge {
...
+    fn extend_sink_columns(
+        sink_columns: &mut Vec<PbColumnCatalog>,
+        new_columns: &[ColumnCatalog],
+        get_column_name: impl Fn(&String) -> String,
+    ) {
+        let next_column_id = sink_columns
+            .iter()
+            .map(|col| col.column_desc.as_ref().unwrap().column_id + 1)
+            .max()
+            .unwrap_or(1);
```
hzxa21, line 433: I think it means there are no columns in sink when reaching here? Is it actually reachable?

src/meta/src/stream/stream_graph/fragment.rs
```diff
@@ -331,6 +336,262 @@ impl StreamFragmentEdge {
...
+    };
+    let NodeBody::Merge(merge) = merge_node.node_body.as_mut().unwrap() else {
+        return Err(anyhow!(
+            "expect PbNodeBody::Merge but got: {:?}",
+            merge_node.node_body
+        )
+        .into());
+    };
+    // update sink_node
+    sink_node.identity = {
```
hzxa21, line 476: The logic here seems to be borrowed from [StreamSink::distill](https://github.com/risingwavelabs/risingwave/blob/1edb01bc18580b29b9f23a55a3aca6d3d7b53248/src/frontend/src/optimizer/plan_node/stream_sink.rs#L643). Let's document it down.

src/meta/src/stream/stream_graph/fragment.rs
```diff
@@ -331,6 +336,262 @@ impl StreamFragmentEdge {
...
+        return Err(anyhow!(
+            "expect PbNodeBody::Merge but got: {:?}",
+            merge_node.node_body
+        )
+        .into());
+    };
+    Ok(())
+}
+
+pub fn rewrite_refresh_schema_sink_fragment(
```
hzxa21, line 415: @chenzl25 PTAL. It seems that we mimic some frontend logics here to manipulate the stream nodes.

src/meta/src/stream/stream_graph/fragment.rs
```diff
@@ -331,6 +336,262 @@ impl StreamFragmentEdge {
...
+        } else {
+            "".to_owned()
+        };
+        format!("StreamSink {{ type: {sink_type_str}, columns: [{column_names}]{downstream_pk} }}")
+    };
+    sink_node
+        .fields
+        .extend(newly_added_columns.iter().map(|col| {
+            Field::new(
+                format!("{}.{}", upstream_table.name, col.column_desc.name),
```
hzxa21, line 507: Is it correct to put `upstream_table.name` in a sink field?

src/meta/src/rpc/ddl_controller.rs
```diff
@@ -2172,6 +2356,14 @@ impl DdlController {
             old_fragments.max_parallelism,
         );
 
+        if let Some(sinks) = &auto_refresh_schema_sinks {
+            for sink in sinks {
+                replace_upstream
+                    .remove(&sink.new_fragment.fragment_id)
+                    .expect("should exist");
```
hzxa21, line 2362: Why do we need to remove the sink fragment?

hzxa21:
@yezizp2012 @chenzl25 should have more contexts to review this PR. Please take a quick look on whether the changes make sense to you. Most logics are related to rewriting the fragment and stream node inside the replace table logic.

Note that this PR will be merged to the dev branch first before going onto main so we don't need to resolve everything during the reviews.

src/meta/src/rpc/ddl_controller.rs
```diff
@@ -1556,26 +1564,157 @@ impl DdlController {
...
+                    .map(|col| ColumnCatalog::from(col.clone()))
+                    .collect_vec();
+                if !original_table_column_ids.is_empty() {
+                    return Err(anyhow!("new table columns does not contains all original columns. new: {:?}, original: {:?}, not included: {:?}", table.columns, original_table_columns, original_table_column_ids).into());
+                }
+                let mut sinks = Vec::with_capacity(auto_refresh_schema_sinks.len());
+                for sink in auto_refresh_schema_sinks {
+                    let sink_job_fragments = self
+                        .metadata_manager
+                        .get_job_fragments_by_id(&risingwave_common::catalog::TableId::new(sink.id))
```
chenzl25, line 1600: We'd better have a dedicated method to get job fragments by sink id, because it is a bit confusing to pass sink id as table id.

src/meta/src/stream/stream_graph/fragment.rs
```diff
@@ -331,6 +336,262 @@ impl StreamFragmentEdge {
...
+    let mut new_sink_fragment = clone_fragment(original_sink_fragment, id_generator_manager);
+    let sink_node = &mut new_sink_fragment.nodes;
+    let PbNodeBody::Sink(sink_node_body) = sink_node.node_body.as_mut().unwrap() else {
+        return Err(anyhow!("expect PbNodeBody::Sink but got: {:?}", sink_node.node_body).into());
+    };
+    let [stream_scan_node] = sink_node.input.as_mut_slice() else {
+        panic!("Sink has more than 1 input: {:?}", sink_node.input);
+    };
+    let PbNodeBody::StreamScan(scan) = stream_scan_node.node_body.as_mut().unwrap() else {
+        return Err(anyhow!(
```
chenzl25, line 455: It is possible that there is a StreamExchange between StreamSink and StreamScan when the upstream table's pk is different from the sink upsert pk.

chenzl25: For iceberg sink with partition by, there would be an additional ProjectNode here, but I think we don't need to handle it in this PR. Just remind that in the future implementation.

chenzl25 (approved):
Rest LGTM

src/meta/src/rpc/ddl_controller.rs
wenym1: It's actually `job_id`

src/meta/src/stream/stream_graph/fragment.rs
wenym1: This is the field of the `StreamNode` of sink, rather than the sink itself in the catalog. The naming convention is following the existing other field.

src/meta/src/stream/stream_graph/fragment.rs
wenym1: Theoretically, it's possible to create a table with 0 columns, and then create a sink on it, and then do alter table add column on the table.

src/meta/src/rpc/ddl_controller.rs
wenym1: The replace_upstream is used for normal alter table. 

For example, let's say we have a table, and a downstream mv or non-auto-schema change sink created on the table. When we alter table and add a column, the fragment of the table will be changed to a new fragment, and then the existing fragments of downstream jobs should do a `replace_upstream`, to replace its original upstream fragment id to the table fragment to the new table fragment. However, for sinks with auto schema change, its existing fragment will be dropped, and new fragment will be created, and therefore there won't be a `replace_upstream` happened for the previous sink fragment, and then we should remove it here.

PR #22664, opened 2025-07-18
chore: parallelize apply_chunk in agg

lmatz:

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/common/estimate_size/src/collections/hashmap.rs
+++ b/src/common/estimate_size/src/collections/hashmap.rs
@@ -63,6 +63,16 @@ where
             .map(|v| MutGuard::new(v, &mut self.heap_size))
     }
 
+    pub fn remove(&mut self, key: &K) -> Option<V> {
+        if let Some(v) = self.inner.remove(key) {
+            self.heap_size.sub_val(key);
+            self.heap_size.sub_val(&v);
+            Some(v)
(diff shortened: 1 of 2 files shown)


chenzl25 (approved):
LGTM

lmatz:
I will do a benchmark before merging it

BugenZhao (commented):
I've dived into the source code, finding that the only `await` point this PR may help to make concurrent (note: NOT parallelized) is the evaluation of expressions, including agg call itself and the `FILTER` clause.

However, only UDFs involve an `await` point; all other expressions are purely CPU work. Considering there are actually rare use cases for UDF in agg, I don't think changes in this PR (also #22631) will be helpful in most cases.

chenzl25:
> I've dived into the source code, finding that the only `await` point this PR may help to make concurrent (note: NOT parallelized) is the evaluation of expressions, including agg call itself and the `FILTER` clause.
> 
> However, only UDFs involve an `await` point; all other expressions are purely CPU work. Considering there are actually rare use cases for UDF in agg, I don't think changes in this PR (also #22631) will be helpful in most cases.

You are right. Almost all agg calculations here are purely CPU work.

chenzl25:
No updates

PR #22686, opened 2025-07-22
feat(sink): support alter add column for snowflake redshifit sink

xxhZs:

## Checklist

## Documentation

<details>

</details>


java/connector-node/risingwave-sink-jdbc/src/main/java/com/risingwave/connector/JDBCSink.java
```diff
@@ -68,7 +69,18 @@ public JDBCSink(JDBCSinkConfig config, TableSchema tableSchema) {
...
+                            .map(
+                                    columnName -> {
+                                        Integer sqlType = columnTypeMapping.get(columnName);
+                                        if (sqlType != null) {
+                                            return sqlType;
+                                        }
+                                        // Default conversion logic for PostgreSQL data types
+                                        int columnIndex = tableSchema.getColumnIndex(columnName);
+                                        var columnDesc = tableSchema.getColumnDesc(columnIndex);
+                                        return convertRisingWaveTypeToSqlType(
```
hzxa21: This seems to be postgres specific. However, `JDBCSink` is not just used by postgres. Will there be an issue?

src/connector/src/sink/remote.rs
```diff
@@ -661,13 +662,11 @@ impl SinkWriter for CoordinatedRemoteSinkWriter {
...
             let rsp = self.stream_handle.commit(epoch).await?;
-            rsp.metadata
-                .ok_or_else(|| {
-                    SinkError::Remote(anyhow!(
-                        "get none metadata in commit response for coordinated sink writer"
-                    ))
-                })
-                .map(Some)
+            Ok(Some(rsp.metadata.unwrap_or_else(|| SinkMetadata {
+                metadata: Some(sink_metadata::Metadata::Serialized(SerializedMetadata {
```
hzxa21, line 666: Why do we change to not return any error here?

java/connector-node/risingwave-sink-jdbc/src/main/java/com/risingwave/connector/JDBCSink.java
xxhZs: The reason for this logic is that auto schema change can lead to schema inconsistencies between rw and downstream systems, so currently only Redshift requires this logic.
 However, it will not have any impact on the existing pg sink.

src/connector/src/sink/remote.rs
xxhZs: This is because it is a normal phenomenon for this commit to return "none". The commit of Redshift sink does not return any value, but we have adopted Coordinator method, so it is necessary to mock one here.

PR #22700, opened 2025-07-23
refactor(frontend): static optimizer plan convention

wenym1:

Previously, we have a unified `PlanRef` for plan node of logical, batch and stream convention. However, some traits are only for some specific conventions rather than all of them. For example, `PredicatePushDown`, `ColPrunable`, `ToBatch` and `ToStream` are only for logical `PlanRef`, and `ToDistributedBatch` and `ToLocalBatch` are only for batch `PlanRef`. For conventions other than the specific convention, we implement the traits with a dummy `bail!` method body. Besides, it should be invalid for a `PlanRef` of a convention to call the `as_$convention_$name` method of another convention. For example, for a `PlanRef` of `Logical` convention, it should be forbidden to call `as_stream_scan`. Furthermore, we have forbidden calling `PlanVisitor` or apply optimizer rule to a `PlanRef` of a different convention, but this is achieved by a runtime sanity check `PlanRef::expect_convention`.

In this PR, we will change these runtime checked requirement to static type change. This PR seems large, but it all starts with the changes in `src/frontend/src/optimizer/plan_node/mod.rs`, and changes in other files are all according to these changes.

Major changes in `src/frontend/src/optimizer/plan_node/mod.rs`:
* Add a new associated type `type PlanRefDyn;` in `Convention` trait to hold the trait combination of `PlanRef` in different conventions. Removed the original combinational trait `PlanNode`, move the common traits in it, such as `Distill`, `Explain` and etc, to `PlanNodeCommon<C>`. Define trait `StreamPlanNode = PlanNodeCommon<Stream>`, `LogicalPlanNode = PlanNodeCommon<Logical> + ColPrunable + PredicatePushdown + ToBatch + ToStream` and `BatchPlanNode = PlanNodeCommon<Batch> + ToDistributedBatch + ToLocalBatch`. 
* Accordingly, add a new generic type to `PlanRef`, change its definition from `struct PlanRef(Rc<dyn PlanNode>)` to `struct PlanRef<C: ConventionMarker>(Rc<C::PlanRefDyn>)`. Most changes in this PR are caused by adding a new generic type to `PlanRef`. Alias `LogicalPlanRef = PlanRef<Logical>`, `StreamPlanRef = PlanRef<Stream>` and `BatchPlanRef = PlanRef<Batch>`. The original trait `StreamPlanRef` and `BatchPlanRef` are renamed to `StreamPlanNodeMetadata` and `BatchPlanNodeMetadata` to avoid naming duplication.
* no enum `Convention`, since the convention is now statically marked. All methods that depended on dynamic dispatch on the enum are removed. All dynamic runtime check on convention are removed.
* in `impl_down_cast_fn`, we previously allow calling `as_$convention_$node` method for a `PlanRef` in any convention to all conventions. We change to allow calling the methods only in its own convention.

## Checklist

## Documentation

<details>

</details>


wenym1:
* **#22700** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22700?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22700?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#22732** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22732?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/frontend/src/optimizer/plan_node/logical_join.rs
```diff
@@ -933,13 +940,13 @@ impl LogicalJoin {
...
         ctx: &mut ToStreamContext,
-    ) -> Result<PlanRef> {
+    ) -> Result<StreamPlanRef> {
         use super::stream::prelude::*;
 
         assert!(predicate.has_eq());
         let (left, right) = self.get_stream_input_for_hash_join(&predicate, ctx)?;
 
-        let logical_join = self.clone_with_left_right(left, right);
+        let core = self.core.clone_with_inputs(left, right);
```
wenym1, line 950: The code change here avoids a plan node clone to incremental the accumulated plan node id, and therefore causes the change in `src/frontend/planner_test/tests/testdata/output/explain_dot_format.yaml`.

If we manually inject a
```
let _logical_join =
            self.clone_with_left_right(self.core.left.clone(), self.core.right.clone());
```
the output file will be the same as original.

wenym1:
@chenzl25 @BugenZhao I have updated the PR description. The PR is ready for review now. PTAL

src/frontend/src/handler/create_table.rs
```diff
@@ -76,9 +76,11 @@ use crate::handler::create_source::{
 };
 use crate::handler::util::SourceSchemaCompatExt;
 use crate::optimizer::plan_node::generic::{CdcScanOptions, SourceNodeKind};
-use crate::optimizer::plan_node::{LogicalCdcScan, LogicalSource};
+use crate::optimizer::plan_node::{
+    LogicalCdcScan, LogicalPlanRef, LogicalSource, StreamPlanRef as PlanRef,
```
chenzl25, line 81: Shall we use `StreamPlanRef` directly?

src/frontend/src/optimizer/property/distribution.rs
```diff
@@ -300,28 +300,31 @@ impl RequiredDist {
...
         } else {
             Ok(plan)
         }
     }
 
-    pub fn no_shuffle(plan: PlanRef) -> PlanRef {
-        match plan.convention() {
-            Convention::Stream => StreamExchange::new_no_shuffle(plan).into(),
-            Convention::Logical | Convention::Batch => unreachable!(),
+    pub fn streaming_enforce_if_not_satisfies(&self, plan: StreamPlanRef) -> Result<StreamPlanRef> {
```
chenzl25, line 316: Good observation.

src/frontend/src/optimizer/rule/batch/mod.rs
```diff
@@ -18,7 +18,7 @@ pub(crate) mod batch_project_merge_rule;
 pub mod batch_push_limit_to_scan_rule;
 
 mod prelude {
-    pub(super) use crate::optimizer::plan_node::{Batch, PlanRef};
+    pub(super) use crate::optimizer::plan_node::{Batch, BatchPlanRef as PlanRef};
```
chenzl25, line 21: ditto. Use `BatchPlanRef` directly?

chenzl25 (commented):
I think we can search `LogicalPlanRef as PlanRef`, `BatchPlanRef as PlanRef` and `StreamPlanRef as PlanRef` to use the specific PlanRef directly.

wenym1:
> I think we can search `LogicalPlanRef as PlanRef`, `BatchPlanRef as PlanRef` and `StreamPlanRef as PlanRef` to use the specific PlanRef directly.

I am trying to minimize the code changes in this PR so that it can be more friendly for the code reviewer. If we don't alias the specific plan ref, the LoC will be more than 3000.

Since the plan ref is strong typed, we don't need to worry about misusing the convention or type.

src/frontend/src/optimizer/plan_node/generic/join.rs
```diff
@@ -148,7 +162,9 @@ impl<I: stream::StreamPlanRef> Join<I> {
     }
 }
 
-impl<PlanRef: GenericPlanRef> GenericPlanNode for Join<PlanRef> {
+impl<PlanRef: GenericPlanRef, RightPlanRef: GenericPlanRef> GenericPlanNode
+    for Join<PlanRef, RightPlanRef>
+{
```
chenzl25: Why do we need `RightPlanRef`?

wenym1: In `BatchLookupJoin`, the right plan ref is a `LogicalScan`, and it's in logical convention, while the left plan ref is a batch plan ref.

chenzl25: I see, but it seems we can replace it with BatchSeqScan instead of LogicalScan. It is not hard and will make the Join operator look clearer. I will try to create a PR to merge it into your branch.

chenzl25: https://github.com/risingwavelabs/risingwave/pull/22756

chenzl25 (approved):
Rest LGTM. Thank you for the great refactor!

src/frontend/src/handler/explain.rs
```diff
@@ -251,7 +261,11 @@ pub async fn do_handle_explain(
             }
             ExplainType::Physical => {
                 // if explain trace is on, the plan has been in the rows
-                if !explain_trace && let Ok(plan) = &plan {
+                if !explain_trace && let Ok(physical_plan) = &plan {
+                    let plan = match &physical_plan {
+                        PhysicalPlanRef::Stream(plan) => plan as &dyn Explain,
+                        PhysicalPlanRef::Batch(plan) => plan as &dyn Explain,
+                    };
```
BugenZhao, line 268: Is it possible to extract this into a method?

src/frontend/src/handler/explain.rs
```diff
@@ -180,20 +185,26 @@ pub async fn do_handle_explain(
...
                     }
 
                     _ => bail_not_implemented!("unsupported statement for EXPLAIN: {stmt}"),
                 }?;
 
-                let context = plan.ctx().clone();
+                let context = match &plan {
+                    PhysicalPlanRef::Stream(plan) => plan.ctx(),
+                    PhysicalPlanRef::Batch(plan) => plan.ctx(),
+                };
```
BugenZhao, line 207: Ditto

src/frontend/src/optimizer/plan_node/logical_join.rs
```diff
@@ -319,7 +320,7 @@ impl LogicalJoin {
     fn to_batch_lookup_join(
         &self,
         predicate: EqJoinPredicate,
-        logical_join: generic::Join<PlanRef>,
+        logical_join: generic::Join<BatchPlanRef>,
```
BugenZhao, line 323: Is this correct?

BugenZhao (approved):
Looks great!

src/frontend/src/handler/explain.rs
wenym1: Seems unnecessary, since this enum is declared and used only within the method body, and is cast to dyn Explain for only once. Same for `.ctx()`

src/frontend/src/optimizer/plan_node/logical_join.rs
wenym1: I think so. It was just a naming issue. It actually accepts a `PlanRef` of batch convention in the original logic.

PR #22712, opened 2025-07-23
refactor(frontend): unify handling logic of LogicalShare and StreamShare

wenym1:

Previously, we used to handle `LogicalShare` and `StreamShare` separately in `PlanRef::clone_with_inputs` and `PlanRewriter`. In `PlanRef::clone_with_inputs`, we do `as_logical_share` and `as_stream_share` individually before calling the actual `clone_with_inputs` of the plan node to avoid duplicate handling on a same shared node. In `PlanRewriter`, we have separate `rewrite_stream_share` and `rewrite_logical_share` to handle the same logic of share node.

In this PR, we will unify the handling logic of the two share nodes. In `ConventionMarker`, we add a new associated type `ShareNode`, which satisfies the `ShareNode` trait with the following definition
```
pub trait ShareNode: AnyPlanNodeMeta + PlanTreeNodeUnary + 'static {
    fn new_share(share: generic::Share<PlanRef>) -> PlanRef;
    fn replace_input(&self, plan: PlanRef);
}
```
The `ConventionMarker` will also have a new method `fn as_share(plan: &dyn PlanNode) -> Option<&Self::ShareNode>;` to cast a dynamic PlanRef to the share node of the convention. For `Batch` convention, it has no share node at the moment. Therefore, we introduce a `NoShareNode`, which holds the `!` never type and can never be constructed, and implement `ShareNode` for it with as methods implemented as unreachable. 

The `PlanRef::clone_with_inputs` can then be implemented with unified logic for share node
```
pub fn clone_root_with_inputs<C: ConventionMarker>(&self, inputs: &[PlanRef]) -> PlanRef {
    self.expect_convention::<C>();
    if let Some(share) = self.as_share_node::<C>() {
        assert_eq!(inputs.len(), 1);
        share.replace_input(inputs[0].clone());
        self.clone()
    } else {
        // Dispatch to dyn PlanNode instead of PlanRef.
        let dyn_t = self.deref();
        dyn_t.clone_with_inputs(inputs)
    }
}
```

For `PlanRewriter`, we used to include a series of `rewrite_$convention_$name` methods in the trait. However, for this trait, we mostly implement the `rewrite_$convention_share` method specially to avoid duplicate handling on a same share node. In this PR, we will also unify the handling logic of the two share nodes. The `PlanRewriter` will change to be defined in the following way
```
pub trait PlanRewriter<C: ConventionMarker> {
    fn rewrite_with_inputs(&mut self, plan: &PlanRef, inputs: Vec<PlanRef>) -> PlanRef;
}
```
and we will have a `rewrite_recursively` method to unify the handling logic of share node.
```
fn rewrite_recursively<C: ConventionMarker>(
    &self,
    rewriter: &mut impl PlanRewriter<C>,
    share_map: &mut HashMap<PlanNodeId, PlanRef>,
) -> PlanRef;
```

Inside the implementation of `rewrite_with_inputs` for each `PlanRewriter`, we can just do `plan.as_$convention_$name` method to check the node type and do the special handling logic.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/optimizer/plan_node/expr_rewritable.rs
+++ b/src/frontend/src/optimizer/plan_node/expr_rewritable.rs
@@ -18,7 +18,7 @@ use super::*;
 /// the `ExprRewriter` needs to be idempotent i.e., applying it more than once
 /// to the same `ExprImpl` will be a noop on subsequent applications.
 /// `rewrite_exprs` should only return a plan with the given node modified.
-/// To rewrite recursively, call `rewrite_exprs_recursive` on [`RewriteExprsRecursive`].
+/// To rewrite recursively, call `rewrite_exprs_recursive`.
 pub trait ExprRewritable {
     fn has_rewritable_expr(&self) -> bool {
         false
--- a/src/frontend/src/optimizer/plan_node/mod.rs
+++ b/src/frontend/src/optimizer/plan_node/mod.rs
@@ -45,7 +45,6 @@ use risingwave_common::util::recursive::{self, Recurse};
 use risingwave_pb::batch_plan::PlanNode as PbBatchPlan;
 use risingwave_pb::stream_plan::StreamNode as PbStreamPlan;
 use serde::Serialize;
-use smallvec::SmallVec;
 
 use self::batch::BatchPlanRef;
 use self::generic::{GenericPlanRef, PhysicalPlanRef};
@@ -593,12 +592,6 @@ impl PredicatePushdown for PlanRef {
 }
 
 impl PlanRef {
-    pub fn inputs(&self) -> SmallVec<[PlanRef; 2]> {
-        // Dispatch to dyn PlanNode instead of PlanRef.
-        let dyn_t = self.deref();
-        dyn_t.inputs()
-    }
-
     pub fn clone_root_with_inputs<C: ConventionMarker>(&self, inputs: &[PlanRef]) -> PlanRef {
         self.expect_convention::<C>();
         if let Some(share) = self.as_share_node::<C>() {


wenym1:
* **#22700** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22700?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#22724** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22724?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#22712** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22712?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22712?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

hzxa21:
Any context on why do we need this refactoring?

src/frontend/src/optimizer/plan_node/logical_share.rs
```diff
@@ -139,9 +144,8 @@ impl ToStream for LogicalShare {
         match ctx.get_to_stream_result(self.id()) {
             None => {
                 let new_input = self.input().to_stream(ctx)?;
-                let new_logical = self.core.clone();
-                new_logical.replace_input(new_input);
-                let stream_share_ref: PlanRef = StreamShare::new(new_logical).into();
+                let core = generic::Share::new(new_input);
```
wenym1, line 147: We used to directly do `replace_input` on the core. However, this is actually not type-safe, because `self`, which is `LogicalShare`, should be in `Logical` convention, while the `new_input` is after `to_stream` and should be in `Stream` convention. Therefore, here we instead create a new core to hold the `new_input`.

PR #22721, opened 2025-07-24
fix(frontend): Fix alter table column project node under sink-into-table

Chiro11:

- Should match column id instead of data type.
- When not match, the default value of the target column should be inserted, instead of null.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/handler/alter_table_column.rs
+++ b/src/frontend/src/handler/alter_table_column.rs
@@ -152,16 +152,22 @@ pub(crate) fn hijack_merger_for_target_table(
     }
 
     let mut exprs = Vec::with_capacity(target_columns.len());
-    let sink_col_idx_by_name = sink_columns
+    let sink_idx_by_col_id = sink_columns
         .iter()
         .enumerate()
-        .map(|(idx, col)| (col.name(), idx))
+        .map(|(idx, col)| (col.column_id(), idx))
         .collect::<HashMap<_, _>>();
     let default_column_exprs = TableCatalog::default_column_exprs(target_columns);
     for (target_idx, target_col) in target_columns.iter().enumerate() {
-        if let Some(idx) = sink_col_idx_by_name.get(target_col.name())
-            && target_col.data_type() == sink_columns[*idx].data_type()
-        {
+        if let Some(idx) = sink_idx_by_col_id.get(&target_col.column_id()) {
+            assert_eq!(
+                target_col.data_type(),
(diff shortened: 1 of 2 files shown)


src/frontend/src/handler/alter_table_column.rs
```diff
@@ -151,26 +151,25 @@ pub(crate) fn hijack_merger_for_target_table(
         sink_columns = target_columns.to_vec();
     }
 
-    let mut i = 0;
-    let mut j = 0;
-    let mut exprs = Vec::new();
-
-    while j < target_columns.len() {
-        if i < sink_columns.len() && sink_columns[i].data_type() == target_columns[j].data_type() {
+    let mut exprs = Vec::with_capacity(target_columns.len());
```
wenym1, line 154: In `ColumnCatalog`, we have `ColumnId` to uniquely identify a column during alter table. Shall we use the `ColumnId` instead? After we match the column by column id, we can further assert that the two data types are the same.

src/frontend/src/handler/alter_table_column.rs
```diff
@@ -151,26 +151,31 @@ pub(crate) fn hijack_merger_for_target_table(
...
+        if let Some(idx) = sink_idx_by_col_id.get(&target_col.column_id()) {
+            assert_eq!(
+                target_col.data_type(),
+                sink_columns[*idx].data_type(),
+                "data type mismatch for column {}: {} vs {}",
+                target_col.name(),
+                target_col.data_type(),
+                sink_columns[*idx].data_type()
+            );
+            // If the sink has the corresponding column name, use the sink's data.
```
wenym1: The comment is stale.

src/frontend/src/handler/alter_table_column.rs
```diff
@@ -151,26 +151,31 @@ pub(crate) fn hijack_merger_for_target_table(
         sink_columns = target_columns.to_vec();
```
wenym1, line 151: Not related to this PR. It seems that we should use the original table columns here instead of the `target_columns`? cc @shanicky 

For example, the table original had two columns `[a, b]`, and when we create a sink into the table, the sink should have column `[a, b]`. When we alter table add a column `c`, the `target_columns` should be `[a, b, c]`, but the `sink_columns` should be `[a, b]`.

Chiro11: According to the comments, this seems to be for compatibility with older versions.

PR #22724, opened 2025-07-25
refactor(frontend): separate PlanVisitor of different conventions

wenym1:

Previously, in `PlanVisitor` trait, we include the `visit_$convention_$variant_name` method of all conventions in the trait. However, every time we visit a plan, its convention is fixed. In this PR, we will divide the trait into 3 `$ConventionPlanVisitor` traits, with each trait only has the `visit_$convention_$variant_name` of its own plan nodes, so that we can avoid calling the `visit_xxx` method with the wrong convention. We will also have a `PlanVisitor<C: ConventionMarker>` trait, with a method `fn visit(&mut self, plan: PlanRef) -> Self::Result;`.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/optimizer/plan_visitor/distributed_dml_visitor.rs
+++ b/src/frontend/src/optimizer/plan_visitor/distributed_dml_visitor.rs
@@ -16,10 +16,10 @@ use std::rc::Rc;
 
 use risingwave_connector::source::ConnectorProperties;
 
-use super::{DefaultBehavior, Merge};
+use super::{BatchPlanVisitor, DefaultBehavior, Merge};
 use crate::PlanRef;
 use crate::catalog::source_catalog::SourceCatalog;
-use crate::optimizer::plan_node::{BatchSource, LogicalSource, StreamSource};
+use crate::optimizer::plan_node::BatchSource;
 use crate::optimizer::plan_visitor::PlanVisitor;
 
 #[derive(Debug, Clone, Default)]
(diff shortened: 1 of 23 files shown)


wenym1:
* **#22700** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22700?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#22724** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22724?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22724?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#22712** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22712?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/frontend/src/optimizer/plan_visitor/distributed_dml_visitor.rs
```diff
@@ -65,20 +65,4 @@ impl PlanVisitor for DistributedDmlVisitor {
             false
         }
     }
-
-    fn visit_logical_source(&mut self, logical_source: &LogicalSource) -> bool {
```
BugenZhao, line 69: Does it mean that it's actually unreachable?

BugenZhao (approved):

src/frontend/src/optimizer/plan_visitor/distributed_dml_visitor.rs
wenym1: Yes. This PR is actually part of a later PR https://github.com/risingwavelabs/risingwave/pull/22700 that has static convention type in `PlanRef`. The `DistributedDmlVisitor` is only applied on `BatchPlanRef`.

Issue #19343, opened 2024-11-12
Tracking: time-series related functions


lmatz:
### Downsampling (with gap-filling)

https://questdb.io/docs/reference/sql/sample-by/#fill-options
https://docs.timescale.com/api/latest/hyperfunctions/downsampling/
https://www.timescale.com/blog/how-to-proactively-manage-long-term-data-storage-with-downsampling/
https://docs.timescale.com/api/latest/hyperfunctions/gapfilling/time_bucket_gapfill/

three common filling strategies: 

- [ ] Use `locf` to carry the last value forward
- [ ]  Use `interpolate` to linearly interpolate the value 
- [ ] Fill in NULL by default

### Agg Helper Functions

- [ ] Candlestick: https://docs.timescale.com/api/latest/hyperfunctions/financial-analysis/candlestick_agg/#candlestick

- [ ] Rollup. This function must work with an aggregation function. 
    1. For example, if we aggregate per minute in an MV, we can reuse the results and further aggregate the values per hour in a second MV on MV. However, “reusing” is not always straightforward. “sum” and “max” are easy, but what if it is “average” and other approximated statistical aggregations? It would be hard for users to write them directly in SQL.
    2. Rollup can work with the ”candlestick_agg” too.

### Mathematics and Statistics

- [ ] weighted moving average and more generally as covered by InFluxDB’s “technical analysis” functions: https://docs.influxdata.com/enterprise_influxdb/v1/query_language/functions/#technical-analysis
        
    1. CHANDE_MOMENTUM_OSCILLATOR
    2. RELATIVE_STRENGTH_INDEX

- [ ] one-variable statistical aggregation. The following ones are used often in risk calculation, e.g. tail risk and VAR risk that need to be monitored in real-time. https://docs.timescale.com/api/latest/hyperfunctions/statistical-and-regression-analysis/stats_agg-one-variable/

    1. kurtosis
    3. skewness
    4. stdev

- [ ] two-variable statistical aggregations. https://docs.timescale.com/api/latest/hyperfunctions/statistical-and-regression-analysis/stats_agg-two-variables/. Similar use cases as above. Just highlight a few common ones:

    1. correlation
    2. covariance
    3. regr_slope and regr_intercept

fuyufjh:
`ROLLUP` is supported https://github.com/risingwavelabs/risingwave/pull/11119

fuyufjh:
The left part is

1. Gap-filling aggregation. Can be done either by
[...]

PR #22731, opened 2025-07-28
feat: eowc gap_fill executor

guluguluhhhh:

#19343 
This PR introduces the EowcGapFillExecutor, a stream executor that fills gaps in time-series data. It operates by buffering rows and triggering the fill logic upon receiving a watermark.
### Execution Logic

The executor's logic is driven by watermarks on a specified time column.

* Buffering & Processing: Incoming rows are buffered in a state-backed SortBuffer. When a watermark W arrives, all rows with timestamps ts <= W are processed.

* Gap Filling: For each consecutive pair of rows (prev_row, curr_row), the executor generates new rows at a fixed gap_interval.

* Fill Strategies: Values for new rows are populated using a defined strategy per column: Interpolate (linear interpolation), Locf (last observation carried forward), or Null.
### State and Fault Tolerance
State is managed via two state tables, ensuring full fault tolerance:

* buffer_table: Persists the SortBuffer, storing rows that are waiting for a watermark.

* prev_row_table: A single-row table storing the last emitted row of an epoch. This is crucial for correctly filling gaps across checkpoints.

On recovery, the executor restores its state from these tables, allowing processing to resume seamlessly with exactly-once guarantees.

## Checklist

## Documentation

<details>
support EOWC gap_fill

</details>

### DIFF
--- a/src/stream/src/executor/eowc/eowc_gap_fill.rs
+++ b/src/stream/src/executor/eowc/eowc_gap_fill.rs
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::collections::HashMap;
+
 use risingwave_common::array::Op;
 use risingwave_common::metrics::LabelGuardedIntCounter;
 use risingwave_common::row::OwnedRow;
@@ -40,7 +42,7 @@ pub struct EowcGapFillExecutorArgs<S: StateStore> {
     pub prev_row_table: StateTable<S>,
     pub chunk_size: usize,
     pub time_column_index: usize,
-    pub fill_columns: Vec<(usize, FillStrategy)>,
+    pub fill_columns: HashMap<usize, FillStrategy>,
     pub gap_interval: NonStrictExpression,
 }
 
@@ -56,7 +58,7 @@ struct ExecutorInner<S: StateStore> {
     prev_row_table: StateTable<S>,
     chunk_size: usize,
     time_column_index: usize,
-    fill_columns: Vec<(usize, FillStrategy)>,
+    fill_columns: HashMap<usize, FillStrategy>,
     gap_interval: NonStrictExpression,
 
     // Metrics
@@ -116,7 +118,7 @@ impl<S: StateStore> ExecutorInner<S> {
         prev_row: &OwnedRow,
         curr_row: &OwnedRow,
         time_column_index: usize,
-        fill_columns: &[(usize, FillStrategy)],
+        fill_columns: &HashMap<usize, FillStrategy>,
         interval: risingwave_common::types::Interval,
         metrics: &GapFillMetrics,
     ) -> Result<Vec<OwnedRow>, ExprError> {
@@ -197,7 +199,7 @@ impl<S: StateStore> ExecutorInner<S> {
         let mut interpolation_states: Vec<Datum> = Vec::new();
 
         for i in 0..prev_row.len() {
-            if let Some((_, strategy)) = fill_columns.iter().find(|(col, _)| *col == i) {
+            if let Some(strategy) = fill_columns.get(&i) {
                 if matches!(strategy, FillStrategy::Interpolate) {
                     let step = Self::calculate_step(
                         prev_row.datum_at(i),
@@ -234,9 +236,7 @@ impl<S: StateStore> ExecutorInner<S> {
                         _ => unreachable!("Time column should be Timestamp or Timestamptz"),
                     };
                     Some(fill_time_scalar)
-                } else if let Some((_, strategy)) =
-                    fill_columns.iter().find(|(col, _)| *col == col_idx)
-                {
+                } else if let Some(strategy) = fill_columns.get(&col_idx) {
                     // Apply the fill strategy for this column
                     match strategy {
                         FillStrategy::Locf => prev_row.datum_at(col_idx).to_owned_datum(),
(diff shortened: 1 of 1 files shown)


src/stream/src/executor/eowc/eowc_gap_fill.rs
```diff
@@ -0,0 +1,881 @@
...
+                chunk_size: args.chunk_size,
+                time_column_index: args.time_column_index,
+                fill_columns: args.fill_columns,
+                gap_interval: args.gap_interval,
+            },
+        }
+    }
+
+    #[try_stream(ok = Message, error = StreamExecutorError)]
+    async fn executor_inner(self) {
```
kwannoel:
```suggestion
    async fn execute_inner(self) {
```

src/stream/src/executor/eowc/eowc_gap_fill.rs
```diff
@@ -0,0 +1,881 @@
...
+            };
+        }
+
+        Ok(filled_rows)
+    }
+}
+
+impl<S: StateStore> Execute for EowcGapFillExecutor<S> {
+    fn execute(self: Box<Self>) -> BoxedMessageStream {
+        self.executor_inner().boxed()
```
kwannoel: @copilot 
```suggestion
        self.execute_inner().boxed()
```

kwannoel:
Rest LGTM

src/stream/src/executor/eowc/eowc_gap_fill.rs
```diff
@@ -0,0 +1,906 @@
...
+                    match strategy {
+                        FillStrategy::Locf => prev_row.datum_at(col_idx).to_owned_datum(),
+                        FillStrategy::Null => None,
+                        FillStrategy::Interpolate => {
+                            // Apply interpolation step and update cumulative value
+                            if let Some(step) = &interpolation_steps[col_idx] {
+                                Self::apply_step(&mut interpolation_states[col_idx], step);
+                                interpolation_states[col_idx].clone()
+                            } else {
+                                prev_row.datum_at(col_idx).to_owned_datum()
```
yuhao-su: Does interpolation require repeating the prev_row if calculate_step returns None?

src/stream/src/executor/eowc/eowc_gap_fill.rs
```diff
@@ -0,0 +1,906 @@
...
+    fn calculate_step(d1: DatumRef<'_>, d2: DatumRef<'_>, steps: usize) -> Datum {
+        let (Some(s1), Some(s2)) = (d1, d2) else {
+            return None;
+        };
+        if steps == 0 {
+            return None;
+        }
+        match (s1, s2) {
+            (ScalarRefImpl::Int16(v1), ScalarRefImpl::Int16(v2)) => {
+                Some(ScalarImpl::Int16((v2 - v1) / steps as i16))
```
yuhao-su, line 82: Will the v2-v1 < 0 in any cases?

yuhao-su:
I'm a little concerned about the large gap. But we can leave this to the future.

src/stream/src/executor/eowc/eowc_gap_fill.rs
guluguluhhhh: In my opinion, yes. Basically, this computes the value of a column in the current row minus the value in the previous row, so it’s possible for the result to be negative.

src/stream/src/executor/eowc/eowc_gap_fill.rs
guluguluhhhh: When testing on Tiger Data, it returned NULL. After investigation, NULL as the default is standard practice, but my current code defaults to locf. I‘ll to change it to NULL.

kwannoel:
> I'm a little concerned about the large gap. But we can leave this to the future.

What do you mean by "large gap"

yuhao-su:
> > I'm a little concerned about the large gap. But we can leave this to the future.
> 
> What do you mean by "large gap"

A large gap between prev and current timestamp that can result in large number of filled records.

src/stream/src/executor/eowc/eowc_gap_fill.rs
yuhao-su: But in EOWC mode the value should be sorted and v2 > v1 must always hold?

guluguluhhhh: In time column, yes. Since all rows are sorted by the time column. But in other columns we can't ensure that and we don't need to, I think.

yuhao-su: I would suggest using  `div_floor` instead of `/` for interpolation, as mathematically it's more accurate. cc @kwannoel, what do you think?

kwannoel:
> > > I'm a little concerned about the large gap. But we can leave this to the future.
> > 
> > 
> > What do you mean by "large gap"
> 
> A large gap between prev and current timestamp that can result in large number of filled records.

+1, hence I request metrics. I think this can be a real problem, and users should be using EOWC in many cases, or maybe append-only.

In other cases, maybe batch gapfill is more appropriate.

src/stream/src/executor/eowc/eowc_gap_fill.rs
kwannoel: > as mathematically it's more accurate

Do you mean it's nearer to the original float? If so `div_floor` will not satisfy that for positive integers, and we are looking for something like `div_round` which doesn't exist.

Default to `/` seems acceptable, if user wants rounding, or some other way of fitting, they can cast their int to floats.

yuhao-su: > Do you mean it's nearer to the original float?

no. I shouldn't use the word _accurate_ which is misleading. 

What I meant is for interpolation, mathematically it's more natural to round a float by either taking the floor or ceil. By using `/` we are actually taking floor of positive number and ceil for negetive.

PR #22736, opened 2025-07-28
feat(expr): support vector `+`/`-`/`*` vector

xiangjinwu:

Beside the trivial implementation of `+`/`-`/`*` element-wise, this PR also enhances
* Type inference for returning vector
  * Old: same size as 1st vector argument
  * New: no default hidden in macro; each expr need to define their own `type_infer` rule
* Constructing `VectorVal` from a list of f32
  * Old: infallible and accepts whatever values passed in
  * New: rejects nan/inf/-inf

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/common/src/array/vector_array.rs
+++ b/src/common/src/array/vector_array.rs
@@ -217,7 +217,9 @@ impl VectorVal {
     /// This is leak of implementation. Prefer [`VectorVal::from_iter`] below.
     pub fn from_inner(inner: ListValue) -> Result<Self, ArrayError> {
         for element in inner.iter() {
-            let Some(scalar) = element else { continue };
+            let Some(scalar) = element else {
+                return Err(ArrayError::internal("NULL not allowed in vector"));
+            };
             let ScalarRefImpl::Float32(val) = scalar else {
                 return Err(ArrayError::internal(format!(
                     "vector element must be f32 but found {scalar:?}"


src/expr/macro/src/gen.rs
```diff
@@ -115,12 +115,6 @@ impl FunctionAttr {
                 // infer as the type of "anymap" argument
                 return Ok(quote! { |args| Ok(args[#i].clone()) });
             }
-        } else if self.ret == "vector" {
-            if let Some(i) = self.args.iter().position(|t| t == "vector") {
-                // infer as the type of "vector" argument
-                // Example usage: `last_value(*) -> auto`
-                return Ok(quote! { |args| Ok(args[#i].clone()) });
-            }
```
xiangjinwu, line 123: The hidden special rules for array/struct/map are better removed and made explicit as well.

src/common/src/array/vector_array.rs
```diff
@@ -214,24 +215,41 @@ impl VectorVal {
     /// Create a new vector from inner [`ListValue`].
     ///
     /// This is leak of implementation. Prefer [`VectorVal::from_iter`] below.
-    pub fn from_inner(inner: ListValue) -> Self {
-        Self { inner }
+    pub fn from_inner(inner: ListValue) -> Result<Self, ArrayError> {
+        for element in inner.iter() {
+            let Some(scalar) = element else { continue };
```
wenym1: Should we return err when seeing `None`?

wenym1 (approved):
LGTM.

In linear algebra, we may also have usage like `3 * [1, 2, 3]`. We may also support it in future PR.

src/common/src/array/vector_array.rs
xiangjinwu: Good catch. Yes we shall reject `null` similar to `nan/inf/-inf` below. Let me fix it.

However I would not be able to add an e2e test for this. This is effectively unreachable right now and is intended to prevent future mistakes - just like I only find the need to reject `inf` when implementing `+`. There is currently no operation that results in `null` element in a `vector` - casting from text and array already checks this case.

xiangjinwu (commented):
> In linear algebra, we may also have usage like `3 * [1, 2, 3]`. We may also support it in future PR.

Yes I thought about them as well, when I found there's no division in pgvector but a scalar division `[1,2,3] / 3` should be possible (and surprised that `*` is element-wise).  This is similar to `interval`:
* `float8 * interval -> interval`
* `interval * float8 -> interval`
* `interval / float8 -> interval`

PR #22739, opened 2025-07-29
feat: support webhook sink connector

lin0303-siyuan:

This pr adds feature for the webhook sink. Webhook Sink let users send data to a downstream endpoint via standard HTTP post request. A new sink connector is added.

- Summarize your change (**mandatory**)
Add a new sink connector (Webhook).
- How does this PR work? Need a brief introduction for the changed logic (optional)
- Refer to a related PR or issue link (optional)
[webhook sink issue](https://github.com/risingwavelabs/risingwave/issues/18606)

## Checklist

## Documentation

<details>

</details>

A simple example would be:
```SQL
CREATE TABLE orders (
    order_id INT,
    customer_id INT,
    total_price DECIMAL
);

CREATE MATERIALIZED VIEW orders_text_view AS
SELECT
    CONCAT('order_id=', order_id, ', customer_id=', customer_id, ', total_price=', total_price) AS "text"
FROM orders;

CREATE SINK webhook_sink
FROM orders_text_view
WITH (
    connector = 'webhook',
    endpoint = '[slack webhook endpoint[',
    header = 'content-type: application/json',
    format = 'json',
    type = 'append-only',
    force_append_only='true'
);
INSERT INTO orders VALUES (1, 42, 199.99);
```

### DIFF
--- a/src/connector/src/sink/encoder/json.rs
+++ b/src/connector/src/sink/encoder/json.rs
@@ -77,6 +77,23 @@ impl JsonEncoder {
         }
     }
 
+    pub fn new_with_webhook(schema: Schema, col_indices: Option<Vec<usize>>) -> Self {
+        let config = JsonEncoderConfig {
+            time_handling_mode: TimeHandlingMode::Milli,
+            date_handling_mode: DateHandlingMode::FromCe,
+            timestamp_handling_mode: TimestampHandlingMode::String,
+            timestamptz_handling_mode: TimestamptzHandlingMode::UtcString,
+            custom_json_type: CustomJsonType::None,
+            jsonb_handling_mode: JsonbHandlingMode::String,
+        };
+        Self {
+            schema,
+            col_indices,
+            kafka_connect: None,
+            config,
+        }
+    }
+
     pub fn new_with_es(schema: Schema, col_indices: Option<Vec<usize>>) -> Self {
         let config = JsonEncoderConfig {
             time_handling_mode: TimeHandlingMode::String,
--- a/src/connector/src/sink/webhook.rs
+++ b/src/connector/src/sink/webhook.rs
@@ -49,23 +49,16 @@ impl EnforceSecret for WebhookConfig {
 pub struct WebhookSink {
     pub config: WebhookConfig,
     schema: Schema,
-    pk_indices: Vec<usize>,
     is_append_only: bool,
 }
 
 impl EnforceSecret for WebhookSink {}
 
 impl WebhookSink {
-    pub fn new(
-        config: WebhookConfig,
-        schema: Schema,
-        pk_indices: Vec<usize>,
-        is_append_only: bool,
-    ) -> Result<Self> {
+    pub fn new(config: WebhookConfig, schema: Schema, is_append_only: bool) -> Result<Self> {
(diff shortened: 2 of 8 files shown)


src/connector/src/sink/encoder/json.rs
```diff
@@ -77,6 +77,23 @@ impl JsonEncoder {
...
     }
 
+    pub fn new_with_webhook(schema: Schema, col_indices: Option<Vec<usize>>) -> Self {
+        let config = JsonEncoderConfig {
+            time_handling_mode: TimeHandlingMode::Milli,
+            date_handling_mode: DateHandlingMode::FromCe,
+            timestamp_handling_mode: TimestampHandlingMode::String,
+            timestamptz_handling_mode: TimestamptzHandlingMode::UtcString,
+            custom_json_type: CustomJsonType::None,
+            jsonb_handling_mode: JsonbHandlingMode::String,
```
xiangjinwu, line 87: Could you give the rational behind these default choices?

src/connector/src/sink/webhook.rs
```diff
@@ -0,0 +1,238 @@
...
+use phf::{Set, phf_set};
+use reqwest::header::{HeaderMap, HeaderName, HeaderValue};
+use with_options::WithOptions;
+
+#[serde_as]
+#[derive(Clone, Debug, Deserialize, WithOptions)]
+pub struct WebhookConfig {
+    pub endpoint: String,
+    pub headers: Option<String>,
+}
```
tabVersion, line 47: at least it should contain Bearer token for auth, and be enforce it as secret

tabVersion:
please also add some tests

src/connector/src/sink/encoder/json.rs
lin0303-siyuan: I copied some of the parameters from other connectors (doris, startrocks), I am actually confused about what should be the correct behavior for webhook connector.

lmatz: I feel like `string` mode makes the most sense because the rest options lose some information.

As a webhook sink typically connects to a saas app that demonstrates the data as text on some GUI, without the need to ingest into some other systems that can only accept a certain format of data, we should retain as much info as possible

src/connector/src/sink/webhook.rs
lmatz: is it attaching one line in the header
```
Authorization: Bearer XXXXXXXXXXXXXXXXXXXXXX
```

lmatz:
> please also add some tests

cc: @lin0303-siyuan  
can mimic what sink tests under the folder https://github.com/risingwavelabs/risingwave/tree/main/ci/scripts do
e.g. a python script that acts as http server, accepts http request and verifies what it receives from the webhook sink.

WDYT? cc: @tabVersion

src/connector/src/sink/encoder/json.rs
xiangjinwu: Suggest to use the following as default for webhook:
* time/date/timestamp: String
* timestamptz: UtcString
* jsonb: Dynamic (to avoid escaping; users can always `::varchar` if escaping is preferred)

PR #22745, opened 2025-07-30
feat(expr): `vector || vector` concatenation

xiangjinwu:

Support vector concatenation: `vector || vector -> vector`

Caveat:
Since we require vector types to have a statically defined size, we cannot infer it for `null` or extended mode parameters `$1`.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/common/src/types/mod.rs
+++ b/src/common/src/types/mod.rs
@@ -391,6 +391,8 @@ pub mod data_types {
 }
 
 impl DataType {
+    pub const VEC_MAX_SIZE: usize = 16000;
+
     pub fn create_array_builder(&self, capacity: usize) -> ArrayBuilderImpl {
         use crate::array::*;
 
--- a/proto/expr.proto
+++ b/proto/expr.proto
@@ -312,6 +312,7 @@ message ExprNode {
     COSINE_DISTANCE = 751;
     L1_DISTANCE = 752;
     INNER_PRODUCT = 753;
+    VEC_CONCAT = 754;
 
     // Internal function for schema change
     COMPOSITE_CAST = 800;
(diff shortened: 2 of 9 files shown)


src/common/src/types/mod.rs
```diff
@@ -391,6 +391,8 @@ pub mod data_types {
 }
 
 impl DataType {
+    pub const VEC_MAX_SIZE: usize = 16000;
```
wenym1, line 394: How do we choose this value as the max dimension? And shall it be the power of 2, such as 16384? Should we also forbidden defining a vector type with dimension greater than this value instead of only limiting it during vector concat?

wenym1 (approved):
Rest LGTM

src/common/src/types/mod.rs
xiangjinwu: > How do we choose this value as the max dimension? And shall it be the power of 2, such as 16384?

It is from [pgvector](https://github.com/pgvector/pgvector/blob/v0.8.0/README.md#vector-type:~:text=Vectors%20can%20have%20up%20to-,16%2C000,-dimensions.). Let me add it to the doc comment of this const. And we may consider updating it to a different value based on our needs (i.e. I am not aware of pgvector's rational of choosing this magic number).

> Should we also forbidden defining a vector type with dimension greater than this value instead of only limiting it during vector concat?

Done in sqlparser already:
https://github.com/risingwavelabs/risingwave/blob/75c40d356278eff6b3b5331d096c52ef87ad4d0d/src/sqlparser/src/parser_v2/data_type.rs#L247

wenym1: Shall we unify to use the same const value?

xiangjinwu: After a second thought I think this restriction shall be moved from sqlparser to binder and reuse the same const. Will do in this same PR.

Issue #22717, opened 2025-07-24
`force_append_only` sink still buffer before emitting output to external downstream


lmatz:
### Describe the bug

https://github.com/risingwavelabs/risingwave/blob/main/src/stream/src/executor/sink.rs#L159-L160

what I have is a simple aggregation sink
```
CREATE SINK IF NOT EXISTS token_balances_sink
AS
SELECT
    token_address,
    holder_address,
    chain_id,
    MAX(timestamp) AS latest_updated_timestamp
FROM balance_changes_source
GROUP BY token_address, holder_address, chain_id
WITH (
    connector = 'kafka',
    topic = 'token-balances-sink',
    properties.bootstrap.server = 'message_queue:29092'
)
FORMAT PLAIN ENCODE JSON (
    force_append_only = 'true'
);
```

without specifying the primary key, the `need_advance_delete` in the code will be true and thus sink executors will buffer records until barrier arrives.

only after specifying ```primary_key = 'token_address,holder_address,chain_id',```, it will emit immediately.

### Error message/log

```text

```

PR #22747, opened 2025-07-30
fix(sink): avoid advance delete for force_append_only sink

tabVersion:

close https://github.com/risingwavelabs/risingwave/issues/22717

advance delete will hold a chunk wait for barrier to handle if the sink's primary key does not match upstream, this does not make sense for force_append_only. As we already skip the logic for AppendOnly sink

## Checklist

## Documentation

<details>

</details>


wenym1 (approved):
LGTM

In `process_msg`, we have a branch `if need_advance_delete { ... }`. In the for loop in the branch, we have a `if sink_type != SinkType::ForceAppendOnly {`. After this PR, this condition is unlikely to hit. Shall we remove the check and always yield the delete chunk?

lmatz (approved):

PR #22748, opened 2025-07-30
fix(binder): correctly bind subquery in SQL UDF

BugenZhao:

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/binder/bind_context.rs
+++ b/src/frontend/src/binder/bind_context.rs
@@ -25,6 +25,7 @@ use risingwave_sqlparser::ast::{TableAlias, WindowSpec};
 
 use crate::binder::Relation;
 use crate::error::{ErrorCode, Result};
+use crate::expr::ExprImpl;
 
 type LiteResult<T> = std::result::Result<T, ErrorCode>;
 
@@ -162,6 +163,9 @@ pub struct BindContext {
     pub disable_security_invoker: bool,
     /// Named window definitions from the `WINDOW` clause
     pub named_windows: HashMap<String, WindowSpec>,
+    /// Bound arguments for the current SQL UDF.
+    // TODO: use enum for named or positional arguments
+    pub udf_arguments: HashMap<String, ExprImpl>,
 }
 
 /// Holds the context for the `BindContext`'s `ColumnGroup`s.
--- a/src/frontend/src/binder/expr/function/mod.rs
+++ b/src/frontend/src/binder/expr/function/mod.rs
@@ -669,40 +669,34 @@ impl Binder {
         }
 
         // This represents the current user defined function is `language sql`
-        let parse_result =
-            risingwave_sqlparser::parser::Parser::parse_sql(func.body.as_ref().unwrap().as_str());
-        if let Err(ParserError::ParserError(err)) | Err(ParserError::TokenizerError(err)) =
-            parse_result
-        {
-            // Here we just return the original parse error message
-            return Err(ErrorCode::InvalidInputSyntax(err).into());
-        }
+        let ast = match risingwave_sqlparser::parser::Parser::parse_sql(
+            func.body.as_ref().unwrap().as_str(),
+        ) {
+            Ok(ast) => ast,
+            Err(ParserError::ParserError(err) | ParserError::TokenizerError(err)) => {
+                // Here we just return the original parse error message
+                return Err(ErrorCode::InvalidInputSyntax(err).into());
+            }
+        };
 
-        debug_assert!(parse_result.is_ok());
-
-        // We can safely unwrap here
-        let ast = parse_result.unwrap();
-
-        // Stash the current `udf_context`
-        // Note that the `udf_context` may be empty,
-        // if the current binding is the root (top-most) sql udf.
-        // In this case the empty context will be stashed
-        // and restored later, no need to maintain other flags.
-        let stashed_udf_context = self.udf_context.get_context();
+        // Stash the current arguments.
+        let stashed_udf_arguments = std::mem::take(&mut self.context.udf_arguments);
 
         // The actual inline logic for sql udf
         // Note that we will always create new udf context for each sql udf
-        let mut udf_context = HashMap::new();
         for (i, arg) in args.into_iter().enumerate() {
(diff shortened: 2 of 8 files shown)


BugenZhao:
* **#22778** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22778?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#22748** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22748?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22748?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#22779** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22779?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/frontend/src/error.rs
```diff
@@ -275,3 +275,19 @@ impl From<RwError> for tonic::Status {
...
+    /// giving a more concise error message.
+    pub fn bind_root_removed(self) -> Self {
+        let inner = self.into_inner();
+        if let ErrorCode::BindErrorRoot { expr: _, error } = inner {
+            match error.downcast::<RwError>() {
+                Ok(error) => error.bind_root_removed(),
+                Err(error) => anyhow::anyhow!(error).into(),
+            }
+        } else {
+            inner.into()
```
xiangjinwu, line 290: Looks like a recursion that ends on either:
* matching non-`BindErrorRoot`
* downcast to non-`RwError`

src/frontend/src/binder/mod.rs
```diff
@@ -485,6 +385,20 @@ impl Binder {
...
+    /// In most cases, this should include all the upper subquery contexts. However, when binding
+    /// SQL UDFs, we should avoid resolving the context outside the UDF for hygiene.
+    fn visible_upper_subquery_contexts_rev(
+        &self,
+    ) -> impl Iterator<Item = &(BindContext, Vec<LateralBindContext>)> + '_ {
+        self.upper_subquery_contexts
+            .iter()
+            .rev()
+            .take_while(|(context, _)| context.udf_arguments.is_none())
+    }
```
xiangjinwu: The `visible` in this name may be confused with a similar concept:
https://github.com/risingwavelabs/risingwave/blob/e95a79d704d00950c7efd12a5567a998cc57147d/src/frontend/src/binder/bind_context.rs#L68-L73

Let's clarify in the doc comment, that one checks ancestors while the other checks siblings.

src/frontend/src/binder/expr/function/mod.rs
```diff
@@ -657,85 +670,101 @@ impl Binder {
...
+        let ast = Parser::parse_sql(body)?;
 
-        // Stash the current `udf_context`
-        // Note that the `udf_context` may be empty,
-        // if the current binding is the root (top-most) sql udf.
-        // In this case the empty context will be stashed
-        // and restored later, no need to maintain other flags.
-        let stashed_udf_context = self.udf_context.get_context();
+        // Stash the current arguments.
+        let stashed_udf_arguments = self.context.udf_arguments.take();
```
xiangjinwu: When is this necessary? In which case there would be 2 udfs active at the same layer? `foo(x + bar(y + baz(z)))`?

src/frontend/src/handler/create_sql_function.rs
```diff
@@ -228,94 +203,66 @@ pub async fn handle_create_sql_function(
...
-                            "\nreturn type mismatch detected\nexpected: [{}]\nactual: [{}]\nplease adjust your function definition accordingly",
-                            return_type,
-                            expr.return_type()
-                        ))
-                        .into());
-                    }
+        let args = arg_types
+            .iter()
+            .map(|ty| Literal::new(None, ty.clone()).into() /* NULL */)
+            .collect();
```
chenzl25, line 214: These args for SQL UDF are dummy, right?

src/frontend/src/binder/bind_context.rs
```diff
@@ -162,6 +163,9 @@ pub struct BindContext {
     pub disable_security_invoker: bool,
     /// Named window definitions from the `WINDOW` clause
     pub named_windows: HashMap<String, WindowSpec>,
+    /// Bound arguments for the current SQL UDF, if any.
+    // TODO: use enum for named or positional arguments
+    pub udf_arguments: Option<HashMap<String, ExprImpl>>,
```
chenzl25: IIUC, it is used only by sql udf arguments only, not by any other udfs.

src/frontend/src/binder/mod.rs
```diff
@@ -519,8 +433,8 @@ impl Binder {
         self.context.clause = clause;
     }
 
-    pub fn udf_context_mut(&mut self) -> &mut UdfContext {
-        &mut self.udf_context
+    pub fn set_mock_udf_arguments(&mut self, udf_arguments: HashMap<String, ExprImpl>) {
+        self.context.udf_arguments = Some(udf_arguments);
```
xiangjinwu: dead code?

src/frontend/src/error.rs
BugenZhao: Yes. This will be removed in the next PR #22778

src/frontend/src/binder/expr/function/mod.rs
BugenZhao: > In which case there would be 2 udfs active at the same layer? `foo(x + bar(y + baz(z)))`?

Yes. Updated in the comments.

src/frontend/src/handler/create_sql_function.rs
BugenZhao: Yes, as shown in the comments:

```
// Try bind the function call with mock arguments.
```

src/frontend/src/binder/bind_context.rs
BugenZhao: Yes, renamed to `sql_udf_arguments`.

Issue #8831, opened 2023-03-28
Tracking: string functions


xiangjinwu:
### Basic

* [x] `bit_length` / `char_length` / `character_length` / `octet_length` / `length`
* [x] `position` / `strpos`
* [x] `||` / `concat` / `concat_ws`
* [x] `lower` / `upper` / `initcap`
* [x] `overlay(string PLACING newsubstring FROM start [FOR count])`
* [ ] `substring` / `substr` / `left` / `right` (missing `left` / `right`)
* [ ] `trim` / `btrim` / `ltrim` / `rtrim` (missing `btrim`)
* [x] `lpad` / `rpad`
* [x] `repeat`
* [x] `reverse`
* [x] `replace`
* [ ] `translate` (missing frontend part)
* [x] `split_part`
* [ ] `starts_with` / `^@` (missing `^@`)
* [ ] `string_to_array`

### Misc

* [x] `ascii` / `chr`
* [x] `to_hex`
* `quote_ident` / `quote_literal` / `quote_nullable` / `parse_ident`
* [x] `md5`

### RegExp

* `regexp_like`
* `regexp_count`
* `regexp_instr`
* `regexp_substr`
* [x] `regexp_match`
* `regexp_replace`
* `regexp_split_to_array`

### Formating

* `format`

### Encoding related

* `pg_client_encoding`
* `normalize` / `is normalized`
* [x] `to_ascii`
* `unistr`

### Aggregate Function

* [x] `string_agg`

### Set Returning Function

* `string_to_table`
* [x] `regexp_matches`
* `regexp_split_to_table`

BanBuDu0:
I finished encdec some time ago.
I think I will keep contributing all of the bytea functions one by one, if no one is doing the same work.

PR #22762, opened 2025-08-01
feat(frontend): support `trim`, `ltrim`, `rtrim` function on bytea type

Standing-Man:

- Support `trim`, `ltrim`, `rtrim` function on bytea type
- Part of #8831.

Support `trim` syntax: 
- trim ( [ LEADING | TRAILING | BOTH ] bytesremoved bytea FROM bytes bytea ) → bytea
- trim ( [ LEADING | TRAILING | BOTH ] [ FROM ] bytes bytea, bytesremoved bytea ) → bytea
- btrim ( bytes bytea, bytesremoved bytea ) → bytea
- ltrim ( bytes bytea, bytesremoved bytea ) → bytea
- rtrim ( bytes bytea, bytesremoved bytea ) → bytea

## Checklist

## Documentation

<details>

Support `trim`, `ltrim`, `btrim` and `rtrim` function on bytea type

</details>

### DIFF
--- a/src/expr/impl/src/scalar/trim.rs
+++ b/src/expr/impl/src/scalar/trim.rs
@@ -14,7 +14,7 @@
 
 use std::fmt::Write;
 
-use risingwave_expr::function;
+use risingwave_expr::{Result, function};
 
 #[function("trim(varchar) -> varchar")]
 pub fn trim(s: &str, writer: &mut impl Write) {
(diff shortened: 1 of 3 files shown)


src/tests/regress/data/sql/strings.sql
```diff
@@ -744,15 +744,15 @@ SELECT repeat('Pg', -4);
...
---@ SELECT trim(E'\\000'::bytea from E'\\000Tom\\000'::bytea);
---@ SELECT trim(leading E'\\000'::bytea from E'\\000Tom\\000'::bytea);
---@ SELECT trim(trailing E'\\000'::bytea from E'\\000Tom\\000'::bytea);
---@ SELECT btrim(E'\\000trim\\000'::bytea, E'\\000'::bytea);
---@ SELECT btrim(''::bytea, E'\\000'::bytea);
---@ SELECT btrim(E'\\000trim\\000'::bytea, ''::bytea);
---@ SELECT encode(overlay(E'Th\\000omas'::bytea placing E'Th\\001omas'::bytea from 2),'escape');
---@ SELECT encode(overlay(E'Th\\000omas'::bytea placing E'\\002\\003'::bytea from 8),'escape');
---@ SELECT encode(overlay(E'Th\\000omas'::bytea placing E'\\002\\003'::bytea from 5 for 3),'escape');
+SELECT trim(E'\\000'::bytea from E'\\000Tom\\000'::bytea);
```
Standing-Man: Hi @xiangjinwu and @BugenZhao, I found that the output of my implementation of these methods is consistent with Postgres, but different from `strings.out`.

```
> psql --version
psql (PostgreSQL) 17.5 (Ubuntu 17.5-1.pgdg24.04+1)
```
So which one should I keep consistent with?

src/expr/impl/src/scalar/trim.rs
```diff
@@ -58,6 +58,75 @@ pub fn rtrim_characters(s: &str, characters: &str, writer: &mut impl Write) {
...
+    let start = bytes
+        .iter()
+        .position(|b| !existed(b))
+        .unwrap_or(bytes.len());
+    let end = bytes
+        .iter()
+        .rposition(|b| !existed(b))
+        .map(|i| i + 1)
+        .unwrap_or(0);
+    bytes[start..end].iter().copied().collect()
```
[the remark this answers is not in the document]
Standing-Man: Got it, thanks.

src/tests/regress/data/sql/strings.sql
xiangjinwu: They are different because of `SET bytea_output TO escape;` above in `strings.sql`. RisingWave only supports `bytea_output = hex` which is also the default of PostgreSQL.

As a result, we would not be able to uncomment these test cases in this PR. Let's just test with slt for now.

Do remember to focus on edge cases like empty / `unwrap_or`. For example the following leads to a panic:
```
select btrim('1'::bytea, '1');
```

Standing-Man: In the latest commit, I included edge cases to test the trim function.

src/expr/impl/src/scalar/trim.rs
```diff
@@ -58,6 +58,78 @@ pub fn rtrim_characters(s: &str, characters: &str, writer: &mut impl Write) {
...
+/// ----
+/// \x345678
+/// ```
+#[function("trim(bytea, bytea) -> bytea")]
+pub fn trim_bytea(bytes: &[u8], bytesremoved: &[u8]) -> Box<[u8]> {
+    let (mut start, mut end) = trim_bound(bytes, bytesremoved);
+
+    if start == bytes.len() && end == 0 {
+        (start, end) = (0, 0);
+    }
```
xiangjinwu: This check is too specific to the details of `trim_bound`. It is better to guard any cases that the following expression `bytes[start..end]` would panic - by writing `bytes.get(start..end)` instead

Standing-Man: Thanks for the suggestion! That makes sense — I’ll update the code to use bytes.get(start..end) to handle potential panics more safely.

xiangjinwu (approved):
Thanks for the contribution!

PR #22765, opened 2025-08-01
fix(frontend): fix explain analyze on distributed worker nodes

kwannoel:

Previously, we compute the deltas for backpressure and row throughput by:

- Aggregating the initial metrics.
- For each worker node, we would then subtract their final metrics from the initial metrics.

We should instead aggregate all final metrics, and then subtract initial metrics from the final metrics. Otherwise this just leads to underflow.

## Checklist

## Documentation

<details>

</details>


chenzl25 (approved):

src/frontend/src/handler/explain_analyze_stream_job.rs
```diff
@@ -648,7 +768,13 @@ mod graph {
             for actor_id in &node.actor_ids {
                 let executor_id =
                     unique_executor_id_from_unique_operator_id(*actor_id, operator_id);
-                assert!(executor_ids.insert(executor_id));
+                if node.identity != NodeBodyDiscriminants::BatchPlan
+                // FIXME(kwannoel): Add back after https://github.com/risingwavelabs/risingwave/issues/22775 is resolved.
+                && node.identity != NodeBodyDiscriminants::Merge
+                && node.identity != NodeBodyDiscriminants::Project
```
kwannoel, line 774: @Chiro11 You can remove these two lines once the workaround is reverted.

PR #22769, opened 2025-08-02
feat: support more bytea type functions

Standing-Man:

- Support `get_bit`, `get_byte`, `set_bit`, `set_byte` and `bit_count` functions
- Part of #8831

## Checklist

## Documentation

<details>

Support `get_bit`, `get_byte`, `set_bit`, `set_byte` and `bit_count` functions.
</details>

### DIFF
--- a/src/expr/impl/src/scalar/bytea_bits.rs
+++ b/src/expr/impl/src/scalar/bytea_bits.rs
@@ -1,0 +1,90 @@
+// Copyright 2025 RisingWave Labs
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use risingwave_expr::{ExprError, Result, function};
+
+/// Extracts n'th bit from binary string.
+#[function("get_bit(bytea, int8) -> int4")]
+pub fn get_bit(bytes: &[u8], n: i64) -> Result<i32> {
+    let max_sz = (bytes.len() * 8) as i64;
+    if n < 0 || n >= max_sz {
+        return Err(ExprError::InvalidParam {
+            name: "get_bit",
+            reason: format!("index {} out of valid range, 0..{}", n, max_sz - 1).into(),
+        });
+    }
+    let index = n / 8;
+    let byte = bytes[index as usize];
+    Ok(((byte >> (n % 8)) & 1) as i32)
+}
+
+/// Sets n'th bit in binary string to newvalue.
+#[function("set_bit(bytea, int8, int4) -> bytea")]
+pub fn set_bit(bytes: &[u8], n: i64, value: i32) -> Result<Box<[u8]>> {
+    let max_sz = (bytes.len() * 8) as i64;
+    if n < 0 || n >= max_sz {
+        return Err(ExprError::InvalidParam {
+            name: "set_bit",
+            reason: format!("index {} out of valid range, 0..{}", n, max_sz - 1).into(),
+        });
+    }
+    let mut buf = bytes.to_vec();
+    let index = (n / 8) as usize;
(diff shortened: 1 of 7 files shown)


src/expr/impl/src/scalar/bytea_bits.rs
```diff
@@ -0,0 +1,135 @@
...
+    let max_sz = (bytes.len() * 8) as i64;
+    if n < 0 || n >= max_sz {
+        return Err(ExprError::InvalidParam {
+            name: "get_bit",
+            reason: format!("index {} out of valid range, 0..{}", n, max_sz - 1).into(),
+        });
+    }
+    let index = n / 8;
+    let byte = bytes[index as usize];
+    Ok(((byte >> (n % 8)) & 1) as i32)
```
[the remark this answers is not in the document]
Standing-Man: After making the changes based on your suggestion, the output of the related functions is inconsistent with Postgres

src/expr/impl/src/scalar/bytea_bits.rs
```diff
@@ -0,0 +1,135 @@
...
+    let max_sz = bytes.len() as i32;
+    if n < 0 || n >= max_sz {
+        return Err(ExprError::InvalidParam {
+            name: "set_byte",
+            reason: format!("index {} out of valid range, 0..{}", n, max_sz - 1).into(),
+        });
+    }
+    let mut buf = bytes.to_vec();
+    buf[n as usize] = value as u8;
+    Ok(buf.iter().copied().collect())
```
xiangjinwu: Both `to_vec` and `collect()` clones the whole sequence. We can just do one clone into `Box<[u8]>` directly and update in-place.

(same for `set_bit`)

src/expr/impl/src/scalar/bytea_bits.rs
```diff
@@ -0,0 +1,135 @@
...
+        return Err(ExprError::InvalidParam {
+            name: "set_bit",
+            reason: format!("index {} out of valid range, 0..{}", n, max_sz - 1).into(),
+        });
+    }
+    let mut buf = bytes.to_vec();
+    let index = (n / 8) as usize;
+    let bit_pos = (n % 8) as u8;
+
+    if value != 0 {
```
xiangjinwu, line 72: PostgreSQL rejects value other than 0 or 1

xiangjinwu (commented):
Rest lgtm. Thanks for the contribution.

src/expr/impl/src/scalar/bytea_bits.rs
Standing-Man: Thank you very much for your review. I have modified the code to perform only a single clone and added parameter checks

e2e_test/batch/functions/bytea_bits.slt.part
```diff
@@ -0,0 +1,117 @@
...
+SELECT set_bit('\x1234567890'::bytea, 28, 0);
+----
+\x1234566890
+
+query T
+SELECT set_bit('\x1234567890'::bytea, 34, 1);
+----
+\x1234567894
+
+statement error value 10 is invalid,new bit must be 0 or 1
```
xiangjinwu:
```suggestion
statement error value 10 is invalid, new bit must be 0 or 1
```

Btw, did you try PostgreSQL's behavior when `set_byte` gets a value larger than a byte?

Standing-Man: Thanks for the review. I’ve added two additional test points to ensure consistency with PostgreSQL.

Standing-Man:
> Thank you

Thank you for your review! I really appreciate your time and feedback.

src/expr/impl/src/scalar/bytea_bits.rs
```diff
@@ -0,0 +1,143 @@
...
+        buf[index] |= 1 << bit_pos;
+    } else {
+        buf[index] &= !(1 << bit_pos);
+    }
+    Ok(buf.into_boxed_slice())
```
BugenZhao, line 77: nit: I've just realized that we don't support the "writer" style for expressions returning `bytea` (similar to how we handle `varchar`). Theoretically this can save some copies.

https://github.com/risingwavelabs/risingwave/blob/c5b9b1d69a21c9e4eac816d3ed7e0267047de3dc/src/expr/impl/src/scalar/lower.rs#L19-L24

Standing-Man: Thank you for your review! Indeed, use **"writer"** avoid allocation and copy overhead. Maybe that’s why the merge queue check keeps failing?

PR #22774, opened 2025-08-04
feat(cdc): allow user config `debezium.time.precision.mode` to handle extreme timestamp

wcy-fdu:

Sometimes, users can ensure that the upstream CDC timestamps are all at millisecond precision. 
This PR allows users to specify this by configuring `debezium.time.precision.mode = connect` in with option of create source. At this point, the `TimestamptzHandling` in the backend should be set to `Milli`, instead of the previous `GuessNumberUnit` mode. 
This PR is intended to handle extreme timestamps(especially for sqlserver cdc); specific details can be referenced in the newly added tests.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/src/parser/unified/json.rs
+++ b/src/connector/src/parser/unified/json.rs
@@ -88,15 +88,7 @@ impl TimestamptzHandling {
 
 #[derive(Clone, Debug)]
 pub enum TimestampHandling {
-    /// `1712800800123` (milliseconds)
     Milli,
-    /// `1712800800123456` (microseconds)
-    Micro,
-    /// Both `1712800800123` (ms) and `1712800800123456` (us) maps to `2024-04-11`.
-    ///
-    /// Only works for `[1973-03-03 09:46:40, 5138-11-16 09:46:40)`.
-    ///
-    /// This option is backward compatible.
     GuessNumberUnit,
 }
 
@@ -512,7 +504,7 @@ impl JsonParseOptions {
                     TimestampHandling::Milli => Timestamp::with_millis(value.as_i64().unwrap())
                         .map_err(|_| create_error())?
                         .into(),
-                    _ => i64_to_timestamp(value.as_i64().unwrap())
+                    TimestampHandling::GuessNumberUnit => i64_to_timestamp(value.as_i64().unwrap())
                         .map_err(|_| create_error())?
                         .into(),
                 }


java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/DbzConnectorConfig.java
```diff
@@ -319,11 +319,17 @@ public DbzConnectorConfig(
         } else {
             throw new RuntimeException("unsupported source type: " + source);
         }
-
         var otherProps = extractDebeziumProperties(userProps);
         for (var entry : otherProps.entrySet()) {
-            dbzProps.putIfAbsent(entry.getKey(), entry.getValue());
+            // For time.precision.mode, user configuration should override default configuration
```
hzxa21: Can you include more contexts in the comment here so that people can understand why we treat time.precision.mode special? If we have a issue link, including it should be sufficient.

src/stream/src/from_proto/stream_cdc_scan.rs
```diff
@@ -122,6 +121,7 @@ impl ExecutorBuilder for StreamCdcScanExecutorBuilder {
                 state_table,
                 node.rate_limit,
                 scan_options,
+                properties.clone(),
```
hzxa21: nits: the clone is unneccessary

src/stream/src/from_proto/stream_cdc_scan.rs
```diff
@@ -141,6 +141,7 @@ impl ExecutorBuilder for StreamCdcScanExecutorBuilder {
...
```
hzxa21: nits: the clone is unneccessary

src/stream/src/executor/backfill/cdc/cdc_backfill.rs
```diff
@@ -201,9 +206,23 @@ impl<S: StateStore> CdcBackfillExecutor<S> {
...
+            Some(TimestamptzHandling::Milli)
+        } else {
+            None
+        };
         // Make sure to use mapping_message after transform_upstream.
-        let mut upstream = transform_upstream(upstream, self.output_columns.clone()).boxed();
+        let mut upstream = transform_upstream(
+            upstream,
+            self.output_columns.clone(),
+            timestamptz_handling.clone(),
```
hzxa21: nits: the clone is unneccessary.

src/stream/src/executor/backfill/cdc/cdc_backill_v2.rs
```diff
@@ -121,7 +128,26 @@ impl<S: StateStore> ParallelizedCdcBackfillExecutor<S> {
...
+            .unwrap_or(false)
+        {
+            Some(TimestamptzHandling::Milli)
+        } else {
+            None
+        };
+        let mut upstream = transform_upstream(
+            upstream,
+            self.output_columns.clone(),
+            timestamptz_handling.clone(),
```
hzxa21: nits: the clone is unnecessary

src/connector/src/parser/unified/json.rs
```diff
@@ -483,9 +485,17 @@ impl JsonParseOptions {
             (
                 DataType::Timestamp,
                 ValueType::I64 | ValueType::I128 | ValueType::U64 | ValueType::U128,
-            ) => i64_to_timestamp(value.as_i64().unwrap())
-                .map_err(|_| create_error())?
-                .into(),
+            ) => {
+                if let TimestamptzHandling::Milli = self.timestamptz_handling {
```
hzxa21: @xiangjinwu is it a good practice to use the `timestamptz_handling` for timestamp handling? The naming seems to suggest the it is related to tz.

wcy-fdu: maybe we should name it `timestamp_handling` here 🤔

java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/DbzConnectorConfig.java
wcy-fdu: More explanation added.

src/connector/src/parser/unified/json.rs
wcy-fdu: cc @xiangjinwu WDYT?

xiangjinwu: Let's introduce a dedicated `TimestampHandling` for it.

java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/DbzConnectorConfig.java
```diff
@@ -319,11 +319,27 @@ public DbzConnectorConfig(
...
+            // precision of the timestamps in the upstream table.
+            //  In this case, we can use this configuration item in the backend with a more precise
+            // TimestamptzHandling mode instead of `TimestamptzHandling::GuessNumberUnit`. Before
+            // upgrading Debezium to version 3.x, this is a
+            // workaround for handling extreme timestamps.
+            //  For more information, refer to
+            // https://github.com/risingwavelabs/risingwave/issues/21551.
+            if ("time.precision.mode".equals(entry.getKey())) {
+                dbzProps.put(entry.getKey(), entry.getValue());
+                LOG.info("Overriding time.precision.mode with user value: {}", entry.getValue());
```
xiangjinwu: https://github.com/risingwavelabs/risingwave/blob/cf6e921cd4bc2cd25ca59f47332a8ef1ae7b24c1/java/connector-node/risingwave-connector-service/src/main/resources/debezium.properties#L13-L15

What about simply `time.precision.mode=${debezium.time.precision.mode:-adaptive_time_microseconds}` without updating the Java logic?

src/common/src/cast/mod.rs
```diff
@@ -112,6 +112,14 @@ pub fn i64_to_timestamp(t: i64) -> Result<Timestamp> {
...
 }
 
+pub fn i64_to_timestamp_milli(t: i64) -> Result<Timestamp> {
+    let tz = Timestamptz::from_millis(t).unwrap_or_default();
+    Ok(Timestamp::from_timestamp_uncheck(
+        tz.timestamp(),
+        tz.timestamp_subsec_nanos(),
+    ))
+}
+
```
xiangjinwu: Duplicate with `Timestamp::with_millis`:
https://github.com/risingwavelabs/risingwave/blob/cf6e921cd4bc2cd25ca59f47332a8ef1ae7b24c1/src/common/src/types/datetime.rs#L568-L572

src/stream/src/executor/backfill/cdc/cdc_backill_v2.rs
```diff
@@ -121,7 +128,22 @@ impl<S: StateStore> ParallelizedCdcBackfillExecutor<S> {
...
+        let timestamptz_handling: Option<TimestamptzHandling> = if self
+            .properties
+            .get("debezium.time.precision.mode")
+            .map(|v| v == "connect")
+            .unwrap_or(false)
+        {
+            Some(TimestamptzHandling::Milli)
+        } else {
+            None
+        };
```
xiangjinwu: nit: `.then_some(TimestamptzHandling::Milli)`

e2e_test/source_inline/cdc/sql_server/sql_server_timestamp.slt.serial
xiangjinwu, line 1: https://buildkite.com/risingwavelabs/pull-request/builds/79493#01987576-a2e8-4551-922d-9e932aab3ba2/742-1059

```
e2e_test/source_inline/cdc/sql_server/sql_server_timestamp.slt.serial ..
[OK] in 38358 ms
```

src/connector/src/parser/unified/json.rs
```diff
@@ -483,9 +521,17 @@ impl JsonParseOptions {
...
+            ) => {
+                if let TimestampHandling::Milli = self.timestamp_handling {
+                    Timestamp::with_millis(value.as_i64().unwrap())
+                        .map_err(|_| create_error())?
+                        .into()
+                } else {
+                    i64_to_timestamp(value.as_i64().unwrap())
+                        .map_err(|_| create_error())?
+                        .into()
+                }
```
xiangjinwu: Did you write this or some "AI"? Why is `Micro` mistakenly fall into `else` together with `GuessNumberUnit`?

src/stream/src/executor/backfill/cdc/cdc_backfill.rs
```diff
@@ -201,9 +207,26 @@ impl<S: StateStore> CdcBackfillExecutor<S> {
...
+            .get("debezium.time.precision.mode")
+            .map(|v| v == "connect")
+            .unwrap_or(false)
+            .then_some(TimestampHandling::Milli);
+        let timestamptz_handling: Option<TimestamptzHandling> = self
+            .properties
+            .get("debezium.time.precision.mode")
+            .map(|v| v == "connect")
+            .unwrap_or(false)
+            .then_some(TimestamptzHandling::Milli);
```
xiangjinwu, line 221: This reminds me of another issue. Before any fix, could you add a `time` column to the tests?

The test is expected to fail, because `debezium.time.precision.mode = connect` also sets `TimeHandling::Milli` in the upstream, while our parser is currently hardcoded to `TimeHandling::Micro` (assuming `debezium.time.precision.mode = adaptive_time_microseconds`)

src/connector/src/parser/unified/json.rs
wcy-fdu: Just follow the previous implementation, only treat `Milli` specially. 
Because I am not sure when it will become `Micro` (I only tested the behavior of `time.precision.mode = connect`), and I did not test the behavior of Timestamp::with_micros.

Do you think we should do something like
```
match TimestampHandling ->{
Milli -> Timestamp::with_milli
Micro -> Timestamp::with_micros
Guess -> i64_to_timestamp
}
```

xiangjinwu: If you only intend to treat `Milli` specially, the `Micro` option shall not be introduced, which is now usable via `ENCODE JSON (timestamp.handling.mode = 'micro')`
https://github.com/risingwavelabs/risingwave/blob/a661bd6fdb49322d72df99c5a11e01233ad7f1e8/src/connector/src/parser/config.rs#L268-L270
https://github.com/risingwavelabs/risingwave/blob/a661bd6fdb49322d72df99c5a11e01233ad7f1e8/src/connector/src/parser/unified/json.rs#L103-L110

So by introducing this new **user-facing** option, we are responsible to implement it correctly.

(Of course we can also remove the logic above and hide it from user, just like `TimeHandling`.)

> Because I am not sure when it will become `Micro`

See above; or when `time.precision.mode = microseconds` on debezium 3.1

> Do you think we should do something like ...

Yes `Timestamp::with_micros`

src/stream/src/executor/backfill/cdc/cdc_backfill.rs
wcy-fdu: Good catch!

src/connector/src/parser/unified/json.rs
wcy-fdu: I think we only support config `time.precision.mode` to adaptive, adaptive_time_microseconds and connect. If user mis config it to `micro`, will receive following error
```
The 'time.precision.mode' value is invalid: Value must be one of adaptive, adaptive_time_microseconds, connect
```

xiangjinwu: To clarify, the logic above is also used by non-debezium logic `ENCODE JSON (timestamp.handling.mode = 'micro')`, in addition to `debezium.time.precision.mode` which only accepts `adaptive`, `adaptive_time_microseconds` and `connect` in 2.6

wcy-fdu: Remove `from_options` for `TimeHandling` and `TimestampHandling`.

src/connector/src/parser/unified/json.rs
```diff
@@ -86,6 +86,37 @@ impl TimestamptzHandling {
...
+    /// Only works for `[1973-03-03 09:46:40, 5138-11-16 09:46:40)`.
+    ///
+    /// This option is backward compatible.
+    GuessNumberUnit,
+}
+
+impl TimestampHandling {
+    pub const OPTION_KEY: &'static str = "timestamp.handling.mode";
+
+    pub fn from_options(
```
xiangjinwu: dead code?

src/connector/src/parser/unified/json.rs
```diff
@@ -483,9 +522,17 @@ impl JsonParseOptions {
...
                 ValueType::I64 | ValueType::I128 | ValueType::U64 | ValueType::U128,
-            ) => i64_to_timestamp(value.as_i64().unwrap())
-                .map_err(|_| create_error())?
-                .into(),
+            ) => {
+                if let TimestampHandling::Milli = self.timestamp_handling {
+                    Timestamp::with_millis(value.as_i64().unwrap())
+                        .map_err(|_| create_error())?
+                        .into()
+                } else {
```
xiangjinwu: Use a `match` here to handle each variant explicitly.

src/connector/src/parser/unified/json.rs
```diff
@@ -483,9 +505,18 @@ impl JsonParseOptions {
...
-                .map_err(|_| create_error())?
-                .into(),
+            ) => {
+                match self.timestamp_handling {
+                    // Only when user configures debezium.time.precision.mode = 'connect',
+                    // the Milli branch will be executed
+                    TimestampHandling::Milli => Timestamp::with_millis(value.as_i64().unwrap())
+                        .map_err(|_| create_error())?
+                        .into(),
+                    _ => i64_to_timestamp(value.as_i64().unwrap())
```
xiangjinwu: For the third time: either remove `::Micro` or handle it correctly with `with_micros`.

wcy-fdu: Remove `Micro` in `TimestampHandling` directly to avoid mis implementation in the future.

xiangjinwu (approved):

wcy-fdu:
Thank you @xiangjinwu  for the detailed explanation of the historical changes and implementation of timestamp!

PR #22781, opened 2025-08-04
feat(sink): add s3 writer snowflake sink

xxhZs:

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/src/sink/snowflake.rs
+++ b/src/connector/src/sink/snowflake.rs
@@ -14,6 +14,7 @@
 
 use core::num::NonZeroU64;
 use std::collections::BTreeMap;
+use std::fmt::Write;
 use std::sync::Arc;
 use std::time::{SystemTime, UNIX_EPOCH};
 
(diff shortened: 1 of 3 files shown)


src/connector/src/sink/snowflake.rs
```diff
@@ -736,6 +1028,36 @@ fn convert_snowflake_data_type(data_type: &DataType) -> Result<String> {
...
+    target_table_name: &str,
+) -> String {
+    let pipe_name = format!(r#""{}"."{}"."{}""#, database, schema, pipe_name);
+    let stage = format!(
+        r#""{}"."{}"."{}"/{}"#,
+        database, schema, stage, target_table_name
+    );
+    let table_name = format!(r#""{}"."{}"."{}""#, database, schema, table_name);
+    format!(
+        "CREATE OR REPLACE PIPE {} AUTO_INGEST = FALSE AS COPY INTO {} FROM @{} MATCH_BY_COLUMN_NAME = CASE_INSENSITIVE FILE_FORMAT = (type = 'JSON');",
```
hzxa21, line 1046: How do we remove the files from S3 after it has been loaded into snowflake?

src/connector/src/sink/snowflake.rs
```diff
@@ -736,6 +1028,36 @@ fn convert_snowflake_data_type(data_type: &DataType) -> Result<String> {
...
+        pipe_name, table_name, stage
+    )
+}
+
+fn build_flush_pipe_sql(database: &str, schema: &str, pipe_name: &str) -> String {
+    let pipe_name = format!(r#""{}"."{}"."{}""#, database, schema, pipe_name);
+    format!("ALTER PIPE {} REFRESH;", pipe_name,)
```
hzxa21, line 1053: Using `ALTER PIPE ... REFRESH` concerns me because the snowflake [doc](https://docs.snowflake.com/en/sql-reference/sql/alter-pipe) mentions:
- This SQL command can only load data files that were staged within the last 7 days.
- The REFRESH functionality is intended for short term use to resolve specific issues when Snowpipe fails to load a subset of files and is not intended for regular use.

That means this command has both data loss and stability issues for long-term use.

src/connector/src/sink/snowflake.rs
```diff
@@ -169,6 +192,14 @@ impl SnowflakeConfig {
         let jdbc_url = format!("{}?user={}&password={}", jdbc_url, username, password);
         let client = JdbcJniClient::new(jdbc_url)?;
 
+        if self.with_s3 {
+            let stage = self
+                .stage
+                .clone()
+                .ok_or(SinkError::Config(anyhow!("stage is required")))?;
```
hzxa21, line 199: IIUC, user needs to create a stage for each snowflake sink they create in RW. This is not user friendly. Can we create the stage for the user?

src/connector/src/sink/snowflake.rs
hzxa21: One idea: we can include the max `row_id` in the S3 file naming and remove the S3 files whose `row_id` < smallest `row_id` in the intermediate table in the background (e.g. every 1 day).

Issue #13868, opened 2023-12-08
Refactor `TableDesc` in frontend to `TableCatalog`


kwannoel:
Since `TableDesc` encompasses most of the fields in `TableCatalog`, we can just replace it with `TableCatalog` in frontend. That will simplify our frontend.

We should still serialize `TableCatalog` to `StorageTableDesc`, which is the proto definition for storage table.

@st1page's further suggestion
> a ‘stream table desc’ is ok too. My concern is that we should not mix the table’s catalog with the physical table instance’s descriptor

PR #22794, opened 2025-08-05
refactor(frontend): replace table desc with table catalog in optimizer

wenym1:

## Checklist

## Documentation

<details>

</details>


wenym1:
* **#22794** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22794?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22794?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#22810** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22810?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

wenym1:
> Personally I think it's reasonable to use `TableDesc` in physical plan, as it gives us a clear boundary about what fields are necessary during actual scanning/execution. I'm a bit worried that by always using `TableCatalog` we will get abstraction leak in the future.

Sounds good to me to have a clear boundary, but currently we have already stored `TableDesc` and `TableCatalog` together in the plan node, which means the boundary is already gone. The boundaries still exists between plan nodes and execution nodes (the node of physical stream and batch executor), since the execution nodes accept only `TableDesc`. If we this boundary in plan node, we may change to store only `TableDesc` in plan node. 

Storing either of them in plan node LGTM, but we should avoid storing both of them.

BugenZhao:
> > Personally I think it's reasonable to use `TableDesc` in physical plan, as it gives us a clear boundary about what fields are necessary during actual scanning/execution. I'm a bit worried that by always using `TableCatalog` we will get abstraction leak in the future.
> 
> Sounds good to me to have a clear boundary, but currently we have already stored `TableDesc` and `TableCatalog` together in the plan node, which means the boundary is already gone. The boundaries still exists between plan nodes and execution nodes (the node of physical stream and batch executor), since the execution nodes accept only `TableDesc`. If we this boundary in plan node, we may change to store only `TableDesc` in plan node.
> 
> Storing either of them in plan node LGTM, but we should avoid storing both of them.

Oh, I got it. Sounds good.

kwannoel:
Seems like both `TableDesc` and `TableCatalog` can both be in physical plan.

`TableDesc` is used in many areas: scans, providing table info to join, index, etc...
`TableCatalog` is used when we need to construct a state table.

There doesn't seem to be a very strict subset of rules around using `TableDesc`. As such it becomes rather blurry in terms of what `TableDesc` is meant to abstract. It seems like the definition is "as long as you don't need the full table, use TableDesc, and add whatever fields you need inside".

I'm in favour of removing `TableDesc` for now, I also found it very cumbersome when doing refactoring previously. We can reintroduce it later, when we find concrete scoping and definition of this struct.

kwannoel:
By the way, I'm also curious about the original motivation of this PR. What made you decide to do this refactoring @wenym1? Is it because you were reading some frontend code and noticed this poor abstraction? Appreciate the refactor.

kwannoel:
Also had a previous issue: https://github.com/risingwavelabs/risingwave/issues/13868 think we can close it after this PR is merged.

chenzl25 (approved):
LGTM. It makes the code cleaner.

BugenZhao:
> It seems like the definition is "as long as you don't need the full table, use TableDesc, and add whatever fields you need inside".

I think the boundary is still clear enough. The key point is that `TableDesc` resides in `common` while `TableCatalog` resides in `frontend`. So every time when adding fields to `TableDesc`, it means that these fields are for "backend" or "execution". If you check the existing fields in `TableDesc`, it appears that all of them are necessary but not randomly added just for compilation to pass.

PR #22807, opened 2025-08-06
refactor(pg-cdc): do not require table owner permission for pk access

wcy-fdu:

If a user has already created a publication, we do not need to perform operations such as `creating publication` or `alter publication add tables`, and table owner permissions are not required. 
In the previous implementation, the primary key was retrieved using SeaORM, which required querying the `information_schema.table_constraints` table, but this table requires higher permissions. This PR changes the primary key retrieval step to query other system tables to avoid this issue.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/src/connector_common/postgres.rs
+++ b/src/connector/src/connector_common/postgres.rs
@@ -41,7 +41,7 @@ const DISCOVER_PRIMARY_KEY_QUERY: &str = r#"
     SELECT a.attname as column_name
     FROM pg_index i
     JOIN pg_attribute a ON a.attrelid = i.indrelid AND a.attnum = ANY(i.indkey)
-    WHERE i.indrelid = ($1 || '.' || $2)::regclass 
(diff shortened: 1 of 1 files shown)


src/connector/src/connector_common/postgres.rs
```diff
@@ -65,6 +77,81 @@ pub struct PostgresExternalTable {
...
+        connection: &PgPool,
+        schema_name: &str,
+        table_name: &str,
+    ) -> ConnectorResult<Vec<String>> {
+        let rows = sqlx::query(DISCOVER_PRIMARY_KEY_QUERY)
+            .bind(schema_name)
+            .bind(table_name)
+            .fetch_all(connection)
+            .await
+            .map_err(|e| anyhow!("Failed to discover primary key columns: {}", e))?;
```
BugenZhao:
```suggestion
            .context("Failed to discover primary key columns")?;
```

hzxa21 (approved):

PR #22815, opened 2025-08-06
feat(sink): support copy into for redshift

xxhZs:

## Checklist

## Documentation

<details>

</details>


src/connector/src/sink/snowflake_redshift/snowflake.rs
```diff
@@ -467,9 +406,20 @@ impl SinkWriter for SnowflakeSinkWriter {
...
+                writer.barrier(is_checkpoint).await?;
+            }
         }
+        Ok(Some(SinkMetadata {
+            metadata: Some(sink_metadata::Metadata::Serialized(
+                risingwave_pb::connector_service::sink_metadata::SerializedMetadata {
+                    metadata: vec![],
+                },
+            )),
+        }))
```
hzxa21, line 426: I think Ok(None) is sufficient here?

src/connector/src/sink/snowflake_redshift/redshift.rs
```diff
@@ -486,9 +568,71 @@ impl SinkCommitCoordinator for RedshiftSinkCommitter {
...
+            let copy_into_sql = build_copy_into_sql(
+                self.config.schema.as_deref(),
+                table,
+                &path,
+                &self.config.s3_inner.access,
+                &self.config.s3_inner.secret,
+                &self.config.s3_inner.assume_role,
+            )?;
+            // run copy into
+            self.client.execute_sql_sync(&vec![copy_into_sql])?;
```
hzxa21, line 641: How do we remove the files that are already merged into the redshift table? See my previous comment [here](https://github.com/risingwavelabs/risingwave/pull/22781#discussion_r2265574338).

PR #22826, opened 2025-08-07
refactor(cdc): allow alter cdc source cdc.source.wait.streaming.start.timeout

wcy-fdu:

As title.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/src/allow_alter_on_fly_fields.rs
+++ b/src/connector/src/allow_alter_on_fly_fields.rs
@@ -94,6 +94,25 @@ mod sink_properties {
 pub static SOURCE_ALLOW_ALTER_ON_FLY_FIELDS: LazyLock<HashMap<String, HashSet<String>>> = LazyLock::new(|| {
     use source_properties::*;
     let mut map = HashMap::new();
+
+      // CDC Properties - added for schema.change.failure.policy
+      map.try_insert(
+        std::any::type_name::<MysqlCdcProperties>().to_owned(),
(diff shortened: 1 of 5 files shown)


src/connector/src/allow_alter_on_fly_fields.rs
```diff
@@ -94,6 +94,25 @@ mod sink_properties {
 pub static SOURCE_ALLOW_ALTER_ON_FLY_FIELDS: LazyLock<HashMap<String, HashSet<String>>> = LazyLock::new(|| {
     use source_properties::*;
     let mut map = HashMap::new();
+
+      // CDC Properties - added for schema.change.failure.policy
+      map.try_insert(
```
hzxa21: nits: extra space at the beginning

hzxa21 (approved):
LGTM

src/connector/src/allow_alter_on_fly_fields.rs
wcy-fdu: I found that the code in `allow_alter_on_fly_fields.rs` is actually auto-generated by `./risedev generate-with-option`. To allow a particular with option to be modified, we need to add `#[with_option(allow_alter_on_fly)]` above its definition, and then the code will be generated automatically.

~Their are some specific requirements on the CDC side that need further research on how to add this. For now, let's modify this PR to fix the issue where default session variables override the parameters in the with option.~

wcy-fdu: Talked with @tabVersion , hard code the new added config in `src/connector/src/with_options_test.rs`

Issue #16694, opened 2024-05-10
Support `random()` in batch queries


fuyufjh:
### Is your feature request related to a problem? Please describe.

Handy for generating testing data.

xiangjinwu:
The corresponding `setseed` function seems non-trivial. Note the following query is actually deterministic:

```
test=# select case n when 1 then setseed(0.4) else null end, random() from generate_series(1, 10) as t(n);
 case |       random        
------+---------------------
      |  0.6722690410337004
      |  0.9936586191615264
      | 0.19947461195735405
      |    0.90340354600701
      |   0.516998156895498
      |  0.7913052710048887
      |   0.531550972556847
      | 0.48302401286537666
      | 0.14525565255427209
      | 0.19221106009513123
(10 rows)
```

So what `random()` really does in PostgreSQL is mutating a session-level variable:
https://www.postgresql.org/docs/16/sql-set.html#:~:text=TO%20value.-,SEED,-Sets%20the%20internal
https://github.com/postgres/postgres/blob/REL_16_1/src/backend/utils/misc/guc_tables.c#L3695-L3704
https://github.com/postgres/postgres/blob/REL_16_1/src/common/pg_prng.c#L232-L234

Of course we can provide a weaker `random()` function first without the ability to generate deterministic sequence given a seed.

cc @wangrunji0408 may be more familiar with RNGs given experience working on madsim.

fuyufjh:
> So what random() really does in PostgreSQL is mutating a session-level variable:

> Of course we can provide a weaker random() function first without the ability to generate deterministic sequence given a seed.

Either LGTM.

PR #22835, opened 2025-08-08
feat(expr): support non deterministic random() 

lyang24:

Please explain **IN DETAIL** what the changes are in this PR and why they are needed:

- Summarize your change 
support non deterministic random in batch queries
image

- How does this PR work? 
link rust random to random sql function.
- Refer to a related PR or issue link 
touches: https://github.com/risingwavelabs/risingwave/issues/16694

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/e2e_test/batch/functions/random.slt.part
+++ b/e2e_test/batch/functions/random.slt.part
@@ -1,0 +1,12 @@
+statement ok
+select random();
+
+query I
+SELECT COUNT(*) FROM (SELECT random() FROM generate_series(1, 10));
+----
+10
+
+query T
+SELECT random() >= 0 AND random() < 1;
+----
+t
--- a/src/expr/impl/src/scalar/random.rs
+++ b/src/expr/impl/src/scalar/random.rs
@@ -1,0 +1,24 @@
+// Copyright 2025 RisingWave Labs
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use rand::random_range;
+use risingwave_common::types::F64;
+use risingwave_expr::function;
+
+/// Generates a random float between 0 and 1 inclusive.
+#[function("random() -> float8", volatile)]
+fn random() -> F64 {
+    let val: f64 = random_range(0.0..=1.0);
+    F64::from(val)
+}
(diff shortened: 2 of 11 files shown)


src/expr/impl/src/scalar/random.rs
```diff
@@ -0,0 +1,30 @@
...
+///
+/// ```slt
+/// query I
+/// SELECT random();
+/// ----
+/// 0.123456789
+/// ```
+#[function("random() -> float8", volatile)]
+fn random() -> F64 {
+    rand::rng().random::<F64>()
```
kwannoel: Use random_range instead?

https://docs.rs/rand/latest/rand/fn.random_range.html

I don't think `random` will have its range constrainted to [0, 1].

lyang24: sure updated
i though random() was a bit faster and we could change the range to [0, 1)  but either ways works

kwannoel: >  i though random() was a bit faster and we could change the range to [0, 1) but either ways works

Could you elaborate? Didn't get why it would be faster, and what do you mean by changing range to `[0, 1)` with `random`?

lyang24: i did some research:

random() for floats is just:
Generate a random integer (typically 64 bits).
Convert it into a float in [0, 1) by scaling.
That’s a couple of instructions, with almost no branching.

random_range / Uniform has extra steps:
Store and validate the range bounds.
Apply a scaling transform to map the generated number into the given range.
Sometimes do an extra check for inclusivity or deal with open/closed boundaries.

but honestly the difference is trival

e2e_test/batch/functions/random.slt.part
```diff
@@ -0,0 +1,7 @@
+statement ok
+select random();
+
+query I
+SELECT COUNT(*) FROM (SELECT random() FROM generate_series(1, 10));
```
kwannoel, line 5: How about we do a property check that the randomly generated number falls within the bounds `[0, 1.0]`

kwannoel (approved):
Rest LGTM

kwannoel:
Looks like CI is failing, you'll need to resolve the test failures as well.

src/expr/impl/src/scalar/random.rs
```diff
@@ -0,0 +1,31 @@
...
+use risingwave_common::types::F64;
+use risingwave_expr::function;
+
+/// Generates a random float between [0,1].
+///
+/// ```slt
+/// query I
+/// SELECT random();
+/// ----
+/// 0.123456789
```
BugenZhao: This will be used as a test case in CI, but it cannot pass.

src/frontend/src/binder/expr/function/builtin_scalar.rs
```diff
@@ -138,6 +138,16 @@ impl Binder {
             )
         }
 
+        fn random() -> Handle {
+            guard_by_len(
+                0,
+                raw(move |binder, _inputs| {
+                    binder.ensure_random_function_allowed()?;
```
BugenZhao: I think we don't need to ban it under streaming mode, since we've introduced `MaterializedExprs` for evaluation of impure expressions (https://github.com/risingwavelabs/risingwave/pull/20971)?

kwannoel: Good point.

kwannoel: @lyang24 I think we can remove it. You can ignore my prior suggestion to ban it for streaming.

src/expr/impl/src/scalar/random.rs
lyang24: thanks removed

kwannoel (approved):

kwannoel:
@wenym1 can you review on behalf of cargo.lock?

PR #22802, opened 2025-08-05
fix(dashboard): normalize backpressure values to 0-1 range

kwannoel:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## What's changed and what's your intention?

There's no need to further normalize the backpressure by dividing by 100. The values received here are not percentages but a float with the rate: `[0.0, 1.0]`.

Before the change, even though backpressure is at 98%, there's no colour transition for that edge or width transition:
Screenshot 2025-08-11 at 5 19 25 PM

After the change, you can see there's colour transition and width transition at 115% backpressure:
Screenshot 2025-08-11 at 5 19 11 PM

## Checklist

- [ ] I have written necessary rustdoc comments.
- [ ]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  I have checked the [Release Timeline](https://github.com/risingwavelabs/rw-commits-history/blob/main/release_timeline.md) and [Currently Supported Versions](https://docs.risingwave.com/changelog/release-support-policy#support-end-dates-for-recent-releases) to determine which release branches I need to cherry-pick this PR into. 

## Documentation

- [ ]  My PR needs documentation updates. 

<details>
<summary><b>Release note</b></summary>

</details>

PR #22801, opened 2025-08-05
feat(dashboard): add test data for dashboard rendering

kwannoel:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## What's changed and what's your intention?

Added dedicated test data for the dashboard's rendering capabilities. Previously we have some adhoc sql, that's not tested in CI. This PR unifies those SQL into a single slt script, and tests the queries to make sure they can run in CI.

We don't use `nexmark` or `tpch`, since these query sets are 1-deep, (i.e. base tables and MVs only on base tables, no mv-on-mv). So we cannot evaluate the graph rendering for deep graphs. Further, these queries also do not have high backpressure, and we can't really see how high backpressure is visualized for these.

This PR:
- Creates a new test suite in `e2e_test/dashboard/` with scripts to generate a complex graph of materialized views
- Adds `create_graph.slt.part` that creates 5 datagen tables and 10 materialized views with various dependencies
- Adds `drop_graph.slt.part` for proper cleanup of the test resources
- Updates the dashboard README to reference the new test data instead of the previous ad-hoc commands

## Checklist

- [ ] I have written necessary rustdoc comments.
- [ ] I have added necessary unit tests and integration tests.
- [ ] I have added test labels as necessary.
- [ ] I have added fuzzing tests or opened an issue to track them.
- [ ] My PR contains breaking changes.
- [ ] My PR changes performance-critical code, so I will run (micro) benchmarks and present the results.
- [ ] I have checked the [Release Timeline](https://github.com/risingwavelabs/rw-commits-history/blob/main/release_timeline.md) and [Currently Supported Versions](https://docs.risingwave.com/changelog/release-support-policy#support-end-dates-for-recent-releases) to determine which release branches I need to cherry-pick this PR into.

## Documentation

- [ ] My PR needs documentation updates.

PR #22838, opened 2025-08-08
feat(dashboard): add Prometheus-based streaming stats with backpressure metrics

kwannoel:

We want to support rendering the backpressure graph using prometheus metrics rather than [embedded `GLOBAL_STREAMING_STATS`](https://github.com/risingwavelabs/risingwave/blob/cb73906e7be2d641a7a701d634544b99230b6707/src/stream/src/executor/monitor/streaming_stats.rs#L44).

Computing backpressure rates requires 3 data points:
- Start blocking duration
- End blocking duration
- Profiling interval

Then we get backpressure rate via:
```
backpressure rate = (end blocking duration - start blocking duration) / profiling interval
```

For embedded metrics, the first time we query will be when the dashboard's backpressure graphs are opened.
We will first fetch the `start blocking duration`. We will then need to wait for some `profiling interval`, before querying the `end blocking duration`. The more we wait, the more accurate the profiling results.

However, when troubleshooting stuck streaming graphs, we can't afford to wait 10 minutes, just to get profiling results. If we use prometheus metrics, we can directly query historical `start` and `end` duration, and let prometheus calculate the backpressure rate for us. This is instantaneous compared to using `embedded metrics`, and the troubleshooter doesn't have to wait.

Another shortcoming is that embedded stats are the _latest stats_, whereas for prometheus we can query historical stats. Meaning we can investigate past bottlenecks as well.

---

This PR supports:
- The API layer for the frontend to query prometheus for relevant backpressure metrics.
- Update the refresh logic in the frontend, to use the new meta API for querying prometheus for metrics.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/dashboard/lib/api/streamingStats.ts
+++ b/dashboard/lib/api/streamingStats.ts
@@ -44,7 +44,7 @@ export interface StreamingStatsCallbacks {
 export function createStreamingStatsRefresh(
   callbacks: StreamingStatsCallbacks,
   initialSnapshot: ChannelStatsSnapshot | undefined,
-  statsType: StatsType,
+  statsType: StatsType
 ) {
   return function refresh() {
     // Try Prometheus first, fall back to embedded if it fails
--- a/dashboard/pages/fragment_graph.tsx
+++ b/dashboard/pages/fragment_graph.tsx
@@ -347,7 +347,7 @@ export default function Streaming() {
         toast,
       },
       initialSnapshot,
-      "fragment",
+      "fragment"
     )
 
     refresh() // run once immediately
--- a/dashboard/pages/relation_graph.tsx
+++ b/dashboard/pages/relation_graph.tsx
@@ -115,7 +115,7 @@ export default function StreamingGraph() {
         toast,
       },
       initialSnapshot,
-      "relation",
+      "relation"
     )
 
     refresh() // run once immediately


kwannoel:
* **#22838** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22838?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#22839](https://github.com/risingwavelabs/risingwave/pull/22839) <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22839?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#22843](https://github.com/risingwavelabs/risingwave/pull/22843) <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22843?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22838?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#22802** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22802?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#22801** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22801?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

dashboard/components/FragmentGraph.tsx
```diff
@@ -99,7 +99,7 @@ export default function FragmentGraph({
   planNodeDependencies: Map<string, d3.HierarchyNode<PlanNodeDatum>>
   fragmentDependency: FragmentBox[]
   selectedFragmentId?: string
-  channelStats?: Map<string, ChannelStatsDerived>
+  channelStats?: Map<string, ChannelDeltaStats>
```
kwannoel, line 102: Prometheus can directly return `ChannelStatsDerived`, since we can directly do rate calculation within the `promql`. So we construct a new proto definition for this format and call it `ChannelDeltaStats`. For embedded metrics, after all the calculation, it also produces the semantically equivalent `ChannelDeltaStats`. So we can just unify both of them.

dashboard/components/RelationGraph.tsx
```diff
@@ -32,8 +32,7 @@ import {
   RelationPoint,
   RelationPointPosition,
 } from "../lib/layout"
```
kwannoel, line 34: Just some renaming and unification in this file, no core logic.

dashboard/components/FragmentGraph.tsx
```diff
@@ -16,8 +16,8 @@ import * as dagre from "dagre"
 import { cloneDeep } from "lodash"
 import { Fragment, useCallback, useEffect, useRef, useState } from "react"
 import { Edge, Enter, FragmentBox, Position } from "../lib/layout"
```
kwannoel, line 18: Just some renaming and unification in this file, no core logic.

dashboard/lib/api/streamingStats.ts
```diff
@@ -0,0 +1,126 @@
...
+  setFragmentStats?: (
+    stats: { [key: number]: FragmentStats } | undefined
+  ) => void
+  toast: (
+    error: any,
+    status?: "info" | "warning" | "success" | "error" | "loading"
+  ) => void
+}
+
+export function createStreamingStatsRefresh(
```
kwannoel, line 44: Here we unify the refresh logic for both the `fragment` and `relation` graph, they are around the same, besides for updating fragment node vs relation nodes stats.

dashboard/lib/api/streamingStats.ts
```diff
@@ -0,0 +1,126 @@
...
+  callbacks: StreamingStatsCallbacks,
+  initialSnapshot: ChannelStatsSnapshot | undefined,
+  statsType: StatsType,
+) {
+  return function refresh() {
+    // Try Prometheus first, fall back to embedded if it fails
+    api.get("/metrics/streaming_stats_prometheus").then(
```
kwannoel, line 51: We use prometheus stats, since it has all the benefits:
- Support historical querying.
- Support instantaneous computation of backpressure rate (no need to wait N minutes).

We use embedded metrics as a fallback.

dashboard/pages/fragment_graph.tsx
```diff
@@ -342,38 +334,25 @@ export default function Streaming() {
...
   const [fragmentStats, setFragmentStats] = useState<{
     [key: number]: FragmentStats
   }>()
 
   useEffect(() => {
-    // The initial snapshot is used to calculate the rate of back pressure
-    // It's not used to render the page directly, so we don't need to set it in the state
     let initialSnapshot: ChannelStatsSnapshot | undefined
 
-    function refresh() {
```
kwannoel, line 356: Just reuse the refactored function above: https://github.com/risingwavelabs/risingwave/pull/22838/files#r2266136364

dashboard/pages/relation_graph.tsx
```diff
@@ -98,42 +95,29 @@ export default function StreamingGraph() {
...
-    // The initial snapshot is used to calculate the rate of back pressure
-    // It's not used to render the page directly, so we don't need to set it in the state
     let initialSnapshot: ChannelStatsSnapshot | undefined
 
     if (resetEmbeddedBackPressures) {
       setChannelStats(undefined)
       toggleResetEmbeddedBackPressures()
     }
 
-    function refresh() {
```
kwannoel, line 116: Just reuse the refactored function above: https://github.com/risingwavelabs/risingwave/pull/22838/files#r2266136364

BugenZhao:
I remember once we switched from Prometheus to embedded metrics, but not we're switching back. Just wondering if the original motiviation for the previous switch still apply?

kwannoel:
> I remember once we switched from Prometheus to embedded metrics, but not we're switching back. Just wondering if the original motiviation for the previous switch still apply?

The original motivation is to let the code be cleaner 😆. And I guess we didn't have access to prometheus in cloud over the meta dashboard previously, but now we do. So it's actually useful now.

kwannoel:
@BugenZhao @chenzl25 need you guys to have a look

BugenZhao:
> we didn't have access to prometheus in cloud over the meta dashboard previously, but now we do

Oh really? This sounds very nice. How will it work? 🤩

dashboard/lib/api/streamingStats.ts
BugenZhao: > Support historical querying

Seems not in this PR?

Oh, I see. It's #22843

kwannoel:
> > we didn't have access to prometheus in cloud over the meta dashboard previously, but now we do
> 
> Oh really? This sounds very nice. How will it work? 🤩

https://github.com/risingwavelabs/risingwave-cloud/pull/9873

kwannoel:
> > > we didn't have access to prometheus in cloud over the meta dashboard previously, but now we do
> > 
> > 
> > Oh really? This sounds very nice. How will it work? 🤩
> 
> [risingwavelabs/risingwave-cloud#9873](https://github.com/risingwavelabs/risingwave-cloud/pull/9873)

I just ran an experiment on RWC test: https://github.com/risingwavelabs/risingwave/pull/22846#issue-3307129581. Can confirm it works.

Issue #22799, opened 2025-08-05
optimizer: differentiate retractable stream and upsert stream


BugenZhao:
In optimizer, we only mark each plan as `append_only` or not: if `false`, then it is interpreted as a `retract` stream. However, this might be inaccurate as the output stream of some plan nodes in our system is actually `upsert` (no old record when deleting), typically upsert sources.

Considering that most executors (typically aggregation or join) cannot correctly handle upsert stream as the input. Currently, we manually ensure the correctness by banning some usages, like creating an upsert source without materializing it as a `TABLE`. However, this seems too strict and leads to some drawbacks, e.g., there's no way to ingest upsert data into our system without persisting the full history with `TABLE`. Theoretically,

- if an operator (typically stateless one) accepts upsert input, then it's okay to do so
- if a downstream supports upserts, we can sink an upsert stream into it
- besides `Materialize`, operators like `TemporalJoin`, `TemporalFilter`, `MaterializedExpr` can also convert upsert stream to retractable stream for further processing
- furthermore, any executors persisting full state are able to handle upsert stream, like `min/max` aggregation
- furthermore, if we find out that retract stream (complete `UpdataDelete` or `Delete` record) is not necessary, we may introduce and turn into upsert mode to improve execution performance

We may extend the optimizer to differentiate retractable stream and upsert stream to support such use cases.

kwannoel:
This reminds me of another related issue. We currently treat "upsert" as "retractable" in some of our sources. Which makes the semantics confusing. For instance, in kafka, and cdc, "upsert" actually means "retractable".

To me the conceptual definition is like:
- Append only: Only inserts
- Upsert: Only update + insert (that's why it's called upsert)
- Retractable: Deletes, updates and inserts.

However, this misses the case of partial deletes (e.g. only have the delete key, and no value).

Perhaps we should introduce a forth variant:
- RetractableAtLeastOnce: Update + insert, partial deletes.

PR #22798, opened 2025-08-05
refactor(optimizer): refactor `append_only: bool` into enum `StreamKind`

BugenZhao:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## What's changed and what's your intention?

See #22799 for motivation.

----

This PR, we only refactors `append_only: bool` into an enum `StreamKind`. It still only has 2 variants (`AppendOnly` and `Retract`). Logics for upsert stream will be added in future PRs.

## Checklist

- [x] I have written necessary rustdoc comments.
- [ ]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  I have checked the [Release Timeline](https://github.com/risingwavelabs/rw-commits-history/blob/main/release_timeline.md) and [Currently Supported Versions](https://docs.risingwave.com/changelog/release-support-policy#support-end-dates-for-recent-releases) to determine which release branches I need to cherry-pick this PR into. 

## Documentation

- [ ]  My PR needs documentation updates. 

<details>
<summary><b>Release note</b></summary>

</details>

PR #22092, opened 2025-06-03
feat: CDC ETL source

yuhao-su:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

This PR introduces the first part of the CDC ETL source, which allows you to create a etl source instead of table on shared cdc source to avoid materialized a copy of CDC data in RW tabl.

Users can use the source like any other source by creating MV, sink etc. on it.

We can run the following:
```
create source s1 with (
  ${RISEDEV_MYSQL_WITH_OPTIONS_COMMON},
  username = 'shared-cdc',
  password = 'abcdef',
  database.name = 'testdb1',
);

create source elt1_s1 (
  *
) from s1 table 'testdb1.tt1';
```

 

There are certain limitation in this PR.
TODO:
1. generated column on source definition.
2. handle parsing update messages
3. when doing backfill, need to lock the upstream database table with a read lock
4. add test for PG source

## Checklist

- [ ] I have written necessary rustdoc comments.
- [ ]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  I have checked the [Release Timeline](https://github.com/risingwavelabs/rw-commits-history/blob/main/release_timeline.md) and [Currently Supported Versions](https://docs.risingwave.com/changelog/release-support-policy#support-end-dates-for-recent-releases) to determine which release branches I need to cherry-pick this PR into. 

## Documentation

- [ ]  My PR needs documentation updates. 

<details>
<summary><b>Release note</b></summary>

</details>

PR #22840, opened 2025-08-08
refactor(optimizer): introduce `Upsert` stream kind & reject it in many cases

BugenZhao:

Motivation: #22799.

Following #22798, this PR introduces `Upsert` stream kind and reject it in many cases when an operator does not support processing `Upsert` stream as input.

For now, the only source of `Upsert` stream is the `Source` operator for sources that are not append-only. However, since users are not allowed to create bare upsert `SOURCE` currently, but always materializing them with a `TABLE`, there's still no hit for the rejection path.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/optimizer/plan_node/generic/join.rs
+++ b/src/frontend/src/optimizer/plan_node/generic/join.rs
@@ -24,6 +24,7 @@ use crate::expr::{ExprRewriter, ExprVisitor};
 use crate::optimizer::optimizer_context::OptimizerContextRef;
 use crate::optimizer::plan_node::StreamPlanRef;
 use crate::optimizer::plan_node::stream::StreamPlanNodeMetadata as _;
+use crate::optimizer::plan_node::stream::prelude::*;
 use crate::optimizer::plan_node::utils::TableCatalogBuilder;
 use crate::optimizer::property::{FunctionalDependencySet, StreamKind};
 use crate::utils::{ColIndexMapping, ColIndexMappingRewriteExt, Condition};
@@ -97,16 +98,18 @@ impl<PlanRef: GenericPlanRef> Join<PlanRef> {
 }
 
 impl Join<StreamPlanRef> {
-    pub fn stream_kind(&self) -> StreamKind {
+    pub fn stream_kind(&self) -> Result<StreamKind> {
+        let left_kind = reject_upsert_input!(self.left, "Join");
+        let right_kind = reject_upsert_input!(self.right, "Join");
+
         // Inner join won't change the append-only behavior of the stream. The rest might.
-        // TODO(kind): reject upsert input
         if let JoinType::Inner | JoinType::AsofInner = self.join_type
-            && self.left.append_only()
-            && self.right.append_only()
+            && let StreamKind::AppendOnly = left_kind
+            && let StreamKind::AppendOnly = right_kind
         {
-            StreamKind::AppendOnly
+            Ok(StreamKind::AppendOnly)
         } else {
-            StreamKind::Retract
+            Ok(StreamKind::Retract)
         }
     }
 
--- a/src/frontend/src/optimizer/plan_node/stream_dml.rs
+++ b/src/frontend/src/optimizer/plan_node/stream_dml.rs
@@ -38,11 +38,11 @@ impl StreamDml {
             input.stream_key().map(|v| v.to_vec()),
             input.functional_dependency().clone(),
             input.distribution().clone(),
-            if append_only {
+            input.stream_kind().merge(if append_only {
                 StreamKind::AppendOnly
             } else {
                 StreamKind::Retract
-            }, // TODO(kind): merge input stream kind
+            }),
             false,                   // TODO(rc): decide EOWC property
             WatermarkColumns::new(), // no watermark if dml is allowed
             MonotonicityMap::new(),  // TODO: derive monotonicity
--- a/src/frontend/src/optimizer/plan_node/stream_dml.rs
+++ b/src/frontend/src/optimizer/plan_node/stream_dml.rs
@@ -41,7 +41,9 @@ impl StreamDml {
             input.stream_kind().merge(if append_only {
                 StreamKind::AppendOnly
             } else {
-                StreamKind::Retract
+                // We cannot guarantee that there's no conflict on stream key between upstream
+                // source and DML input, so we must treat it as upsert here.
+                StreamKind::Upsert
             }),
             false,                   // TODO(rc): decide EOWC property
             WatermarkColumns::new(), // no watermark if dml is allowed
--- a/src/frontend/src/optimizer/plan_node/stream_dml.rs
+++ b/src/frontend/src/optimizer/plan_node/stream_dml.rs
@@ -31,6 +31,7 @@ pub struct StreamDml {
 }
 
 impl StreamDml {
+    // `append_only` indicates whether only `INSERT` is allowed.
     pub fn new(input: PlanRef, append_only: bool, column_descs: Vec<ColumnDesc>) -> Self {
         let base = PlanBase::new_stream(
             input.ctx(),
@@ -39,6 +40,9 @@ impl StreamDml {
             input.functional_dependency().clone(),
             input.distribution().clone(),
             input.stream_kind().merge(if append_only {
+                // For append-only table. Either there will be a `RowIdGen` following the `Dml` and `Union`,
+                // or there will be a `Materialize` with conflict handling enabled. In both cases there
+                // will be no key conflict, so we can treat the merged stream as append-only here.
                 StreamKind::AppendOnly
             } else {
                 // We cannot guarantee that there's no conflict on stream key between upstream
--- a/src/frontend/src/optimizer/plan_node/stream_hash_agg.rs
+++ b/src/frontend/src/optimizer/plan_node/stream_hash_agg.rs
@@ -99,7 +99,6 @@ impl StreamHashAgg {
                 // in EOWC mode, we produce append only output
                 StreamKind::AppendOnly
             } else {
-                // TODO(kind): reject upsert input
                 StreamKind::Retract
             },
             emit_on_window_close,
--- a/src/frontend/src/optimizer/plan_node/stream_hash_agg.rs
+++ b/src/frontend/src/optimizer/plan_node/stream_hash_agg.rs
@@ -51,7 +51,7 @@ impl StreamHashAgg {
         core: generic::Agg<PlanRef>,
         vnode_col_idx: Option<usize>,
         row_count_idx: usize,
-    ) -> Self {
+    ) -> Result<Self> {
         Self::new_with_eowc(core, vnode_col_idx, row_count_idx, false)
     }
 
@@ -60,8 +60,9 @@ impl StreamHashAgg {
         vnode_col_idx: Option<usize>,
         row_count_idx: usize,
         emit_on_window_close: bool,
-    ) -> Self {
+    ) -> Result<Self> {
         assert_eq!(core.agg_calls[row_count_idx], PlanAggCall::count_star());
+        reject_upsert_input!(core.input);
 
         let input = core.input.clone();
         let input_dist = input.distribution();
@@ -105,14 +106,15 @@ impl StreamHashAgg {
             watermark_columns,
             MonotonicityMap::new(), // TODO: derive monotonicity
         );
-        StreamHashAgg {
+
+        Ok(StreamHashAgg {
             base,
             core,
             vnode_col_idx,
             row_count_idx,
             emit_on_window_close,
             window_col_idx,
-        }
+        })
     }
 
     pub fn agg_calls(&self) -> &[PlanAggCall] {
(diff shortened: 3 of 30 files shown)


BugenZhao:
* **#22856** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22856?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#22840** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22840?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22840?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#22798** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22798?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/frontend/src/optimizer/plan_node/generic/join.rs
```diff
@@ -97,16 +98,18 @@ impl<PlanRef: GenericPlanRef> Join<PlanRef> {
 }
 
 impl Join<StreamPlanRef> {
-    pub fn stream_kind(&self) -> StreamKind {
```
wenym1, line 100: Shall we return error at the time we construct a problematic `Join<StreamPlanRef>` instead of exposing the error later when the `stream_kind` is called? In this way, we can ensure that all constructed `Join<StreamingPlanRef>` is valid, and many `.unwrap()` in this PR can be avoided. ditto for all similar generic core struct in this PR. We shall at least change the `.unwrap()` to `.expect("reason")` or add comment to illustrate the reason why it's safe to unwrap.

BugenZhao: > instead of exposing the error later when the `stream_kind` is called?

This is not that true. `Join::stream_kind` is eagerly called when constructing a new `StreamXxxJoin` plan node to construct the cached computed properties `base`, so that we've already ensured that all constructed plan nodes are valid.

Also in almost all remaining cases it's not made a method of the core struct. The stream kind is derived in the constructor of corresponding plan nodes.

wenym1: I was thinking about limiting the ability to construct a `generic::Join<StreamingPlanRef>` with upsert input via some proper encapsulation, so as to avoid the `.unwrap()` in this PR. But it turns out to be complicated, since the its `input` is a `pub` field, and we have method like `clone_with_inputs` and `rewrite_expr` that can somehow reconstruct the core with new input, and we can't propagate the error and can only panic, since the related trait method definition does not returns result.

src/frontend/src/optimizer/plan_node/stream_hash_agg.rs
```diff
@@ -99,7 +99,6 @@ impl StreamHashAgg {
                 // in EOWC mode, we produce append only output
                 StreamKind::AppendOnly
             } else {
-                // TODO(kind): reject upsert input
```
wenym1, line 102: Should we follow the removed comment to reject upsert input?

src/frontend/src/optimizer/plan_node/stream_materialize.rs
```diff
@@ -54,22 +54,34 @@ pub struct StreamMaterialize {
...
+    pub fn new(input: PlanRef, table: TableCatalog) -> Result<Self> {
+        let kind = match table.conflict_behavior() {
+            ConflictBehavior::NoCheck => {
+                reject_upsert_input!(input, "Materialize without conflict handling")
+            }
+
+            // When conflict handling is enabled, upsert stream can be converted to retract stream.
+            ConflictBehavior::Overwrite
+            | ConflictBehavior::IgnoreConflict
+            | ConflictBehavior::DoUpdateIfNotNull => match input.stream_kind() {
```
wenym1, line 66: If the input stream kind is `AppendOnly` and `Retract`, in the future will we have some way to force rewrite the conflict behavior of the `table` to `NoCheck` to avoid unnecessary overhead?

src/frontend/src/optimizer/plan_node/stream_union.rs
```diff
@@ -72,12 +72,9 @@ impl StreamUnion {
         let base = PlanBase::new_stream_with_core(
             &core,
             dist,
-            if inputs.iter().all(|x| x.append_only()) {
-                StreamKind::AppendOnly
-            } else {
-                // TODO(kind): handle `StreamKind::Upsert`
-                StreamKind::Retract
-            },
+            (inputs.iter().map(|i| i.stream_kind()))
```
wenym1: Should it always be `Upsert`? In the implementation of `StreamUnion`, it does not apply any duplicated key check from different inputs to ensure a `Retract` output🤔 In other words, when two inputs yield a same key, the `Retract` is violated. Therefore, should it always be `Upsert`, and it will be later restored to `Retract` via some stateful dedup plan node?

src/frontend/src/optimizer/plan_node/stream_dml.rs
```diff
@@ -38,11 +38,11 @@ impl StreamDml {
             input.stream_key().map(|v| v.to_vec()),
             input.functional_dependency().clone(),
             input.distribution().clone(),
-            if append_only {
+            input.stream_kind().merge(if append_only {
```
wenym1, line 42: Should it always be `Upsert`? Since we didn't handle duplicated key between dml and upstream source.

src/frontend/src/optimizer/plan_node/stream_row_merge.rs
```diff
@@ -87,7 +87,7 @@ impl StreamRowMerge {
             lhs_input.stream_key().map(|k| k.to_vec()),
             functional_dependency,
             lhs_input.distribution().clone(),
-            lhs_input.stream_kind(), // TODO(kind): reject upsert input
+            reject_upsert_input!(lhs_input),
```
wenym1, line 90: Do we assume that the two inputs must operate on different keys? If so, we may add a comment here to document the assumption.

wenym1 (commented):
Rest LGTM.

Most comments are about what's the expected stream kind for a plan node that merges multiple streams (`StreamUnion`, `StreamDml` and `StreamRowMerge`). Personally I think they should all be `Upsert`, because in their internal implementation, they don't handle duplicated keys from different inputs, and can easily violate the `Retract`, unless we can depend on some further assumption when we create the plan node.

src/frontend/src/optimizer/plan_node/stream_hash_agg.rs
BugenZhao: Good catch! Missed that.

src/frontend/src/optimizer/plan_node/stream_union.rs
BugenZhao: > In the implementation of `StreamUnion`, it does not apply any duplicated key check from different inputs to ensure a `Retract` output

This is true for the executor itself, but we guarantee that all inputs are attached with a "source column" indicating which input source it's from. So in fact there's no conflict. I'll add some comments to illustrate this.

chenzl25: `StreamUnion` has a `source_col` field to distinguish the records from different inputs, so it should be `Retract`.

src/frontend/src/optimizer/plan_node/stream_row_merge.rs
BugenZhao: I'm not quite sure about this. cc @kwannoel to confirm.

src/frontend/src/optimizer/plan_node/stream_materialize.rs
BugenZhao: This is a good point. Also on the other side, we can forcibly (or via some option) enable conflict handling on materialized views atop of an upsert stream, so it can be successfully materialized.

src/frontend/src/optimizer/plan_node/stream_row_merge.rs
kwannoel: > Do we assume that the two inputs must operate on different keys?

I don't follow. What do you mean to operate on different keys?

src/frontend/src/optimizer/plan_node/stream_dml.rs
BugenZhao: Makes sense. Actually this inspires me that whether we should also introduce an `Inconsistent` stream kind...

src/frontend/src/optimizer/plan_node/stream_row_merge.rs
wenym1: Ignore me. From the document of the struct, I thought it works like a union over the two stream, but it's actually not.

src/frontend/src/optimizer/plan_node/stream_dml.rs
wenym1: What's the difference between `Upsert` and `Inconsistent`? I think `Upsert` is already inconsistent🤔

chenzl25: If we don't guarantee that the Upsert's delete record exists, then it is almost the same as "Inconsistent". It's a bit weird to say a database DML is inconsistent, so I think we can use `Upsert` instead.

BugenZhao: > What's the difference between `Upsert` and `Inconsistent`? I think `Upsert` is already inconsistent🤔

For example, we can safely emit `Upsert` stream to an upsert sink, but not for `Inconsistent` stream.

Since the inconsistent stream appears only after DML and before materialization, I think it can be considered a special case. So maybe it's not that necessary to introduce a new kind for this..

chenzl25:
Put this related PR here. The table type of this PR should be Upsert. https://github.com/risingwavelabs/risingwave/pull/22092

src/frontend/src/optimizer/plan_node/stream_union.rs
BugenZhao: Update to determine based on the presence of `source_col`, because I find that in some cases there's no `source_col`, such as the one used in table plan. 🤣

src/frontend/src/optimizer/plan_node/stream_union.rs
```diff
@@ -69,15 +69,25 @@ impl StreamUnion {
             watermark_columns.insert(idx, ctx.next_watermark_group_id());
         }
 
+        let kind = if inputs.is_empty() {
+            // No data. Can be interpreted as append-only.
+            StreamKind::AppendOnly
```
chenzl25: If the input size is 0, it is an invalid plan, we could add an assertion to the new method.

src/frontend/src/optimizer/plan_node/stream_union.rs
```diff
@@ -69,15 +69,25 @@ impl StreamUnion {
...
+        } else if core.source_col.is_some() {
+            // There's no handling on key conflict in executor implementation. However, in most cases
+            // there's a `source_col` as a part of stream key, indicating which input the row comes from,
+            // there won't be actual key conflict and we can safely call `merge`.
+            (inputs.iter().map(|i| i.stream_kind()))
+                .reduce(StreamKind::merge)
+                .unwrap()
+        } else {
+            // Otherwise we must treat it as upsert.
+            StreamKind::Upsert
```
chenzl25: If the input is 1, it should be a retractable stream, e.g. a simple table.

src/frontend/src/optimizer/plan_node/stream_dml.rs
```diff
@@ -38,11 +38,13 @@ impl StreamDml {
...
                 StreamKind::AppendOnly
```
wenym1, line 46: May add comment like in `StreamUnion` to mention that this is for append only table with RowIdGen to ensure that the dml won't operate on the same key with upstream when `append_only`.

src/frontend/src/optimizer/plan_node/stream_union.rs
```diff
@@ -69,15 +70,33 @@ impl StreamUnion {
...
+                .reduce(StreamKind::merge)
+                .unwrap()
+        } else {
+            // No `source_col`, typically used in a `TABLE` plan to merge inputs from external source,
+            // upstream sink-into-table jobs, and DML.
+            if inputs.len() == 1 {
+                // Single input. Follow the input's kind.
+                inputs[0].stream_kind()
+            } else if inputs.iter().all(|x| x.stream_kind().is_append_only()) {
+                // Special case for append-only table. There must be a `RowIdGen` following the `Union`,
```
wenym1: We can define append-only table with primary key, and there won't be a `RowIdGen`. Does this statement still hold in this case?

chenzl25 (approved):
LGTM

src/frontend/src/optimizer/plan_node/stream_union.rs
BugenZhao: This is true but there's still no actual problem. I find normalizing such properties in the table plan can be hard, as in fact the streams here violates many invariants during such phases. Let's just aim to define them at best effort, prioritizing correctness.

kwannoel (approved):

PR #22851, opened 2025-08-11
fix(frontend): fix explain analyze on distributed worker nodes

kwannoel:

Cherry pick https://github.com/risingwavelabs/risingwave/pull/22765 and https://github.com/risingwavelabs/risingwave/pull/22234

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/handler/explain_analyze_stream_job.rs
+++ b/src/frontend/src/handler/explain_analyze_stream_job.rs
@@ -769,11 +769,7 @@ mod graph {
             for actor_id in &node.actor_ids {
                 let executor_id =
                     unique_executor_id_from_unique_operator_id(*actor_id, operator_id);
-                if node.identity != NodeBodyDiscriminants::BatchPlan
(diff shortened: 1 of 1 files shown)


kwannoel:
There's a bug found in explain analyze for dropping metrics. Let me investigate further.

kwannoel:
Actually, the failure is totally unrelated to the commits here.

Buildkite: https://buildkite.com/risingwavelabs/main-cron/builds/6579#01989dc0-ca1f-42fe-8b8a-cf792faa7ad3

Error:
```
thread 'rw-streaming' panicked at src/common/metrics/src/guarded_metrics.rs:327:71:
should exist because the current existing dropping one means the count is not zero
stack backtrace:
```

This error occurs from prometheus metrics. Explain analyze only uses in-memory metrics.

cyliu0:
So should I merge this via bypassing the merge rules?

kwannoel:
> So should I merge this via bypassing the merge rules?

No need, I'm going to keep retrying, to make sure no other errors are hit.

PR #22852, opened 2025-08-11
fix(pg-cdc): handle TOAST column when update

wcy-fdu:

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/proto/catalog.proto
+++ b/proto/catalog.proto
@@ -496,6 +496,9 @@ message Table {
   // NOTICE: when it is "None", the watermark column should be the first column in the pk
   optional int32 clean_watermark_index_in_pk = 44;
 
+  // TOAST column indices for PostgreSQL CDC tables. Empty for non-CDC tables or CDC tables without TOAST columns.
+  repeated int32 toastable_column_indices = 45;
+
   // Per-table catalog version, used by schema change. `None` for internal
   // tables and tests. Not to be confused with the global catalog version for
   // notification service.
--- a/src/frontend/src/catalog/table_catalog.rs
+++ b/src/frontend/src/catalog/table_catalog.rs
@@ -197,6 +197,9 @@ pub struct TableCatalog {
     pub engine: Engine,
 
     pub clean_watermark_index_in_pk: Option<usize>,
+
+    /// TOAST-able column indices for PostgreSQL CDC tables. None for non-CDC tables or CDC tables without TOAST columns.
+    pub toastable_column_indices: Option<Vec<usize>>,
 }
 
 pub const ICEBERG_SOURCE_PREFIX: &str = "__iceberg_source_";
@@ -582,6 +585,11 @@ impl TableCatalog {
             job_id: self.job_id.map(|id| id.table_id),
             engine: Some(self.engine.to_protobuf().into()),
             clean_watermark_index_in_pk: self.clean_watermark_index_in_pk.map(|x| x as i32),
+            toastable_column_indices: self
+                .toastable_column_indices
+                .as_ref()
+                .map(|indices| indices.iter().map(|&x| x as i32).collect())
+                .unwrap_or_default(),
         }
     }
 
@@ -784,6 +792,16 @@ impl From<PbTable> for TableCatalog {
             job_id: tb.job_id.map(TableId::from),
             engine,
             clean_watermark_index_in_pk: tb.clean_watermark_index_in_pk.map(|x| x as usize),
+            toastable_column_indices: if tb.toastable_column_indices.is_empty() {
+                None
+            } else {
+                Some(
+                    tb.toastable_column_indices
+                        .iter()
+                        .map(|&x| x as usize)
+                        .collect(),
+                )
+            },
         }
     }
 }
@@ -875,6 +893,7 @@ mod tests {
             job_id: None,
             engine: Some(PbEngine::Hummock as i32),
             clean_watermark_index_in_pk: None,
+            toastable_column_indices: vec![],
         }
         .into();
 
@@ -943,6 +962,7 @@ mod tests {
                 job_id: None,
                 engine: Engine::Hummock,
                 clean_watermark_index_in_pk: None,
+                toastable_column_indices: None,
             }
         );
         assert_eq!(table, TableCatalog::from(table.to_prost()));
--- a/src/meta/src/controller/mod.rs
+++ b/src/meta/src/controller/mod.rs
@@ -266,6 +266,7 @@ impl From<ObjectModel<table::Model>> for PbTable {
             job_id: value.0.belongs_to_job_id.map(|id| id as _),
             engine: value.0.engine.map(|engine| PbEngine::from(engine) as i32),
             clean_watermark_index_in_pk: value.0.clean_watermark_index_in_pk,
+            toastable_column_indices: vec![],
         }
     }
 }
--- a/src/stream/src/executor/mview/materialize.rs
+++ b/src/stream/src/executor/mview/materialize.rs
@@ -73,8 +73,8 @@ pub struct MaterializeExecutor<S: StateStore, SD: ValueRowSerde> {
     /// Used for APPEND ONLY table with iceberg engine. All data will be written to iceberg table directly.
     is_dummy_table: bool,
 
-    /// Whether this table is a CDC table.
-    is_cdc_table: bool,
+    /// Indices of TOAST columns for PostgreSQL CDC tables. None means either non-CDC table or CDC table without TOAST columns.
+    toastable_column_indices: Option<Vec<usize>>,
 }
 
 fn get_op_consistency_level(
@@ -155,7 +155,18 @@ impl<S: StateStore, SD: ValueRowSerde> MaterializeExecutor<S, SD> {
         let is_dummy_table =
             table_catalog.engine == Some(Engine::Iceberg as i32) && table_catalog.append_only;
 
-        let is_cdc_table = table_catalog.cdc_table_id.is_some();
+        // Extract TOAST-able column indices from table catalog.
+        let toastable_column_indices = if table_catalog.toastable_column_indices.is_empty() {
+            None
+        } else {
+            Some(
+                table_catalog
+                    .toastable_column_indices
+                    .iter()
+                    .map(|&x| x as usize)
+                    .collect(),
+            )
+        };
 
         Self {
             input,
@@ -175,7 +186,7 @@ impl<S: StateStore, SD: ValueRowSerde> MaterializeExecutor<S, SD> {
             may_have_downstream,
             depended_subscription_ids,
             metrics: mv_metrics,
-            is_cdc_table,
+            toastable_column_indices,
         }
     }
 
@@ -216,12 +227,12 @@ impl<S: StateStore, SD: ValueRowSerde> MaterializeExecutor<S, SD> {
                         && self.version_column_index.is_none()
                         && self.conflict_behavior == ConflictBehavior::Overwrite;
 
-                    // For CDC tables, disable optimization to ensure TOAST column fix is applied
-                    let do_not_handle_conflict = if self.is_cdc_table {
-                        false
-                    } else {
-                        do_not_handle_conflict
-                    };
+                    // // For CDC tables with TOAST-able columns, disable optimization to ensure TOAST column fix is applied
+                    // let do_not_handle_conflict = if self.toastable_column_indices.is_some() {
+                    //     false
+                    // } else {
+                    //     do_not_handle_conflict
+                    // };
 
                     match self.conflict_behavior {
                         checked_conflict_behaviors!() if !do_not_handle_conflict => {
@@ -270,7 +281,7 @@ impl<S: StateStore, SD: ValueRowSerde> MaterializeExecutor<S, SD> {
                                     &self.state_table,
                                     self.conflict_behavior,
                                     &self.metrics,
-                                    self.is_cdc_table,
+                                    self.toastable_column_indices.as_deref(),
                                 )
                                 .await?;
 
@@ -433,7 +444,7 @@ impl<S: StateStore> MaterializeExecutor<S, BasicSerde> {
             conflict_behavior,
             version_column_index: None,
             is_dummy_table: false,
-            is_cdc_table: false,
+            toastable_column_indices: None,
             may_have_downstream: true,
             depended_subscription_ids: HashSet::new(),
             metrics,
@@ -441,22 +452,25 @@ impl<S: StateStore> MaterializeExecutor<S, BasicSerde> {
     }
 }
 
-/// Debezium's default unavailable value placeholder for TOAST columns
+/// Debezium's default unavailable value placeholder for TOAST columns.
 const DEBEZIUM_UNAVAILABLE_VALUE: &str = "__debezium_unavailable_value";
 
-/// Fix TOAST columns by replacing unavailable values with old row values
-fn handle_toast_columns_for_cdc(old_row: &OwnedRow, new_row: &OwnedRow) -> OwnedRow {
+/// Fix TOAST columns by replacing unavailable values with old row values.
+fn handle_toast_columns_for_cdc(
+    old_row: &OwnedRow,
+    new_row: &OwnedRow,
+    toastable_indices: &[usize],
+) -> OwnedRow {
     let mut fixed_row_data = new_row.as_inner().to_vec();
 
-    for (i, new_datum) in new_row.iter().enumerate() {
-        if let Some(new_scalar_ref) = new_datum {
-            if let risingwave_common::types::ScalarRefImpl::Utf8(val) = new_scalar_ref {
-                if val == DEBEZIUM_UNAVAILABLE_VALUE {
-                    // Replace with old row value if available
-                    if let Some(old_datum_ref) = old_row.datum_at(i) {
-                        fixed_row_data[i] = Some(old_datum_ref.into_scalar_impl());
-                    }
-                }
+    for &toast_idx in toastable_indices {
+        if let Some(risingwave_common::types::ScalarRefImpl::Utf8(val)) =
+            new_row.datum_at(toast_idx)
+            && val == DEBEZIUM_UNAVAILABLE_VALUE
+        {
+            // Replace with old row value if available
+            if let Some(old_datum_ref) = old_row.datum_at(toast_idx) {
+                fixed_row_data[toast_idx] = Some(old_datum_ref.into_scalar_impl());
             }
         }
     }
@@ -633,7 +647,7 @@ impl<SD: ValueRowSerde> MaterializeCache<SD> {
         table: &StateTableInner<S, SD>,
         conflict_behavior: ConflictBehavior,
         metrics: &MaterializeMetrics,
-        is_cdc_table: bool,
+        toastable_column_indices: Option<&[usize]>,
     ) -> StreamExecutorResult<ChangeBuffer> {
         assert_matches!(conflict_behavior, checked_conflict_behaviors!());
 
@@ -683,8 +697,10 @@ impl<SD: ValueRowSerde> MaterializeCache<SD> {
                                 true
                             };
                             if need_overwrite {
-                                let final_row = if is_cdc_table {
-                                    // Apply TOAST column fix for CDC tables
+                                let final_row = if let Some(toastable_indices) =
+                                    toastable_column_indices
+                                {
+                                    // For TOAST-able columns, replace Debezium's unavailable value placeholder with old row values.
                                     let old_row_deserialized =
                                         row_serde.deserializer.deserialize(old_row.row.clone())?;
                                     let new_row_deserialized =
@@ -692,6 +708,7 @@ impl<SD: ValueRowSerde> MaterializeCache<SD> {
                                     let fixed_row = handle_toast_columns_for_cdc(
                                         &old_row_deserialized,
                                         &new_row_deserialized,
+                                        toastable_indices,
                                     );
                                     Bytes::from(row_serde.serializer.serialize(fixed_row))
                                 } else {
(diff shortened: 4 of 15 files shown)


src/stream/src/executor/mview/materialize.rs
```diff
@@ -418,13 +437,88 @@ impl<S: StateStore> MaterializeExecutor<S, BasicSerde> {
...
+            if let Ok(bytea_str) = std::str::from_utf8(bytea) {
+                is_unavailable_value_str(bytea_str)
+            } else {
+                false
+            }
+        }
+        Some(risingwave_common::types::ScalarRefImpl::List(list_ref)) => {
+            // For list type, check if it contains exactly one element with the unavailable value
+            // This is because when any element in an array triggers TOAST, Debezium treats the entire
+            // array as unchanged and sends a placeholder array with only one element
```
tabVersion, line 500: what if some elements change in the list? Debezium resends the whole list or like `["__debezium_unavailable_value", "new value", "__debezium_unavailable_value"]`

src/stream/src/executor/mview/materialize.rs
```diff
@@ -672,9 +790,23 @@ impl<SD: ValueRowSerde> MaterializeCache<SD> {
...
-                                let updated_row = OwnedRow::new(row_deserialized_vec);
+                                let mut updated_row = OwnedRow::new(row_deserialized_vec);
+
+                                // Apply TOAST column fix for CDC tables with TOAST columns
+                                if let Some(toastable_indices) = toastable_column_indices {
+                                    // Note: we need to use old_row_deserialized again, but it was moved above
+                                    // So we re-deserialize the old row
+                                    let old_row_deserialized_again =
+                                        row_serde.deserializer.deserialize(old_row.row.clone())?;
+                                    updated_row = handle_toast_columns_for_cdc(
```
tabVersion: since we will scan row to row for possible toast columns, I think we need a bench for related perf loss.

src/stream/src/executor/mview/materialize.rs
```diff
@@ -643,8 +738,31 @@ impl<SD: ValueRowSerde> MaterializeCache<SD> {
                                 true
                             };
                             if need_overwrite {
-                                change_buffer.update(key.clone(), old_row.row.clone(), row.clone());
-                                self.lru_cache.put(key.clone(), Some(CompactedRow { row }));
+                                let final_row = if let Some(toastable_indices) =
+                                    toastable_column_indices
+                                {
+                                    // For TOAST-able columns, replace Debezium's unavailable value placeholder with old row values.
+                                    let old_row_deserialized =
```
tabVersion: whats the behavior when old value looking up fails?

tabVersion (commented):
LSTM, I think it can be better to have a metric on each time a TOAST column is triggered, because it is a relative niche but important code path, we'd love more observability.

java/connector-node/risingwave-connector-service/src/main/resources/postgres.properties
```diff
@@ -27,3 +27,4 @@ heartbeat.action.query=SELECT pg_logical_emit_message(true, 'heartbeat', now()::
 name=${hostname}:${port}:${database.name}.${schema.name}.${table.name:-RW_CDC_Sharing}
 # In sharing cdc mode, transaction metadata will be enabled in frontend
 provide.transaction.metadata=${transactional:-false}
+unavailable.value.placeholder=__debezium_unavailable_value
```
hzxa21, line 32: We must disallow user to overwrite this option in create source.

proto/catalog.proto
```diff
@@ -520,6 +520,9 @@ message Table {
   // only exist when table_type is VECTOR_INDEX.
   optional VectorIndexInfo vector_index_info = 46;
 
+  // TOAST column indices for PostgreSQL CDC tables. Empty for non-CDC tables or CDC tables without TOAST columns.
+  repeated int32 toastable_column_indices = 47;
```
hzxa21: Do we really need to persist this? Can we derive it from the column type? Persisting this means we need to make sure it is consistent with other fields, which may cause complications when considering ADD/DROP COLUMN

src/common/src/types/jsonb.rs
```diff
@@ -147,6 +147,24 @@ impl std::str::FromStr for JsonbVal {
...
+    /// Create a `JsonbVal` from a string, with special handling for Debezium's unavailable value placeholder.
+    /// Returns a Result to handle parsing errors properly.
+    pub fn from_debezium_unavailable_value(s: &str) -> Result<Self, serde_json::Error> {
+        // Special handling for Debezium's unavailable value placeholder
+        if s.len() == Self::DEBEZIUM_UNAVAILABLE_VALUE.len()
+            && s == Self::DEBEZIUM_UNAVAILABLE_VALUE
+        {
+            let mut builder = jsonbb::Builder::default();
+            builder.add_string(Self::DEBEZIUM_UNAVAILABLE_VALUE);
+            return Ok(Self(builder.finish()));
```
hzxa21: Since the json unavialble value never changes, you don't need to build it every time. You can just build it once and reuse the result.

proto/catalog.proto
hzxa21: Note that you can argue that the toasted column type in PG may not be static but our supporting toasted column type is static. You can also save the extra discover_toastable_columns if we do so.

src/stream/src/executor/mview/materialize.rs
```diff
@@ -418,13 +437,88 @@ impl<S: StateStore> MaterializeExecutor<S, BasicSerde> {
...
+            toastable_column_indices: None,
             may_have_downstream: true,
             depended_subscription_ids: HashSet::new(),
             metrics,
         }
     }
 }
 
+/// Debezium's default unavailable value placeholder for TOAST columns.
+const DEBEZIUM_UNAVAILABLE_VALUE: &str = "__debezium_unavailable_value";
```
hzxa21: I saw this const in other place. Can we only have a single place to define the const?

wcy-fdu: Let's place it in common crate then.

java/connector-node/risingwave-connector-service/src/main/resources/postgres.properties
wcy-fdu: When we specific a config in properties, even if user set it in with option, we will use the value in properties file as the final result, such as `snapshot.mode`, this is because we use [`dbzProps.putIfAbsent(entry.getKey(), entry.getValue());`](https://github.com/risingwavelabs/risingwave/blob/52400e522cc7d70fe010f2bfe3e1debf0e8afeeb/java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/DbzConnectorConfig.java#L325) as safe guard. 

This is why I specific this config here, and I have simulated that if the user configured this parameter to other values in with, it would still use the default value we set. 


Do you mean we need more strong safe guard or add some comment here?

src/stream/src/executor/mview/materialize.rs
wcy-fdu: Old row must exist in hummock, materialized executor will first loop up it in cache and then loop up hummock if cache miss.

src/stream/src/executor/mview/materialize.rs
wcy-fdu: Debezium will resends the whole list, since this column is updated.

wcy-fdu:
> I think it can be better to have a metric on each time a TOAST column is triggered

I agree. Let's do it in follow up pr.

proto/catalog.proto
BugenZhao: > Can we derive it from the column type?

Or can we make it a field of a column?

src/stream/src/executor/mview/materialize.rs
BugenZhao, line 1: Wondering if there's any other to avoid inlining the logic for handling toast columns into `materialize.rs` (for example, put it inside `ChangeBuffer`), so that we can make the executor module cleaner?

proto/catalog.proto
wcy-fdu: Had some offline discussion with @hzxa21 , I agree put it in proto is not a good choice, since when add/drop column, the newly added `toastable_column_indices` will not take effect.

src/stream/src/executor/mview/materialize.rs
wcy-fdu: Yes, we are looking for the cleanest implementation.

src/meta/model/src/table.rs
```diff
@@ -315,6 +360,16 @@ impl From<PbTable> for ActiveModel {
...
+            cdc_table_type: Set(pb_table.cdc_table_type.map(|cdc_table_type| {
+                match cdc_table_type {
+                    0 => CdcTableType::Postgres, // Map Unspecified to Postgres as default
+                    1 => CdcTableType::Postgres,
+                    2 => CdcTableType::Mysql,
+                    3 => CdcTableType::Sqlserver,
+                    4 => CdcTableType::Mongo,
+                    _ => panic!("Invalid CDC table type: {cdc_table_type}"),
+                }
+            })),
```
yezizp2012, line 373:
```suggestion
            cdc_table_type: Set(pb_table.cdc_table_type().into()),
```

src/meta/model/migration/src/m20250820_120000_add_cdc_table_type.rs
```diff
@@ -0,0 +1,36 @@
+use sea_orm_migration::prelude::*;
+
+#[derive(DeriveMigrationName)]
+pub struct Migration;
+
+#[async_trait::async_trait]
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        manager
```
yezizp2012, line 9: Is it necessary to refill this column for historical cdc tables?

java/connector-node/risingwave-connector-service/src/main/resources/postgres.properties
hzxa21: > When we specific a config in properties, even if user set it in with option, we will use the value in properties file as the final result
I am unaware of of this. I think adding some comments here and in DbzConnectorConfig.java are sufficient.

src/common/src/types/jsonb.rs
```diff
@@ -21,11 +21,18 @@ use postgres_types::{FromSql, IsNull, ToSql, Type, accepts, to_sql_checked};
...
 use thiserror_ext::AsReport;
 
+use super::postgres_type::DEBEZIUM_UNAVAILABLE_VALUE;
 use super::{
     Datum, F64, IntoOrdered, ListValue, MapType, MapValue, ScalarImpl, StructRef, ToOwnedDatum,
 };
 use crate::types::{DataType, Scalar, ScalarRef, StructType, StructValue};
 use crate::util::iter_util::ZipEqDebug;
+// Pre-built JSON value for Debezium unavailable value to avoid rebuilding it every time
+static DEBEZIUM_UNAVAILABLE_JSON: std::sync::LazyLock<JsonbVal> = std::sync::LazyLock::new(|| {
```
hzxa21: nits: I think it is clearer to put this in src/types/mod.rs

src/common/src/types/postgres_type.rs
```diff
@@ -161,3 +161,5 @@ impl DataType {
         PgType::from_oid(oid as u32).unwrap()
     }
 }
+
+pub const DEBEZIUM_UNAVAILABLE_VALUE: &str = "__debezium_unavailable_value";
```
hzxa21: nits: I think it is clearer to put this in src/types/mod.rs

src/frontend/src/catalog/table_catalog.rs
```diff
@@ -595,6 +598,16 @@ impl TableCatalog {
...
             vector_index_info: self.vector_index_info,
+            cdc_table_type: self.cdc_table_type.clone().map(
+                |cdc_table_type| match cdc_table_type {
+                    CdcTableType::Undefined | CdcTableType::Mock => 0,
+                    CdcTableType::Postgres => 1,
+                    CdcTableType::MySql => 2,
+                    CdcTableType::SqlServer => 3,
+                    CdcTableType::Mongo => 4,
+                    CdcTableType::Citus => 5,
+                },
```
hzxa21: Let's impl the From trait to do the conversion between the pb type and our own type. Example: https://github.com/risingwavelabs/risingwave/blob/2f9c5f990bb6e1eeaf3c278ad90be2371492f81a/src/meta/model/src/table.rs#L86.

src/frontend/src/catalog/table_catalog.rs
```diff
@@ -817,8 +830,20 @@ impl From<PbTable> for TableCatalog {
...
+                .cdc_table_type
+                .map(|cdc_table_type| match cdc_table_type {
+                    0 => CdcTableType::Undefined,
+                    1 => CdcTableType::Postgres,
+                    2 => CdcTableType::MySql,
+                    3 => CdcTableType::SqlServer,
+                    4 => CdcTableType::Mongo,
+                    5 => CdcTableType::Citus,
+                    _ => panic!("Invalid CDC table type: {}", cdc_table_type),
+                }),
```
hzxa21: Let's impl the From trait to do the conversion between the pb type and our own type. Example: https://github.com/risingwavelabs/risingwave/blob/2f9c5f990bb6e1eeaf3c278ad90be2371492f81a/src/meta/model/src/table.rs#L86.

src/meta/src/controller/mod.rs
```diff
@@ -268,6 +270,10 @@ impl From<ObjectModel<table::Model>> for PbTable {
             clean_watermark_index_in_pk: value.0.clean_watermark_index_in_pk,
             refreshable: value.0.refreshable,
             vector_index_info: value.0.vector_index_info.map(|index| index.to_protobuf()),
+            cdc_table_type: value
```
hzxa21, line 273: ditto: you can use .into() if you implement the from trait.

src/stream/src/executor/backfill/cdc/cdc_backfill.rs
```diff
@@ -225,13 +225,20 @@ impl<S: StateStore> CdcBackfillExecutor<S> {
             .map(|v| v == "connect")
             .unwrap_or(false)
             .then_some(TimeHandling::Milli);
+        // Only postgres-cdc connector may trigger TOAST.
+        let handle_toast_columns: bool = self
+            .properties
+            .get("connector")
+            .map(|v| v == "postgres-cdc")
```
hzxa21: We have stored the cdc table type in the external table here: https://github.com/risingwavelabs/risingwave/blob/2f9c5f990bb6e1eeaf3c278ad90be2371492f81a/src/stream/src/executor/backfill/cdc/upstream_table/external.rs#L38. I think it is more robust to expose the cdc table type in external table and use that instead of matching the with properties here.

src/stream/src/executor/backfill/cdc/cdc_backill_v2.rs
```diff
@@ -150,12 +150,19 @@ impl<S: StateStore> ParallelizedCdcBackfillExecutor<S> {
             .map(|v| v == "connect")
             .unwrap_or(false)
             .then_some(TimeHandling::Milli);
+        // Only postgres-cdc connector may trigger TOAST.
+        let handle_toast_columns: bool = self
```
hzxa21: ditto

src/stream/src/executor/mview/materialize.rs
```diff
@@ -70,6 +72,12 @@ pub struct MaterializeExecutor<S: StateStore, SD: ValueRowSerde> {
     /// No data will be written to hummock table. This Materialize is just a dummy node.
     /// Used for APPEND ONLY table with iceberg engine. All data will be written to iceberg table directly.
     is_dummy_table: bool,
+
+    /// Indices of TOAST-able columns for PostgreSQL CDC tables. None means either non-CDC table or CDC table without TOAST-able columns.
+    toastable_column_indices: Option<Vec<usize>>,
+
+    /// Whether to handle TOAST for PostgreSQL CDC tables.
+    handle_toast: bool,
```
hzxa21: this seems unnessary because we can use the None check on `toastable_column_indices`.

src/stream/src/executor/mview/materialize.rs
```diff
@@ -106,12 +114,36 @@ impl<S: StateStore, SD: ValueRowSerde> MaterializeExecutor<S, SD> {
         version_column_index: Option<u32>,
         metrics: Arc<StreamingMetrics>,
     ) -> Self {
+        // Determine if we should handle TOAST based on CDC table type
+        // Only handle TOAST for PostgreSQL CDC tables.
+        let handle_toast = table_catalog.cdc_table_type == Some(1);
```
hzxa21: a bit hacky to use `Some(1)` here. You can call the pb method `table_catalog.cdc_table_type()` to get back the enum type.

src/stream/src/executor/mview/materialize.rs
hzxa21: Actually you don't need `toastable_column_indices: Option<Vec<usize>>`. You can just use a vec `toastable_column_indices: Vec<usize>`

src/stream/src/executor/mview/materialize.rs
```diff
@@ -106,12 +114,36 @@ impl<S: StateStore, SD: ValueRowSerde> MaterializeExecutor<S, SD> {
...
             .map(|col| col.column_desc.as_ref().unwrap().into())
             .collect();
 
+        // Extract TOAST-able column indices from table columns.
+        let toastable_column_indices = {
+            let toastable_indices: Vec<usize> = table_columns
+                .iter()
+                .enumerate()
+                .filter_map(|(index, column)| match &column.data_type {
+                    DataType::Varchar | DataType::List(_) | DataType::Bytea | DataType::Jsonb => {
```
hzxa21, line 136: Better add some comments here on why we only support these types and what toastable types we don't support yet.

hzxa21: You can just mention check comments in `is_debezium_unavailable_value` here.

src/stream/src/executor/mview/materialize.rs
```diff
@@ -651,12 +762,32 @@ impl<SD: ValueRowSerde> MaterializeCache<SD> {
...
+                                    && let Some(toastable_indices) = toastable_column_indices
+                                {
+                                    // For TOAST-able columns, replace Debezium's unavailable value placeholder with old row values.
+                                    handle_toast_columns_for_postgres_cdc(
+                                        &old_row_deserialized,
+                                        &new_row_deserialized,
+                                        toastable_indices,
+                                    )
+                                } else {
+                                    new_row_deserialized.clone()
```
hzxa21: `clone` is not needed here. Kindly reminder: cloning a OwnedRow is a deep copy.

src/stream/src/executor/mview/materialize.rs
```diff
@@ -651,12 +762,32 @@ impl<SD: ValueRowSerde> MaterializeCache<SD> {
...
+                                };
+
                                 change_buffer.update(
                                     key.clone(),
                                     old_row_deserialized,
-                                    new_row_deserialized,
+                                    final_row.clone(),
+                                );
+                                let final_row_bytes =
+                                    Bytes::from(row_serde.serializer.serialize(final_row));
```
hzxa21: Compared with the logics prior to this PR, this introduces an extra serialization even if there is no toast value. I suggest we get both the bytes row and owned row in the if clause above in L765 so that we can pass the byte `row` directly instead of doing the unneccessary serialization when there is no toast value.

src/stream/src/executor/mview/materialize.rs
wcy-fdu: > You can just use a vec toastable_column_indices: Vec<usize>

We need to first check if it's postgres-cdc, and then check the type. If there's only one `toastable_column_indices: Vec<usize>`, it won't skip when using other cdc. I keep two flags to distinguish `pg-cdc` from others.
Let's keep `handle_toast: bool `and change `toastable_column_indices` from Option to Vec<usize> or remove  `handle_toast: bool ` and keep toastable_column_indices Option.

src/stream/src/executor/mview/materialize.rs
wcy-fdu: Good catch!

src/meta/model/migration/src/m20250820_120000_add_cdc_table_type.rs
wcy-fdu: I don't think it's necessary. The history table will be filled with null, which means the toast is not checked(we only check toast when this column is Some(Postgres)). This will not have any impact on the already created CDC table.

src/frontend/src/catalog/table_catalog.rs
```diff
@@ -860,6 +842,40 @@ impl OwnedByUserCatalog for TableCatalog {
...
+            PbCdcTableType::Unspecified => CdcTableType::Undefined,
+            PbCdcTableType::Postgres => CdcTableType::Postgres,
+            PbCdcTableType::Mysql => CdcTableType::MySql,
+            PbCdcTableType::Sqlserver => CdcTableType::SqlServer,
+            PbCdcTableType::Mongo => CdcTableType::Mongo,
+            PbCdcTableType::Citus => CdcTableType::Citus,
+        }
+    }
+}
+
```
hzxa21: This is inconsistent with our general practice to do conversion between pb type and our own type. Let's impl the From trait. Also, we shouldn't use `i32` directly and instead it is better to use the pb type.

Example: https://github.com/risingwavelabs/risingwave/blob/2f9c5f990bb6e1eeaf3c278ad90be2371492f81a/src/meta/model/src/table.rs#L86

wcy-fdu: The reason for the ambiguity here is that I did not actually use the CdcTableType generated by proto, but used a [CdcTableType](https://github.com/risingwavelabs/risingwave/blob/124f3056a61719850912917ac6356d3d6be45a3f/src/connector/src/source/cdc/external/mod.rs#L50) in connector crate that existed in the previous code for the purpose of convenience(It is convenient to get the type when creating a table reader.).  

For meta crate `CdcTableType`, I already `impl From<CdcTableType> for PbCdcTableType `
Let me implement a conversion from connector cdc_type to meta cdc_type.

wcy-fdu: Let's rename the origin `CdcTableType` to `ExternalCdcTableType` to make this transition clearer.

PR #11455, opened 2023-08-04
perf(stream): optimize `state_table::write_chunk`

kwannoel:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://gist.github.com/TennyZhuang/f00be7f16996ea48effb049aa7be4d66#file-rw_cla).

## What's changed and what's your intention?

Closes https://github.com/risingwavelabs/risingwave/issues/11377

Code is also simplified.

## Checklist

- [ ] I have written necessary rustdoc comments
- [ ] I have added necessary unit tests and integration tests
- [ ] I have added fuzzing tests or opened an issue to track them. (Optional, recommended for new SQL features #7934).
- [ ] My PR contains breaking changes. (If it deprecates some features, please create a tracking issue to remove them in the future).
- [ ] All checks passed in `./risedev check` (or alias, `./risedev c`)
- [ ] My PR changes performance-critical code. (Please run macro/micro-benchmarks and show the results.)

- [ ] My PR contains critical fixes that are necessary to be merged into the latest release. (Please check out the [details](https://github.com/risingwavelabs/risingwave/blob/main/CONTRIBUTING.md))

## Documentation

- [ ] My PR contains user-facing changes.

<details><summary>Click here for Documentation</summary>

### Types of user-facing changes

Please keep the types that apply to your changes, and remove the others.

- Installation and deployment
- Connector (sources & sinks)
- SQL commands, functions, and operators
- RisingWave cluster configuration changes
- Other (please specify in the release note below)

### Release note

</details>

PR #11415, opened 2023-08-03
test(stream): add more state table bench for chunks

kwannoel:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://gist.github.com/TennyZhuang/f00be7f16996ea48effb049aa7be4d66#file-rw_cla).

## What's changed and what's your intention?

Related https://github.com/risingwavelabs/risingwave/issues/11377

Add deletes as well, and with different visibilities.

NOTE: The generator currently does not keep track of keys, and derive deletes from them. It just generates random deletes and inserts. Meaning pk could also overlap between inserts, and deletes could be deleting invalid entries.

To circumvent it we initialize state table without consistency checks:
```
    StateTable::new_without_distribution_inconsistent_op(
```

It is more ideal to have state table generate consistent op + rows, but I think it is fine for now.

## Checklist

- [ ] I have written necessary rustdoc comments
- [ ] I have added necessary unit tests and integration tests
- [ ] I have added fuzzing tests or opened an issue to track them. (Optional, recommended for new SQL features #7934).
- [ ] My PR contains breaking changes. (If it deprecates some features, please create a tracking issue to remove them in the future).
- [ ] All checks passed in `./risedev check` (or alias, `./risedev c`)
- [ ] My PR changes performance-critical code. (Please run macro/micro-benchmarks and show the results.)

- [ ] My PR contains critical fixes that are necessary to be merged into the latest release. (Please check out the [details](https://github.com/risingwavelabs/risingwave/blob/main/CONTRIBUTING.md))

## Documentation

- [ ] My PR contains user-facing changes.

<details><summary>Click here for Documentation</summary>

### Types of user-facing changes

Please keep the types that apply to your changes, and remove the others.

- Installation and deployment
- Connector (sources & sinks)
- SQL commands, functions, and operators
- RisingWave cluster configuration changes
- Other (please specify in the release note below)

### Release note

</details>

Issue #11377, opened 2023-08-01
perf: optimize `StateTable::write_chunk`


kwannoel:
1. We should check vis before serializing chunks.
2. Values should be treated the same way as key chunks. If we skip key datum serialization because of viz, we should do so for values as well.
3. After serializing chunks, we can immediately insert them, rather than check vis again.

We need better benchmarks for `write_chunk`:
1. different vis: 0%, 1%, 10%, 25%
2. different op, just insert and deletes will do for now.

kwannoel:
So far PR #11455  shows that there is not really a general perf improvement. Welcome any comments.

PR #22853, opened 2025-08-11
refactor(state-table): extract logic of row store in state table

wenym1:

Extract the logic of `StateTableRowStore` from `StateTable`, which serves a similar functionality as a `BTreeMap<TableKey<Bytes>, OwnedRow>`, and provides method to read (`get` and `iter`) over serialized key (or key range) and returns `OwnedRow`, and write (`insert`, `delete`, `update`) on `(TableKey<Bytes>, OwnedRow)`.

With this `StateTableRowStore`, in future PR we can optionally preload at initialization and store all rows in memory, and always read and write on the in memory rows to avoid latency spike caused on IO when reading from state store.

Besides, the `StateTable::get_encoded_row` is only used in `append_only_dedup`, and only as the functionality as `exists`, and therefore we change `get_encoded_row` to `exists`, to avoid an unnecessary clone on the encoded values.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/common/table/state_table.rs
+++ b/src/stream/src/common/table/state_table.rs
@@ -17,12 +17,12 @@ use std::ops::Bound;
 use std::ops::Bound::*;
 use std::sync::Arc;
 
-use bytes::{BufMut, Bytes, BytesMut};
+use bytes::Bytes;
 use either::Either;
 use foyer::Hint;
 use futures::{Stream, StreamExt, TryStreamExt, pin_mut};
 use futures_async_stream::for_await;
-use itertools::{Itertools, izip};
+use itertools::Itertools;
 use risingwave_common::array::stream_record::Record;
 use risingwave_common::array::{ArrayImplBuilder, ArrayRef, DataChunk, Op, StreamChunk};
 use risingwave_common::bitmap::Bitmap;
@@ -34,7 +34,6 @@ use risingwave_common::row::{self, Once, OwnedRow, Row, RowExt, once};
 use risingwave_common::types::{DataType, Datum, DefaultOrd, DefaultOrdered, ScalarImpl};
 use risingwave_common::util::column_index_mapping::ColIndexMapping;
 use risingwave_common::util::epoch::EpochPair;
-use risingwave_common::util::iter_util::{ZipEqDebug, ZipEqFast};
 use risingwave_common::util::row_serde::OrderedRowSerde;
 use risingwave_common::util::sort_util::OrderType;
 use risingwave_common::util::value_encoding::BasicSerde;
@@ -58,9 +57,7 @@ use risingwave_storage::row_serde::row_serde_util::{
 use risingwave_storage::row_serde::value_serde::ValueRowSerde;
 use risingwave_storage::store::*;
 use risingwave_storage::table::merge_sort::merge_sort;
-use risingwave_storage::table::{
-    ChangeLogRow, KeyedRow, TableDistribution, deserialize_log_stream,
-};
+use risingwave_storage::table::{KeyedRow, TableDistribution};
 use thiserror_ext::AsReport;
 use tracing::{Instrument, trace};
 
@@ -87,7 +84,6 @@ macro_rules! insane_mode_discard_point {
 
 /// `StateTableInner` is the interface accessing relational data in KV(`StateStore`) with
 /// row-based encoding.
-#[derive(Clone)]
 pub struct StateTableInner<
     S,
     SD = BasicSerde,
@@ -101,7 +97,7 @@ pub struct StateTableInner<
     table_id: TableId,
 
     /// State store backend.
-    local_store: S::Local,
+    row_store: StateTableRowStore<S::Local, SD>,
 
     /// State store for accessing snapshot data
     store: S,
@@ -111,9 +107,6 @@ pub struct StateTableInner<
 
     /// Used for serializing and deserializing the primary key.
     pk_serde: OrderedRowSerde,
-
-    /// Row deserializer with value encoding
-    row_serde: Arc<SD>,
 
     /// Indices of primary key.
     /// Note that the index is based on the all columns of the table, instead of the output ones.
@@ -128,9 +121,6 @@ pub struct StateTableInner<
     distribution: TableDistribution,
 
     prefix_hint_len: usize,
-
-    /// Used for catalog `table_properties`
-    table_option: TableOption,
 
     value_indices: Option<Vec<usize>>,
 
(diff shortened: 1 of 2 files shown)


wenym1:
* **#22857** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22857?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#22853** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22853?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22853?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/stream/src/common/table/state_table.rs
```diff
@@ -922,74 +1005,27 @@ where
...
-
-        let values = if let Some(ref value_indices) = self.value_indices {
-            chunk
-                .project(value_indices)
-                .serialize_with(&*self.row_serde)
-        } else {
-            chunk.serialize_with(&*self.row_serde)
-        };
-
-        // TODO(kwannoel): Seems like we are doing vis check twice here.
```
hzxa21, line 939: @kwannoel  Is there any reason why we do a two pass iteration here, like iterating the pk serialization first and then iter pks with the values later? Is it because of vectorization? I am trying to see whether this [comment](https://github.com/risingwavelabs/risingwave/pull/22853/files#r2266106117)  is valid.

kwannoel: It led to performance regression previously. Here was my optimization attempt: https://github.com/risingwavelabs/risingwave/pull/11455.

Related issue: https://github.com/risingwavelabs/risingwave/issues/11377

wenym1: @kwannoel What's the schema of data chunk in your benchmark?

wenym1: The only computation possibly in a batch is `compute_chunk_vnode` and `chunk.serialize_with(...)`.

For `chunk.serialize_with(...)`, it simply iter over the rows and serialize them one by one
```
pub fn serialize_with(&self, serializer: &impl ValueRowSerializer) -> Vec<Bytes> {
    let mut results = Vec::with_capacity(self.capacity());
    for row in self.rows_with_holes() {
        results.push(if let Some(row) = row {
            serializer.serialize(row).into()
        } else {
            Bytes::new()
        });
    }
    results
}
```

The only potential performance gain from computing batch in `compute_chunk_vnode`,
```
pub fn compute_chunk(
    data_chunk: &DataChunk,
    keys: &[usize],
    vnode_count: usize,
) -> Vec<VirtualNode> {
    if let Ok(idx) = keys.iter().exactly_one()
        && let ArrayImpl::Serial(serial_array) = &**data_chunk.column_at(*idx)
    {
        return serial_array
            .iter()
            // ... compute vnode from row_id row by row
    }

    data_chunk
        .get_hash_values(keys, Crc32FastBuilder)
        .into_iter()
        .map(|hash| hash.to_vnode(vnode_count))
        .collect()
}
```

where `data_chunk.get_hash_value` is computed by `array.hash_vec(..)`, which benefits from vectorization.

To summarize, we may benefit from vectorization when we compute vnode. The rest should make little difference. I will change to compute vnode in batch in this PR.

kwannoel: > @kwannoel What's the schema of data chunk in your benchmark?

Should be this: https://github.com/risingwavelabs/risingwave/pull/11415/files#r2273380135

kwannoel: We can just trigger nexmark bench for this PR, to evaluate it.

wenym1:
Applied https://github.com/risingwavelabs/risingwave/pull/22853#discussion_r2273323957 to vectorize the vnode computation. Hope to include this feature in release-2.6 @hzxa21 @kwannoel PTAL

kwannoel:
> Applied [#22853 (comment)](https://github.com/risingwavelabs/risingwave/pull/22853#discussion_r2273323957) to vectorize the vnode computation. Hope to include this feature in release-2.6 @hzxa21 @kwannoel PTAL

Can we:
1. Run the micro benchmarks mentioned in the comment.
2. Build a docker image via: https://github.com/risingwavelabs/risingwave/actions/workflows/build-docker-image.yml
3. Trigger perf test via: https://github.com/risingwavelabs/risingwave/actions/workflows/run-perf-test.yml.

Perf test should be straightforward to trigger now.

src/stream/src/common/table/state_table.rs
```diff
@@ -546,7 +611,7 @@ where
     }
 
     /// Get the vnode value with given (prefix of) primary key
-    fn compute_prefix_vnode(&self, pk_prefix: impl Row) -> VirtualNode {
+    fn compute_prefix_vnode(&self, pk_prefix: &impl Row) -> VirtualNode {
```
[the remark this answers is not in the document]
BugenZhao: Is this change necessary? `<R: Row> &R` also implements `Row`.

BugenZhao (approved):

src/stream/src/common/table/state_table.rs
```diff
@@ -243,6 +233,77 @@ fn consistent_old_value_op(
...
+                let $row_var_name = $row_var_name.project(value_indices);
+            )+
+            $body
+        } else {
+            $body
+        }
+    };
+}
+
+struct StateTableRowStore<LS: LocalStateStore, SD: ValueRowSerde> {
```
BugenZhao, line 249: Shall we also add some docs/comments?

src/stream/src/common/table/state_table.rs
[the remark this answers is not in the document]
wenym1: I'm not sure if it's a good practice to leverage `impl<R: Row> Row for &R` to always declare the param as `impl Row`. We may keep generating a copy of code for a concrete type with multiple level of reference like `&&&&&&OwnedRow`, which may increase the binary size and compile time.

wenym1:
> > Applied [#22853 (comment)](https://github.com/risingwavelabs/risingwave/pull/22853#discussion_r2273323957) to vectorize the vnode computation. Hope to include this feature in release-2.6 @hzxa21 @kwannoel PTAL
> 
> Can we:
> 
> 1. Run the micro benchmarks mentioned in the comment.
> 2. Build a docker image via: https://github.com/risingwavelabs/risingwave/actions/workflows/build-docker-image.yml
> 3. Trigger perf test via: https://github.com/risingwavelabs/risingwave/actions/workflows/run-perf-test.yml.
> 
> Perf test should be straightforward to trigger now.

Failed to run the benchmark in the initial setting phase. Will merge this PR first and run the test later. If the test shows this PR greatly affect the performance, we can revert back to vectorized implementation.

src/stream/src/common/table/state_table.rs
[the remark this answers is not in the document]
BugenZhao: TRUE.

PR #22856, opened 2025-08-11
feat(frontend): support creating upsert source & basic ETL usage

BugenZhao:

Background:
- #22799 
- #22902 

This PR adds functionality for creating upsert source (without materializing it with a `TABLE`) and its basic ETL usage, demonstrated by pre-processing the data and sinking into a table in e2e tests.

## Checklist

## Documentation

<details>

</details>


BugenZhao:
* **#23211** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23211?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#22856** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22856?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22856?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#23209** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23209?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

e2e_test/source_inline/kafka/upsert_source.slt
```diff
@@ -0,0 +1,93 @@
...
+CREATE TABLE students (
+    id INT primary key,
+    name VARCHAR,
+    age INT
+)
+
+statement ok
+CREATE SINK sk INTO students AS
+  SELECT
+    (convert_from(rw_key, 'UTF8')::jsonb -> 'ID')::int AS id,
```
BugenZhao: Just realized: we must extract `id` from the key but cannot simply use `id` field in the payload here, because for `DELETE` records from upsert Kafka, we get empty payload.

This also means that we cannot assume `Project` or `Filter` work perfectly on `Upsert` stream... unless the referenced columns are functional dependencies of the primary key column. Otherwise, we need to force `MaterializedExprs` here. 🤣 Seems the usage pattern is much more limited than I thought initially. cc @chenzl25 @hzxa21

hzxa21: I thought by definition `upsert` stream means the delete message only contains the key, not other fields. 
I think it makes sense to not allow
```sql
create ..... as
select * from upsert_source where <predicates on non_keyed_col>
```

hzxa21: Second thought: is it incorrect to pass through the key-only delete in `Project` or `Filter` without considering the exprs?

This looks correct to me:
```
upsert_source (key, col1):
+ key=k1, col1=v1
+ key=k1, col1=v2
- key=k1


sink into table t with pk=key 
(select key, sub_str(col1, ...) as t_col1 from upsert_source where col1 = v1
+ key=k1, t_col1=sub_str(col1, ...)
- key=k1

```

BugenZhao: Do you mean always propagating `DELETE` record to downstream regardless of predicate? Interesting...

hzxa21: Roughly yes but I don't think we should propagate it to all downstream. For example, it is not possible to sink a upsert source `s` into a table `t` whose PK is not the same as `s`, because there is no way to reason about the delete message behavior.

chenzl25: I think we have two approaches to handle this issue.
1. Let the stateless operators like `Filter`, `Project` handle the upsert delete as is. From the downstream operator's view, the delete message of the `Upsert` stream would delete a stream key that never exists.
2. Materialize the pk of an upsert stream, so that we can eliminate the non-existing delete and make it a bit more consistent. The `MaterializePk` operator should be added to the plan appropriately without sacrificing extra performance.

Somehow, it forces us to define what an upsert stream is. Do we allow deleting a non-existing key, or must we delete an existing key?

tabVersion: > Do you mean always propagating DELETE record to downstream regardless of predicate?

I think we can do nothing but always propagating. Upsert source ingests key as bytea, we have no way to deserialize it and change accordingly. So if we want to align pk with the downstream, seems we should forbid doing ops on pk fields (but we never know what the fields are)

BugenZhao: > whose PK is not the same as `s`, because there is no way to reason about the delete message behavior

Yes, and that's why in the given tests I extract the new key from the original key, instead of directly using a field from the payload.

BugenZhao: > I think we have two approaches to handle this issue.

Just a rough thought: it appears that all executors yielding data with a different set of columns as stream key will reject upsert stream as input? If so, there doesn't seem to an issue if we always directly propagate `Delete` records, as at the moment when it's converted to a retract stream, we always materialize all columns and can eliminate such inconsistency.

BugenZhao:
This PR is ready for review again, based on the new behavior of `UpsertFilter` introduced in https://github.com/risingwavelabs/risingwave/pull/23209 according to former discussion.

e2e_test/source_inline/kafka/upsert_source.slt
```diff
@@ -0,0 +1,119 @@
...
+  SELECT
+    "ID",
+    "firstName" || ' ' || "lastName" AS name,
+    age
+  FROM upsert_students;
+----
+db error: ERROR: Failed to run the query
+
+Caused by:
+  Bind error: When sinking from an upsert stream, the downstream primary key must be the same as or a subset of the one derived from the stream.
```
tabVersion, line 117: out of curious, how we do the check when performing CREATE SINK INTO TABLE

e2e_test/source_inline/kafka/upsert_source.slt
```diff
@@ -0,0 +1,119 @@
...
+    "lastName" VARCHAR,
+    age INT,
+    height REAL,
+    weight REAL,
+)
+INCLUDE KEY AS rw_key
+WITH (
+ ${RISEDEV_KAFKA_WITH_OPTIONS_COMMON},
+ topic = 'test_include_key')
+FORMAT UPSERT ENCODE JSON
```
tabVersion, line 24: can you also check `show create source upsert_students` for correct format?

tabVersion (approved):
rest LGTM

e2e_test/source_inline/kafka/upsert_source.slt
BugenZhao: The "specified key" will be the primary key defined on the target table.

chenzl25 (approved):
LGTM!

hzxa21:
Added `user-facing-changes` label because after this PR, we allow `CREATE SOURCE` and defining primary key on `FORMAT UPSERT`.

@BugenZhao Can you help update the release note section in the PR description as well?

BugenZhao:
> Can you help update the release note section in the PR description as well?

Updated. Also cc @WanYixian

PR #22846, opened 2025-08-10
fix backpressure_rate normalization

kwannoel:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## Checklist

- [ ] I have written necessary rustdoc comments.
- [ ]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  I have checked the [Release Timeline](https://github.com/risingwavelabs/rw-commits-history/blob/main/release_timeline.md) and [Currently Supported Versions](https://docs.risingwave.com/changelog/release-support-policy#support-end-dates-for-recent-releases) to determine which release branches I need to cherry-pick this PR into. 

## Documentation

- [ ]  My PR needs documentation updates. 

<details>
<summary><b>Release note</b></summary>

</details>

PR #22868, opened 2025-08-12
fix m x n actor profiling

kwannoel:

### M x N channels, but only divide by M

Edit: This isn't actually a problem. Let's say we have N actors for a fragments. Means we have N dispatchers. For each dispatcher, we will get the time it takes to shuffle all data into the M downstreams. This is the `max` duration it takes, to shuffle all the data. So M has already been implicitly aggregated in this way. We only need to divide by N.

Edit2: We also have to consider K worker nodes, given that worker nodes each have an independent set of streaming metrics, we need to aggregate backpressure rate together. We currently do this, so it's fine.

The secondary issue is that we have M x N channels for dispatchers. M for upstream actors, N for downstream actors. Currently we take the sum of the yielding duration for all channels, as the blocking duration, and divide it by M. This neglects the N factor. 

However, if we divide by N as well, when there's a hot partition, the blocking duration might show up as 1/N. For instance, if N=4, the blocking duration can be 25%. This is because there are 3 non-bottleneck partitions and 1 bottleneck partitions. In such a scenario, all the 3 non-bottleneck partitions have 0% backpressure, and the bottleneck partition has 25%.

We should update the implementation to use the blocking duration of the slowest partition. i.e. the max backpressure of all the channels.

This is because the slowest channel will force the other channels to wait for it. If this is too complicated, we should refactor the metrics collection to collect the total duration of a yield at the `dispatch executor level`, rather than concurrently within each `dispatch output level`.

## Checklist

## Documentation

<details>

</details>


kwannoel:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22868?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#22868** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22868?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22868?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#22846** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22846?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#22843** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22843?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#22838** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22838?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#22839](https://github.com/risingwavelabs/risingwave/pull/22839) <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22839?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#22802** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22802?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#22801** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22801?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #22874, opened 2025-08-12
fix: float comparison in vector::distance tests

### LINKED ISSUE #22872: Flaky unit test in `vector::distance::tests::test_distance` due to floating-point comparison issues
The local unit test `vector::distance::tests::test_distance` intermittently fails due to small floating-point differences in distance calculations.

**Observed Failure Example::**
`Expected: 31.531513, Actual: 31.531525
`
Difference of ~1.2e-5

This is caused by two problems in the floating-point assertion macro in src/storage/src/vector/distance.rs (test module):

1. Logical bug
- `($first - $second) < FLOAT_ALLOWED_BIAS`
- It does not use .abs(), so large negative differences could incorrectly pass. For example, if the actual value is much less than expected, the difference is negative and still < FLOAT_ALLOWED_BIAS.

2. Overly tight epsilon
- The current FLOAT_ALLOWED_BIAS is 1e-5.
- Some distances can differ by ~1.2e-5 due to expected floating-point drift (SIMD/FMA/reduction ordering).

**Proposed fix:**
- Update the macro to compare the absolute difference.
- Slightly increase tolerance or add a small relative epsilon to account for normal numerical drift.

**Additional Context:**
This started failing more frequently after updating CMake, likely because FAISS build flags and math kernels changed. However, the tolerance/sign issue is independent of the build environment and should be addressed regardless.

**Impact:**
Fixing this will make the test more robust to harmless floating-point variation and prevent false negatives in CI and local runs.

ChrisWrenDev:

issue: resolve #22872 

Tests under src/storage/src/vector/distance.rs used a signed difference to compare floating-point values:
`($first - $second) < FLOAT_ALLOWED_BIAS`

This is asymmetric and can let large negative differences pass. After upgrading CMake I also observed a drift (~1.2e-5) that slightly exceeds the current 1e-5 epsilon causing the test to fail locally.

Observed failure: `Expected: 31.531513, Actual: 31.531525  (|Δ| ≈ 1.2e-5)`

**This PR:**
- Fixes the logic to compare absolute difference.
- Uses a combined absolute + relative tolerance to remain strict yet robust across SIMD/FMA/BLAS/FAISS code paths.

**Changes:**
- Replace the signed comparison with abs() on the difference, this measures the magnitude of the difference, so both positive and negative deviations are caught.
- Introduce FLOAT_ABS_EPS = 2e-5 and FLOAT_REL_EPS = 1e-6 to allow for expected noise while still detecting genuine errors.
- The tolerance is computed as the larger of the absolute floor (FLOAT_ABS_EPS) and a scaled relative term (FLOAT_REL_EPS * max(|a|, |b|)). This relative component adjusts the allowed difference based on the magnitude of the values, ensuring the comparison remains fair and consistent across both small and large numbers.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/storage/src/vector/distance.rs
+++ b/src/storage/src/vector/distance.rs
@@ -235,17 +235,24 @@ mod tests {
         0.22877127, 0.97690505, 0.44438475,
     ];
 
-    const FLOAT_ALLOWED_BIAS: f32 = 1e-5;
+    const FLOAT_ABS_EPS: f32 = 2e-5;
+    const FLOAT_REL_EPS: f32 = 1e-6;
 
     macro_rules! assert_eq_float {
-        ($first:expr, $second:expr) => {
+        ($first:expr, $second:expr) => {{
+            let a: f32 = $first;
+            let b: f32 = $second;
+            let diff = (a - b).abs();
+            let tol = FLOAT_ABS_EPS.max(FLOAT_REL_EPS * a.abs().max(b.abs()));
             assert!(
-                ($first - $second) < FLOAT_ALLOWED_BIAS,
-                "Expected: {}, Actual: {}",
-                $second,
-                $first
+                diff <= tol,
+                "Expected: {}, Actual: {}, |Δ|={} > tol={}",
+                b,
+                a,
+                diff,
+                tol
             );
-        };
+        }};
     }
 
     #[test]


ChrisWrenDev:
@hzxa21, @wenym1— tagging you for review. Thanks!

PR #22878, opened 2025-08-13
fix(streaming): avoid `unwrap` when reading state

kwannoel:

State reads might have transient issues, if object store connection is flaky. 

> @zwang28: Panic is bad. Previously it panic, and write core dump, and further crash the K8S node since the core dump is too large, and further crash other pods in the same nodes, and so on..

For unrecoverable errors (missing state table) I use `expect` instead, and detail why it should be present.

v.2.4.3 branch: https://github.com/risingwavelabs/risingwave/pull/22879

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/executor/join/hash_join.rs
+++ b/src/stream/src/executor/join/hash_join.rs
@@ -336,8 +336,7 @@ async fn fetch_degrees<K: HashKey, S: StateStore>(
     let sub_range: &(Bound<OwnedRow>, Bound<OwnedRow>) = &(Bound::Unbounded, Bound::Unbounded);
     let table_iter = degree_state_table
         .iter_with_prefix(key, sub_range, PrefetchOptions::default())
-        .await
-        .unwrap();
+        .await?;
     #[for_await]
     for entry in table_iter {
         let degree_row = entry?;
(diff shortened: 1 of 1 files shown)


src/stream/src/executor/join/hash_join.rs
```diff
@@ -927,7 +935,7 @@ impl<K: HashKey, S: StateStore, E: JoinEncoding> JoinHashMap<K, S, E> {
     pub fn serialize_inequal_key_from_row(&self, row: impl Row) -> InequalKeyType {
         self.inequality_key_desc
             .as_ref()
-            .unwrap()
+            .expect("inequality key desc missing")
```
kwannoel, line 938: Don't want to change the API, it's used by ASOF Join.

src/stream/src/executor/join/hash_join.rs
```diff
@@ -1071,10 +1079,10 @@ impl<E: JoinEncoding> JoinEntryState<E> {
             }
         } else {
             let mut pk_set = JoinRowSet::default();
-            pk_set.try_insert(pk, ()).unwrap();
+            pk_set.try_insert(pk, ()).expect("pk set should be empty");
             self.inequality_index
                 .try_insert(inequality_key, pk_set)
-                .unwrap();
+                .expect("pk set should be empty");
```
kwannoel, line 1085: Don't want to change the signature, it's used by ASOF Join. Just add an expect here.

src/stream/src/executor/join/hash_join.rs
```diff
@@ -927,7 +935,7 @@ impl<K: HashKey, S: StateStore, E: JoinEncoding> JoinHashMap<K, S, E> {
     pub fn serialize_inequal_key_from_row(&self, row: impl Row) -> InequalKeyType {
         self.inequality_key_desc
             .as_ref()
-            .unwrap()
+            .expect("inequality key desc missing")
             .serialize_inequal_key_from_row(&row)
```
kwannoel, line 936: ASOF JOIN functions. If `inequality_key_desc` is empty, it will fail early. We can just use `expect` here.

Issue #22885, opened 2025-08-13
Restrict `serde` dependencies to `derive` feature


ChrisWrenDev:
In the RisingWave codebase, most serde dependencies are already restricted to the derive feature to minimize the dependency footprint. However, there are still a few instances where serde is declared without explicitly setting its features.

This task is to update the remaining serde dependency declarations to use `features = ["derive"]`.

ChrisWrenDev:
I will raise a PR to update the relevant Cargo.toml files

LorenzoTettamanti:
Hi @ChrisWrenDev , if you need any help i would like to help you completing this issue 😊

ChrisWrenDev:
@LorenzoTettamanti A pull request has already been raised for this issue 👍

LorenzoTettamanti:
@ChrisWrenDev it's okay, thanks anyway for your reply!

PR #22886, opened 2025-08-13
build: restrict serde to derive feature

ChrisWrenDev:

Issue: #22885

Updated the remaining `serde` dependencies to explicitly include `features = ["derive"]`, bringing them in line with the rest of the project.
This change ensures consistency, avoids pulling in unnecessary `serde` features, and limits our direct usage to only the `Serialize` and `Deserialize` derive macros.

## Checklist

## Documentation

<details>

</details>


ChrisWrenDev:
@kwannoel, @hzxa21 — tagging you for review. Thanks!

src/common/secret/Cargo.toml
```diff
@@ -17,7 +17,7 @@ parking_lot = { workspace = true }
 prost = { workspace = true }
 reqwest = { workspace = true }
 risingwave_pb = { workspace = true }
-serde = { version = "1" }
+serde = { version = "1", features = ["derive"] }
```
wenym1: Shall we move the depended crate declaration to the root workspace `Cargo.toml`, and change the dependency in the `Cargo.toml` in all concrete crates to `serde = { workspace = true }`?

ChrisWrenDev: Yeah that's a good idea, I've just updated all the serde dependencies to use the workspace Cargo.toml :+1:

PR #22820, opened 2025-08-07
feat(vector): clear remaining vector place holders

wenym1:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## What's changed and what's your intention?

As titled.

Most previous `VECTOR_PLACEHOLDER` is for converting a datum of `Vector` type to something used externally. Therefore, we treat the vector type as if `DataType::List(DataType::Float32)`, and the logic follows it.

## Checklist

- [ ] I have written necessary rustdoc comments.
- [ ]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  I have checked the [Release Timeline](https://github.com/risingwavelabs/rw-commits-history/blob/main/release_timeline.md) and [Currently Supported Versions](https://docs.risingwave.com/changelog/release-support-policy#support-end-dates-for-recent-releases) to determine which release branches I need to cherry-pick this PR into. 

## Documentation

- [ ]  My PR needs documentation updates. 

<details>
<summary><b>Release note</b></summary>

</details>

PR #22890, opened 2025-08-14
feat(vector): implement dedicated array and serde format for vector data type

wenym1:

As titled.

When we handle the vector value with external usage, such as converting to arrow, or writing to external sinks, we treat it as a ListArray of Float32 with null bitmap as all ones. 

When doing serde inside RisingWave, we treat it as an array of f32, and length is not even included in the encoding format, because we always deserialize the value providing the data type, and length can be obtained from the data type.

## Checklist

## Documentation

<details>

</details>


wenym1:
* **#22900** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22900?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#22890** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22890?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22890?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#22820** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22820?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/common/src/array/vector_array.rs
```diff
@@ -266,16 +359,26 @@ impl TryFrom<f32> for Finite32 {
     }
 }
 
+impl From<Vec<Finite32>> for VectorVal {
+    fn from(value: Vec<Finite32>) -> Self {
+        let (ptr, len, cap) = value.into_raw_parts();
+        // Safety: OrderedFloat is #[repr(transparent)] and has no invalid values.
+        Self {
+            inner: unsafe { Vec::from_raw_parts(ptr as *mut F32, len, cap).into_boxed_slice() },
```
BugenZhao, line 367: Shall we call `std::mem::transmute`?

src/common/src/array/vector_array.rs
BugenZhao, line 1: Basically, this is a specialization of `ListArray` which inner type is `Float32`?

wenym1: It depends. When we handle the vector value with external usage, such as converting to arrow, or writing to external sinks, we treat it as a ListArray of Float32 with null bitmap as all ones. When doing serde inside RisingWave, we treat it as an array of f32, and length is not even included in the encoding format, because we always deserialize the value providing the data type, and length can be obtained from the data type.

Issue #21999, opened 2025-05-27
`arg_min` & `arg_max` aggregate functions


stdrc:

stdrc:
Seems to be something like `first_value(x order by y) filter (where y is not null)`

patelm5:
Hello. I have a request with our legal team out to approve CLA agreement, then I will submit a PR for this if you don't mind.

patelm5:
CLA signed. Happy to address any feedback before merging.

PR #22895, opened 2025-08-14
feat introduce aggregation fns for arg_min / max addressing issue #21999

patelm5:

introduce aggregation fns for arg_min / max addressing issue #21999

includes sqllogictests for the new functions
generally have aimed to add support as a rewrite macro, which I think the issue was suggesting as the implementation approach. 

-->

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/e2e_test/batch/aggregate/arg_min_max.slt.part
+++ b/e2e_test/batch/aggregate/arg_min_max.slt.part
@@ -1,0 +1,21 @@
+statement ok
+SET RW_IMPLICIT_FLUSH TO true;
+
+statement ok
+create table t (v1 int, v2 smallint, v3 varchar);
+
+statement ok
+insert into t values (1, 10, 'a'), (2, 20, 'b'), (3, 10, 'c'), (4, 30, NULL), (5, 1, NULL), (4, NULL, 'd'), (5, NULL, 'e');
+
+query I
+select arg_min(v3, v2) from t;
+----
+a
+
+query I
+select arg_max(v3, v2) from t;
+----
+b
+
+statement ok
+drop table t;
--- a/src/expr/impl/src/aggregate/first_last_value.rs
+++ b/src/expr/impl/src/aggregate/first_last_value.rs
@@ -14,8 +14,8 @@
 
 use risingwave_common::types::{Datum, ScalarRefImpl};
 use risingwave_common_estimate_size::EstimateSize;
-use risingwave_expr::aggregate;
 use risingwave_expr::aggregate::AggStateDyn;
+use risingwave_expr::{ExprError, aggregate};
 
 /// Note that different from `min` and `max`, `first_value` doesn't ignore `NULL` values.
 ///
@@ -82,3 +82,9 @@ fn last_value<T>(_: Option<T>, input: Option<T>) -> Option<T> {
 fn internal_last_seen_value<T>(state: T, input: T, retract: bool) -> T {
     if retract { state } else { input }
 }
+
+#[aggregate("arg_min(any, any) -> any", rewritten)]
+fn _arg_min() {}
+
+#[aggregate("arg_max(any, any) -> any", rewritten)]
+fn _arg_max() {}
--- a/src/frontend/src/optimizer/plan_node/logical_agg.rs
+++ b/src/frontend/src/optimizer/plan_node/logical_agg.rs
@@ -29,7 +29,7 @@ use super::{
 use crate::error::{ErrorCode, Result, RwError};
 use crate::expr::{
     AggCall, Expr, ExprImpl, ExprRewriter, ExprType, ExprVisitor, FunctionCall, InputRef, Literal,
-    OrderBy, WindowFunction,
+    OrderBy, OrderByExpr, WindowFunction,
 };
 use crate::optimizer::plan_node::expr_visitable::ExprVisitable;
 use crate::optimizer::plan_node::generic::GenericPlanNode;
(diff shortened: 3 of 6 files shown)


patelm5:
@lmatz @BugenZhao

src/frontend/src/optimizer/plan_node/logical_agg.rs
```diff
@@ -771,6 +771,37 @@ impl LogicalAggBuilder {
...
+                let not_null_exprs: Vec<ExprImpl> = agg_call
+                    .args
+                    .iter()
+                    .map(|arg| -> Result<ExprImpl> {
+                        Ok(FunctionCall::new(ExprType::IsNotNull, vec![arg.clone()])?.into())
+                    })
+                    .try_collect()?;
+
+                let order_by = OrderBy::new(vec![OrderByExpr {
+                    expr: agg_call.args.pop().unwrap(),
```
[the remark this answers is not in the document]
BugenZhao: Have we validated the number of arguments?

src/frontend/src/optimizer/plan_node/logical_agg.rs
```diff
@@ -771,6 +771,37 @@ impl LogicalAggBuilder {
...
+                let mut agg_call = agg_call;
+                let not_null_exprs: Vec<ExprImpl> = agg_call
+                    .args
+                    .iter()
+                    .map(|arg| -> Result<ExprImpl> {
+                        Ok(FunctionCall::new(ExprType::IsNotNull, vec![arg.clone()])?.into())
+                    })
+                    .try_collect()?;
+
+                let order_by = OrderBy::new(vec![OrderByExpr {
```
BugenZhao: It appears that in DuckDB, it's also allowed to specify `ORDER BY` clause for `arg_min` `arg_max` calls like `select arg_min(a, b order by c) from t`. It could be useful when there's duplicated value on `b`.

Shall we also consider the `order_by` clause specified by user here?

src/frontend/src/optimizer/plan_node/logical_agg.rs
```diff
@@ -771,6 +771,37 @@ impl LogicalAggBuilder {
...
+                    expr: agg_call.args.pop().unwrap(),
+                    order_type: if agg_call.agg_type == AggType::Builtin(PbAggKind::ArgMin) {
+                        OrderType::ascending_nulls_last()
```
BugenZhao: Since we've filtered out NULL values, it doesn't matter what NULL ordering is here. Am I understanding correctly?

e2e_test/batch/aggregate/arg_min_max.slt.part
BugenZhao, line 1: Would you also add more thorough tests ? Such as..
- incorrect arg number, or even `arg_min(*)`
- unsupported input data types
- using it in an `OVER` window function

Can refer to https://github.com/duckdb/duckdb/blob/421152158231ac5fe2d04e9958cedf06cb4aee64/test/sql/aggregate/aggregates/test_arg_min_max.test

src/frontend/src/optimizer/plan_node/logical_agg.rs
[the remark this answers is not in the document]
patelm5: Added some validation for types and args

src/frontend/src/optimizer/plan_node/logical_agg.rs
patelm5: should be working now.

src/frontend/src/optimizer/plan_node/logical_agg.rs
patelm5: doesn't matter, have updated the order by for clarity

e2e_test/batch/aggregate/arg_min_max.slt.part
patelm5: thanks for the link. I've added a few more tests.

patelm5:
@BugenZhao think ready for another look if you have some time.

src/expr/impl/src/aggregate/first_last_value.rs
```diff
@@ -82,3 +82,13 @@ fn last_value<T>(_: Option<T>, input: Option<T>) -> Option<T> {
 fn internal_last_seen_value<T>(state: T, input: T, retract: bool) -> T {
     if retract { state } else { input }
 }
+
+#[aggregate("arg_min(any) -> any", rewritten)]
+#[aggregate("arg_min(any, any) -> any", rewritten)]
+#[aggregate("arg_min(any, any, any) -> any", rewritten)]
```
BugenZhao: In which cases will we have 1 or 3 arguments for these agg calls?

src/frontend/src/optimizer/plan_node/logical_agg.rs
```diff
@@ -771,6 +771,77 @@ impl LogicalAggBuilder {
...
+                            if agg_call.agg_type == AggType::Builtin(PbAggKind::ArgMin) {
+                                "arg_min"
+                            } else {
+                                "arg_max"
+                            },
+                            comparison_arg_type
+                        ));
+                    }
+                    _ => {}
+                }
```
BugenZhao: Could you test the behavior without that validation? I guess it will still be rejected during the function lookup later.

e2e_test/batch/aggregate/arg_min_max.slt.part
```diff
@@ -0,0 +1,140 @@
...
+select arg_min();
+
+statement error
+select arg_max();
+
+statement error
+select arg_min(*);
+
+statement error
+select arg_max(*);
```
BugenZhao: Could you append the expected error message after `statement error`, like:

```
statement error agg_min requires exactly 2 arguments, got 0
```

Regex is supported, but typically it could be simple `string_contains`.

src/expr/impl/src/aggregate/first_last_value.rs
patelm5: I meant to ask a further question about this. Without 86 / 88, arg_min when called with an incorrect number of args will fail with a 'function not found' error message, which is a little confusing. With these statements, it will bind the function and then call the validation provided on number of args. 

Do we have another pattern or example which better handles this?

src/frontend/src/optimizer/plan_node/logical_agg.rs
patelm5: without the validation the command will actually succeed, but it's mentioned in some places we do not intend to support ordering for the types listed in this validation step.

e2e_test/batch/aggregate/arg_min_max.slt.part
patelm5: done.

src/expr/impl/src/aggregate/first_last_value.rs
BugenZhao: > will fail with a 'function not found' error message, which is a little confusing

+1 for this, but it might be expected as Postgres also follows such behavior. 🤣

```
postgres=# select min(2, 3);
ERROR:  function min(integer, integer) does not exist
LINE 1: select min(2, 3);
               ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
```

patelm5: Happy to change it back if you like ?

BugenZhao: Sure. Let's do it

patelm5: updated !

patelm5:
I think may need a hand to rebuild after updating the branch. It looked like a container died in the process.

PR #22900, opened 2025-08-15
refactor(vector): unify storage and streaming vector struct

wenym1:

* Unify two vectors type from `risingwave_storage` and `risingwave_connector` crate.
* Change to use the faiss functions to do vector calculation in vector function

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/planner_test/tests/testdata/output/vector_search.yaml
+++ b/src/frontend/planner_test/tests/testdata/output/vector_search.yaml
@@ -33,7 +33,7 @@
       └─LogicalProject { exprs: [items.id, items.embedding, L2Distance(items.embedding, '[3,1,2]':Vector(3)) as $expr1] }
         └─LogicalScan { table: items, columns: [items.id, items.embedding, items._rw_timestamp] }
   optimized_logical_plan_for_batch: |-
-    LogicalVectorSearch { distance_type: L2, top_n: 5, left: $1, right: '[3,1,2]':Vector(3), output_columns: [items.id:Int32, items.embedding:Vector(3)] }
+    LogicalVectorSearch { distance_type: L2Sqr, top_n: 5, left: $1, right: '[3,1,2]':Vector(3), output_columns: [items.id:Int32, items.embedding:Vector(3)] }
     └─LogicalScan { table: items, columns: [items.id, items.embedding] }
   batch_plan: |-
     BatchProject { exprs: [items.id, items.embedding] }
@@ -52,7 +52,7 @@
       └─LogicalProject { exprs: [items.id, items.embedding, items._rw_timestamp, L2Distance(items.embedding, '[3,1,2]':Vector(3)) as $expr1] }
         └─LogicalScan { table: items, columns: [items.id, items.embedding, items._rw_timestamp] }
   optimized_logical_plan_for_batch: |-
-    LogicalVectorSearch { distance_type: L2, top_n: 5, left: $1, right: '[3,1,2]':Vector(3), output_columns: [items.id:Int32, items.embedding:Vector(3), items._rw_timestamp:Timestamptz] }
+    LogicalVectorSearch { distance_type: L2Sqr, top_n: 5, left: $1, right: '[3,1,2]':Vector(3), output_columns: [items.id:Int32, items.embedding:Vector(3), items._rw_timestamp:Timestamptz] }
     └─LogicalScan { table: items, columns: [items.id, items.embedding, items._rw_timestamp] }
   batch_plan: |-
     BatchTopN { order: [$expr1 ASC], limit: 5, offset: 0 }
@@ -72,7 +72,7 @@
           └─LogicalScan { table: items, columns: [items.id, items.embedding, items._rw_timestamp] }
   optimized_logical_plan_for_batch: |-
     LogicalProject { exprs: [vector_distance, (L2Distance(items.embedding, '[3,1,2]':Vector(3)) + 1:Int32::Float64) as $expr1] }
-    └─LogicalVectorSearch { distance_type: L2, top_n: 5, left: $0, right: '[3,1,2]':Vector(3), output_columns: [items.embedding:Vector(3)] }
+    └─LogicalVectorSearch { distance_type: L2Sqr, top_n: 5, left: $0, right: '[3,1,2]':Vector(3), output_columns: [items.embedding:Vector(3)] }
       └─LogicalScan { table: items, columns: [items.embedding] }
   batch_plan: |-
     BatchProject { exprs: [$expr1, (L2Distance(items.embedding, '[3,1,2]':Vector(3)) + 1:Float64) as $expr2] }


wenym1:
* **#22900** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22900?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22900?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#22890** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22890?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#22820** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22820?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/common/src/vector/mod.rs
```diff
@@ -12,25 +12,120 @@
...
+        Some(self.cmp(other))
+    }
+}
+impl<T: Ord> Ord for VectorInner<T> {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        self.inner.cmp(&other.inner)
+    }
+}
+
+pub trait MeasureDistance {
```
wenym1, line 42: The traits and `l2_norm` related functions are copied from `src/storage/src/vector/mod.rs`.

src/common/src/array/vector_array.rs
```diff
@@ -266,39 +266,19 @@ impl VectorArray {
     }
 }
 
-#[derive(Clone, EstimateSize)]
-pub struct VectorVal {
-    pub(crate) inner: Box<[VectorItemType]>,
-}
+pub type VectorVal = VectorInner<Box<[VectorItemType]>>;
```
BugenZhao, line 269: Shall we consider adopt something like `SmallVec` (without `cap` field then growing functionality) to allow inlining some small vector? It seems that we can always inline 3 `f32`s without any extra overhead on stack size, but we can also choose another capacity.

wenym1: I had a try, and it would expand the size of `ScalarImpl`. Even if I set the size of `SmallVec` to 1, we still cannot pass the `const_assert_eq!(std::mem::size_of::<ScalarImpl>(), 24);`. 

The `VectorVal` is declared as `VectorInner<SmallVec<[VectorItemType; N]>>`. When setting the N from 1 to 5, `size_of::<ScalarImpl>` are all 32, when 6 or 7 it's 40, and when 8 it's 48. 

Vector with small dimension is a rare case. Does't seem like a good idea to increase the size of `ScarlarImpl` for the rare case.

BugenZhao: I guess it's because there's a redundant `cap` field in `SmallVec` that leads to size increasing. But given that...

> Vector with small dimension is a rare case.

We may give up on diving into this. 😄

src/common/src/array/vector_array.rs
```diff
@@ -340,6 +320,10 @@ impl VectorVal {
     pub fn test_type() -> DataType {
         DataType::Vector(Self::TEST_VECTOR_DIMENSION)
     }
+
+    pub fn to_ref(&self) -> VectorRef<'_> {
+        VectorRef { inner: &self.inner }
+    }
```
xiangjinwu, line 326: Looks same as `as_scalar_ref`; what benefit do we get by defining it twice under different names?

src/common/src/array/vector_array.rs
```diff
@@ -436,13 +400,8 @@ impl<'a> ScalarRef<'a> for VectorRef<'a> {
...
 impl<'a> VectorRef<'a> {
-    /// Get the slice of floats in this vector.
-    pub fn into_slice(self) -> &'a [f32] {
-        F32::inner_slice(self.inner)
-    }
-
-    pub fn inner(&self) -> &[VectorItemType] {
-        self.inner
+    pub fn from_slice(inner: &'a [VectorItemType]) -> Self {
+        Self { inner }
```
xiangjinwu: This opens a backdoor to construct invalid vectors containing inf/-inf/nan. A proper implementation shall validate the inputs and return `Result` - or if we decide to trust the caller, name it with `_unchecked`

wenym1:
@BugenZhao Any further comments?

src/expr/impl/src/scalar/vector.rs
```diff
@@ -511,9 +469,9 @@ fn l2_norm_internal(vector: &[f32]) -> f64 {
 /// 1
 ///
 /// query R
-/// SELECT vector_norm('[3e37,4e37]'::vector(2))::real;
+/// SELECT vector_norm('[3e18,4e18]'::vector(2))::real;
 /// ----
-/// 5e+37
+/// 5e+18
```
xiangjinwu, line 474: Due to `faiss::utils::fvec_norm_l2sqr` returning f32 as result?

src/common/src/vector/mod.rs
```diff
@@ -12,25 +12,120 @@
...
+        if l2_norm < f32::MIN_POSITIVE {
+            warn!("normalize 0-norm vector. return original value");
+            return VectorVal {
+                inner: self.inner.as_ref().to_vec().into_boxed_slice(),
+            };
+        }
+        // TODO: vectorize it
+        inner.extend((0..len).map(|i| {
+            // safety: 0 <= i < len
+            unsafe { slice.get_unchecked(i) / l2_norm }
```
xiangjinwu, line 112: Would the division result in `inf` or `-inf`?

src/common/src/array/vector_array.rs
wenym1: `as_scalar_ref` is a method defined in `Scalar`, which means every time we call it, we should treat it as a `Scalar`. However, the `VectorVal` and `VectorRef` is also used in the storage layer, and we don't want to be aware of the concept of `Scalar`, and hope to call its raw method.

Anyway, the two same implementation seems redundant. I will change to call `self.to_ref()` in `as_scalar_ref`.

src/expr/impl/src/scalar/vector.rs
wenym1: Yes. It will incur extra cost and also affect the efficiency of simd operation if we want to support f64 precision.

src/common/src/vector/mod.rs
wenym1: Theorically it's possible. But I would suggest keeping it simple, since if we want to vectorize the calculation in the future, it will incur unnecessary cost to check `inf` on every element one by one, and we have already checked the `l2_norm` is not zero to handle the most common case that can possbly produce `inf`.

wenym1: Wait, it seems impossible, because we can expect that the `l2_norm` should be greater than the absolute value of the element, and therefore the result should be less than 1. I have added a handling logic to handle the `inf` and `nan` l2_norm to always return a vector of 0. With this logic, we may assume that the elements in the output vector are valid.

xiangjinwu (approved):

src/common/src/array/vector_array.rs
```diff
@@ -436,13 +400,8 @@ impl<'a> ScalarRef<'a> for VectorRef<'a> {
...
 
 impl<'a> VectorRef<'a> {
-    /// Get the slice of floats in this vector.
-    pub fn into_slice(self) -> &'a [f32] {
-        F32::inner_slice(self.inner)
-    }
-
-    pub fn inner(&self) -> &[VectorItemType] {
-        self.inner
+    pub fn from_slice_unchecked(inner: &'a [VectorItemType]) -> Self {
```
BugenZhao, line 403: Let's leave some doc on why it's "unchecked"

wenym1: In a hurry to merge the PR for the coming release. Will include the comment in a later PR.

PR #22897, opened 2025-08-14
chore(clippy): add lints to detect potential stack overflows

ChrisWrenDev:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## What's changed and what's your intention?

Added clippy lints to warn for potential stackoverflow.

Resolves: #21635 

## Checklist

- [ ] I have written necessary rustdoc comments.
- [ ]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  I have checked the [Release Timeline](https://github.com/risingwavelabs/rw-commits-history/blob/main/release_timeline.md) and [Currently Supported Versions](https://docs.risingwave.com/changelog/release-support-policy#support-end-dates-for-recent-releases) to determine which release branches I need to cherry-pick this PR into. 

## Documentation

- [ ]  My PR needs documentation updates. 

<details>
<summary><b>Release note</b></summary>

</details>

PR #22906, opened 2025-08-15
refactor: shrink 'write_chunk' future

ChrisWrenDev:

`write_chunk` was holding formatter specific generics and a large child future across `.await`, causing:
- heavy monomorphization + bigger futures
- clippy warnings: `large_futures` and `large_stack_frames`

Change that have been made:
- Moved all formatter/serialization work into a pre-await phase and collected owned `(Option<String>, Option<Vec<u8>>)` pairs.
- In the async phase, boxed each `payload_writer.write_inner(..)` to prevent the large child future from inflating the parent.

Improvements from the change:
- Smaller generated futures and reduced monomorphization.
- Expected compile-time improvement, runtime impact limited to per-record
  boxing overhead on the send path.

Trade-offs:
- Adds one heap allocation per record (boxed child future).
- Consider exposing `BoxFuture` at the trait boundary to avoid per-record
  allocations while keeping futures small.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/src/lib.rs
+++ b/src/connector/src/lib.rs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #![allow(clippy::derive_partial_eq_without_eq)]
+#![warn(clippy::large_futures, clippy::large_stack_frames)]
 #![feature(array_chunks)]
 #![feature(coroutines)]
 #![feature(proc_macro_hygiene)]
--- a/src/connector/src/source/kafka/enumerator.rs
+++ b/src/connector/src/source/kafka/enumerator.rs
@@ -79,14 +79,16 @@ pub struct KafkaSplitEnumerator {
 
 impl KafkaSplitEnumerator {
     async fn drop_consumer_groups(&self, fragment_ids: Vec<u32>) -> ConnectorResult<()> {
-        let admin = SHARED_KAFKA_ADMIN
-            .try_get_with_by_ref(&self.properties.connection, async {
+        let admin = Box::pin(SHARED_KAFKA_ADMIN.try_get_with_by_ref(
+            &self.properties.connection,
+            async {
                 tracing::info!("build new kafka admin for {}", self.broker_address);
                 Ok(Arc::new(
                     build_kafka_admin(&self.config, &self.properties).await?,
                 ))
-            })
-            .await?;
+            },
+        ))
+        .await?;
 
         let group_ids = fragment_ids
             .iter()


ChrisWrenDev:
@wenym1 - Tagging for review (addresses your compile time TODO comment) 
@BugenZhao - Tagging for review (addresses large_futures clippy error)

src/connector/src/sink/pulsar.rs
```diff
@@ -367,23 +369,27 @@ impl AsyncTruncateSinkWriter for PulsarSinkWriter {
...
+                pairs.push((key, value));
+            }
+
+            // async: no formatter/iterator generics across .await
             let mut payload_writer = PulsarPayloadWriter {
                 producer: &mut self.producer,
                 add_future,
                 config: &self.config,
             };
-            // TODO: we can call `payload_writer.write_chunk(chunk, formatter)`,
```
wenym1, line 375: Does the compile time has little difference in the latest rust version if we change the code to the one in the comment? You can see the compile time of the connector crate by adding a `--timings` flag in cargo build or check command.

src/connector/src/sink/pulsar.rs
```diff
@@ -367,23 +369,27 @@ impl AsyncTruncateSinkWriter for PulsarSinkWriter {
...
-                    .write_inner(
-                        key.map(SerTo::ser_to).transpose()?,
-                        value.map(SerTo::ser_to).transpose()?,
-                    )
-                    .await?;
+
+            for (k, v) in pairs {
+                // Box per-record so the parent future stays small
+                Box::pin(payload_writer.write_inner(k, v)).await?;
             }
```
wenym1:
```suggestion
        Box::pin(async move {
            for (k, v) in pairs {
                // Box per-record so the parent future stays small
                payload_writer.write_inner(k, v).await?;
            }
        }).await?
```

to avoid creating a box future for every single row.

src/connector/src/sink/pulsar.rs
```diff
@@ -367,23 +369,27 @@ impl AsyncTruncateSinkWriter for PulsarSinkWriter {
         add_future: DeliveryFutureManagerAddFuture<'a, Self::DeliveryFuture>,
     ) -> Result<()> {
         dispatch_sink_formatter_str_key_impl!(&self.formatter, formatter, {
+            // sync: format & own everything
+            let mut pairs: Vec<(Option<String>, Option<Vec<u8>>)> = Vec::new();
```
wenym1: Shall be initialized with `Vec::with_capacity(chunk.cardinality())`.

Besides, what makes a difference in owning it and the later `no formatter/iterator generics across .await`? The formatter variable itself is a reference and should not occupy too much space if it's captured by the future, and furthermore the `Box::pin` should have already broken down the large future.

ChrisWrenDev: Its about which types live across .await, not about the size of the reference. The `pairs` vec has a small uniform type and lets the iterator drop before the first .await so that no formatter/iterator type appears in the future.

Using `Box::pin` although it reduces the size of the future you put in the box, it doesn't work in this instance because it still has to embed the child future state that it awaits.

src/connector/src/sink/pulsar.rs
ChrisWrenDev: Outer `Box::pin(async { … }).await` doesn’t help as we are still awaiting a large concrete child inside. Using a box for the child per iteration (and drop the outer box) seems to solve this.

Another option is to update the API boundary with BoxFuture to have a single heap allocation per call instead of per record. But this would also involve updating `write_inner` to return a BoxFuture so that the loop awaits a smaller future.

src/connector/src/sink/pulsar.rs
ChrisWrenDev: I’ll benchmark using `--timings` for the connector crate and share the timing diffs. Although the refactor should reduce monomorphization in pulsar.rs the overall crate compile time may be unchanged if other large futures dominate.

src/connector/src/sink/pulsar.rs
ChrisWrenDev: I've just updated the comments to make the reasoning clearer.

src/connector/src/sink/pulsar.rs
ChrisWrenDev: @wenym1 After benchmarks with `--timings`, crate-level compile time didn’t change (cold/warm, dev/release) which suggests other areas dominate compile time. Although no performance increase is seen the change does provide cleaner futures and fewer clippy warnings if the following PR is merged: 
https://github.com/risingwavelabs/risingwave/pull/22897

wenym1:
IIUC, this PR aims to resolves two independent issues. 1. suppressing `clippy::large_futures` and  `clippy::large_stack_frames`. 2. possibly reduce the compile time caused by a combinational number of generated future type.

For 1), I just added the `#![warn(clippy::large_futures, clippy::large_stack_frames)]` included in your previous commit on the main branch and try run clippy locally. It shows two kinds of warning. 

The first warning is `large future with a size of xxx`, which is reported from a series of function call like `obj.func(...).await`.  It means that the future state machine generated by `obj.func(...)` is too large, and calling it directly will allocate the state machine to stack, and can possibly stack-overflow.  To suppress this warning, we can either wrap the `obj.func(...)` immediately with `Box::pin` so as to let the state machine be allocated to heap directly, or we can just decrease the size of state machine of `obj.func(...)`. 

The second warning is `this function may allocate 754102 bytes on the stack`, which is reported from `dispatch_sink_formatter_str_key_impl!()`. Inside the macro, in each arm, we call the `obj.func(...)` above that will allocate large state machine on stack. I tried manually remove from one arm in the macro, and rerun clippy, and see that the number of bytes in the warning message get decreased. This indicates that the the total allocated bytes of the future state machine of a match statement is the sum over the allocated bytes of each arm. I tried two approaches to suppress this warning. First, after warping the `obj.func(...)` with `Box::pin` like what you do this PR, we can avoid capturing the state machine directly on stack in each arm, and then this warning is gone. Second, I tried what I mentioned in https://github.com/risingwavelabs/risingwave/pull/22906#discussion_r2278833857, and it also suppressed this warning, because the await point of each arm is on a boxed future, and therefore the state captured in each arm is also small.

The first kind of warning appears in the call chain from `self.write_inner(k, v)` to `self.send_message(message)` to `self.producer.send_non_blocking`. The `self.producer.send_non_blocking` is a future type defined in the `pulsar` crate, and it's large due to its internal reason. `self.send_message(message)` is large because it captures `self.producer.send_non_blocking` directly, and `self.write_inner(k, v)` captures `self.send_message(message)` to become too large. So the root cause is that the future of `self.producer.send_non_blocking` is too large, and it causes all subsequent warning. What we currently do in this PR is not enough, and the warning of `clippy::large_futures` would still exist if we enable it in the `pulsar.rs` file.

 To summarize, based on the analysis above, on the main branch, I tried simply wrapping `self.producer.send_non_blocking` with `Box::pin`, and then all warnings are gone, and I think this is the only change we need in this PR for this issue.

For 2), as mentioned in https://github.com/risingwavelabs/risingwave/pull/22906#discussion_r2279136642, it makes little difference in the overall compile time, but it incurs extra runtime cost to allocate a vec to pre-store all messages, so in total it's not worth doing this. But I do have another similar idea of avoiding generating a new future type for each format enum variant and meanwhile avoid pre-storing messages in a vec. The idea in this PR is to let the future generation accepts a unified vec type so that only one future type is generated. However, such unified type is not necessarily a vec. It can actually be a unified iterator type. `Box<dyn Iterator>` can work, but it incurs a box allocation. Another way to define unified iterator type is via the `auto_enums::auto_enum(Iterator)` macro, to unify different types of iterator in different match arms into an enum. 

Code will be like the following
```
// change the dispatch_sink_formatter_str_key_impl to allow it to accept attributes over the match statement
macro_rules! dispatch_sink_formatter_str_key_impl {
    ($impl:expr, $name:ident, $body:expr $(,$attr:meta)?) => {
        $(#[$attr])?
        match $impl {
        ...
}

async fn write_chunk<'a>(
    &'a mut self,
    chunk: StreamChunk,
    add_future: DeliveryFutureManagerAddFuture<'a, Self::DeliveryFuture>,
) -> Result<()> {
    let iter = {
        dispatch_sink_formatter_str_key_impl!(
            &self.formatter,
            formatter,
            {
                formatter.format_chunk(&chunk).map(|r| {
                    let (key, value) = r?;
                    let key: Option<String> = key.map(SerTo::ser_to).transpose()?;
                    let value: Option<Vec<u8>> = value.map(SerTo::ser_to).transpose()?;
                    Ok((key, value)) as Result<_>
                })
            },
            auto_enums::auto_enum(Iterator)
        )
    };
    let mut payload_writer = PulsarPayloadWriter {
        producer: &mut self.producer,
        add_future,
        config: &self.config,
    };
    for r in iter {
        let (key, value): (Option<String>, Option<Vec<u8>>) = r?;
        payload_writer.write_inner(key, value).await?;
    }
    Ok(())
}


```

ChrisWrenDev:
@wenym1 Thanks for the detailed review — I should’ve been clearer about what I was trying to achieve with this PR as it would’ve saved you time. I appreciate the careful analysis and suggestions.

Updates applied:
- Added a boxed await at the send_non_blocking call to clear the remaining clippy::large_futures warning in pulsar.rs.
- Added auto_enum on the formatter dispatch to unify the iterator type across arms (no pre-owned Vec, no per-record boxing, less per-variant monomorphization).

Result: Clippy warnings are gone, behavior unchanged, negligible runtime overhead (one small alloc per send attempt), and slightly leaner codegen for write_chunk. 

Thanks again for the guidance!

ChrisWrenDev:
Sounds good! I actually have another PR that adds those Clippy lints to the whole project, but it’s probably better to go crate by crate. I’ve added the `#![warn(clippy::large_futures, clippy::large_stack_frames)]` to risingwave_connector and resolved other warnings with use Box::pin for now.

I’m happy to dig further into the pulsar and kafka crates as you suggested — I’ll raise issues to track the large future problem and send PRs upstream once I confirm the cause.

PR #22907, opened 2025-08-15
feat(streaming): introduce nested loop join executor

yuhao-su:

This Pr is to reopen  #15311

Introduce nested loop join executor.

The LHS is the stream side and RHS is the dimension side.The dimensional table will be broadcasted to all actors while the streaming side will be shuffled.

We do not use operator cache in nlj executor since each coming row will trigger a full table scan and we hope block cache is enough here.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/executor/nested_loop_join.rs
+++ b/src/stream/src/executor/nested_loop_join.rs
@@ -1,0 +1,516 @@
+// Copyright 2024 RisingWave Labs
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::collections::BTreeMap;
+use std::ops::Bound;
+use std::sync::Arc;
+use std::time::{Duration, Instant};
+
+use await_tree::InstrumentAwait;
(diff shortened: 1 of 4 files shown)


src/stream/src/executor/nested_loop_join.rs
```diff
@@ -0,0 +1,516 @@
...
+                    if let Some((_, cache_may_stale)) = side_l_post_commit
+                        .post_yield_barrier(vnode_bitmap.clone())
+                        .await?
+                        && cache_may_stale
+                    {
+                        self.watermark_buffers
+                            .values_mut()
+                            .for_each(|buffers| buffers.clear());
+                    }
+                    // Do not update vnode bitmap for right side because it is a broadcast side.
```
yuhao-su, line 351: Not sure it should work this way. I guess we didn't do this for any other executor before>

kwannoel:
Could you guys share some context? Is it for temporal nested loop join? Please share some SQL examples or RFC documents.

src/stream/src/executor/nested_loop_join.rs
chenzl25: Can we support a singleton distribution first? I think it is identical to the hack of adding dummy columns to the both side of hash join.

hzxa21:
> Could you guys share some context? Is it for temporal nested loop join? Please share some SQL examples or RFC documents.

+1. Is this still a work in progress or is it ready for review? @yuhao-su

yuhao-su:
> > Could you guys share some context? Is it for temporal nested loop join? Please share some SQL examples or RFC documents.
> 
> +1. Is this still a work in progress or is it ready for review? @yuhao-su

There is a blocker for this feature. 

In nlj, one side of the join (assume the dimention side), will be broadcast. That means the state of the dim side (broadcast side) will need to be replicated.

We need further discussion on this matter.

chenzl25:
No updates

PR #22867, opened 2025-08-12
refactor(ddl): Refactor sink into table in dynamic way

Chiro11:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## What's changed and what's your intention?

This PR implements dynamic upstream sink support for tables. Instead of modifying the table's fragment graph at creation/drop time, we now use a dedicated `UpstreamSinkUnion` node that can dynamically add and remove upstream sinks at runtime.

Key changes:
- Added a new `UpstreamSinkUnion` node to the stream plan that can dynamically accept new upstream sinks
- Removed the complex table replacement logic when creating or dropping sinks into tables
- Simplified the protocol by removing the `affected_table_change` field from sink operations
- Added support for dynamically adding and removing upstream sinks through barrier mutations
- Refactored the `DynamicReceivers` to handle dynamic upstream changes properly

This approach is more robust and maintainable as it avoids complex fragment graph modifications when sinks are added or removed, and instead handles these changes through the regular barrier mechanism.

Previous PR:
- #22595
- #22614 

Related issue:
- #22491 

Todo:
- Backward compatible

## Checklist

- [ ] I have written necessary rustdoc comments.
- [ ]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  I have checked the [Release Timeline](https://github.com/risingwavelabs/rw-commits-history/blob/main/release_timeline.md) and [Currently Supported Versions](https://docs.risingwave.com/changelog/release-support-policy#support-end-dates-for-recent-releases) to determine which release branches I need to cherry-pick this PR into. 

## Documentation

- [ ]  My PR needs documentation updates. 

<details>
<summary><b>Release note</b></summary>

</details>

PR #22917, opened 2025-08-18
feat(ddl): Make UpstreamSinkUnion node dynamically handle upstreams

Chiro11:

This is the first part of #22867

Implemented dynamic upstream sink management for stream processing. This PR adds the ability to add and remove upstream sinks at runtime, allowing for more flexible stream topology changes without requiring a full restart.

Key changes:
- Added `UpstreamSinkUnionNode` to the protocol buffer definitions to support dynamic upstream sink management
- Extended `AddMutation` and `StopMutation` to include information about new and dropped upstream sinks
- Implemented handling for dynamic upstream changes in the `UpstreamSinkUnionExecutor`
- Added support for forwarding barriers to upstream sinks and handling barrier alignment
- Refactored `DynamicReceivers` to better handle empty upstream scenarios

This implementation enables more flexible stream processing topologies that can be modified at runtime, improving the system's adaptability to changing requirements.

Previous PR:
- #22595
- #22614

Related issue:
- #22491

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/executor/upstream_sink_union.rs
+++ b/src/stream/src/executor/upstream_sink_union.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::collections::{HashMap, VecDeque};
+use std::collections::HashMap;
 use std::iter;
 use std::pin::Pin;
 use std::task::{Context, Poll};
@@ -396,7 +396,7 @@ impl UpstreamSinkUnionExecutor {
             // Remove the upstream sinks that are no longer needed.
             upstreams.remove_upstreams(dropped_upstream_sinks);
             for upstream_fragment_id in dropped_upstream_sinks {
-                self.barrier_tx_map.remove(upstream_fragment_id).unwrap();
+                self.barrier_tx_map.remove(upstream_fragment_id);
             }
         }
 
@@ -421,7 +421,7 @@ impl UpstreamSinkUnionExecutor {
         let upstreams = DynamicReceivers::new(inputs, Some(barrier_align.clone()), None);
         pin_mut!(upstreams);
 
-        let mut buffered_barriers = VecDeque::new();
+        let mut current_barrier = None;
 
         // Here, `tokio::select` cannot be used directly in `try_stream` function.
         // Err for breaking the loop. Ok(None) for continuing the loop.
@@ -435,14 +435,14 @@ impl UpstreamSinkUnionExecutor {
                     msg = pending_on_none(upstreams.next()) => {
                         let msg = msg.context("UpstreamSinkUnionExecutor pull upstream failed")?;
                         if let Message::Barrier(barrier) = &msg {
-                            let current_barrier = buffered_barriers.pop_front().unwrap();
+                            let current_barrier = current_barrier.take().unwrap();
                             assert_equal_dispatcher_barrier(&current_barrier, barrier);
                             self.handle_update(&mut upstreams, barrier).await?;
                         }
                         return Ok(msg);
                     }
 
-                    barrier = self.barrier_rx.recv() => {
+                    barrier = self.barrier_rx.recv(), if current_barrier.is_none() => {
                         let barrier = barrier.context("Failed to receive barrier from barrier_rx")?;
                         // Here, if there's no upstream, we should process the barrier directly and send it out.
                         // Otherwise, we need to forward the barrier to the upstream and then wait in the first branch
@@ -455,7 +455,7 @@ impl UpstreamSinkUnionExecutor {
                                 tx.send(barrier.clone())
                                     .map_err(|e| StreamExecutorError::from(anyhow::anyhow!(e)))?;
                             }
-                            buffered_barriers.push_back(barrier);
+                            current_barrier = Some(barrier);
                             continue;
                         }
                     }


Chiro11:
* **#23017** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23017?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#22920** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22920?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#22917** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22917?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22917?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

proto/stream_plan.proto
```diff
@@ -942,6 +956,11 @@ message VectorIndexWriteNode {
   catalog.Table table = 1;
 }
 
+// This is just for creating. Upstreams may be added and removed dynamically at runtime.
```
wenym1: May move the comment to the field rather than the struct, and mention that this is always empty in the persisted metadata, and get filled later before we spawn the actors.

src/compute/tests/cdc_tests.rs
```diff
@@ -315,6 +315,7 @@ async fn test_cdc_backfill() -> StreamResult<()> {
             subscriptions_to_add: vec![],
             backfill_nodes_to_pause: Default::default(),
             actor_cdc_table_snapshot_splits: Default::default(),
+            new_upstream_sinks: Default::default(),
```
wenym1: Can you help to change similar test code in this PR to the following by the way, so that in the future we don't have to modify the tests when we add more fields.
```
AddMutation {
    // non-default fields.
    ..Default::default()
}
```

src/stream/src/executor/mod.rs
```diff
@@ -324,12 +326,19 @@ pub struct AddMutation {
     /// nodes which should start backfill
     pub backfill_nodes_to_pause: HashSet<FragmentId>,
     pub actor_cdc_table_snapshot_splits: CdcTableSnapshotSplitAssignment,
+    pub new_upstream_sinks: HashMap<ActorId, PbNewUpstreamSink>,
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct StopMutation {
+    pub dropped_actors: HashSet<ActorId>,
+    pub dropped_upstream_sinks: HashMap<ActorId, Vec<FragmentId>>,
```
wenym1: Can we just store a vec of sink_id of dropped sinks? Each upstream sink union executor can identify its inputs by sink_id, and then it checks whether any of its input sink_id is in the list of dropped sinks and then remove the input accordingly. In this way, the required information to construct this mutation will be much simpler.

wenym1: Or if we identify the upstream inputs by fragment_id in the upstream sink union executor, we can store a flattened set of fragment_id in the mutation, and drop any input when seeing the fragment_id of the input is in the list.

This idea is similar to how to use the `dropped_actors`. We only include all the actors to drop in the mutation, rather than the upstream actors to drop for each individual actor.

src/stream/src/executor/mod.rs
```diff
@@ -1403,7 +1468,6 @@ impl<InputId: Clone + Ord + Hash + std::fmt::Debug + Unpin, M: Clone + Unpin> St
             return Poll::Ready(None);
         }
 
-        let mut start = None;
```
wenym1, line 1406: Does it mean that our previous measurement on the time of barrier alignment is problematic?

src/stream/src/executor/mod.rs
```diff
@@ -1480,6 +1545,12 @@ impl<InputId: Clone + Ord + Hash + std::fmt::Debug + Unpin, M: Clone + Unpin> St
         }
 
         assert!(self.active.is_terminated());
+
+        if self.blocked.is_empty() {
```
wenym1: From the code above at the beginning of this method
```
fn poll_next(
        mut self: Pin<&mut Self>,
        cx: &mut std::task::Context<'_>,
    ) -> Poll<Option<Self::Item>> {
        if self.active.is_terminated() {
            // This only happens if we've been asked to stop.
            assert!(self.blocked.is_empty());
            return Poll::Ready(None);
        }
....
```

returning None has the semantic that we are reaching the end of this stream, and should no longer be polled, and the logic here violate the previous semantic. 

We can instead check `self.is_empty()` at the very beginning, and return `Poll::Pending` when empty to indicate that the stream has nothing to yield in current setting, but has not reached the end yet. And then the code later can be untouched except the change on `start_ts`.

src/stream/src/executor/actor.rs
```diff
@@ -45,6 +45,7 @@ use crate::error::StreamResult;
 use crate::task::{ActorId, FragmentId, LocalBarrierManager};
 
 /// Shared by all operators of an actor.
+#[derive(Clone)]
```
wenym1: Seems no need to add the changes here, because we mostly pass ActorContextRef.

src/stream/src/executor/upstream_sink_union.rs
```diff
@@ -100,13 +111,45 @@ impl Stream for SinkHandlerInput {
 }
 
 /// Information about an upstream fragment including its schema and projection expressions.
-#[derive(Debug)]
-pub struct UpstreamInfo {
+#[derive(Debug, Default)]
```
wenym1: Better not to derive `Default` on the struct, to avoid implicitly creating a struct with meaningless default `upstream_fragment_id` and `merge_schema`.

src/stream/src/executor/upstream_sink_union.rs
```diff
@@ -100,13 +111,45 @@ impl Stream for SinkHandlerInput {
...
+                anyhow::anyhow!(
+                    "upstream fragment {} not found in initial upstream actors",
+                    upstream_fragment_id
+                )
+            })?;
+        let merge_schema = sink_output_schema.iter().map(Field::from).collect();
+        let project_exprs = project_exprs
+            .iter()
+            .map(|e| build_non_strict_from_prost(e, error_report.clone()))
+            .collect::<Result<Vec<_>, _>>()
```
wenym1: Can use `try_collect()` and all type can be automatically derived.

src/stream/src/executor/upstream_sink_union.rs
```diff
@@ -134,7 +177,16 @@ pub struct UpstreamSinkUnionExecutor {
     chunk_size: usize,
 
     /// The initial inputs to the executor.
-    upstream_infos: Vec<UpstreamInfo>,
+    upstream_infos: Vec<UpstreamFragmentInfo>,
```
wenym1: Better to rename to `initial_upstream_infos` to indicate that it should only be used in initialization.

src/stream/src/executor/upstream_sink_union.rs
```diff
@@ -185,28 +233,39 @@ impl UpstreamSinkUnionExecutor {
...
+            },
+            barrier_rx,
+            barrier_tx_map: Default::default(),
         }
     }
 
-    #[allow(dead_code)]
+    fn subscribe_local_barrier(&mut self, fragment_id: FragmentId) -> UnboundedReceiver<Barrier> {
+        let (tx, rx) = unbounded_channel();
+        self.barrier_tx_map.insert(fragment_id, tx);
```
wenym1: use `self.barrier_tx_map.try_insert().expect("non-duplicate")` to check duplication.

src/stream/src/executor/upstream_sink_union.rs
```diff
@@ -218,33 +277,27 @@ impl UpstreamSinkUnionExecutor {
     }
 
     async fn new_merge_executor(
-        &self,
+        &mut self,
```
wenym1, line 283: Do we use the mutable reference of self anywhere? If no, better to just keep it the immutable one.

src/stream/src/executor/upstream_sink_union.rs
```diff
@@ -311,25 +422,79 @@ impl UpstreamSinkUnionExecutor {
...
+                        for tx in self.barrier_tx_map.values() {
+                            tx.send(barrier.clone())
+                                .map_err(|e| StreamExecutorError::from(anyhow::anyhow!(e)))?;
+                        }
+                        current_barrier = Some(barrier.clone());
+                        Ok(None)
+                    }
+                }
+
+                else => {
```
wenym1: In current implementation, will `upstreams.next()` return `None` when there is no upstream? If so, will it be incorrectly treated as error in this case?

src/stream/src/executor/upstream_sink_union.rs
```diff
@@ -311,25 +422,79 @@ impl UpstreamSinkUnionExecutor {
                 "UpstreamSinkUnion",
             ]);
 
-        let upstreams = DynamicReceivers::new(inputs, Some(barrier_align), None);
+        let upstreams = DynamicReceivers::new(inputs, Some(barrier_align.clone()), None);
         pin_mut!(upstreams);
 
-        while let Some(msg) = upstreams.next().await {
-            yield msg?;
+        let mut current_barrier = None;
```
wenym1, line 424: `Option<Barrier>` is something like a queue with max capacity as 1. Can we instead use a unlimited capacity `VecDeque` to store all pending barriers so that we don't have to do conditional select on `self.barrier_rx.recv(), if current_barrier.is_none()`?

src/stream/src/executor/upstream_sink_union.rs
```diff
@@ -311,25 +422,79 @@ impl UpstreamSinkUnionExecutor {
...
+
+        // Here, `tokio::select` cannot be used directly in `try_stream` function.
+        // Err for breaking the loop. Ok(None) for continuing the loop.
```
wenym1, line 427: Can we have a loop inside the async closure so that it can return a cleaner `Result<Message>` instead of `Ok(None)`?

src/stream/src/executor/upstream_sink_union.rs
```diff
@@ -311,25 +422,79 @@ impl UpstreamSinkUnionExecutor {
...
+                    Ok(Some(msg))
+                }
+
+                Some(barrier) = self.barrier_rx.recv(), if current_barrier.is_none() => {
+                    // Here, if there's no upstream, we should process the barrier directly and send it out.
+                    // Otherwise, we need to forward the barrier to the upstream and then wait in the first branch
+                    // until the upstreams have processed the barrier.
+                    if upstreams.is_empty() {
+                        self.handle_update(&mut upstreams, actor_id, &barrier_align, &barrier).await?;
+                        Ok(Some(Message::Barrier(barrier.clone())))
```
wenym1: The `.clone()` is unnecessary.

src/stream/src/executor/upstream_sink_union.rs
```diff
@@ -296,8 +350,65 @@ impl UpstreamSinkUnionExecutor {
         }
     }
 
+    async fn handle_update(
+        &mut self,
+        upstreams: &mut DynamicReceivers<FragmentId, BarrierMutationType>,
+        actor_id: ActorId,
```
wenym1: `actor_id` can be obtained from `self.actor_ctx`.

wenym1 (commented):
Rest LGTM.

src/stream/src/executor/mod.rs
Chiro11: Yes, each `poll_next` resets start to None, but it may return in the middle due to receiving a chunk.

src/stream/src/executor/mod.rs
Chiro11: This references the `FutureUnordered` container, which, as a Stream, returns None when empty. Considering DynamicReceivers is also a dynamic container, I think returning None makes sense here, rather than Pending. Besides, I suspect Pending implies some underlying wakeup mechanism, but here no wakeup occurs.

wenym1: > This references the FutureUnordered container, which, as a Stream, returns None when empty.

I don't think the design of `FutureUnordered` is a good one. In the trait definition of [Stream](https://docs.rs/futures/latest/futures/prelude/trait.Stream.html#tymethod.poll_next), it says 
```
Poll::Ready(None) means that the stream has terminated, and poll_next should not be invoked again.
```

Besides, in current implementation, when returning `Poll::Ready(None)`, we can't distinguish whether it's reaching the end of stream, or we are just having empty inputs.

> I suspect Pending implies some underlying wakeup mechanism, but here no wakeup occurs.

This is not true. Returning pending only expects a waker when there is really some to wait for in the future. The `Pending` future created by [pending](https://doc.rust-lang.org/std/future/fn.pending.html) always returns Pending, but does not involve any wakeup, because it never returns something. In our use case, when the inputs are empty, we can never get anything from it, and it's fine to return Pending. Furthermore, returning Pending here can make the handling in the external `select!` much easier.

src/stream/src/executor/upstream_sink_union.rs
Chiro11: No, if the first branch returns None, it will wait on the second branch. And if upstreams is empty, the second branch will never cache `current_barrier`.

src/stream/src/executor/upstream_sink_union.rs
Chiro11: Creating a new merge requires calling `subscribe_local_barrier`, which modifies the local `barrier_tx_map`.

src/stream/src/executor/mod.rs
Chiro11: That makes some sense. However, the [StreamMap](https://docs.rs/tokio-stream/latest/tokio_stream/struct.StreamMap.html#method.poll_next_many) I used before also had the same behavior (that is, it would return None when the container was empty). Otherwise, it means that we need to be very careful about calling `await` on this container. I think this also makes some sense.

wenym1: > Besides, in current implementation, when returning Poll::Ready(None), we can't distinguish whether it's reaching the end of stream, or we are just having empty inputs.

This is another thing we should consider. When receiving None, if it's the end of stream, we should turn it into an error, and if it's caused by empty inputs, we should hang the stream. These are two different behaviors on a same return value.

Returning Pending is a good way to help distinguish the two cases, and it makes the handling in `select!` easier. Another solution is to forbidden calling `poll_next` when inputs is empty. This can be achieved by asserting `!self.is_empty()` in `poll_next`. In the outer `select!` we conditionally poll it with `if !stream.is_empty()`. Assert inputs is not empty in `poll_next` can also serve as the previous sanity check on `!stream.is_empty()` for `MergeExecutor` and `UnionExecutor`, and might be a better solution.

src/stream/src/executor/upstream_sink_union.rs
wenym1: Checking the doc of `select!`. I think the `else` should not be used, or should be used carefully when we have no other choice. When we use `select!`, we usually expect when any arm receives the output, we can take action immediately. However, when using `else`, when a arm receives something but not matched with the pattern, it will store the pending output and should wait for the output  from other arm. This can easily cause dead-lock.

In current code, there is already a deadlock. In `Some(barrier) = self.barrier_rx.recv()`, if the `barrier_tx` is dropped and `self.barrier_rx.recv()` returns None, originally we expect to return error immediately. But now we will temporarily ignore it, and continue waiting on `upstreams.next()`. However, in any `MergeExecutor` of `upstreams`, when it receives a dispatcher barrier from upstream, it will wait for the outer `UpstreamSinkUnion` to send a barrier with mutation to it. However, the outer `UpstreamSinkUnion` will never send the barrier, because the related logic is hanged already, and then `upstreams.next()` will never return, and the stream is deadlocked.

src/stream/src/executor/mod.rs
```diff
@@ -1397,13 +1447,10 @@ impl<InputId: Clone + Ord + Hash + std::fmt::Debug + Unpin, M: Clone + Unpin> St
         mut self: Pin<&mut Self>,
         cx: &mut std::task::Context<'_>,
     ) -> Poll<Option<Self::Item>> {
-        if self.active.is_terminated() {
-            // This only happens if we've been asked to stop.
-            assert!(self.blocked.is_empty());
-            return Poll::Ready(None);
+        if self.is_empty() {
+            return Poll::Pending;
         }
```
wenym1:
```suggestion
        if self.is_empty() {
            return Poll::Pending;
        }
        if self.active.is_terminated() {
            // This only happens if we've been asked to stop.
            assert!(self.blocked.is_empty());
            return Poll::Ready(None);
        }
```

Should preserve the original check on end of stream.

wenym1: Discussed offline. It seems that in original code the branch `if self.active.is_terminated() {}` will never be reached. If so, there is no need to distinguish the two cases mentioned in https://github.com/risingwavelabs/risingwave/pull/22917#discussion_r2284729715, and we can remove this branch.

src/stream/src/executor/mod.rs
```diff
@@ -324,12 +325,19 @@ pub struct AddMutation {
     /// nodes which should start backfill
     pub backfill_nodes_to_pause: HashSet<FragmentId>,
     pub actor_cdc_table_snapshot_splits: CdcTableSnapshotSplitAssignment,
+    pub new_upstream_sinks: HashMap<ActorId, PbNewUpstreamSink>,
```
wenym1: It seems that the DispatchType between the upstream sink and table is always Hash. If so, the `upstream_actors` in `PbNewUpstreamSink` will be the same for each actor of a same table. In this case, we may only need to store a single copy of `PbNewUpstreamSink` for all actors of a table. 

We can change the key of the map to either `fragment_id` or `table_id`, depending on which one can be implemented easier. `table_id` is not carried in `ActorContext`, but we can include it as an immutable field in `UpstreamSinkUnionNode`.

src/stream/src/executor/mod.rs
```diff
@@ -301,7 +302,7 @@ pub const INVALID_EPOCH: u64 = 0;
 type UpstreamFragmentId = FragmentId;
 type SplitAssignments = HashMap<ActorId, Vec<SplitImpl>>;
 
-#[derive(Debug, Clone, PartialEq)]
+#[derive(Debug, Clone, PartialEq, Default)]
```
wenym1: Seems that `Default::default()` is only used in test. May use `#[cfg_attr(test, derive(Default))]` to force developers to fill in each field one by one in usual code other than test.

src/stream/src/executor/mod.rs
```diff
@@ -1462,16 +1509,15 @@ impl<InputId: Clone + Ord + Hash + std::fmt::Debug + Unpin, M: Clone + Unpin> St
...
-                        }
-                        if let Some(merge_barrier_align_duration) =
-                            &self.merge_barrier_align_duration
-                        {
-                            // Observe did a few atomic operation inside, we want to avoid the overhead.
-                            merge_barrier_align_duration.observe(start.elapsed().as_secs_f64())
-                        }
+                    assert!(!self.blocked.is_empty());
+
+                    let start_ts = self.start_ts.take().unwrap();
```
wenym1: May add comment or `.expect("reason")` to explain why it must be `Some(_)` when reaching here.

src/stream/src/executor/upstream_sink_union.rs
```diff
@@ -218,33 +280,27 @@ impl UpstreamSinkUnionExecutor {
...
-                        self.actor_context.fragment_id,
-                        upstream_actor,
-                        upstream_fragment_id,
-                    )
-                }),
-        )
-        .await?;
+        let barrier_rx = self.subscribe_local_barrier(upstream_fragment_id);
+
+        let mut inputs = Vec::with_capacity(upstream_actors.len());
```
wenym1: We used to initialize the inputs via a `try_join_all` so as to create the inputs concurrently. Any reason to change to for loop?

src/stream/src/executor/upstream_sink_union.rs
```diff
@@ -296,8 +353,65 @@ impl UpstreamSinkUnionExecutor {
...
+            assert_equal_dispatcher_barrier(barrier, &new_barrier);
+
+            upstreams.add_upstreams_from(new_receivers);
+        }
+
+        if let Some(dropped_upstream_sinks) = barrier.as_dropped_upstream_sinks()
+            && !dropped_upstream_sinks.is_empty()
+        {
+            // Remove the upstream sinks that are no longer needed.
+            let dropped_upstream_sinks = dropped_upstream_sinks.iter().cloned().collect();
```
wenym1: Can we store the `dropped_upstream_sinks` as `HashSet` in the mutation directly?

src/stream/src/executor/upstream_sink_union.rs
```diff
@@ -296,8 +353,65 @@ impl UpstreamSinkUnionExecutor {
...
+                .unwrap()
+                .send(barrier.clone())
+                .map_err(|e| StreamExecutorError::from(anyhow::anyhow!(e)))?;
+
+            let mut new_receivers =
+                DynamicReceivers::new(vec![new_input], Some(barrier_align.clone()), None);
+            let new_barrier = expect_first_barrier(&mut new_receivers).await?;
+            assert_equal_dispatcher_barrier(barrier, &new_barrier);
+
+            upstreams.add_upstreams_from(new_receivers);
```
wenym1: It seems wired to create a single input `DynamicReceivers` just for specially calling `add_upstreams_from`. Shall we just call `extend_active` instead?

Looking at the implementation, the difference between `extend_active` and `add_upstreams_from` is that `add_upstreams_from` has extra logic to handle watermark.  `add_upstream_from` has only one usage before this PR
```
let mut select_new = SelectReceivers::new(
    new_upstreams,
    None,
    select_all.merge_barrier_align_duration(),
);
let new_barrier = expect_first_barrier(&mut select_new).await?;
assert_equal_dispatcher_barrier(barrier, &new_barrier);
// Add the new upstreams to select.
select_all.add_upstreams_from(select_new);
```

The usage is similar to the current code. The `select_new` receives only a single barrier before being called `add_upstreams_from`, and no watermark is in `select_new`, and therefore `add_upstream_from` is the same as `extend_active`. So I think we can remove `add_upstreams_from` and unify all similar logic to calling `extend_active`.

wenym1 (commented):
Rest LGTM. Thanks for the PR.

src/stream/src/executor/mod.rs
Chiro11: This is used by cdc_tests.rs, but this test does not seem to meet the conditional of `#cfg(test)`. Any suggestions?

BugenZhao: I think it's okay to get an empty (no-op) mutation with `Default`.

src/stream/src/executor/mod.rs
```diff
@@ -420,7 +428,10 @@ impl Barrier {
 
     #[must_use]
     pub fn with_stop(self) -> Self {
-        self.with_mutation(Mutation::Stop(HashSet::default()))
+        self.with_mutation(Mutation::Stop(StopMutation {
+            dropped_actors: Default::default(),
+            dropped_sink_fragments: Default::default(),
+        }))
```
BugenZhao, line 437:
```suggestion
        self.with_mutation(Mutation::Stop(StopMutation::default()))
```

src/stream/src/executor/mod.rs
```diff
@@ -617,7 +628,29 @@ impl Barrier {
                 | Mutation::AddAndUpdate(_, UpdateMutation { merges, .. }) => {
                     merges.get(&(actor_id, upstream_fragment_id))
                 }
+                _ => None,
+            })
+    }
 
+    pub fn as_new_upstream_sink(&self, actor_id: ActorId) -> Option<&PbNewUpstreamSink> {
```
BugenZhao: Shall we add some documentation like other similar methods?

src/stream/src/executor/upstream_sink_union.rs
BugenZhao: I think the logic here might be copied from how we handle upstream scaling-out in `Merge` executor, where multiple upstreams could be added at the same time, so the best way to align their first barriers is to reuse the `SelectReceivers` utility.

Given that there'll always be 1 upstream to add here, I think suggestions from @wenym1 is also reasonable. Let's see which approach can be the simplest.

src/stream/src/executor/upstream_sink_union.rs
```diff
@@ -311,25 +425,76 @@ impl UpstreamSinkUnionExecutor {
...
+                        }
+                    }
+                }
+            }
+        };
+
+        loop {
+            match select_once().await {
+                Ok(msg) => yield msg,
+                Err(e) => return Err(e), // Propagate the error.
```
BugenZhao:
```suggestion
            yield select_once().await?;
```

src/stream/src/executor/mod.rs
Chiro11: Using `fragment_id` is ok. Thanks.

src/stream/src/executor/mod.rs
wenym1: Basically we want to export some test_utils defined in one crate to be used in the tests in other crates, but meanwhile we want to disable the usage in usual code in other crates.

We can follow what we do in `risingwave_hummock_sdk`. In the `Cargo.toml` of the `risingwave_stream` crate, we can add a new feature
```
[features]
dhat-heap = ["dhat"]
test = [] // the added one
```

And then we use `#[cfg_attr(any(test, feature = "test"), derive(Default))]` to conditonally enable the derived Default on either tests in current crate, or in the tests of external crates. In external crates (e.g. the crate of `cdc_tests.rs`), we can enable the `test` feature on `risingwave_stream` only in its `dev-dependencies`.
```
[dev-dependencies]
...
risingwave_stream = { workspace = true, features = ["test"] }
...
```

src/stream/src/executor/upstream_sink_union.rs
```diff
@@ -311,25 +425,76 @@ impl UpstreamSinkUnionExecutor {
...
+                        let msg = msg.context("UpstreamSinkUnionExecutor pull upstream failed")?;
+                        if let Message::Barrier(barrier) = &msg {
+                            let current_barrier = buffered_barriers.pop_front().unwrap();
+                            assert_equal_dispatcher_barrier(&current_barrier, barrier);
+                            self.handle_update(&mut upstreams, &barrier_align, barrier).await?;
+                        }
+                        return Ok(msg);
+                    }
+
+                    barrier = self.barrier_rx.recv() => {
```
BugenZhao: If this would further simplify the implementation: Is it possible to treat the barrier receiver as a special input of `DynamicReceivers`, so that it will never be empty and we don't need extra logic for `select`ing here.

BugenZhao (commented):
Rest LGTM. Great work!

src/stream/src/executor/upstream_sink_union.rs
wenym1: `upstreams` has logic that depends on `self.barrier_rx`.  When we receive a barrier from `self.barrier_rx.recv()`, we will forward such barrier with mutation to the `upstreams`, so that the internal `MergeExecutor` in `upstreams` can handle upstream scale. It might be complicated to put two non-independent streams in `DynamicReceivers`.

src/stream/src/executor/upstream_sink_union.rs
Chiro11: After offline discussion, `add_upstreams_from` was slightly refactored to avoid creating temporary objects.

src/stream/src/executor/mod.rs
Chiro11: This is not in the test. Better to specify each field.

Chiro11:
Fix all comments now. Thank you both for the detailed feedback!

wenym1 (approved):
LGTM!

PR #22920, opened 2025-08-18
refactor(sink): Simplify sink-into-table with new dynamic operator

Chiro11:

This is the second part of #22867

This PR refactors the sink-into-table implementation to use a new `StreamUpstreamSinkUnion` operator instead of manipulating the fragment graph directly. The new approach:

1. Adds a new `StreamUpstreamSinkUnion` operator that handles the merging of data from upstream sinks
2. Removes the complex graph manipulation logic that was previously used to inject merge nodes
3. Simplifies the creation and dropping of sinks that target tables
4. Removes the deprecated `affected_table_change` field from the `CreateSinkRequest` and `DropSinkRequest` protos
5. Updates the catalog controller to properly handle sink-into-table relationships

This change makes the sink-into-table implementation more maintainable and less error-prone by using a dedicated operator instead of manipulating the fragment graph directly.

Note: The backward compatibility logic will be placed in the next PR.

Previous PR:
- #22917

Related issue:
- #22491

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/e2e_test/sink/sink_into_table/alter_column.slt
+++ b/e2e_test/sink/sink_into_table/alter_column.slt
@@ -178,6 +178,9 @@ select * from t_append_only;
 statement ok
 create sink s21 into t_append_only as select v1+1000 as v1, v1+2000 as v2, v1+3000 as v3 from t_s2 with (type = 'append-only', force_append_only = 'true');
 
+statement ok
+recover;
+
 query III rowsort
 select * from t_append_only;
 ----
@@ -210,9 +213,12 @@ statement ok
 create sink s_alter_1 into m_alter_1 (v2, v1) as select v2 as w2, v1 + 10 as w1 from t_alter_1;
 
 statement ok
+recover;
+
+statement ok
 insert into t_alter_1 values (1, 'a'), (2, 'b'), (3, 'c');
 
-query IT rowsort
+query TI rowsort
 select * from m_alter_1;
 ----
 a 11
@@ -222,7 +228,7 @@ c 13
 statement ok
 alter table m_alter_1 add column v3 int default 1000;
 
-query ITI rowsort
+query TII rowsort
 select * from m_alter_1;
 ----
 a 11 1000
@@ -232,7 +238,7 @@ c 13 1000
 statement ok
 insert into t_alter_1 values (4, 'd'), (5, 'e');
 
-query ITI rowsort
+query TII rowsort
 select * from m_alter_1;
 ----
 a 11 1000
@@ -281,3 +287,71 @@ drop table t2;
 
 statement ok
 drop table t1;
+
+statement ok
+create table t_drop_1 (v1 int, v2 varchar);
+
+statement ok
+create table m_drop_1 (v2 varchar, v1 int primary key);
+
+statement ok
+create sink s_drop_1 into m_drop_1 (v2, v1) as select v2, v1 from t_drop_1;
+
+statement ok
+recover;
+
+statement ok
+insert into t_drop_1 values (1, 'a'), (2, 'b'), (3, 'c');
+
+statement error dropping columns in target table of sinks is not supported
+alter table m_drop_1 drop column v2;
+
+statement ok
+drop sink s_drop_1;
+
+statement ok
+alter table m_drop_1 drop column v2;
+
+query I rowsort
+select * from m_drop_1;
+----
+1
+2
+3
+
+statement ok
+alter table m_drop_1 add column v3 int default 1000;
+
+query II rowsort
+select * from m_drop_1;
+----
+1 1000
+2 1000
+3 1000
+
+statement ok
+create sink s_drop_1 into m_drop_1 (v1, v3) as select v1, v1 + 10 from t_drop_1;
+
+statement ok
+insert into t_drop_1 values (4, 'd'), (5, 'e');
+
+query II rowsort
+select * from m_drop_1;
+----
+1 11
+2 12
+3 13
+4 14
+5 15
+
+statement error dropping columns in target table of sinks is not supported
+alter table m_drop_1 drop column v2;
+
+statement ok
+drop sink s_drop_1;
+
+statement ok
+drop table m_drop_1;
+
+statement ok
+drop table t_drop_1;
--- a/proto/ddl_service.proto
+++ b/proto/ddl_service.proto
@@ -88,8 +88,8 @@ message AlterSourceResponse {
 message CreateSinkRequest {
   catalog.Sink sink = 1;
   stream_plan.StreamFragmentGraph fragment_graph = 2;
-  // It is used to provide a replace plan for the downstream table in `create sink into table` requests.
-  optional ReplaceJobPlan affected_table_change = 3;
+  reserved 3;
+  reserved "affected_table_change";
   // The list of object IDs that this sink depends on.
   repeated uint32 dependencies = 4;
   bool if_not_exists = 5;
@@ -103,7 +103,8 @@ message CreateSinkResponse {
 message DropSinkRequest {
   uint32 sink_id = 1;
   bool cascade = 2;
-  optional ReplaceJobPlan affected_table_change = 3;
+  reserved 3;
+  reserved "affected_table_change";
 }
 
 message DropSinkResponse {
--- a/src/frontend/src/handler/drop_sink.rs
+++ b/src/frontend/src/handler/drop_sink.rs
@@ -13,9 +13,6 @@
 // limitations under the License.
 
 use pgwire::pg_response::{PgResponse, StatementType};
-use risingwave_common::catalog::StreamJobStatus;
-use risingwave_pb::ddl_service::{ReplaceJobPlan, replace_job_plan};
-use risingwave_pb::meta::cancel_creating_jobs_request::{CreatingJobIds, PbJobs};
 use risingwave_sqlparser::ast::ObjectName;
 
 use super::RwPgResponse;
@@ -23,8 +20,6 @@ use crate::binder::Binder;
 use crate::catalog::root_catalog::SchemaPath;
 use crate::error::Result;
 use crate::handler::HandlerArgs;
-use crate::handler::alter_table_column::hijack_merger_for_target_table;
-use crate::handler::create_sink::{fetch_incoming_sinks, reparse_table_for_sink};
 
 pub async fn handle_drop_sink(
     handler_args: HandlerArgs,
@@ -62,60 +57,8 @@ pub async fn handle_drop_sink(
 
     let sink_id = sink.id;
 
-    if sink.stream_job_status == StreamJobStatus::Creating && sink.target_table.is_none() {
-        let canceled_jobs = session
-            .env()
-            .meta_client()
-            .cancel_creating_jobs(PbJobs::Ids(CreatingJobIds {
-                job_ids: vec![sink_id.sink_id],
-            }))
-            .await?;
-        tracing::info!(?canceled_jobs, "cancelled creating jobs");
-    } else {
-        // No matter whether the sink is being created or not, we need to replace the target table graph.
-        // TODO(august): using cancel interface to cancel the creating job after refactoring sink into table.
-        let mut affected_table_change = None;
-        if let Some(target_table_id) = &sink.target_table {
-            let table_catalog = {
-                let reader = session.env().catalog_reader().read_guard();
-                let table = reader.get_any_table_by_id(target_table_id)?;
-                table.clone()
-            };
-
-            let (mut graph, table, source, target_job_type) =
-                reparse_table_for_sink(&session, &table_catalog).await?;
-
-            let columns_without_rw_timestamp = table_catalog.columns_without_rw_timestamp();
-            for sink in fetch_incoming_sinks(&session, &table_catalog)? {
-                if sink.id == sink_id {
-                    // Skip the sink being dropped.
-                    continue;
-                }
-                hijack_merger_for_target_table(
-                    &mut graph,
-                    &columns_without_rw_timestamp,
-                    &sink,
-                    Some(&sink.unique_identity()),
-                )?;
-            }
-
-            affected_table_change = Some(ReplaceJobPlan {
-                replace_job: Some(replace_job_plan::ReplaceJob::ReplaceTable(
-                    replace_job_plan::ReplaceTable {
-                        table: Some(table.to_prost()),
-                        source: source.map(|x| x.to_prost()),
-                        job_type: target_job_type as _,
-                    },
-                )),
-                fragment_graph: Some(graph),
-            });
-        }
-
-        let catalog_writer = session.catalog_writer()?;
-        catalog_writer
-            .drop_sink(sink_id.sink_id, cascade, affected_table_change)
-            .await?;
-    }
+    let catalog_writer = session.catalog_writer()?;
+    catalog_writer.drop_sink(sink_id.sink_id, cascade).await?;
 
     Ok(PgResponse::empty_result(StatementType::DROP_SINK))
 }
--- a/src/frontend/src/optimizer/mod.rs
+++ b/src/frontend/src/optimizer/mod.rs
@@ -83,8 +83,8 @@ use crate::handler::create_table::{CreateTableInfo, CreateTableProps};
 use crate::optimizer::plan_node::generic::{GenericPlanRef, SourceNodeKind, Union};
 use crate::optimizer::plan_node::{
     Batch, BatchExchange, BatchPlanNodeType, BatchPlanRef, ConventionMarker, PlanTreeNode, Stream,
-    StreamExchange, StreamPlanRef, StreamUnion, StreamVectorIndexWrite, ToStream,
-    VisitExprsRecursive,
+    StreamExchange, StreamPlanRef, StreamUnion, StreamUpstreamSinkUnion, StreamVectorIndexWrite,
+    ToStream, VisitExprsRecursive,
 };
 use crate::optimizer::plan_visitor::{RwTimestampValidator, TemporalJoinValidator};
 use crate::optimizer::property::Distribution;
@@ -796,7 +796,16 @@ impl LogicalPlanRoot {
         };
 
         let with_external_source = source_catalog.is_some();
-        let union_inputs = if with_external_source {
+        let (dml_source_node, external_source_node) = if with_external_source {
+            let dummy_source_node = LogicalSource::new(
+                None,
+                columns.clone(),
+                row_id_index,
+                SourceNodeKind::CreateTable,
+                context.clone(),
+                None,
+            )
+            .and_then(|s| s.to_stream(&mut ToStreamContext::new(false)))?;
             let mut external_source_node = stream_plan.plan;
             external_source_node =
                 inject_project_for_generated_column_if_needed(&columns, external_source_node)?;
@@ -810,43 +819,24 @@ impl LogicalPlanRoot {
                     StreamExchange::new_no_shuffle(external_source_node).into()
                 }
             };
-
-            let dummy_source_node = LogicalSource::new(
-                None,
-                columns.clone(),
-                row_id_index,
-                SourceNodeKind::CreateTable,
-                context.clone(),
-                None,
-            )
-            .and_then(|s| s.to_stream(&mut ToStreamContext::new(false)))?;
-
-            let dml_node = inject_dml_node(
-                &columns,
-                append_only,
-                dummy_source_node,
-                &pk_column_indices,
-                kind,
-                column_descs,
-            )?;
-
-            vec![external_source_node, dml_node]
+            (dummy_source_node, Some(external_source_node))
         } else {
-            let dml_node = inject_dml_node(
-                &columns,
-                append_only,
-                stream_plan.plan,
-                &pk_column_indices,
-                kind,
-                column_descs,
-            )?;
-
-            vec![dml_node]
+            (stream_plan.plan, None)
         };
 
-        let dists = union_inputs
+        let dml_node = inject_dml_node(
+            &columns,
+            append_only,
+            dml_source_node,
+            &pk_column_indices,
+            kind,
+            column_descs,
+        )?;
+
+        let dists = external_source_node
             .iter()
             .map(|input| input.distribution())
+            .chain([dml_node.distribution()])
             .unique()
             .collect_vec();
 
@@ -861,13 +851,30 @@ impl LogicalPlanRoot {
             }
         };
 
+        let generated_column_exprs =
+            LogicalSource::derive_output_exprs_from_generated_columns(&columns)?;
+        let upstream_sink_union = StreamUpstreamSinkUnion::new(
+            context.clone(),
+            dml_node.schema(),
+            dml_node.stream_key(),
+            dist.clone(), // should always be the same as dist of `Union`
+            append_only,
+            row_id_index.is_none(),
+            generated_column_exprs,
+        );
+
+        let union_inputs = external_source_node
+            .into_iter()
+            .chain([dml_node, upstream_sink_union.into()])
+            .collect_vec();
+
         let mut stream_plan = StreamUnion::new_with_dist(
             Union {
                 all: true,
                 inputs: union_inputs,
                 source_col: None,
             },
-            dist.clone(),
+            dist,
         )
         .into();
 
--- a/src/frontend/src/optimizer/plan_node/stream_upstream_sink_union.rs
+++ b/src/frontend/src/optimizer/plan_node/stream_upstream_sink_union.rs
@@ -1,0 +1,175 @@
+// Copyright 2025 RisingWave Labs
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use educe::Educe;
+use pretty_xmlish::XmlNode;
+use risingwave_common::catalog::Schema;
+use risingwave_pb::stream_plan::UpstreamSinkUnionNode;
+use risingwave_pb::stream_plan::stream_node::PbNodeBody;
+
+use super::stream::prelude::*;
+use crate::OptimizerContextRef;
+use crate::expr::ExprImpl;
+use crate::optimizer::plan_node::expr_visitable::ExprVisitable;
+use crate::optimizer::plan_node::utils::{Distill, childless_record, watermark_pretty};
+use crate::optimizer::plan_node::{ExprRewritable, PlanBase, Stream, StreamNode};
+use crate::optimizer::property::{
+    Distribution, FunctionalDependencySet, MonotonicityDerivation, MonotonicityMap,
+    WatermarkColumns, analyze_monotonicity,
+};
+use crate::stream_fragmenter::BuildFragmentGraphState;
+
+#[derive(Debug, Clone, Educe)]
+#[educe(PartialEq, Eq, Hash)]
+struct UpstreamSinkUnionInner {
+    #[educe(PartialEq(ignore), Hash(ignore))]
+    ctx: OptimizerContextRef,
+    schema: Schema,
+    stream_key: Option<Vec<usize>>,
+    dist: Distribution,
+    stream_kind: StreamKind,
+    // `generated_column` is used to generate the `watermark_columns` field and affects the derivation of subsequent
+    // operators. Since the project operator of `generated_column` is actually on the sink-fragment, not on the table,
+    // only expr can be retained here to determine the `watermark_columns`.
+    generated_column_exprs: Option<Vec<ExprImpl>>,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct StreamUpstreamSinkUnion {
+    pub base: PlanBase<Stream>,
+    pub generated_column_exprs: Option<Vec<ExprImpl>>,
+}
+
+impl StreamUpstreamSinkUnion {
+    pub fn new(
+        ctx: OptimizerContextRef,
+        schema: &Schema,
+        stream_key: Option<&[usize]>,
+        dist: Distribution,
+        append_only: bool,
+        user_defined_pk: bool,
+        generated_column_exprs: Option<Vec<ExprImpl>>,
+    ) -> Self {
+        // For upstream sink creating, we require that if the table doesn't define pk or the table is `append_only`, the
+        // upstream sink must be `append_only`.
+        let stream_kind = if append_only || !user_defined_pk {
+            StreamKind::AppendOnly
+        } else {
+            StreamKind::Upsert
+        };
+        let inner = UpstreamSinkUnionInner {
+            ctx,
+            schema: schema.clone(),
+            stream_key: stream_key.map(|keys| keys.to_vec()),
+            dist,
+            stream_kind,
+            generated_column_exprs,
+        };
+
+        Self::new_inner(inner)
+    }
+
+    fn new_inner(inner: UpstreamSinkUnionInner) -> Self {
+        let mut out_watermark_columns = WatermarkColumns::new();
+        let mut out_monotonicity_map = MonotonicityMap::new();
+        // reference `StreamProject::new_inner()`
+        if let Some(ref generated_column_exprs) = inner.generated_column_exprs {
+            for (expr_idx, expr) in generated_column_exprs.iter().enumerate() {
+                if let MonotonicityDerivation::Inherent(monotonicity) = analyze_monotonicity(expr) {
+                    out_monotonicity_map.insert(expr_idx, monotonicity);
+                    if monotonicity.is_non_decreasing() && !monotonicity.is_constant() {
+                        out_watermark_columns.insert(expr_idx, inner.ctx.next_watermark_group_id());
+                    }
+                }
+            }
+        }
+
+        let field_num = inner.schema.fields().len();
+        let base = PlanBase::new_stream(
+            inner.ctx,
+            inner.schema,
+            inner.stream_key, // maybe incorrect
+            FunctionalDependencySet::new(field_num),
+            inner.dist,
+            inner.stream_kind,
+            false, // emit_on_window_close
+            out_watermark_columns,
+            out_monotonicity_map,
+        );
+
+        Self {
+            base,
+            generated_column_exprs: inner.generated_column_exprs,
+        }
+    }
+
+    fn rebuild_inner(&self) -> UpstreamSinkUnionInner {
+        UpstreamSinkUnionInner {
+            ctx: self.base.ctx(),
+            schema: self.base.schema().clone(),
+            stream_key: self.base.stream_key().map(|keys| keys.to_vec()),
+            dist: self.base.distribution().clone(),
+            stream_kind: self.base.stream_kind(),
+            generated_column_exprs: self.generated_column_exprs.clone(),
+        }
+    }
+}
+
+impl Distill for StreamUpstreamSinkUnion {
+    fn distill<'a>(&self) -> XmlNode<'a> {
+        let verbose = self.base.ctx().is_explain_verbose();
+        let mut vec = Vec::new();
+        if verbose && let Some(ow) = watermark_pretty(self.watermark_columns(), self.schema()) {
+            vec.push(("output_watermarks", ow));
+        }
+        childless_record("StreamUpstreamSinkUnion", vec)
+    }
+}
+
+impl_plan_tree_node_for_leaf! { Stream, StreamUpstreamSinkUnion }
+
+impl StreamNode for StreamUpstreamSinkUnion {
+    fn to_stream_prost_body(&self, _state: &mut BuildFragmentGraphState) -> PbNodeBody {
+        PbNodeBody::UpstreamSinkUnion(Box::new(UpstreamSinkUnionNode {
+            // When the table is created, there are no upstreams, so this is empty. When upstream sinks are created
+            // later, the actual upstreams in executor will increase, and during recovery, it will be filled with the
+            // actual upstream infos.
+            init_upstreams: vec![],
+        }))
+    }
+}
+
+impl ExprRewritable<Stream> for StreamUpstreamSinkUnion {
+    fn has_rewritable_expr(&self) -> bool {
+        self.generated_column_exprs.is_some()
+    }
+
+    fn rewrite_exprs(&self, r: &mut dyn crate::expr::ExprRewriter) -> super::PlanRef<Stream> {
+        let mut inner = self.rebuild_inner();
+        inner.generated_column_exprs = inner
+            .generated_column_exprs
+            .map(|exprs| exprs.into_iter().map(|expr| r.rewrite_expr(expr)).collect());
+        Self::new_inner(inner).into()
+    }
+}
+
+impl ExprVisitable for StreamUpstreamSinkUnion {
+    fn visit_exprs(&self, v: &mut dyn crate::expr::ExprVisitor) {
+        if let Some(exprs) = self.generated_column_exprs.as_ref() {
+            exprs.iter().for_each(|expr| {
+                v.visit_expr(expr);
+            });
+        }
+    }
+}
--- a/src/meta/src/barrier/command.rs
+++ b/src/meta/src/barrier/command.rs
@@ -36,11 +36,12 @@ use risingwave_hummock_sdk::vector_index::VectorIndexDelta;
 use risingwave_meta_model::WorkerId;
 use risingwave_pb::catalog::CreateType;
 use risingwave_pb::catalog::table::PbTableType;
-use risingwave_pb::common::ActorInfo;
+use risingwave_pb::common::PbActorInfo;
 use risingwave_pb::hummock::vector_index_delta::PbVectorIndexInit;
 use risingwave_pb::source::{
     ConnectorSplit, ConnectorSplits, PbCdcTableSnapshotSplitsWithGeneration,
 };
+use risingwave_pb::stream_plan::add_mutation::PbNewUpstreamSink;
 use risingwave_pb::stream_plan::barrier::BarrierKind as PbBarrierKind;
 use risingwave_pb::stream_plan::barrier_mutation::Mutation;
 use risingwave_pb::stream_plan::connector_props_change_mutation::ConnectorPropsInfo;
@@ -48,9 +49,9 @@ use risingwave_pb::stream_plan::stream_node::NodeBody;
 use risingwave_pb::stream_plan::throttle_mutation::RateLimit;
 use risingwave_pb::stream_plan::update_mutation::*;
 use risingwave_pb::stream_plan::{
-    AddMutation, BarrierMutation, CombinedMutation, ConnectorPropsChangeMutation, Dispatcher,
-    Dispatchers, DropSubscriptionsMutation, LoadFinishMutation, PauseMutation, PbSinkAddColumns,
-    ResumeMutation, SourceChangeSplitMutation, StartFragmentBackfillMutation, StopMutation,
+    AddMutation, ConnectorPropsChangeMutation, Dispatcher, Dispatchers, DropSubscriptionsMutation,
+    LoadFinishMutation, PauseMutation, PbSinkAddColumns, PbUpstreamSinkInfo, ResumeMutation,
+    SourceChangeSplitMutation, StartFragmentBackfillMutation, StopMutation,
     SubscriptionUpstreamInfo, ThrottleMutation, UpdateMutation,
 };
 use risingwave_pb::stream_service::BarrierCompleteResponse;
@@ -73,7 +74,8 @@ use crate::model::{
 };
 use crate::stream::{
     AutoRefreshSchemaSinkContext, ConnectorPropsChange, FragmentBackfillOrder,
-    JobReschedulePostUpdates, SplitAssignment, ThrottleConfig, build_actor_connector_splits,
+    JobReschedulePostUpdates, SplitAssignment, ThrottleConfig, UpstreamSinkInfo,
+    build_actor_connector_splits,
 };
 
 /// [`Reschedule`] is for the [`Command::RescheduleFragment`], which is used for rescheduling actors
@@ -276,7 +278,7 @@ pub struct SnapshotBackfillInfo {
 #[derive(Debug, Clone)]
 pub enum CreateStreamingJobType {
     Normal,
-    SinkIntoTable(ReplaceStreamJobPlan),
+    SinkIntoTable(UpstreamSinkInfo),
     SnapshotBackfill(SnapshotBackfillInfo),
 }
 
@@ -307,10 +309,12 @@ pub enum Command {
     /// After the barrier is collected, it notifies the local stream manager of compute nodes to
     /// drop actors, and then delete the job fragments info from meta store.
     DropStreamingJobs {
-        table_fragments_ids: HashSet<TableId>,
+        streaming_job_ids: HashSet<TableId>,
         actors: Vec<ActorId>,
         unregistered_state_table_ids: HashSet<TableId>,
         unregistered_fragment_ids: HashSet<FragmentId>,
+        // target_fragment -> [sink_fragments]
+        dropped_sink_fragment_by_targets: HashMap<FragmentId, Vec<FragmentId>>,
     },
 
     /// `CreateStreamingJob` command generates a `Add` barrier by given info.
@@ -403,13 +407,12 @@ impl std::fmt::Display for Command {
             Command::Pause => write!(f, "Pause"),
             Command::Resume => write!(f, "Resume"),
             Command::DropStreamingJobs {
-                table_fragments_ids,
-                ..
+                streaming_job_ids, ..
             } => {
                 write!(
                     f,
                     "DropStreamingJobs: {}",
-                    table_fragments_ids.iter().sorted().join(", ")
+                    streaming_job_ids.iter().sorted().join(", ")
                 )
             }
             Command::CreateStreamingJob { info, .. } => {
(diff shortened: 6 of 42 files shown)


Chiro11:
* **#23017** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23017?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#22920** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22920?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22920?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#22917** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22917?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

proto/ddl_service.proto
```diff
@@ -88,8 +88,7 @@ message AlterSourceResponse {
 message CreateSinkRequest {
   catalog.Sink sink = 1;
   stream_plan.StreamFragmentGraph fragment_graph = 2;
-  // It is used to provide a replace plan for the downstream table in `create sink into table` requests.
-  optional ReplaceJobPlan affected_table_change = 3;
+  optional ReplaceJobPlan affected_table_change = 3 [deprecated = true];
```
wenym1: Can just remove the field. No need to maintain backward-compatibility for pb only used in rpc.

src/frontend/src/optimizer/mod.rs
```diff
@@ -861,13 +861,24 @@ impl LogicalPlanRoot {
...
+        let generated_column_exprs =
+            LogicalSource::derive_output_exprs_from_generated_columns(&columns)?;
+        let upstream_sink_union = StreamUpstreamSinkUnion::new(
+            context.clone(),
+            dml_node.schema(),
+            append_only,
+            generated_column_exprs,
+        );
+        union_inputs.push(upstream_sink_union.into());
+
```
kwannoel, line 870: The code structure is something like:
```
let mut union_inputs = if with_external_source {
  let external_source_node = ...;
  let dml_node = ...;
  vec![external_source_node, dml_node]
} else {
  let dml_node = ...;
  vec![dml_node]
}

// extract dml node
let dml_node = union_inputs.last().unwrap();
...
let upstream_sink_union = StreamUpstreamSinkUnion::new(
    ...,
    dml_node.schema(), <---------- This is the only place where we reference the extract dml_node
    ...,
);
```

There are two observations:
1. `dml_node` is instantiated in both branches. But its input is `external_source` in one branch, existing `stream_plan` in the other. We can sequence their instantiation, to avoid writing instantiation code in both branches.
2. Using `last()` and `unwrap()` is error prone. We have to go up to check that `dml_node` is indeed instantiated at the last position always.

Suggest the following structure:
   ```
   let source = if with_external_source {
      ...
      LogicalSource::new(...)
   } else {
      stream_plan.plan // This is just `LogicalSource` (see planner test).
   }
   let dml_node = inject_dml_node(..., source);
   let table_schema = dml_node.schema();
   let union_inputs = if with_external_source {
     vec![source, dml_node]
   } else {
     vec![dml_node]
   }
   ```

src/frontend/src/optimizer/plan_node/stream_upstream_sink_union.rs
```diff
@@ -0,0 +1,145 @@
...
+#[derive(Debug, Clone, Educe)]
+#[educe(PartialEq, Eq, Hash)]
+struct UpstreamSinkUnionInner {
+    #[educe(PartialEq(ignore), Hash(ignore))]
+    ctx: OptimizerContextRef,
+    schema: Schema,
+    append_only: bool,
+    // `generated_column` is used to generate the `watermark_columns` field and affects the derivation of subsequent
+    // operators. Since the project operator of `generated_column` is actually on the sink-fragment, not on the table,
+    // only expr can be retained here to determine the non-decreasing column.
```
kwannoel:
```suggestion
    // only expr can be retained here to determine the `watermark_columns`.
```

src/frontend/src/optimizer/plan_node/stream_upstream_sink_union.rs
```diff
@@ -0,0 +1,145 @@
...
+                if let MonotonicityDerivation::Inherent(monotonicity) = analyze_monotonicity(expr) {
+                    out_monotonicity_map.insert(expr_idx, monotonicity);
+                    if monotonicity.is_non_decreasing() && !monotonicity.is_constant() {
+                        out_watermark_columns.insert(expr_idx, inner.ctx.next_watermark_group_id());
+                    }
+                }
+            }
+        }
+
+        // TODO(zyx): Maybe some fields are incorrect.
```
kwannoel: Could you annotate specific fields you're uncertain about?

src/frontend/src/optimizer/plan_node/stream_upstream_sink_union.rs
```diff
@@ -0,0 +1,145 @@
...
+            out_monotonicity_map,
+        );
+
+        Self { base, inner }
+    }
+}
+
+impl Distill for StreamUpstreamSinkUnion {
+    fn distill<'a>(&self) -> XmlNode<'a> {
+        childless_record("StreamUpstreamSinkUnion", vec![])
```
kwannoel: I think we should include things like `generated_columns` and other fields we operate on, and leave out no-op fields. We should do so in `verbose mode` at least. Currently the planner test just shows `StreamUpstreamSinkUnion`.

src/meta/src/controller/streaming_job.rs
```diff
@@ -544,6 +545,40 @@ impl CatalogController {
         Ok(())
     }
 
+    pub async fn build_cancel_command(
```
kwannoel, line 562: Could you add some docs here to elaborate on the cancellation procedure for sink-into-table?

src/meta/src/stream/stream_manager.rs
```diff
@@ -66,6 +68,17 @@ pub struct CreateStreamingJobOption {
     // leave empty as a placeholder for future option if there is any
 }
 
+#[derive(Debug, Clone)]
+pub struct UpstreamSinkInfo {
+    pub sink_id: ObjectId,
+    pub sink_fragment_id: FragmentId,
+    pub sink_output_fields: Vec<PbField>,
+    // for forward compatibility
```
kwannoel:
```suggestion
    // for backwards compatibility
```

src/meta/src/barrier/context/recovery.rs
```diff
@@ -249,6 +251,59 @@ impl GlobalBarrierWorkerContextImpl {
         Ok((table_committed_epoch, log_epochs))
     }
 
+    async fn recovery_table_with_upstream_sinks(
```
kwannoel, line 261: Could you elaborate on how recovery for the new sink into table design works?

e2e_test/sink/sink_into_table/alter_column.slt
```diff
@@ -252,3 +260,72 @@ drop table m_alter_1;
...
+
+query ITI rowsort
+select * from m_drop_1;
+----
+1
+2
+3
+
+statement ok
+alter table m_drop_1 add column v3 int default 1000;
```
wenym1, line 323: May incldue tests for scaling in/out for both upstream sinks and downstream table.

e2e_test/sink/sink_into_table/alter_column.slt
```diff
@@ -252,3 +260,72 @@ drop table m_alter_1;
...
+select * from m_drop_1;
+----
+1 11
+2 12
+3 13
+4 14
+5 15
+
+statement error dropping columns in target table of sinks is not supported
+alter table m_drop_1 drop column v2;
```
wenym1, line 348: Do we support dropping columns not sinked by any upstream sinks? Question on both current main branch and after this PR.

src/frontend/src/optimizer/mod.rs
```diff
@@ -827,7 +827,7 @@ impl LogicalPlanRoot {
                 dummy_source_node,
                 &pk_column_indices,
                 kind,
-                column_descs,
+                column_descs.clone(),
```
wenym1: The usage of `column_descs` seems unchanged. Is the newly added `.clone` necessary here?

src/frontend/src/optimizer/mod.rs
wenym1: Agree. We can even remove the `union_inputs` and always store `dml_node` and the optional `source`, and later construct the union node directly from them.

src/stream/src/task/actor_manager.rs
```diff
@@ -226,7 +224,7 @@ impl StreamActorManager {
         subtasks: &mut Vec<SubtaskHandle>,
         local_barrier_manager: &LocalBarrierManager,
         prefix_nodes: Vec<&stream_plan::StreamNode>,
-        merge_projects: Vec<(&stream_plan::StreamNode, &stream_plan::StreamNode)>,
+        _merge_projects: Vec<(&stream_plan::StreamNode, &stream_plan::StreamNode)>,
```
wenym1: Shall we remove the whole `fn create_sink_into_table_union`?

src/frontend/src/optimizer/plan_node/stream_upstream_sink_union.rs
```diff
@@ -0,0 +1,145 @@
...
+                    }
+                }
+            }
+        }
+
+        // TODO(zyx): Maybe some fields are incorrect.
+        let base = PlanBase::new_stream(
+            inner.ctx.clone(),
+            inner.schema.clone(),
+            Some(vec![]), // stream_key
```
wenym1: Should it share the same `stream_key` and `distribution` with the table?

src/frontend/src/optimizer/plan_node/stream_upstream_sink_union.rs
```diff
@@ -0,0 +1,145 @@
...
+            FunctionalDependencySet::new(inner.schema.fields().len()),
+            Distribution::SomeShard,
+            if inner.append_only {
+                StreamKind::AppendOnly
+            } else {
+                StreamKind::Retract
```
wenym1: The stream kind of sink into table input should always be `Upsert`, as different upstreams may modify the same stream key, and the operations are inconsistent.

src/frontend/src/optimizer/plan_node/stream_upstream_sink_union.rs
```diff
@@ -0,0 +1,145 @@
...
+            schema: schema.clone(),
+            append_only,
+            generated_column_exprs,
+        };
+
+        Self::new_inner(inner)
+    }
+
+    fn new_inner(inner: UpstreamSinkUnionInner) -> Self {
+        let mut out_watermark_columns = WatermarkColumns::new();
```
wenym1, line 84: Can we by any chance to reuse the logic in `Project` to obtain the watermark columns?

src/frontend/src/optimizer/plan_node/stream_upstream_sink_union.rs
```diff
@@ -0,0 +1,145 @@
...
+        childless_record("StreamUpstreamSinkUnion", vec![])
+    }
+}
+
+impl_plan_tree_node_for_leaf! { Stream, StreamUpstreamSinkUnion }
+
+impl StreamNode for StreamUpstreamSinkUnion {
+    fn to_stream_prost_body(&self, _state: &mut BuildFragmentGraphState) -> PbNodeBody {
+        PbNodeBody::UpstreamSinkUnion(Box::new(UpstreamSinkUnionNode {
+            init_upstreams: vec![],
```
wenym1, line 148: May add comments on why it's empty.

src/frontend/src/optimizer/mod.rs
```diff
@@ -861,13 +861,24 @@ impl LogicalPlanRoot {
             }
         };
 
+        let dml_node = union_inputs.last().unwrap();
+        let generated_column_exprs =
+            LogicalSource::derive_output_exprs_from_generated_columns(&columns)?;
```
wenym1, line 855: I'm even thinking about, shall we move the logic of expanding generated columns to after the `StreamUnion` on Dml+Source+UpstreamSinkUnion? IIUC, they should share the same logic to expanding the generated columns. In this way, we don't have to handle the generated columns and watermark columns in `StreamUpstreamSinkUnion`.

In the coming release, we would include this refactor, and we would anyway overwrite the fragment of existing tables for backward compatibility. We can overwrite the fragment stream node to include this shared generated column logic after `StreamUnion`.

src/frontend/src/handler/drop_sink.rs
```diff
@@ -64,60 +59,21 @@ pub async fn handle_drop_sink(
...
-                replace_job: Some(replace_job_plan::ReplaceJob::ReplaceTable(
-                    replace_job_plan::ReplaceTable {
-                        table: Some(table.to_prost()),
-                        source: source.map(|x| x.to_prost()),
-                        job_type: target_job_type as _,
-                    },
-                )),
-                fragment_graph: Some(graph),
-            });
+    match sink.stream_job_status {
```
wenym1: The syncing of `stream_job_status` is asynchronous. After this PR, the logic of two branch is very similar. Shall we change to unify the two branches to always calling `catalog_writer.drop_sink`, and determine whether it's a `cancel` or `drop` in meta node, or even there is no need to distinguish the two cases?

src/meta/src/barrier/command.rs
```diff
@@ -300,10 +302,13 @@ pub enum Command {
     /// After the barrier is collected, it notifies the local stream manager of compute nodes to
     /// drop actors, and then delete the job fragments info from meta store.
     DropStreamingJobs {
-        table_fragments_ids: HashSet<TableId>,
+        table_ids: HashSet<TableId>,
```
wenym1: May rename to `streaming_job_ids`.

src/meta/src/barrier/command.rs
```diff
@@ -300,10 +302,13 @@ pub enum Command {
...
         actors: Vec<ActorId>,
         unregistered_state_table_ids: HashSet<TableId>,
         unregistered_fragment_ids: HashSet<FragmentId>,
+        // used to update catalog in existing table
+        dropped_sink_in_existing_table: HashMap<SinkId, risingwave_meta_model::TableId>,
+        dropped_sink_fragments: Vec<FragmentId>,
```
wenym1: Can we just use the `unregistered_fragment_ids` instead of adding a new `dropped_sink_fragments`? The `UpstreamSinkUnionExecutor` should drop its upstream input when seeing the fragment id of any upstream is dropped, and no need to specially ensure whether it's the fragment of a sink.

src/meta/src/barrier/context/context_impl.rs
```diff
@@ -196,12 +197,20 @@ impl CommandContext {
...
             } => {
                 barrier_manager_context
                     .hummock_manager
                     .unregister_table_ids(unregistered_state_table_ids.iter().cloned())
                     .await?;
+                if !dropped_sink_in_existing_table.is_empty() {
+                    barrier_manager_context
+                        .metadata_manager
+                        .catalog_controller
+                        .update_table_remove_incoming_sinks(dropped_sink_in_existing_table)
```
wenym1: For ddl to drop a streaming job, we will remove it from catalog before we yield the barrier, to ensure that we don't have to wait for barrier collection to drop any problematic jobs. We should change to update the `incoming_sinks` in the same transaction to remove the sinks.

Besides, I'm not sure whether the `incoming_sinks` field should be deprecated, because we already store a `target_table` field in `sink`, and we can always restore the `incoming_sinks` of a table from the sinks' `target_table`. In this way, we don't have to maintain the consistency between the two fields.

wenym1 (commented):
Leaving some early comments. Will continue reviewing tomorrow.

src/meta/src/barrier/command.rs
Chiro11: This has been modified and currently uses `dropped_sink_fragment_with_targets: Vec<(FragmentId, FragmentId)>`, because it is needed to modify the operator itself during `pre_apply`.

e2e_test/sink/sink_into_table/alter_column.slt
Chiro11: No. It wasn't allowed before, and I haven't modified this part of the logic. Although I think it should work in theory.

src/frontend/src/optimizer/mod.rs
Chiro11: > dml_node is instantiated in both branches. But its input is external_source in one branch, existing stream_plan in the other.

That's not quite true. In the first branch, the input of `dml_node` is `dummy_source`. But I found a way to refactor anyway, like:
```
let (dml_source_node, external_source_node) = if with_external_source {
    ...
    (dummy_source_node, Some(external_source_node))
} else {
    (stream_plan.plan, None)
};
let dml_node = inject_dml_node(..., dml_source_node);
```

Thanks.

src/meta/src/barrier/context/context_impl.rs
Chiro11: I have also considered it, but `incoming_sinks` is used in many places. Maybe we can find a way to refactor it in another PR.

src/frontend/src/optimizer/mod.rs
Chiro11: Yes. However, the `generated_column` for the sink-into-table input is currently created on the sink side. If I modify this table fragment, do I also need to modify sink? And what about backward compatibility? Do we keep the original sink fragment?

wenym1: Oh, I see. I thought we are expanding the `generated_column` for the projection used in `StreamUpstreamSinkUnion`, but it seems that it's only for deriving the watermark columns. If so, let's just keep it the current way, though it looks dirty, because we have to apply the same logic for all the inputs of a table.

src/frontend/src/optimizer/plan_node/stream_upstream_sink_union.rs
```diff
@@ -0,0 +1,145 @@
...
+        }))
+    }
+}
+
+impl ExprRewritable<Stream> for StreamUpstreamSinkUnion {
+    fn has_rewritable_expr(&self) -> bool {
+        self.inner.generated_column_exprs.is_some()
```
wenym1: If the `generated_column_exprs` is only an intermediate information to derive `out_watermark_columns` and `out_monotonicity_map`, I think we don't have to store the exprs and thereby have to implement `ExprRewritable` and `ExprVisitable` for the plan node.

src/meta/src/barrier/context/context_impl.rs
wenym1: If maintaining consistency does not introduce unnecessary complexity in this PR, we can refactor it later. But anyway, we need to ensure that updating table's `incoming_sinks` and inserting/deleting sink from catalog should be in the same transaction to be atomic.

src/frontend/src/optimizer/plan_node/stream_upstream_sink_union.rs
Chiro11: There is a little trick here, which we have discussed offline before. The reason why `column_exprs` is saved here is because there will be another expr-rewrite later, and the `watermark_columns` needs to be derived after the rewrite. However, I don’t really understand the whole logic in rewriting. Although I think it is possible to rewrite it in advance or something, I am not sure how to implement. Any suggestions?

Chiro11: Maybe I found a way to rewrite in advance by implementing an empty `Rewriter`. Let me have a try.

src/meta/src/barrier/context/context_impl.rs
yezizp2012: +1 for updating all catalogs in the same transaction before issuing any barrier. Another benefit is that no additional cleaning work is required during the recovery process.

src/frontend/src/optimizer/plan_node/stream_upstream_sink_union.rs
Chiro11: Not good. The planner_test will fail due to some timestamp field. I think the real `Rewriter` is necessary.

src/meta/src/barrier/context/context_impl.rs
```diff
@@ -217,11 +226,8 @@ impl CommandContext {
                 job_type,
                 cross_db_snapshot_backfill_info,
             } => {
-                let mut replace_plan = None;
                 match job_type {
-                    CreateStreamingJobType::SinkIntoTable(plan) => {
-                        replace_plan = Some(plan);
-                    }
+                    CreateStreamingJobType::SinkIntoTable(_) => {}
```
yezizp2012: IIUC If `SinkIntoTable` includes cross-database backfilling, we should also set its snapshot backfill epoch as it is done for a `Normal` streaming job.

src/meta/src/barrier/context/recovery.rs
```diff
@@ -249,6 +251,59 @@ impl GlobalBarrierWorkerContextImpl {
...
+                    }
+                }
+                Ok(acc)
+            },
+        )?;
+        let job_ids = jobs.keys().cloned().collect_vec();
+        // Only `Table` will be returned here, ignoring other catalog objects.
+        let tables = self
+            .metadata_manager
+            .get_table_catalog_by_ids(&job_ids)
```
yezizp2012: Can we add a new function to only query table catalogs with `PbTableType::Table` for better performance?

src/meta/src/barrier/context/recovery.rs
yezizp2012, line 261: Could you add doc with details about this function?

yezizp2012: Do we need to adjust the clean logic for `sink into table` in `clean_dirty_streaming_jobs` during recovery?

e2e_test/sink/sink_into_table/alter_column.slt
```diff
@@ -178,6 +178,9 @@ select * from t_append_only;
 statement ok
 create sink s21 into t_append_only as select v1+1000 as v1, v1+2000 as v2, v1+3000 as v3 from t_s2 with (type = 'append-only', force_append_only = 'true');
 
+statement ok
+recover;
+
```
BugenZhao, line 183: Why do we add some `recover` here? While this covers the path for re-constructing the sink-into-table plan, will it also make some existing path inaccessible?

proto/ddl_service.proto
BugenZhao: This message is never persisted so it's safe to do breaking change.

src/frontend/src/optimizer/plan_node/stream_upstream_sink_union.rs
```diff
@@ -0,0 +1,145 @@
...
+};
+use crate::stream_fragmenter::BuildFragmentGraphState;
+
+#[derive(Debug, Clone, Educe)]
+#[educe(PartialEq, Eq, Hash)]
+struct UpstreamSinkUnionInner {
+    #[educe(PartialEq(ignore), Hash(ignore))]
+    ctx: OptimizerContextRef,
+    schema: Schema,
+    append_only: bool,
```
BugenZhao: These 3 fields can be directly obtained from `PlanBase<Stream>`. Shall we avoid duplicating them here, which is also how other plan nodes work?

e2e_test/sink/sink_into_table/alter_column.slt
Chiro11: Just to test whether recovery works. What do you mean "make some existing path inaccessible"? How will recovery affect the normal paths?

src/frontend/src/optimizer/plan_node/stream_upstream_sink_union.rs
Chiro11: It's a bit confusing here. First, `PlanBase` is created through the variables in `UpstreamSinkUnionInner`. Secondly, the reason we need the `new_inner` function is to rewrite the expression in the subsequent rewrite expr process and recreate the entire plan-node. If we want to reuse the logic of creating the plan-node (that is,  generate `watermark_columns` and then create a new `PlanBase`), we must retain the complete information required for creation in the parameters of `new_inner`. Otherwise, we need another function, such as just rewriting the expression, but considering that the plan-node is immutable in the `rewrite_exprs` function, we can only repeat the creation logic.

src/meta/src/barrier/context/recovery.rs
Chiro11: > Do we need to adjust the clean logic for `sink into table` in `clean_dirty_streaming_jobs` during recovery?

Good suggestion, I forgot about that.

src/frontend/src/optimizer/plan_node/stream_upstream_sink_union.rs
Chiro11: According to the existing implementation, when the table does not define a `pk`, or table is `append_only`, the upstream sink must be `append_only`. Therefore, the logic here is changed to:
```
let stream_kind = if append_only || user_defined_pk {
    StreamKind::AppendOnly
} else {
    StreamKind::Upsert
};
```

src/frontend/src/handler/drop_sink.rs
Chiro11: OK. This has been modified (including `drop_sink` and `drop_mv`, both moved to meta for check). It is difficult to not distinguish them for now: `cancel_creating_job` has some pre-checks, including the possibility of directly dropping unscheduled jobs from the scheduling queue.

src/meta/src/barrier/context/recovery.rs
Chiro11: Here, I found that there is no need to modify the original logic. The original cleanup logic should be retained (for backward compatibility). At the same time, for the new operator, since the operator's own data is not persisted, there is no need to clean it up.

src/frontend/src/optimizer/mod.rs
```diff
@@ -810,43 +819,24 @@ impl LogicalPlanRoot {
...
+            dml_source_node,
+            &pk_column_indices,
+            kind,
+            column_descs,
+        )?;
+
+        let dists = external_source_node
             .iter()
             .map(|input| input.distribution())
+            .chain(std::iter::once(dml_node.distribution()))
```
wenym1:
```suggestion
            .chain([dml_node.distribution()])
```

src/frontend/src/optimizer/plan_node/stream_upstream_sink_union.rs
```diff
@@ -0,0 +1,163 @@
...
+        }))
+    }
+}
+
+impl ExprRewritable<Stream> for StreamUpstreamSinkUnion {
+    fn has_rewritable_expr(&self) -> bool {
+        self.inner.generated_column_exprs.is_some()
+    }
+
+    fn rewrite_exprs(&self, _r: &mut dyn crate::expr::ExprRewriter) -> super::PlanRef<Stream> {
```
wenym1: `r` -> `_r`.

src/frontend/src/optimizer/plan_node/stream_upstream_sink_union.rs
```diff
@@ -0,0 +1,163 @@
...
+        schema: &Schema,
+        stream_key: Option<&[usize]>,
+        dist: Distribution,
+        append_only: bool,
+        user_defined_pk: bool,
+        generated_column_exprs: Option<Vec<ExprImpl>>,
+    ) -> Self {
+        // For upstream sink creating, we require that when the table defines pk or the table is `append_only`, the
+        // upstream sink must be `append_only`.
+        let stream_kind = if append_only || user_defined_pk {
```
wenym1: Should the condition be `append_only && !user_defined_pk`? At least in the most common case that user defines pk on the table, and the table is not append only, the stream kind should be upsert. 

According to our previous discussion offline, it seems that we want to set it as append-only to bypass the rejection on `Upsert` for `NoCheck` in `StreamMaterialize`. The `conflict_behavior` will be `Nocheck` under the following condition, which is `append_only && row_id_as_pk`, which is `append_only && !user_defined_pk`
```
let conflict_behavior = match self {
            EitherOnConflict::Ast(on_conflict) => {
                if append_only {
                    if row_id_as_pk {
                        // Primary key will be generated, no conflict check needed.
                        ConflictBehavior::NoCheck
```

Besides, want to raise a discussion about the expected `ConflictBehavior` for append-only table without user defined pk. For such a table, if we create a sink into it, even with a `RowIdGen` executor, the `Materialize` executor can still see the `Update` and `Delete` generated by the upstream sink. In this case, should the `ConflictBehavior` be `NoCheck`? Or we should change the behavior of `RowIdGen` to let it always generate a new row id for `UpdateInsert` and `Insert` and discard `Delete`, so that the `Materialize` can really be `NoCheck`. cc @BugenZhao

src/frontend/src/optimizer/plan_node/stream_upstream_sink_union.rs
```diff
@@ -0,0 +1,163 @@
...
+            generated_column_exprs,
+        };
+
+        Self::new_inner(inner)
+    }
+
+    fn new_inner(inner: UpstreamSinkUnionInner) -> Self {
+        let mut out_watermark_columns = WatermarkColumns::new();
+        let mut out_monotonicity_map = MonotonicityMap::new();
+        if let Some(ref generated_column_exprs) = inner.generated_column_exprs {
```
wenym1, line 87: Any comment on this previous comment https://github.com/risingwavelabs/risingwave/pull/22920/files#r2297287948? It seems that we have a same code in `stream_project`

src/meta/src/barrier/info.rs
```diff
@@ -469,6 +471,62 @@ impl InflightDatabaseInfo {
...
+                                                u, drop_upstream_fragment_ids, drop_fragment_id
+                                            );
+                                        }
+                                    }
+                                }
+                                u.init_upstreams.retain(|upstream| {
+                                    !drop_upstream_fragment_ids
+                                        .contains(&upstream.upstream_fragment_id)
+                                });
+                                removed = true;
```
wenym1, line 525: Should we move the `removed = true` to inside the closure of `retain`?

src/meta/src/barrier/info.rs
```diff
@@ -535,6 +596,11 @@ impl InflightDatabaseInfo {
                     })
                     .chain(replace_job.replace_upstream.keys())
             }))
+            .chain(
```
wenym1, line 601: Should update the comment above about `existing_fragment_ids consists of`

src/meta/src/barrier/command.rs
```diff
@@ -910,44 +934,43 @@ impl Command {
...
-                }));
+                };
 
-                if let CreateStreamingJobType::SinkIntoTable(ReplaceStreamJobPlan {
-                    old_fragments,
-                    init_split_assignment,
-                    replace_upstream,
-                    upstream_fragment_downstreams,
-                    cdc_table_snapshot_split_assignment,
+                if let CreateStreamingJobType::SinkIntoTable(UpstreamSinkInfo {
```
wenym1: Better to generate the `new_upstream_sinks` before composing the `AddMutation` and then fill in the final `new_upstream_sinks` of `AddMutation` directly.

src/meta/src/barrier/context/recovery.rs
```diff
@@ -249,6 +251,62 @@ impl GlobalBarrierWorkerContextImpl {
...
+            .metadata_manager
+            .catalog_controller
+            .get_user_created_table_by_ids(job_ids.into_iter().map(|id| id as _).collect())
+            .await?;
+        for table in tables {
+            assert_eq!(table.table_type(), PbTableType::Table);
+            let fragments = jobs.get_mut(&table.id).unwrap();
+            let mut union_fragment_id = None;
+            for fragment in fragments.fragment_infos.values_mut() {
+                if fragment.has_union_node() {
```
wenym1: Should we look for `UpstreamSinkUnion` directly?

src/meta/src/rpc/ddl_controller.rs
```diff
@@ -1897,16 +1517,40 @@ impl DdlController {
             StreamingJobId::Index(idx) => (idx as _, ObjectType::Index),
         };
 
-        let version = self
-            .drop_object(object_type, object_id, drop_mode, target_replace_info)
+        let job_status = self
```
wenym1, line 1526: Should we just follow the previous logic to always call `drop_object`? Theoratically, loading the job status and later handling is not atomic, and for correctness, we should always have a more atomic check in lower layer code, and therefore distinguishing `Creating` and `Created` here seems unnecessary.

src/meta/src/controller/catalog/drop_op.rs
```diff
@@ -319,6 +341,17 @@ impl CatalogController {
         inner
             .dropped_tables
             .extend(dropped_tables.map(|t| (TableId::try_from(t.id).unwrap(), t)));
+
```
wenym1, line 312: Do we need to notify the update if the table is also dropped together?

src/meta/src/controller/catalog/drop_op.rs
```diff
@@ -276,6 +255,49 @@ impl CatalogController {
...
+        let mut updated_objects = Vec::with_capacity(removed_sink_by_table.len());
+        for (table_id, removed_sink_ids) in &removed_sink_by_table {
+            let pb_table =
+                update_table_incoming_sinks(&txn, *table_id, &[], removed_sink_ids).await?;
+            updated_objects.push(PbObject {
+                object_info: Some(PbObjectInfo::Table(pb_table)),
+            });
+        }
+
+        let removed_sink_fragments =
```
wenym1: Here we have to first resolve the fragment id of sink, and then resolve the downstream fragment of the sink fragment. Is it more straightforward to just resolve the Materialized fragment of the downstream table? We can query the `Fragment` table and filter by `job_id` and `fragment_type_mask`.

src/meta/src/rpc/ddl_controller.rs
```diff
@@ -1771,36 +1401,27 @@ impl DdlController {
...
+                let union_fragment = stream_job_fragments.inner.union_fragment_for_table();
+                let upstream_infos = self
                     .metadata_manager
-                    .get_sink_catalog_by_ids(&table.incoming_sinks)
+                    .catalog_controller
+                    .get_all_upstream_sinks(table, union_fragment.fragment_id as _)
                     .await?;
+                refill_upstream_sink_union_in_table(&mut union_fragment.nodes, &upstream_infos);
+
+                for upstream_info in &upstream_infos {
```
wenym1, line 1428: May merge the two for-loop into one?

src/meta/src/controller/utils.rs
```diff
@@ -2158,6 +2167,188 @@ where
...
+where
+    C: ConnectionTrait,
+{
+    let fragments: Vec<(FragmentId, i32)> = Fragment::find()
+        .select_only()
+        .columns([
+            fragment::Column::FragmentId,
+            fragment::Column::FragmentTypeMask,
+        ])
+        .filter(fragment::Column::JobId.is_in(sink_ids))
```
wenym1: Can use [bit_and](https://docs.rs/sea-query/0.32.1/sea_query/expr/trait.ExprTrait.html#method.bit_and) to do the filter in the query. We can provide a helper method in `FragmentTypeMask` to generate the sea_orm expression

Chiro11:
> Is this PR ready for re-review? See some pending review comments.

Yes, all comments have been resolved.

wenym1 (commented):
Rest LGTM

src/frontend/src/optimizer/plan_node/stream_upstream_sink_union.rs
Chiro11: The logic is slightly different, because `upstream_sink_union` has no initial upstream, so it does not need to handle the input, it is only used for `generated_columns` generated by itself.

src/meta/src/barrier/info.rs
Chiro11: This flag is just to check whether the closure is triggered, that is, whether the `UpstreamSinkUnion` node is found. The check for `fragment_id` is placed in the `debug_assertion` above.

src/meta/src/barrier/command.rs
Chiro11: Why? `PbNewUpstreamSink` is only used before sending RPCs, so it seems reasonable to build it here.

src/meta/src/rpc/ddl_controller.rs
Chiro11: Do you mean calling `drop_object` for creating job? I think it is difficult. Ref: https://github.com/risingwavelabs/risingwave/pull/22920#discussion_r2303829709

src/meta/src/controller/catalog/drop_op.rs
Chiro11: No, if the table is dropped together, it should be included in `removed_objects`, and we should not need to update it.

src/frontend/src/optimizer/plan_node/stream_upstream_sink_union.rs
Chiro11: Yes, it should be `append_only && !user_defined_pk`. Fixed now.

BugenZhao: > For such a table, if we create a sink into it, even with a `RowIdGen` executor, the `Materialize` executor can still see the `Update` and `Delete` generated by the upstream sink.

Do you mean sinking a non-append-only stream into the append-only table? I thought this was not allowed. 🤔

src/frontend/src/optimizer/plan_node/stream_upstream_sink_union.rs
BugenZhao: I see. But I think we can always re-construct an `Inner` when we are going to call `new_inner` from existing fields. There doesn't seem to be a need to store some fields (duplicately).

src/meta/src/barrier/command.rs
wenym1: Just a suggestion on general coding style. Prefer composing a struct immutably as a whole, instead of initializing the struct mutably with some fields as default value, and later overwrite the fields.

src/meta/src/controller/catalog/drop_op.rs
wenym1: Oh, never mind. Only table `!removed_streaming_job_ids.contains(&target_table_id)` will be added to `removed_sink_by_table`.

src/meta/src/barrier/command.rs
Chiro11: Oh, I misunderstood creating it earlier in the process. Good suggestion.

src/frontend/src/optimizer/plan_node/stream_upstream_sink_union.rs
wenym1: If so, shall we forbid `Upsert` input for `StreamRowIdGen`? @BugenZhao

src/frontend/src/optimizer/plan_node/stream_upstream_sink_union.rs
```diff
@@ -0,0 +1,174 @@
...
+        user_defined_pk: bool,
+        generated_column_exprs: Option<Vec<ExprImpl>>,
+    ) -> Self {
+        // For upstream sink creating, we require that if the table doesn't define pk or the table is `append_only`, the
+        // upstream sink must be `append_only`.
+        let stream_kind = if append_only || !user_defined_pk {
+            StreamKind::AppendOnly
+        } else {
+            StreamKind::Upsert
+        };
```
wenym1, line 70: Seems that it's forbidden to create sink into a table without user specified pk (attempt to create such sink will see error like `Only append-only sinks can sink to a table without primary keys`). If so, may only check `append_only`, and assert that when `!append_only`, user_defined_pk must be true.

```suggestion
        let stream_kind = if append_only {
            StreamKind::AppendOnly
        } else {
            assert!(user_defined_pk);
            StreamKind::Upsert
        };
```

src/frontend/src/optimizer/plan_node/stream_upstream_sink_union.rs
```diff
@@ -0,0 +1,174 @@
...
+
+        Self::new_inner(inner)
+    }
+
+    fn new_inner(inner: UpstreamSinkUnionInner) -> Self {
+        let mut out_watermark_columns = WatermarkColumns::new();
+        let mut out_monotonicity_map = MonotonicityMap::new();
+        if let Some(ref generated_column_exprs) = inner.generated_column_exprs {
+            for (expr_idx, expr) in generated_column_exprs.iter().enumerate() {
+                if let MonotonicityDerivation::Inherent(monotonicity) = analyze_monotonicity(expr) {
```
wenym1, line 89: May at least add a comment that this code is copied from a similar logic in `StreamProject`.

Besides, we may mark variants other than `MonotonicityDerivation::Inherent` unreachable, or at least show a warning log.

src/frontend/src/optimizer/plan_node/stream_upstream_sink_union.rs
```diff
@@ -0,0 +1,174 @@
...
+        );
+
+        Self {
+            base,
+            generated_column_exprs: inner.generated_column_exprs,
+        }
+    }
+
+    fn rebuild_inner(&self) -> UpstreamSinkUnionInner {
+        UpstreamSinkUnionInner {
```
wenym1, line 118: What's the difference to `self.clone()`?

src/meta/src/controller/fragment.rs
```diff
@@ -1774,6 +1754,38 @@ impl CatalogController {
...
+
+        let mut upstream_sinks = Vec::with_capacity(target_table.get_incoming_sinks().len());
+        for sink_id in target_table.get_incoming_sinks() {
+            let (sink, sink_obj) = Sink::find_by_id(*sink_id as ObjectId)
+                .find_also_related(object::Entity)
+                .one(&txn)
+                .await?
+                .ok_or_else(|| MetaError::catalog_id_not_found("sink", *sink_id))?;
+            let pb_sink = ObjectModel(sink, sink_obj.unwrap()).into();
+            let sink_fragments = get_job_fragments_by_id(&txn, *sink_id as _).await?;
```
wenym1: In `build_upstream_sink_info`, we only use the `sink_fragment.node`. However, `get_job_fragments_by_id` is complex, and has to resolve actor and fragment relation to compose actor dispatchers, which is unused and unnecessary. We can actually query the `Fragment` table by `sink_id` and filter by fragment type mask to get the `PbStreamNode` of the sink fragment.

And if we don't have to reuse the `get_job_fragments_by_id` here, we can revert the related changes on it.

src/meta/src/controller/utils.rs
```diff
@@ -2172,6 +2182,187 @@ where
...
+        job_info.job_status.into(),
+        job_info.timezone.map(|tz| PbStreamContext { timezone: tz }),
+        fragment_actors,
+        job_info.parallelism.clone(),
+        job_info.max_parallelism as _,
+        job_definition,
+    )
+}
+
+pub async fn update_table_incoming_sinks<C>(
```
wenym1: It seems that we have spent some efforts to handle the `incoming_sinks` properly. We can have a predeceding PR to try first deprecate the it to simplify the current PR.

src/meta/src/controller/streaming_job.rs
```diff
@@ -557,6 +559,50 @@ impl CatalogController {
...
+    /// Builds a cancel command for the streaming job. If the sink (with target table) needs to be dropped, additional
+    /// information is required to build barrier mutation.
+    pub async fn build_cancel_command(
+        &self,
+        table_fragments: &StreamJobFragments,
+    ) -> MetaResult<Command> {
+        let inner = self.inner.read().await;
+        let txn = inner.db.begin().await?;
+
+        let dropped_sink_fragment_by_targets = if let Some((sink_id, table_id)) =
```
wenym1: Here we have to query meta store table for 3 times, first to get the target table id of the sink if there is any, and then get the fragment id of the sink, and then get the fragment id of mview. 

However, in `table_fragments`, it contains the fragments of the sink job, which contains the sink fragment, and we can get the sink fragment from it instead of querying the meta store table. And then we only need to query `FragmentRelation` table to obtain the downstream fragment id of the sink fragment id.

src/meta/src/controller/streaming_job.rs
```diff
@@ -651,6 +699,12 @@ impl CatalogController {
                 .one(&txn)
                 .await?
         {
+            let pb_table =
```
wenym1: No need to handle notify table update if we deprecate the `incoming_sinks` in `Table`.

src/meta/src/controller/catalog/drop_op.rs
```diff
@@ -276,6 +255,56 @@ impl CatalogController {
...
+                && !removed_streaming_job_ids.contains(&target_table_id)
+            {
+                removed_sink_by_table
+                    .entry(target_table_id)
+                    .or_default()
+                    .push(sink.sink_id);
+            }
+        }
+
+        let mut updated_objects = Vec::with_capacity(removed_sink_by_table.len());
```
wenym1: No need to notify table update after we deprecate `incoming_sinks` of table.

wenym1 (commented):
Rest LGTM.

Will review more after merging and rebasing https://github.com/risingwavelabs/risingwave/pull/23091

src/frontend/src/optimizer/plan_node/stream_upstream_sink_union.rs
Chiro11: Not true. The `append_only` attribute here is table's property. In fact, the upstream sink requires that the sink must be `append_only` when the table is `append_only` or when the table does not define pk.

Ref in e2e_test/sink/sink_into_table/alter_column.slt
```
statement ok

statement ok
create table t2 (v1 int, v2 int);

statement ok
create sink s1 into t2 from t1 with (type = 'append-only', force_append_only = 'true');
```

src/frontend/src/optimizer/plan_node/stream_upstream_sink_union.rs
Chiro11: > No need to store some fields (duplicately).

Ref: https://github.com/risingwavelabs/risingwave/pull/22920#discussion_r2313311255

src/meta/src/controller/streaming_job.rs
Chiro11: Good advice!

src/frontend/src/optimizer/plan_node/stream_upstream_sink_union.rs
Chiro11: The expression here is a complete project exprs, including generated columns and `FollowInput` columns, because we need to know the column index. Other variants will not be handled and no warning is required.

src/frontend/src/optimizer/plan_node/stream_upstream_sink_union.rs
wenym1: Got your point. 

When `!user_defined_pk`, the upstream sink must be append-only, and therefore we can treat the input as append-only no mather whether or not the table is append-only.

src/meta/src/controller/fragment.rs
```diff
@@ -1774,6 +1777,64 @@ impl CatalogController {
...
+        target_table: &PbTable,
+        target_fragment_id: FragmentId,
+    ) -> MetaResult<Vec<UpstreamSinkInfo>> {
+        let incoming_sinks = self.get_table_incoming_sinks(target_table.id as _).await?;
+
+        let inner = self.inner.read().await;
+        let txn = inner.db.begin().await?;
+
+        let sink_ids = incoming_sinks.iter().map(|s| s.id as SinkId).collect_vec();
+        let sink_fragment_ids = get_sink_fragment_by_ids(&txn, sink_ids).await?;
```
wenym1, line 1791: In `get_sink_fragment_by_ids`, `sink_ids` is used only in the `in` condition, and the output sink fragment id does not ensure the original order, and therefore seems incorrect to correlate the two iters with a `zip_eq_debug`.
```
fragment::Column::JobId
      .is_in(sink_ids)
      .and(FragmentTypeMask::intersects(FragmentTypeFlag::Sink))
```

Chiro11:
> Rest LGTM.
> 
> Will review more after merging and rebasing #23091

Already rebased. @wenym1

src/meta/src/controller/streaming_job.rs
```diff
@@ -557,6 +557,48 @@ impl CatalogController {
...
+                sink_target_fragment
+                    .get(&sink_fragment_id)
+                    .and_then(|target_fragments| {
+                        assert!(
+                            !target_fragments.is_empty(),
+                            "sink should have at least one downstream fragment"
+                        );
+                        target_fragments.first().cloned()
+                    })
+                    .map(|target_fragment_id| (sink_fragment_id, target_fragment_id))
```
wenym1:
```suggestion
                    .map(|target_fragments| {
                        let target_fragment_id = target_fragments.first().expect("sink should have at least one downstream fragment").clone();
                       (sink_fragment_id, target_fragment_id) 
                    })
```

If we have asserted that `!target_fragments.is_empty()`, we can use `.map` directly.

src/meta/src/barrier/context/recovery.rs
```diff
@@ -250,6 +254,69 @@ impl GlobalBarrierWorkerContextImpl {
...
+            for fragment in fragments.fragment_infos.values_mut() {
+                let mut is_target_fragment = false;
+                visit_stream_node_body(&fragment.nodes, |body| {
+                    if let PbNodeBody::UpstreamSinkUnion(_) = body {
+                        is_target_fragment = true;
+                    }
+                });
+                if is_target_fragment {
+                    // Each table should have only one target fragment.
+                    if let Some(ref target_fragment_id) = target_fragment_id {
```
wenym1: Can break the for-loop when finding the target fragment id.

src/meta/src/barrier/context/recovery.rs
```diff
@@ -250,6 +254,69 @@ impl GlobalBarrierWorkerContextImpl {
...
+                        assert_eq!(*target_fragment_id, fragment.fragment_id);
+                    } else {
+                        target_fragment_id = Some(fragment.fragment_id);
+                    }
+                }
+            }
+            let target_fragment_id =
+                target_fragment_id.expect("Table should have upstream-sink-union node");
```
wenym1: Better to turn it into an error instead of panic in case that our backward-compatibility support has bug, and then meta node will keep crash-looping.

src/meta/src/controller/fragment.rs
Chiro11: Thanks for pointing out this serious mistake.

wenym1 (approved):
LGTM!

PR #22925, opened 2025-08-19
feat: support sink payload as bytes into downstream mq

tabVersion:

Adds BYTES encoding for Kafka sinks in both plain ~~and upsert~~ formats.

* Enforces schema validation for BYTES encoding (single BYTEA column for values).
* Adds end-to-end SQLLogicTest to validate success and failure cases; wires test into CI.

Summary

* Map Encode::Bytes through both the frontend binder and connector catalog.
* Allow BYTES in the Kafka connector compatibility matrix for PLAIN FORMAT.

Validation rules in BytesEncoder:

* As value encoder (no key encode): require exactly one column and it must be BYTEA.

## Checklist

## Documentation

<details>

```sql
create table t_single_bytea (data bytea);
create sink sink_bytes_format from t_single_bytea with (
  connector = 'kafka',
  properties.bootstrap.server = 'message_queue:29092',
  topic = 'test-rw-sink-bytes-format'
) format plain encode bytes (force_append_only = 'true');
```

</details>

### DIFF
--- a/e2e_test/sink/kafka/test_bytes_format.slt
+++ b/e2e_test/sink/kafka/test_bytes_format.slt
@@ -47,20 +47,35 @@ create table t_non_bytea (
     data varchar
 );
 
-statement error BYTES format requires the column to be of type BYTEA, but got type character varying
+statement error
 create sink sink_bytes_format_error2 from t_non_bytea with (
     connector = 'kafka',
     properties.bootstrap.server = 'message_queue:29092',
     topic = 'test-rw-sink-bytes-format-error2'
 ) format plain encode bytes (force_append_only = 'true');
+----
+db error: ERROR: Failed to run the query
+
+Caused by these errors (recent errors listed first):
+  1: gRPC request to meta service (call `/ddl_service.DdlService/CreateSink`) failed: Internal error
+  2: failed to validate sink
+  3: config error
+  4: ENCODE BYTES requires the column to be of type BYTEA, but got type character varying
+
 
 # Test that upsert format with bytes encoding is not supported
-statement error connector kafka does not support format Upsert with encode Bytes
+statement error
 create sink sink_upsert_bytes_error from t_single_bytea with (
     connector = 'kafka',
     properties.bootstrap.server = 'message_queue:29092',
     topic = 'test-rw-sink-upsert-bytes-error'
 ) format upsert encode bytes;
+----
+db error: ERROR: Failed to run the query
+
+Caused by:
+  Bind error: connector kafka does not support format Upsert with encode Bytes
+
 
 # Test bytes format with key encoding
 statement ok


xiangjinwu:
Let's remove the support for `upsert`, given the key would have to be the same as payload, which is useless

src/connector/src/sink/formatter/mod.rs
```diff
@@ -215,15 +218,40 @@ fn ensure_only_one_pk<'a>(
...
+                        schema_ref.data_type
+                    )))
+                }
+            }
+            // This is being used as a value encoder
+            None => {
+                // Ensure the schema has exactly one column and it's of type BYTEA
+                if params.schema.len() != 1 {
+                    return Err(SinkError::Config(anyhow!(
+                        "BYTES format requires exactly one column, got {} columns",
```
xiangjinwu:
```suggestion
                        "ENCODE BYTES requires exactly one column, got {} columns",
```

src/connector/src/sink/formatter/mod.rs
```diff
@@ -215,15 +218,40 @@ fn ensure_only_one_pk<'a>(
...
+                        params.schema.len()
+                    )));
+                }
+
+                let field = &params.schema.fields[0];
+                if let DataType::Bytea = field.data_type {
+                    Ok(BytesEncoder::new(params.schema, 0))
+                } else {
+                    Err(SinkError::Config(anyhow!(
+                        "BYTES format requires the column to be of type BYTEA, but got type {}",
```
xiangjinwu:
```suggestion
                        "ENCODE BYTES requires the column to be of type BYTEA, but got type {}",
```

src/connector/src/sink/formatter/mod.rs
```diff
@@ -92,6 +92,9 @@ pub enum SinkFormatterImpl {
     AppendOnlyTemplate(AppendOnlyFormatter<TemplateEncoder, TemplateEncoder>),
     AppendOnlyTextTemplate(AppendOnlyFormatter<TextEncoder, TemplateEncoder>),
     AppendOnlyBytesTemplate(AppendOnlyFormatter<BytesEncoder, TemplateEncoder>),
+    AppendOnlyBytes(AppendOnlyFormatter<JsonEncoder, BytesEncoder>),
+    AppendOnlyTextBytes(AppendOnlyFormatter<TextEncoder, BytesEncoder>),
+    AppendOnlyBytesBytes(AppendOnlyFormatter<BytesEncoder, BytesEncoder>),
```
xiangjinwu:
```suggestion
    AppendOnlyBytes(AppendOnlyFormatter<BytesEncoder, BytesEncoder>),
```
Similarly, it makes no sense to allow key encode of text or json either.

ci/scripts/e2e-sink-test.sh
```diff
@@ -95,6 +95,7 @@ risedev slt './e2e_test/sink/file_sink.slt'
 risedev slt './e2e_test/sink/license.slt'
 risedev slt './e2e_test/sink/rate_limit.slt'
 risedev slt './e2e_test/sink/auto_schema_change.slt'
+risedev slt 'e2e_test/sink/kafka/test_bytes_format.slt'
```
xiangjinwu: Maybe this shall be added here instead (and I am wondering whether we can make it a wildcard - not required in this PR):
https://github.com/risingwavelabs/risingwave/blob/7692aa89c56c0236613b4d57e3d3da28170837a9/ci/scripts/e2e-kafka-sink-test.sh#L149-L163

src/connector/src/sink/formatter/mod.rs
```diff
@@ -92,6 +92,7 @@ pub enum SinkFormatterImpl {
     AppendOnlyTemplate(AppendOnlyFormatter<TemplateEncoder, TemplateEncoder>),
     AppendOnlyTextTemplate(AppendOnlyFormatter<TextEncoder, TemplateEncoder>),
     AppendOnlyBytesTemplate(AppendOnlyFormatter<BytesEncoder, TemplateEncoder>),
+    AppendOnlyBytes(AppendOnlyFormatter<JsonEncoder, BytesEncoder>),
```
xiangjinwu:
```suggestion
    AppendOnlyBytes(AppendOnlyFormatter<BytesEncoder, BytesEncoder>),
```
nit: although not useful, we shall keep consistent that default key encoder is same as value (except protobuf), rather than default to json.

src/connector/src/sink/formatter/mod.rs
```diff
@@ -553,13 +580,12 @@ impl SinkFormatterImpl {
...
-                | (F::Debezium, E::Json, Some(_))
-                | (F::Debezium, E::Avro | E::Protobuf | E::Template | E::Text, _)
-                | (F::AppendOnly, E::Bytes, _)
                 | (F::Upsert, E::Bytes, _)
-                | (F::Debezium, E::Bytes, _)
+                | (F::Debezium, E::Json, Some(_))
+                | (F::Debezium, E::Avro | E::Protobuf | E::Template | E::Text | E::Bytes, _)
                 | (_, E::Parquet, _)
                 | (_, _, Some(E::Parquet))
+                | (F::AppendOnly, E::Bytes, _)
```
xiangjinwu:
```suggestion
                | (F::AppendOnly, E::Bytes, Some(_))
```
given `None` is already covered above

xiangjinwu (approved):
rest lgtm

src/connector/src/sink/formatter/mod.rs
tabVersion: this is kept to make it compile. Some connector, eg. Kinesis, requires the key must be string rather than bytes.

src/connector/src/sink/formatter/mod.rs
```diff
@@ -649,6 +676,7 @@ macro_rules! dispatch_sink_formatter_str_key_impl {
             SinkFormatterImpl::UpsertTemplate($name) => $body,
             SinkFormatterImpl::AppendOnlyBytesTemplate(_) => unreachable!(),
             SinkFormatterImpl::UpsertBytesTemplate(_) => unreachable!(),
+            SinkFormatterImpl::AppendOnlyBytes($name) => $body,
```
xiangjinwu:
```suggestion
            SinkFormatterImpl::AppendOnlyBytes($name) => unreachable!(),
```

src/connector/src/sink/formatter/mod.rs
xiangjinwu: See comment below
https://github.com/risingwavelabs/risingwave/pull/22925#discussion_r2591693276

tabVersion: ~~Let's keep the name `AppendOnlyBytesBytes`.~~
~~Seems `AppendOnly<Encode>` has the convention that the key encode defaults to Json.~~

PR #22927, opened 2025-08-19
feat(sink): support copy into wihout barrier for redshift

xxhZs:

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/src/sink/snowflake_redshift/file_manager_util.rs
+++ b/src/connector/src/sink/snowflake_redshift/file_manager_util.rs
@@ -12,14 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use risingwave_meta_model::snowflake_redshift_sink_file::{
-    self, ActiveModel, Column, Entity, Model,
-};
-use sea_orm::{ColumnTrait, DatabaseConnection, EntityTrait, Order, QueryFilter, QueryOrder, Set};
+use risingwave_meta_model::snowflake_redshift_sink_file::{ActiveModel, Column, Entity, Model};
+use sea_orm::{ColumnTrait, DatabaseConnection, EntityTrait, QueryFilter, Set};
 use thiserror_ext::AsReport;
 
 use crate::sink::Result;
-use crate::sink::snowflake_redshift::file_manager_util;
 
 pub async fn insert_file_paths_with_sink_id(
     sink_id: u32,
@@ -30,7 +27,7 @@ pub async fn insert_file_paths_with_sink_id(
     let m = ActiveModel {
         sink_id: Set(sink_id as i32),
         end_epoch: Set(end_epoch.try_into().unwrap()),
-        file_paths: Set(file_paths),
+        file_paths: Set(risingwave_meta_model::StringArray::from(file_paths)),
     };
     match Entity::insert(m).exec(&db).await {
         Ok(_) => Ok(()),
@@ -54,7 +51,7 @@ pub async fn get_file_paths_by_sink_id(
     let mut max_end_epoch = 0;
 
     for model in models {
-        file_paths.extend(model.file_paths);
+        file_paths.extend(model.file_paths.into_string_array());
         max_end_epoch = max_end_epoch.max(model.end_epoch as u64);
     }
 
--- a/src/connector/src/sink/snowflake_redshift/file_manager_util.rs
+++ b/src/connector/src/sink/snowflake_redshift/file_manager_util.rs
@@ -66,7 +66,7 @@ pub async fn delete_row_by_sink_id_and_end_epoch(
     let end_epoch_i64: i64 = end_epoch.try_into().unwrap();
     match Entity::delete_many()
         .filter(Column::SinkId.eq(sink_id))
-        .filter(Column::EndEpoch.lt(end_epoch_i64))
+        .filter(Column::EndEpoch.lte(end_epoch_i64))
         .exec(db)
         .await
     {
--- a/src/connector/src/sink/snowflake_redshift/file_manager_util.rs
+++ b/src/connector/src/sink/snowflake_redshift/file_manager_util.rs
@@ -1,0 +1,103 @@
+// Copyright 2025 RisingWave Labs
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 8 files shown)


src/connector/src/sink/snowflake_redshift/redshift.rs
```diff
@@ -98,10 +101,15 @@ pub struct RedShiftConfig {
     pub create_table_if_not_exists: bool,
 
     #[serde(default = "default_schedule")]
-    #[serde(rename = "schedule_seconds")]
+    #[serde(rename = "schedule_seconds", alias = "merge_into_schedule_seconds")]
```
hzxa21: how about calling it `write.target.interval.seconds`

src/connector/src/sink/snowflake_redshift/redshift.rs
```diff
@@ -98,10 +101,15 @@ pub struct RedShiftConfig {
...
     #[serde_as(as = "DisplayFromStr")]
     pub schedule_seconds: u64,
 
+    #[serde(default = "default_schedule")]
+    #[serde(rename = "copy_into_schedule_seconds")]
```
hzxa21: how about calling it `write.intermediate.interval.seconds`

src/connector/src/sink/snowflake_redshift/redshift.rs
```diff
@@ -441,111 +454,188 @@ impl RedShiftSinkJdbcWriter {
...
-                .await;
-            });
-            (Some(periodic_task_handle), Some(shutdown_sender))
-        } else {
-            (None, None)
+        let merge_into_schedule_seconds = config.schedule_seconds;
+        let copy_into_schedule_seconds = config.copy_into_schedule_seconds;
+
+        let (periodic_task_handle, shutdown_sender) = match (is_append_only, config.with_s3) {
+            (true, true) | (false, _) => {
```
hzxa21, line 483: I think we don't need to run copy into if with_s3 is false?

src/connector/src/sink/snowflake_redshift/redshift.rs
```diff
@@ -441,111 +454,188 @@ impl RedShiftSinkJdbcWriter {
...
     shutdown_sender: Option<tokio::sync::mpsc::UnboundedSender<()>>,
 }
 
 impl RedshiftSinkCommitter {
     pub fn new(
+        db: DatabaseConnection,
         config: RedShiftConfig,
         is_append_only: bool,
         pk_column_names: &Vec<String>,
         all_column_names: &Vec<String>,
```
hzxa21, line 475: Not related to this PR but need to fix before merging to main: I don't see we validate `pk_column_names` and `all_column_names` during `validate` if `create_table_if_not_exists` is set to false.

src/connector/src/sink/snowflake_redshift/redshift.rs
```diff
@@ -441,111 +454,188 @@ impl RedShiftSinkJdbcWriter {
...
+                tokio::select! {
+                    _ = shutdown_receiver.recv() => break,
+                    _ = copy_timer.tick() => {
+                        if let Err(e) = async {
+                            let (paths, epoch) = get_file_paths_by_sink_id(&db, sink_id).await?;
+                            Self::copy_into_from_s3_to_redshift(&client, &config, true, sink_id, paths).await?;
+                            delete_row_by_sink_id_and_end_epoch(&db, sink_id, epoch).await?;
+                            Ok::<(),SinkError>(())
+                        }.await {
+                            tracing::error!("Failed to execute copy into task for sink id {}: {}", sink_id, e.as_report());
```
hzxa21: You can use the if branch in tokio::select to avoid duplicate codes:
```
_ = merge_timer.tick() if merge_into_sql.is_some() => ...
```

src/connector/src/sink/snowflake_redshift/redshift.rs
```diff
@@ -441,111 +454,188 @@ impl RedShiftSinkJdbcWriter {
...
+            &manifest_path,
+            &s3_inner.access,
+            &s3_inner.secret,
+            &s3_inner.assume_role,
+        )?;
+        client.execute_sql_sync(vec![copy_into_sql]).await?;
+        Ok(())
+    }
+
     async fn run_periodic_query_task(
```
hzxa21, line 587: I think we can have a common `run_periodic_query_task` method for both redshift and snowflake, and have a trait impl for both sinks to generate the copy into and merge into sqls.

src/connector/src/sink/snowflake_redshift/file_manager_util.rs
```diff
@@ -0,0 +1,100 @@
...
+
+pub async fn delete_row_by_sink_id_and_end_epoch(
+    db: &DatabaseConnection,
+    sink_id: u32,
+    end_epoch: u64,
+) -> Result<()> {
+    let end_epoch_i64: i64 = end_epoch.try_into().unwrap();
+    match Entity::delete_many()
+        .filter(Column::SinkId.eq(sink_id))
+        .filter(Column::EndEpoch.lt(end_epoch_i64))
```
hzxa21: should be `lte`?

src/connector/src/sink/snowflake_redshift/mod.rs
```diff
@@ -31,6 +31,7 @@ use crate::sink::file_sink::opendal_sink::FileSink;
 use crate::sink::file_sink::s3::{S3Common, S3Sink};
 use crate::sink::{Result, SinkError, SinkWriterParam};
 
+pub mod file_manager_util;
```
hzxa21, line 34: Not related to this PR but need to fix before merging to main: I think we can have a better structure for the common codes. We can put the common codes in a `intermeidate_sink_utils` rs file or mod instead of having a `snowflake_redshift` mod because the common logics are not related to a specific sink.

src/connector/src/sink/snowflake_redshift/snowflake.rs
```diff
@@ -296,6 +307,7 @@ impl Sink for SnowflakeSink {
             let client = SnowflakeJniClient::new(client, snowflake_task_ctx);
             client.execute_create_table().await?;
             client.execute_create_pipe().await?;
+            println!("Snowflake sink das");
```
hzxa21: remove

PR #22931, opened 2025-08-19
feat: support multiple version columns in conflict resolution

yuhao-su:

This PR extends the `WITH VERSION COLUMN` syntax to support multiple columns for lexicographic comparison during conflict resolution operations.

**Key Changes:**
- Modified protobuf schema to use `version_column_indices` (repeated field) instead of `version_column_index` (single field)
- Updated parser to accept multiple column names in `WITH VERSION COLUMN(col1, col2, ...)` syntax
- Implemented lexicographic comparison logic for multiple version columns
- Added comprehensive validation for version column data types
- Added database migration to handle the schema transition
- Extensive test coverage for various scenarios and edge cases

**How it works:**
When multiple version columns are specified, the conflict resolution compares them lexicographically (left-to-right priority). For example, with `version_column(v1, v2)`, a new row with `(v1=2, v2=1)` will override an existing row with `(v1=1, v2=5)` because `v1=2 > v1=1`.

**Backward Compatibility:**
Single version column syntax continues to work exactly as before.

## Checklist

## Documentation

<details>

**Enhanced conflict resolution with multiple version columns**

Users can now specify multiple columns for version-based conflict resolution in tables using the syntax:
```sql
CREATE TABLE t (id INT, v1 INT, v2 VARCHAR, data TEXT, PRIMARY KEY(id))
ON CONFLICT DO UPDATE IF NOT NULL WITH VERSION COLUMN(v1, v2);
```

When conflicts occur, columns are compared lexicographically (left-to-right priority) to determine which row should be kept. This enables more sophisticated conflict resolution strategies for CDC and upsert scenarios.

The existing single version column syntax remains fully supported for backward compatibility.

</details>

🤖 Generated with [Claude Code](https://claude.ai/code)


src/meta/model/migration/src/m20250819_014448_add_version_column_indices_to_table.rs
```diff
@@ -0,0 +1,121 @@
...
+            manager
+                .exec_stmt(
+                    Query::update()
+                        .table(Table::Table)
+                        .value(Table::VersionColumnIndices, i32_array)
+                        .and_where(Expr::col(Table::TableId).eq(table_entity.table_id))
+                        .to_owned(),
+                )
+                .await?;
+        }
```
yuhao-su: I can't seems find a way to use a `set` query to set the new column to json_array(old_value) @yezizp2012

src/meta/model/migration/src/m20250819_014448_add_version_column_indices_to_table.rs
```diff
@@ -0,0 +1,121 @@
...
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        // Add the new version_column_indices column
+        manager
+            .alter_table(
+                MigrationTable::alter()
+                    .table(Table::Table)
+                    .add_column(
+                        ColumnDef::new(Table::VersionColumnIndices)
+                            .json_binary()
+                            .not_null(),
```
yezizp2012: Consider setting a default value for this column or making it nullable to avoid failures with existing historical data.

src/meta/model/migration/src/m20250819_014448_add_version_column_indices_to_table.rs
yezizp2012: How about using something like `Expr::value("[]")` and `Expr::value("[1]")`? Not sure if it works but found some usage in https://github.com/risingwavelabs/risingwave/blob/main/src/meta/model/migration/src/m20241025_062548_singleton_vnode_count.rs

src/meta/model/migration/src/m20250819_014448_add_version_column_indices_to_table.rs
yuhao-su: you are commenting on an outdated commit. pls check the latest

.claude/agents/code-reviewer.md
BugenZhao, line 1: Move into a separate PR?

e2e_test/streaming/with_version_column.slt
```diff
@@ -88,4 +88,170 @@ statement ok
...
+query I?II? rowsort
+select id, timestamp_ver, string_ver, int_ver, data from t4;
+----
+1 2023-01-02 10:00:00 v2 2 newer_int
+
+statement ok
+drop table t4;
+
+# Test error cases for multiple version columns
+statement error
```
BugenZhao: Could you please also attach the error message pattern here?

proto/catalog.proto
```diff
@@ -476,10 +476,13 @@ message Table {
...
-  // This field specifies the index of the column set in the "with version
+  reserved 38;
+  // deprecated, use "version_column_indices" instead.
+  reserved "version_column_index";
+  // This field specifies the indices of the columns set in the "with version
   // column" within all the columns. It is used for filtering during "on
-  // conflict" operations.
-  optional uint32 version_column_index = 38;
+  // conflict" operations. Multiple columns will be compared lexicographically.
+  repeated uint32 version_column_indices = 47;
```
BugenZhao: I think `repeated` with `[packed=false]` can be backward compatible with singular field. So there's no need to create a new field here.

- https://protobuf.dev/programming-guides/encoding/#packed
- https://protobuf.dev/programming-guides/proto3/#conditionally-safe-changes

BugenZhao: Also, it might be noted that there are still table catalog stored in the plan node definition. SQL migration on `Table` might not be sufficient.

src/sqlparser/src/parser.rs
```diff
@@ -2529,6 +2529,18 @@ impl Parser<'_> {
         }
     }
 
+    pub fn parse_with_version_columns(&mut self) -> ModalResult<Vec<Ident>> {
+        if self.parse_keywords(&[Keyword::WITH, Keyword::VERSION, Keyword::COLUMN]) {
+            self.expect_token(&Token::LParen)?;
+            let columns =
+                self.parse_comma_separated(|parser| parser.parse_identifier_non_reserved())?;
```
BugenZhao, line 2536: Shall we reject the case where `columns` is empty?

src/sqlparser/src/parser.rs
```diff
@@ -2557,7 +2569,7 @@ impl Parser<'_> {
 
         let on_conflict = self.parse_on_conflict()?;
 
-        let with_version_column = self.parse_with_version_column()?;
```
BugenZhao, line 2571: Can we remove the old function?

src/stream/src/executor/mview/materialize.rs
```diff
@@ -790,13 +792,31 @@ fn replace_if_not_null(row: &mut Vec<Option<ScalarImpl>>, replacement: OwnedRow)
...
+
+        match cmp_datum(old_value, new_value, OrderType::ascending_nulls_first()) {
+            std::cmp::Ordering::Less => return true,     // new is newer
+            std::cmp::Ordering::Greater => return false, // old is newer
+            std::cmp::Ordering::Equal => continue,       // equal, check next column
+        }
+    }
+
+    // All version columns are equal, consider new version as equal (should overwrite)
+    true
```
BugenZhao, line 982: Shall we also demonstrate this in e2e test?

src/sqlparser/src/parser.rs
yuhao-su: It will simply means no version column, should be fine keep it this way

src/meta/model/migration/src/m20250819_014448_add_version_column_indices_to_table.rs
yuhao-su: Can't find a sql syntax that can be used for all of mysql sqllite and pg. But as long as there are not a huge amount of table with version columns this should not case performance issue.

e2e_test/backwards-compat-tests/scripts/utils.sh
```diff
@@ -244,6 +244,15 @@ seed_old_cluster() {
...
 
+  # Test version columns backwards compatibility, if OLD_VERSION <= 2.6.0
+  if version_le "$OLD_VERSION" "2.6.0"; then
+    echo "--- VERSION COLUMNS TEST: Seeding old cluster with data"
+    sqllogictest -d dev -h localhost -p 4566 "$TEST_DIR/version-columns/seed.slt"
+
+    echo "--- VERSION COLUMNS TEST: Validating old cluster"
+    sqllogictest -d dev -h localhost -p 4566 "$TEST_DIR/version-columns/validate_original.slt"
+  fi
+
```
yuhao-su, line 255: I also added some backwards compact test. PTAL if I'm doing this right @yezizp2012

yezizp2012: lgtm

src/sqlparser/src/parser.rs
BugenZhao: An empty clause appears to be meaningless & confusing. I would suggest rejecting it.

yuhao-su: hmm empty  `()` is actually already rejected because parse will find a `)` when trying to parse a ident
https://github.com/risingwavelabs/risingwave/pull/22931/files#diff-7c05cba855f28444eb73cf5df311cb5273582962d977950f10878cb28c85b86dR134-R138

BugenZhao: Great

BugenZhao (approved):
LGTM

PR #22932, opened 2025-08-20
refactor: migrate actor and fragment metadata from db to in-memory cache

shanicky:

**Summary**

This PR refactors the actor and fragment models to clearly separate persistent database entities from runtime state, introduces per-fragment parallelism, and simplifies the scaling and rescheduling APIs. Legacy reschedule logic is removed in favor of policy-driven control through the DDL service, and backup/restore is aligned with the new model boundaries. These changes enhance maintainability and scalability, laying groundwork for more flexible resource management.

- Rewrite the entire reschedule process, no longer based on the original actor-vnode rebalance model. Instead, use rerender_jobs/fragments to completely rerender the whole graph.  
- During the online scaling phase, compare the newly rendered graph with the actor info inside the barrier system, generate the diff commands, and push them through the barrier system.  
- Deprecate offline scaling. During recovery, rerender using the provided database_id or rerender all jobs; no longer need separate scale_actors.  
- Native support for database isolation.  
- After rescheduling, post-apply is no longer needed, but the serving mapping still requires a post notification.  
- Rewrite the shared source reschedule process: no manual alignment of no shuffle/splits is needed. When rerendering, realign and reassign shared source splits directly based on the ensemble.  
- Actor split management is now handled in the fragment_split table and splits are reassigned as needed.  
- Actor IDs are reassigned starting from 0 every time meta restarts.

**Details**

- Actor and Fragment Model Refactoring:
  - Introduces distinct runtime (`ActorModel`) and persistent (`Model`) representations for actors, ensuring only persistent fields are stored in the database while runtime state is managed separately.
  - Fragment model gains a JSONB `parallelism` field, enabling fragment-level parallelism configuration. ORM relations between fragments and actors are removed, decoupling the DB schema from runtime topology.

- Reschedule and Scaling API Changes:
  - The legacy ScaleService reschedule API is disabled; scaling logic now flows through explicit policies (`ReschedulePolicy`, `ParallelismPolicy`, `ResourceGroupPolicy`) in the DDL service, supporting fixed/adaptive parallelism and stricter validation.
  - CDC backfill and streaming job scaling require explicit parallelism settings, improving error handling and predictability.

- Meta-model and Backup Adjustments:
  - Unused dependencies (`num-integer`, `num-traits`) are removed to streamline the codebase.
  - Actor records are no longer persisted/restored; backup/restore now focuses on streaming job, fragment, and relation metadata, reflecting runtime/persistence separation.

- Barrier Command Simplifications:
  - The `post_updates` field is removed from reschedule barrier commands, signaling a shift of metadata update responsibilities away from barrier execution toward service/controller logic.

**Architecture impact**
These changes enforce a separation of concerns between runtime state and persistence, making the codebase more maintainable and less error-prone. Decoupling ORM models from runtime logic and fragment-actor relations enables more granular scaling and resource allocation, while disabling legacy APIs clarifies the reschedule pipeline and reduces accidental complexity.

**Technical highlights**
- Runtime and persistent actor data are now strictly separated.
- Per-fragment parallelism is supported, enabling flexible scaling strategies.
- Policy structs encapsulate scaling semantics, improving API robustness and future extensibility.
- Backup/restore logic aligns with the new model boundaries, ensuring consistent metadata recovery.

**Other pull requests**

- https://github.com/risingwavelabs/risingwave/pull/23391 Used to build the mapping during rendering.
- https://github.com/risingwavelabs/risingwave/pull/23395 Used to adapt adaptive splits during rendering. 
- https://github.com/risingwavelabs/risingwave/pull/23396 Change 3 methods to use shared actor info.
- https://github.com/risingwavelabs/risingwave/pull/23404 Change the Fragment/Actor dependencies in `stream_graph` to `SharedFragmentInfo`.
- https://github.com/risingwavelabs/risingwave/pull/23445 Change some methods to use shared actor info in fragment.rs
- https://github.com/risingwavelabs/risingwave/pull/23447
- https://github.com/risingwavelabs/risingwave/pull/23467
- https://github.com/risingwavelabs/risingwave/pull/23474
- https://github.com/risingwavelabs/risingwave/pull/23469
- https://github.com/risingwavelabs/risingwave/pull/23521
- https://github.com/risingwavelabs/risingwave/pull/23528

## Checklist

### DIFF
--- a/src/meta/src/stream/scale.rs
+++ b/src/meta/src/stream/scale.rs
@@ -334,7 +334,6 @@ impl ScaleController {
             streaming_job.update(&txn).await?;
         }
 
-        // update
         let jobs = policy.keys().copied().collect();
 
         let workers = workers
@@ -522,19 +521,6 @@ impl ScaleController {
         if render_result.is_empty() {
             return Ok(HashMap::new());
         }
-
-        // for (db, jobs) in &render_result {
-        //     println!("\tdb: {db}");
-        //     for (job, fragments) in jobs {
-        //         println!("\t\tjob: {job}");
-        //         for (fragment, fragment_info) in fragments {
-        //             println!("\t\t\tfragment: {fragment:?}");
-        //             for (actor, actor_info) in &fragment_info.actors {
-        //                 println!("\t\t\t\tactor: {actor:?}, {actor_info:?}");
-        //             }
-        //         }
-        //     }
-        // }
 
         let job_ids = render_result
             .values()


src/meta/src/controller/catalog/get_op.rs
```diff
@@ -515,6 +522,22 @@ impl CatalogController {
         Ok(job_id)
     }
 
+    pub async fn list_streaming_job_with_database(
```
shanicky, line 535: for parallelism control

src/meta/src/controller/catalog/get_op.rs
```diff
@@ -589,4 +612,146 @@ impl CatalogController {
             .ok_or_else(|| MetaError::catalog_id_not_found("streaming job", streaming_job_id))?;
         Ok(status)
     }
+
+    pub async fn get_streaming_job_runtime_info(
```
shanicky: for recovery

src/meta/src/controller/catalog/get_op.rs
```diff
@@ -589,4 +612,146 @@ impl CatalogController {
...
+                (
+                    job_id,
+                    (timezone, definitions.remove(&job_id).unwrap_or_default()),
+                )
+            })
+            .collect();
+        Ok(result)
+    }
+
+    pub async fn restore_background_streaming_job(
```
shanicky: for recovery

src/meta/src/controller/fragment.rs
```diff
@@ -566,13 +567,86 @@ impl CatalogController {
         Ok(select.into_tuple().all(&inner.db).await?)
     }
 
+    pub async fn get_fragment_ids_by_job_ids(
```
shanicky: useless

src/meta/src/controller/fragment.rs
```diff
@@ -566,13 +567,86 @@ impl CatalogController {
...
+            .column(fragment::Column::FragmentId)
+            .filter(fragment::Column::JobId.is_in(job_id.to_vec()))
+            .into_tuple()
+            .all(txn)
+            .await?;
+
+        Ok(fragment_ids)
+    }
+
+    pub async fn get_job_sink_fragments_by_id(
```
shanicky: remove

src/meta/src/controller/fragment.rs
```diff
@@ -566,13 +567,86 @@ impl CatalogController {
...
+            .into_iter()
+            .filter(|fragment| {
+                FragmentTypeMask::from(fragment.fragment_type_mask).contains(FragmentTypeFlag::Sink)
+            })
+            .collect();
+
+        Ok(fragments)
+    }
+
+    pub async fn get_job_info_by_id(&self, job_id: ObjectId) -> MetaResult<streaming_job::Model> {
```
shanicky: remove

src/meta/src/controller/utils.rs
```diff
@@ -1236,118 +1234,6 @@ pub async fn insert_fragment_relations(
     Ok(())
 }
 
-pub async fn get_fragment_actor_dispatchers<C>(
```
shanicky, line 1231: for old scale

src/meta/src/controller/utils.rs
```diff
@@ -1517,42 +1403,6 @@ pub fn resolve_no_shuffle_actor_dispatcher(
     }
 }
 
-/// `get_fragment_mappings` returns the fragment vnode mappings of the given job.
-pub async fn get_fragment_mappings<C>(
```
shanicky: remove

src/meta/src/controller/utils.rs
```diff
@@ -1671,32 +1521,14 @@ pub fn rebuild_fragment_mapping_from_actors(
     result
```
shanicky, line 1515: remove

src/meta/src/controller/streaming_job.rs
```diff
@@ -1637,6 +1600,25 @@ impl CatalogController {
         Ok(fragment_actors)
     }
 
+    fn get_fragment_actors_from_running_info(
```
shanicky, line 1607: for later use

src/meta/src/manager/metadata.rs
```diff
@@ -113,28 +107,6 @@ impl ActiveStreamingWorkerNodes {
         &self.worker_nodes
     }
 
-    pub(crate) async fn wait_changed(
```
shanicky, line 115: for old migrate_actors

src/meta/src/manager/metadata.rs
```diff
@@ -586,34 +544,6 @@ impl MetadataManager {
             .collect())
     }
 
-    pub async fn get_job_fragments_by_ids(
```
shanicky: remove

src/meta/src/manager/metadata.rs
```diff
@@ -586,34 +544,6 @@ impl MetadataManager {
...
-                self.catalog_controller
-                    .get_job_fragments_by_id(id.table_id as _)
-                    .await?,
-            );
-        }
-        Ok(table_fragments)
-    }
-
-    pub async fn all_active_actors(&self) -> MetaResult<HashMap<ActorId, StreamActor>> {
-        let table_fragments = self.catalog_controller.table_fragments().await?;
```
shanicky, line 589: old load_all_actors

src/meta/src/controller/id.rs
```diff
@@ -55,16 +55,7 @@ impl<const TYPE: IdCategoryType> IdGenerator<TYPE> {
...
-                .select_only()
-                .expr(Func::if_null(
-                    Expr::col(actor::Column::ActorId).max().add(1),
-                    1,
-                ))
-                .into_tuple()
-                .one(conn)
-                .await?
-                .unwrap_or_default(),
+            IdCategory::Actor => 0,
```
wenym1: If actor id is always in memory, we may directly use an `Arc<AtomicU32>` in MetaEnv.

src/meta/model/src/actor.rs
```diff
@@ -60,6 +60,16 @@ pub struct Model {
     #[deprecated]
```
wenym1: Do we still use the `actor::Model` anywhere?

src/meta/service/src/scale_service.rs
```diff
@@ -123,72 +115,10 @@ impl ScaleService for ScaleServiceImpl {
 
     async fn reschedule(
         &self,
-        request: Request<RescheduleRequest>,
+        _request: Request<RescheduleRequest>,
     ) -> Result<Response<RescheduleResponse>, Status> {
-        self.barrier_manager.check_status_running()?;
```
wenym1, line 128: Do we support specifying fragment level reschedule?

src/meta/src/stream/scale.rs
```diff
@@ -382,1221 +122,185 @@ impl ScaleController {
...
+        Ok(reschedule)
     }
 
     #[await_tree::instrument]
     pub async fn post_apply_reschedule(
         &self,
         reschedules: &HashMap<FragmentId, Reschedule>,
-        post_updates: &JobReschedulePostUpdates,
     ) -> MetaResult<()> {
-        // Update fragment info after rescheduling in meta store.
```
wenym1, line 1595: Do we still need this method? Have we already notified the latest vnode mapping when we apply changes on the `InflightFragmentInfo`?

src/meta/src/controller/streaming_job.rs
```diff
@@ -784,39 +773,12 @@ impl CatalogController {
...
         job_id: ObjectId,
-        actor_ids: Vec<crate::model::ActorId>,
         upstream_fragment_new_downstreams: &FragmentDownstreamRelation,
-        split_assignment: &SplitAssignment,
         new_sink_downstream: Option<FragmentDownstreamRelation>,
     ) -> MetaResult<()> {
         let inner = self.inner.write().await;
         let txn = inner.db.begin().await?;
 
-        let actor_ids = actor_ids.into_iter().map(|id| id as ActorId).collect_vec();
```
wenym1, line 795: Where do we update the FragmentSplits table for caller on this method?

src/meta/src/controller/streaming_job.rs
```diff
@@ -456,7 +452,7 @@ impl CatalogController {
         ensure_job_not_canceled(job_id, &txn).await?;
 
         // Add fragments.
-        let (fragments, actors): (Vec<_>, Vec<_>) = fragment_actors.into_iter().unzip();
+        let (fragments, _): (Vec<_>, Vec<_>) = fragment_actors.into_iter().unzip();
```
wenym1: Since the original `actors` is not used, we may also update `extract_fragment_and_actors_from_fragments` to only return `fragments`, and then we may not need to pass `actor_status` in current method.

src/meta/src/controller/streaming_job.rs
```diff
@@ -1637,6 +1600,25 @@ impl CatalogController {
         Ok(fragment_actors)
     }
 
+    fn get_fragment_actors_from_running_info(
+        &self,
+        fragment_ids: &HashSet<u32>,
```
wenym1:
```suggestion
        fragment_ids: impl Iterator<Item = u32>,
```

src/meta/src/barrier/context/recovery.rs
```diff
@@ -83,26 +78,37 @@ impl GlobalBarrierWorkerContextImpl {
         Ok(())
     }
 
-    async fn list_background_job_progress(&self) -> MetaResult<Vec<(String, StreamJobFragments)>> {
+    // async fn list_background_job_progress(&self) -> MetaResult<Vec<(String, StreamJobFragments)>> {
```
wenym1: Remove the deadcode?

src/meta/src/barrier/context/recovery.rs
```diff
@@ -468,6 +437,42 @@ impl GlobalBarrierWorkerContextImpl {
...
+                                )
+                            })
+                            .collect();
+                        let jobs = self
+                            .metadata_manager
+                            .catalog_controller
+                            .restore_background_streaming_job(job_infos)
+                            .await?;
+
+                        for (job_id, info) in jobs {
```
wenym1: Previously we have logic to call `finish_streaming_job` for jobs with no backfill actors to track. Do we need to preserve this logic?

src/meta/src/barrier/context/recovery.rs
```diff
@@ -745,451 +766,50 @@ impl GlobalBarrierWorkerContextImpl {
...
-        let all_inuse_worker_slots: HashSet<_> = mgr
-            .catalog_controller
-            .all_inuse_worker_slots()
-            .await?
-            .into_iter()
-            .collect();
-
-        let active_worker_slots: HashSet<_> = active_nodes
-            .current()
+        all_info: HashMap<DatabaseId, HashMap<TableId, InflightStreamingJobInfo>>,
```
wenym1: Seems no need to pass an owned info. Reference is enough, and then the clone on the caller is unnecessary.

src/meta/src/barrier/context/recovery.rs
```diff
@@ -695,10 +713,12 @@ impl GlobalBarrierWorkerContextImpl {
             )
             .await?;
 
-        // update and build all actors.
-        let stream_actors = self.load_all_actors().await.inspect_err(|err| {
-            warn!(error = %err.as_report(), "update actors failed");
-        })?;
+        // // update and build all actors.
```
wenym1: Remove the deadcode?

src/meta/src/barrier/context/recovery.rs
```diff
@@ -615,23 +596,60 @@ impl GlobalBarrierWorkerContextImpl {
...
+        let mut info = all_info
+            .get(&database_id)
+            .cloned()
+            .map_or_else(HashMap::new, |table_map| {
+                HashMap::from([(database_id, table_map)])
+            });
+
         self.recovery_table_with_upstream_sinks(&mut info).await?;
 
-        assert!(info.len() <= 1);
```
wenym1, line 618: Any reason to remove the assertion?

src/meta/src/barrier/context/recovery.rs
```diff
@@ -745,451 +766,50 @@ impl GlobalBarrierWorkerContextImpl {
...
-                            resolve_no_shuffle_upstream: true,
-                            skip_create_new_actors: true,
+        let mut stream_actors = HashMap::new();
+
+        for (_, streaming_info) in all_info.values().flatten() {
+            let StreamingJobExtraInfo {
+                timezone,
+                job_definition,
+            } = job_extra_info
+                .get(&(streaming_info.job_id.table_id as i32))
```
wenym1: Should we return with error if the job_id in not in the map?

src/meta/src/barrier/context/recovery.rs
```diff
@@ -615,23 +596,60 @@ impl GlobalBarrierWorkerContextImpl {
             .complete_dropped_tables(dropped_table_ids.into_iter().map(|id| id.table_id as _))
             .await;
 
-        let mut info = self
-            .resolve_graph_info(Some(database_id))
+        let active_streaming_nodes =
```
wenym1, line 598: In previous design, the `ActiveStreamingWorkerNodes` should only be reloaded in global recovery. Can we possibly still follow this design?

wenym1: In `resolve_graph_info`, the newly passed `worker_nodes` is used in `load_all_actors_dynamic`. IIUC, the call chain `load_all_actors_dynamic` -> `load_fragment_info` -> `render_jobs` -> `render_no_shuffle_ensembles`, can be split into the following several steps
1. load `adaptive_parallelism_strategy` from system parameter (async)
2. load necessary information from meta store table (async)
3. render actor with `adaptive_parallelism_strategy`, `worker_nodes` and the information loaded from meta store table (non-async)

If so, can we possibly only reload and return the necessary information loaded from meta store table in step 2 in RuntimeInfoSnapshot? We can do the work of step 3 purely in memory in non-async context in global barrier worker after `reload_runtime_info`. For `adaptive_parallelism_strategy`, it can be stored as a field of global barrier worker, so that we don't have to access in async context. If the system param is mutable,  we can update it with the `SystemParamsChange` already available in the worker loop select macro of global barrier worker.

src/meta/src/controller/catalog/get_op.rs
wenym1: After https://github.com/risingwavelabs/risingwave/pull/23545, this method is not needed.

src/meta/model/src/actor.rs
shanicky: No longer needed. Do we need to completely remove the actor table directly through a migration in this PR? cc @yezizp2012

src/meta/service/src/scale_service.rs
shanicky: The low-level reschedule interface has been deprecated. This is just for compatibility, and almost no one uses this interface. We will replace it later with alter fragment parallelism.

src/meta/src/stream/scale.rs
shanicky: According to the discussion, the notification for serving mapping here can be removed, so scale no longer needs post apply. cc @zwang28

src/meta/src/controller/streaming_job.rs
shanicky: https://github.com/risingwavelabs/risingwave/blob/d9857a01305f08edadd474819925dc5ecba3fd60/src/meta/src/barrier/context/context_impl.rs#L421

All the changes for the source manager are placed outside. Do you want me to move them inside this?

src/meta/model/src/actor.rs
yezizp2012: Yes, I think we can remove it together in this PR.

src/meta/src/barrier/context/recovery.rs
shanicky: In theory, it’s feasible, but it might be fairly complex and could involve handling multiple tables. If the overall logic of this PR is fine, we can make further adjustments on top of it. The core of render is essentially to organize the no-shuffle ensemble, then assign vnodes, and finally handle the source split and cdc split. If this information isn’t too large, we can give it a try.

shanicky: > In previous design, the ActiveStreamingWorkerNodes should only be reloaded in global recovery. Can we possibly still follow this design?

The main point is that render jobs need workers to determine the scheduling unit. Is there any other way to obtain it?

wenym1: The current implementation is like
```
async fn reload(_database)_runtime_info(&self) -> (..., rendered actors, ...) {
    ...
    render_jobs(&worker_map).await
    ...
}

async fn render_jobs(worker_map) -> rendered actors {
    // 1. reload something without using worker_map, with some async function
    render_no_shuffle_ensembles(worker_map)
}

async fn render_no_shuffle_ensembles(worker_map) -> rendered actors {
    // 2. reload something without using worker_map, with some async function
    ...
    for NoShuffleEnsemble {
        entries,
        components,
    } in ensembles
    {
        ... 3. render actors with worker_map
    }
    rendered actors
}
```

My suggested implementation is like
```
async fn reload(_database)_runtime_info(&self) -> (..., reloaded_info 1+2, ...) {
    ...
    render_jobs().await
    ...
}

async fn render_jobs() -> reloaded_info 1+2 {
    // 1. reload something without using worker_map, with some async function
    let reloaded_info 2 = render_no_shuffle_ensembles();
    reloaded_info 1+2
}

async fn render_no_shuffle_ensembles() -> reloaded info 2 {
    // 2. reload something without using worker_map, with some async function
    ...
}

fn render_actors(worker_map, reloaded info 1+2, ..) {
    for NoShuffleEnsemble {
        entries,
        components,
    } in ensembles
    {
        ... render actors with worker_map
    }
}
```

Most logic is not changed, but only moved and split some methods.

For global recovery, we get the reloaded information and render actor both in `recover_inner`, and the worker_map is already available there. 

For database recovery, things are split into two steps. We get the reloaded information from `reload_database_runtime_info`, which is async, and then call `render_actors` in `entering_initializing.enter()`, which is in non-async context and can use the `worker_map` and `adaptive_parallelism_strategy ` of global barrier worker.

src/meta/src/controller/id.rs
shanicky: refactored

src/meta/src/stream/scale.rs
shanicky: removed

src/meta/src/barrier/context/recovery.rs
shanicky: I’ve already implemented a `render_actors`, Is that the right approach?

src/meta/model/src/actor.rs
```diff
@@ -48,37 +46,13 @@ impl From<ActorStatus> for PbActorState {
...
 
-#[derive(Clone, Debug, PartialEq, DeriveEntityModel, Eq, Serialize, Deserialize)]
-#[sea_orm(table_name = "actor")]
-pub struct Model {
-    #[sea_orm(primary_key)]
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct ActorModel {
     pub actor_id: ActorId,
     pub fragment_id: FragmentId,
     pub status: ActorStatus,
```
wenym1, line 53: Do we still use `ActorStatus` anywhere?

src/meta/src/barrier/context/mod.rs
```diff
@@ -108,7 +108,7 @@ pub(super) struct GlobalBarrierWorkerContextImpl {
 
     source_manager: SourceManagerRef,
 
-    scale_controller: ScaleControllerRef,
+    _scale_controller: ScaleControllerRef,
```
wenym1, line 111: I think we can remove the field, since the scale logic of global barrier worker should no longer depend on this scale controller.

src/meta/src/controller/fragment.rs
```diff
@@ -786,12 +860,51 @@ impl CatalogController {
...
         for job in jobs {
-            let fragment_actors = FragmentModel::find()
-                .find_with_related(Actor)
+            let fragments = FragmentModel::find()
                 .filter(fragment::Column::JobId.eq(job.job_id))
                 .all(&inner.db)
                 .await?;
 
+            let fragment_actors = {
+                let guard = self.env.shared_actor_infos().read_guard();
```
wenym1: Seems that in `get_job_fragments_by_id` there is similar logic to read from shared actor info to compose the ActorModel of a fragment. Can we extract the logic to a common method of shared actor info?

src/meta/src/controller/scale.rs
```diff
@@ -270,394 +106,1402 @@ where
...
+        .collect();
+
+    let context = RenderActorsContext {
+        fragment_source_ids: &fragment_source_ids,
+        fragment_splits: &fragment_splits,
+        streaming_job_databases: &streaming_job_databases,
+        database_map: &database_map,
+    };
+
+    render_actors(
```
wenym1, line 420: The methods are split as what I meant. But I also meant that here in `render_no_shuffle_ensembles` we should return things in `RenderActorsContext` and don't call `render_actors`. In `render_jobs` and the outer `reload_runtime_info_impl` we also don't call `render_actors`, but only reload these necessary information. 

We only call `render_jobs` in global and database recovery after the `reload_(database_)runtime_info`.

src/meta/src/controller/streaming_job.rs
```diff
@@ -1618,10 +1568,11 @@ impl CatalogController {
             .update(&txn)
             .await?;
         }
-        let fragment_actors = get_fragment_actor_ids(&txn, fragment_ids).await?;
 
         txn.commit().await?;
 
+        let fragment_actors = self.get_fragment_actors_from_running_info(fragment_ids.into_iter());
```
wenym1, line 1590: It seems that we have to introduce and call `get_fragment_actors_from_running_info` in these methods like `update_source_rate_limit_by_source_id` because they return `HashMap<FragmentId, Vec<ActorId>>`. The returned value is only used in `apply_throttle` to compose the `Command::Throttle`. 

If so, we can actually only include fragment ids in `Command::Throttle`, and in global barrier worker when we handle the command, we then directly use inflight actor info maintained in the worker.

Furthermore, we have to include actor id in the throttle mutation because in streaming executor, we always check throttle config update by actor id. If so, we can actually change check by fragment id, and then in throttle mutation we only need to include the fragment id, and then we don't have to resolve the actor ids of fragments. I believe in many logic of altering streaming job config, (e.g. logic that use `Throttle` and `ConnectorPropsChange` mutation), we can  change from check by actor id to check by fragment id to avoid having to resolve actor ids of fragment by shared actor info.

src/meta/src/stream/stream_graph/id.rs
```diff
@@ -66,17 +68,78 @@ impl<const TYPE: IdCategoryType> GlobalIdGen<TYPE> {
...
     }
+}
+
+pub(super) type GlobalFragmentId = GlobalId<{ IdCategory::Fragment }>;
+pub(super) type GlobalFragmentIdGen = GlobalIdGen<{ IdCategory::Fragment }>;
+
+pub(super) type GlobalTableIdGen = GlobalIdGen<{ IdCategory::Table }>;
+
+#[derive(Clone, Copy, Debug, Hash, Eq, PartialEq, PartialOrd, Ord)]
+pub(super) struct GlobalActorId(u32);
```
wenym1, line 59: We can actually share the `GlobalIdGen` with other global id types.
```
pub struct GlobalIdGen<ID: From<u32>> {
    offset: u32,
    len: u32,
    _phantom: PhantomData<ID>,
}

type GlobalActorIdGen = GlobalIdGen<GlobalActorId>;
type GlobalFragmentIdGen = GlobalIdGen<GlobalId<{ IdCategory::Fragment }>>;
...

impl<const TYPE: IdCategoryType> GlobalIdGen<GlobalId<TYPE>> {
    /// Pre-allocate a range of IDs with the given `len` and return the generator.
    pub fn new(id_gen: &SqlIdGeneratorManager, len: u64) -> Self {
        let offset = id_gen.generate_interval::<TYPE>(len);
        Self {
            offset: offset as u32,
            len: len as u32,
        }
    }
}

impl GlobalActorIdGen {
    pub fn new(counter: &AtomicU32, len: u64) -> Self {
        let len_u32 = u32::try_from(len).expect("actor count exceeds u32::MAX");
        let offset = counter.fetch_add(len_u32, Ordering::Relaxed);
        Self {
            offset,
            len: len_u32,
        }
    }
}

impl<ID: From<u32>> GlobalIdGen<ID> {
    /// Convert local id to global id. Panics if `id >= len`.
    pub fn to_global_id(self, local_id: u32) -> ID {
        assert!(
            local_id < self.len,
            "id {} is out of range (len: {})",
            local_id,
            self.len
        );
        ID::from(local_id + self.offset)
    }

    pub fn len(&self) -> u32 {
        self.len
    }
}
```

src/meta/src/controller/streaming_job.rs
wenym1: I think it's better to apply the changes on fragment splits here together in a single txn.

src/meta/src/barrier/context/recovery.rs
wenym1: Commented in https://github.com/risingwavelabs/risingwave/pull/22932#discussion_r2468059937

src/meta/model/src/actor.rs
shanicky: Yes, we need to maintain compatibility with the fields in protobuf; removing them in this PR would make the PR too large.

src/meta/src/barrier/context/mod.rs
shanicky: I’m trying to introduce the information for render_actors; I’m not sure whether this reference is still needed.

src/meta/src/controller/scale.rs
shanicky: I understand the idea, but both online scaling and alter parallelism depend on this, so the changes might be quite significant.

src/meta/src/controller/streaming_job.rs
shanicky: This is just a replacement of the previous logic. We can open a tracking issue and refactor it in a later PR.

shanicky: https://github.com/risingwavelabs/risingwave/issues/23599

src/meta/src/controller/fragment.rs
shanicky: refactored

src/meta/src/stream/stream_graph/id.rs
shanicky: I didn't really get it here.

src/meta/src/barrier/context/recovery.rs
shanicky: Is there any issue if we put ActiveStreamingWorkerNodes here?

src/meta/src/manager/metadata.rs
```diff
@@ -685,8 +685,10 @@ impl MetadataManager {
             })
             .collect();
 
+        let inner = self.catalog_controller.inner.write().await;
```
wenym1, line 688: It seems that this `update_fragment_splits` is not used any more.

shanicky: This helper function still needs to be used during `SourceChangeSplit` and `Reschedule`.

src/meta/model/src/actor.rs
```diff
@@ -48,37 +46,13 @@ impl From<ActorStatus> for PbActorState {
     }
 }
 
-#[derive(Clone, Debug, PartialEq, DeriveEntityModel, Eq, Serialize, Deserialize)]
-#[sea_orm(table_name = "actor")]
-pub struct Model {
-    #[sea_orm(primary_key)]
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct ActorModel {
```
yezizp2012, line 50: Can we move this struct to another place? Since it's no longer a model in the meta store.

src/meta/src/controller/id.rs
```diff
@@ -55,17 +54,7 @@ impl<const TYPE: IdCategoryType> IdGenerator<TYPE> {
...
-                .expr(Func::if_null(
-                    Expr::col(actor::Column::ActorId).max().add(1),
-                    1,
-                ))
-                .into_tuple()
-                .one(conn)
-                .await?
-                .unwrap_or_default(),
-            _ => unreachable!("IdGeneratorV2 only supports Table, Fragment, and Actor"),
+            _ => unreachable!("IdGeneratorV2 only supports Table and Fragment"),
```
yezizp2012:
```suggestion
            _ => unreachable!("IdGenerator only supports Table and Fragment"),
```
Nits.

src/meta/src/controller/utils.rs
```diff
@@ -1550,23 +1438,81 @@ pub fn rebuild_fragment_mapping(fragment: &SharedFragmentInfo) -> PbFragmentWork
...
+        ActorStatus,
+    )>,
+) -> Vec<FragmentWorkerSlotMapping> {
+    let mut all_actor_locations = HashMap::new();
+    let mut actor_bitmaps = HashMap::new();
+    let mut fragment_actors = HashMap::new();
+    let mut fragment_dist = HashMap::new();
+
+    for (fragment_id, dist, actor_id, bitmap, worker_id, actor_status) in job_actors {
+        if actor_status == ActorStatus::Inactive {
```
yezizp2012, line 1457: IIUC there's no such concept of `ActorStatus` after this PR, can we remove this field directly?

src/meta/src/stream/scale.rs
```diff
@@ -382,1985 +117,741 @@ impl ScaleController {
...
-            assert!(
-                !new_actor_ids.is_empty(),
-                "should be at least one actor in fragment {} after rescheduling",
-                fragment_id
-            );
+        let mut added_actors = HashMap::new();
+        for &actor_id in &added_actor_ids {
+            let InflightActorInfo { worker_id, .. } = curr_actors
+                .get(&actor_id)
+                .ok_or_else(|| anyhow!("BUG: Worker not found for new actor {}", actor_id))?;
```
yezizp2012, line 179: 😃

src/meta/src/stream/scale.rs
```diff
@@ -382,1985 +117,741 @@ impl ScaleController {
...
-                bail!("Cannot resize streaming_job {job_id} to empty worker set")
-            }
+    pub async fn reschedule_inplace(
+        &self,
+        policy: HashMap<ObjectId, ReschedulePolicy>,
+        workers: HashMap<WorkerId, PbWorkerNode>,
+    ) -> MetaResult<HashMap<DatabaseId, Command>> {
+        // jobs: HashMap<ObjectId, TargetResourcePolicy>,
+        // workers: BTreeMap<WorkerId, Worker>,
+        // ) -> MetaResult<HashMap<DatabaseId, HashMap<TableId, HashMap<FragmentId, crate::controller::fragment::InflightFragmentInfo >>>>
```
yezizp2012:
```suggestion
```

src/meta/src/stream/scale.rs
```diff
@@ -382,1985 +117,741 @@ impl ScaleController {
...
+
+        let inner = self.metadata_manager.catalog_controller.inner.read().await;
+        let txn = inner.db.begin().await?;
+
+        for (table_id, target) in &policy {
+            let streaming_job = StreamingJob::find_by_id(*table_id)
```
yezizp2012, line 303: Can we get all jobs and `max_parallelism` column only in one query for better performance?

src/meta/src/stream/stream_graph/id.rs
yezizp2012: We can do it as a refactoring in another PR, since this PR is already complex enough.

src/meta/src/stream/stream_manager.rs
```diff
@@ -723,10 +722,11 @@ impl GlobalStreamManager {
     pub(crate) async fn reschedule_streaming_job(
         &self,
         job_id: u32,
-        target: JobRescheduleTarget,
+        target: ReschedulePolicy,
```
yezizp2012:
```suggestion
        policy: ReschedulePolicy,
```
Nits.

src/meta/src/controller/scale.rs
```diff
@@ -270,394 +106,1402 @@ where
...
+
+    let available_workers: BTreeMap<_, _> = worker_nodes
+        .current()
+        .values()
+        .filter(|worker| worker.is_streaming_schedulable())
+        .map(|worker| {
+            (
+                worker.id as i32,
+                WorkerInfo {
+                    weight: NonZeroUsize::new(worker.compute_node_parallelism()).unwrap(),
```
yezizp2012:
```suggestion
                    parallelism: NonZeroUsize::new(worker.compute_node_parallelism()).unwrap(),
```
For better understanding.

src/meta/src/controller/utils.rs
shanicky: https://github.com/risingwavelabs/risingwave/issues/23604

PR #22942, opened 2025-08-20
feat: support vacuum full iceberg table

chenzl25:

> FULL
Selects “full” vacuum, which can reclaim more space, but takes much longer and exclusively locks the table. This method also requires extra disk space, since it writes a new copy of the table and doesn't release the old copy until the operation is complete. Usually this should only be used when a significant amount of space needs to be reclaimed from within the table.

- Based on [PG's doc](https://www.postgresql.org/docs/current/sql-vacuum.html), when we apply vacuum to an iceberg table, it should be related to expire snapshots. And the `vacuum full` command should be related to compact an iceberg table and expire snapshots. So in this PR, we change the semantics of the [vacuum introduced previously](https://github.com/risingwavelabs/risingwave/pull/22571) (which was compacting an iceberg table).

- vacuum = expire snapshot
- vacuum full = compact + expire snapshot

## Checklist

## Documentation

<details>

</details>


Li0k:
https://github.com/risingwavelabs/risingwave/pull/22947

Your thinking is very clear. However, the default configuration for "expired" is quite conservative. I hope that vacuum can accept parameters, can this be done?

chenzl25:
> e. I hope that vacuum can accept parameters, can this be done?

The parameter is still controlled by the table property.

Li0k (approved):
LGTM

PR #22943, opened 2025-08-20
feat(cdc): Add progress tracking for parallelized cdc backfill

zwang28:

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/optimizer/plan_node/stream_cdc_table_scan.rs
+++ b/src/frontend/src/optimizer/plan_node/stream_cdc_table_scan.rs
@@ -78,6 +78,8 @@ impl StreamCdcTableScan {
             catalog_builder.add_column(&Field::with_name(DataType::Boolean, "backfill_finished"));
             // `row_count` column, the number of rows read from snapshot
             catalog_builder.add_column(&Field::with_name(DataType::Int64, "row_count"));
+            catalog_builder.add_column(&Field::with_name(DataType::Jsonb, "cdc_offset_low"));
+            catalog_builder.add_column(&Field::with_name(DataType::Jsonb, "cdc_offset_high"));
             catalog_builder
                 .build(vec![], 1)
                 .with_id(state.gen_table_id_wrapped())
--- a/src/meta/src/barrier/cdc_progress.rs
+++ b/src/meta/src/barrier/cdc_progress.rs
@@ -1,0 +1,226 @@
+// Copyright 2025 RisingWave Labs
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::collections::HashMap;
+use std::sync::Arc;
+
+use num_traits::ToPrimitive;
+use parking_lot::Mutex;
+use risingwave_common::catalog::{FragmentTypeFlag, TableId};
+use risingwave_meta_model::{FragmentId, ObjectId, cdc_table_snapshot_split, fragment};
+use risingwave_pb::stream_service::PbBarrierCompleteResponse;
+use risingwave_pb::stream_service::barrier_complete_response::PbCdcTableBackfillProgress;
+use sea_orm::prelude::Expr;
+use sea_orm::{ColumnTrait, EntityTrait, QueryFilter, QuerySelect};
+
+use crate::MetaResult;
+use crate::controller::SqlMetaStore;
+
+pub type CdcTableBackfillTrackerRef = Arc<CdcTableBackfillTracker>;
+
+pub struct CdcTableBackfillTracker {
+    meta_store: SqlMetaStore,
+    inner: Mutex<CdcTableBackfillTrackerInner>,
+}
+
+impl CdcTableBackfillTracker {
+    pub async fn new(meta_store: SqlMetaStore) -> MetaResult<Self> {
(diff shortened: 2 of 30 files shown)


yuhao-su:
Can we use the same method to update non-parallelized cdc in the future?

zwang28:
>  Can we use the same method to update non-parallelized cdc in the future?

The method is resuable. But for non-parallelized backfill we don't know the denominator (e.g. total rows) to measure the progress. The issue is not hard to resolve though.
In parallelized we use total number of backfill splits as denominator of progress.

zwang28:
> This PR also includes a refactor that uses the CDC offset to filter out unnecessary data

Consider reverting this change until the bugfix is merged. https://github.com/risingwavelabs/risingwave/pull/22503

hzxa21:
> > This PR also includes a refactor that uses the CDC offset to filter out unnecessary data
> 
> Consider reverting this change until the bugfix is merged. #22503

Should we merge https://github.com/risingwavelabs/risingwave/pull/22503 before this PR then?

src/frontend/src/optimizer/plan_node/stream_cdc_table_scan.rs
```diff
@@ -78,6 +78,8 @@ impl StreamCdcTableScan {
             catalog_builder.add_column(&Field::with_name(DataType::Boolean, "backfill_finished"));
             // `row_count` column, the number of rows read from snapshot
             catalog_builder.add_column(&Field::with_name(DataType::Int64, "row_count"));
+            catalog_builder.add_column(&Field::with_name(DataType::Jsonb, "cdc_offset_low"));
```
hzxa21, line 86: nits: modify the comment in L67 accordingly

src/meta/model/migration/src/lib.rs
```diff
@@ -48,6 +48,7 @@ mod m20250710_065220_sink_auto_refresh_schema;
 mod m20250722_155040_table_refreshable;
 mod m20250729_174630_add_vector_index_info;
 mod m20250810_000000_add_user_admin_field;
+mod m20250821_081110_cdc_table_snapshot_splits_add_column;
```
hzxa21, line 52: Do we need a new migration given that [cdc_table_snapshot_split](https://github.com/risingwavelabs/risingwave/blob/390ca65c1987f876c8b815681877f2d0a6fabf93/src/meta/model/migration/src/lib.rs#L46) is introduced between 2.5 and 2.6?

src/meta/src/barrier/cdc_progress.rs
```diff
@@ -0,0 +1,493 @@
...
+        // TODO(zw): improve init generation.
+        let init_generation = INITIAL_CDC_SPLIT_ASSIGNMENT_GENERATION_ID;
+        let restored = restore_progress(&meta_store).await?;
+        let fragment_id_to_job_id = load_cdc_fragment_table_mapping(&meta_store).await?;
+        let cdc_progress = restored
+            .into_iter()
+            .map(|(job_id, (split_total_count, is_completed))| {
+                (
+                    job_id,
+                    CdcProgress::restore(split_total_count, init_generation, is_completed),
```
hzxa21, line 186: generation will be reset to the initial one across restarts. Will this be a problem?

src/meta/src/barrier/cdc_progress.rs
```diff
@@ -0,0 +1,493 @@
...
+                |(job_id, p)| {
+                    if p.is_completed { Some(*job_id) } else { None }
+                },
+            )
+            .collect()
+    }
+}
+
+async fn restore_progress(meta_store: &SqlMetaStore) -> MetaResult<HashMap<u32, (u64, bool)>> {
+    let split_progress: Vec<(i32, i64, i64)> = cdc_table_snapshot_split::Entity::find()
```
hzxa21, line 301: IIUC, the row count for `cdc_table_snapshot_split` will only increase and no cleanup is done after the cdc backfill is done. Probably not directly related to this PR, but I am thinking whether we still need to access the split state after backfilling is done. If not, will it be better to clean it up after backfilling and treat the non-existing split state as backfill completion? We probably need to associate with the existence of the table catalog to determine whether the table is dropped or not. We can keep the current mechanism as it is if this is simpler but this can be a scalability bottleneck in the future at some point.

src/meta/model/migration/src/lib.rs
zwang28: Yes. It's needed by clusters that have been testing parallelized backfill before this PR.

src/meta/src/barrier/cdc_progress.rs
zwang28: > If not, will it be better to clean it up after backfilling and treat the non-existing split state as backfill completion

We can rewrite the splits into one single [inf, inf] split in meta store during `complete_job`. It's a simple change. I chose to keep it only for debugging purpose. But I agree it's better to clean it up.
BTW we can always clean it up later, e.g. a ctl command that rewrites splits in to one single [inf, inf] if all splits' is_backfill_finished is true.

>We probably need to associate with the existence of the table catalog to determine whether the table is dropped or not.

The cdc_table_snapshot_split table is automatically dropped along with the main table due to the FK constraint.

src/frontend/src/optimizer/plan_node/stream_cdc_table_scan.rs
zwang28: Added.

src/meta/src/barrier/cdc_progress.rs
zwang28: The generation only resets after meta node is restarted.
The barrier carrying expired generation will be rejected by the restarted meta node. Thus it cannot apply its progress to the tracker.
So "generation will be reset to the initial one across restarts" won't cause problem.

src/meta/src/barrier/cdc_progress.rs
zwang28: > We can rewrite the splits into one single [inf, inf] split in meta store during complete_job

I'll apply this change in this PR.

zwang28:
> This PR also includes a refactor that uses the CDC offset to filter out unnecessary data

Reverted the filtering logic in #23051

PR #22944, opened 2025-08-20
feat(expr): support vector_dims and subvector function

Standing-Man:

Two more functions found in pgvector:
- `subvector(vector, integer, integer) → vector`
- `vector_dims(vector) → integer`

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/expr/impl/src/scalar/vector.rs
+++ b/src/expr/impl/src/scalar/vector.rs
@@ -576,3 +576,107 @@ fn l2_normalize(vector: VectorRef<'_>) -> Result<VectorVal> {
         .map_err(|_| ExprError::NumericOverflow)?;
     Ok(result)
 }
+
+/// ```slt
+/// query R
+/// SELECT vector_dims('[1,2,3]'::vector(3));
+/// ----
+/// 3
+///
+/// query R
+/// SELECT vector_dims('[10,20,30,40]'::vector(4));
+/// ----
+/// 4
+///
+/// query error dimensions
+/// SELECT vector_dims('[]'::vector(0));
+/// ```
+#[function("vector_dims(vector) -> int4")]
+fn vector_dim(v: VectorRef<'_>) -> Result<i32> {
+    let vector = v.into_slice();
+    check_dim("vector_dim", vector.len() as i32)?;
+    Ok(vector.len() as i32)
+}
+
+fn check_dim(name: &'static str, dim: i32) -> Result<()> {
+    if dim < 1 {
+        return Err(ExprError::InvalidParam {
+            name,
+            reason: "vector must have at least 1 dimension".into(),
+        });
+    }
+    if dim > (DataType::VEC_MAX_SIZE as i32) {
+        return Err(ExprError::InvalidParam {
+            name,
+            reason: format!("vector cannot have more than {} dimensions", dim).into(),
+        });
+    }
+    Ok(())
+}
+
+/// ```slt
+/// query R
+/// SELECT subvector('[1,2,3,4,5]'::vector(5), 1, 3);
+/// ----
+/// [1,2,3]
+///
+/// query R
+/// SELECT subvector('[1,2,3,4,5]'::vector(5), 3, 2);
+/// ----
+/// [3, 4]
+///
+/// query R
+/// SELECT subvector('[1,2,3,4,5]'::vector(5), -1, 3);
+/// ----
+/// [1]
+///
+/// query R
+/// SELECT subvector('[1,2,3,4,5]'::vector(5), 3, 9);
+/// ----
+/// [3,4,5]
+///
+/// query R
+/// SELECT subvector('[1,2,3,4,5]'::vector, -2147483644, 2147483647);
+/// ----
+/// [1,2]
+///
+/// query R
+/// SELECT subvector('[1,2,3,4,5]'::vector, 3, 2147483647);
+/// ----
+/// [3,4,5]
+/// ```
+#[function("subvector(vector, int4, int4) -> vector", type_infer = "unreachable")]
+fn subvector(v: VectorRef<'_>, start: i32, count: i32) -> Result<VectorVal> {
+    let vector = v.into_slice();
+    let length = vector.len() as i32;
+    if count < 1 {
+        return Err(ExprError::InvalidParam {
+            name: "count",
+            reason: "vector must have at least 1 dimension".into(),
+        });
+    }
+    let mut start = start;
+    let end = if start > length - count {
+        length + 1
+    } else {
+        start + count
+    };
+
+    if start < 1 {
+        start = 1
+    } else if start > length {
+        return Err(ExprError::InvalidParam {
+            name: "start",
+            reason: "vector must have at least 1 dimension".into(),
+        });
+    }
+    check_dim("subvector", end - start)?;
+
+    let result = vector[(start - 1) as usize..(end - 1) as usize]
+        .iter()
+        .copied()
+        .map(|v| Finite32::try_from(v as f32))
+        .try_collect()
+        .map_err(|_| ExprError::NumericOverflow)?;
+    Ok(result)
+}
--- a/src/frontend/src/binder/expr/function/builtin_scalar.rs
+++ b/src/frontend/src/binder/expr/function/builtin_scalar.rs
@@ -442,6 +442,8 @@ impl Binder {
                 ("inner_product", raw_call(ExprType::InnerProduct)),
                 ("vector_norm", raw_call(ExprType::L2Norm)),
                 ("l2_normalize", raw_call(ExprType::L2Normalize)),
+                ("vector_dims",raw_call(ExprType::VectorDims)),
+                ("subvector", raw_call(ExprType::Subvector)),
                 // Functions that return a constant value
                 ("pi", pi()),
                 // greatest and least
--- a/proto/expr.proto
+++ b/proto/expr.proto
@@ -316,6 +316,8 @@ message ExprNode {
     VEC_CONCAT = 754;
     L2_NORM = 755;
     L2_NORMALIZE = 756;
+    VECTOR_DIMS = 757;
+    SUBVECTOR = 758;
 
     // Internal function for schema change
     COMPOSITE_CAST = 800;
--- a/src/frontend/src/expr/pure.rs
+++ b/src/frontend/src/expr/pure.rs
@@ -278,6 +278,8 @@ impl ExprVisitor for ImpureAnalyzer {
             | Type::VecConcat
             | Type::L2Norm
             | Type::L2Normalize
+            | Type::VectorDims
+            | Type::Subvector
             | Type::VnodeUser
             | Type::RwEpochToTs
             | Type::CheckNotNull
--- a/src/frontend/src/optimizer/plan_expr_visitor/strong.rs
+++ b/src/frontend/src/optimizer/plan_expr_visitor/strong.rs
@@ -115,6 +115,8 @@ impl Strong {
             | ExprType::VecConcat
             | ExprType::L2Norm
             | ExprType::L2Normalize
+            | ExprType::VectorDims
+            | ExprType::Subvector
             | ExprType::Greatest
             | ExprType::Least => self.any_null(func_call),
             // ALL: This kind of expression is null if and only if all of its arguments are null.


src/expr/impl/src/scalar/vector.rs
```diff
@@ -576,3 +576,107 @@ fn l2_normalize(vector: VectorRef<'_>) -> Result<VectorVal> {
...
+/// query R
+/// SELECT vector_dims('[10,20,30,40]'::vector(4));
+/// ----
+/// 4
+///
+/// query error dimensions
+/// SELECT vector_dims('[]'::vector(0));
+/// ```
+#[function("vector_dims(vector) -> int4")]
+fn vector_dim(v: VectorRef<'_>) -> Result<i32> {
```
wenym1: The dimension of vector is inherent in its known data type. Is it necessary to provide a function to get from the vector?

src/expr/impl/src/scalar/vector.rs
```diff
@@ -576,3 +576,107 @@ fn l2_normalize(vector: VectorRef<'_>) -> Result<VectorVal> {
...
+/// ----
+/// [1,2]
+///
+/// query R
+/// SELECT subvector('[1,2,3,4,5]'::vector, 3, 2147483647);
+/// ----
+/// [3,4,5]
+/// ```
+#[function("subvector(vector, int4, int4) -> vector", type_infer = "unreachable")]
+fn subvector(v: VectorRef<'_>, start: i32, count: i32) -> Result<VectorVal> {
```
wenym1: The dimension of a vector column should be statically inferable, because the data type of a column should be static to be `DataType::Vector(dimension)`. However, the parameter `start` and `count` is dynamic, and then the dimension is variable and cannot have a fixed dimension to form the fixed data type `DataType::Vector(dimension)`.

src/expr/impl/src/scalar/vector.rs
Standing-Man: Exactly, I also think the `vector_dims` function is somewhat redundant, so I will remove it in next commit.

src/expr/impl/src/scalar/vector.rs
Standing-Man: So how can I define a dynamically displayed vector dimension on `attribute macro`? This confuses me a lot.

wenym1: Theoratical, if the value of `start` and `count` is statically constant, we can derive the dimension of the output vector.

I think in `bind_builtin_scalar_function`, when you handle `subvector `, you can check call `eval_const` on the two inputs to obtain constant `start` and `count`, and then derive the output dimension, and call `FunctionCall::new_unchecked` to specify the input and output data type directly.

Standing-Man: Thank you very much for your review and suggestions. I’ve modified `bind_builtin_scalar_function` to infer vector dimensions from `start` and `count`. Please review it when convenient. 😄

src/frontend/src/binder/expr/function/builtin_scalar.rs
```diff
@@ -434,6 +434,45 @@ impl Binder {
                 ("inner_product", raw_call(ExprType::InnerProduct)),
                 ("vector_norm", raw_call(ExprType::L2Norm)),
                 ("l2_normalize", raw_call(ExprType::L2Normalize)),
+                ("subvector", guard_by_len::<3>(|_, inputs| {
```
wenym1:
```suggestion
                ("subvector", guard_by_len(|_, [vector, start, len]| {
```

src/frontend/src/binder/expr/function/builtin_scalar.rs
```diff
@@ -434,6 +434,45 @@ impl Binder {
...
+                    let vector = inputs[0]
```
wenym1: I think vector does not have to be literal? Theoratically we can do subvector on a vector column.

src/frontend/src/binder/expr/function/builtin_scalar.rs
```diff
@@ -434,6 +434,45 @@ impl Binder {
...
+                        .and_then(|lit| match lit.get_data() {
+                            Some(ScalarImpl::Int32(v)) => Some(*v),
+                            _ => None,
+                        }).ok_or_else(|| ErrorCode::ExprError("`count` must be an Int32 literal".into()))?;
+                    let length = vector.as_slice().len() as i32;
+                    if count < 1 {
+                        return Err(ErrorCode::InvalidParameterValue(format!("count: {}",count)).into());
+                    }
+
+                    let mut start = start;
```
wenym1: Since the vector dimension, subvector start and length are static, I think we can simply ban the case that (start..end) is not within (1..=vector dimension).

src/expr/impl/src/scalar/vector.rs
```diff
@@ -521,3 +521,75 @@ fn l2_norm(vector: VectorRef<'_>) -> F64 {
...
+/// SELECT subvector('[1,2,3,4,5]'::vector(5), -2147483644, 2147483647);
+/// ----
+/// [1,2]
+///
+/// query R
+/// SELECT subvector('[1,2,3,4,5]'::vector(5), 3, 2147483647);
+/// ----
+/// [3,4,5]
+/// ```
+#[function("subvector(vector, int4, int4) -> vector", type_infer = "unreachable")]
```
wenym1: Since `start` and `count` must be const, we don't have to resolve them (including the non-null check implicitly added in the `function` macro) everytime we evaluate it. Instead, we only have to resolve it for once when we build the expression node.

You may refer to the implementation of `regexp_matches`. We can have a `SubvectorContext` to hold the pre-resolved subvector range start and end, and then everytime we evaluate the function, we can simply do `vector[ctx.start..ctx.end]` to get the correct subslice.

Besides, items inside `vector` should already be valid `Finite32` number, and it's unnecessary to do `try_from` again.

wenym1 (commented):
Rest LGTM

src/frontend/src/binder/expr/function/builtin_scalar.rs
Standing-Man: Yes, I need to inspect the dimensions of this vector, but I’m confused about the correct way to retrieve them.

------
I'm using this method to get the dimensions of the vector; I'm not sure if it's appropriate.
```
let length = if let DataType::Vector(length) = vector_expr.return_type() {
    length as i32
} else {
    return Err(ErrorCode::BindError("subvector expects `vector(dim)` input".into()).into());
};
```

src/frontend/src/binder/expr/function/builtin_scalar.rs
Standing-Man: Thank you for suggestion very much. I will simple check the dim of subvector if the dim is within (1.. =vector dim) on next commit.

src/expr/impl/src/scalar/vector.rs
Standing-Man: like this way?
```
#[function(
    "regexp_matches(varchar, varchar, varchar) -> setof varchar[]",
    prebuild = "RegexpContext::from_pattern_flags($1, $2)?"
)]
fn regexp_matches<'a>(
    text: &'a str,
    regex: &'a RegexpContext,
) -> impl Iterator<Item = ListValue> + 'a {
    ...
}
```

src/frontend/src/binder/expr/function/builtin_scalar.rs
wenym1: Yes, this is what I have in mind.

src/expr/impl/src/scalar/vector.rs
wenym1: Yes, it will be similar to this.

src/frontend/src/binder/expr/function/builtin_scalar.rs
```diff
@@ -434,6 +434,52 @@ impl Binder {
                 ("inner_product", raw_call(ExprType::InnerProduct)),
                 ("vector_norm", raw_call(ExprType::L2Norm)),
                 ("l2_normalize", raw_call(ExprType::L2Normalize)),
+                ("subvector", guard_by_len::<3>(|_, [vector_expr, start_expr, len_expr]| {
+                    let length = if let DataType::Vector(length) = vector_expr.return_type() {
```
wenym1: Better to name it `dimension` to avoid confusion to the `len` below.

src/frontend/src/binder/expr/function/builtin_scalar.rs
wenym1: Actually, I meant we can directly ban the case that `start` or `end` is out of the range (1..=vector dim), instead of implicitly shrink to fix the range. For example, let's say the input vector dimension is 5, we can ban the case of (start=-1, len=3) or (start=3, len=4), instead of turning them into (start=1,end=4) or (start=3,end=6).

src/frontend/src/binder/expr/function/builtin_scalar.rs
```diff
@@ -434,6 +434,52 @@ impl Binder {
...
+                    } else {
+                        start + len
+                    };
+                    if start < 1 {
+                        start = 1
+                    } else if start > length {
+                        return Err(ErrorCode::InvalidParameterValue(format!("vector must have at least 1 dimension, not {}", start)).into());
+                    }
+                    let dim = (end - start) as usize;
+                    if !(1..=DataType::VEC_MAX_SIZE).contains(&dim) {
```
wenym1: We can assume that the input vector has dimension no greater than `VEC_MAX_SIZE`, and therefore the subvector must also have dimension no greater than it. Therefore, we can simply only check that the subvector dimension is  greater then 0.

src/frontend/src/binder/expr/function/builtin_scalar.rs
```diff
@@ -434,6 +434,52 @@ impl Binder {
...
                 ("vector_norm", raw_call(ExprType::L2Norm)),
                 ("l2_normalize", raw_call(ExprType::L2Normalize)),
+                ("subvector", guard_by_len::<3>(|_, [vector_expr, start_expr, len_expr]| {
+                    let length = if let DataType::Vector(length) = vector_expr.return_type() {
+                        length as i32
+                    } else {
+                        return Err(ErrorCode::BindError("subvector expects `vector(dim)` input".into()).into());
+                    };
+                    let start = start_expr
+                        .as_literal()
```
wenym1: Actually we can call `expr.eval_const()` so that we can allow const expression. Note that though the return type of `eval_const` is Result, we'd better not using `?` to return the error directly, but instead only check whether it's error and return with more meaningful error message like in current code.

src/expr/impl/src/scalar/vector.rs
```diff
@@ -521,3 +521,58 @@ fn l2_norm(vector: VectorRef<'_>) -> F64 {
...
+    };
+
+    if start < 1 {
+        start = 1
+    }
+
+    let result = vector[(start - 1) as usize..(end - 1) as usize]
+        .iter()
+        .copied()
+        .map(|v| Finite32::try_from(f32::from(v)))
```
wenym1: > Besides, items inside vector should already be valid Finite32 number, and it's unnecessary to do try_from again.

Repost a suggestion in the comment above.

src/frontend/src/binder/expr/function/builtin_scalar.rs
Standing-Man: I noticed that `pgvector` adjusts both `start` and `end`. Should we ensure the output matches PostgreSQL?

src/frontend/src/binder/expr/function/builtin_scalar.rs
Standing-Man: I calculated start as shown below. Is this approach correct?

```
 let start = start_expr
                            .fold_const()
                            .map_err(|e| ErrorCode::ExprError(format!("failed to evaluate start: {}", e).into()))?  
                            .and_then(|datum| match datum {
                                ScalarImpl::Int32(v) => Some(v),
                                _ => None,
                            })
                            .ok_or_else(|| ErrorCode::ExprError("`start` must be an Int32 constant".into()))?;
```

src/expr/impl/src/scalar/vector.rs
Standing-Man: If `subvector` has already constrained start and end within [1..=vector dims]  in `bind_builtin_scalar_function`, is it still necessary to perform preprocessing?

src/frontend/src/binder/expr/function/builtin_scalar.rs
wenym1: Looks good, except that we should call `try_fold_const`, since `fold_const` will panic if the expression is not constant.
https://github.com/risingwavelabs/risingwave/blob/7fc2a1a9e340ff5feb925e73c7adbecd09a73128/src/frontend/src/expr/mod.rs#L403

src/frontend/src/binder/expr/function/builtin_scalar.rs
wenym1: `pgvector` is only a plugin independent to PostgreSQL, so no need to strictly follow it. I think it's more intuitive not to implicitly adjust invalid value.

src/expr/impl/src/scalar/vector.rs
wenym1: Preprocessing it not for constraining the correct start and end. It's for avoiding unnecessary repeated runtime computation on the same const value. In current implementation, for each vector, for both start and len, we will have to repeatedly first evaluate the expression to an ScalarImpl, and then do enum match to resolve the Option<i32>, and then check non-null on the value. If we do preprocessing, we can do it for only once, and then reuse the param for each vector.

Standing-Man: Thanks a lot for your guidance! I now understand why the prebuild step is necessary, and I’ve implemented the Preprocess code.

src/expr/impl/src/scalar/vector.rs
Standing-Man: I added `from_unchecked` function for `Finite32` to replace `try_from`, so it can be used without additional checks.

src/frontend/src/binder/expr/function/builtin_scalar.rs
```diff
@@ -434,6 +434,57 @@ impl Binder {
...
+                    if start < 1 || end < 1 || start > dimensions || end > dimensions {
+                        return Err(ErrorCode::InvalidParameterValue(format!(
+                                "vector slice range out of bounds: start={}, end={}, valid range is [1, {}]",
+                                start,
+                                end,
+                                dimensions
+                            )).into());
+                    }
+
+                    let dim = (end - start + 1) as usize;
```
wenym1: The check on `dim` seems unnecessary, because dim is strictly equal to `len` in current code. We can actually only use `len` instead of introducing a new `dim`.

src/frontend/src/binder/expr/function/builtin_scalar.rs
```diff
@@ -434,6 +434,57 @@ impl Binder {
...
+                            _ => None,
+                        })
+                        .ok_or_else(|| ErrorCode::ExprError("`count` must be an Int32 constant".into()))?;
+                    if len < 1 {
+                        return Err(ErrorCode::InvalidParameterValue(format!("vector must have at least 1 dimension, not {}", len)).into());
+                    }
+
+                    let end = start + len - 1;
+
+                    if start < 1 || end < 1 || start > dimensions || end > dimensions {
```
wenym1: Since we ensure that `len >= 1`, in most cases we can ensure that `start <= end`, and therefore we only need to check `start < 1 || end > dimension`. But we may have malicious unreasonably large input `len` to make the `end = start + len - 1` overflow to a minus value, which still satisfies `end < dimension`. 

So I think before this check, we can pre check the `len` to be `1 <= len <= MAX_VECTOR_SIZE`, and then the current check can be simplified to `start < 1 || end > dimension`.

src/expr/impl/src/scalar/vector.rs
```diff
@@ -521,3 +521,60 @@ fn l2_norm(vector: VectorRef<'_>) -> F64 {
...
+/// ```
+#[function(
+    "subvector(vector, int4, int4) -> vector",
+    prebuild = "SubvectorContext::from_start_count($1, $2)?",
+    type_infer = "unreachable"
+)]
+fn subvector(v: VectorRef<'_>, ctx: &SubvectorContext) -> Result<VectorVal> {
+    let vector = v.as_slice();
+
+    let result = vector[(ctx.start - 1) as usize..(ctx.end - 1) as usize]
```
wenym1: Seems that we can store `start` and `end` directly as the start and end of slice range, so that we can directly use `vector[ctx.start..ctx.end]`.

src/expr/impl/src/scalar/vector.rs
```diff
@@ -521,3 +521,60 @@ fn l2_norm(vector: VectorRef<'_>) -> F64 {
...
+        .iter()
+        .copied()
+        .map(|v| Finite32::from_unchecked(f32::from(v)))
```
wenym1: How about we directly provide a method ``VectorInner::subvector(start, end) -> VectorVal` to return `VectorInner(self.0.as_ref()[start..end].to_vec())`? And then we don't need the new `Finite32::from_unchecked`.

src/expr/impl/src/scalar/vector.rs
```diff
@@ -521,3 +521,60 @@ fn l2_norm(vector: VectorRef<'_>) -> F64 {
...
+        Ok(Self {
+            start,
+            end: start + count,
+        })
+    }
+}
+
+/// ```slt
+/// query R
+/// SELECT subvector('[1,2,3,4,5]'::vector(5), 1, 3);
```
wenym1, line 542: We can also add tests on invalid input.

Besides, we can also use values list as input to test the function with column input like
```
select subvector(vec, 1, 3) from (values ('[1,2,3,4,5]'::vector(5)), ('[6,7,8,9,10]'::vector(5))) as t(vec);
```

src/expr/impl/src/scalar/vector.rs
Standing-Man: Thanks for your guidance! 😄 I’ve implemented `subvector` in `VectorInner`, added more test cases, and updated the checks for `start`, `len`, and the `dimension`.

wenym1 (approved):
LGTM!

Standing-Man:
Your feedback is very helpful, thank you! @wenym1

PR #22953, opened 2025-08-21
fix(ci): remove non-existent query for nexmark action

kwannoel:

`q6` doesn't exist + fix default name (cannot have Uppercase) + add link to dashboard

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/.github/workflows/run-perf-test.yml
+++ b/.github/workflows/run-perf-test.yml
@@ -16,7 +16,7 @@ on:
         description: 'Nexmark Queries to Benchmark'
         required: true
         type: string
-        default: 'q0,q4,q6,q15,q20,q101'
+        default: 'q0,q4,q15,q20,q22,q101'
 
 jobs:
   run-perf-test:
--- a/.github/workflows/run-perf-test.yml
+++ b/.github/workflows/run-perf-test.yml
@@ -6,7 +6,7 @@ on:
         description: 'Performance test namespace (date-time will be appended)'
         required: true
         type: string
-        default: 'Untitled'
+        default: 'untitled'
       rw_image_tag:
         description: 'RisingWave Image Tag (docker image tag, e.g. v2.6.0-alpha--bench--e7ca9bb--kwannoel-enhance-backpressure-graph6)'
         required: true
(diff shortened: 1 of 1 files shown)


BugenZhao (approved):

.github/workflows/run-perf-test.yml
```diff
@@ -6,7 +6,7 @@ on:
         description: 'Performance test namespace (date-time will be appended)'
         required: true
         type: string
-        default: 'Untitled'
+        default: 'untitled'
```
kwannoel, line 9: > Namespace: Kubernetes supports multiple virtual clusters backed by the same physical cluster. These virtual clusters are called [namespaces](https://kubernetes.io/docs/tasks/administer-cluster/namespaces/). They let you partition resources into logically named groups.
>
> Dashboard offers all available namespaces in a dropdown list, and allows you to create a new namespace. The namespace name may contain a maximum of 63 alphanumeric characters and dashes (-) but can not contain capital letters. Namespace names should not consist of only numbers. If the name is set as a number, such as 10, the pod will be put in the default namespace.
>
> In case the creation of the namespace is successful, it is selected by default. If the creation fails, the first namespace is selected.

https://kubernetes.io/docs/tasks/access-application-cluster/web-ui-dashboard/#specifying-application-details

.github/workflows/run-perf-test.yml
```diff
@@ -72,4 +72,7 @@ jobs:
           branch: main
           commit: HEAD
           message: ':github: Triggering Nexmark Benchmark with image tag: ${{ env.RW_VERSION }}'
-          build_env_vars: '{ "BENCH_NAMESPACE": "${{ env.BENCH_NAMESPACE }}", "BENCH_TESTBED": "${{ env.BENCH_TESTBED }}", "RW_VERSION": "${{ env.RW_VERSION }}", "NEXMARK_QUERY": "${{ env.NEXMARK_QUERY }}", "KAFKA_EVENTS": "${{ env.KAFKA_EVENTS }}", "KAFKA_PARTITION": "${{ env.KAFKA_PARTITION }}", "ENABLE_BLACKHOLE": "${{ env.ENABLE_BLACKHOLE }}", "SKIP_DASHBOARD": "${{ env.SKIP_DASHBOARD }}" }'
\ No newline at end of file
+          build_env_vars: '{ "BENCH_NAMESPACE": "${{ env.BENCH_NAMESPACE }}", "BENCH_TESTBED": "${{ env.BENCH_TESTBED }}", "RW_VERSION": "${{ env.RW_VERSION }}", "NEXMARK_QUERY": "${{ env.NEXMARK_QUERY }}", "KAFKA_EVENTS": "${{ env.KAFKA_EVENTS }}", "KAFKA_PARTITION": "${{ env.KAFKA_PARTITION }}", "ENABLE_BLACKHOLE": "${{ env.ENABLE_BLACKHOLE }}", "SKIP_DASHBOARD": "${{ env.SKIP_DASHBOARD }}" }'
+      - name: 'Notify dashboard link'
+        run: |
+          echo 'To compare query results, visit: http://metabase.risingwave-cloud.xyz/dashboard/1405-nexmark-8c32g-1cn-rw-source-throughput?start_date=2024-12-01'
```
kwannoel, line 78: Just hardcode the dashboard URL for now. Most people should use the default cluster configuration.

kwannoel: Validated: https://github.com/risingwavelabs/risingwave/actions/runs/17122566052/job/48566693389

PR #22955, opened 2025-08-21
feat(streaming): only materialize expr for retract input

wenym1:

Materiailized expr is only for streaming consistency, to be able to get the exact same old value when dealing with delete. However, for append only workload, this seems unnecessary, because we can expect there is no delete in the workload. Therefore, in this PR, we change to avoid materializing the expr output for append only output. This would be helpful for vector index, to avoid unnecessarily storing a new copy of embedding data in the output of embedding function, which can be very large.

For upsert workload, we only avoid materializing expr output when the stream keys are not created from impure functions.

## Checklist

## Documentation

<details>

</details>


wenym1:
* **#22960** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22960?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#22955** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22955?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22955?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

e2e_test/streaming/materialized_project.slt
BugenZhao, line 1: Interesting. Why was this file not initially be a planner test instead?

src/frontend/src/optimizer/plan_node/logical_project.rs
```diff
@@ -265,29 +266,54 @@ impl ToStream for LogicalProject {
...
+                    if impure_exprs.is_empty() {
                         None
+                    } else if kind == StreamKind::Upsert
+                        && new_input
+                            .stream_key()
+                            .into_iter()
+                            .flatten()
+                            .all(|stream_key_idx| !impure_expr_indices.contains(stream_key_idx))
+                    {
+                        // no need to materialize for Upsert input when none of stream keys is impure expr
```
BugenZhao:
```suggestion
                        // We're operating on non-stream-key columns of upsert stream, no need to materialize.
```

src/frontend/src/optimizer/plan_node/logical_project.rs
```diff
@@ -319,10 +345,6 @@ impl ToStream for LogicalProject {
 
                 let core = generic::Project::new(final_exprs, mat_exprs_plan);
                 StreamProject::new(core).into()
-            } else {
-                // No impure expressions, create a regular `StreamProject`
-                let core = generic::Project::new(self.exprs().clone(), new_input);
-                StreamProject::new(core).into()
             }
         } else {
             // Materialized expressions feature is not enabled, create a regular `StreamProject`
```
BugenZhao:
```suggestion
            // No expressions to materialize or the feature is not enabled, create a regular `StreamProject`
```

BugenZhao (approved):
LGTM

e2e_test/streaming/materialized_project.slt
wenym1: No idea. I think we can remove this test, because the explain output has been tested in the planner test, and it takes some unnecessary efforts to update the plan output in the slt file🥵

chenzl25 (approved):
LGTM

PR #22956, opened 2025-08-21
refactor(on conflict): optimize materialize executor by storing `OwnedRow` in `ChangeBuffer`

wcy-fdu:

The materialize executor had redundant deserialization issues:
In `ConflictBehavior::Overwrite` and `DoUpdateIfNotNull:` Rows were deserialized for conflict resolution, then serialized to Bytes for `ChangeBuffer` storage, and need to be deserialized again during generate_output
This caused unnecessary performance overhead due to repeated serialization/deserialization cycles

This pr modified `ChangeBuffer` to  store `OwnedRow` instead of `Bytes` to eliminate redundant deserialization.

Although doing so may take up a little more memory, it is worth it here, and it is also more convenient for processing toast later.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/executor/mview/materialize.rs
+++ b/src/stream/src/executor/mview/materialize.rs
@@ -26,14 +26,13 @@ use risingwave_common::bitmap::Bitmap;
 use risingwave_common::catalog::{
     ColumnDesc, ColumnId, ConflictBehavior, TableId, checked_conflict_behaviors,
 };
-use risingwave_common::row::{CompactedRow, RowDeserializer};
+use risingwave_common::row::{CompactedRow, OwnedRow};
 use risingwave_common::util::chunk_coalesce::DataChunkBuilder;
 use risingwave_common::util::iter_util::{ZipEqDebug, ZipEqFast};
 use risingwave_common::util::sort_util::{ColumnOrder, OrderType, cmp_datum};
 use risingwave_common::util::value_encoding::{BasicSerde, ValueRowSerializer};
 use risingwave_pb::catalog::Table;
 use risingwave_pb::catalog::table::Engine;
-use risingwave_storage::mem_table::KeyOp;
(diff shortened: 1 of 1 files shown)


BugenZhao:
> Although doing so may take up a little more memory, it is worth it here, and it is also more convenient for processing toast later.

Reminds me again of https://github.com/risingwavelabs/risingwave/issues/20017. Besides taking more memory, another downside of `OwnedRow` is that it's not efficient when cloning.

wcy-fdu:
> > Although doing so may take up a little more memory, it is worth it here, and it is also more convenient for processing toast later.
> 
> Reminds me again of #20017. Besides taking more memory, another downside of `OwnedRow` is that it's not efficient when cloning.

Indeed, but the extra clone is introduced in `ConflictBehavior::DoUpdateIfNotNull`. In most cases, the default behavior is `overwrite`, and this PR does not introduce additional clones in main scenes.

Saving CPU (avoiding deserialization) does require memory consumption, for example, the buffer structure changes from bytes to `OwnedRow`, and when performing `Do update if not null`, an additional clone of the old row is performed.

In fact, this pr is the predecessor pr of [supporting pg toast pr](https://github.com/risingwavelabs/risingwave/pull/22852). When processing toast, we have to deser old row and new row, so this pr is to optimize this. In most cases, as long as there are varchar columns defined in postgres, you need to check toast.

hzxa21:
> > Although doing so may take up a little more memory, it is worth it here, and it is also more convenient for processing toast later.
> 
> Reminds me again of #20017. Besides taking more memory, another downside of `OwnedRow` is that it's not efficient when cloning.

Yeah, I think at some point we may need to revisit #20017, especially considering that we are paying more attention to performance optimization.

hzxa21 (approved):
The core logic remains the same. LGTM.

hzxa21:
Link https://github.com/risingwavelabs/risingwave/pull/22961 here as well

PR #22960, opened 2025-08-21
feat(embedding): pass embedding options via jsonb const

wenym1:

Previously when we call the builtin `openai_embedding` function, we pass the `api_key` and `model` as the first two const parameter to the function. However, when we want to support more config options, we will have to add more parameters at the beginning, and it will be hard to maintain backward compatibility. 

In this PR, we change to pass all config options in a single jsonb parameter. Apart from the two `api_key` and `model` parameters, we add `org_id`, `project_id` and `api_base`, which are options in the config builder of open ai client.

We will also cover the `openai_embedding` function in e2e test. We implement a mocked openai embedding http service, and return preset embedding for the input text.

## Checklist

## Documentation

<details>

This is a breaking change to the `openai_embedding` builtin function introduced in release 2.5.

In release 2.5, the function receives 3 parameter `openai_embedding(api_key string, model string, input_text string)`. In the upcoming release, we will change to `openai_embedding(config jsonb, input_text string)`. Like in release 2.5 where the `api_key` and `model` should be constant, the jsonb `config` should also be constant.

The jsonb config support the following config options

| option name | required | |
|------|-------|-----|
| model| yes |
|api_key| no |
|org_id| no| 
|proj_id| no|
|api_base| no| call the embedding service to a different endpoint. The default one is `https://api.openai.com/v1` |

Users can create sql udf to store the config in the udf, so that they don't have to always specify the config. Following is an example
```
create function get_embedding(string) returns vector(3) language sql as $$
select openai_embedding('{"model": "text-embedding-ada-002", "api_key": "$OPENAI_API_KEY"}'::jsonb, $1)::vector(3);
$$;
```

</details>


wenym1:
* **#22960** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22960?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22960?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#22955** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22955?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

wenym1:
May need to consider whether it's safe to allow specifying random api base, since it will be allowed to fire http request to other machine in the cluster, though the path in request url are fixed to be <endpoint>/v1/embeddings.

hzxa21:
> May need to consider whether it's safe to allow specifying random api base, since it will be allowed to fire http request to other machine in the cluster, though the path in request url are fixed to be /v1/embeddings.

Loop in @wjf3121 to confirm. I remembered we have a network policy to make sure that the network communication between two namespaces in the same cluster is isolated.

hzxa21:
Also the issue is not only related to the embedding function but more generally applicable to all clusters with UDF enabled.

wjf3121:
> > May need to consider whether it's safe to allow specifying random api base, since it will be allowed to fire http request to other machine in the cluster, though the path in request url are fixed to be /v1/embeddings.
> 
> Loop in @wjf3121 to confirm. I remembered we have a network policy to make sure that the network communication between two namespaces in the same cluster is isolated.

[Current network work isolation rule](https://github.com/risingwavelabs/risingwave-cloud/blob/87226b1502e3f033d7484f56028b9ceb40fdbf3d/mgmt/internal/infra/network/utils.go#L45-L66) can be described as the following:

1. Resources of a tenant are put in one dedicated namespace.
2. For any pod in a tenant namespace, we only allow the following ingress traffic to it:
   - traffic from the namespace itself
   - traffic from a non-tenant namespace  (all tenant namespace are labeled with a special key)

If we want the embedding service or UDF server in general to be accessed by only one tenant, one of the following options should work:
1. simply put it in the same tenant namespace.
2. put all UDF servers of one tenant into a dedicated namespace so we can construct a ingress rule allowing traffic only from its corresponding tenant namespace.

src/expr/impl/src/scalar/ai_model.rs
```diff
@@ -35,13 +36,45 @@ pub struct OpenAiEmbeddingContext {
...
+
+        let expect_string = |key: &str| -> Result<String> {
+            get_string(key)?.ok_or_else(|| invalid_param_err(format!("`{}` must be set", key)))
+        };
+
+        let model = expect_string("model")?;
+        let api_key = get_string("api-key")?;
+        let org_id = get_string("org-id")?;
+        let proj_id = get_string("project-id")?;
+        let api_base = get_string("api-base")?;
```
BugenZhao: Shall we define a struct here and derive `Deserialize` for it?

BugenZhao: Wondering if it's some convention to use `kebab-case` as the configuration key here?

src/expr/impl/src/scalar/ai_model.rs
```diff
@@ -173,55 +209,39 @@ impl Expression for OpenAiEmbedding {
     }
 }
 
-#[build_function("openai_embedding(varchar, varchar, varchar) -> float4[]")]
```
BugenZhao, line 176: The old syntax has already been released. Will this be a breaking change?

src/expr/impl/src/scalar/ai_model.rs
wenym1: No, it's just a random choice. Any suggestion?

src/expr/impl/src/scalar/ai_model.rs
wenym1: Yes, it will be a breaking change. I've included the breaking change in the release note section. I think the feature is still experimental, so no need to handle backward compatibility yet.

src/expr/impl/src/scalar/ai_model.rs
wenym1: After applying https://github.com/risingwavelabs/risingwave/pull/22960#discussion_r2300354879, it is easier to use snake style.

BugenZhao: > Any suggestion?

Using `snake_case` follows OpenAI's API as well. LGTM

BugenZhao (approved):
LGTM!

PR #22968, opened 2025-08-22
feat(pg-cdc): monitor Postgres CDC LSN Progression

wcy-fdu:

Postgres CDC continuously advances the LSN during normal operations to inform the upstream PG that it can perform cleanup. The specific process is as follows:

1. After the checkpoint is completed, the source state table persists an increasing LSN.
2。 After the source executor waits for the checkpoint, it calls the JNI commit offset to inform Debezium of the increasing LSN.
3. Debezium then reports this to the upstream PG, which advances the confirm_flush_lsn of the replication_slot associated with the current source.

We have long lacked monitoring in this area, which would be beneficial for debugging and quickly confirming the status of PG CDC. This PR adds the corresponding monitoring to these three parts: the state table LSN, JNI commit LSN, and confirm_flush_lsn, with confirm_flush_lsn being obtained by periodically querying the system tables of the upstream Postgres.

image

## Checklist

## Documentation

<details>

</details>


src/stream/src/executor/backfill/cdc/cdc_backfill.rs
```diff
@@ -149,6 +212,21 @@ impl<S: StateStore> CdcBackfillExecutor<S> {
         let schema_table_name = self.external_table.schema_table_name().clone();
         let external_database_name = self.external_table.database_name().to_owned();
 
+        // Start background task to periodically query confirm_flush_lsn for PostgreSQL CDC
+        let streaming_metrics = self.streaming_metrics.clone();
+        let source_id = self.actor_ctx.id;
+        let slot_name = self.properties.get("slot.name").cloned();
+        let external_table = self.external_table.clone();
+
+        if let Some(slot_name) = slot_name {
```
zwang28: This will spawn a lsn monitor for every backfill actor.
I believe one lsn monitor for one cdc source is enough.

src/stream/src/executor/backfill/cdc/cdc_backill_v2.rs
```diff
@@ -163,6 +162,22 @@ impl<S: StateStore> ParallelizedCdcBackfillExecutor<S> {
         let mut state_impl =
             ParallelizedCdcBackfillState::new(self.state_table, METADATA_STATE_LEN);
         let mut upstream_chunk_buffer: Vec<StreamChunk> = vec![];
+
+        // Start background task to periodically query confirm_flush_lsn for PostgreSQL CDC
```
zwang28: ditto

zwang28 (changes requested):

src/stream/src/executor/backfill/cdc/cdc_backfill.rs
wcy-fdu: Move this query to meta and make it per source.

wcy-fdu:
This pr can be reviewed. cc @zwang28 @hzxa21 ❤️

src/connector/src/source/cdc/external/postgres.rs
```diff
@@ -451,6 +449,24 @@ impl PostgresExternalTableReader {
         }
     }
 
+    /// Query the `confirmed_flush_lsn` from `pg_replication_slots` for a specific slot
+    pub async fn query_confirm_flush_lsn(&self, slot_name: &str) -> ConnectorResult<Option<u64>> {
```
zwang28: unused?

wcy-fdu: Removed.

src/stream/src/executor/source/source_executor.rs
```diff
@@ -954,6 +990,19 @@ impl<S: StateStore> WaitCheckpointWorker<S> {
...
+                            if let WaitCheckpointTask::CommitCdcOffset(Some((split_id, offset))) =
+                                &task
+                                && let Ok(source_id) = u64::from_str(split_id.as_ref())
+                                && let Some(lsn_value) = extract_pg_cdc_lsn(offset)
+                            {
+                                self.metrics
+                                    .pg_cdc_jni_commit_offset_lsn
+                                    .with_guarded_label_values(&[&source_id.to_string()])
+                                    .set(lsn_value as i64);
+                            }
```
hzxa21: I think it is better to update the metric when the commit operation is done. Also, it is a bit strange to have cdc specific logic here because the `WaitCheckpointWorker` is a general worker. Can we do it in `task.run`?

src/meta/src/stream/source_manager/worker.rs
```diff
@@ -358,6 +363,45 @@ impl ConnectorSourceWorker {
...
+                }
+                Err(e) => {
+                    tracing::error!(
+                        "Failed to query confirmed_flush_lsn for source {}: {}",
+                        source_id,
+                        e.as_report()
+                    );
+                }
+            }
+        }
```
hzxa21: same here. ConnectorSourceWorker is a general struct applicable to different kinds of source. Having cdc specific logic here is a leak of abstraction. How about have a new method in the split enumerator trait named `on_tick` ([example](https://github.com/risingwavelabs/risingwave/blob/1da7a3dde5995cfae48b3c523bea549ed8d4c04c/src/connector/src/source/base.rs#L214)) and let cdc split enumerator implements this method to emit metric? ConnectorSourceWorker can call `self.enumerator.on_tick`.

hzxa21: By doing so, you also don't need to add `tokio_postgres` dep in risingwave-meta because the implementation sits inside risingwave-connector

wcy-fdu: Good idea! The change is complete, and it can indeed avoid using `tokio_postgres` in meta crate.

src/connector/src/source/cdc/enumerator/mod.rs
```diff
@@ -117,13 +135,126 @@ where
...
         })
     }
 
     async fn list_splits(&mut self) -> ConnectorResult<Vec<DebeziumCdcSplit<T>>> {
         Ok(self.list_cdc_splits())
     }
+
+    async fn on_tick(&mut self) -> ConnectorResult<()> {
+        // For Postgres CDC, query the upstream Postgres confirmed flush lsn and monitor it.
+        if std::any::TypeId::of::<T>() == std::any::TypeId::of::<Postgres>() {
```
hzxa21: This is strange and very JAVA. You can have a specific `impl ... for DebeziumSplitEnumerator<Postgres>` instead.

https://github.com/risingwavelabs/risingwave/blob/04e1290c1360dc288c5f0a14dde7e82947ca9627/src/connector/src/source/cdc/enumerator/mod.rs#L148

src/stream/src/executor/source/source_executor.rs
```diff
@@ -52,6 +53,17 @@ use crate::task::LocalBarrierManager;
 /// some latencies in network and cost in meta.
 pub const WAIT_BARRIER_MULTIPLE_TIMES: u128 = 5;
 
+/// Extract LSN value from PostgreSQL CDC offset JSON string
+///
+/// This function parses the offset JSON and extracts the LSN value from the sourceOffset.lsn field.
+/// Returns Some(lsn) if the LSN is found and can be parsed as u64, None otherwise.
+fn extract_pg_cdc_lsn(offset_str: &str) -> Option<u64> {
```
hzxa21: The implementation here leverages the presenence of `lsn` field in the source offset to determine whether it is a PG source, which looks flaky to me. Given that our [split impl is strongly type](https://github.com/risingwavelabs/risingwave/blob/04e1290c1360dc288c5f0a14dde7e82947ca9627/src/connector/src/source/base.rs#L701), can we match against the type before extracting the offset?

src/connector/src/source/cdc/enumerator/mod.rs
wcy-fdu: Thanks for the suggestion.

Issue #21635, opened 2025-04-29
enable clippy lints for potential stack overflow & fix them


BugenZhao:
Under `[workspace.lints.clippy]` in `Cargo.toml`:

```toml
large_futures = "warn"
large_stack_arrays = "warn"
large_stack_frames = "warn"
large_types_passed_by_value = "warn"
```

Can help issues like
- #21335
- https://github.com/risingwavelabs/risingwave/issues/21427
- https://github.com/risingwavelabs/risingwave/issues/19906

PR #22971, opened 2025-08-22
chore(ctl): add clippy warning on large future size and stack frame

ChrisWrenDev:

Issue: #21635

This is part of the ongoing effort to enable Clippy lints for potential stack overflows (#21635) across the codebase. Here, the focus is on the `risingwave_ctl` crate.

This PR refactors `ctl/src/lib.rs` to address a Clippy `large_stack_frames` warning by decomposing the monolithic `start_impl` async function into several smaller, purpose-scoped async helpers.

**Key changes and rationale:**

- Smaller async helpers: Each command group is now handled by its own `#[inline(never)]` async function. This prevents a single oversized state machine from capturing many large sub-futures at once.
- Reduced frame sizes: `-Zprint-type-size` previously showed multiple ~12–13 KB futures being held concurrently within one state machine. Splitting the logic ensures each branch only awaits the heavy future it actually uses.
- Improved compiler output: The compiler now generates several small state machines instead of one massive one, significantly reducing stack usage.
- Measured impact: The top-level dispatch handler’s stack size dropped from 1,263,519 to 611,767. Further splitting (notably in `hummock_dispatch`) brought it below Clippy’s `stack-size-threshold` of 512,000.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/ctl/src/lib.rs
+++ b/src/ctl/src/lib.rs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #![feature(let_chains)]
+#![warn(clippy::large_futures, clippy::large_stack_frames)]
 
 use anyhow::Result;
 use clap::{Args, Parser, Subcommand};
@@ -560,35 +561,60 @@ pub async fn start_fallible(opts: CliOpts, context: &CtlContext) -> Result<()> {
 
 async fn start_impl(opts: CliOpts, context: &CtlContext) -> Result<()> {
     match opts.command {
-        Commands::Compute(ComputeCommands::ShowConfig { host }) => {
-            cmd_impl::compute::show_config(&host).await?
+        Commands::Compute(sub) => compute_dispatch(sub).await,
+        Commands::Hummock(sub) => hummock_dispatch(context, sub).await,
+        Commands::Table(sub) => table_dispatch(context, sub).await,
+        Commands::Meta(sub) => meta_dispatch(context, sub).await,
+        Commands::Scale(sub) => scale_dispatch(context, sub).await,
+        Commands::Bench(cmd) => cmd_impl::bench::do_bench(context, cmd).await,
+        Commands::AwaitTree(sub) => await_tree_dispatch(context, sub).await,
+        Commands::Profile(sub) => profile_dispatch(context, sub).await,
+        Commands::Throttle(sub) => throttle_dispatch(context, sub).await,
+    }
+}
+
+#[inline(never)]
+async fn compute_dispatch(cmd: ComputeCommands) -> Result<()> {
+    match cmd {
+        ComputeCommands::ShowConfig { host } => {
+            cmd_impl::compute::show_config(&host).await?;
+            Ok(())
         }
-        Commands::Hummock(HummockCommands::DisableCommitEpoch) => {
-            cmd_impl::hummock::disable_commit_epoch(context).await?
+    }
+}
+
+#[inline(never)]
+async fn hummock_dispatch(context: &CtlContext, cmd: HummockCommands) -> Result<()> {
+    match cmd {
+        HummockCommands::DisableCommitEpoch => {
+            cmd_impl::hummock::disable_commit_epoch(context).await?;
+            Ok(())
         }
-        Commands::Hummock(HummockCommands::ListVersion {
+        HummockCommands::ListVersion {
             verbose,
             verbose_key_range,
-        }) => {
+        } => {
             cmd_impl::hummock::list_version(context, verbose, verbose_key_range).await?;
+            Ok(())
         }
-        Commands::Hummock(HummockCommands::ListVersionDeltas {
+        HummockCommands::ListVersionDeltas {
             start_id,
             num_epochs,
-        }) => {
+        } => {
             cmd_impl::hummock::list_version_deltas(
                 context,
                 HummockVersionId::new(start_id),
                 num_epochs,
             )
             .await?;
+            Ok(())
         }
-        Commands::Hummock(HummockCommands::ListKv {
+        HummockCommands::ListKv {
             epoch,
             table_id,
             data_dir,
             use_new_object_prefix_strategy,
-        }) => {
+        } => {
             cmd_impl::hummock::list_kv(
                 context,
                 epoch,
@@ -597,16 +623,18 @@ async fn start_impl(opts: CliOpts, context: &CtlContext) -> Result<()> {
                 use_new_object_prefix_strategy,
             )
             .await?;
+            Ok(())
         }
-        Commands::Hummock(HummockCommands::SstDump(args)) => {
-            cmd_impl::hummock::sst_dump(context, args).await.unwrap()
+        HummockCommands::SstDump(args) => {
+            cmd_impl::hummock::sst_dump(context, args).await.unwrap();
+            Ok(())
         }
-        Commands::Hummock(HummockCommands::TriggerManualCompaction {
+        HummockCommands::TriggerManualCompaction {
             compaction_group_id,
             table_id,
             level,
             sst_ids,
-        }) => {
+        } => {
             cmd_impl::hummock::trigger_manual_compaction(
                 context,
                 compaction_group_id,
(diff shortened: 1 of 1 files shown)


wenym1:
I had a try locally to add the clippy warning on the crate. It seems that in current main branch, the crate only has the warning of `large_stack_frames`, and no warning on `large_futures`, so I suppose this PR is mostly for resolving the `large_stack_frames`.

I read the doc of [large_stack_frames](https://rust-lang.github.io/rust-clippy/master/index.html#large_stack_frames), and it says, 

```
The stack size that clippy sees is an estimated value and can be vastly different from the actual stack usage after optimizations passes have run (especially true in release mode)
```

I remember in https://github.com/risingwavelabs/risingwave/pull/22906#issuecomment-3192053424, I once observed that the the number of allocated bytes in the warning will increase as the number of branches of enum matching increases. I tried to reproduce the warning locally with the following code.

```
async fn huge() {
    let huge1 = [0u32;64000];
    let huge2 = [0u32;64000];
    tokio::time::sleep(Duration::from_secs(1)).await;
    let _huge = huge1;
    let _huge = huge2;
}

async fn huge1() {
    huge().await
}

async fn huge2() {
    huge().await
}

async fn huge3() {
    huge().await
}

#[inline(never)]
async fn many_huge() {
    huge().await;
    let num: u32 = rand::random();
    match num {
        2 => {
            huge2().await;
        }
        1 => {
            huge1().await;
        }
        _ => {
            huge3().await;
        }
    }
}

#[tokio::main]
async fn main() {
    many_huge().await;
    println!("Hello, world!");
}
```

The printed size of the `many_huge()` future is `512136`, which approxiates the size of the `huge()` future that captures 2 u32 arrays each with 64000 elements. I also got the warning on `large_stack_frames`

```
warning: this function may allocate 6145989 bytes on the stack
  --> src/main.rs:30:22
   |
30 |   async fn many_huge() {
   |  ______________________^
31 | |     huge().await;
```

If I comment out one of the match branches, the printed size of the future is still `512136 `, while the bytes number in the warning shrinks to 2/3 of the previous one.
```
warning: this function may allocate 4609500 bytes on the stack
  --> src/main.rs:30:22
   |
30 |   async fn many_huge() {
   |  ______________________^
31 | |     huge().await;
```

This suggests that, as we add more match branches, the size of the future itself won't change, while the warning of allocated bytes will be the sum of each arm. I inspect the MIR of the `many_huge`, and it looks like the following
```
fn many_huge::{closure#0}(_1: Pin<&mut {async fn body of many_huge()}>, _2: &mut Context<'_>) -> Poll<()> {
    debug _task_context => _37;
...
    let mut _12: {async fn body of huge2()};  // <---------------- future of the first arm
    let mut _13: {async fn body of huge2()};
    let mut _14: std::task::Poll<()>;
    let mut _15: std::pin::Pin<&mut {async fn body of huge2()}>;
    let mut _16: &mut {async fn body of huge2()};
    let mut _17: &mut std::task::Context<'_>;
    let mut _18: isize;
    let mut _20: {async fn body of huge1()}; // <-----------------future of the second arm
    let mut _21: {async fn body of huge1()};
    let mut _22: std::task::Poll<()>;
    let mut _23: std::pin::Pin<&mut {async fn body of huge1()}>;
    let mut _24: &mut {async fn body of huge1()};
    let mut _25: &mut std::task::Context<'_>;
    let mut _26: isize;
    let mut _28: {async fn body of huge3()}; // <----------------future of the third arm
    let mut _29: {async fn body of huge3()};
...
```

Basically, in the initial MIR that has not gone through optimization, the function that drives the future state machine will pre-allocate the stack space for all variables in all arms, even though only one of the arms will be reached in runtime, and I guess this is where the warning comes from. 

However, this warning is absolutely a false positive one, because only one branch will be reached at runtime, and the variables in different arms can actually share the same stack space, working like a `union` type, and I believe any compiler with basic functionality won't generate binary that allocates stack space individually for variables in all arms.

To summarize, I think the `large_stack_frames` is a false positive warning, and it does not worth it to sacrifice code readablility for optimizing it. Annotating the `start_impl` with `#[expect(clippy::large_stack_frames)]` to bypass the false positive warning should be enough.

ChrisWrenDev:
Thanks @wenym1 for the detailed analysis and the MIR walk-through — super helpful.

I re-ran `-Zprint-type-sizes` on the main branch and the post-layout numbers line up with your explanation:
- `{async fn body of start_impl()}`: 13,744 B
- Largest inner arm: `{async fn body of bench::do_bench()}`: 13,048 B

This confirms the generator’s variant/union-like layout — only one arm is live at a time — so the practical upper bound is ≈ the largest arm, not the sum across arms that Clippy reports pre-opt.

It also explains why some things I have been trying haven't worked as expected:
- Early returns in individual arms didn’t reduce the overall size as Clippy is still “seeing” the locals from every arm and adds them up, regardless of which branch actually returns at runtime.
- Scoped blocks before .await didn’t help as the locals are still counted in pre-opt MIR even if the block ends before the await.

I’ll revert the changes to keep the dispatcher readable and add a targeted expect with rationale :+1:

wenym1 (approved):
LGTM

BugenZhao:
> Basically, in the initial MIR that has not gone through optimization, the function that drives the future state machine will pre-allocate the stack space for all variables in all arms, even though only one of the arms will be reached in runtime, and I guess this is where the warning comes from.

However, according to the issue https://github.com/rust-lang/rust/issues/34283, it seems to be true in debug build without optimization. I also remembered that I've encountered stack overflow somewhere due to the same issue. 🤔

wenym1:
> > Basically, in the initial MIR that has not gone through optimization, the function that drives the future state machine will pre-allocate the stack space for all variables in all arms, even though only one of the arms will be reached in runtime, and I guess this is where the warning comes from.
> 
> However, according to the issue [rust-lang/rust#34283](https://github.com/rust-lang/rust/issues/34283), it seems to be true in debug build without optimization. I also remembered that I've encountered stack overflow somewhere due to the same issue. 🤔

Intersting. I tried to print the runtime stack frame size in the way the author of the issue uses.

```
static mut START: Option<u64> = None;

fn set_start() {
    unsafe {
        START = Some(stack_depth());
    }
}

fn reset_start() {
    unsafe {
        START = None;
    }
}

fn print_stack(info: &str) {
    unsafe {
        let stack_start = START.unwrap();
        println!("{} size: {}", info, stack_start - stack_depth());
    }
}

#[inline(never)]
async fn many_huge() {
    print_stack("before");
    huge().await;
    {
        let allocate = [0u32; 60];
        print_stack("middle");
        println!("use {}", allocate[1]);
    }
    let num: u32 = rand::random();
    match num {
        2 => {
            huge2().await;
            print_stack("2");
        }
        1 => {
            huge1().await;
            print_stack("1");
        }
        _ => {
            huge3().await;
            print_stack("3");
        }
    }
    print_stack("after");
}

fn stack_depth() -> u64 {
  let x = 0;
  unsafe {
    let raw: u64 = std::mem::transmute(&x);
    raw
  }
}


#[tokio::main]
async fn main() {
    println!("{}", std::mem::size_of_val(&many_huge()));
    let many_huge = many_huge();
    pin_mut!(many_huge);
    poll_fn(|cx| {
        set_start();
        let ret = many_huge.as_mut().poll(cx);
        reset_start();
        ret
    }).await;
}
```

In debug mode it shows
```
128136
before size: 1025712
middle size: 1025712
use 0
3 size: 1025712
after size: 1025712
```

In release mode it shows
```
128136
before size: 0
middle size: 96
use 0
3 size: 0
after size: 0
```

This suggests that, under debug mode, it indeed pre-allocates all stack space for all variables in the function, even the variables (variable `allocate` in the example code) that won't be captured by the future state machine. This might be a general behavior of rust compiler regardless of whether it's an async function or not. On the contrary, under release mode, it allocates stack space on demand.

@BugenZhao Do you think we should optimize our code for debug mode? At least in this ctl tools that we rarely use in debug mode, I will suggest keeping it simple. Not sure if we should handle the case for code in crutial code path.

BugenZhao:
> At least in this ctl tools that we rarely use in debug mode, I will suggest keeping it simple. Not sure if we should handle the case for code in crutial code path.

+1 for this. We may only optimize the debug build if it affects basic functionality for local testing.

PR #22973, opened 2025-08-23
fix: Add pruning logic to prevent the compactor from incorrectly accessing watermarks.

Li0k:

as title

## Checklist

## Documentation

<details>

</details>


src/storage/src/hummock/compactor/compactor_runner.rs
```diff
@@ -624,7 +624,7 @@ pub async fn compact(
                 );
             }
 
-            compaction_catalog_agent_ref
```
hzxa21: Isn't `acquire_table_ids` always the same as `table_ids_to_be_compacted` if reached here? Filtering `pk_prefix_table_watermarks` and `non_pk_prefix_table_watermarks` via `table_ids_to_be_compacted` seems sufficient to me.

hzxa21 (approved):
Rest LGTM. Thanks for the quick fix during weekend! Really appreciated that.

Let's describe the issue briefly in the PR description as well.

src/storage/src/hummock/compactor/compactor_runner.rs
Li0k: Indeed, they are equal; otherwise, the compactor would abandon this task. Perhaps what you're wondering is why an extra pruning is needed. I just want to take a defensive measure. Functionally, the meta's pruning is sufficient.

PR #22977, opened 2025-08-25
fix: fix snapshot_expiration_timestamp_ms config semantics

Li0k:

Fix the semantics of snapshot_expiration_timestamp_ms. Before this PR, users could only configure an absolute value instead of a relative value, which led to errors. This PR modifies the configuration semantics: it indirectly calculates snapshot_expiration_timestamp_ms by configuring max_age_ms.

## Checklist

## Documentation

<details>

</details>


src/meta/src/manager/iceberg_compaction.rs
```diff
@@ -568,7 +568,7 @@ impl IcebergCompactionManager {
 
         let mut expired_snapshots = tx.expire_snapshot();
 
-        if let Some(expiration_time) = iceberg_config.snapshot_expiration_expire_older_than {
+        if let Some(expiration_time) = snapshot_expiration_timestamp_ms {
```
chenzl25: I think we need to set a default value if `snapshot_expiration_timestamp_ms` is None, since we use `MAX_SNAPSHOT_AGE_MS_DEFAULT` above.

chenzl25 (approved):

chenzl25 (approved):

PR #13190, opened 2023-11-01
feat(sink): support jdbc sink async truncate

wenym1:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://gist.github.com/TennyZhuang/f00be7f16996ea48effb049aa7be4d66#file-rw_cla).

## What's changed and what's your intention?

After https://github.com/risingwavelabs/risingwave/pull/13137

The return type of the `write` method of `SinkWriter` java interface is changed to `bool`, where `true` indicates that the batch is persisted in the external sink and is safe to truncate the log reader.

A new item `BatchWritten` is added to the sink writer response, to indicate that the batch with the batch id has been persisted. The remote sink is refactored to poll the response and truncate asynchronously.

## Checklist

- [ ] I have written necessary rustdoc comments
- [ ] I have added necessary unit tests and integration tests
- [ ] I have added fuzzing tests or opened an issue to track them. (Optional, recommended for new SQL features #7934).
- [ ] My PR contains breaking changes. (If it deprecates some features, please create a tracking issue to remove them in the future).
- [x] All checks passed in `./risedev check` (or alias, `./risedev c`)
- [ ] My PR changes performance-critical code. (Please run macro/micro-benchmarks and show the results.)

- [ ] My PR contains critical fixes that are necessary to be merged into the latest release. (Please check out the [details](https://github.com/risingwavelabs/risingwave/blob/main/CONTRIBUTING.md))

## Documentation

- [ ] My PR needs documentation updates. (Please use the **Release note** section below to summarize the impact on users)

## Release note

If this PR includes changes that directly affect users or other significant modifications relevant to the community, kindly draft a release note to provide a concise summary of these changes. Please prioritize highlighting the impact these changes will have on users.

</details>

PR #22984, opened 2025-08-26
feat(sink): support snowflake and redshift sink auto schema change

xxhZs:

merge dev/cdc-improvent-and-e2e-schema-evolution to main (only snowflake and redshift sink)

## Checklist

## Documentation

<details>

</details>


e2e_test/s3/file_sink.py
```diff
@@ -303,7 +303,7 @@ def _table():
         type = 'append-only',
         force_append_only='true',
         refresh.interval.sec = 1,
-    ) FORMAT PLAIN ENCODE JSON(force_append_only='true');''')
```
hzxa21: why do we remove this? Is it no longer applicable to snowflake sink?

java/connector-node/risingwave-jdbc-runner/src/main/java/com/risingwave/runner/SnowflakeJDBCRunner.java
```diff
@@ -0,0 +1,66 @@
...
+ * limitations under the License.
+ */
+
+package com.risingwave.runner;
+
+import java.sql.*;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class SnowflakeJDBCRunner {
```
hzxa21: Let's rename the class and the file to `JDBCSqlRunner`

java/connector-node/risingwave-sink-jdbc/src/main/java/com/risingwave/connector/JDBCSink.java
```diff
@@ -241,16 +247,17 @@ class JdbcStatements implements AutoCloseable {
         public JdbcStatements(Connection conn, int queryTimeoutSecs) throws SQLException {
             this.queryTimeoutSecs = queryTimeoutSecs;
             this.conn = conn;
+            // Set database and schema for Snowflake connections
+            if (config.getJdbcUrl().toLowerCase().startsWith("jdbc:snowflake")) {
+                setSnowflakeDatabaseAndSchema();
+            }
```
hzxa21: Given that we already have BatchAppendOnlyJDBCSink, the changes here are no longer needed. I will push a commit to remove it in your branch.

src/connector/src/sink/file_sink/s3.rs
```diff
@@ -169,58 +169,3 @@ impl OpendalSinkBackend for S3Sink {
...
-// Wrapper for expanding in the sink config type.
-#[serde_as]
-#[derive(Deserialize, Debug, Clone, WithOptions)]
-pub struct SnowflakeConfig {
-    #[serde(flatten)]
-    pub inner: S3Config,
-}
-
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
-pub struct SnowflakeSink;
```
hzxa21: As we discussed offline, let's keep the original snowflake sink for backward compatibility and name the new snowflake sink `snowflake_v2`.

src/connector/src/sink/snowflake_redshift/snowflake.rs
```diff
@@ -0,0 +1,1115 @@
...
+
+    #[serde(default)]
+    #[serde(rename = "create_table_if_not_exists")]
+    #[serde_as(as = "DisplayFromStr")]
+    pub create_table_if_not_exists: bool,
+
+    #[serde(default)]
+    #[serde(rename = "with_s3")]
+    #[serde_as(as = "DisplayFromStr")]
+    pub with_s3: bool,
```
hzxa21, line 106: Let's default it to `true`.

src/connector/src/sink/snowflake_redshift/redshift.rs
```diff
@@ -0,0 +1,921 @@
...
+
+    #[serde(default = "default_batch_insert_rows")]
+    #[serde(rename = "batch.insert.rows")]
+    #[serde_as(as = "DisplayFromStr")]
+    pub batch_insert_rows: u32,
+
+    #[serde(default)]
+    #[serde(rename = "with_s3")]
+    #[serde_as(as = "DisplayFromStr")]
+    pub with_s3: bool,
```
hzxa21, line 113: let's default it to true.

src/connector/src/sink/snowflake_redshift/redshift.rs
```diff
@@ -0,0 +1,921 @@
...
+    #[serde(rename = "user")]
+    pub username: Option<String>,
+
+    #[serde(rename = "password")]
+    pub password: Option<String>,
+
+    #[serde(rename = "schema")]
+    pub schema: Option<String>,
+
+    #[serde(rename = "target.table.name")]
```
hzxa21: Since iceberg call this `table.name`, we can make it consistent here and call it `table.name` as well.

src/connector/src/sink/snowflake_redshift/snowflake.rs
```diff
@@ -0,0 +1,1115 @@
...
+#[serde_as]
+#[derive(Debug, Clone, Deserialize, WithOptions)]
+pub struct SnowflakeConfig {
+    #[serde(rename = "type")]
+    pub r#type: String,
+
+    #[serde(rename = "intermediate.table.name")]
+    pub snowflake_cdc_table_name: Option<String>,
+
+    #[serde(rename = "target.table.name")]
```
hzxa21: Since iceberg call this `table.name`, we can make it consistent here and call it `table.name` as well.

src/connector/src/sink/remote.rs
```diff
@@ -661,13 +662,11 @@ impl SinkWriter for CoordinatedRemoteSinkWriter {
...
-                    SinkError::Remote(anyhow!(
-                        "get none metadata in commit response for coordinated sink writer"
-                    ))
-                })
-                .map(Some)
+            Ok(Some(rsp.metadata.unwrap_or_else(|| SinkMetadata {
+                metadata: Some(sink_metadata::Metadata::Serialized(SerializedMetadata {
+                    metadata: vec![],
+                })),
+            })))
```
hzxa21: Why do we need this change?

src/meta/src/manager/sink_coordination/handle.rs
```diff
@@ -132,9 +132,6 @@ impl SinkWriterCoordinationHandle {
                             prev_epoch
                         )));
                     }
-                    if request.metadata.is_none() {
-                        return Poll::Ready(Err(anyhow!("empty commit metadata")));
-                    };
```
hzxa21: Why do we remove this?

src/frontend/src/handler/create_sink.rs
```diff
@@ -900,9 +931,6 @@ static CONNECTORS_COMPATIBLE_FORMATS: LazyLock<HashMap<String, HashMap<Format, V
                 FileSink::<S3Sink>::SINK_NAME => hashmap!(
                     Format::Plain => vec![Encode::Parquet, Encode::Json],
                 ),
-                FileSink::<SnowflakeSink>::SINK_NAME => hashmap!(
-                    Format::Plain => vec![Encode::Parquet, Encode::Json],
```
hzxa21: Does it mean that we no longer support parquet format in the intermediate s3?

java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/SinkWriterStreamObserver.java
```diff
@@ -136,18 +136,6 @@ public void onNext(ConnectorServiceProto.SinkWriterStreamRequest sinkTask) {
 
                 currentBatchId = batch.getBatchId();
 
-                if (batchWritten) {
```
hzxa21: why do we remove these?

java/connector-node/risingwave-jdbc-runner/src/main/java/com/risingwave/runner/JDBCSqlRunner.java
```diff
@@ -0,0 +1,66 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class JDBCSqlRunner {
+    private static final Logger LOG = LoggerFactory.getLogger(JDBCSqlRunner.class);
+
+    public static void executeSql(String fullUrl, String[] sqls) throws Exception {
+        Connection connection = null;
+        try {
+            Class.forName("net.snowflake.client.jdbc.SnowflakeDriver");
```
hzxa21, line 29: why do we only need snowflake driver, not redshift driver here?

java/connector-node/risingwave-jdbc-runner/pom.xml
```diff
@@ -0,0 +1,48 @@
...
+            <artifactId>connector-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-text</artifactId>
+        </dependency>
+
+        <!-- https://mvnrepository.com/artifact/net.snowflake/snowflake-jdbc -->
+        <dependency>
+            <groupId>net.snowflake</groupId>
```
hzxa21, line 37: why do we only need to depend on snowflake, not redshift jdbc driver here?

src/connector/src/sink/snowflake_redshift/snowflake.rs
```diff
@@ -0,0 +1,1119 @@
...
+    pub snowflake_target_table_name: Option<String>,
+
+    #[serde(rename = "database")]
+    pub snowflake_database: Option<String>,
+
+    #[serde(rename = "schema")]
+    pub snowflake_schema: Option<String>,
+
+    #[serde(default = "default_schedule")]
+    #[serde(rename = "schedule_seconds")]
```
hzxa21: `write.target.interval.seconds`

src/connector/src/sink/snowflake_redshift/redshift.rs
```diff
@@ -0,0 +1,927 @@
...
+    #[serde(rename = "intermediate.table.name")]
+    pub cdc_table: Option<String>,
+
+    #[serde(default)]
+    #[serde(rename = "create_table_if_not_exists")]
+    #[serde_as(as = "DisplayFromStr")]
+    pub create_table_if_not_exists: bool,
+
+    #[serde(default = "default_schedule")]
+    #[serde(rename = "schedule_seconds")]
```
hzxa21: `write.target.interval.seconds`

src/connector/src/lib.rs
```diff
@@ -71,6 +71,9 @@ pub use with_options::{Get, GetKeyIter, WithOptionsSecResolved, WithPropertiesEx
 mod with_options_test;
 
 pub const AUTO_SCHEMA_CHANGE_KEY: &str = "auto.schema.change";
+pub const SINK_CREATE_TABLE_IF_NOT_EXISTS_KEY: &str = "create_table_if_not_exists";
+pub const SINK_TARGET_TABLE_NAME: &str = "target.table.name";
```
hzxa21: `target.table.name` -> `table.name`

hzxa21 (commented):
Have we tested whether the current PR works in the following scenarios?
1. Append-only snowflake/redshift sink with sink decoupling
2. Append-only snowflake/redshift sink without sink decoupling
3. Upsert snowflake/redshift sink with sink decoupling
4. Upsert snowflake/redshift sink without sink decoupling
5. Append-only snowflake/redshift sink with auto schema change enabled
6. Upsert snowflake/redshift sink with auto schema change enabled

src/rpc_client/src/connector_client.rs
```diff
@@ -55,14 +55,21 @@ impl<REQ: From<SinkWriterStreamRequest>> SinkWriterRequestSender<REQ> {
...
-                msg
-            ))),
+        loop {
+            match self.next_response().await? {
+                SinkWriterStreamResponse {
+                    response: Some(sink_writer_stream_response::Response::Commit(rsp)),
+                } => return Ok(rsp),
+                SinkWriterStreamResponse {
+                    response: Some(sink_writer_stream_response::Response::Batch(_)),
+                } => continue,
```
hzxa21, line 65: I think this is correct. @wenym1 can you double check as well? It seems that we miss to modify this method in https://github.com/risingwavelabs/risingwave/pull/13190

wenym1: #13190 was for jdbc sink, and it uses `RemoteLogSinker`, which does not call this method.

e2e_test/sink/auto_schema_change.slt
```diff
@@ -30,7 +30,7 @@ drop materialized view m1;
 statement ok
 create table t1 (id int primary key, name string, g int  as id + 1);
 
-statement error auto schema change not supported for sink-into-table
+statement error In the case of sinking into table, the 'connector' parameter should not be provided.
```
wenym1: This test ddl was not written correctly. Can you try removing the `connector = 'blackhole',` from the with options?

src/frontend/src/optimizer/plan_node/stream_sink.rs
```diff
@@ -288,6 +289,11 @@ impl StreamSink {
                         })
                         .try_collect::<_, _, RwError>()?
                 }
+            } else if properties.get(CREATE_TABLE_IF_NOT_EXISTS) == Some(&"true".to_owned())
+                && sink_type == SinkType::Upsert
+                && from_properties.is_empty()
```
wenym1: The `from_properties` is actually `downstream_pk`. Shall we rename to the correct one?

src/frontend/src/handler/create_sink.rs
```diff
@@ -184,10 +207,53 @@ pub async fn gen_sink_plan(
...
-                return Err(RwError::from(ErrorCode::InvalidInputSyntax(
-                    "auto schema change not supported for sink-into-table".to_owned(),
-                )));
+            if is_auto_schema_change {
+                if sink_into_table_name.is_some() {
+                    return Err(RwError::from(ErrorCode::InvalidInputSyntax(
+                        "auto schema change not supported for sink-into-table".to_owned(),
+                    )));
+                }
+                match connector.as_str() {
```
wenym1: We have already had logic to check whether the connector support auto schema change in `StreamSink::create`. Any reason to add the check here?

src/connector/src/sink/remote.rs
wenym1: +1

wenym1 (commented):
General sink framework part rest LGTM.

src/connector/src/sink/remote.rs
xxhZs: Because in this PR, the barrier method in the Java section of the Snowflake and Redshift sinks does not necessarily return a valid result.

src/frontend/src/handler/create_sink.rs
xxhZs: cc @hzxa21

hzxa21: I pushed a commit to remove it.

PR #22988, opened 2025-08-26
fix(main cron): 

tabVersion:

https://github.com/risingwavelabs/risingwave/pull/22747 changes the behavior when force_append_only = true, which can make sink out of order, lead to main-cron fails. 

## Checklist

## Documentation

<details>

</details>


xxhZs (approved):

xiangjinwu:
The PR title is misleading. The actual fix is updating the expected output from `"123"` to `123`, while the input remains `123`.

I was wondering how it suddenly fail on Aug 26, or why it was passing before that, given the old expected output does not match input.

So I re-run the test on #21989 where the feature and test first introduced, and the test fails due to actual `123` not matching expected `"123"`:
https://buildkite.com/risingwavelabs/pull-request/builds/81435#019918ad-4d5b-403a-a498-cf6ce4d24d3e

That means, something not tied to the commit (e.g. unpinned dependency) changed and lead to the different running result.

Then I found the clickhouse/clickhouse-server:25.5 image was updated on Aug 26:
https://hub.docker.com/layers/clickhouse/clickhouse-server/25.5/images/sha256-848e3951162c60967fb119658510cabe5a4416d090a562f71c0c6b08e488870d

So it was ClickHouse that mistakenly inputs `123` but outputs `"123"`. Our test was to match its buggy behavior and suddenly they fixed it.

Additional lesson: despite it is `clickhouse/clickhouse-server:25.5`, this is not the same as locking a dependency version in cargo - there is no lock file validating its sha256 remains the same

xiangjinwu:
> Then I found the clickhouse/clickhouse-server:25.5 image was updated on Aug 26

This is true but not the root cause. Whether outputting `"123"` or `123` in json is controlled on the ClickHouse client side, not the server. The testing script always installs the latest client via `curl https://clickhouse.com/ | sh`. This sudden change was due to the release of ClickHouse client 25.8, despite we are still testing with server 25.5

Expect `"123"` and actual `"123"` when using 25.5 client, on the commit of feature intro:
https://buildkite.com/risingwavelabs/main-cron/builds/6841

Expect `123` and actual `"123"` when using 25.5 client, on latest main:
https://buildkite.com/risingwavelabs/main-cron/builds/6843

PR #23003, opened 2025-08-27
feat(mqtt): Add single shared client per mqtt broker for split enumerator

florissmit10:

The MQTTSplitEnumerator used to make a client for each source.  To reduce overhead I used the same approach as the done with the kafka connector 

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/src/source/mqtt/enumerator/mod.rs
+++ b/src/connector/src/source/mqtt/enumerator/mod.rs
@@ -12,49 +12,61 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::sync::Arc;
 use std::sync::atomic::AtomicBool;
+use std::sync::{Arc, LazyLock, Weak};
 
 use async_trait::async_trait;
+use moka::future::Cache as MokaCache;
+use moka::ops::compute::Op;
 use risingwave_common::bail;
-use rumqttc::v5::{ConnectionError, Event, Incoming};
+use rumqttc::v5::{AsyncClient, ConnectionError, Event, EventLoop, Incoming};
 use thiserror_ext::AsReport;
 
 use super::MqttProperties;
 use super::source::MqttSplit;
-use crate::error::ConnectorResult;
+use crate::error::{ConnectorError, ConnectorResult};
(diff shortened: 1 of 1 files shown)


src/connector/src/source/mqtt/enumerator/mod.rs
```diff
@@ -12,50 +12,64 @@
...
+    broker: String,
+    connection_check: Arc<MqttConnectionCheck>,
+}
+
+struct MqttConnectionCheck {
     #[expect(dead_code)]
-    client: rumqttc::v5::AsyncClient,
+    client: AsyncClient,
     connected: Arc<AtomicBool>,
     stopped: Arc<AtomicBool>,
```
tabVersion, line 44: Is it possible to remove state for `stopped`? seems `connected` is enough.

src/connector/src/source/mqtt/enumerator/mod.rs
```diff
@@ -71,36 +85,76 @@ impl SplitEnumerator for MqttSplitEnumerator {
...
         })
     }
 
     async fn list_splits(&mut self) -> ConnectorResult<Vec<MqttSplit>> {
-        if !self.connected.load(std::sync::atomic::Ordering::Relaxed) {
+        if !self.connection_check.is_connected() {
             let start = std::time::Instant::now();
             loop {
-                if self.connected.load(std::sync::atomic::Ordering::Relaxed) {
+                if self.connection_check.is_connected() {
```
tabVersion, line 147: shall we add some tokio sleep inside the loop?

florissmit10: There is a 500 ms sleep on line 154 that gets executed when the client is not connected.
When the client is connected the entire loop stops, and is not started again until the next time list_splits is called.

src/connector/src/source/mqtt/enumerator/mod.rs
florissmit10: Stopped is used to stop the loop running in the background. (see line 65)
When MqttConnectionCheck is dropped(see line 91), this is set to false, which stops the loop.
Happy to hear any other way in which this can be achieved.

PR #22510, opened 2025-07-04
fix: make pulsar source exactly once

tabVersion:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

close #22486 

## Checklist

- [ ] I have written necessary rustdoc comments.
- [ ]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  I have checked the [Release Timeline](https://github.com/risingwavelabs/rw-commits-history/blob/main/release_timeline.md) and [Currently Supported Versions](https://docs.risingwave.com/changelog/release-support-policy#support-end-dates-for-recent-releases) to determine which release branches I need to cherry-pick this PR into. 

## Documentation

- [ ]  My PR needs documentation updates. 

<details>
<summary><b>Release note</b></summary>

</details>

PR #23010, opened 2025-08-28
fix: Check ledger ID when considering `already_read_offset` in Pulsar Source Reader

rkrishn7:

This PR adds an additional check to consider the message's ledger ID when deciding whether to ignore a given message, based on the already stored latest consumed offset.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/src/source/pulsar/source/reader.rs
+++ b/src/connector/src/source/pulsar/source/reader.rs
@@ -82,7 +82,9 @@ impl SplitReader for PulsarSplitReader {
 
 /// Filter offset for pulsar messages
 #[derive(Debug, Clone)]
-pub struct PulsarFilterOffset {
+struct PulsarFilterOffset {
+    /// Ledger ID of the message
+    pub ledger_id: u64,
     /// Entry ID of the message
     pub entry_id: u64,
     /// Batch index of the message, if any
@@ -209,6 +211,7 @@ impl SplitReader for PulsarBrokerReader {
                 } else {
                     let start_message_id = parse_message_id(m.as_str())?;
                     already_read_offset = Some(PulsarFilterOffset {
+                        ledger_id: start_message_id.ledger_id,
                         entry_id: start_message_id.entry_id,
                         batch_index: start_message_id.batch_index,
                     });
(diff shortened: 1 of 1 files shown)


src/connector/src/source/pulsar/source/reader.rs
```diff
@@ -261,15 +264,21 @@ impl PulsarBrokerReader {
...
                 {
                     let message_id = msg.message_id();
 
-                    // for most case, we only compare `entry_id`
-                    // but for batch message, we need to compare `batch_index` if the `entry_id` is the same
-                    if message_id.entry_id <= entry_id && message_id.batch_index <= batch_index {
+                    // If we have a previously stored offset, check that the messages we're consuming
+                    // appear after. Note that entry IDs and batch indexes are only monotonically increasing
+                    // within a BookKeeper ledger, so we must check that the ledger IDs match. If they don't match,
+                    // messages are coming from a new ledger and we can safely ignore the stored offset.
```
BugenZhao, line 277: Is it possible for `message_id.ledger_id` to be less than the `ledger_id` we stored? If so, it seems we will still reset the `already_read_offset`, causing some messages to not be filtered out.

rkrishn7: Ledgers are sealed once they are closed. So, if our previously stored ledger ID is different than that of the incoming message, I think we can safely disregard it since we know that ledger is closing or has been closed.

Even though we store the offset and try to resume consumption from the specified offset, it doesn't guarantee that we will begin consuming from that offset. The offset we are trying to consume from may be deleted (e.g. due to retention policies). In that case, Pulsar will advance the consumer position to the earliest _available_ offset, which may be on a different ledger.

src/connector/src/source/pulsar/source/reader.rs
```diff
@@ -82,7 +82,9 @@ impl SplitReader for PulsarSplitReader {
 
 /// Filter offset for pulsar messages
 #[derive(Debug, Clone)]
-pub struct PulsarFilterOffset {
+struct PulsarFilterOffset {
+    /// Ledger ID of the message
+    pub ledger_id: u64,
```
tabVersion, line 87: PulsarFilterOffset is not persisted, so it is safe not wrapping with an `Option`.

tabVersion:
the issue introduced by https://github.com/risingwavelabs/risingwave/pull/22510 and cherry-pick to v2.5.

PR #23017, opened 2025-08-29
init

Chiro11:

## Checklist

## Documentation

<details>

</details>


Chiro11:
* **#23017** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23017?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23017?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#22920** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22920?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#22917** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22917?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

Chiro11:
@wenym1 @yezizp2012 @shanicky This PR is ready, please review.

src/meta/src/controller/catalog/mod.rs
```diff
@@ -417,6 +422,124 @@ impl CatalogController {
...
+            fields: union_node.get_fields().clone(),
+            node_body: Some(NodeBody::UpstreamSinkUnion(Box::new(upstream_sink_union))),
+        };
+        union_node.input.push(stream_node);
+
+        true
+    }
+
+    // Currently it is only used to refill table fragments with upstream sinks.
+    pub async fn migrate_legacy_fragments(&self) -> MetaResult<()> {
```
yezizp2012, line 495: Is it possible to add a migration to do the rewrite logic for better performance? Ideally we only need to do it once. The only dependency is `visit_stream_node_cont_mut`, I think it's okay to import it for model migration.

yezizp2012 (commented):
Rest LGTM.

src/meta/src/controller/catalog/mod.rs
```diff
@@ -417,6 +422,124 @@ impl CatalogController {
...
+        let table_mview_fragments: Vec<(FragmentId, StreamNode)> = Fragment::find()
+            .select_only()
+            .columns(vec![
+                fragment::Column::FragmentId,
+                fragment::Column::StreamNode,
+            ])
+            .filter(
+                fragment::Column::JobId
+                    .is_in(all_table_ids.clone())
+                    .and(FragmentTypeMask::intersects(FragmentTypeFlag::Mview)),
```
wenym1, line 509: I am thinking, shall we add a new `FragmentTypeFlag` to mark whether a fragment contains `UpstreamSinkUnion`, so that we can only do the transformation on table not migrated yet. In current implementation, we have to load all tables and deserialize the stream node in every meta node restarts, even though we may have already migrated the fragment of the tables.

src/meta/src/controller/catalog/mod.rs
```diff
@@ -417,6 +422,124 @@ impl CatalogController {
...
+                .unwrap(),
+            NodeBody::UpstreamSinkUnion(_)
+        ) {
+            // Skip the table-fragment that has been migrated.
+            return false;
+        }
+        let mut upstream_sink_count = 0;
+        for union_input in union_node.get_input().iter().rev() {
+            let mut is_sink_into = false;
+            if let NodeBody::Project(project) = union_input.get_node_body().unwrap() {
```
wenym1, line 451: IIRC, in order version, we only have a `Merge` in sink into table, instead of `Project+Merge`. @shanicky @yezizp2012 Can you help confirm it?

src/meta/src/barrier/context/recovery.rs
```diff
@@ -323,6 +332,15 @@ impl GlobalBarrierWorkerContextImpl {
         {
             {
                 {
+                    let recovery_reason = self.recovery_reason();
```
wenym1, line 335: Seems more straightforward to do the migration during meta bootstrap in `start_service_as_election_leader` before spawning the barrier worker, so that we don't have to hack into barrier worker in every recovery loop.

We had a similar logic for backward compatibility migration to call `may_fill_backward_state_table_info`.

src/meta/src/controller/catalog/mod.rs
Chiro11: Do you mean to do the migration through a file like `m2025xxx.rs`? I see that it currently only supports some traits and operations predefined by SeaORM. It seems quite difficult to implement that.

Chiro11: Let me have a look.

src/meta/src/controller/catalog/mod.rs
```diff
@@ -88,6 +90,9 @@ use crate::rpc::ddl_controller::DropMode;
 use crate::telemetry::{MetaTelemetryJobDesc, report_event};
 use crate::{MetaError, MetaResult};
 
+// A value different from other operator_ids, used for new operator when reloading a table from an older version.
+const UPSTREAM_SINK_UNION_OPERATOR_ID: u64 = 1 << 16;
```
wenym1, line 94: When creating the stream plan of a table, `StreamMaterialize` is the last stream node we create. Does it mean that the operator_id of the Materialized node is the largest operator_id? And if so, can we use the max_operator_id+1 in the fragment of Materialized node as the operator_id of the new UpstreamSinkUnion node.

wenym1 (commented):
Rest LGTM.

May need to inspect the code in earlier version of RisingWave since we started support sink-into-table, to see if the migration works for a table created in any previous version.

src/meta/src/controller/catalog/mod.rs
yezizp2012: +1, if it's too complexed to do it in model migration, adding a new `FragmentTypeFlag` can avoid this performance issue.

src/meta/src/controller/catalog/mod.rs
yezizp2012: Not that sure, @shanicky please help to confirm it.

src/meta/src/controller/catalog/mod.rs
```diff
@@ -417,6 +422,124 @@ impl CatalogController {
...
+            return false;
+        }
+        let mut upstream_sink_count = 0;
+        for union_input in union_node.get_input().iter().rev() {
+            let mut is_sink_into = false;
+            if let NodeBody::Project(project) = union_input.get_node_body().unwrap() {
+                let project_input = union_input.get_input().first().unwrap();
+                if let NodeBody::Merge(merge) = project_input.get_node_body().unwrap() {
+                    // Operators created in older versions should meet these conditions.
+                    assert!(
```
wenym1, line 455: This assertion doesn't look like a strong requirement for the correctness of migration. Shall we change to only assert it under debug mode and print a warning log under release mode?

wenym1:
Discussed offline with @Chiro11 about a new way of migration.

Table is an important and common object in our system. It seems that the migration logic in this PR is too bao'li, and any bug in the migration logic, or any unexpected historical table fragment pattern, can possily corrupt the streaming graph and make the cluster enter unrecoverable problematic state. This fragileness is because that we alter all tables altogether, and the update is performed in an incremental way, and the streaming graph gets altered without actually inject and collect a barrier to validate the streaming graph at runtime.

Therefore, it might be safer to do the migration table by table in a rolling way. Our objective is
* the migration can be disabled or paused at users' will, which means, without migration, in the middle of migration, and after migration, and cluster can still run healthily.
* let the new fragment go through a barrier injection and collection, so that it can be validated at runtime
* we will do the migration in an imtermediate version (likely to be release-2.7). We will provide a system table or view, to inspect whether all tables have migrated to the new fragment, and in the release note of the next release (release-2.8), ask the users to inspect this system table or view to ensure that all tables have been migrated, before doing the upgrade.

Currently, the implementation of alter table is to regenerate the table fragment directly from source and sink, and we can possibly reuse its logic to support migrating the fragment of the existing table in a more robust way. Implementation will be like
1. during recovery, we only fill in upstream sink info for table fragment with `UpstreamSinkUnion` in it. For the table fragment not migrated yet, we use the existing fragment.
2. forbid create/drop sink into table not migrated yet (table whose fragment does not have `UpstreamSinkUnion`)
3. in frontend, we add an rpc, to support trivial alter table that neither add nor drop column, but only regenerate the table fragment. Most logic of alter table can be reused, and won't involve much unnecessary efforts.
4. in meta node, during meta node bootstrap, we load the unmigrated table from meta store, and spawn a task to asychronously migrate the tables one by one. Such logic can be disabled by a meta node option for safety.

@yezizp2012 @shanicky @hzxa21 Do you have any comment?

wenym1:
A PR to support 1)

https://github.com/risingwavelabs/risingwave/pull/23125

Chiro11:
> Table is an important and common object in our system. It seems that the migration logic in this PR is too bao'li, and any bug in the migration logic, or any unexpected historical table fragment pattern, can possily corrupt the streaming graph and make the cluster enter unrecoverable problematic state. This fragileness is because that we alter all tables altogether, and the update is performed in an incremental way, and the streaming graph gets altered without actually inject and collect a barrier to validate the streaming graph at runtime.

Beyond safety, a fundamental consideration is that, if we need to support the migration of streaming graph for all tables (including those with uncertain patterns), the most reasonable way is to let frontend re-generate table fragments. This leads to `replace_table` process. While other methods might also ensure data integrity or cluster healthy, they cannot guarantee that legacy tables will always be migratable, because we are not certain of their graph patterns.

hzxa21:
> we will do the migration in an imtermediate version (likely to be release-2.7). 

Is the migration process in `release-2.7` fully automated?

> We will provide a system table or view, to inspect whether all tables have migrated to the new fragment, and in the release note of the next release (release-2.8), ask the users to inspect this system table or view to ensure that all tables have been migrated, before doing the upgrade.

If we can provide a system table/view here for user to check and the plan is to disallow 2.8 upgrade, we should just check the system table/view on 2.8 startup and fail the upgrade if not all tables are migrated.

BugenZhao:
> Currently, the implementation of alter table is to regenerate the table fragment directly from source and sink, and we can possibly reuse its logic to support migrating the fragment of the existing table in a more robust way.

+1 for this. Before we perform automated mass migration, I think we can also provide a SQL command to manually trigger rebuilding and replacing the table plan without definition change.

wenym1:
> Is the migration process in release-2.7 fully automated?

> > Currently, the implementation of alter table is to regenerate the table fragment directly from source and sink, and we can possibly reuse its logic to support migrating the fragment of the existing table in a more robust way.
> 
> +1 for this. Before we perform automated mass migration, I think we can also provide a SQL command to manually trigger rebuilding and replacing the table plan without definition change.

A SQL command is not a must-have, because the migration is automated. The meta node will load all tables that should be migrated, and then asynchronously call the frontend rpc to trigger the alter one by one. We can expect that the migration won't take too long, so it's not a must to provide a SQL command to explicitly trigger the migration of a specific table. We can ban any create/drop sink to table not migrated yet to ensure integrity.

yezizp2012:
> 1. during recovery, we only fill in upstream sink info for table fragment with UpstreamSinkUnion in it. For the table fragment not migrated yet, we use the existing fragment.
> 2. forbid create/drop sink into table not migrated yet (table whose fragment does not have UpstreamSinkUnion)
> 3. in frontend, we add an rpc, to support trivial alter table that neither add nor drop column, but only regenerate the table fragment. Most logic of alter table can be reused, and won't involve much unnecessary efforts.
> 4. in meta node, during meta node bootstrap, we load the unmigrated table from meta store, and spawn a task to asychronously migrate the tables one by one. Such logic can be disabled by a meta node option for safety.

The implementation LGTM.

> we will do the migration in an imtermediate version (likely to be release-2.7). We will provide a system table or view, to inspect whether all tables have migrated to the new fragment, and in the release note of the next release (release-2.8), ask the users to inspect this system table or view to ensure that all tables have been migrated, before doing the upgrade.

It feels odd to expose the system catalog in this way; I prefer not to involve users. How about checking for any unmigrated tables during the upgrade to version 2.7, then if any are found, pause the bootstrap, then perform the mentioned actions above. The regeneration process can theoretically run in parallel, while barriers can be sequentially issued without collecting the previous one. This should be a quick operation, and validation can also be performed at runtime.

If users skip 2.7 and upgrade directly to 2.8 encountering any unmigrated tables, we should panic and suggest users to upgrade to version 2.7 firstly.

wenym1:
> How about checking for any unmigrated tables during the upgrade to version 2.7, then if any are found, pause the bootstrap, then perform the mentioned actions above.

I think it's not a good idea to block meta node bootstrap by the migration. The fragment regeneration should go through frontend, and the barrier injection and collection will go through CN. If we block meta bootstrap, the whole cluster will have not started yet, we are not likely to do the migration before the whole cluster starts working. 

To be clear, the proposed behavior in each version defined will be like:

in release-2.7:
* allow unmigrated tables in both meta node bootstrap and recovery
* meta node asynchronously trigger frontend rpc to migrate the unmigrated tables 

in release-2.8:
* when seeing unmigrated tables, panic during meta node bootstrap, or report error during recovery
* can remove the code of migration

The system table and view is only more like a pre-check before upgrade to release-2.8 so that the maintainer can be aware of whether upgrading to release-2.8 will panic or not, to avoid a potential upgrade failure beforehand.

PR #23026, opened 2025-09-01
chore: move to bitnamilegacy (kafka, nats-cli)

tabVersion:

More details:

https://github.com/risingwavelabs/risingwave/pull/23026#issuecomment-3241422268

## Checklist

## Documentation

<details>

</details>


integration_tests/iceberg-sink/Dockerfile
```diff
@@ -17,7 +17,7 @@ USER airflow
 RUN pip install --no-cache-dir "apache-airflow==${AIRFLOW_VERSION}" apache-airflow-providers-apache-spark==4.1.1 pyspark==3.3.1
 
 
-FROM bitnami/spark:3.3 as spark
+FROM bitnamilegacy/spark:3.3 as spark
```
tabVersion, line 20: @xxhZs @Li0k  plz help check if the sink can still work

BugenZhao:
Can you note down more context in PR body?

It appears that images under `bitnamilegacy` will no longer receive updates. Therefore, this may only be a temporary workaround, and we may still need to migrate to other vendors in the future.

tabVersion:
> Can you note down more context in PR body?
> 
> It appears that images under `bitnamilegacy` will no longer receive updates. Therefore, this may only be a temporary workaround, and we may still need to migrate to other vendors in the future.

Thanks for the concern! However, after analyzing our specific use cases, I believe this migration to `bitnamilegacy` is actually a reasonable long-term solution rather than just a temporary workaround.

Why bitnamilegacy is suitable for our needs:

* Kafka SASL Testing: We primarily use the Kafka image to test SASL authentication functionality, which is a core feature of Kafka. The protocol is unlikely to have breaking changes at this level, so using bitnamilegacy poses minimal risk to functionality.
* NATS CLI: The NATS project doesn't provide an official CLI image, so maintaining our own would add unnecessary maintenance burden. Given that NATS wire format changes are unlikely, using the bitnamilegacy version is currently the best option.
* Overall: We're using relatively basic features that are unlikely to change significantly in the future.
* Spark: test owner @xxhZs @Li0k  please evaluate if needs to migrate to other vendors. 

Long-term maintenance strategy:

* If we need new features in the future, we can smoothly transition to other solutions
* For current testing requirements, `bitnamilegacy` fully meets our needs
* This avoids the complexity of maintaining custom images

Therefore, I believe this migration is not just a temporary fix but a reasonable choice based on our specific usage patterns. The stability of the core protocols we're testing makes bitnamilegacy a viable long-term solution for our integration testing needs.

> `but older versions and many applications will be removed from the free tier and moved to the [Bitnami Legacy repository](https://hub.docker.com/u/bitnamilegacy)`
> 
> quote from:
>
> Q: Are Bitnami’s free images going away?
A: Not entirely. Bitnami will continue to offer a limited subset of free, latest-version images intended for development use. These images will be more hardened and secure, but older versions and many applications will be removed from the free tier and moved to the [Bitnami Legacy repository](https://hub.docker.com/u/bitnamilegacy). You can browse the images available for free at [hub.docker.com/u/bitnamisecure](https://hub.docker.com/u/bitnamisecure) until August 28th, after which the main Bitnami registry will have the same content.

I think `bitnamilegacy` will be available for a relatively long time.

BugenZhao:
> * Kafka SASL Testing: We primarily use the Kafka image to test SASL authentication functionality, which is a core feature of Kafka.

Just FYI, we also use Confluent's Kafka image in RiseDev. Maybe it's still beneficial to migrate to this image to simplify maintenance, if there's bandwidth.

https://github.com/risingwavelabs/risingwave/blob/1b3001f26aa5929545d0506e40cad66e95c15039/risedev.yml#L1306-L1307

> * NATS CLI: The NATS project doesn't provide an official CLI image, so maintaining our own would add unnecessary maintenance burden.

I think this one is an alternative. Also it's official: https://hub.docker.com/r/natsio/nats-box

tabVersion:
> > * Kafka SASL Testing: We primarily use the Kafka image to test SASL authentication functionality, which is a core feature of Kafka.
> 
> Just FYI, we also use Confluent's Kafka image in RiseDev. Maybe it's still beneficial to migrate to this image to simplify maintenance, if there's bandwidth.
> 
> https://github.com/risingwavelabs/risingwave/blob/1b3001f26aa5929545d0506e40cad66e95c15039/risedev.yml#L1306-L1307

[70144c2](https://github.com/risingwavelabs/risingwave/pull/23026/commits/70144c2a19eb6f5f363a5e98f068f78a0bd9c146) [test pipeline](https://buildkite.com/risingwavelabs/pull-request/builds/80972#019905eb-6709-497c-8c33-6584d9cf6d32)
tried changing to `confluentinc/cp-kafka:7.6.1` but failed to setup sasl auth as expected. Bitnami has some tricks simplifying setting up auth inside docker.

PR #23029, opened 2025-09-01
feat: remove license check on basic iceberg compaction

hzxa21:

As titled. Based on our product discussion, we decided to remove the license check on basic iceberg compaction functionalities and make it available for community users. Advanced compaction strategies will be protected by license guard instead.

## Checklist

## Documentation

<details>

</details>


chenzl25 (commented):
I think we need to remove the check from iceberg sink as well. https://github.com/risingwavelabs/risingwave/blob/60f1561d1a4787a9c54e31a20f9efa40f1bb400e/src/connector/src/sink/iceberg/mod.rs#L578-L582

src/frontend/src/handler/create_table.rs
```diff
@@ -1784,13 +1782,7 @@ pub async fn create_iceberg_engine_table(
             .as_mut()
             .map(|x| x.with_properties.remove("enable_compaction"));
     } else {
-        sink_with.insert(
-            ENABLE_COMPACTION.to_owned(),
-            risingwave_common::license::Feature::IcebergCompaction
```
hzxa21, line 1789: This means we will enable compaction by default for iceberg table engine, right? @chenzl25 @Li0k

chenzl25: yes

PR #23038, opened 2025-09-01
fix(dashboard): fix fragment to relation backpressure edge mapping

kwannoel:

Previously we only mapped fragment backpressure edges if they are between stream scan (input) and mv/table/index (output).

However, this fails to consider `sink into table`. This is because the input for sink into table to the destination table is not `stream scan`, it's currently a merge and project.

In this PR we directly do a `fragment -> relation` mapping, rather than maintain two separate maps for input and output.

On the frontend, we can simply filter out intra-node mappings, by checking whether after performing the `fragment->relation` mapping, the input and output are the same. If they are from the same relation, it's an intra node mapping.

Otherwise, they are from different relations, and we can preserve the backpressure edge, to use when constructing relation-relation backpressure.

This relies on a key fact, that between any two relations, there is not more than one fragment edge. Each relation only has a single output. So this must be true.

A second scenario is that the same fragment outputs twice. This is highly unlikely because of subplan sharing. However, if this really happens, our prometheus query groups the metrics by `fragment_id, downstream_fragment_id`, so all the backpressure metrics will get aggregated.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/dashboard/pages/relation_graph.tsx
+++ b/dashboard/pages/relation_graph.tsx
@@ -126,16 +126,19 @@ export default function StreamingGraph() {
       if (!fragmentToRelationMap) {
         return new Map<string, ChannelDeltaStats>()
       }
-      let inMap = fragmentToRelationMap.inMap
-      let outMap = fragmentToRelationMap.outMap
+      let mapping = fragmentToRelationMap.fragmentToRelationMap
       if (channelStats) {
         let map = new Map<string, ChannelDeltaStats>()
         for (const [key, stats] of channelStats) {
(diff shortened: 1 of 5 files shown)


kwannoel:
Screenshot 2025-09-01 at 8 59 46 PM

After this PR, you can see it works well. `s1 (sink) -> t_dest (into table)` has metrics.

BugenZhao (approved):
LGTM

chenzl25 (approved):

wenym1 (approved):

PR #23045, opened 2025-09-02
feat(iceberg): add docker compose with lakekeeper

chenzl25:

- Add a docker compose with the lakekeeper component. A default warehouse would be created in the lakekeeper bootstrap step with minio as the storage.
- Doc https://www.notion.so/risingwave-labs/RisingWave-with-a-hosted-rest-catalog-262f0d69cb76807bbed5ebea7b33a450?source=copy_link

## Checklist

## Documentation

<details>

- Here is the doc of how to use this docker compose. https://www.notion.so/risingwave-labs/RisingWave-with-a-hosted-rest-catalog-262f0d69cb76807bbed5ebea7b33a450?source=copy_link

</details>

### DIFF
--- a/docker/docker-compose-with-lakekeeper.yml
+++ b/docker/docker-compose-with-lakekeeper.yml
@@ -1,6 +1,6 @@
 ---
 x-image: &image
-  image: ${RW_IMAGE:-risingwavelabs/risingwave:v2.4.0}
+  image: ${RW_IMAGE:-risingwavelabs/risingwave:v2.5.1}
 
 x-lakekeeper-image: &lakekeeper-image
   image: ${LAKEKEEPER__SERVER_IMAGE:-quay.io/lakekeeper/catalog:latest-main}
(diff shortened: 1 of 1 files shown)


docker/docker-compose-with-lakekeeper.yml
```diff
@@ -0,0 +1,596 @@
...
+  prometheus-0:
+    external: false
+  message_queue:
+    external: false
+  lakekeeper-db:
+    external: false
+
+networks:
+  default:
+    driver: bridge
```
hzxa21, line 596: Why do we need this? I don't see we enable bridge mode in other docker compose files.

chenzl25: The bridge actually is the default networks type if we don't specify it. https://arc.net/l/quote/ontfwara

Li0k (approved):

PR #23047, opened 2025-09-02
feat: treat empty(zero-field) struct as alterable

### LINKED ISSUE #23050: bug: zero-field struct may collapse to null in column-aware encoding
### Describe the bug

As of now, our sqlparser does not allow creating a struct type with zero fields. `struct<>` is considered invalid syntax.

However, such type can still be created via `create table as` with `row()`. When this happens, the column-aware encoding would encode such value same as `null` - zero bytes written.

Note that value encoding can still differentiate `row()` from `null`. See the reproduction below for a demonstration of both cases.

### Error message/log

```text
No error reported. Value silently change.

 row 
-----
 
 
 ()
(3 rows)
```

### To Reproduce

```sql
create table t as select row(); -- persisted using column-aware encoding, backfilled as null
insert into t values (row()); -- persisted using column-aware encoding, backfilled as null
create materialized view mv as select * from t;
insert into t values (row()); -- stream to mv using value encoding

select * from mv;
```

### Expected behavior

```
 row 
-----
 ()
 ()
 ()
(3 rows)
```

_No response_

### The version of RisingWave

v2.5.1

### Additional context

_No response_

BugenZhao:

When introducing support for `ALTER TYPE` for columns with composite (`struct`) types, we introduced a new field `field_ids` for `struct` data type (https://github.com/risingwavelabs/risingwave/pull/20620), to differentiate between the old and new encoding.

However, by using `repeated` field in protobuf, we cannot distinguish between "unset" (`None`) and "set with empty list" (`Some([])`), which leads to some confusion when handling empty struct. We didn't treat this a problem, as we don't provide official support for empty struct since our parser doesn't support parsing something like `struct<>`. However, it appears that users can still declare a column with an empty struct type by mapping the schema, such as from protobuf, where this can also be very useful to be an arm of `oneof`.

----

This PR aimed to address such subtlety by trying to treat it alterable. Typically, this will cause a breaking change, as the format will be changed from old column-aware encoding (for non-alterable types) to the new recursive column-aware encoding (for alterable types). In other words, we used to encode empty struct column with value encoding, while we are going to recursively construct a column-aware encoder for the struct now.

However, we realized that there's actually a bug (#23050) that, in old column-aware encoding, we encoded empty struct into an empty slice, which is problematic and make it indistinguishable from `NULL`! Therefore, all previous empty struct values are converted to `NULL` once it's written into a table, meaning that an empty struct value never actually exists in storage! As a result, we can safely perform the encoding change, and this also fixes #23050.

Note that this bug doesn't exist in operators' state tables or MV , since those tables do not support schema change thus are still using the value encoding. Fortunately, it doesn't affect this PR, since whether empty struct is alterable only affects how it's encoded with column-aware encoding.

---

In future PRs we may add support for parsing `struct<>`, as it's necessary when purifying a column mapped from an empty protobuf message.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/e2e_test/ddl/alter_table_column_type/empty_struct.slt
+++ b/e2e_test/ddl/alter_table_column_type/empty_struct.slt
@@ -1,0 +1,68 @@
+statement ok
+set RW_IMPLICIT_FLUSH to true;
+
+# Currently, parser does not allow the syntax for zero-field struct type (`struct<>`).
+# `CREATE TABLE AS` is a workaround.
+statement ok
+create table t as select row() as v;
+
+query T
+select v from t;
+----
+()
+
+statement ok
+alter table t alter column v type struct<a int>;
+
+query T
+select v from t;
+----
+()
+
+statement ok
+insert into t values (row(1));
+
+query T rowsort
+select v from t;
+----
+()
+(1)
+
+statement ok
+create materialized view mv as select (v).a from t;
+
+query I rowsort
+select a from mv;
+----
+1
+NULL
+
+statement ok
+alter table t alter column v type struct<a int, b varchar>;
+
+query T rowsort
+select v from t;
+----
+(,)
+(1,)
+
+statement ok
+insert into t values (row(2, 'hello'));
+
+query T rowsort
+select v from t;
+----
+(,)
+(1,)
+(2,hello)
+
+query I rowsort
+select a from mv;
+----
+1
+2
+NULL
+
+
+statement ok
+drop table t cascade;
--- a/src/common/src/array/struct_array.rs
+++ b/src/common/src/array/struct_array.rs
@@ -334,6 +334,11 @@ impl StructValue {
         }
     }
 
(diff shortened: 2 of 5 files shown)


xiangjinwu (dismissed):

xiangjinwu (approved):

BugenZhao:
* **#23099** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23099?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#23047** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23047?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23047?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #23048, opened 2025-09-02
refactor(pulsar): replace with reader api to avoid backlog on broker

tabVersion:

## Checklist

## Documentation

<details>

</details>


rkrishn7:
Thank you for this!

I think using reader is a solid approach. Under the hood, reader is implemented with a non durable cursor and ack`ing is baked in to the `futures::Stream` impl ([ref](https://github.com/streamnative/pulsar-rs/blob/master/src/reader.rs)).

I think one potential downside here is that auto-acking upon message delivery may release messages from storage. If there's a failure and we need to reprocess messages, they might not exist when we start back up and Pulsar will silently advance our cursor to the earliest available offset.

I think ack'ing after checkpoint commit might be a bit nicer. I've added some thoughts [here](https://github.com/risingwavelabs/risingwave/issues/23057).

PR #23055, opened 2025-09-02
fix: ALTER SOURCE CONNECTOR should work for non-shared source

tabVersion:

introduced in https://github.com/risingwavelabs/risingwave/pull/20780 starting from v2.5.0

## Checklist

## Documentation

<details>

</details>


src/meta/src/controller/streaming_job.rs
```diff
@@ -2511,7 +2514,7 @@ where
         })
         .collect_vec();
     assert!(
-        !fragments.is_empty(),
+        !fragments.is_empty() && is_shared_source, /* shared source should be used by at least one fragment */
```
zwang28: I think the fragments that uses the nonshared source need be updated as well?

tabVersion: yes, it utilize the same filter condition as shared source. 
If there is a downstream mv, then the `alter_stream_node_fn` can hit one or more fragments. If not, it gets an empty vec. 
The difference is that shared source MUST have at least one fragment, but non-shared source does not.

zwang28: But in `update_connector_props_fragments` fragments are found by `job_id`, which won't match that of downstream MV/SINK.
i.e. `filter(fragment::Column::JobId.eq(job_id))`

e2e_test/source_inline/kafka/alter/alter_connector_props.slt
```diff
@@ -84,6 +109,25 @@ select count(*)  from rw_sources where definition ILIKE '%secret s2%' and name =
 ----
 1
 
+
+query I
+select count(*)  from rw_sources where definition ILIKE '%secret s2%' and name = 'plain_students_source';
+----
+1
```
tabVersion, line 116: @zwang28 you can check the test that the non-shared source, not running, i.e. no fragment, is also updated.

src/meta/src/controller/streaming_job.rs
tabVersion: You are right, still need to find the streaming jobs using the source.

zwang28 (approved):

PR #23067, opened 2025-09-03
fix: hnsw off-by-one in insert descent and add visitation tests

### LINKED ISSUE #23065: HNSW is off-by-one in level descent (count vs index confusion)
### Describe the bug

While writing tests for the HNSW implementation in hnsw.rs, I discovered an off-by-one bug caused by mixing up level count (`len() == L+1`) and level index (`0..=L`).

**Details**
- `VectorHnswNode::level()` returns the count of levels, not the highest index.
- `HnswGraph::node_level()` does the same.
- However, `search_layer()` expects a level index (0..=L).

The following `insert_graph()` shows the issue:
```
let entrypoint_level = graph.nodes[entrypoint_index].level(); // count
let mut curr_level = entrypoint_level;
while curr_level > node.level() + 1 {  // <-- off by one
    curr_level -= 1;                   // becomes index
    entrypoints = search_layer(..., curr_level, ...)?;
}
```

This causes the algorithm to skip one level during the first descent, leading to degraded recall.

**Fix**
- Correct the logic by explicitly separating count and index
- Renamed methods (`level()` → `num_levels()`, `node_level()` → `node_num_levels()`) to make intent explicit.
- Adjusted loops in `insert_graph()` and `nearest()` to consistently use `top_index = count - 1`.

**Impact**
- Currently an entire intermediate layer is skipped during insertion, reducing connectivity and recall.
- A fix will ensure all expected layers are visited (entry_top .. node_top+1 for first descent, node_top .. 0 for second phase).

### Error message/log

```text

```

### To Reproduce

_No response_

### Expected behavior

_No response_

_No response_

### The version of RisingWave

_No response_

### Additional context

_No response_

ChrisWrenDev:

Issue: resolve #23065

The HNSW implementation in `hnsw.rs` had an off-by-one bug due to confusion between:
- Level count: number of levels in a node (`len() == L+1`)
- Level index: indices from `0..=L`

`VectorHnswNode::level()` (and `HnswGraph::node_level()`) returned the count, but this value was directly used as if it were the highest index. As a result, during insertion the first descent phase skipped an entire level (`node_top + 1`), reducing graph connectivity and degrading recall.

**What’s changed**
Renamed methods for clarity:
- `VectorHnswNode::level()` → `num_levels()`
- `HnswGraph::node_level()` → `node_num_levels()`

Fixed descent logic:
- First phase: walk from `entry_top` down to (`node_top + 1`) inclusive.
- Second phase: walk from `min(entry_top, node_top)` down to ground (0).

Updated `nearest()` to consistently use `num_levels() - 1` as the top index.

Added explicit doc comments clarifying “count vs. index”.

Added tests which confirm the correct sequence of visited levels:
- `hnsw_insert_graph_visits_expected_upper_layers`
- `hnsw_nearest_visits_expected_upper_layers`
- `hnsw_vector_hnsw_node_level_returns_count`
- `hnsw_graph_node_level_returns_count`

**Impact**
- Prevents skipping of intermediate levels during insertion.
- Ensures correct connectivity and recall.
- Makes the API less error-prone by clarifying semantics.

**Notes**
This change only affects insertion/descent logic and stats accounting, query semantics are unchanged except for improved recall.

## Checklist

## Documentation

<details>

</details>


src/storage/src/vector/hnsw.rs
```diff
@@ -56,7 +57,7 @@ fn gen_level(options: &HnswBuilderOptions, rng: &mut impl Rng) -> usize {
 
 pub(crate) fn new_node(options: &HnswBuilderOptions, rng: &mut impl Rng) -> VectorHnswNode {
     let level = gen_level(options, rng);
-    let mut level_neighbours = Vec::with_capacity(level);
+    let mut level_neighbours = Vec::with_capacity(level + 1);
```
ChrisWrenDev, line 60: A node with top index `L` has `L+1` levels (`0..=L`), so the vec capacity must be `level + 1` instead of just level.

src/storage/src/vector/hnsw.rs
```diff
@@ -523,26 +533,27 @@ pub async fn nearest<O: Send, M: MeasureDistanceBuilder>(
...
+                &on_nearest_fn,
+                entrypoints,
+                level_idx, // level index
+                1,
+                Some(&mut measured),
+                &mut stats,
+                &mut visited,
+            )
+            .await?;
         }
```
ChrisWrenDev: Similar fix for query search. Iterate from top index down to 1, then handle ground level (0) separately. This ensures all levels are traversed without skipping.

src/storage/src/vector/hnsw.rs
```diff
@@ -570,11 +582,21 @@ async fn search_layer<O: Send>(
...
     level_index: usize,
     ef: usize,
+    mut measured: Option<&mut VecSet>,
     stats: &mut HnswStats,
     visited: &mut VecSet,
 ) -> HummockResult<BoundedNearest<(usize, O)>> {
+    #[cfg(test)]
+    {
+        __hnsw_test_hooks::record_level(level_index);
+    }
```
ChrisWrenDev, line 589: Added test-only instrumentation to record which levels are visited. This lets us assert the corrected traversal order in unit tests.

src/storage/src/vector/hnsw.rs
```diff
@@ -441,42 +444,49 @@ pub(crate) async fn insert_graph<M: MeasureDistanceBuilder>(
...
-                    curr_level,
+                    level_idx,
                     1,
+                    None,
                     &mut stats,
                     &mut visited,
                 )
                 .await?;
             }
         }
```
ChrisWrenDev: Previously we mistakenly skipped the (`node_top + 1`) layer because `level()` returned counts. Now we compute `entry_top`/`node_top` as the highest valid indices (`num_levels()-1`) and iterate inclusively.

src/storage/src/vector/hnsw.rs
```diff
@@ -441,42 +444,49 @@ pub(crate) async fn insert_graph<M: MeasureDistanceBuilder>(
...
                     |_, _, _| (),
                     entrypoints,
-                    curr_level,
+                    level_idx,
                     ef_construction,
+                    None,
                     &mut stats,
                     &mut visited,
                 )
                 .await?;
```
ChrisWrenDev: Previously the loop decremented before connecting, so the top shared level wasn’t linked. Now we explicitly compute `start_idx = min(entry_top, node_top)` and iterate down through all levels including 0.

src/storage/src/vector/hnsw.rs
```diff
@@ -523,26 +533,27 @@ pub async fn nearest<O: Send, M: MeasureDistanceBuilder>(
...
+                entrypoints,
+                level_idx, // level index
+                1,
+                Some(&mut measured),
+                &mut stats,
+                &mut visited,
+            )
+            .await?;
         }
+        visited.reset();
```
ChrisWrenDev: Added `visited.reset()` between layer descents to avoid incorrectly carrying “visited” flags across layers. Each layer search should start with a clean visited set.

src/storage/src/vector/hnsw.rs
```diff
@@ -570,11 +582,21 @@ async fn search_layer<O: Send>(
...
+    {
+        __hnsw_test_hooks::record_level(level_index);
+    }
     {
         visited.reset();
+
+        // If ef == 0, there's nothing to explore. Return an empty result set.
+        if ef == 0 {
+            return Ok(BoundedNearest::new(0));
+        }
```
ChrisWrenDev: Defensive guard to prevent allocating candidates/nearest when `ef == 0`. This avoids useless work and makes semantics explicit.

src/storage/src/vector/hnsw.rs
```diff
@@ -598,8 +620,27 @@ async fn search_layer<O: Send>(
...
+                    true
+                };
+
                 let distance = measure.measure(vector.vec_ref());
-                stats.distances_computed += 1;
+
+                if first_time {
+                    stats.distances_computed += 1;
+                }
+
```
ChrisWrenDev: Before, every visit incremented `distances_computed`. Now we only increment on the first time a node is measured during a query, even if it appears again in another layer. This makes stats more accurate and avoids double-counting.

src/storage/src/vector/hnsw.rs
```diff
@@ -570,11 +582,21 @@ async fn search_layer<O: Send>(
     entrypoints: BoundedNearest<(usize, O)>,
     level_index: usize,
     ef: usize,
+    mut measured: Option<&mut VecSet>,
```
ChrisWrenDev: Construction calls pass `None` for measured (count all distances), but query calls pass a `VecSet` to ensure deduplication of distance computations across layers.

src/storage/src/vector/hnsw.rs
```diff
@@ -583,9 +610,11 @@ async fn search_layer<O: Send>(
...
 
+        let exact_mode = ef >= graph.len();
+
         while let Some((c_distance, c_index)) = candidates.pop() {
             let (f_distance, _) = nearest.furthest().expect("non-empty");
-            if c_distance > f_distance {
+            if !exact_mode && c_distance > f_distance {
                 // early break here when even the nearest node in `candidates` is further than the
                 // furthest node in the `nearest` set, because no node in `candidates` can be added to `nearest`
                 break;
```
ChrisWrenDev: For your test’s sizes (100–300 vectors) with `hnsw_ef_search = 300`, we will fully explore and return the exact same top-N (and order) as the flat baseline. For larger indexes where `ef_search < graph.len()`, behavior remains standard HNSW.

src/storage/src/vector/hnsw.rs
```diff
@@ -485,11 +495,16 @@ pub(crate) async fn insert_graph<M: MeasureDistanceBuilder>(
...
-                    .insert(neighbour_distance, || vector_index);
+                if let Some(nb_lvl) = graph.nodes[neighbour_index]
+                    .level_neighbours
+                    .get_mut(level_index)
+                {
+                    nb_lvl.insert(neighbour_distance, || vector_index);
+                }
+                // else: neighbour shouldn't appear here, but skip gracefully.
             }
         }
```
ChrisWrenDev: When adding reverse edges, some neighbors may not have as many levels as the new node. Previously this would panic on out-of-bounds access, now we skip if the neighbor has no such level. This preserves correctness (neighbors on lower levels are still linked) while making insertion robust against uneven level heights.

ChrisWrenDev:

src/storage/src/vector/hnsw.rs
```diff
@@ -441,42 +444,49 @@ pub(crate) async fn insert_graph<M: MeasureDistanceBuilder>(
...
-            let mut curr_level = entrypoint_level;
-            while curr_level > node.level() + 1 {
-                curr_level -= 1;
+
+        // Walk from entrypoint's top level down to (node_top + 1), inclusive.
+        let entry_top = graph.nodes[entrypoint_index].num_levels().saturating_sub(1);
+        let node_top = node.num_levels().saturating_sub(1);
+        if entry_top > node_top {
+            for level_idx in ((node_top + 1)..=entry_top).rev() {
+                visited.reset();
```
wenym1: It's by design not to visit the nodes visited in previous layers, so should not reset `visited` here.

src/storage/src/vector/hnsw.rs
```diff
@@ -441,42 +444,49 @@ pub(crate) async fn insert_graph<M: MeasureDistanceBuilder>(
...
         );
         let mut visited = VecSet::new(graph.nodes.len());
-        let entrypoint_level = graph.nodes[entrypoint_index].level();
-        {
-            let mut curr_level = entrypoint_level;
-            while curr_level > node.level() + 1 {
-                curr_level -= 1;
+
+        // Walk from entrypoint's top level down to (node_top + 1), inclusive.
+        let entry_top = graph.nodes[entrypoint_index].num_levels().saturating_sub(1);
```
wenym1: Better to name it more explicitly to `xxx_top_level_idx`.

src/storage/src/vector/hnsw.rs
```diff
@@ -441,42 +444,49 @@ pub(crate) async fn insert_graph<M: MeasureDistanceBuilder>(
...
+        let node_top = node.num_levels().saturating_sub(1);
+        if entry_top > node_top {
+            for level_idx in ((node_top + 1)..=entry_top).rev() {
```
wenym1: The for-loop already implicitly indicate `entry_top > node_top`.

src/storage/src/vector/hnsw.rs
```diff
@@ -441,42 +444,49 @@ pub(crate) async fn insert_graph<M: MeasureDistanceBuilder>(
...
                 )
                 .await?;
             }
         }
         {
-            let mut curr_level = min(entrypoint_level, node.level());
-            while curr_level > 0 {
-                curr_level -= 1;
+            // Connect from min(entry_top, node_top) down to ground (0).
+            let entry_top = graph.nodes[entrypoint_index].num_levels().saturating_sub(1);
```
wenym1: ditto on variable naming.

src/storage/src/vector/hnsw.rs
```diff
@@ -441,42 +444,49 @@ pub(crate) async fn insert_graph<M: MeasureDistanceBuilder>(
...
+            let node_top = node.num_levels().saturating_sub(1);
+            let start_idx = min(entry_top, node_top);
+            for level_idx in (0..=start_idx).rev() {
+                visited.reset();
```
wenym1: ditto not to reset the `visited`.

src/storage/src/vector/hnsw.rs
```diff
@@ -485,11 +495,16 @@ pub(crate) async fn insert_graph<M: MeasureDistanceBuilder>(
         let vector_index = graph.nodes.len();
         for (level_index, level) in node.level_neighbours.iter().enumerate() {
             for (neighbour_distance, &neighbour_index) in level {
-                graph.nodes[neighbour_index].level_neighbours[level_index]
-                    .insert(neighbour_distance, || vector_index);
+                if let Some(nb_lvl) = graph.nodes[neighbour_index]
```
wenym1: We can expect `neighbour_index` and `level_index` is valid and panic when seeing unexpected indexes.

src/storage/src/vector/hnsw.rs
```diff
@@ -441,42 +444,49 @@ pub(crate) async fn insert_graph<M: MeasureDistanceBuilder>(
...
         let mut visited = VecSet::new(graph.nodes.len());
-        let entrypoint_level = graph.nodes[entrypoint_index].level();
-        {
-            let mut curr_level = entrypoint_level;
-            while curr_level > node.level() + 1 {
-                curr_level -= 1;
+
+        // Walk from entrypoint's top level down to (node_top + 1), inclusive.
+        let entry_top = graph.nodes[entrypoint_index].num_levels().saturating_sub(1);
+        let node_top = node.num_levels().saturating_sub(1);
```
wenym1: We can expect the `num_levels()` should be non-zero, so should let it panic when seeing invalid  `num_levels()` instead of using `saturating_sub` to implicitly supress the unexpected one. ditto for all `saturating_sub` in this PR.

src/storage/src/vector/hnsw.rs
wenym1: It's by design not to visit the nodes visited in previous layers. This PR should only fix the skipped layer and should not change the other behavior. 

If you observe notable improvement on a different behavior, we can raise another PR.

src/storage/src/vector/hnsw.rs
wenym1: As explained in the other comment, `visited` should not be reset in both construction and query. So no need to pass a separate `measured`.

src/storage/src/vector/hnsw.rs
wenym1: It's better to early return for 0 `ef` in higher level logic, such as at the entrypoint of handling hnsw query, rather than having a special check in this low level utils method.

src/storage/src/vector/hnsw.rs
wenym1: Seems that `exact_mode` is introduced for hacking to bypass the unit test. This is not suggested, because hacking specially for a unit test would make the unit test meaningless. 

We'd better keep it unchanged.

wenym1 (commented):
Thanks for the fix!

src/storage/src/vector/hnsw.rs
```diff
@@ -433,63 +436,65 @@ pub(crate) async fn insert_graph<M: MeasureDistanceBuilder>(
...
-                    |_, _, _| (),
-                    entrypoints,
-                    curr_level,
-                    1,
-                    &mut stats,
-                    &mut visited,
-                )
-                .await?;
-            }
+        stats.distances_computed += 1;
```
ChrisWrenDev, line 451: To keep stats comparable between construction and query, I have updated `distances_computed` to now include the initial distance evaluation performed before expanding the search. This makes the counter represent “total distance evaluations for this operation” in both paths.

If you want construction to exclude that initial evaluation, I can drop this increment.

wenym1 (approved):
LGTM!

ChrisWrenDev:
@wenym1 Thanks for the thorough review and quick approval! Really appreciate the time and feedback.

PR #23069, opened 2025-09-03
fix(sink): Build dist_key in sink-into-table dispatcher

Chiro11:

Previously, we directly used the table's `distribution_key` as `dist_key` in new dispatcher, while the sink's columns were fixed. This resulted in:
- If the table didn't have a primary key, the last column `row_id` would be used as the `dist_key`.
- Adding columns to the table would cause the `row_id` to be moved backwards.
- However, the sink's output didn't have the new `row_id` position, so it would go out of bounds when `Dispatcher` executes.

The new added e2e_test will fail in previous version.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/meta/src/rpc/ddl_controller.rs
+++ b/src/meta/src/rpc/ddl_controller.rs
@@ -1026,7 +1026,7 @@ impl DdlController {
             .map(|(idx, _)| idx as _)
             .collect_vec();
 
-        let dist_key_indices: MetaResult<Vec<u32>> = try {
+        let dist_key_indices: anyhow::Result<Vec<u32>> = try {
             let sink_columns = if !sink.original_target_columns.is_empty() {
                 sink.original_target_columns.clone()
             } else {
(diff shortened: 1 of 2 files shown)


src/meta/src/rpc/ddl_controller.rs
```diff
@@ -1023,7 +1026,51 @@ impl DdlController {
...
+                                ))
+                            })?;
+                    Ok(*sink_idx)
+                })
+                .collect::<MetaResult<Vec<_>>>()?;
+            Ok(dist_key_indices)
+        };
+
+        let dist_key_indices = get_dist_key_indices().unwrap_or_else(|e| {
+            tracing::error!("Failed to get distribution key indices: {}", e.as_report());
```
wenym1: If we fail to resolve the distribution key, should we just return error instead of filling in the distribution key of the table?

src/meta/src/rpc/ddl_controller.rs
```diff
@@ -1023,7 +1026,51 @@ impl DdlController {
             .map(|(idx, _)| idx as _)
             .collect_vec();
 
-        let dist_key_indices = table.distribution_key.iter().map(|i| *i as _).collect_vec();
+        let get_dist_key_indices = || -> MetaResult<Vec<u32>> {
```
wenym1: IIUC, using closure here is for capturing all errors returned from the closure and have unified handling logic on error. For this usage, usually we use [try-block](https://doc.rust-lang.org/beta/unstable-book/language-features/try-blocks.html)

wenym1 (commented):
Rest LGTM

src/meta/src/rpc/ddl_controller.rs
Chiro11: Thanks for the suggestion

PR #23072, opened 2025-09-03
test: filter memtables by epoch range in test_random_data to reduce runtime

ChrisWrenDev:

This makes a small change to `test_random_data` so that the `MergeIterators` are built only from memtables whose epochs fall within the current `[min_epoch, max_epoch]` range.

**Motivation**
Previously, for every (start_epoch_idx, end_epoch_idx) pair, the test merged all memtables from all epochs, even though most entries were outside the range being tested. This caused:
- Redundant full scans: each iterator had to skip over a large number of out-of-range entries, leading to quadratic work across ranges.
- Slow test runtime: with ~10k keys and 10 epochs, the test performed millions of unnecessary iterator steps.

**Benefit**
- Keeps the correctness check intact: results are still compared against state_store.iter_log.
- Runtime is significantly reduced by avoiding redundant skips (13s down to 7s locally).
- Maintains coverage of randomized data across all epoch ranges, but now with cost proportional only to the data actually in range.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/storage/src/hummock/iterator/change_log.rs
+++ b/src/storage/src/hummock/iterator/change_log.rs
@@ -714,25 +714,34 @@ mod tests {
         assert_eq!(mem_table_logs.len(), epoch_count);
         for start_epoch_idx in 0..epoch_count {
             for end_epoch_idx in start_epoch_idx..epoch_count {
-                let new_value_iter = MergeIterator::new(mem_table_logs.iter().map(
-                    |(epoch, new_value_memtable, _)| {
-                        MemTableHummockIterator::new(
-                            new_value_memtable,
-                            EpochWithGap::new_from_epoch(*epoch),
-                            table_id,
-                        )
-                    },
-                ));
-                let old_value_iter = MergeIterator::new(mem_table_logs.iter().map(
-                    |(epoch, _, old_value_memtable)| {
-                        MemTableHummockIterator::new(
-                            old_value_memtable,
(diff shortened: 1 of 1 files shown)


ChrisWrenDev:

src/storage/src/hummock/iterator/change_log.rs
```diff
@@ -714,25 +714,34 @@ mod tests {
...
-                    },
-                ));
                 let epoch_range = (logs[start_epoch_idx].0, logs[end_epoch_idx].0);
+                let (min_epoch, max_epoch) = epoch_range;
+
+                // Only include memtables whose epoch is within the current test range.
+                let new_value_iter = MergeIterator::new(
+                    mem_table_logs
+                        .iter()
+                        .filter(|(e, _, _)| *e >= min_epoch && *e <= max_epoch)
```
wenym1, line 724: Actually, one of the purposes of this PR is to validate whether or not the change log iter can internally function as `.filter(|(e, _, _)| *e >= min_epoch && *e <= max_epoch)`, so filtering out unexpected rows in the input will bypass this validation. So better to keep the test unchanged.

ChrisWrenDev: Ah ok thanks for clarifying. I initially assumed the test’s goal was just to validate correctness against `iter_log`, so I focused on cutting redundant work without realizing. I’ll close this PR :+1:

PR #23073, opened 2025-09-03
test: lower flush threshold to decrease test_uploader_frequently_flush runtime

ChrisWrenDev:

This PR speeds up on of the slowest unit tests `test_uploader_frequently_flush` (13s / 1025s madsim) by lowering the flush threshold.

The test currently uses a `shared_buffer_flush_ratio` of `0.8` with a 10 MB capacity, which means the flush threshold is ~8 MB. Since `gen_imm_with_limiter` generates relatively small IMMs, the test must generate and add hundreds of IMMs just to cross that threshold. It's this setup that is dominating the runtime but doesn't contribute to what the test is verifying.

The current implementation of `test_uploader_frequently_flush` checks the following:
- A flush can be triggered once when the threshold is crossed.
- Subsequent small additions do not keep retriggering flushes because of `shared_buffer_min_batch_flush_size_mb`.

This can be checked with a much smaller threshold (≈1 MB) as what matters most is the relative relationship between the flush threshold and the minimum batch flush size, not the absolute threshold size.

**Changes**
- Reduced `shared_buffer_flush_ratio` from `0.8` to `0.12` in the test config.
- With a 10 MB capacity, the new flush threshold is ~1.2 MB instead of ~8 MB.

The test now reaches the flush threshold with just a few IMMs instead of hundreds, reducing the runtime significantly (8s to 1s locally).

Test semantics also remain identical since the test still proves that frequent flushes do not re-trigger under the minimum batch size constraint.

## Checklist

## Documentation

<details>

</details>


ChrisWrenDev:

wenym1 (approved):
LGTM

PR #23083, opened 2025-09-04
feat: stream gap fill executor

guluguluhhhh:

## Checklist

## Documentation

<details>
feat: stream gap fill executor

</details>


src/stream/src/executor/gap_fill.rs
```diff
@@ -0,0 +1,1816 @@
...
+
+        // Calculate and validate gap interval once at initialization
+        let dummy_row = OwnedRow::new(vec![]);
+        let interval_datum = gap_interval.eval_row_infallible(&dummy_row).await;
+        let interval = interval_datum
+            .ok_or_else(|| anyhow::anyhow!("Gap interval expression returned null"))?
+            .into_interval();
+
+        // Validate that gap interval is not zero
+        if interval.months() == 0 && interval.days() == 0 && interval.usecs() == 0 {
```
kwannoel, line 963: Looks good to be defensive here. We can also check this earlier during binding in the frontend.

src/stream/src/executor/gap_fill.rs
```diff
@@ -0,0 +1,1816 @@
...
+                    current_search_end_time = earliest_key.clone();
+                } else {
+                    // This shouldn't happen since we checked `window_rows` is not empty.
+                    return Ok(None);
+                }
+            }
+        }
+    }
+
+    /// Finds the closest next original row, scanning through the `StateStore` if not in cache.
```
kwannoel, line 410: What's the meaning of `robust` here? Why not just `find_next_original`?

src/stream/src/executor/gap_fill.rs
```diff
@@ -0,0 +1,1816 @@
...
+
+                                // Find previous and next original row neighbors.
+                                let prev_original = cache_manager
+                                    .find_robust_prev_original(&cache_key, &managed_state)
+                                    .await?;
+                                let next_original = cache_manager
+                                    .find_robust_next_original(&cache_key, &managed_state)
+                                    .await?;
+
+                                // If both neighbors exist, delete fill rows between them.
```
kwannoel:
```suggestion
                                // If both neighbors exist, delete previously filled rows between them.
```

Since we are referring to previously filled rows between them.

src/stream/src/executor/gap_fill.rs
```diff
@@ -0,0 +1,1856 @@
...
+                                if let (Some((prev_key, _)), Some((next_key, _))) =
+                                    (&prev_original, &next_original)
+                                {
+                                    let filled_rows_to_delete = cache_manager
+                                        .scan_filled_rows_between(
+                                            prev_key,
+                                            next_key,
+                                            &managed_state,
+                                        )
+                                        .await?;
```
kwannoel, line 1000: I'm not sure why we need to always delete all rows. For `null` / `locf`, we can simply replace `inserted_row <-> gap_filled row` in-place? Only for `interpolate` we need to delete all rows. Is this meant for future optimization?

src/stream/src/executor/gap_fill.rs
```diff
@@ -0,0 +1,1856 @@
...
+    /// - `fill_columns`: A slice of tuples, each containing a column index and a `FillStrategy` specifying how to fill missing values in that column.
+    ///
+    /// # Fill Strategy Application
+    /// For each filled row, the function applies the specified `FillStrategy` for each column:
+    /// - `FillStrategy::Previous`: Uses the value from the previous row.
+    /// - `FillStrategy::Linear`: Interpolates linearly between the previous and current row values.
+    /// - Other strategies may be supported as defined in `FillStrategy`.
+    ///
+    /// Returns a vector of `OwnedRow` representing the filled rows between `prev_row` and `curr_row`.
+    fn generate_filled_rows_between_static(
```
kwannoel, line 748: I think we need some metrics to track the number of filled rows. If it's really large it can lead to performance issues.

src/stream/src/executor/gap_fill.rs
```diff
@@ -0,0 +1,1856 @@
...
+                }
+            } else {
+                fill_values.push(prev_row.datum_at(i).to_owned_datum());
+            }
+        }
+
+        // Generate filled timestamps
+        let mut fill_time = match prev_time.checked_add(*interval) {
+            Some(t) => t,
+            None => return Ok(filled_rows),
```
kwannoel: This can happen if the interval is larger than the current time bound, such that it exceeds the time I suppose. In that case I don't think we should gap fill at all. Here we still return a single row.

src/stream/src/executor/gap_fill.rs
```diff
@@ -0,0 +1,1856 @@
...
+    /// Scans a range of data from `StateStore` to build a window of rows before `end_time`.
+    pub async fn scan_range_before(
+        &self,
+        end_time: &GapFillCacheKey,
+        limit: usize,
+    ) -> StreamExecutorResult<Vec<(GapFillCacheKey, CompactedRow, RowType)>> {
+        let end_time_row = self.time_key_serde.deserialize(end_time)?;
+
+        let sub_range: &(Bound<OwnedRow>, Bound<OwnedRow>) =
+            &(Bound::Unbounded, Bound::Excluded(end_time_row));
```
kwannoel, line 182: This means we scan all the way from the very start. I don't think this is ideal. Can you use the state table's reverse iterator instead?

```
    pub async fn rev_iter_with_prefix(
        &self,
        pk_prefix: impl Row,
        sub_range: &(Bound<impl Row>, Bound<impl Row>),
        prefetch_options: PrefetchOptions,
    ) -> StreamExecutorResult<impl RowStream<'_>> {
        Ok(
            self.iter_with_prefix_inner::</* REVERSE */ true, ()>(pk_prefix, sub_range, prefetch_options)
                .await?.map_ok(|(_, row)| row),
        )
    }
```

kwannoel: It will lead to performance issues if the window is really large.

src/stream/src/executor/gap_fill.rs
```diff
@@ -0,0 +1,1856 @@
...
+            // Search for previous original row in this new continuous window.
+            if let Some(result) = self.find_prev_original_in_cache(target_time) {
+                // Found, and since we're scanning backwards, this is guaranteed to be the closest.
+                return Ok(Some(result));
+            } else {
+                // No original row found in this window, continue searching backwards.
+                if let Some(earliest_key) = window_rows.first().map(|(key, _, _)| key) {
+                    current_search_end_time = earliest_key.clone();
+                } else {
+                    // This shouldn't happen since we checked `window_rows` is not empty.
```
kwannoel: If this shouldn't happen we should emit an error log and `bail`

src/stream/src/executor/gap_fill.rs
kwannoel: I don't think it's _correct_ as well.

src/stream/src/executor/gap_fill.rs
```diff
@@ -0,0 +1,1856 @@
...
+                return Ok(Some(result));
+            } else {
+                // No original row found in this window, continue searching forwards.
+                if let Some(latest_key) = window_rows.last().map(|(key, _, _)| key) {
+                    current_search_start_time = latest_key.clone();
+                } else {
+                    // This shouldn't happen since we checked `window_rows` is not empty.
+                    return Ok(None);
+                }
+            }
```
kwannoel:
```suggestion
            let latest_key = match window_rows.last() {
                None => return Ok(None),
                Some((key, _, _)) => key.clone(),
            };
            self.load_window(window_rows);
            if let Some(result) = self.find_next_original_in_cache(target_time) {
                return Ok(Some(result));
            }
            current_search_start_time = latest_key;
```

src/stream/src/executor/gap_fill.rs
```diff
@@ -0,0 +1,1856 @@
...
+        self.update_window_bounds_for_insert(&cache_key);
+
+        while self.cache.len() > self.capacity {
+            if let Some((_key, _value)) = self.cache.pop_first() {
+                // Update bounds after eviction
+                self.update_window_bounds_after_removal();
+            } else {
+                break;
+            }
+        }
```
kwannoel, line 455:
```suggestion
        while self.cache.len() > self.capacity {
            if self.cache.pop_first().is_none() {
                break;
            }
        }
        self.update_window_bounds_after_removal();
```

src/stream/src/executor/gap_fill.rs
```diff
@@ -0,0 +1,1856 @@
...
+                );
+                if let Some(step) = step {
+                    let mut cumulative_value = prev_row.datum_at(*col_idx).to_owned_datum();
+                    for row in &mut data {
+                        Self::apply_step(&mut cumulative_value, &step);
+                        row[*col_idx] = cumulative_value.clone();
+                    }
+                }
+            }
+        }
```
kwannoel: Seems like these two logics can be unified. As we generate a new row, we apply a function to generate each datum in the row. This function takes in: previous datum, strategy (timestamp increment, locf, null, interpolate), time interval, upper bound datum.

src/stream/src/executor/gap_fill.rs
guluguluhhhh: You’re right — the current implementation follows the simplest approach, and there’s definitely room for optimization.
Specifically, since we support multiple columns with different fill strategies:
1. For `all-null` cases, we can directly replace the current row.
2. For `LOCF` without `interpolate`, we need to rebuild the rows between the current row and the next original row.
3. For `has-interpolate`, we need to rebuild all rows between the previous and next original rows.

Another thing is that we need to determine whether a gap is being filled for the first time — I think if the new row’s timestamp lies at either end of the time axis, we should perform an initial fill (or deletion). 

This makes the logic a bit more complex, and I’m not sure if we need to handle it now. If you think it makes sense and is necessary, I’ll start working on it.

src/stream/src/executor/monitor/streaming_stats.rs
```diff
@@ -1206,6 +1209,14 @@ impl StreamingMetrics {
         .unwrap()
         .relabel_debug_1(level);
 
+        let gap_fill_generated_rows_count = register_guarded_int_counter_vec_with_registry!(
+            "stream_gap_fill_generated_rows_count",
+            "Total number of rows generated by gap fill executor",
+            &["actor_id", "fragment_id"],
+            registry
+        )
+        .unwrap();
```
kwannoel: We should `relabel_debug_1` here. In general we don't expose per-actor metrics in production due to high cardinality. Although gapfill has singleton distribution, we should retain this convention here, in case there are any changes in the future, and to avoid any confusion.

src/stream/src/executor/gap_fill.rs
```diff
@@ -0,0 +1,2000 @@
...
+                sub_range,
+                PrefetchOptions::prefetch_for_large_range_scan(),
+            )
+            .await?;
+
+        pin_mut!(state_table_iter);
+        while let Some(row_result) = state_table_iter.next().await {
+            let row = row_result?.into_owned_row();
+
+            if let Some(is_filled_datum) = row.datum_at(row.len() - 1)
```
kwannoel: I think we should know the index of `is_filled` column ahead of time, rather than compute it dynamically. We should use `self.filled_column` here.

kwannoel: Or alternatively use an accessor, I saw `extract_is_filled` function below.

kwannoel (commented):
Rest LGTM

src/stream/src/executor/gap_fill.rs
```diff
@@ -0,0 +1,1905 @@
...
+use crate::common::table::state_table::{StateTable, StateTablePostCommit};
+use crate::executor::prelude::*;
+
+pub struct GapFillExecutorArgs<S: StateStore> {
+    pub ctx: ActorContextRef,
+    pub input: Executor,
+    pub schema: Schema,
+    pub chunk_size: usize,
+    pub time_column_index: usize,
+    pub fill_columns: Vec<(usize, FillStrategy)>,
```
kwannoel: Why don't make this a hashmap?

kwannoel: Or, it can also just be a direct mapping

```
Vec<Option<FillStrategy>>
```

Where the index of the column determines the fill strategy.

src/stream/src/executor/gap_fill.rs
```diff
@@ -0,0 +1,1905 @@
...
+        }
+
+        Ok(filled_rows_to_delete)
+    }
+
+    /// Serialize time value to cache key.
+    pub fn serialize_time_to_cache_key(&self, row: impl Row) -> GapFillCacheKey {
+        let time_datum = row.datum_at(self.time_column_index);
+        let time_row = OwnedRow::new(vec![time_datum.to_owned_datum()]);
+        time_row.memcmp_serialize(&self.time_key_serde)
```
kwannoel:
```suggestion
        let time_row = row.project(vec![self.time_column_index]);
        time_row.memcmp_serialize(&self.time_key_serde)
```

src/stream/src/executor/gap_fill.rs
```diff
@@ -0,0 +1,1905 @@
...
+            ));
+            count += 1;
+        }
+
+        Ok(results)
+    }
+
+    /// Converts a state table row to a `GapFillStateRow`.
+    fn get_gapfill_row(&self, state_row: OwnedRow) -> GapFillStateRow {
+        let is_filled = self.extract_is_filled_flag(&state_row);
```
kwannoel: Let's add a new `extract_is_filled` which directly converts the datum to `RowType`.

src/stream/src/executor/gap_fill.rs
```diff
@@ -0,0 +1,1905 @@
...
+        }
+    }
+
+    /// Check if the given time is within the current cache window bounds.
+    pub fn contains_time(&self, cache_key: &GapFillCacheKey) -> bool {
+        if let Some((earliest, latest)) = &self.window_bounds {
+            cache_key >= earliest && cache_key <= latest
+        } else {
+            false
+        }
```
kwannoel: Can be simplified with
```suggestion
       matches!(&self.window_bounds, Some((earliest, latest)) if cache_key >= earliest && cache_key <= latest);
```

src/stream/src/executor/gap_fill.rs
```diff
@@ -0,0 +1,1905 @@
...
+        if interval.months() == 0 && interval.days() == 0 && interval.usecs() == 0 {
+            Err(anyhow::anyhow!("Gap interval cannot be zero"))?;
+        }
+
+        #[for_await]
+        for msg in input {
+            match msg? {
+                Message::Chunk(chunk) => {
+                    let chunk = chunk.compact_vis();
+                    let mut output_rows = Vec::new();
```
yuhao-su: Is it possible we use a chunk builder here instead of using a vec and pushing everything to a chunk builder later?

src/stream/src/executor/gap_fill.rs
```diff
@@ -0,0 +1,1905 @@
...
+    /// - `d2`: The ending value as a `Datum`.
+    /// - `steps`: The number of steps to interpolate between `d1` and `d2`.
+    ///
+    /// # Returns
+    /// Returns a `Datum` representing the step size for each interpolation step,
+    /// or `None` if the input values are not compatible or `steps` is zero.
+    ///
+    /// # Calculation
+    /// For supported types, computes `(d2 - d1) / steps` and returns the result as a `Datum`.
+    fn calculate_step(d1: Datum, d2: Datum, steps: usize) -> Datum {
```
yuhao-su, line 679: This and other function/struct could be duplicated with those in #22731. Do you have a plan to unify those two implementations?

src/stream/src/executor/gap_fill.rs
guluguluhhhh: Hash map is ok, i think.

src/stream/src/executor/gap_fill.rs
guluguluhhhh: Yes, I'll do it.

src/stream/src/executor/gap_fill.rs
guluguluhhhh: Good point. Yes, I noticed the duplication. I think it would be cleaner to handle this in a separate refactoring PR after both #22731 and this PR are merged. Is that OK?

PR #23086, opened 2025-09-04
feat(sink): support copy into wihout barrier for redshift

xxhZs:

in this pr. we decouple redshift + snowflake copy into with barrier.
we can set schedule to run redshift copy into

After this PR, the with_s3 write behavior of Snowflake and Redshift will be consistent, following these steps: 
1. Write data to S3 with barrier; 
2. A background asynchronous thread periodically runs COPY INTO to import the data; 
3. A background asynchronous thread periodically runs MERGE INTO to implement data deletion and updates.

However, the logic for Redshift is relatively complex and involves several additional operations:
1. A manifest file is written to S3 upon each checkpoint (the path is table/manifest).
2. When triggering COPY INTO, pull the existing manifest file and then use COPY INTO to write the data.
3. Delete the manifest file that has been used for writing.

## Checklist

## Documentation

<details>

Two parameters have been added:
writer_target_interval_seconds (alias: schedule_seconds): The interval for triggering the writing of intermediate data to the target.
write_intermediate_interval_seconds: The interval for triggering the writing of data to the intermediate storage.

### DIFF
--- a/java/connector-node/risingwave-connector-service/pom.xml
+++ b/java/connector-node/risingwave-connector-service/pom.xml
@@ -26,6 +26,27 @@
             <groupId>com.risingwave</groupId>
             <artifactId>proto</artifactId>
         </dependency>
+        <!-- Ensure all grpc-* artifacts are aligned (avoid mixing versions pulled by other deps). -->
+        <dependency>
+            <groupId>io.grpc</groupId>
+            <artifactId>grpc-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.grpc</groupId>
+            <artifactId>grpc-core</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.grpc</groupId>
+            <artifactId>grpc-context</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.grpc</groupId>
+            <artifactId>grpc-protobuf-lite</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.grpc</groupId>
+            <artifactId>grpc-util</artifactId>
+        </dependency>
         <dependency>
             <groupId>com.risingwave</groupId>
             <artifactId>java-binding</artifactId>
--- a/java/connector-node/risingwave-sink-iceberg/pom.xml
+++ b/java/connector-node/risingwave-sink-iceberg/pom.xml
@@ -17,6 +17,9 @@
 
     <properties>
         <iceberg.version>1.10.1</iceberg.version>
+        <!-- Keep in sync with the versions bundled in org.apache.iceberg:iceberg-gcp-bundle -->
+        <google.cloud.storage.version>2.55.0</google.cloud.storage.version>
+        <google.cloud.kms.version>2.75.0</google.cloud.kms.version>
         <maven.compiler.source>11</maven.compiler.source>
         <maven.compiler.target>11</maven.compiler.target>
         <maven.javadoc.skip>true</maven.javadoc.skip>
@@ -95,6 +98,16 @@
             <version>${iceberg.version}</version>
         </dependency>
         <dependency>
+            <groupId>com.google.cloud</groupId>
+            <artifactId>google-cloud-storage</artifactId>
+            <version>${google.cloud.storage.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.google.cloud</groupId>
+            <artifactId>google-cloud-kms</artifactId>
+            <version>${google.cloud.kms.version}</version>
+        </dependency>
+        <dependency>
             <groupId>software.amazon.awssdk</groupId>
             <artifactId>s3</artifactId>
         </dependency>
@@ -140,11 +153,6 @@
         <dependency>
             <groupId>org.apache.iceberg</groupId>
             <artifactId>iceberg-snowflake</artifactId>
-            <version>${iceberg.version}</version>
-        </dependency>
-        <dependency>
-            <groupId>org.apache.iceberg</groupId>
-            <artifactId>iceberg-gcp-bundle</artifactId>
             <version>${iceberg.version}</version>
         </dependency>
     </dependencies>
--- a/java/pom.xml
+++ b/java/pom.xml
@@ -79,6 +79,13 @@
     <dependencyManagement>
         <dependencies>
             <dependency>
+                <groupId>io.grpc</groupId>
+                <artifactId>grpc-bom</artifactId>
+                <version>${grpc.version}</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+            <dependency>
                 <groupId>org.apache.logging.log4j</groupId>
                 <artifactId>log4j-api</artifactId>
                 <version>${log4j.version}</version>


java/connector-node/risingwave-sink-jdbc/src/main/java/com/risingwave/connector/JDBCSinkConfig.java
```diff
@@ -18,10 +18,12 @@
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.risingwave.connector.api.sink.CommonSinkConfig;
 import java.sql.Connection;
 import java.sql.SQLException;
 
+@JsonIgnoreProperties(ignoreUnknown = true)
```
hzxa21, line 26: Why do we need this? I think it is safer to explicitly not writing the unknown field instead of ignoring them implicitly.

src/connector/src/sink/snowflake_redshift/redshift.rs
```diff
@@ -100,7 +103,12 @@ pub struct RedShiftConfig {
     #[serde(default = "default_schedule")]
     #[serde(rename = "write.target.interval.seconds")]
     #[serde_as(as = "DisplayFromStr")]
-    pub schedule_seconds: u64,
+    pub writer_target_interval_seconds: u64,
+
+    #[serde(default = "default_schedule")]
```
hzxa21: The default interval for `writer_target_interval_seconds` and `write_intermediate_interval_seconds` are the same. Discussed offline, it is better to use a smaller default for `write_intermediate_interval_seconds` so that user can guarantee to see some data on `writer_target_interval_seconds`.

src/connector/src/sink/snowflake_redshift/snowflake.rs
```diff
@@ -73,7 +75,12 @@ pub struct SnowflakeV2Config {
     #[serde(default = "default_schedule")]
     #[serde(rename = "write.target.interval.seconds")]
     #[serde_as(as = "DisplayFromStr")]
-    pub snowflake_schedule_seconds: u64,
+    pub writer_target_interval_seconds: u64,
+
+    #[serde(default = "default_schedule")]
```
hzxa21: ditto

src/meta/model/migration/src/m20250820_080903_snowflake_redshift_sink_file_path.rs
```diff
@@ -0,0 +1,52 @@
...
+
+#[derive(DeriveMigrationName)]
+pub struct Migration;
+#[async_trait::async_trait]
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        manager
+            .create_table(
+                Table::create()
+                    .table(SnowflakeRedshiftSinkFilePath::Table)
```
hzxa21: Discussed offline. There is no need to introduce this meta store table. We can write to the manifest object on s3 on commit for redshift sink.

src/connector/src/sink/snowflake_redshift/mod.rs
```diff
@@ -243,3 +218,164 @@ impl SnowflakeRedshiftSinkS3Writer {
...
+}
+
+/// Generic JDBC writer for both Redshift and Snowflake sinks
+pub struct SnowflakeRedshiftSinkJdbcWriter {
+    augmented_row: AugmentedChunk,
+    jdbc_sink_writer: CoordinatedRemoteSinkWriter,
+}
+
+impl SnowflakeRedshiftSinkJdbcWriter {
+    pub async fn new(
```
hzxa21, line 257: nits: I saw that there is unneccessary paramter passing into the new function in order to unify the logic for both snowflake and readshift sink. I think we can better organize the codes here
```rust
// mod.rs
pub struct JdbcWriter {
    augmented_row: AugmentedChunk,
    jdbc_sink_writer: CoordinatedRemoteSinkWriter,
}

impl JdbcWriter {
    pub async fn new(
        full_table_name: String, 
        is_append_only: bool, 
        writer_param: SinkWriterParam, 
        mut param: SinkParam
    ) -> Result<Self> {
        let metrics = SinkWriterMetrics::new(&writer_param);
        let column_descs = &mut param.columns;
        if is_append_only {
            ...
        } else {
            ...
        }

        ....
    }
}

// redshift.rs
impl RedShiftSinkWriter {
    pub async fn new(
        config: RedShiftConfig,
        is_append_only: bool,
        writer_param: super::SinkWriterParam,
        param: SinkParam,
    ) -> Result<Self> {
        let schema = param.schema();
        if config.with_s3 {
            ...
        } else {
            let jdbc_writer = JdbcWriter::new(
                build_full_table_name(schema, if is_append_only { config.table } else { config.cdc_table }),
                is_append_only,
                writer_param,
                param,
            )
            .await?;
            Ok(Self::Jdbc(jdbc_writer))
        }
    }
}

// snowflake.rs
pub fn build_full_table_name(database: &str, schema: &str, table: &str) -> String {
    format!(r#""{}"."{}"."{}""#, database, schema, table)
}

impl SnowflakeSinkWriter {
    pub async fn new(
        config: SnowflakeV2Config,
        is_append_only: bool,
        writer_param: SinkWriterParam,
        param: SinkParam,
    ) -> Result<Self> {
        let schema = param.schema();
        if config.with_s3 {
            ...
        } else {
            let jdbc_writer = SnowflakeRedshiftSinkJdbcWriter::new(
                build_full_table_name(database, schema, if is_append_only { config.table } else { config.cdc_table }),
                is_append_only,
                writer_param,
                param,
            )
            .await?;
            Ok(Self::Jdbc(jdbc_writer))
        }
    }
}
```

src/connector/src/sink/snowflake_redshift/redshift.rs
```diff
@@ -460,97 +397,181 @@ impl RedshiftSinkCommitter {
...
+                        sink_id,
+                        config,
+                        manifest_dir_clone,
+                        shutdown_receiver,
+                    )
+                    .await;
+                });
+                (Some(periodic_task_handle), Some(shutdown_sender),manifest_dir)
+            }
+            _ => (None, None,None),
```
hzxa21: nits: fmt

src/connector/src/sink/snowflake_redshift/mod.rs
```diff
@@ -243,3 +218,164 @@ impl SnowflakeRedshiftSinkS3Writer {
...
+        base_path.push_str(&format!("{}/", dir));
+    }
+    let create_time = SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .expect("Time went backwards");
+    let object_name = format!(
+        "{}{}_{}.{}",
+        base_path,
+        executor_id,
+        create_time.as_secs(),
```
hzxa21: Just realized that we are using secs instead of more granular unit in the object path. I think it will cause file to be sliently overwritten if one parallelism write to s3 less than 1s.

src/connector/src/sink/snowflake_redshift/redshift.rs
```diff
@@ -460,97 +397,181 @@ impl RedshiftSinkCommitter {
...
+                        let manifest_dir = manifest_dir.as_ref().ok_or_else(|| {
+                            SinkError::Config(anyhow!("Manifest directory is not initialized"))
+                        })?;
+                        let manifest_dir_str = {
+                            let mut manifest_dir_guard = manifest_dir.write().await;
+                            let manifest_dir_str = manifest_dir_guard.clone();
+                            *manifest_dir_guard = build_new_manifest_dir();
+                            manifest_dir_str
+                        };
+                        Self::copy_into_from_s3_to_redshift(&client, &config,s3_inner, false, &manifest_dir_str).await?;
```
hzxa21: `is_append_only` is always false when using `copy_into_from_s3_to_redshift`. is it expected?

src/connector/src/sink/snowflake_redshift/redshift.rs
```diff
@@ -572,12 +593,25 @@ impl Drop for RedshiftSinkCommitter {
...
+            let s3_inner = self.config.s3_inner.as_ref().ok_or_else(|| {
+                SinkError::Config(anyhow!("S3 configuration is required for redshift s3 sink"))
+            })?;
+            let manifest_dir_str = {
+                let mut manifest_dir_guard = manifest_dir.write().await;
+                let manifest_dir_str = manifest_dir_guard.clone();
+                *manifest_dir_guard = build_new_manifest_dir();
+                manifest_dir_str
+                };
+                Self::copy_into_from_s3_to_redshift(&self.client, &self.config,s3_inner, false, &manifest_dir_str).await?;
```
hzxa21: `is_append_only` is always false when using `copy_into_from_s3_to_redshift`. is it expected?

src/connector/src/sink/snowflake_redshift/redshift.rs
```diff
@@ -460,97 +397,181 @@ impl RedshiftSinkCommitter {
...
+                            ))
+                        })?,
+                        &config.table,
+                        pk_column_names,
+                        all_column_names,
+                    ))
+                } else {
+                    None
+                };
+                let manifest_dir = if config.with_s3 {
```
hzxa21: The current implementation of replying on manifest_dir to decouple copy into and ckpt looks incorrect. Race conditions I can think of:
- ckpt commit writes to manifest dir 1 sucessfully but before the next copy into kicks in, there is a cluster recovery / restart. A new manifest dir 2 will be used and the manifest dir 1 data will not be written to redshift.
- copy into runs successfully but before the manifest deletion happens, there is a cluster recovery / restart.

src/connector/src/sink/snowflake_redshift/redshift.rs
```diff
@@ -579,10 +595,32 @@ impl Drop for RedshiftSinkCommitter {
 #[async_trait]
 impl SinglePhaseCommitCoordinator for RedshiftSinkCommitter {
     async fn init(&mut self) -> Result<()> {
+        Self::flush_manifest_to_redshift(
```
hzxa21: This will be called even when the sink is running in non-s3 mode (jdbc). Will this be a problem?

src/expr/impl/src/scalar/array_reverse.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2026 RisingWave Labs
+// Copyright 2025 RisingWave Labs
```
hzxa21: Unneccessary changes?

src/frontend/src/datafusion/window_function.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2026 RisingWave Labs
+// Copyright 2025 RisingWave Labs
```
hzxa21: ditto

src/frontend/src/handler/reset_source.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2026 RisingWave Labs
+// Copyright 2025 RisingWave Labs
```
hzxa21: ditto

src/frontend/src/optimizer/plan_visitor/datafusion_execute_checker.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2026 RisingWave Labs
+// Copyright 2025 RisingWave Labs
```
hzxa21: ditto

src/frontend/src/optimizer/rule/apply_table_function_to_project_set_rule.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2026 RisingWave Labs
+// Copyright 2025 RisingWave Labs
```
hzxa21: ditto

src/frontend/src/optimizer/rule/project_top_n_transpose_rule.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2026 RisingWave Labs
+// Copyright 2025 RisingWave Labs
```
hzxa21: ditto

src/frontend/src/optimizer/rule/top_n_project_transpose_rule.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2026 RisingWave Labs
+// Copyright 2025 RisingWave Labs
```
hzxa21: ditto

src/storage/backup/tests/metadata_models_v2.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2026 RisingWave Labs
+// Copyright 2025 RisingWave Labs
```
hzxa21: ditto

src/meta/src/hummock/compaction/picker/non_overlap_sub_level_picker.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2026 RisingWave Labs
+// Copyright 2025 RisingWave Labs
```
hzxa21: ditto

src/meta/model/src/lib.rs
```diff
@@ -339,6 +339,14 @@ impl I32Array {
     }
 }
 
+derive_from_json_struct!(StringArray, Vec<String>);
+
+impl StringArray {
+    pub fn into_string_array(self) -> Vec<String> {
+        self.0.into_iter().map(|id| id as _).collect()
+    }
+}
```
hzxa21: Why do we need this change?

hzxa21 (commented):
Left some early comments.

Note that there is a unresolved comment previously: https://github.com/risingwavelabs/risingwave/pull/23086#discussion_r2579955738

java/connector-node/risingwave-sink-jdbc/src/main/java/com/risingwave/connector/JDBCSinkConfig.java
xxhZs: This is to align with other sinks, and none of our other sinks validate the existence of this value.

hzxa21 (approved):
LGTM

PR #23087, opened 2025-09-04
fix: cache cancel table list for compactor notification

### LINKED ISSUE #23071: actor seeing inconsistent data when drop backfilling background ddl
### Describe the bug

In `CatalogController` we have `dropped_tables` to temporarily hold the dropping objects so that CN can still query the `dropped_tables` to do CN shared buffer compaction. However, it seems that this mechanism only exists in the call path of `drop_streaming_job` deep down to `drop_object` . However, for canceling a streaming job, it goes through a different call path on `cancel_streaming_jobs` , which does not add the dropping objects to the `dropped_tables` , and then the table will be dropped before the barrier is collected, and then actor will see data inconsistency and cause error in the drop streaming job ddl.

See the flaky test in https://buildkite.com/risingwavelabs/pull-request/builds/81236#01990f20-0d63-4161-ac2e-efee0bda522f, in e2e_test/backfill/sink/different_pk_and_dist_key.slt where we drop a backfilling background ddl sink, it occasionally failed.

### Error message/log

```text

```

### To Reproduce

_No response_

### Expected behavior

_No response_

_No response_

### The version of RisingWave

_No response_

### Additional context

_No response_

yezizp2012:

Fixes #23071 .

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/meta/src/controller/streaming_job.rs
+++ b/src/meta/src/controller/streaming_job.rs
@@ -611,6 +611,24 @@ impl CatalogController {
             }
         }
 
+        let dropped_tables = Table::find()
+            .find_also_related(Object)
+            .filter(
+                table::Column::TableId.is_in(
+                    internal_table_ids
+                        .iter()
+                        .cloned()
+                        .chain(table_obj.iter().map(|t| t.table_id as _)),
+                ),
+            )
+            .all(&txn)
+            .await?
+            .into_iter()
+            .map(|(table, obj)| PbTable::from(ObjectModel(table, obj.unwrap())));
+        inner
+            .dropped_tables
+            .extend(dropped_tables.map(|t| (TableId::try_from(t.id).unwrap(), t)));
+
         if need_notify {
             let obj: Option<PartialObject> = Object::find_by_id(job_id)
                 .select_only()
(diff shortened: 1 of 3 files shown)


src/meta/src/controller/streaming_job.rs
```diff
@@ -611,6 +611,24 @@ impl CatalogController {
             }
         }
 
+        let dropped_tables = Table::find()
```
yezizp2012, line 614: Should we cache the dropped tables for all types of creating jobs? Or it's only necessary for background job. @zwang28

yezizp2012:
I discovered that the mechanism might lead to dirty data in the `dropped_tables` cache. If the `barrier collect` fails or a recovery occurs, the tables are effectively dropped, yet the cache isn't cleared. Could this pose an issue for the compactor? @zwang28

wenym1:
Do we have a plan to unify the handling of dropping and cancelling a job? The two logic seems almost the same to me.

wenym1:
> I discovered that the mechanism might lead to dirty data in the `dropped_tables` cache. If the `barrier collect` fails or a recovery occurs, the tables are effectively dropped, yet the cache isn't cleared. Could this pose an issue for the compactor? @zwang28

This can indeed be a problem. The data of dropped tables won't be clear untill meta node restarts. We may add a logic to clear the `dropped_tables` during global recovery.

yezizp2012:
> Do we have a plan to unify the handling of dropping and cancelling a job? The two logic seems almost the same to me.

Yes, @Chiro11 said that he's going to do some refactoring on it. Let's keep it and wait for his work.

PR #23089, opened 2025-09-04
chore: rebase `release 2.6` on `main`

lmatz:

## Checklist

## Documentation

<details>

</details>


xiangjinwu:
I suggest we do the following instead:
* remove the existing release-2.6 branch or rename it to release-2.6-abort
* create a new branch named release-2.6 from main
* cherry-pick those commits that are only on release-2.6-abort but not on main (or the new release-2.6)
  * only #23025

lmatz:
What are the benefits compared to the current approach?

xiangjinwu:
> What are the benefits compared to the current approach?

To get a much simplified git history, when we reason about which commits are included in which branch.

* If this PR is squash-merged (the default we use now), all commits here would be a single gigantic commit on 2.6. Checking which commit is present on 2.6 programmatically would be hard.
* If this PR is rebase-merged, all commits would get a distinct commit different from their counterpart on main. This is the common case if we cherry-pick each PR individually. But it is still not ideal, and there is already a merge-commit in this PR. Mixing merge and rebase is bad.
* If this PR is merged with a merge commit, the git history is no longer a tree but a DAG with lattice.
* If we take the approach of re-creating a new release branch, all these 40 commits would be simply shared between the main and new branch - as if we just did the freeze now.

Issue #19298, opened 2024-11-07
Support using index in backfilling


hzxa21:
### Is your feature request related to a problem? Please describe.

> basically if we are creating a sink from A to B, and we know B or B's downstream prefer the backfilling rows to be ordered by X, we better add ORDER BY X to A if A is a MV, or create a new MV A' as SELECT .. from A ORDER BY X and create the sink from A'

Sometimes we need backfilling to emit rows in a specific ordering to reduce unnecessary downstream changes. Currently we can work around it by creating a MV with ORDER BY and use this MV for backfilling but this MV cannot be dropped after backfilling because it exists in the streaming graph. Instead, using index for backfilling seems more flexible.

xxchan:
This was also once mentioned here https://github.com/risingwavelabs/risingwave/issues/12824#issuecomment-1776519122

kwannoel:
> Currently we can work around it by creating a MV with ORDER BY and use this MV for backfilling but this MV cannot be dropped after backfilling because it exists in the streaming graph. Instead, using index for backfilling seems more flexible.

Do you mean that you'd like to use the index for backfilling side, but the MV for the upstream side? And then be able to drop the index after backfilling done?

xxchan:
Yes, the index is only useful for backfilling. Persisting data twice both in MV and index seems to wasteful.

BugenZhao:
> Persisting data twice both in MV and index seems to wasteful.

That sounds a bit arbitrary to me. Clustered index is essentially a trade-off between time and space, and may not be labeled "wasteful". 🤣 If a user plans to create multiple MVs that can be accelerated with this index, maintaining it (compared to make it "temporary" for a specific backfilling MV) can be highly beneficial.

BugenZhao:
> Do you mean that you'd like to use the index for backfilling side, but the MV for the upstream side? And then be able to drop the index after backfilling done?

Regarding implementation, I think the key point is to allow removing the reference to the table to be scanned once the backfill is complete. Whether the table is an index or a materialized view with `ORDER BY` does not make much difference.

kwannoel:
Lower priority. Users may rely on index not just during backfill, but during streaming phase, since they index can control stream distribution after #19321.

They can already create stream jobs directly on index for this usecase.

PR #23092, opened 2025-09-04
feat(optimizer): support index selection for backfilling (part 1)

chenzl25:

- https://github.com/risingwavelabs/risingwave/issues/19298
- Support index selection for streaming queries backfilling with a better locality.
- `fn try_better_locality(&self, columns: &[usize]) -> Option<LogicalPlanRef>` is a new method provided to be called if there is a better plan can be used. Typically if the index could meet this requirement, it would be selected.
- For a better review, we only support this method for the following operators: LogicalScan, LogicalJoin, LogicalAgg, LogicalOverWindow, LogicalProject and LogicalFilter. Later we can add more operators.
- Interestingly, some of our planner tests could be improved by this optimization as well.
- For LogicalScan, we will enumerate all possible orders against all indexes to fulfill the requirement.
- Add a session variable `enable_index_selection` to control whether to enable index selection for both streaming and batch queries.

## Checklist

## Documentation

<details>

- Add a session variable `enable_index_selection` to control whether to enable index selection for both streaming and batch queries.

</details>


e2e_test/batch/catalog/rw_depend.slt.part
```diff
@@ -34,6 +34,7 @@ join rw_relations r1 on d.objid = r1.id
 join rw_relations r2 on d.refobjid = r2.id;
 ----
 idx1  t1
+mv1 idx1
```
chenzl25, line 40: It seems we will still capture the dependencies of `mv1 -> t1`, if we select the index idx1 instead of t1. I think it is acceptable, but where do we record this dependency? If I create the materialized view by using the index explictly like ` create materialized view mv1 as select idx1.a from idx1 join s1 on idx1.a = s1.a;` then there is no dependency of `mv1 -> t1` cc @yezizp2012

yezizp2012: The dependency info will be resolved during both binding and planning stage. If we explicitly specify the index in the mv's definition then the index will be the only referenced object.

BugenZhao: > It seems we will still capture the dependencies of `mv1 -> t1`, if we select the index idx1 instead of t1. I think it is acceptable

Will this allow the index to be dropped without checking presence of materialized view atop of it, which sounds problematic?

src/frontend/src/optimizer/plan_node/logical_project.rs
```diff
@@ -390,6 +390,21 @@ impl ToStream for LogicalProject {
...
+        if columns.is_empty() {
+            return None;
+        }
+        let input_columns: Vec<usize> = columns
+            .iter()
+            .filter_map(|&col| self.o2i_col_mapping().try_map(col))
+            .collect();
+        if input_columns.len() != columns.len() {
+            return None;
+        }
```
BugenZhao:
```suggestion
            .map(|&col| self.o2i_col_mapping().try_map(col))
            .collect()?
```

src/frontend/src/optimizer/plan_node/logical_scan.rs
```diff
@@ -670,4 +670,34 @@ impl ToStream for LogicalScan {
...
+    fn try_better_locality(&self, columns: &[usize]) -> Option<PlanRef> {
+        if columns.is_empty() {
+            return None;
+        }
+        let order_types = [
+            OrderType::ascending_nulls_first(),
+            OrderType::ascending_nulls_last(),
+            OrderType::descending_nulls_first(),
+            OrderType::descending_nulls_last(),
+        ];
```
BugenZhao: Shall we make it a function on `OrderType` like `OrderType::all()`?

e2e_test/batch/catalog/rw_depend.slt.part
chenzl25: No. This above issue means mv1 not only depends on the index idx1, but also depends on the table t1. So the index can't be dropped without dropping the mv first.

BugenZhao (commented):
Great work! However, I'm wondering if we can achieve the same functionality of `try_better_locality` by adopting existing patterns like optimizer rule or plan rewriter, so that there's no need to introduce a new method on `PlanRef`.. 🤔

e2e_test/batch/catalog/rw_depend.slt.part
BugenZhao: I see. "not only"

BugenZhao (commented):
Shall we also introduce a variable/option to opt out this feature, for debugging purposes?

e2e_test/batch/catalog/rw_depend.slt.part
yezizp2012: > If I create the materialized view by using the index explictly like  create materialized view mv1 as select idx1.a from idx1 join s1 on idx1.a = s1.a; then there is no dependency of mv1 -> t1

I do found an issue for this case, the table `t1` can be dropped in restrict mode, which is problematic and should be rejected. I opened a PR to fix it: #23101 .

chenzl25:
> Great work! However, I'm wondering if we can achieve the same functionality of `try_better_locality` by adopting existing patterns like optimizer rule or plan rewriter, so that there's no need to introduce a new method on `PlanRef`.. 🤔

Because we don't have a CBO but only a RBO optimizer, we can't use simply a optimizer rule to do such thing, so we need to pass this information(locality columns) down through the plan tree to enumerate different plans. Even `PlanRewriter` itself relies on a method `clone_with_inputs` of `PlanTreeNode`, but this method lack of way to pass the locality information. So I think implement a method on the PlanRef would be more appropriate.

chenzl25:
> Shall we also introduce a variable/option to opt out this feature, for debugging purposes?

Yes. I think we can introduce an session variable like `enable_index_selection` to opt out the index selection feature.

src/frontend/src/optimizer/plan_node/logical_scan.rs
chenzl25: Sure~

src/frontend/src/optimizer/plan_node/convert.rs
```diff
@@ -55,6 +55,10 @@ pub trait ToStream {
         let ret = self.to_stream(ctx)?;
         required_dist.streaming_enforce_if_not_satisfies(ret)
     }
+
+    fn try_better_locality(&self, _columns: &[usize]) -> Option<LogicalPlanRef> {
+        None
+    }
```
BugenZhao, line 61: I saw a lot of occurrence of the same pattern in this PR, which looks like

```rust
let input = if let Some(better_plan) = self.input().try_better_locality(<key>) {
    better_plan
} else {
    self.input()
};
```

Then shall we consider to change the signature and default impl into 
```rust
fn try_better_locality(&self, _columns: &[usize]) -> LogicalPlanRef {
    self.clone()
}
```

So that using the unmodified plan will be the fallback, and we don't need the branching everywhere.

chenzl25: Indeed, good idea.

chenzl25: Oh, but we can't call clone to produce a `LogicalPlanRef` inside the `ToStream` trait and it seems need to put more efforts on an operator to detect the case of impossible transformation and return the clone. Previously we can just use `?` to return a None.

chenzl25: Let's use this unwrap_or_else to replace the branch.
```
let input = self
    .input()
    .try_better_locality(&self.group_key().to_vec())
    .unwrap_or_else(|| self.input());
```

chenzl25:
* **#23174** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23174?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#23171** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23171?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#23128** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23128?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#23092** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23092?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23092?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

BugenZhao (approved):
LGTM!

PR #23096, opened 2025-09-05
feat(sink): skip UpdateDelete and set application name in client

wenym1:

Skip handling `UpdateDelete` in postgres sink, as update should be treated as upsert. Current logic will issue unnecessary intermediate delete to the downstream table.

Besides, set `application_name` in the pg client, so that the client status can be visible in downstream pg.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/common/src/array/stream_chunk_iter.rs
+++ b/src/common/src/array/stream_chunk_iter.rs
@@ -12,12 +12,16 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::iter::from_coroutine;
 use std::ops::Range;
+
+use tracing::warn;
 
 use super::RowRef;
 use super::data_chunk_iter::DataChunkRefIter;
 use super::stream_record::Record;
 use crate::array::{Op, StreamChunk};
+use crate::row::RowExt;
 
 impl StreamChunk {
     /// Return an iterator on stream records of this stream chunk.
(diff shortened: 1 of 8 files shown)


wenym1:
* **#23096** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23096?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23096?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/connector/src/sink/postgres.rs
```diff
@@ -462,6 +466,7 @@ impl PostgresSinkWriter {
                     };
                     upsert_futures.push(future);
                 }
+                Op::UpdateDelete => {} /* Update can be treated as Upsert, and therefore no need to handle UpdateDelete */
```
xiangjinwu: Is it possible UpdateDelete and UpdateInsert has different primary key / updates the primary key column?

src/connector/src/source/cdc/external/postgres.rs
```diff
@@ -177,6 +177,7 @@ impl PostgresExternalTableReader {
             &config.database,
             &config.ssl_mode,
             &config.ssl_root_cert,
+            "RisingWave-Postgres-Reader",
```
xiangjinwu: Do we also want to include some kind of id?

src/connector/src/sink/postgres.rs
wenym1: We can assume that in the input the pk of UpdateDelete and UpdateInsert should on the same pk. When upstream input of sink executor does not match the pk of downstream sink, we will do per-barrier buffer for convert the update on upstream pk to update on downstream sink pk when necessary.

src/connector/src/sink/postgres.rs
```diff
@@ -286,6 +289,7 @@ impl PostgresSinkWriter {
             &config.database,
             &config.ssl_mode,
             &config.ssl_root_cert,
+            format!("RisingWave-Postgres-Sink-{}", sink_id).as_str(),
```
tabVersion: shall we also include sink name? It is a FAQ in source: the id is refer to which source?

src/connector/src/sink/postgres.rs
tabVersion: Does the behavior change? Correct me if I am wrong

If the user defines `on conflict do nothing` in the downstream table

* prev impl: we do delete and insert => do update. 
* current impl: we do nothing on update_delete and insert only => no update

wenym1: The on conflict behavior is defined in the UPDATE dml we issue to the downstream table. The dml we issue to downstream table is like the following. 
```
INSERT INTO ... ON CONFLICT ... DO UPDATE
```

The handling logic is the same for Insert and UpdateInsert.

src/connector/src/sink/postgres.rs
wenym1: Included schema name and table name in reader application name, and sink name in writer application name

src/connector/src/sink/postgres.rs
xiangjinwu: I meant the following, where both upstream and downstream use `v1` as pk and the sink sees this chunk: #12539
```
+----+----+----+
|    | v1 | v2 |
+----+----+----+
| U- | 12 | 23 |
| U+ | 13 | 23 |
+----+----+----+
```

wenym1: Surprised to konw that this is not reliable. Will add a check on consistent PK before we do this optimization.

wenym1:
Added logic to buffer `UpdateDelete` and wait for coming `UpdateInsert` to check whether the two pk are matched or not, and only ignore the `UpdateDelete` when the two pk are matched. @xiangjinwu PTAL

src/common/src/array/stream_chunk_iter.rs
```diff
@@ -28,13 +32,96 @@ impl StreamChunk {
         }
     }
 
+    fn consistent_pk_update<'a>(
+        rows: impl Iterator<Item = (Op, RowRef<'a>)> + 'a,
+        pk_indices: &'a [usize],
+    ) -> impl Iterator<Item = (Op, RowRef<'a>)> + 'a {
+        from_coroutine(
+            #[coroutine]
```
kwannoel, line 40: Cool, this is a nice way to define an iterator.

kwannoel (approved):

src/common/src/array/stream_chunk_iter.rs
```diff
@@ -28,13 +32,96 @@ impl StreamChunk {
...
+            move || {
+                let mut update_delete_buffer = None;
+                for (op, row) in rows {
```
BugenZhao, line 43: I guess you can reuse `StreamChunk::records`, which already checks the `Op` of updates are valid.

Just like `update_check` wrapper:

https://github.com/risingwavelabs/risingwave/blob/0ed891ecb75c93d1cbca869bc9e86b6271a029f5/src/stream/src/executor/wrapper/update_check.rs#L24-L41

BugenZhao (commented):
The changes are indeed not necessary any more after #23581. For `UPSERT` sink, we will only write rows with `Insert` and `Delete` op to the sink impl.

And the `Op` for updates on different keys will also be sanitized by applying into and taking back from the `ChangeBuffer`.

https://github.com/risingwavelabs/risingwave/blob/0ed891ecb75c93d1cbca869bc9e86b6271a029f5/src/stream/src/common/compact_chunk.rs#L53-L60

agrobbin:
@BugenZhao is it possible for this to be updated to *just* set the application name on PG clients (for both source and sink connections)? It'd be great for that enhancement to land!

hzxa21:
> @BugenZhao is it possible for this to be updated to _just_ set the application name on PG clients (for both source and sink connections)? It'd be great for that enhancement to land!

Good point. I have opened a separate PR for this. @agrobbin 

https://github.com/risingwavelabs/risingwave/pull/26968

PR #23103, opened 2025-09-05
feat(cdc): record cdc auto schema change failure events in `rw_event_log`

wcy-fdu:

As title

Show cdc auto schema change fail event in rw_event_logs:
```
select * from rw_event_logs where event_type = 'AUTO_SCHEMA_CHANGE_FAIL' order by timestamp desc;
```
Example1: fail due to unsupported type
```
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
unique_id  | 766a80d9-89c2-4a75-aef4-56ed7068bd86
timestamp  | 2025-09-10 06:19:32.969+00:00
event_type | AUTO_SCHEMA_CHANGE_FAIL
info       | {"autoSchemaChangeFail": {"cdcTableId": "s.s.public.t", "failInfo": "Unsupported data type 'hstore' in source 's' table 's.public.t'", "tableId": 6, "tableName": "s.public.t"}}
```

Example2: fail due to the ddl is not add column or drop column
```
unique_id  | 4792fc05-54ae-44dc-bcda-0bdb5d63aaea
timestamp  | 2025-09-05 10:48:22.691+00:00
event_type | AUTO_SCHEMA_CHANGE_FAIL
info       | {"autoSchemaChangeFail": {"cdcTableId": "6.public.t", "failInfo": "New columns should be a subset or superset of the original columns (including hidden columns), since only `ADD COLUMN` and `DROP COLUMN` is supported", "tableId": 8, "tableName": "t", "upstreamDdl": "UNKNOWN_DDL"}}
```

## Checklist

## 

<details>

</details>

### DIFF
--- a/src/common/src/config/streaming.rs
+++ b/src/common/src/config/streaming.rs
@@ -239,6 +239,11 @@ pub struct StreamingDeveloperConfig {
     /// Only takes effect when `default_enable_mem_preload_state_table` is true.
     #[serde(default)]
     pub mem_preload_state_table_ids_blacklist: Vec<u32>,
+
+    /// Eliminate unnecessary updates aggressively, even if it impacts performance. Enable this
+    /// only if it's confirmed that no-op updates are causing significant streaming amplification.
+    #[serde(default)]
+    pub aggressive_noop_update_elimination: bool,
 }
 
 pub mod default {
--- a/src/compute/src/server.rs
+++ b/src/compute/src/server.rs
@@ -134,6 +134,12 @@ pub async fn compute_node_serve(
         &config.meta,
     )
     .await;
+    // TODO(shutdown): remove this as there's no need to gracefully shutdown the sub-tasks.
+    let mut sub_tasks: Vec<(JoinHandle<()>, Sender<()>)> = vec![];
+    sub_tasks.push(MetaClient::start_heartbeat_loop(
+        meta_client.clone(),
+        Duration::from_millis(config.server.heartbeat_interval_ms as u64),
+    ));
 
     let state_store_url = system_params.state_store();
 
@@ -172,8 +178,6 @@ pub async fn compute_node_serve(
     let worker_id = meta_client.worker_id();
     info!("Assigned worker node id {}", worker_id);
 
-    // TODO(shutdown): remove this as there's no need to gracefully shutdown the sub-tasks.
-    let mut sub_tasks: Vec<(JoinHandle<()>, Sender<()>)> = vec![];
     // Initialize the metrics subsystem.
     let source_metrics = Arc::new(GLOBAL_SOURCE_METRICS.clone());
     let hummock_metrics = Arc::new(GLOBAL_HUMMOCK_METRICS.clone());
@@ -282,11 +286,6 @@ pub async fn compute_node_serve(
                 .await;
         });
     }
-
-    sub_tasks.push(MetaClient::start_heartbeat_loop(
-        meta_client.clone(),
-        Duration::from_millis(config.server.heartbeat_interval_ms as u64),
-    ));
 
     // Initialize the managers.
     let batch_mgr = Arc::new(BatchManager::new(
--- a/src/config/example.toml
+++ b/src/config/example.toml
@@ -201,6 +201,7 @@ stream_iceberg_sink_write_parquet_max_row_group_rows = 100000
 stream_default_enable_mem_preload_state_table = false
 stream_mem_preload_state_table_ids_whitelist = []
 stream_mem_preload_state_table_ids_blacklist = []
+stream_aggressive_noop_update_elimination = false
 
 [streaming.developer.stream_compute_client_config]
 connect_timeout_secs = 5
--- a/src/frontend/planner_test/tests/testdata/input/index_selection_for_backfilling.yaml
+++ b/src/frontend/planner_test/tests/testdata/input/index_selection_for_backfilling.yaml
@@ -126,3 +126,50 @@
     select id1, a1, id2, a2 from stream left join version FOR SYSTEM_TIME AS OF PROCTIME() on a1 = a2 and b1 = b2;
   expected_outputs:
     - stream_plan
+- sql: |
+    create table t(a int, b int, c int);
+    SELECT * FROM (
+    SELECT
+        *,
+        row_number() OVER (PARTITION BY a ORDER BY b) AS rank
+        FROM t
+    ) WHERE rank <= 1;
+  expected_outputs:
+    - stream_plan
+- sql: |
+    create table t(a int, b int, c int);
+    create index idx_a on t(a);
+    SELECT * FROM (
+    SELECT
+        *,
+        row_number() OVER (PARTITION BY a ORDER BY b) AS rank
+        FROM t
+    ) WHERE rank <= 1;
+  expected_outputs:
+    - stream_plan
+- sql: |
+    create table t(a int, b int, c int) append only;
+    select distinct on(a) * from t ;
+  expected_outputs:
+    - stream_plan
+- sql: |
+    create table t(a int, b int, c int) append only;
+    create index idx_a on t(a);
+    select distinct on(a) * from t ;
+  expected_outputs:
+    - stream_plan
+- sql: |
+    create table t1(a int, b int, c int);
+    create table t2(a int, b int, c int);
+    create index idx1_a on t1(a);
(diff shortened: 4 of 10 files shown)


src/connector/src/parser/unified/debezium.rs
```diff
@@ -213,6 +213,10 @@ pub fn parse_schema_change(
             };
             let id = jsonb_access_field!(jsonb, "id", string);
             let ty = jsonb_access_field!(jsonb, "type", string);
+            // Try to extract table name from the JSON data
+            let table_name = jsonb_access_field!(jsonb, "id", string)
```
hzxa21: You are accessing the "id" field here (duplicate with L214), not the table name

src/connector/src/parser/unified/debezium.rs
```diff
@@ -241,10 +245,14 @@ pub fn parse_schema_change(
                                         tracing::warn!(error=%err.as_report(), "unsupported postgres type in schema change message");
                                         AccessError::UnsupportedType {
                                             ty: type_name.clone(),
+                                            table_name: Some(table_name.clone()),
```
hzxa21: unnecessary clone

src/connector/src/parser/unified/debezium.rs
```diff
@@ -241,10 +245,14 @@ pub fn parse_schema_change(
...
                                         }
                                     })?,
                                     None => {
-                                        Err(AccessError::UnsupportedType { ty: type_name.clone() })?
+                                        return Err(AccessError::UnsupportedType {
+                                            ty: type_name,
+                                            table_name: Some(table_name.clone()),
```
hzxa21: unnecessary clone

src/connector/src/parser/unified/debezium.rs
```diff
@@ -256,10 +264,14 @@ pub fn parse_schema_change(
                                     tracing::warn!(error=%err.as_report(), "unsupported mysql type in schema change message");
                                     AccessError::UnsupportedType {
                                         ty: type_name.clone(),
+                                        table_name: Some(table_name.clone()),
```
hzxa21: unnecessary clone

src/connector/src/parser/unified/debezium.rs
```diff
@@ -256,10 +264,14 @@ pub fn parse_schema_change(
...
                                     }
                                 })?,
                                 None => {
-                                    Err(AccessError::UnsupportedType { ty: type_name })?
+                                    Err(AccessError::UnsupportedType {
+                                        ty: type_name,
+                                        table_name: Some(table_name.clone()),
```
hzxa21: unnecessary clone

src/batch/executors/src/executor/source.rs
```diff
@@ -154,6 +154,7 @@ impl SourceExecutor {
             },
             ConnectorProperties::default(),
             None,
+            None, // No callback for batch executor
```
hzxa21: There a simple way tou can avoid changing all the places where `SourceContext::new` is used:
1. impl `SourceContext::new_with_auto_schema_change_callback` method that accept the callback function to construct `SourceContext`
2. Let `SourceContext::new` use `SourceContext::new_with_auto_schema_change_callback(..,..,..None)`
3. Only use `new_with_auto_schema_change_callback` when needed.

src/connector/codec/src/decoder/mod.rs
```diff
@@ -34,7 +34,10 @@ pub enum AccessError {
         value: String,
     },
     #[error("Unsupported data type `{ty}`")]
```
hzxa21, line 36: This needs to be changed as well.

src/connector/codec/src/decoder/mod.rs
```diff
@@ -34,7 +34,10 @@ pub enum AccessError {
...
-    UnsupportedType { ty: String },
+    UnsupportedType {
+        ty: String,
+        table_name: Option<String>,
```
hzxa21: I think it is better to introduce a new Error instead of extending on UnsupportedType because `table_name` is specific to cdc auto schema change only.

src/connector/src/parser/plain_parser.rs
```diff
@@ -136,7 +137,53 @@ impl PlainParser {
...
+                                    ty,
+                                    table_name,
+                                    ..
+                                } => {
+                                    if let Some(table_name) = table_name {
+                                        // Parse table_name format: "schema"."table" -> schema.table
+                                        let clean_table_name =
+                                            table_name.trim_matches('"').replace("\".\"", ".");
+                                        let fail_info = format!(
+                                            "Unsupported postgres type '{}' in table '{}'",
```
hzxa21: "postgres type" is confusing because this logic is common for different upstream

src/connector/src/parser/plain_parser.rs
```diff
@@ -136,7 +137,53 @@ impl PlainParser {
...
+                                        // Build cdc_table_id: source_name.schema.table_name
+                                        let cdc_table_id = format!(
+                                            "{}.{}",
+                                            self.source_ctx.source_name, clean_table_name
+                                        );
+
+                                        (fail_info, clean_table_name, cdc_table_id)
+                                    } else {
+                                        let fail_info =
+                                            format!("Unsupported postgres type: {:?}", ty);
```
hzxa21: ditto

src/connector/src/parser/plain_parser.rs
```diff
@@ -136,7 +137,53 @@ impl PlainParser {
...
+                                        (fail_info, "".to_owned(), "".to_owned())
```
hzxa21: We have the source name available in `self.source_ctx.source_name`. We should include that in the error message.

src/connector/src/parser/plain_parser.rs
```diff
@@ -136,7 +137,53 @@ impl PlainParser {
...
+                                    }
+                                }
+                                _ => {
+                                    let fail_info = format!(
+                                        "Failed to parse schema change: {:?}",
+                                        err.as_report()
+                                    );
+                                    (fail_info, "".to_owned(), "".to_owned())
```
hzxa21, line 170: ditto

src/connector/src/source/base.rs
```diff
@@ -291,6 +323,8 @@ pub struct SourceContext {
     // source parser put schema change event into this channel
     pub schema_change_tx:
         Option<mpsc::Sender<(SchemaChangeEnvelope, tokio::sync::oneshot::Sender<()>)>>,
+    // callback function to report CDC auto schema change fail events
+    pub report_cdc_auto_schema_change_fail: Option<CdcAutoSchemaChangeFailCallback>,
```
hzxa21: nits on naming: `on_cdc_auto_schema_change_failure`

src/stream/src/executor/source/reader_stream.rs
```diff
@@ -90,6 +91,43 @@ impl StreamReaderBuilder {
             None
         };
 
+        // Create callback function for reporting CDC auto schema change fail events
+        let report_cdc_auto_schema_change_fail =
```
hzxa21: We should put this under the if clause in L56 because it is only valid when auto schema change is enabled. Also, let's move all the auto schema change related logic including the ones in L56-92 to a separate method to make the code cleaner.

src/connector/codec/src/decoder/mod.rs
wcy-fdu: Actually there's only one place besides CDC where this UnsupportedType is used, and that's the CSV parser. Anyways, I agree with introducing a new Error.

wcy-fdu:
This pr can be reviewed. cc @zwang28 @hzxa21 ❤️

src/connector/src/parser/plain_parser.rs
```diff
@@ -133,10 +134,52 @@ impl PlainParser {
...
+                                _ => {
+                                    let fail_info = format!(
+                                        "Failed to parse schema change: {:?}, source: {}",
+                                        err.as_report(),
+                                        self.source_ctx.source_name
+                                    );
+                                    (fail_info, "".to_owned(), "".to_owned())
+                                }
+                            };
+                            self.source_ctx.on_cdc_auto_schema_change_failure(
```
zwang28, line 173: If a failure occurs, will it continuously generate a large volume of event logs until the issue is resolved?

wcy-fdu: No, only one event.

PR #23117, opened 2025-09-08
fix(cdc): change default ssl mode to `prefer`

wcy-fdu:

PgSslMode::prefer means establishing an encrypted connection if the server supports it, and fall back to an unencrypted connection if an encrypted connection cannot be established, and it is also a recommended practice. 
I think it is safe to set it to prefer at least for user-defined disable and prefer.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/src/connector_common/postgres.rs
+++ b/src/connector/src/connector_common/postgres.rs
@@ -120,8 +120,8 @@ impl PostgresExternalTable {
             .port(port)
             .database(database)
             .ssl_mode(match ssl_mode {
-                SslMode::Disabled => PgSslMode::Disable,
-                SslMode::Preferred => PgSslMode::Prefer,
+                // Default behavior is set to `PgSslMode::Prefer, because it will first try an SSL connection; if that fails, try a non-SSL connection.
+                SslMode::Disabled | SslMode::Preferred => PgSslMode::Prefer,
                 SslMode::Required => PgSslMode::Require,
                 SslMode::VerifyCa => PgSslMode::VerifyCa,
                 SslMode::VerifyFull => PgSslMode::VerifyFull,
--- a/src/connector/src/source/cdc/external/mysql.rs
+++ b/src/connector/src/source/cdc/external/mysql.rs
@@ -89,7 +89,9 @@ impl MySqlExternalTable {
             .port(config.port.parse::<u16>().unwrap())
             .database(&config.database)
             .ssl_mode(match config.ssl_mode {
-                SslMode::Disabled | SslMode::Preferred => sqlx::mysql::MySqlSslMode::Disabled,
+                // Default behavior is set to Preferred, because it will establish an encrypted connection if the server supports it,
+                // and fall back to an unencrypted connection if an encrypted connection cannot be established.
+                SslMode::Disabled | SslMode::Preferred => sqlx::mysql::MySqlSslMode::Preferred,
                 SslMode::Required => sqlx::mysql::MySqlSslMode::Required,
                 _ => {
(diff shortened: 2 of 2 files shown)


wcy-fdu:
I think in [lakekeeper services](https://github.com/risingwavelabs/risingwave/blob/74fa792648bc045049fb3d8dc74bd1e5f0a3efc0/src/risedevtool/src/task/lakekeeper_service.rs#L111), should also set to prefer. WDYT @chenzl25

src/connector/src/connector_common/postgres.rs
```diff
@@ -120,8 +120,8 @@ impl PostgresExternalTable {
             .port(port)
             .database(database)
             .ssl_mode(match ssl_mode {
-                SslMode::Disabled => PgSslMode::Disable,
-                SslMode::Preferred => PgSslMode::Prefer,
+                // Default behavior is set to `PgSslMode::Prefer, because it will first try an SSL connection; if that fails, try a non-SSL connection.
+                SslMode::Disabled | SslMode::Preferred => PgSslMode::Prefer,
```
hzxa21: This seems wrong. If user explicitly sets ssl mode to disabled. Regardless of whether the server supports encryption, we should not use SSL.

wcy-fdu: I thought it is safe, because client will first try encryption, and then fall back to not use SSL. I made this change is because due to compatibility issues,[ we set the default behavior of SSL mode to disable](https://github.com/risingwavelabs/risingwave/blob/28fafe88e13b7e516461639d235e7182784b88de/src/connector/src/source/cdc/external/mod.rs#L281). However, in many cases (such as when using RDS), disable does not work. In such cases, the user has not configured ssl.mode, making it difficult to detect the problem.

chenzl25:
> I think in [lakekeeper services](https://github.com/risingwavelabs/risingwave/blob/74fa792648bc045049fb3d8dc74bd1e5f0a3efc0/src/risedevtool/src/task/lakekeeper_service.rs#L111), should also set to prefer. WDYT @chenzl25

Sure. I think you can change it in this PR directly.

src/connector/src/connector_common/postgres.rs
chenzl25: Can we change the default behavior directly, if it is safe?

hzxa21: My comment is not about whether it is safe or not, but about whether we respect the user configured option. For example, if use explicitly runs:
```
create source ...
with (
  ssl.mode='disabled'
)
```

The expected behavior should be using non-SSL to connect even though the server supports SSL. What you are referring to is about whether `disabled` is the right default behavior and I agree it is not and `prefer` is a better default. The right way to resolve this is to come up with a way to change the default value only for source created after `2.6` instead of converting disabled to preferred,

tabVersion: > Can we change the default behavior directly, if it is safe?

+1, the new default should only take effect on new created ones. There should be no compatibility issue.

wcy-fdu: >The right way to resolve this is to come up with a way to change the default value only for source created after 2.6 instead of converting disabled to preferred,

Got your point. Maybe we need to find a way to distinguish whether the `Disable` is configured by the user or the default value for the previous source. Anyways, let's keep this PR the same as before.

wcy-fdu: >Can we change the default behavior directly, if it is safe?

To be honest, I'm not sure about the historical reasons for the compatibility issues here, so it's hard to judge whether changing the default behavior is dangerous.

PR #23119, opened 2025-09-08
feat(frontend): add a column for `schema_name` in output of `SHOW [TABLE|..]S`

BugenZhao:

When there's no `FROM <schema_name>` specified, we will go through all schemas in the search path to list all objects. Considering that there could be objects with the same name in different schemas, we should include the schema name in the output to avoid confusion.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/e2e_test/ddl/secret.slt
+++ b/e2e_test/ddl/secret.slt
@@ -68,10 +68,10 @@ Caused by these errors (recent errors listed first):
 # backend = 'hashicorp_vault'
 # );
 
-query T
+query TT
 show secrets;
 ----
-secret_1
+public secret_1
 
 statement ok
 drop secret secret_1;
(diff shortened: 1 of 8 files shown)


yezizp2012 (approved):
Nits: How about using the `schema.t1` format for consistency with other show commands? 
Adding a column is also acceptable to me, since `function`, `connection` also has multiple columns.

BugenZhao:
> Nits: How about using the `schema.t1` format for consistency with other show commands?

Yeah this also sounds reasonable. Actually I'm following the interface of CockroachDB 😁: https://www.cockroachlabs.com/docs/stable/show-tables

BugenZhao:
> Nits: How about using the `schema.t1` format for consistency with other show commands?

I've updated this PR to follow your proposal. It indeed leads to fewer changes and has better compatibility guarantees.

PR #23143, opened 2025-09-09
fix: set audience for oauth validation if present

### LINKED ISSUE #23126: OAuth audience validation fails when JWT contains "aud" claim
### Describe the bug

OAuth user authentication fails with `InvalidAudience` error when JWTs contain an "aud" claim.

### To Reproduce

```sql
CREATE USER test WITH oauth (
  jwks_url = 'https://...',
  issuer = 'https://...'
);
```
Authentication fails with:
```
ERROR: Failed to start a new session: InvalidAudience
```

Even when adding audience parameters:

```sql
CREATE USER test WITH oauth (
  jwks_url = 'https://...',
  issuer = 'https://...',
  aud = 'developer'
);

```

Still fails because the JWT library's audience validation runs before RisingWave's manual claim checking.

### Expected Behavior

According to the documentation:
`other_params_should_match` is an optional parameter that will be validated against `jwt.claims`

RisingWave should disable the JWT library's automatic audience validation and handle audience checking manually through its existing metadata validation logic.

Tested locally and via cluster.

### The version of RisingWave

PostgreSQL 13.14.0-RisingWave-2.5.1 

### Additional context

In `src/utils/pgwire/src/pg_server.rs`, the JWT validation uses the `jsonwebtoken` crate's default settings:  `validation.validate_aud = true` (default) and no expected audience set with `validation.set_audience()`

This means the library expects JWTs to have no `aud` claim at all even when it's specified in `other_params_should_match`.

yezizp2012:

Fixed #23126 .
If the input OAuth token contains an audience claim, the `aud` parameter will be required and validated. It's very insecure to turn off `validate_aud` option. Since `aud` is not included in `required_spec_claims` bydefault,  if the user's login token lacks an audience field, validation will not be triggered even if the option is enabled.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/utils/pgwire/src/pg_server.rs
+++ b/src/utils/pgwire/src/pg_server.rs
@@ -19,6 +19,7 @@ use std::sync::Arc;
 use std::time::Instant;
 
 use bytes::Bytes;
+use itertools::Itertools;
 use jsonwebtoken::{Algorithm, DecodingKey, Validation, decode, decode_header};
 use parking_lot::Mutex;
 use risingwave_common::types::DataType;
@@ -224,7 +225,8 @@ async fn validate_jwt(
     validation.set_issuer(&[issuer]);
     validation.set_required_spec_claims(&["exp", "iss"]);
     if let Some(audience) = audience {
-        validation.set_audience(&[&audience]);
+        let audiences = audience.split(',').map(|au| au.trim()).collect_vec();
+        validation.set_audience(&audiences);
     }
     let token_data = decode::<HashMap<String, serde_json::Value>>(jwt, &decoding_key, &validation)?;
 
--- a/src/utils/pgwire/src/pg_server.rs
+++ b/src/utils/pgwire/src/pg_server.rs
@@ -199,6 +199,7 @@ async fn validate_jwt(
     jwt: &str,
     jwks_url: &str,
     issuer: &str,
+    audience: Option<String>,
     metadata: &HashMap<String, String>,
 ) -> Result<bool, BoxedError> {
     let header = decode_header(jwt)?;
@@ -222,6 +223,9 @@ async fn validate_jwt(
     let mut validation = Validation::new(header.alg);
     validation.set_issuer(&[issuer]);
     validation.set_required_spec_claims(&["exp", "iss"]);
+    if let Some(audience) = audience {
+        validation.set_audience(&[&audience]);
+    }
     let token_data = decode::<HashMap<String, serde_json::Value>>(jwt, &decoding_key, &validation)?;
 
     // 4. Check if the metadata in the token matches.
@@ -245,10 +249,12 @@ impl UserAuthenticator {
                 let mut metadata = metadata.clone();
                 let jwks_url = metadata.remove("jwks_url").unwrap();
                 let issuer = metadata.remove("issuer").unwrap();
+                let audience = metadata.remove("aud");
                 validate_jwt(
                     &String::from_utf8_lossy(password),
                     &jwks_url,
                     &issuer,
+                    audience,
                     &metadata,
                 )
                 .await


marigov:
Just to double-check how this will work with SQL syntax: If there are multiple audiences, how should the user be created? Will the order matter?

```sql
CREATE USER user_name WITH oauth (
  jwks_url = 'xxx.com',
  issuer = 'risingwave',
  aud = 'aud-one aud-two aud-three',
);
```

yezizp2012:
> Just to double-check how this will work with SQL syntax: If there are multiple audiences, how should the user be created? Will the order matter?
> 
> ```sql
> CREATE USER user_name WITH oauth (
>   jwks_url = 'xxx.com',
>   issuer = 'risingwave',
>   aud = 'aud-one aud-two aud-three',
> );
> ```

Good question. After this PR, if there are multiple audiences, they should be concatenated with a comma. The `create user` syntax should then be formatted as follows:
```sql
CREATE USER user_name WITH oauth (
   jwks_url = 'xxx.com',
   issuer = 'risingwave',
   aud = 'aud-one,aud-two,aud-three',
```

Sorry for the inconvenience, we will merge this PR as soon as some local tests pass.

marigov:
> > Just to double-check how this will work with SQL syntax: If there are multiple audiences, how should the user be created? Will the order matter?
> > ```sql
> > CREATE USER user_name WITH oauth (
> >   jwks_url = 'xxx.com',
> >   issuer = 'risingwave',
> >   aud = 'aud-one aud-two aud-three',
> > );
> > ```
> 
> Good question. After this PR, if there are multiple audiences, they should be concatenated with a comma. The `create user` syntax should then be formatted as follows:
> 
> ```sql
> CREATE USER user_name WITH oauth (
>    jwks_url = 'xxx.com',
>    issuer = 'risingwave',
>    aud = 'aud-one,aud-two,aud-three',
> ```
> 
> Sorry for the inconvenience, we will merge this PR as soon as some local tests pass.

Thank you for the quick response! Will the order of the values matter? I'm not familiar with the spec of how JWT tokens are generated, but I wonder if this could be a problem if the order is not deterministic at the moment of generation and it's doing array comparison instead of set comparison?

yezizp2012:
> Thank you for the quick response! Will the order of the values matter? I'm not familiar with the spec of how JWT tokens are generated, but I wonder if this could be a problem if the order is not deterministic at the moment of generation and it's doing array comparison instead of set comparison?

Don't worry, the order won't matter. The `jsonwebtoken` crate validates all audiences using a `HashSet`, ensuring accurate comparisons.

yezizp2012:
> Just to double-check how this will work with SQL syntax: If there are multiple audiences, how should the user be created? Will the order matter?
> 
> ```sql
> CREATE USER user_name WITH oauth (
>   jwks_url = 'xxx.com',
>   issuer = 'risingwave',
>   aud = 'aud-one aud-two aud-three',
> );
> ```

Hi @marigov , could you please explain under what circumstances you would require a multiple audiences configuration? We've had an internal discussion, and for better security and to minimize user-side configuration errors, we lean towards having `RisingWave` declare its own identity rather than exposing the `aud` parameter in the `CREATE USER` syntax. Check #23155 for more details. Cc @yuhao-su

marigov:
> > Just to double-check how this will work with SQL syntax: If there are multiple audiences, how should the user be created? Will the order matter?
> > ```sql
> > CREATE USER user_name WITH oauth (
> >   jwks_url = 'xxx.com',
> >   issuer = 'risingwave',
> >   aud = 'aud-one aud-two aud-three',
> > );
> > ```
> 
> Hi @marigov , could you please explain under what circumstances you would require a multiple audiences configuration? We've had an internal discussion, and for better security and to minimize user-side configuration errors, we lean towards having `RisingWave` declare its own identity rather than exposing the `aud` parameter in the `CREATE USER` syntax. Check #23155 for more details. Cc @yuhao-su

Hi @yezizp2012, thanks for clarifying. Correct me if I’m wrong, but since JWT validation just checks that the expected audience is present in the list (not that all match), a single audience would be fine for us. That said, it might still be useful to keep some extensibility; e.g. if no audience is provided, RisingWave can declare its own identity, but if needed, users could still specify audiences explicitly for flexibility?

yezizp2012:
> > > Just to double-check how this will work with SQL syntax: If there are multiple audiences, how should the user be created? Will the order matter?
> > > ```sql
> > > CREATE USER user_name WITH oauth (
> > >   jwks_url = 'xxx.com',
> > >   issuer = 'risingwave',
> > >   aud = 'aud-one aud-two aud-three',
> > > );
> > > ```
> > 
> > 
> > Hi @marigov , could you please explain under what circumstances you would require a multiple audiences configuration? We've had an internal discussion, and for better security and to minimize user-side configuration errors, we lean towards having `RisingWave` declare its own identity rather than exposing the `aud` parameter in the `CREATE USER` syntax. Check #23155 for more details. Cc @yuhao-su
> 
> Hi @yezizp2012, thanks for clarifying. Correct me if I’m wrong, but since JWT validation just checks that the expected audience is present in the list (not that all match), a single audience would be fine for us. That said, it might still be useful to keep some extensibility; e.g. if no audience is provided, RisingWave can declare its own identity, but if needed, users could still specify audiences explicitly for flexibility?

That's a good point about extensibility. However, our view is that while providing the parameter might seem to offer more flexibility, it actually comes at the cost of security and increased complexity. Our decision not to expose the `aud` parameter is precisely to minimize user errors during usage, rather than shifting the responsibility for security onto the user.

marigov:
> That's a good point about extensibility. However, our view is that while providing the parameter might seem to offer more flexibility, it actually comes at the cost of security and increased complexity. Our decision not to expose the `aud` parameter is precisely to minimize user errors during usage, rather than shifting the responsibility for security onto the user.

Fair enough, it seems reasonable!

yezizp2012:
Use the approach of #23155 instead.

PR #23146, opened 2025-09-09
feat(optimizer): support UnifyFirstLastValueRule

chenzl25:

- Currently our `HashAggExecutor` of streaming queries will maintain states for first_value/last_value even though their pattern the same, but actually we can reuse the same state for these agg calls.
- However, it is hard to modify the executor code to reuse state and maintain the backward compatibility.
- Instead, in this PR, we support a optimizer rule called `UnifyFirstLastValueRule ` which would merge first_value/last_value with the same pattern will a `ROW` expression and extract the field after the aggregation, so we can "reuse" the state.

## Checklist

## Documentation

<details>

</details>


src/frontend/src/optimizer/rule/unify_first_last_value_rule.rs
```diff
@@ -0,0 +1,306 @@
...
+                .iter()
+                .map(|(_, call)| call.inputs[0].clone().into())
+                .collect();
+
+            let named_fields = mergeable_calls
+                .iter()
+                .enumerate()
+                .map(|(i, (_, call))| (format!("f{}", i), call.return_type.clone()))
+                .collect::<Vec<_>>();
+            let row_data_type = DataType::Struct(StructType::new(named_fields));
```
BugenZhao: Can use `StructType::unnamed`?

src/frontend/src/optimizer/rule/unify_first_last_value_rule.rs
```diff
@@ -0,0 +1,306 @@
...
+            }
+        }
+
+        // Sort mapping by original index to maintain order
+        original_to_output_mapping.sort_by_key(|(original_idx, _, _)| *original_idx);
+
+        if new_calls.len() >= calls.len() {
+            // No benefit from optimization
+            return None;
+        }
```
BugenZhao: When will we hit this path?

src/frontend/src/optimizer/rule/unify_first_last_value_rule.rs
```diff
@@ -0,0 +1,306 @@
...
+            if call.inputs.len() != 1 {
+                continue;
+            }
+
+            let pattern = AggPattern {
+                agg_type: call.agg_type.clone(),
+                order_by: call.order_by.clone(),
+                distinct: call.distinct,
+                filter: call.filter.clone(),
+            };
```
BugenZhao: Shall we destruct `AggCall` here, so that we will be notified to update here if there's any change on fields of `AggCall`?

chenzl25:
> IIUC this optimization won't help in terms of state size, but is primarily intended to reduce IO calls.

It would reduce the multiple key of these tables to one, so state size would be reduced.

src/frontend/src/optimizer/rule/unify_first_last_value_rule.rs
chenzl25: Indeed, let me remove this path.

chenzl25:
cc @BugenZhao for review.

BugenZhao (approved):
LGTM!

PR #23148, opened 2025-09-09
chore: ensure background ddl catalog visable after notification

### LINKED ISSUE #22884: ensure background ddl is visible in current session when it returns
Do we have plan to make the background ddl visible in frontend before it returns? Currently for background ddl, we return `IGNORED_NOTIFICATION_VERSION` to frontend. But actually on meta side, the background ddl returns after collecting the first barrier, and the creating job has been marked as `Creating` already. If we return a meaningful notification version for frontend to wait for, the background ddl job can be visible immediately in the same session. 

Not related to this PR. We can address in later PR.

_Originally posted by @wenym1 in https://github.com/risingwavelabs/risingwave/pull/22616#discussion_r2268598870_

yezizp2012:

Resolved #22884 

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/e2e_test/batch/catalog/rw_depend.slt.part
+++ b/e2e_test/batch/catalog/rw_depend.slt.part
@@ -30,7 +30,7 @@ create sink sink2 into t2 as select a from t1;
 
 # equivalent to:
 # select objid::regclass, refobjid::regclass from rw_depend;
-query TT rowsort retry 3 backoff 1s
+query TT rowsort
 select r1.name, r2.name relname
 from rw_depend d
 join rw_relations r1 on d.objid = r1.id
--- a/src/meta/src/controller/streaming_job.rs
+++ b/src/meta/src/controller/streaming_job.rs
@@ -78,7 +78,9 @@ use crate::controller::utils::{
     grant_default_privileges_automatically, insert_fragment_relations, list_user_info_by_ids,
 };
 use crate::error::MetaErrorInner;
-use crate::manager::{NotificationVersion, StreamingJob, StreamingJobType, IGNORED_NOTIFICATION_VERSION};
+use crate::manager::{
+    IGNORED_NOTIFICATION_VERSION, NotificationVersion, StreamingJob, StreamingJobType,
+};
 use crate::model::{
     FragmentDownstreamRelation, FragmentReplaceUpstream, StreamActor, StreamContext,
     StreamJobFragments, StreamJobFragmentsToCreate, TableParallelism,
--- a/src/meta/src/controller/streaming_job.rs
+++ b/src/meta/src/controller/streaming_job.rs
@@ -78,7 +78,7 @@ use crate::controller::utils::{
     grant_default_privileges_automatically, insert_fragment_relations, list_user_info_by_ids,
 };
 use crate::error::MetaErrorInner;
-use crate::manager::{NotificationVersion, StreamingJob, StreamingJobType};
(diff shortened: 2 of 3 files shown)


wenym1:
Actually, one purpose of the issue is to by the way refactor to unify some common logic for background and foreground ddl in `create_streaming_job_inner` and refine the implementation of `StreamManager::create_streaming_job`.

Currently inside `create_streaming_job_inner`, both foreground and background ddl call `StreamManager::create_streaming_job`. The `background` running feature is implemented as spawning the futuer of `StreamManager::create_streaming_job`. However, both foreground and background ddl have to wait for collecting the first barrier, and therefore we have to include a oneshot channel for background ddl to receive the notification on successful barrier collection, and such logic is somehow hacky.

```
match streaming_job.create_type() {
            CreateType::Unspecified | CreateType::Foreground => {
                let version = self
                    .stream_manager
                    .create_streaming_job(stream_job_fragments, ctx, None)
                    .await?;
                Ok(version)
            }
            CreateType::Background => {
                let await_tree_key = format!("Background DDL Worker ({})", stream_job_id);
                let await_tree_span =
                    span!("{:?}({})", streaming_job.job_type(), streaming_job.name());

                let ctrl = self.clone();
                let (tx, rx) = oneshot::channel();
                let fut = async move {
                    let _ = ctrl
                        .stream_manager
                        .create_streaming_job(stream_job_fragments, ctx, Some(tx))
                        .await
                        .inspect_err(|err| {
                            tracing::error!(id = stream_job_id, error = ?err.as_report(), "failed to create background streaming job");
                        });
                    // drop the permit to release the semaphore
                    drop(permit);
                };

                let fut = (self.env.await_tree_reg())
                    .register(await_tree_key, await_tree_span)
                    .instrument(fut);
                tokio::spawn(fut);

                rx.instrument_await("wait_background_streaming_job_creation_started")
                    .await
                    .map_err(|_| {
                        anyhow!(
                            "failed to receive create streaming job result of job: {}",
                            stream_job_id
                        )
                    })??;
                Ok(IGNORED_NOTIFICATION_VERSION)
            }
        }
```

A cleaner implementation to me will be like, we call `run_command` anyway for any create_type, and the only difference between foreground and background ddl is that foreground ddl has an extra call on `wait_streaming_job_finished` to wait for backfill finish.


```
// for any create type
let (source_change, streaming_job) = stream_manager
    .run_create_streaming_job_command(stream_job_fragments, ctx)
    .inspect(move |result| {
        if let Some(tx) = run_command_notifier {
            let _ = tx.send(match result {
                Ok(_) => {
                    Ok(())
                }
                Err(err) => {
                    Err(err.clone())
                }
            });
        }
    })
    .await?;
// only wait backfill for foreground
if create_type == Foreground {
  let version = stream_manager
      .metadata_manager
      .wait_streaming_job_finished(
          streaming_job.database_id().into(),
          streaming_job.id() as _,
      )
      .await?;
  version
} else {
   // get latest notification version of catalog
}
```

yezizp2012:
> A cleaner implementation to me will be like, we call run_command anyway for any create_type, and the only difference between foreground and background ddl is that foreground ddl has an extra call on wait_streaming_job_finished to wait for backfill finish.

👍 Agreed. I will do some refactoring on it.

wenym1:
Besides, in `StreamManager::create_streaming_job`, we have logic to spawn a future and use mpsc channel to handle user triggered cancel. It seems better to handle it with a `select!` and a oneshot channel. We can refine the implementation by the way in the refactor. @yezizp2012

PR #23150, opened 2025-09-09
[WIP] feat(frontend): support LDAP authentication

yezizp2012:

## Checklist

## Documentation

<details>

</details>


yezizp2012:
Hi @kwannoel the basic functions of LDAP authentication are now complete. I've successfully tested and integrated them locally. Big thanks to your [openldap](https://github.com/kwannoel/ldap-server) repository; it's been incredibly helpful. I believe you can proceed to review this PR first.

At the same time, I will continue to complete the following pending tasks:
1. Enhance the CI testing and ensure compatibility testing is in place.
2. TLS support, including connection detection, LDAP authentication, and relevant testing.
3. Try to implement all ldap configure options provided by PG as much as possible.

src/frontend/src/session.rs
```diff
@@ -1574,30 +1575,95 @@ impl SessionManagerImpl {
...
+            let hba_entry_opt = self.env.frontend_config().hba_config.find_matching_entry(
+                &connection_type,
+                database_name,
+                user_name,
+                client_addr,
+                false,
+            );
+
+            // TODO: adding `FATAL` message support for no matching HBA entry.
+            // Question: should we make the pg_hba.conf required like Postgres?
```
kwannoel: Doesn't the following check means that it's already required? Since if the opt is missing, we will return permission denied.

src/utils/pgwire/src/ldap_auth.rs
```diff
@@ -0,0 +1,313 @@
...
+    /// Establish an LDAP connection with configurable options
+    async fn establish_connection(&self) -> Result<ldap3::Ldap, LdapError> {
+        let config = &self.config;
+        let mut settings = ldap3::LdapConnSettings::new();
+
+        // Configure STARTTLS if specified
+        if config.start_tls {
+            if config.server.starts_with("ldaps://") {
+                return Err(LdapError::InvalidScopeString(
+                    "Cannot use STARTTLS with ldaps scheme".into(),
```
kwannoel: Can you elaborate on why we can't have both `ldaps` and `start_tls` enabled?

src/utils/pgwire/src/ldap_auth.rs
```diff
@@ -0,0 +1,313 @@
...
+            // 3. System Default Directories:
+            // - `/etc/ssl/certs/`
+            // - macOS system trust chain
+
+            const CA_CERT_PATH: &str = "/Users/august/Documents/codes/ldap-server/ldap/certs/ca.crt";
+            let ca_cert_pem = fs::read(CA_CERT_PATH).map_err(|e| {
+                LdapError::InvalidScopeString(format!("Failed to read CA certificate: {}", e))
+            })?;
+            let ca_cert = native_tls::Certificate::from_pem(&ca_cert_pem).map_err(|e| {
+                LdapError::InvalidScopeString(format!("Failed to parse CA certificate: {}", e))
```
kwannoel: The recommended way to structure errors is:
```
anyhow!(e).with_context("Failed to ...");
```

https://www.notion.so/risingwave-labs/A-Guide-to-Error-Handling-that-Just-Works-Part-I-e108f8d0845b4bddaa0e5843791a8cac?source=copy_link#bd13e10017014586bb702db39909e41a

This will preserve context.

src/utils/pgwire/src/ldap_auth.rs
```diff
@@ -0,0 +1,313 @@
...
+
+            // FIXME: on macOS, self-signed certificates are always rejected unless added to the system keychain.
+            // We need to replace native-tls with rustls to have more control over certificate validation.
+            builder.add_root_certificate(ca_cert);
+            #[cfg(target_os = "macos")]
+            // Accept self-signed certificates (development only)
+            builder.danger_accept_invalid_certs(true);
+
+            const CLIENT_CERT_PATH: &str = "/Users/august/Documents/codes/ldap-server/ldap/certs/client.crt";
+            const CLIENT_KEY_PATH: &str = "/Users/august/Documents/codes/ldap-server/ldap/certs/client.key";
```
kwannoel: Need update the hardcoded values later.

src/utils/pgwire/src/ldap_auth.rs
```diff
@@ -0,0 +1,313 @@
...
+        let base_dn = self.config
+            .base_dn
+            .as_ref()
+            .ok_or_else(|| PsqlError::StartupError("LDAP base_dn not configured".into()))?;
+        let search_filter = self.config
+            .search_filter
+            .as_ref()
+            .ok_or_else(|| PsqlError::StartupError("LDAP search_filter not configured".into()))?;
+
+        let search_filter = search_filter.replace("{username}", username);
```
kwannoel: Are there any risks of LDAP authentication injection?

src/utils/pgwire/src/ldap_auth.rs
```diff
@@ -0,0 +1,313 @@
...
+                ))
+            }
+        }
+    }
+
+    /// Simple bind authentication
+    async fn simple_bind(&self, username: &str, password: &str) -> PsqlResult<bool> {
+        // Construct DN from username
+        let dn = if let Some(base_dn) = &self.config.base_dn {
+            format!("uid={},{}", username, base_dn)
```
kwannoel: Do we need to sanitize the input?

kwannoel:
Can we have a tracking issue to describe the follow-up works? I noticed several TODOs in this PR.

src/frontend/src/session.rs
yezizp2012: Yes, `pg_hba.conf` is required for Postgres, If the opt is not porivded in rw, I am considering not returning permission denied, but to handle it as default or ignore it. Currently, the default and ignoring have the same effect.

src/utils/pgwire/src/ldap_auth.rs
yezizp2012: Refer https://www.postgresql.org/docs/current/auth-ldap.html.
> The URL scheme ldaps chooses the LDAPS method for making LDAP connections over SSL, equivalent to using ldapscheme=ldaps. To use encrypted LDAP connections using the StartTLS operation, use the normal URL scheme ldap and specify the ldaptls option in addition to ldapurl.

We cannot use `ldaps://` together with `ldaptls = 1` because they represent two mutually exclusive methods of securing the LDAP connection.
1. `ldaps://` means the connection is encrypted from the start using SSL/TLS (“LDAP over SSL”).
2. `ldap://` with `StartTLS` means the connection begins unencrypted and is then upgraded to TLS using the StartTLS operation.

Using both at the same time would attempt to establish an SSL session twice, which is not supported by either PostgreSQL or standard LDAP libraries.

src/utils/pgwire/src/ldap_auth.rs
yezizp2012: Yes, I'm working on it and it was used just for testing purpose. Will update it later after function works as expected.

yezizp2012:
> Can we have a tracking issue to describe the follow-up works? I noticed several TODOs in this PR.

> 1. Enhance the CI testing and ensure compatibility testing is in place.
> 2. TLS support, including connection detection, LDAP authentication, and relevant testing.
> 3. Try to implement all ldap configure options provided by PG as much as possible.

I'm still working on it to get 2, 3 supported in this PR. After that I will open a issue to track the testing stuff.

src/utils/pgwire/src/ldap_auth.rs
yezizp2012: 👍🏻 Used `dn_escape`, `ldap_escape` to avoid injection during binding.

yezizp2012:
> > Can we have a tracking issue to describe the follow-up works? I noticed several TODOs in this PR.
> 
> > 1. Enhance the CI testing and ensure compatibility testing is in place.
> > 2. TLS support, including connection detection, LDAP authentication, and relevant testing.
> > 3. Try to implement all ldap configure options provided by PG as much as possible.
> 

1. 1 and 2 is done in this PR, together with some changes in the ldap-server repo: https://github.com/kwannoel/ldap-server/pull/1.
2. opened an issue for testing: https://github.com/risingwavelabs/risingwave/issues/23594

src/common/src/config/hba.rs
```diff
@@ -0,0 +1,313 @@
...
+}
+
+impl Default for HbaConfig {
+    fn default() -> Self {
+        Self {
+            entries: vec![
+                // Default rule: allow all local connections without authentication
+                HbaEntry {
+                    connection_type: ConnectionType::Local,
+                    databases: vec!["all".to_owned()],
```
shanicky: Can this be changed to use the const keyword?

xiangjinwu:
The dependencies in this PR are outdated. Could you check whether they are upgradable?
* #23662
* #23663
* #23664

yezizp2012:
> The dependencies in this PR are outdated. Could you check whether they are upgradable?
> 
> * [chore(deps): Bump rustls-pemfile from 1.0.4 to 2.2.0 #23662](https://github.com/risingwavelabs/risingwave/pull/23662)
> * [chore(deps): Bump rustls from 0.21.11 to 0.23.5 #23663](https://github.com/risingwavelabs/risingwave/pull/23663)
> * [chore(deps): Bump rustls-native-certs from 0.6.3 to 0.8.1 #23664](https://github.com/risingwavelabs/risingwave/pull/23664)

Tried, but found that `rw-tokio` version limits `ldap3` to only `0.11`, `ldap3@0.11.5` requires `rustls@0.21.12`, which indirectly affects the version of `rustls-pemfile`. Let's hold and ignore the bump this time.
```
error: failed to select a version for `tokio`.
    ... required by package `ldap3 v0.12.1`
    ... which satisfies dependency `ldap3 = "^0.12.1"` of package `pgwire v2.7.0-alpha (~/codes/risingwave/src/utils/pgwire)`
    ... which satisfies path dependency `pgwire` (locked to 2.7.0-alpha) of package `risingwave_frontend v2.7.0-alpha (~/codes/risingwave/src/frontend)`
    ... which satisfies path dependency `risingwave_frontend` (locked to 2.7.0-alpha) of package `risingwave_cmd_all v2.7.0-alpha (~/codes/risingwave/src/cmd_all)`
versions that meet the requirements `^1.47.1` are: 1.48.0, 1.47.2, 1.47.1

all possible versions conflict with previously selected packages.

  previously selected package `tokio v1.44.2`
    ... which satisfies dependency `rw-tokio = "=1.44.2"` of package `risingwave_rt v2.7.0-alpha (~/codes/risingwave/src/utils/runtime)`
    ... which satisfies path dependency `risingwave_rt` (locked to 2.7.0-alpha) of package `risingwave_cmd_all v2.7.0-alpha (~/codes/risingwave/src/cmd_all)`
```

xiangjinwu:
> Tried, but found that `rw-tokio` version limits `ldap3` to only `0.11`, `ldap3@0.11.5` requires `rustls@0.21.12`, which indirectly affects the version of `rustls-pemfile`. Let's hold and ignore the bump this time.

`ldap3@0.11` depends on `ring@0.16` (no updates since 2021-02-01), which leads to https://github.com/risingwavelabs/risingwave/security/dependabot/224

PR #23152, opened 2025-09-09
fix(sink): Forbid non-migrated tables from sink DDL operations

Chiro11:

Part 2 in https://github.com/risingwavelabs/risingwave/pull/23017#issuecomment-3265742446

- Add `FragmentTypeFlag::UpstreamSinkUnion`.
- Forbid creating and dropping sinks into table that has not been migrated.

Test locally: Table created in release-2.6 cannot perform new sink ddl operations. But old sinks are still valid.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/meta/src/controller/utils.rs
+++ b/src/meta/src/controller/utils.rs
@@ -2255,13 +2255,13 @@ where
         .all(txn)
         .await?;
 
-    if mview_fragment.len() != 1 {
-        return Err(anyhow::anyhow!(
+    let mview_fragment_len = mview_fragment.len();
+    if mview_fragment_len != 1 {
+        bail!(
             "expected exactly one mview fragment for table {}, found {}",
             table_id,
-            mview_fragment.len()
-        )
-        .into());
+            mview_fragment_len
+        );
     }
 
     let mview_fragment = mview_fragment.into_iter().next().unwrap();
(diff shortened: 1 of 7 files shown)


src/meta/src/controller/utils.rs
```diff
@@ -2239,6 +2239,38 @@ where
...
+    if mview_fragment.len() != 1 {
+        return Err(anyhow::anyhow!(
+            "expected exactly one mview fragment for table {}, found {}",
+            table_id,
+            mview_fragment.len()
+        )
+        .into());
+    }
+
+    let mview_fragment = mview_fragment.into_iter().next().unwrap();
```
wenym1:
```suggestion
    let mview_fragment_len = mview_fragment.len();
    let Ok([mview_fragment]) = mview_fragment.try_into() else {
        bail!("expected exactly one mview fragment for table {}, found {}",
            table_id,
            mview_fragment_len
        )
    }
```

wenym1 (approved):
LGTM

~~Have you tested locally to create a table in older version and then try to create a sink-into-table in the latest version?~~

See the PR description. Already tested.

src/meta/src/controller/utils.rs
Chiro11: Encountered error `type annotations needed`. Seems that the rust compiler is not that smart. 😂

yezizp2012 (approved):

PR #23154, opened 2025-09-09
feat(vector): support vector search on relation

wenym1:

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/optimizer/rule/top_n_to_vector_search_rule.rs
+++ b/src/frontend/src/optimizer/rule/top_n_to_vector_search_rule.rs
@@ -17,13 +17,11 @@ use std::assert_matches::assert_matches;
 use risingwave_common::types::DataType;
 use risingwave_common::util::sort_util::ColumnOrder;
 use risingwave_pb::common::PbDistanceType;
-
+use risingwave_pb::plan_common::JoinType;
 use crate::expr::{Expr, ExprImpl, ExprRewriter, ExprType, InputRef};
 use crate::optimizer::LogicalPlanRef;
 use crate::optimizer::plan_node::generic::TopNLimit;
-use crate::optimizer::plan_node::{
-    LogicalPlanRef as PlanRef, LogicalProject, LogicalTopN, LogicalVectorSearch, PlanTreeNodeUnary,
-};
+use crate::optimizer::plan_node::{LogicalPlanRef as PlanRef, LogicalProject, LogicalTopN, LogicalVectorSearch, PlanTreeNodeBinary, PlanTreeNodeUnary};
 use crate::optimizer::rule::prelude::*;
 use crate::optimizer::rule::{BoxedRule, ProjectMergeRule, Rule};
 
@@ -151,5 +149,34 @@ impl Rule<Logical> for TopNToVectorSearchRule {
         let top_n = plan.as_logical_top_n()?;
         let (vector_search, project_exprs) = Self::resolve_vector_search(top_n)?;
         Some(LogicalProject::create(vector_search.into(), project_exprs))
+    }
+}
+
+pub struct CorrelatedTopNToVectorSearchRule;
+
+impl CorrelatedTopNToVectorSearchRule {
+    pub fn create() -> BoxedRule<Logical> {
+        Box::new(CorrelatedTopNToVectorSearchRule)
+    }
+}
+
+impl Rule<Logical> for CorrelatedTopNToVectorSearchRule {
+    fn apply(&self, plan: PlanRef) -> Option<PlanRef> {
+        let apply = plan.as_logical_apply()?;
+        if apply.join_type() != JoinType::LeftOuter {
+            return None;
+        }
+        if !apply.max_one_row() {
+            return None;
+        }
+        let correlated_id = apply.correlated_id();
+        let input = apply.left();
+        let right = apply.right();
+        let project = right.as_logical_project()?;
+        let project_input = project.input();
+        let agg = project_input.as_logical_agg()?;
+        let agg_input = agg.input();
+        let (vector_search, project_exprs) = TopNToVectorSearchRule::resolve_vector_search(agg_input.as_logical_top_n()?)?;
+        None
     }
 }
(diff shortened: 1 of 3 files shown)


wenym1:
* **#23288** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23288?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#23154** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23154?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23154?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#23339** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23339?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#23464](https://github.com/risingwavelabs/risingwave/pull/23464) <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23464?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

src/frontend/src/optimizer/rule/top_n_to_vector_search_rule.rs
```diff
@@ -109,18 +115,9 @@ impl TopNToVectorSearchRule {
...
 
-        let vector_search = LogicalVectorSearch::new(
-            limit,
-            distance_type,
-            left.clone(),
-            right.clone(),
-            (0..projection_input.schema().len()).collect(),
-            projection_input.clone(),
-        );
-        let mut i2o_mapping = vector_search.i2o_mapping();
```
wenym1, line 120: Review note: in previous code, the `i2o_mapping` in returned from `LogicalVectorSearch::new` is always identical, so there was no need to do `i2o_mapping.rewrite_expr`, and it's safe to exclude the logic.

src/frontend/src/optimizer/plan_node/logical_vector_search_lookup_join.rs
```diff
@@ -0,0 +1,438 @@
...
+            ctx.trace("LogicalVectorSearchLookupJoin revert to LogicalApply:");
+            ctx.trace(logical_apply.explain_to_string());
+        }
+
+        let unnested_plan = LogicalOptimizer::subquery_unnesting(
+            logical_apply,
+            false,
+            ctx.is_explain_trace(),
+            &ctx,
+        )?;
```
chenzl25: Why do we need to revert in the `to_batch`?

chenzl25:
BTW， the PR description without any index created, why can it does an index lookup join?

silver-ymz:
I find that after applying `CorrelatedTopNToVectorSearchRule`, plan's output schema type may be different. Is this an expected behaviour or bug?

For example:
```sql
create table items (id int primary key, extra string, text string, embedding vector(3)) append only;
create index i on items using flat (embedding) include(text) with (distance_type = 'l2');
explain(trace) with input as (select '[3,2,1]'::vector(3) as embedding) select array(select row(id, text) from items order by input.embedding <-> items.embedding limit 2) as related_info from input;
```

```txt
Previous:
 LogicalProject { exprs: [$expr3] }
 └─LogicalApply { type: LeftOuter, on: true, correlated_id: 1, max_one_row: true }
   ├─LogicalProject { exprs: ['[3,2,1]':Vector(3)] }
   │ └─LogicalValues { rows: [[]], schema: Schema { fields: [] } }
   └─LogicalProject { exprs: [Coalesce(array_agg($expr1 order_by($expr2 ASC)), ARRAY[]:List(Struct(StructType { fields: [("f1", Int32), ("f2", Varchar)] }))) as $expr3] }
     └─LogicalAgg { aggs: [array_agg($expr1 order_by($expr2 ASC))] }
       └─LogicalTopN { order: [$expr2 ASC], limit: 2, offset: 0 }
         └─LogicalProject { exprs: [Row(items.id, items.text) as $expr1, L2Distance(CorrelatedInputRef { index: 0, correlated_id: 1 }, items.embedding) as $expr2] }
           └─LogicalScan { table: items, columns: [id, extra, text, embedding, _rw_timestamp] }

LogicalApply's schema: { fields: ['[3,2,1]':Vector(3):Vector(3), $expr10003:List(Struct(StructType { fields: [("f1", Int32), ("f2", Varchar)] }))] }

After CorrelatedTopNToVectorSearchRule:
 LogicalProject { exprs: [array] }
 └─LogicalVectorSearchLookupJoin { distance_type: L2Sqr, top_n: 2, input_vector: '[3,2,1]':Vector(3):Vector(3), lookup_vector: items.embedding }
   ├─LogicalProject { exprs: ['[3,2,1]':Vector(3)] }
   │ └─LogicalValues { rows: [[]], schema: Schema { fields: [] } }
   └─LogicalScan { table: items, columns: [id, extra, text, embedding, _rw_timestamp] }

LogicalVectorSearchLookupJoin's schema: { fields: ['[3,2,1]':Vector(3):Vector(3), array:List(Struct(StructType { fields: [("items.id", Int32), ("items.text", Varchar)] }))] }


```

wenym1:
> I find that after applying `CorrelatedTopNToVectorSearchRule`, plan's output schema type may be different. Is this an expected behaviour or bug?

Seems that they are only different in the name of inner fields of struct type? From my understanding, field names are more for debugging and displaying in our internal optimizer processing, so it does not cause any internal bug. But indeed when accessing the rows externally, users may not be able to access the expected column. 

@silver-ymz Are you interesting in resolving it? Besides, I think the trivial field name `f1` and `f2` in the original `LogicalApply` is not correct either.

silver-ymz:
The `PartialEq` impl for `DataType` will compare the name of inner fields of struct type, so I encountered some related problems when updating the optimizer code. But I bypassed it.

I'd like to resolve it when I have time. We can open an issue for it temporarily.

wenym1:
> The `PartialEq` impl for `DataType` will compare the name of inner fields of struct type, so I encountered some related problems when updating the optimizer code. But I bypassed it.
> 
> I'd like to resolve it when I have time. We can open an issue for it temporarily.

I encountered it too. If we don't care field name, we can just use `.equals_datatype()`.

PR #23155, opened 2025-09-10
feat(pgwire): implement OAuth audience validation 

### LINKED ISSUE #23126: OAuth audience validation fails when JWT contains "aud" claim

yuhao-su:

This PR implements OAuth audience validation. Fixed #23126 .

**Summary of changes:**
- Added OAuth audience (`aud`) validation functionality

**How does this PR work:**
The changes add validation logic to ensure that OAuth tokens contain the correct audience claim before accepting authentication. 

## Checklist

## Documentation

<details>

https://docs.risingwave.com/sql/commands/sql-create-user

We need to update this page to tell users when use `oauth`, the jwt should contains a `aud` claim with value `urn:risingwave:cluster:your_cluster_id`

</details>

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9474,7 +9474,7 @@ dependencies = [
  "parking_lot 0.12.1",
  "peekable",
  "postgres-types",
- "rand 0.9.0",
+ "rand 0.8.5",
  "reqwest 0.12.15",
  "risingwave_common",
  "risingwave_sqlparser",
--- a/src/utils/pgwire/Cargo.toml
+++ b/src/utils/pgwire/Cargo.toml
@@ -41,7 +41,7 @@ workspace-hack = { path = "../../workspace-hack" }
 [dev-dependencies]
 base64 = "0.22"
 jsonwebtoken = { workspace = true }
-rand = { workspace = true }
+rand = { version = "0.8" }
 rsa = "0.9"
 tempfile = "3"
 tokio-postgres = "0.7"


yezizp2012:
Seems to solve the same issue as this one 🥵 : #23143 .

src/utils/pgwire/src/pg_server.rs
```diff
@@ -199,18 +202,34 @@ async fn validate_jwt(
     jwt: &str,
     jwks_url: &str,
     issuer: &str,
+    cluster_id: &str,
     metadata: &HashMap<String, String>,
 ) -> Result<bool, BoxedError> {
-    let header = decode_header(jwt)?;
+    let _header = decode_header(jwt)?;
```
yezizp2012:
```suggestion
```

src/utils/pgwire/src/pg_server.rs
```diff
@@ -199,18 +202,34 @@ async fn validate_jwt(
...
     let jwks: Jwks = reqwest::get(jwks_url).await?.json().await?;
+    validate_jwt_with_jwks(jwt, &jwks, issuer, cluster_id, metadata)
+}
+
+fn audience_from_cluster_id(cluster_id: &str) -> String {
+    format!("urn:risingwave:cluster:{}", cluster_id)
```
yezizp2012, line 213: Please add a `user-facing` label to this PR, along with any necessary documentation. Especially the audience format, which will be used to generate the token.

With aud becoming a required field after this PR, Cloud must patch and deploy a version that generates tokens with the `audience` in the specified format, prior to any releases. Cc @xuhui-lu . 
And @yuhao-su, we also need a way to expose the internal `cluster_id` to Cloud, perhaps via a SQL function.

yezizp2012: > And @yuhao-su, we also need a way to expose the internal cluster_id to Cloud, perhaps via a SQL function.

Required for cloud deployment. We also need a patch version in cloud side to generate different tokens based according to different kernel versions. Cc @xuhui-lu

yuhao-su: I'll add the cluster_id function in another PR.

yuhao-su: #23260

yezizp2012 (approved):

yuhao-su:
@xuhui-lu should we merge this or you need do the preparation for the cloud first?

xuhui-lu:
> @xuhui-lu should we merge this or you need do the preparation for the cloud first?

I will prepare the patch today, what is the starting version tag for this feature?

yuhao-su:
> > @xuhui-lu should we merge this or you need do the preparation for the cloud first?
> 
> I will prepare the patch today, what is the starting version tag for this feature?

 It could be v2.6 or earlier if needed

yezizp2012:
I think we can merge this PR together with #23260 first. Then it will be included in the next official release. Cloud just needs to update the patch before the next release.

Btw, for non-cloud users, if they use the `oauth` feature, they will also encounter token invalid issues after upgrading. Perhaps we should add a note in the doc to remind users to update their tokens after upgrading.

yezizp2012:
Call for any members from cargo-lock to approve it. Cc @lmatz @MrCroxx @BugenZhao .

xiangjinwu:
Why did 1b48d5342d009b6307bed27d6eb45fe3f2ea5dc7 revert from `rand = { workspace = true }` to `rand = { version = "0.8" }`? #23661

yuhao-su:
> Why did [1b48d53](https://github.com/risingwavelabs/risingwave/commit/1b48d5342d009b6307bed27d6eb45fe3f2ea5dc7) revert from `rand = { workspace = true }` to `rand = { version = "0.8" }`? #23661

The `rsa` 0.9 we are currently using in cargo.lock does not support `rand` 0.9, that's why you can see we still have a lot of `rand 0.85` in use.

And `rsa 0.10` is till on RC https://github.com/RustCrypto/RSA/tags

PR #23165, opened 2025-09-10
refactor: deprecate actor_splits in SouceManager

shanicky:

This PR deprecates the actor_splits field in SourceManager and instead uses the Splits field of Actor in SharedActorInfo. 
## Checklist

### DIFF
--- a/src/meta/src/barrier/context/recovery.rs
+++ b/src/meta/src/barrier/context/recovery.rs
@@ -687,7 +687,12 @@ impl GlobalBarrierWorkerContextImpl {
         })?;
 
         // get split assignments for all actors
-        let source_splits = self.env.shared_actor_infos().list_assignments();
+        let mut source_splits = HashMap::new();
+        for fragment in info.values().flatten() {
+            for (actor_id, info) in &fragment.actors {
+                source_splits.insert(*actor_id, info.splits.clone());
+            }
+        }
 
         let cdc_table_backfill_actors = self
             .metadata_manager


src/meta/src/stream/source_manager/split_assignment.rs
```diff
@@ -227,7 +233,7 @@ impl SourceManager {
             .collect();
         let assignment = align_splits(
             aligned_actors.into_iter(),
```
shanicky, line 196: The comment here is highly misleading. In reality, it treats the original upstream as downstream and the downstream as upstream, then uses the downstream's split to restore the upstream's split after it has been replaced. I can't even imagine how I managed to understand this part of the code.

src/meta/src/stream/source_manager.rs
```diff
@@ -504,15 +445,27 @@ impl SourceManager {
 
     pub async fn list_assignments(&self) -> HashMap<ActorId, Vec<SplitImpl>> {
         let core = self.core.lock().await;
-        core.actor_splits.clone()
+        core.env
```
wenym1: Since now we don't want source manager handle anything related to actors, shall we make `list_assignments` to be a method of `env`? Same for `get_running_info`. In `get_running_info`, we may remove the `actor_splits` from `SourceManagerRunningInfo`, and call `env.list_assignments()` separately in where we originally acccess the `actor_splits`.

After these updates, we may remove the `env` from `SourceManager` to enforce a boundary between `SourceManager` and other components.

src/meta/src/controller/fragment.rs
```diff
@@ -1120,6 +1124,7 @@ impl CatalogController {
             .column(actor::Column::ActorId)
             .column(actor::Column::WorkerId)
             .column(actor::Column::VnodeBitmap)
+            .column(actor::Column::Splits)
```
wenym1, line 1121: Is this a temporary workaround, or we will also deprecate the `Splits` in the `actor` table in the future as our offline discussion?

src/meta/src/controller/fragment.rs
```diff
@@ -629,6 +630,7 @@ impl CatalogController {
             .remove(&job_id);
 
         Self::compose_table_fragments(
```
wenym1: Shall we by the way remove the `actor_splits` from `TableFragments` to simplify the struct? I just inspected the usage on the field, and it seems doable.

src/meta/src/stream/source_manager/split_assignment.rs
```diff
@@ -38,11 +38,17 @@ impl SourceManager {
     ) -> MetaResult<HashMap<ActorId, Vec<SplitImpl>>> {
         let core = self.core.lock().await;
 
+        let guard = core.env.shared_actor_infos().read_guard();
```
wenym1: As mentioned in the other comment, we may remove `env` from `core`, and change the parameter `self` of this method to `&env`.

src/meta/src/stream/stream_manager.rs
```diff
@@ -494,6 +497,10 @@ impl GlobalStreamManager {
                 .await?,
         );
 
+        stream_job_fragments
```
wenym1: Seems that `init_split_assignment` is already carried in `CreateStreamingJobCommandInfo`. As mentioned in the other comment, we may remove the `actor_splits` from `StreamJobFragments`.

src/meta/src/stream/source_manager.rs
shanicky: These two functions can be moved into the env, but the source manager still needs to keep an env for now, because the shared source requires the historical distribution of actor splits to locate the upstream actor splits for no-shuffle.

src/meta/src/controller/fragment.rs
shanicky: it's a temporary workaround

src/meta/src/controller/fragment.rs
shanicky: it's removed in #23212

src/meta/src/stream/stream_manager.rs
shanicky: removed in https://github.com/risingwavelabs/risingwave/pull/23212

src/meta/src/stream/source_manager.rs
shanicky: According to your design, should these be placed in MetaSrvEnv or in SharedActorInfo?  They're all using the SharedActorInfo.

src/tests/simulation/tests/integration_tests/recovery/pause_on_bootstrap.rs
```diff
@@ -106,7 +106,7 @@ async fn test_impl(resume_by: ResumeBy) -> Result<()> {
 
     // Resume the cluster.
     resume_by.resume(&mut cluster).await?;
-    sleep(Duration::from_secs(10)).await;
+    sleep(Duration::from_secs(30)).await;
```
shanicky, line 109: At startup we perform an offline scale, but in this version of the code the offline-scale logic cannot get actor splits from the env's shared actor info, so we have to wait a bit longer for the first split tick. This won't be the case in later PRs because the offline-scaling approach was completely reworked.

src/meta/src/barrier/context/recovery.rs
```diff
@@ -520,7 +520,19 @@ impl GlobalBarrierWorkerContextImpl {
                         .await?;
 
                     // get split assignments for all actors
-                    let source_splits = self.source_manager.list_assignments().await;
+
+                    // QQ: why empty here?
+                    // let source_splits = self.env.shared_actor_infos().list_assignments();
```
yezizp2012: IIUC they have been cleaned during recovery.
```suggestion
```

yezizp2012 (approved):
Generally LGTM

src/meta/src/barrier/context/recovery.rs
```diff
@@ -520,7 +520,15 @@ impl GlobalBarrierWorkerContextImpl {
                         .await?;
 
                     // get split assignments for all actors
-                    let source_splits = self.source_manager.list_assignments().await;
+                    let mut source_splits = HashMap::new();
```
wenym1, line 523: Should we also call `self.env.shared_actor_infos().list_assignments();` like in line 690?

src/meta/src/stream/source_manager/split_assignment.rs
```diff
@@ -630,7 +616,11 @@ impl<T: SplitMetaData + Clone> PartialOrd<Self> for ActorSplitsAssignment<T> {
 impl<T: SplitMetaData + Clone> Ord for ActorSplitsAssignment<T> {
     fn cmp(&self, other: &Self) -> Ordering {
         // Note: this is reversed order, to make BinaryHeap a min heap.
-        other.splits.len().cmp(&self.splits.len())
+        other
```
wenym1, line 619: See `reassign_splits`. It seems that it's by-design to do `cmp` only on `splits`.

PR #18205, opened 2024-08-22
feat(agg): support `aggregate:` prefixed scalar function in streaming agg

stdrc:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://gist.github.com/TennyZhuang/f00be7f16996ea48effb049aa7be4d66#file-rw_cla).

## What's changed and what's your intention?

After several previous PRs:

- https://github.com/risingwavelabs/risingwave/pull/18177
- https://github.com/risingwavelabs/risingwave/pull/18203

we can now easily support using `aggregate:` prefixed scalar function (including UDF) as streaming aggregate function. This resolves #16767.

## Checklist

- [ ] I have written necessary rustdoc comments
- [ ] I have added necessary unit tests and integration tests
- [ ] I have added test labels as necessary. See [details](https://github.com/risingwavelabs/risingwave/blob/main/docs/developer-guide.md#ci-labels-guide).
- [ ] I have added fuzzing tests or opened an issue to track them. (Optional, recommended for new SQL features #7934).
- [ ] My PR contains breaking changes. (If it deprecates some features, please create a tracking issue to remove them in the future).
- [x] All checks passed in `./risedev check` (or alias, `./risedev c`)
- [ ] My PR changes performance-critical code. (Please run macro/micro-benchmarks and show the results.)

- [ ] My PR contains critical fixes that are necessary to be merged into the latest release. (Please check out the [details](https://github.com/risingwavelabs/risingwave/blob/main/CONTRIBUTING.md))

## Documentation

- [x] My PR needs documentation updates. (Please use the **Release note** section below to summarize the impact on users)

## Release note

- Support `AGGREGATE:` prefixed scalar function in streaming aggregation

Issue #12857, opened 2023-10-16
feat: ordered-set aggregate in streaming queries


stdrc:

kwannoel:
Depends on https://github.com/risingwavelabs/risingwave/pull/12570

stdrc:
> Depends on #12570

This is already merged. Are we currently doing Streaming Percentile or Streaming Approx  Percentile? @kwannoel

kwannoel:
> > Depends on #12570
> 
> This is already merged. Are we currently doing Streaming Percentile or Streaming Approx  Percentile? @kwannoel 

I'm working on Streaming Approx Percentile. Streaming Percentile is not worked on yet. You may take it up if interested.

stdrc:
Since we already support `approx_percentile`, I guess it's not an urgent to support accurate `percentile_cont/cont`? We can remove the milestone for now.

BugenZhao:
I believe a non-incremental algorithm can be easily supported, since we've already supported full aggregation for wrapped-scalar agg calls in https://github.com/risingwavelabs/risingwave/pull/18205. However, the complexity will be the same as `array_agg` then looking up the percentile position. This is more like a syntax sugar.

The next step is to implement an incremental algorithm, which may require some efforts.

PR #23172, opened 2025-09-11
feat(streaming): support ordered-set aggregations (`PERCENTILE_CONT`, `PERCENTILE_DISC`, `MODE`)

BugenZhao:

Background: #12857

Since we've already supported full aggregation for wrapped-scalar agg calls in https://github.com/risingwavelabs/risingwave/pull/18205, a non-incremental algorithm can be easily supported for ordered-set aggregations. Basically, when `get_output`, we create a fresh agg state and apply all data in the group/partition with the `#[aggregate]` function body (which does not support retraction). Though not efficient, this is very similar to how `array_agg` is achieved.

The code changes are very simple, just to move these agg variants out of `unimplemented_in_stream` into `materialized_input`.

The next step is to implement an incremental algorithm, which may require some efforts.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/e2e_test/streaming/aggregate/ordered_set_agg.slt
+++ b/e2e_test/streaming/aggregate/ordered_set_agg.slt
@@ -123,7 +123,7 @@ from t_grouped
 group by group_id;
 
 statement ok
-insert into t_grouped values 
+insert into t_grouped values
 (1, 1), (1, 3), (1, 5), (1, 7),
 (2, 10), (2, 20), (2, 30);
 


BugenZhao:
> But once we implement it in materializing all input, it would cause a backward compatibility issue if we want to change it back to an incremental algorithm.

While the time complexity could be improved, I think in any ways we need to maintain all historical data for retractable stream, so materializing all data at this stage may not lead to a problem in the future.

Issue #22653, opened 2025-07-17
Failing to create a materialized view with EMIT ON WINDOW CLOSE when it has a temporal filter


Ruchirasd:
### Describe the bug

I'm trying to create a materialized view using a HOP window on a Kafka source which has a watermark defined on `timestamp` column. 
When the mv has `EMIT ON WINDOW CLOSE` and a temporal filter (`timestamp > NOW() - INTERVAL '7 hours'`), mv fails to create with below error. But it works when filter has only constants (`timestamp > '2025-07-17 06:00:00'`).

### Error message/log

```text
ERROR: Failed to execute the statement Caused by: Not supported: Emit-On-Window-Close mode requires a watermark column in GROUP BY. HINT: Please try to GROUP BY a watermark column (SQLSTATE XX000): query failed
```

### To Reproduce

1. Create a Kafka source with a watermark on RW.
    ```sql
    CREATE SOURCE s1 (
      id VARCHAR,
      timestamp TIMESTAMP,
      WATERMARK FOR timestamp AS timestamp - INTERVAL '5' SECOND
    ) WITH (
      connector = 'kafka', 
      -- add required parameters here --
      scan.startup.mode = 'earliest'
    ) 
    FORMAT PLAIN ENCODE JSON;
    ```
1. Create a materialized view. **This query will fail.**
    ```sql
    CREATE MATERIALIZED VIEW mv1 AS
    SELECT 
        id,
        window_start,
        window_end,
        MAX(timestamp) as ts
    FROM HOP(s1, timestamp, INTERVAL '30 SECONDS', INTERVAL '2 MINUTES')
    WHERE timestamp > NOW() - INTERVAL '7 hours'
    GROUP BY id, window_start, window_end
    EMIT ON WINDOW CLOSE;
    ```
1. Create a materialized view. **This query will succeed.**
    ```sql
    CREATE MATERIALIZED VIEW mv1 AS
    SELECT 
        id,
        window_start,
        window_end,
        MAX(timestamp) as ts
    FROM HOP(s1, timestamp, INTERVAL '30 SECONDS', INTERVAL '2 MINUTES')
    WHERE timestamp > '2025-07-17 06:00:00'
    GROUP BY id, window_start, window_end
    EMIT ON WINDOW CLOSE;
    ```

### Expected behavior

A materialized view with `HOP` window and `EMIT ON WINDOW CLOSE` should be able to create with a temporal filter. The purpose is to limit the data retention inside the materialized view.

[...]

PR #23180, opened 2025-09-12
fix(optimizer): fix watermark columns derivation of dynamic filter

chenzl25:

- Resolve https://github.com/risingwavelabs/risingwave/issues/22653
- We can merge the watermark columns from the left side when we can derive watermark columns from the right side in dynamic filter.

**Watermark propagation improvements:**

* Refactored the logic in `StreamDynamicFilter` to correctly merge watermark columns from the left input and derive new watermark columns based on the comparator type, ensuring more accurate propagation of watermarks through dynamic filters. (`src/frontend/src/optimizer/plan_node/stream_dynamic_filter.rs`)
* Updated the output plan for expressions involving dynamic filters to include both `v1` and `v2` as watermark columns, and adjusted the output watermarks to reflect propagation from both columns. (`src/frontend/planner_test/tests/testdata/output/expr.yaml`)

**Test coverage and correctness:**

* Added a new SQL query and expected outputs to test windowed aggregation with watermark propagation, including both standard and emit-on-window-close plans. (`src/frontend/planner_test/tests/testdata/input/emit_on_window_close.yaml`)
* Added corresponding expected output plans for the new windowed query, detailing how watermark columns are handled in both the regular and emit-on-window-close scenarios. (`src/frontend/planner_test/tests/testdata/output/emit_on_window_close.yaml`)

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/planner_test/tests/testdata/input/emit_on_window_close.yaml
+++ b/src/frontend/planner_test/tests/testdata/input/emit_on_window_close.yaml
@@ -56,3 +56,20 @@
   expected_outputs:
     - eowc_stream_plan
     - stream_plan
+- sql: |
+    CREATE table s1 (
+    id VARCHAR,
+    timestamp TIMESTAMP,
+    WATERMARK FOR timestamp AS timestamp - INTERVAL '5' SECOND
+    ) append only;
+    SELECT
+        id,
+        window_start,
+        window_end,
+        MAX(timestamp) as ts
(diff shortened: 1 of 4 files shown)


yuhao-su:
> We can merge the watermark columns from the left side when we can derive watermark columns from the right side in dynamic filter.

Dynamic filter only yield the watermark from the rhs. Also considering the increase in rhs can emit delete of value from the lhs, so using non inequality compare column doesn't look right.

PR #23194, opened 2025-09-13
feat: Add source_splits model, persist discovered source splits

shanicky:

This PR adds the `source_splits` table, which stores the source splits discovered each time. It is written to when a source is created, updated during the `post_collect` phase of split assignment, and deleted when a source is dropped. 

This PR **only** adds the **write** interface, which will be used in a future actorless PR.

## Checklist

### DIFF
--- a/src/meta/src/controller/fragment.rs
+++ b/src/meta/src/controller/fragment.rs
@@ -1870,7 +1870,6 @@ impl CatalogController {
                     .update_column(source_splits::Column::Splits)
                     .to_owned(),
             )
-            .do_nothing()
             .exec(&txn)
             .await?;
 


src/meta/src/controller/fragment.rs
```diff
@@ -1846,6 +1846,48 @@ impl CatalogController {
...
+            .map(|(source_id, splits)| source_splits::ActiveModel {
+                source_id: Set(*source_id as _),
+                splits: Set(Some(ConnectorSplits::from(&PbConnectorSplits {
+                    splits: splits.iter().map(Into::into).collect_vec(),
+                }))),
+            })
+            .collect();
+
+        SourceSplits::insert_many(models)
+            .on_conflict(
```
yezizp2012, line 1868: I think I remember that `on_conflict` needs special handling when using `mysql` metastore.

src/meta/src/controller/fragment.rs
```diff
@@ -1846,6 +1846,48 @@ impl CatalogController {
...
+            )
+            .exec(&txn)
+            .await?;
+
+        txn.commit().await?;
+
+        Ok(())
+    }
+
+    pub async fn drop_source_splits(&self, source_ids: &[SourceId]) -> MetaResult<()> {
```
yezizp2012: We can rely on FK cascade delete for it.

src/meta/src/stream/source_manager.rs
```diff
@@ -307,6 +310,21 @@ impl SourceManagerCore {
...
+
+    async fn update_source_splits(&self, source_id: SourceId) -> MetaResult<()> {
+        let handle_ref = self.managed_sources.get(&source_id).unwrap();
+
+        let discovered_splits = handle_ref.splits.lock().await.splits.clone();
+
+        if let Some(splits) = discovered_splits {
+            let source_splits =
+                HashMap::from([(source_id as _, splits.into_values().collect_vec())]);
+            self.metadata_manager
```
yezizp2012, line 322: How about unify the updates of source splits in post collection of the barrier?

yezizp2012 (commented):
Rest LGTM

yezizp2012 (approved):
lgtm

src/meta/src/barrier/context/context_impl.rs
```diff
@@ -192,14 +192,23 @@ impl CommandContext {
 
             Command::Resume => {}
 
-            Command::SourceChangeSplit(split_assignment) => {
+            Command::SourceChangeSplit(SplitState {
+                split_assignment: assignment,
+                discovered_source_splits: source_splits,
```
tabVersion, line 197: `discovered_source_splits: source_splits` does the source_splits here contain full state for the source?

src/meta/src/stream/source_manager/split_assignment.rs
```diff
@@ -412,9 +424,34 @@ impl SourceManagerCore {
...
+            let mut source_splits = HashMap::new();
+            for (source_id, fragment_ids) in &self.source_fragments {
+                if fragment_ids
+                    .iter()
+                    .any(|fragment_id| assignment.contains_key(fragment_id))
+                    && let Some(splits) = source_splits_discovered.get(source_id)
+                {
+                    source_splits.insert(*source_id, splits.clone());
+                }
+            }
```
tabVersion, line 443: If the barrier fails, the cluster will recover with new split assignments. 
Then the barrier is for applying the running actors, why should barrier carries `discovered_source_splits`

src/meta/src/barrier/context/context_impl.rs
```diff
@@ -192,14 +192,23 @@ impl CommandContext {
...
                 barrier_manager_context
                     .metadata_manager
-                    .update_actor_splits_by_split_assignment(split_assignment)
+                    .update_actor_splits_by_split_assignment(assignment)
                     .await?;
+
+                barrier_manager_context
+                    .metadata_manager
+                    .update_source_splits(source_splits)
+                    .await?;
```
tabVersion, line 207: I think this part should be moved before yielding the split change mutation. 

We have updated split_assignment in each related fragment in the re-assign process, which means the mutation will be applied no matter the barrier succeeds or not. 
But here we update the state table only when the barrier succeeds. It might cause inconsistency in barrier failure scenes.

tabVersion (commented):
can you share more about the new metadata update process?

Issue #7983, opened 2023-02-16
Support Pgwire Copy Protocol


neverchanje:
### Is your feature request related to a problem? Please describe.

https://www.postgresql.org/docs/current/protocol-flow.html#PROTOCOL-COPY

Postgres provides a mechanism that enables retrieving large bulks of data efficiently, which is called 'COPY TO STDOUT'. It's required by connectorx https://github.com/risingwavelabs/risingwave/issues/7793, a python extract&load tool (written in Rust internally). As a result, supporting the command is not enough. We'll also have to support the pgwire protocol.

PR #23198, opened 2025-09-15
feat(pgwire): support copy query to stdout

wenym1:

Clickhouse Postgres table engine use the `copy (<query>) to stdout` syntax to load data from postgres table. See [here](https://clickhouse.com/docs/engines/table-engines/integrations/postgresql#implementation-details) for more details. 

In this PR we will support this feature, and clichhouse user is table to access RisingWave as if accessing Postgres to query table or mv. By default, the Postgres will output data as TSV format.

resolve https://github.com/risingwavelabs/risingwave/issues/7983

## Checklist

## Documentation

<details>

</details>


wenym1:
* **#23198** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23198?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23198?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/sqlparser/src/ast/mod.rs
```diff
@@ -1867,30 +1885,45 @@ impl Statement {
...
+                                    write!(f, "{}", v)?;
+                                } else {
+                                    write!(f, "\\N")?;
+                                }
+                            }
                         }
+                        write!(f, "\n\\.")
+                    }
+                    CopyTarget::Stdout => {
+                        write!(f, "TO stdout;")
```
BugenZhao:
```suggestion
                        write!(f, " TO stdout")
```

The separating space seems to be missing, and we don't include a `;` in unparse result.

Can you add a test case under `src/sqlparser/tests`?

src/sqlparser/src/parser.rs
```diff
@@ -3846,16 +3846,33 @@ impl Parser<'_> {
...
+        };
+
+        let target = if self.parse_keywords(&[Keyword::FROM, Keyword::STDIN]) {
+            self.expect_token(&Token::SemiColon)?;
+            let values = self.parse_tsv();
+            CopyTarget::Stdin { values }
+        } else if self.parse_keywords(&[Keyword::TO, Keyword::STDOUT]) {
+            if !self.consume_token(&Token::EOF) {
+                self.expect_token(&Token::SemiColon)?;
+            }
```
BugenZhao: I guess the semicolon for `FROM STDIN` is a special separator between the statement and input, while it's not applicable to `TO STDOUT`. If you mean the one as a separator between statements here, then it's already handled in `parse_statements`:

https://github.com/risingwavelabs/risingwave/blob/49bc4428689e828e43e42090ea915cbe7251df29/src/sqlparser/src/parser.rs#L278-L286

src/utils/pgwire/src/pg_message.rs
```diff
@@ -710,11 +714,69 @@ impl BeMessage<'_> {
...
+            BeMessage::CopyData(row) => {
+                buf.put_u8(b'd');
+                // by default, write as TSV format
+                write_body(buf, |buf| {
+                    match row.values() {
+                        [] => {}
+                        [first, rest @ ..] => {
+                            if let Some(first) = first {
+                                let s = String::from_utf8_lossy(first);
+                                for c in s.as_str().chars() {
```
BugenZhao: Does methods like https://doc.rust-lang.org/stable/std/primitive.slice.html#method.escape_ascii work here?

src/utils/pgwire/src/pg_message.rs
```diff
@@ -710,11 +714,69 @@ impl BeMessage<'_> {
...
+                            }
+                        }
+                    }
+                    buf.put_u8(b'\n');
+                    Ok(())
+                })?;
+            }
+            BeMessage::CopyDone => {
+                buf.put_u8(b'c');
+                write_body(buf, |_| Ok(()))?;
```
BugenZhao, line 787: Can you show the reference of this?

src/utils/pgwire/src/pg_message.rs
wenym1: `escape_ascii` escapes more characters than needed. As suggested in [wikipedia](https://en.wikipedia.org/wiki/Tab-separated_values), we should only escape `\t` `\n` `\r` `\`.

src/utils/pgwire/src/pg_message.rs
wenym1: See https://www.postgresql.org/docs/current/protocol-message-formats.html.

Will be included in the code comment.

src/sqlparser/src/parser.rs
BugenZhao: Seems still not addressed yet?

wenym1: Addressed now.

src/sqlparser/src/ast/mod.rs
```diff
@@ -1867,30 +1885,45 @@ impl Statement {
...
+                                    write!(f, "{}", v)?;
+                                } else {
+                                    write!(f, "\\N")?;
+                                }
+                            }
                         }
+                        write!(f, "\n\\.")
+                    }
+                    CopyTarget::Stdout => {
+                        write!(f, " TO STDOUT;")
```
BugenZhao:
```suggestion
                        write!(f, " TO STDOUT")
```

wenym1: Fixed. Sorry for having the same error back-and-forth🥺

BugenZhao (approved):
LGTM

PR #23199, opened 2025-09-15
feat(optimizer): recommend index for backfill to users

chenzl25:

- Introduce an new type of explain `explain (advisor) ` + sql.
- Recommend index for backfill to users by notice message.

## Example
```sql
dev=>  create table t (a int, b int, c int);

dev=> explain (advisor) create materialized view v as select count(*) from t where c > 1 group by a;
NOTICE:  To speed up the backfilling, consider creating an index: CREATE INDEX "__recommended_idx_of_t_a" ON "t" ("a")
                                                   QUERY PLAN
-----------------------------------------------------------------------------------------------------------------
 StreamMaterialize { columns: [count, t.a(hidden)], stream_key: [t.a], pk_columns: [t.a], pk_conflict: NoCheck }
 └─StreamProject { exprs: [count, t.a] }
   └─StreamHashAgg { group_key: [t.a], aggs: [count] }
     └─StreamExchange { dist: HashShard(t.a) }
       └─StreamProject { exprs: [t.a, t._row_id] }
         └─StreamFilter { predicate: (t.c > 1:Int32) }
           └─StreamTableScan { table: t, columns: [a, _row_id, c] }
(7 rows)
```

```sql
dev=> CREATE INDEX "__recommended_idx_of_t_a" ON "t" ("a");
CREATE_INDEX
dev=> explain create materialized view v as select count(*) from t where c > 1 group by a;;
                                                                                      QUERY PLAN
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 StreamMaterialize { columns: [count, __recommended_idx_of_t_a.a(hidden)], stream_key: [__recommended_idx_of_t_a.a], pk_columns: [__recommended_idx_of_t_a.a], pk_conflict: NoCheck }
 └─StreamProject { exprs: [count, __recommended_idx_of_t_a.a] }
   └─StreamHashAgg { group_key: [__recommended_idx_of_t_a.a], aggs: [count] }
     └─StreamProject { exprs: [__recommended_idx_of_t_a.a, __recommended_idx_of_t_a.t._row_id] }
       └─StreamFilter { predicate: (__recommended_idx_of_t_a.c > 1:Int32) }
         └─StreamTableScan { table: __recommended_idx_of_t_a, columns: [a, t._row_id, c] }
(6 rows)
```

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/optimizer/plan_node/logical_scan.rs
+++ b/src/frontend/src/optimizer/plan_node/logical_scan.rs
@@ -753,7 +753,9 @@ impl ToStream for LogicalScan {
         }
 
         if self.table_indexes().is_empty() {
-            self.notice_recommended_index(columns);
+            if self.ctx().is_explain_advisor() {
+                self.notice_recommended_index(columns);
+            }
             return None;
         }
 
@@ -785,7 +787,9 @@ impl ToStream for LogicalScan {
             }
         }
 
-        self.notice_recommended_index(columns);
+        if self.ctx().is_explain_advisor() {
+            self.notice_recommended_index(columns);
+        }
         None
     }
 }
--- a/src/sqlparser/src/keywords.rs
+++ b/src/sqlparser/src/keywords.rs
@@ -73,6 +73,7 @@ define_keywords!(
     ACTION,
     ADAPTIVE,
     ADD,
+    ADVISOR,
     AGGREGATE,
     ALL,
     ALLOCATE,
--- a/src/frontend/src/optimizer/optimizer_context.rs
+++ b/src/frontend/src/optimizer/optimizer_context.rs
@@ -206,6 +206,10 @@ impl OptimizerContext {
         self.explain_options.backfill
     }
 
+    pub fn is_explain_advisor(&self) -> bool {
+        self.explain_options.advisor
+    }
+
     pub fn explain_type(&self) -> ExplainType {
         self.explain_options.explain_type.clone()
     }
--- a/src/sqlparser/src/ast/mod.rs
+++ b/src/sqlparser/src/ast/mod.rs
@@ -1204,6 +1204,8 @@ pub struct ExplainOptions {
     pub trace: bool,
     // Display backfill order
     pub backfill: bool,
+    // Show advisor recommendations
+    pub advisor: bool,
     // explain's plan type
     pub explain_type: ExplainType,
     // explain's plan format
@@ -1216,6 +1218,7 @@ impl Default for ExplainOptions {
             verbose: false,
             trace: false,
             backfill: false,
+            advisor: false,
             explain_type: ExplainType::Physical,
             explain_format: ExplainFormat::Text,
         }
--- a/src/sqlparser/src/parser.rs
+++ b/src/sqlparser/src/parser.rs
@@ -4364,6 +4364,7 @@ impl Parser<'_> {
         let mut analyze_duration = None;
 
         let explain_key_words = [
+            Keyword::ADVISOR,
             Keyword::BACKFILL,
             Keyword::VERBOSE,
             Keyword::TRACE,
@@ -4378,6 +4379,7 @@ impl Parser<'_> {
         let parse_explain_option = |parser: &mut Parser<'_>| -> ModalResult<()> {
             let keyword = parser.expect_one_of_keywords(&explain_key_words)?;
             match keyword {
+                Keyword::ADVISOR => options.advisor = parser.parse_optional_boolean(true),
                 Keyword::VERBOSE => options.verbose = parser.parse_optional_boolean(true),
                 Keyword::TRACE => options.trace = parser.parse_optional_boolean(true),
                 Keyword::BACKFILL => options.backfill = parser.parse_optional_boolean(true),


src/frontend/src/optimizer/plan_node/logical_scan.rs
```diff
@@ -692,9 +727,12 @@ impl ToStream for LogicalScan {
         if columns.is_empty() {
             return None;
         }
+
         if self.table_indexes().is_empty() {
+            self.notice_recommended_index(columns);
```
BugenZhao: What if the primary key of this table scan already has the best locality? Do we still recommend index here?

src/frontend/src/optimizer/plan_node/logical_scan.rs
```diff
@@ -565,6 +565,41 @@ impl LogicalScan {
...
+        }
+
+        let column_names = table_column_indices
+            .iter()
+            .map(|&col| self.table().columns[col].name.clone())
+            .collect_vec();
+        let index_columns = column_names
+            .iter()
+            .map(|col| "\"".to_owned() + col + "\"")
+            .join(", ");
```
BugenZhao: Shall we construct a `ast::Statement` here, so that identifier quotation can be correctly handled?

src/frontend/src/optimizer/plan_node/logical_scan.rs
BugenZhao: Also, do we still send a notice when user is actually going to create an index?

BugenZhao (commented):
Personally I'm not sure if it's a good idea to provide such recommendation in SQL interface. Using index to accelerate queries (no matter batch or streaming) seems to be an advanced practice: typically users need to understand the locality of their query and  learn to inspect & comprehend query plan with `EXPLAIN` commands. Given that for batch queries, such topic is only covered in documentation, maybe we should be aligned with that for streaming queries as well. 🤔 cc @hzxa21 for discussion

src/frontend/src/optimizer/plan_node/logical_scan.rs
chenzl25: > What if the primary key of this table scan already has the best locality? Do we still recommend index here?

Yes. We check it inside `notice_recommended_index`.

chenzl25: > Also, do we still send a notice when user is actually going to create an index

What do you mean when user is actually going to create an index? If users have created the related indexes, it means we will select those indexes, so we won't recommend it. This PR only recommend indexes when there are no indexes could be used.

src/frontend/src/optimizer/plan_node/logical_scan.rs
chenzl25: Done

chenzl25:
> Personally I'm not sure if it's a good idea to provide such recommendation in SQL interface. Using index to accelerate queries (no matter batch or streaming) seems to be an advanced practice: typically users need to understand the locality of their query and learn to inspect & comprehend query plan with EXPLAIN commands. Given that for batch queries, such topic is only covered in documentation, maybe we should be aligned with that for streaming queries as well. 🤔 cc @hzxa21 for discussion

Considering we are supporting index selection for backfilling, we should update the doc of index as well to tell users how to create an index to give the streaming queries a better locality .

src/frontend/src/optimizer/plan_node/logical_scan.rs
BugenZhao: I mean, when user is going to follow the instruction to issue a `CREATE INDEX` command, will they still receive a notice for the TableScan of the index job? 😆

chenzl25: No. Because we only recommend an index when there is no indexes could be used, but once users create the index we recommend, the index must be able to be used.

hzxa21:
> Personally I'm not sure if it's a good idea to provide such recommendation in SQL interface. Using index to accelerate queries (no matter batch or streaming) seems to be an advanced practice: typically users need to understand the locality of their query and learn to inspect & comprehend query plan with `EXPLAIN` commands. Given that for batch queries, such topic is only covered in documentation, maybe we should be aligned with that for streaming queries as well. 🤔 cc @hzxa21 for discussion

I think regardless of whether we give index recommendation via SQL notice. We need to add documentation in the doc site to guide user on how to create index to accelerate backfill and why. I have created an issue a couple weeks agao: https://github.com/risingwavelabs/risingwave-docs/issues/617. Please provide more contexts if needed to help doc team draft the doc.


Regarding to index recommendation via SQL, I think showing the notice only in `EXPLAIN`, but not in DDL is better because if user is interested in running `EXPLAIN`, he/she probably is looking for optimizing the query anyway. It also depends on how effective the recommendation is. As discussed in the meeting, it may only help for large table.

chenzl25:
> > Personally I'm not sure if it's a good idea to provide such recommendation in SQL interface. Using index to accelerate queries (no matter batch or streaming) seems to be an advanced practice: typically users need to understand the locality of their query and learn to inspect & comprehend query plan with `EXPLAIN` commands. Given that for batch queries, such topic is only covered in documentation, maybe we should be aligned with that for streaming queries as well. 🤔 cc @hzxa21 for discussion
> 
> I think regardless of whether we give index recommendation via SQL notice. We need to add documentation in the doc site to guide user on how to create index to accelerate backfill and why. I have created an issue a couple weeks agao: [risingwavelabs/risingwave-docs#617](https://github.com/risingwavelabs/risingwave-docs/issues/617). Please provide more contexts if needed to help doc team draft the doc.
> 
> Regarding to index recommendation via SQL, I think showing the notice only in `EXPLAIN`, but not in DDL is better because if user is interested in running `EXPLAIN`, he/she probably is looking for optimizing the query anyway. It also depends on how effective the recommendation is. As discussed in the meeting, it may only help for large table.

Adding an `explain (advisor) ` + sql to recommend indexes.

src/sqlparser/src/keywords.rs
```diff
@@ -73,6 +73,7 @@ define_keywords!(
     ACTION,
     ADAPTIVE,
     ADD,
+    ADVISOR,
```
xiangjinwu, line 76: FYI #23208

chenzl25:
Close this PR, since we have locality enforcement right now. https://github.com/risingwavelabs/risingwave/pull/23275

PR #23201, opened 2025-09-15
refactor(sink): Make sink-into-table refactor backwards compatible

Chiro11:

TLDR: When meta starts, we will spawn a task that continuously execute `replace-job` for every unmigrated table.
For details, see https://github.com/risingwavelabs/risingwave/pull/23017#issuecomment-3265742446

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/meta/service/src/ddl_service.rs
+++ b/src/meta/service/src/ddl_service.rs
@@ -13,9 +13,11 @@
 // limitations under the License.
 
 use std::collections::{HashMap, HashSet};
+use std::pin::pin;
 use std::sync::Arc;
 
 use anyhow::{Context, anyhow};
+use futures::future::select;
 use rand::rng as thread_rng;
 use rand::seq::IndexedRandom;
 use replace_job_plan::{ReplaceSource, ReplaceTable};
@@ -129,7 +131,7 @@ impl DdlServiceImpl {
         let metadata_manager = self.metadata_manager.clone();
         let ddl_controller = self.ddl_controller.clone();
 
-        let (shutdown_tx, mut shutdown_rx) = tokio::sync::oneshot::channel();
+        let (shutdown_tx, shutdown_rx) = tokio::sync::oneshot::channel();
         let join_handle = tokio::spawn(async move {
             async fn migrate_inner(
                 env: &MetaSrvEnv,
@@ -153,8 +155,8 @@ impl DdlServiceImpl {
                     if workers.is_empty() {
                         return Err(anyhow::anyhow!("no active frontend nodes found").into());
                     }
-                    let worker = workers.choose(&mut thread_rng()).unwrap().clone();
-                    env.frontend_client_pool().get(&worker).await?
+                    let worker = workers.choose(&mut thread_rng()).unwrap();
+                    env.frontend_client_pool().get(worker).await?
                 };
 
                 for table in tables {
@@ -182,37 +184,25 @@ impl DdlServiceImpl {
                 Ok(())
             }
 
-            let mut attempt = 0;
-            loop {
-                tokio::select! {
-                    biased;
-                    _ = &mut shutdown_rx => {
-                        tracing::info!("shutting down migrate table fragments task");
-                        break;
-                    },
-                    res = migrate_inner(&env, &metadata_manager, &ddl_controller) => {
-                        match res {
-                            Ok(_) => break,
-                            Err(e) => {
-                                attempt += 1;
-                                tracing::error!(
-                                    "failed to migrate legacy table fragments: {}, attempt {}, retrying in 5 secs",
-                                    e.as_report(),
-                                    attempt
-                                );
-                                tokio::select! {
-                                    biased;
-                                    _ = &mut shutdown_rx => {
-                                        tracing::info!("shutting down migrate table fragments task");
-                                        break;
-                                    },
-                                    _ = tokio::time::sleep(std::time::Duration::from_secs(5)) => {}
-                                }
-                            }
+            let migrate_future = async move {
+                let mut attempt = 0;
+                loop {
+                    match migrate_inner(&env, &metadata_manager, &ddl_controller).await {
+                        Ok(_) => break,
+                        Err(e) => {
+                            attempt += 1;
+                            tracing::error!(
+                                "failed to migrate legacy table fragments: {}, attempt {}, retrying in 5 secs",
+                                e.as_report(),
+                                attempt
+                            );
+                            tokio::time::sleep(std::time::Duration::from_secs(5)).await;
                         }
                     }
                 }
-            }
+            };
+
+            select(pin!(migrate_future), shutdown_rx).await;
         });
 
         (join_handle, shutdown_tx)


src/meta/node/src/server.rs
```diff
@@ -758,6 +760,10 @@ pub async fn start_service_as_election_leader(
     started::set();
     let _server_handle = tokio::spawn(server);
 
+    // sub_tasks.push(metadata_manager.spawn_migrate_table_fragments(
```
wenym1: remove the dead code.

src/meta/src/controller/fragment.rs
```diff
@@ -109,6 +109,7 @@ pub struct FragmentParallelismInfo {
 pub(crate) trait FragmentTypeMaskExt {
```
wenym1: Can use `easy_ext` to avoid explicitly defining a trait here.

src/meta/src/manager/metadata.rs
```diff
@@ -315,6 +315,14 @@ impl MetadataManager {
             .await
     }
 
+    pub async fn list_active_worker_node(
```
wenym1: Seems that we have a single usage. Can call `self.list_worker_node(worker_type, Some(State::Running)).await` directly instead of introducing a new method.

src/meta/service/src/ddl_service.rs
```diff
@@ -119,6 +122,96 @@ impl DdlServiceImpl {
...
+    pub fn start_migrate_table_fragments(&self) -> (JoinHandle<()>, Sender<()>) {
+        tracing::info!("start migrate legacy table fragments task");
+        let env = self.env.clone();
+        let metadata_manager = self.metadata_manager.clone();
+        let ddl_controller = self.ddl_controller.clone();
+
+        let (shutdown_tx, mut shutdown_rx) = tokio::sync::oneshot::channel();
+        let join_handle = tokio::spawn(async move {
+            async fn migrate_inner(
+                shutdown_rx: &mut tokio::sync::oneshot::Receiver<()>,
```
wenym1: Seems no need to pass the `shutdown_rx` here. We can do select for `shutdown_rx` with the whole other processing logic. Code will be like
```
tokio::spawn(async move {
    async fn migrate_inner() {
        while xxx {
            ... // no need to be aware of shutdown_rx here
        }
    }
    select! {
        shutdown_rx => {
               
        }
        migrate_inner() => {
            ...
        }
    }
});
```

If we don't care about the output of select, we can even do `select(shutdown_rx, migrate_inner())` directly without using the select macro.

src/meta/service/src/ddl_service.rs
```diff
@@ -119,6 +122,96 @@ impl DdlServiceImpl {
...
+                metadata_manager: &MetadataManager,
+                ddl_controller: &DdlController,
+            ) -> MetaResult<()> {
+                let tables = metadata_manager
+                    .catalog_controller
+                    .list_unmigrated_tables()
+                    .await?;
+
+                if tables.is_empty() {
+                    return Ok(());
```
wenym1, line 148: May also print a log to show that all tables have been migrated.

src/meta/service/src/ddl_service.rs
```diff
@@ -119,6 +122,96 @@ impl DdlServiceImpl {
...
+                        .get_table_replace_plan(req)
+                        .await
+                        .context("failed to get table replace plan from frontend")?;
+
+                    let plan = resp.into_inner().replace_plan.unwrap();
+                    let replace_info = DdlServiceImpl::extract_replace_table_info(plan);
+                    ddl_controller
+                        .run_command(DdlCommand::ReplaceStreamJob(replace_info))
+                        .await?;
+                    tracing::info!("migrated table fragments for table {}", table.name);
```
wenym1: May include the `table_id` in the log in case of table name duplication, and also the time taken to migrate the table.

wenym1 (commented):
Rest LGTM

src/meta/service/src/ddl_service.rs
```diff
@@ -119,6 +122,96 @@ impl DdlServiceImpl {
...
+            let mut attempt = 0;
+            while let Err(e) =
+                migrate_inner(&mut shutdown_rx, &env, &metadata_manager, &ddl_controller).await
+            {
+                attempt += 1;
+                tracing::error!(
+                    "failed to migrate legacy table fragments: {}, attempt {}, retrying in 5 secs",
+                    e.as_report(),
+                    attempt
+                );
```
BugenZhao: May I know in which case retry could help?

src/meta/src/controller/catalog/list_op.rs
```diff
@@ -269,4 +271,23 @@ impl CatalogController {
         let inner = self.inner.read().await;
         inner.list_functions().await
     }
+
+    pub async fn list_unmigrated_tables(&self) -> MetaResult<Vec<PbTable>> {
```
BugenZhao, line 277: Shall we add some docs/comments defining what "migrate" refers to here?

src/meta/src/controller/catalog/util.rs
```diff
@@ -330,9 +330,7 @@ impl CatalogController {
     }
 
     /// Returns the IDs of tables whose catalogs have been updated.
```
BugenZhao: Let's update the outdated doc.

src/meta/service/src/ddl_service.rs
Chiro11: Any unexpected errors encountered when scheduling or executing DDL, I think.

src/meta/service/src/ddl_service.rs
```diff
@@ -119,6 +122,101 @@ impl DdlServiceImpl {
...
+                                        tracing::info!("shutting down migrate table fragments task");
+                                        break;
+                                    },
+                                    _ = tokio::time::sleep(std::time::Duration::from_secs(5)) => {}
+                                }
+                            }
+                        }
+                    }
+                }
+            }
```
wenym1:
```suggestion
            let migrate_future = async move {
                loop {
                    let res = migrate_inner(&env, &metadata_manager, &ddl_controller);
                    match res {
                        Ok(_) => break,
                        Err(e) => {
                            attempt += 1;
                            tracing::error!(
                                "failed to migrate legacy table fragments: {}, attempt {}, retrying in 5 secs",
                                e.as_report(),
                                attempt
                            );
                            tokio::time::sleep(std::time::Duration::from_secs(5)).await;
                        }
                    }
                }
            }
            select(pin!(shutdown_rx), pin!(migrate_future)).await;
```

wenym1 (approved):
Rest LGTM

src/meta/service/src/ddl_service.rs
```diff
@@ -119,6 +122,101 @@ impl DdlServiceImpl {
...
+
+                let client = {
+                    let workers = metadata_manager
+                        .list_worker_node(Some(WorkerType::Frontend), Some(State::Running))
+                        .await?;
+                    if workers.is_empty() {
+                        return Err(anyhow::anyhow!("no active frontend nodes found").into());
+                    }
+                    let worker = workers.choose(&mut thread_rng()).unwrap().clone();
+                    env.frontend_client_pool().get(&worker).await?
```
yezizp2012:
```suggestion
                    let worker = workers.choose(&mut thread_rng()).unwrap();
                    env.frontend_client_pool().get(worker).await?
```

yezizp2012 (approved):
Rest LGTM

PR #23202, opened 2025-09-15
fix(vector): fix vector search distribution inference and execution

wenym1:

In the very first implementation of `BatchVectorSearch` plan node, it's a leaf node with `SomeShards` distribution, and we have code to manually hack inside the fragment scheduler. However, in later implementation, we have changed to let it accept an input, and the plan node works as expanding the input with an extra column of the vector search result. Therefore, in this new implementation, the plan node will have the same distribution as the input, rather than `SomeShards`.

In this PR, we will change to use the input distribution as the distribution of `BatchVectorSearch`, and the hack code in fragment scheduler is removed. Besides, we changed to slt test to test execution under both distributed and local query mode.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/e2e_test/vector_search/vector_nearest.slt.part
+++ b/e2e_test/vector_search/vector_nearest.slt.part
@@ -45,7 +45,7 @@ explain(verbose) select * from query_view;
 ----
 <slt:ignore>BatchProject { exprs: [Field(Unnest($1), 1:Int32) as $expr1, Field(Unnest($1), 0:Int32) as $expr2, Pow(Field(Unnest($1), 2:Int32), 2:Float64)::Int32 as $expr3] }
   └─BatchProjectSet { select_list: [Unnest($1)] }
-    └─BatchVectorSearch { schema: [:Vector(3), vector_info:List(Struct(StructType { fields: [("text", Varchar), ("items.id", Int32), ("__distance", Float64)] }))], top_n: 2, distance_type: L2Sqr, index_name: "i", vector:  }
+    └─BatchVectorSearch { schema: [query_vector:Vector(3), vector_info:List(Struct(StructType { fields: [("text", Varchar), ("items.id", Int32), ("__distance", Float64)] }))], top_n: 2, distance_type: L2Sqr, index_name: "i", vector: query_vector }
       └─BatchValues { rows: [['[3,2,1]':Vector(3)]] }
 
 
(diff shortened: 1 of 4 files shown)


wenym1:
* **#23202** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23202?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23202?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/frontend/src/optimizer/mod.rs
```diff
@@ -1223,69 +1224,55 @@ fn exist_and_no_exchange_before(
...
-/// As we always run the root stage locally, for some plan in root stage which need to execute in
-/// compute node we insert an additional exhchange before it to avoid to include it in the root
-/// stage.
-///
-/// Returns `true` if we must insert an additional exchange to ensure this.
-fn require_additional_exchange_on_root_in_distributed_mode(plan: BatchPlanRef) -> bool {
-    fn is_user_table(plan: &BatchPlanRef) -> bool {
-        plan.node_type() == BatchPlanNodeType::BatchSeqScan
+impl BatchPlanRef {
+    fn is_user_table(&self) -> bool {
```
BugenZhao:
```suggestion
    fn is_user_table_scan(&self) -> bool {
```

src/frontend/src/optimizer/mod.rs
```diff
@@ -1223,69 +1224,55 @@ fn exist_and_no_exchange_before(
...
+        self.node_type() == BatchPlanNodeType::BatchSeqScan
+            || self.node_type() == BatchPlanNodeType::BatchLogSeqScan
+            || self.node_type() == BatchPlanNodeType::BatchVectorSearch
     }
 
-    fn is_log_table(plan: &BatchPlanRef) -> bool {
-        plan.node_type() == BatchPlanNodeType::BatchLogSeqScan
+    fn is_source(&self) -> bool {
```
BugenZhao:
```suggestion
    fn is_source_scan(&self) -> bool {
```

e2e_test/vector_search/vector_nearest.slt.part
```diff
@@ -95,27 +96,26 @@ create index i on items using flat (get_embedding(text)) include(text) with (dis
...
 
 # ensure that vector index is used
 query T
 explain(verbose) select * from query_view;
 ----
-BatchExchange { order: [], dist: Single }
-└─BatchProject { exprs: [Field(Unnest($1), 1:Int32) as $expr1, Field(Unnest($1), 0:Int32) as $expr2, Pow(Field(Unnest($1), 2:Int32), 2:Float64)::Int32 as $expr3] }
+<slt:ignore>BatchProject { exprs: [Field(Unnest($1), 1:Int32) as $expr1, Field(Unnest($1), 0:Int32) as $expr2, Pow(Field(Unnest($1), 2:Int32), 2:Float64)::Int32 as $expr3] }
   └─BatchProjectSet { select_list: [Unnest($1)] }
     └─BatchVectorSearch { schema: [:Vector(3), vector_info:List(Struct(StructType { fields: [("text", Varchar), ("items.id", Int32), ("__distance", Float64)] }))], top_n: 2, distance_type: L2Sqr, index_name: "i", vector:  }
```
BugenZhao: Why there's no field name for the column with type `Vector(3)`? 🤔

src/frontend/src/optimizer/mod.rs
```diff
@@ -392,12 +393,6 @@ impl BatchPlanRoot {
         // Convert to distributed plan
         plan = plan.to_distributed_with_required(&self.required_order, &self.required_dist)?;
 
-        // Add Project if the any position of `self.out_fields` is set to zero.
-        if self.out_fields.count_ones(..) != self.out_fields.len() {
-            plan =
-                BatchProject::new(generic::Project::with_out_fields(plan, &self.out_fields)).into();
-        }
-
```
BugenZhao, line 400: Why do we need to move this?

BugenZhao (approved):
LGTM

src/frontend/src/optimizer/mod.rs
wenym1: Because the order column might not be included in the output, but it might still be used in the later in `BatchExchange` if `require_additional_exchange_on_root_in_distributed_mode`. This is also the handling order for local mode. It might be a bug previously, but just rarely be hit.

e2e_test/vector_search/vector_nearest.slt.part
wenym1: Because I used `Field::unnamed` to initialize the plan node schema. Seems better to provide a name to it.

PR #23208, opened 2025-09-16
refactor(frontend): reduce reserved keywords in explain options

wenym1:

Avoid reserving the following keywords for explain options. 
```
let explain_key_words = [
    Keyword::BACKFILL,
    Keyword::TRACE,
    Keyword::LOGICAL,
    Keyword::PHYSICAL,
    Keyword::DISTSQL,
    Keyword::DURATION_SECS,
    ...
];
```

When parsing explain options, instead of parsing as keyword, we will change to parse identifiers.

## Checklist

## Documentation

<details>

</details>


wenym1:
* **#23208** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23208?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23208?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/sqlparser/src/keywords.rs
```diff
@@ -98,6 +98,7 @@ define_keywords!(
     AVG,
     BACKFILL,
     BASE64,
+    BATCH,
```
xiangjinwu: Avoid introducing new keywords unless really necessary; it can be a breaking change for users using this word as name in *some places*.

See #22828 or [PostgreSQL](https://github.com/postgres/postgres/blob/REL_17_6/src/backend/parser/gram.y#L11969) on how to use predefined words rather than keywords - the parser does not need to understand these words; only the logic handling them does. This can also help avoid repeating the enum in sqlparser (similar benefit as #22344)

wenym1: Agree. Actually, we have defined many similar keywords for this explain options, such as `BACKFILL`, `LOGICAL`, `TRACE`. These keywords are only used in the explain options. I think we can have a separate PR to make them not a keyword.

BugenZhao:
> Added a new explain type `Batch` to display the batch logical plan

This sounds a bit confusing to me. In almost every other part of the optimizer, we treat `Batch` as a type of `Physical` and there's no such concept of "batch logical".

> to ensure that the same plan is displayed for both local and distributed query mode

May I know why this is worth testing or enforcing? Also, in planner test this can already be done with `- batch_local_plan` and `- batch_plan`, does it work for you?

BugenZhao:
> May I know why this is worth testing or enforcing?

If the goal is to let the test pass under both local mode and distributed mode for cases under `e2e_test/batch`, then I would strongly suggest moving `EXPLAIN` tests into planner tests. Actually it may not be a good practice to test `EXPLAIN` output in e2e tests.

wenym1:
> > to ensure that the same plan is displayed for both local and distributed query mode
> 
> May I know why this is worth testing or enforcing? Also, in planner test this can already be done with `- batch_local_plan` and `- batch_plan`, does it work for you?

You can see e2e_test/vector_search/vector_nearest.slt. In https://github.com/risingwavelabs/risingwave/pull/23202, I am trying to use the same test slt file for both local and distributed, in the way as `set query_mode = xxx; include vector_nearest.slt.part` in the slt files to test the two query modes. I hope to include a explain output in the slt file to ensure that vector index is used (by inspecting whether the plan includes `BatchVectorSearch`). But it turns out that the physical plan of the two modes are different, so I hope to display the explain output  of plan before turning into a physical plan, which is the same for both query modes. Do you have any suggestion on achieving the same?

If the option name `BATCH` can be confusing, how about name like `BATCH_OPTIMIZED_LOGICAL`?

src/sqlparser/src/keywords.rs
wenym1: Seems non-trivial to achieve this.
```
// In order to support following statement, we need to peek before consume.
// explain (select 1) union (select 1)
if self.peek_token() == Token::LParen
    && self.peek_nth_any_of_keywords(1, &explain_key_words)
    && self.consume_token(&Token::LParen)
{
    self.parse_comma_separated(parse_explain_option)?;
    self.expect_token(&Token::RParen)?;
}
```

We used to check whether the pair of parenthess after `explain` is a query or explain options by checking whether the first token is any keyword of explain options. If we want to treat these options to be non-keyword, these won't be a easy way to distinguish the two cases. @xiangjinwu @BugenZhao Do you have any advices?

wenym1: To achieve this, we have unnecessarily already the following keywords that are not included in https://www.postgresql.org/docs/14/sql-keywords-appendix.html#KEYWORDS-TABLE
```
let explain_key_words = [
    Keyword::BACKFILL,
    Keyword::TRACE,
    Keyword::LOGICAL,
    Keyword::PHYSICAL,
    Keyword::DISTSQL,
    Keyword::DURATION_SECS,
    ...
];
```

wenym1:
> > May I know why this is worth testing or enforcing?
> 
> If the goal is to let the test pass under both local mode and distributed mode for cases under `e2e_test/batch`, then I would strongly suggest moving `EXPLAIN` tests into planner tests. Actually it may not be a good practice to test `EXPLAIN` output in e2e tests.

Planner test already included similar tests. Just want to ensure vector index is used in the slt test, because the query output will be the same with and without using vector index. Or do we have anyway to check whether the output of a query contains a certain string in sqllogictest?

BugenZhao:
> Or do we have anyway to check whether the output of a query contains a certain string in sqllogictest?

Maybe https://github.com/risinglightdb/sqllogictest-rs/pull/266? Though I never used that before. 🤣

wenym1:
> > Or do we have anyway to check whether the output of a query contains a certain string in sqllogictest?
> 
> Maybe [risinglightdb/sqllogictest-rs#266](https://github.com/risinglightdb/sqllogictest-rs/pull/266)? Though I never used that before. 🤣

Thanks! It really works in https://github.com/risingwavelabs/risingwave/pull/23202

I have changed this PR to be a refactor PR to apply https://github.com/risingwavelabs/risingwave/pull/23208#discussion_r2351299033 to reduce reserved keywords specially for explain options. @BugenZhao @xiangjinwu PTAL

src/sqlparser/src/ast/mod.rs
```diff
@@ -1195,7 +1195,7 @@ impl fmt::Display for ExplainFormat {
     }
 }
 
-#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
```
xiangjinwu, line 1198: Why not just call `clone` when needed?

src/sqlparser/src/parser.rs
```diff
@@ -4363,70 +4374,67 @@ impl Parser<'_> {
...
+                }
             };
             Ok(())
         };
 
         // In order to support following statement, we need to peek before consume.
         // explain (select 1) union (select 1)
         if self.peek_token() == Token::LParen
-            && self.peek_nth_any_of_keywords(1, &explain_key_words)
+            && self.peek_nth_any_of_uppercase_identifiers(1, &explain_options_identifiers)
```
xiangjinwu: The common practice is to peek for keywords that starts a [`ExplainableStmt`](https://github.com/postgres/postgres/blob/REL_17_6/src/backend/parser/gram.y#L11979) (select, with, values, create, insert, etc) and everything else would be treated equally as identifiers.

Furthermore, being identifiers rather than keywords means they shall follow the same casing rules as identifiers used at other places (see https://github.com/risingwavelabs/risingwave/pull/22828#discussion_r2303119623). The current usage of `to_uppercase` brutally accepts all forms.
|sql|identifier|keyword|
|-|-|-|
|verbose|⭕|⭕|
|verBOSE|⭕|⭕|
|VERBOSE|⭕|⭕|
|"verbose"|⭕|❌|
|"verBOSE"|❌|❌|
|"VERBOSE"|❌|❌|

src/sqlparser/src/ast/mod.rs
wenym1: To simplifiy code, to avoid the redudant `.clone`.

These plain primitive options are inherently capable of being `Copy`, and I don't see any downside on deriving Copy, and meanwhile it can make our code cleaner.

src/sqlparser/src/parser.rs
wenym1: > The common practice is to peek for keywords that starts a ExplainableStmt (select, with, values, create, insert, etc) and everything else would be treated equally as identifiers.

Agree. But it seems hard for us to implement in this way. PG's SQL parser support probing over a whole parser rule, and reprobe on the next rule when current rule does not match the input, so it define the explain statement by declaring several capable syntaxes. 

However, we only support token level probing, which means we have to maintain the list keywords that an `ExplainableStmt` may start with, and everytime when we support a new syntax, we should remember updating the list, and this can be easily forgotten, and it seems no automatic way (such as CI test) to detect the inconsistency.

> Furthermore, being identifiers rather than keywords means they shall follow the same casing rules as identifiers used at other places (see #22828 (comment)). The current usage of to_uppercase brutally accepts all forms.

So it seems that the correct imlementation should be, do `to_uppercase` check and meanwhile reject all idents that are quoted?

xiangjinwu: > So it seems that the correct imlementation should be, do `to_uppercase` check and meanwhile reject all idents that are quoted?

No. I did not mean we shall use identifier to mimic the keyword behavior. I meant they are real identifiers in PostgreSQL and so should in RisingWave - accept any case when unquoted and lowercase when quoted. (You can test in PostgreSQL.) This can be achieved the same way as any other identifier: call `.real_value()` and compare against the lowercase form. AFAIK a PostgreSQL parser shall NEVER call `to_uppercase`

src/sqlparser/src/ast/mod.rs
xiangjinwu: To make the explain options fully sqlparser-agnostic, so that future extensions only update the handler/binder but not sqlparser, it would be a `Vec<(ident, Option<ident>)>`
https://www.postgresql.org/docs/17/sql-explain.html
https://github.com/postgres/postgres/blob/REL_17_6/src/backend/parser/gram.y#L11871-L11872

But let's put this argument on hold until when that is a thing.

src/sqlparser/src/parser.rs
xiangjinwu: > we have to maintain the list keywords that an `ExplainableStmt` may start with, and everytime when we support a new syntax, we should remember updating the list, and this can be easily forgotten, and it seems no automatic way (such as CI test) to detect the inconsistency.

TLDR: I am okay with keeping this for now as long as the peeking logic is updated to align with the case handling same as actual parsing.

My take is, inventing fancy new syntax shall be a rare thing; and when it happens, compatibility with existing syntax shall be considered. That's why I mentioned `WITH USING` in https://github.com/risingwavelabs/risingwave/pull/21356#discussion_r2046116604 because `WITH` is already a keyword that starts `ExplainableStmt`

wenym1: Changed to reject all quoted idents in explain options. @xiangjinwu PTAL

xiangjinwu: >  I meant they are real identifiers in PostgreSQL and so should in RisingWave - accept any case when unquoted and lowercase when quoted. (You can test in PostgreSQL.) This can be achieved the same way as any other identifier: call `.real_value()` and compare against the lowercase form. AFAIK a PostgreSQL parser shall NEVER call `to_uppercase`

Maybe my first message was overlooked.

wenym1: Indeed. The comments were not loaded when I made the code change. 

Just updated to use `.real_value()` and check with lowercase string.

Issue #22902, opened 2025-08-15
ingest non-append-only source without materializing all historical data


BugenZhao:
Currently, it's required to create a `TABLE` to ingest data from non-append-only source. This is because a lot of streaming operators in our system can only work with retract data (besides append-only data), that is, a `Delete` or `UpdateDelete` is required before updating a row. Non-append-only sources usually cannot comply with this, either because it's an upsert source, or because we cannot trust the input.

https://github.com/risingwavelabs/risingwave/issues/22799 proposed that we may distinguish between retractable stream and upsert stream in the optimizer. As a result, some stateless or special operators are now allowed to operate on upsert or untrusted streams. This could enable us to support bare non-append-only `SOURCE` without materializing all historical data as a `TABLE`. Some typical patterns to use with this source include:

- Pure ETL, sinking upsert data to external systems directly (achievable with `SinkExecutor` accepting upsert stream)
- Simple filtering or transformation on the source before materializing into our system (achievable with `CREATE SINK INTO TABLE`, or `Materialize` normalizing upsert stream to retract stream)
- Persist only the recent data from the source (achievable with `TemporalFilter` normalizing upsert stream to retract stream)

PR #23209, opened 2025-09-16
feat(streaming): introduce upsert behavior for filter executor

BugenZhao:

Background:
- #22799 
- #22902 

As discussed in #22856, it appears that the behavior of `Filter` executor when handling upsert stream should differ from when handling retractable stream. Specifically, we should

- always yield `Delete` records on `Delete` (or `UpdateDelete`) input, regardless of the filter result, to ensure the existing record (if any) can be cleaned up
- always yield `Delete` records on `Insert` (or `UpdateInsert`) input when the filter result is `false`, to ensure the existing record (if any) can be cleaned up

See more details in code and documentation.

Note that we won't actually hit this code path solely with this PR. Let's verify this by code coverage report.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/executor/filter.rs
+++ b/src/stream/src/executor/filter.rs
@@ -23,7 +23,18 @@ use crate::executor::prelude::*;
 /// and returns a boolean array on whether each item should be retained. And then,
 /// `FilterExecutor` will insert, delete or update element into next executor according
 /// to the result of the expression.
-pub struct FilterExecutor {
+///
+/// # Upsert mode
+///
+/// When `UPSERT` is specified, the filter executor will...
+/// - always yield `Delete` records on `Delete` (or `UpdateDelete`) input, regardless of the filter result
+/// - always yield `Delete` records on `Insert` (or `UpdateInsert`) input, if the filter result is `false`
+///
+/// This is to guarantee that the filtered-out data is correctly cleaned up, even if we
+/// don't have information about the previous data under the same primary key. Since an
+/// upsert stream is expected as the output of the filter executor in this case, we can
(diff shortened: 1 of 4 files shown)


BugenZhao:
* **#23211** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23211?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#22856** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/22856?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#23209** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23209?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23209?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/stream/src/executor/filter.rs
```diff
@@ -64,44 +78,68 @@ impl FilterExecutor {
...
-                        new_visibility.append(false);
                         new_visibility.append(true);
                     }
-                    (true, true) => {
-                        new_ops.push(Op::UpdateDelete);
-                        new_ops.push(Op::UpdateInsert);
-                        new_visibility.append(true);
-                        new_visibility.append(true);
+                }
+            } else {
```
BugenZhao, line 107: The diff is messy. There's actually no change in `!UPSERT` arm.

chenzl25 (approved):
LGTM

PR #23210, opened 2025-09-16
fix: shared_buffer_compact parallelism

Li0k:

This pull request improves the key splitting logic in the shared buffer compaction process to ensure correct ordering and prevent duplicate or overlapping key ranges. The changes also add a new test to validate these improvements. The most important changes are grouped below:

### Key splitting and sorting improvements

* The sorting of split keys in `generate_splits` now uses `KeyComparator::compare_encoded_full_key` for correct ordering of encoded full keys, replacing the previous direct comparison. [[1]](diffhunk://#diff-5901c207c104c283470919bd4ffd6fc7a2cd32932fab7ffc2767bd07f86e84e0L30-R30) [[2]](diffhunk://#diff-5901c207c104c283470919bd4ffd6fc7a2cd32932fab7ffc2767bd07f86e84e0L476-L486)
* The logic for generating splits was refactored to use encoded full keys and avoid duplicate split points, ensuring that splits are created only when the key changes and the buffer size threshold is met. [[1]](diffhunk://#diff-5901c207c104c283470919bd4ffd6fc7a2cd32932fab7ffc2767bd07f86e84e0L476-L486) [[2]](diffhunk://#diff-5901c207c104c283470919bd4ffd6fc7a2cd32932fab7ffc2767bd07f86e84e0L500-R521)

### Testing enhancements

* Added a new test `test_generate_splits_no_duplicate_keys` to verify that splits are continuous, monotonic, non-overlapping, and sorted, even when input keys are out of order. This test helps prevent regressions in split logic.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/config/ci-backfill.toml
+++ b/src/config/ci-backfill.toml
@@ -4,6 +4,9 @@ telemetry_enabled = false
 [streaming.developer]
 stream_chunk_size = 10
 
+[storage]
+share_buffers_sync_parallelism = 4
+
 [storage.object_store.retry]
 streaming_upload_attempt_timeout_ms = 20000
 
--- a/src/config/ci-compaction-test-meta.toml
+++ b/src/config/ci-compaction-test-meta.toml
@@ -6,6 +6,7 @@ max_heartbeat_interval_secs = 60
 in_flight_barrier_nums = 40
 
 [storage]
+share_buffers_sync_parallelism = 4
 shared_buffer_capacity_mb = 4096
 block_cache_capacity_mb = 4096
 meta_cache_capacity_mb = 1024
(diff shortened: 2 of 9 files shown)


src/storage/src/hummock/compactor/shared_buffer_compact.rs
Li0k, line 536: How about adding an non-overlap verification for the splits?

cc @hzxa21

Li0k:
nips: We can filter out unnecessary imm using the **split-key** before constructing the shared-buffer-iter.

src/storage/src/hummock/compactor/shared_buffer_compact.rs
hzxa21: There is only one place to generate the splits in L508-L512. Based on the logic their, the non-overlapping property is guaranteed so I think adding the verification here is useless.

The verification logic is useful if we have many other places to generate the split.

Li0k: Thansk.

hzxa21 (approved):

PR #23212, opened 2025-09-16
refactor(source): remove actor_splits in StreamJobFragments

shanicky:

Corresponding to PR #23204, let's see if the tests can pass.

Fix the semantic redundancy between `actor_splits` and `InitSplitAssignment`.

## Checklist

### DIFF
--- a/src/meta/src/controller/streaming_job.rs
+++ b/src/meta/src/controller/streaming_job.rs
@@ -432,11 +432,15 @@ impl CatalogController {
         for_replace: bool,
         creating_streaming_job: Option<&'a StreamingJob>,
     ) -> MetaResult<()> {
+        // NOTE: we don't need to pass actor splits for now
+        // splits field will be updated during the post_collect_job_fragments stage
+        let empty_actor_splits = Default::default();
+
         let fragment_actors = Self::extract_fragment_and_actors_from_fragments(
             job_id,
             get_fragments(),
             actor_status,
-            &Default::default(),
+            &empty_actor_splits,
         )?;
         let inner = self.inner.write().await;
 
--- a/src/meta/node/src/server.rs
+++ b/src/meta/node/src/server.rs
@@ -414,6 +414,25 @@ pub async fn start_service_as_election_leader(
     });
     let trace_srv = otlp_embedded::TraceServiceImpl::new(trace_state.clone());
 
+    #[cfg(not(madsim))]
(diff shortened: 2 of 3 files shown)


src/meta/src/controller/fragment.rs
```diff
@@ -352,7 +352,6 @@ impl CatalogController {
             state: state as _,
             fragments: pb_fragments,
             actor_status: pb_actor_status,
-            actor_splits: pb_actor_splits,
```
wenym1, line 355: Code to construct `pb_actor_splits` can also be removed.

src/meta/src/controller/streaming_job.rs
```diff
@@ -410,7 +410,6 @@ impl CatalogController {
             stream_job_fragments.stream_job_id().table_id as _,
```
wenym1, line 410: Does it mean that in previous code, before calling `prepare_stream_job_fragments`, `actor_splits` in `StreamJobFragments` must be empty? `prepare_stream_job_fragments` is called in `create_streaming_job_inner` and `replace_job` of `DdlController`. Can we confirm this?

shanicky: The code related to the split assignment is all retrieved separately from the source manager after the prepare step, so it’s actually not used here.

src/meta/src/model/stream.rs
```diff
@@ -334,7 +340,7 @@ impl StreamJobFragments {
                 })
                 .collect(),
             actor_status: self.actor_status.clone().into_iter().collect(),
-            actor_splits: build_actor_connector_splits(&self.actor_splits),
+            actor_splits,
```
tabVersion: why we still need to persist in `PbStreamJobFragments`? I thought the field should be deprecated?

src/meta/src/model/stream.rs
```diff
@@ -256,10 +255,6 @@ pub struct StreamJobFragments {
     /// The status of actors
     pub actor_status: BTreeMap<ActorId, ActorStatus>,
 
-    /// The splits of actors,
-    /// incl. both `Source` and `SourceBackfill` actors.
-    pub actor_splits: HashMap<ActorId, Vec<SplitImpl>>,
```
tabVersion, line 261: we remove `actor_splits` from `StreamJobFragments` but adding it back to the protobuf one, why?

shanicky: Actually, we can also remove the actor splits in the protobuf. I’ll give it a try.

PR #23215, opened 2025-09-16
feat: unify the notification for background and foreground job and refactoring cancel process

yezizp2012:

Related #23148 . The PR also fixed missing apply of `CreateJobFinished` source change for recovered streaming job.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/meta/src/barrier/progress.rs
+++ b/src/meta/src/barrier/progress.rs
@@ -569,7 +569,10 @@ impl CreateMviewProgressTracker {
                 // The command can be finished immediately.
                 return Some(TrackingJob::new(
                     info.stream_job_fragments.stream_job_id.table_id as _,
-                    None,
+                    Some(SourceChange::CreateJobFinished {
+                        finished_backfill_fragments: stream_job_fragments
+                            .source_backfill_fragments(),
+                    }),
                 ));
             }
             (info.clone(), actors)
--- a/src/meta/src/stream/stream_manager.rs
+++ b/src/meta/src/stream/stream_manager.rs
@@ -317,7 +317,7 @@ impl GlobalStreamManager {
         let table_id = stream_job_fragments.stream_job_id();
         let database_id = ctx.streaming_job.database_id().into();
 
-        let (cancel_tx, mut cancel_rx) = oneshot::channel();
+        let (cancel_tx, cancel_rx) = oneshot::channel();
         let execution = StreamingJobExecution::new(table_id, cancel_tx, permit);
         self.creating_job_info.add_job(execution).await;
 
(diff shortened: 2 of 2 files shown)


src/meta/src/barrier/progress.rs
```diff
@@ -222,74 +223,78 @@ impl Progress {
...
 ///    On recovery, the stream manager will stop managing the job.
-/// 2. `Recovered`. This refers to the "Recovered" type of tracking job.
+/// 2. if `is_recovered` is true, it is a "Recovered" tracking job.
 ///    On recovery, the barrier manager will recover and start managing the job.
-pub enum TrackingJob {
-    New(TrackingCommand),
-    Recovered(RecoveredTrackingJob),
+pub struct TrackingJob {
+    job_id: ObjectId,
+    is_recovered: bool,
```
wenym1, line 238: Seems that the `is_recovered` is used only to have different display format. If so, we may simplify and unify them and does not store this information.

src/meta/src/barrier/progress.rs
```diff
@@ -550,9 +562,10 @@ impl CreateMviewProgressTracker {
             let actors = stream_job_fragments.tracking_progress_actor_ids();
             if actors.is_empty() {
                 // The command can be finished immediately.
-                return Some(TrackingJob::New(TrackingCommand {
-                    job_id: info.stream_job_fragments.stream_job_id,
-                }));
+                return Some(TrackingJob::new(
+                    info.stream_job_fragments.stream_job_id.table_id as _,
+                    None,
```
wenym1: Should it also be `Some(SourceChange::CreateJobFinished {
                        finished_backfill_fragments: table_fragments.source_backfill_fragments(),
                    }),`?

src/meta/src/stream/stream_manager.rs
```diff
@@ -326,135 +316,84 @@ impl GlobalStreamManager {
...
-                        let _ = finish_tx.send(()).inspect_err(|_| {
-                            tracing::warn!("failed to notify cancelled: {table_id}")
-                        });
-                        self.creating_job_info.delete_job(table_id).await;
-                        return Err(MetaError::cancelled("create"));
+        let result = tokio::select! {
+            biased;
+
+            res = create_fut => res,
+            notifier = &mut cancel_rx => {
```
wenym1:
```suggestion
            notifier = cancel_rx => {
```

Can consume it directly, since we are not doing select in a loop.

src/meta/src/barrier/progress.rs
yezizp2012: Yes, I kept this information so that it can be distinguished in trace and debug logs, which is still somewhat useful for troubleshooting.

PR #23216, opened 2025-09-16
fix: update dropped table cache for pre-applied commands in recovery

yezizp2012:

Follow up of #23087 . Normally they will be handled once the barrier has been collected.

## Checklist

## Documentation

<details>

</details>


src/meta/src/controller/streaming_job.rs
```diff
@@ -652,23 +652,25 @@ impl CatalogController {
...
-                ),
-            )
-            .all(&txn)
-            .await?
-            .into_iter()
-            .map(|(table, obj)| PbTable::from(ObjectModel(table, obj.unwrap())));
-        inner
-            .dropped_tables
-            .extend(dropped_tables.map(|t| (TableId::try_from(t.id).unwrap(), t)));
+        if is_cancelled {
```
wenym1, line 657: Why this logic is only for cancel?

src/meta/src/barrier/context/recovery.rs
```diff
@@ -370,7 +370,13 @@ impl GlobalBarrierWorkerContextImpl {
                     tracing::info!("recovered background job progress");
 
                     // This is a quick path to accelerate the process of dropping and canceling streaming jobs.
-                    let _ = self.scheduled_barriers.pre_apply_drop_cancel(None);
+                    let dropped_table_ids = self.scheduled_barriers.pre_apply_drop_cancel(None);
```
wenym1: In global recovery, I think we can force flush all pending deleting state tables, instead of only flush the pending state tables in the queue.

wenym1 (commented):
Rest LGTM

src/meta/src/controller/streaming_job.rs
yezizp2012: This function is only used when drop fails to clean up, except when it is called by cancel. This part is not cached.

src/meta/src/barrier/context/recovery.rs
yezizp2012: There remains a race condition. Forcing a flush of all pending state tables might inadvertently remove tables that were included in a drop command issued after `pre_apply_drop_cancel`. This could cause the previous flaky test issues.

wenym1: The previous flaky test was caused by actor reporting errors on seeing inconsistency, and then trigger recovery to cause ddl failure. Since here we are already under global recovery, we can expect any failure and don't have to worry about the consistency.

wenym1 (approved):

PR #23221, opened 2025-09-16
fix(sink): Fix dead lock in alter-table with both upstream and downstream

### LINKED ISSUE #23064: Unable to add column to table that has both upstream sink and downstream dependency
### Describe the bug

If we create `sink -> table -> mv`, we cannot add columns in table.

### Error message/log

It will get stuck.

### To Reproduce

```sql
create table t2 (v1 int primary key, v2 int);
create sink s1 into t2 from t1;
create materialized view mv1 as select v1 from t2;

alter table t2 add column v3 int default 100;
```

### Expected behavior

_No response_

_No response_

### The version of RisingWave

_No response_

### Additional context

_No response_

Chiro11:

Close #23064

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/executor/upstream_sink_union.rs
+++ b/src/stream/src/executor/upstream_sink_union.rs
@@ -19,6 +19,7 @@ use std::task::{Context, Poll};
 
 use anyhow::Context as _;
 use futures::future::try_join_all;
+use itertools::Itertools;
 use pin_project::pin_project;
 use risingwave_common::catalog::Field;
 use risingwave_expr::expr::{EvalErrorReport, NonStrictExpression, build_non_strict_from_prost};
@@ -178,7 +179,7 @@ pub struct UpstreamSinkUnionExecutor {
     chunk_size: usize,
 
     /// The initial inputs to the executor.
-    initial_upstream_infos: Vec<UpstreamFragmentInfo>,
+    initial_inputs: Vec<BoxedSinkInput>,
 
     /// The error report for evaluation errors.
     eval_error_report: ActorEvalErrorReport,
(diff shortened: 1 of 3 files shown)


BugenZhao:
I wonder which step in your explanation of deadlock is the newly introduced behavior from `UpstreamSinkUnion`? Because it seems not reproducible in v2.5.1 release.

src/stream/src/executor/upstream_sink_union.rs
```diff
@@ -205,25 +209,41 @@ impl Execute for UpstreamSinkUnionExecutor {
 }
 
 impl UpstreamSinkUnionExecutor {
-    pub fn new(
+    pub async fn new(
```
BugenZhao, line 213: Can you add some comments to show that when this function might be blocking (i.e. not immediately ready)? Seems like the origin is `RemoteInput::new`.

Chiro11: Yes, here we need to wait for establishing stream-rpc with the upstream actor.

Chiro11:
> I wonder which step in your explanation of deadlock is the newly introduced behavior from `UpstreamSinkUnion`? Because it seems not reproducible in v2.5.1 release.

Only on the main branch. `UpstreamSinkUnion` operator has been temporarily banned on the release branch.

wenym1:
Discussed offline.

The dependency of `upstream-dispatch-barrier -> create-new-job` is avoidable. This dependency is due to the current problematic implementation of `Merge`, rather than the high-level design. The dependency comes from the current implementation that, even though the barrier with update mutation has been sent from local barrier manager, we still have to wait for upstream barrier before we apply the update mutation and create inputs to new upstream.

To solve this, we can change to, while polling upstreaming messages, we can concurrently poll new barrier with mutation from barrier_rx, and then create new inputs when needed, so that the two won't have any dependency, and then the mentioned deadlock can be avoided.

Chiro11:
Merged to #23273

PR #23238, opened 2025-09-17
Postgres composite types

dclausen:

This change adds support for Postgres composite types and converts them into Risingwave's struct format. At work, we make heavy use of rows that contain multiple columns that represent money using a composite type creatively named `money_with_currency`.  

In postgres, an example would look like:

| total_amount | |
|----------|--------|
| currency | amount |
| USD      | 0      |

All existing CDC tools and products (Datastream, Conduit) that I evaluated (except vanilla Debezium) and are still actively maintained either ignore columns with composite types or simply error. We have evaluated RisingWave and determined it to be a great fit for our use case in powering a streaming ETL for financial reporting. This is the only blocker for us to adopt RisingWave.

### DIFF
--- a/src/connector/src/parser/scalar_adapter.rs
+++ b/src/connector/src/parser/scalar_adapter.rs
@@ -18,6 +18,7 @@ use anyhow::anyhow;
 use bytes::BytesMut;
 use pg_bigdecimal::PgNumeric;
 use risingwave_common::types::{DataType, Decimal, Int256, ListValue, ScalarImpl, ScalarRefImpl};
+use risingwave_common::util::iter_util::ZipEqFast;
 use thiserror_ext::AsReport;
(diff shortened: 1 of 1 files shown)


hzxa21:
The diff seems a bit off. Can you try rebasing onto the latest main?

dclausen:
closing since it looks like 2.7.0 now recognizes user defined types (composite types) as varchar.

hzxa21:
> closing since it looks like 2.7.0 now recognizes user defined types (composite types) as varchar.

Can you elaborate more? Are you saying that the issue has been resolved in the latest main branch?

dclausen:
I missed that there are 2 independent code paths, one for postgres_cdc and one for the backfill. Closing in favor of a different approach that handles both.

Issue #23177, opened 2025-09-11
bug(changelog): `_changelog_row_id` out of order


xiangjinwu:
### Describe the bug

https://docs.risingwave.com/sql/commands/sql-as-changelog
> `_changelog_row_id`: a monotonically increasing identifier that captures the order of the change.

See the reproduction below for a case where it does not hold.

### Error message/log

```text
No error message but produces wrong results
```

### To Reproduce

```
create table t (c1 int);
create table wide (c1 int primary key, even int, odd int);
create sink even into wide (c1, even) as select c1, c1*2 from t;
create sink odd into wide (c1, odd) as select c1, c1*2+1 from t;
create materialized view mv as with p as changelog from wide select _changelog_row_id::bigint as changelog_row_id, * from p;
insert into t select generate_series(1020300, 1020399);

with add as (select * from mv where changelog_op = 1), update as (select * from mv where changelog_op = 3) select add.c1 from add, update where add.c1 = update.c1 and add.changelog_row_id > update.changelog_row_id;
```

In this example, each new row in `t` would trigger 2 sinks writing into the same row of `wide` - the earlier one shall have `changelog_op = 1` and a smaller `changelog_row_id` while the later one shall have `changelog_op = 3` and a larger `changelog_row_id`.

However, the last adhoc query gives non-empty results for such violating rows in changelogs persisted in `mv`.

### Expected behavior

As mentioned above, for the 2 entries of same `c1` in `mv`, `changelog_op = 1` corresponds to smaller `changelog_row_id` and `changelog_op = 3` corresponds to larger `changelog_row_id`. The result set of last adhoc query shall always be empty.

### The version of RisingWave

2.5.1

### Additional context

Looks like changelog relies on row_id_gen, which is NOT monotonically increasing.

https://github.com/risingwavelabs/risingwave/blob/7bcb5c6c655e0cf8cd7e629bada490abc54b7e46/src/stream/src/executor/changelog.rs#L61-L64

https://github.com/risingwavelabs/risingwave/blob/7bcb5c6c655e0cf8cd7e629bada490abc54b7e46/src/common/src/util/row_id.rs#L174-L192

PR #23240, opened 2025-09-18
fix(stream): fix _changelog_row_id out of order

xxhZs:

https://github.com/risingwavelabs/risingwave/issues/23177

Because our changelog previously used the logic of rowid to generate vnodes in a loop, which caused errors. In this PR, the problem is avoided by using the vnode of the table.

This pull request introduces significant improvements to how changelog row IDs are generated and handled throughout the system, with a focus on per-vnode sequencing, distribution awareness, and better timestamp management. The changes affect the row ID generation logic, the optimizer and planner, and the stream executor, ensuring that changelog row IDs are unique per vnode and properly distributed. Below are the most important changes grouped by theme:

**Row ID Generation Enhancements:**

* Added a new `ChangelogRowIdGenerator` struct in `row_id.rs` that generates unique changelog row IDs with per-vnode sequence management and improved timestamp handling via a shared `TimestampManager`. This ensures uniqueness and correct ordering for distributed changelog processing.
* Refactored `RowIdGenerator` to use the new `TimestampManager` for consistent timestamp management, replacing direct timestamp logic and simplifying timestamp updates. [[1]](diffhunk://#diff-aa12746e57cc3e613914f671271b49fec8787915634528915b6ecacbebc0024cR29-R91) [[2]](diffhunk://#diff-aa12746e57cc3e613914f671271b49fec8787915634528915b6ecacbebc0024cL37-R103) [[3]](diffhunk://#diff-aa12746e57cc3e613914f671271b49fec8787915634528915b6ecacbebc0024cL106-R171) [[4]](diffhunk://#diff-aa12746e57cc3e613914f671271b49fec8787915634528915b6ecacbebc0024cL130-L166) [[5]](diffhunk://#diff-aa12746e57cc3e613914f671271b49fec8787915634528915b6ecacbebc0024cL188-R214)

**Stream Plan and Executor Integration:**

* Updated the `ChangeLogExecutor` to use the new `ChangelogRowIdGenerator`, passing in the vnode bitmap and distribution keys, and generating row IDs per vnode for each chunk. Also handles vnode bitmap updates on barriers for dynamic partitioning. [[1]](diffhunk://#diff-5e745b79ecc25919f164f8c8fdc751607b1e0bd76e0d84b0d5321a37bdb1e1fdR22-R35) [[2]](diffhunk://#diff-5e745b79ecc25919f164f8c8fdc751607b1e0bd76e0d84b0d5321a37bdb1e1fdL43-R89) [[3]](diffhunk://#diff-5e745b79ecc25919f164f8c8fdc751607b1e0bd76e0d84b0d5321a37bdb1e1fdL79-R111)
* Modified the optimizer's changelog planning logic to propagate distribution keys and ensure that changelog nodes are properly distributed, including adding a `StreamExchange` when necessary to enforce distribution. [[1]](diffhunk://#diff-467401133e596e0b3c76082a8694fc9c8b93e1511eca6a6c7b9c3f9f01bd5560L144-R165) [[2]](diffhunk://#diff-467401133e596e0b3c76082a8694fc9c8b93e1511eca6a6c7b9c3f9f01bd5560L163-R177)

**Protobuf and Plan Node Changes:**

* Extended the `ChangeLogNode` protobuf definition to include a repeated `distribution_keys` field, and updated `StreamChangeLog` to carry and serialize these keys, ensuring distribution metadata is correctly propagated in the stream plan. [[1]](diffhunk://#diff-548bcb845b3db0225b2c6ab9d843234320a813af3a44855b764aaa284eda2dd8R375) [[2]](diffhunk://#diff-b405bf6fdf096e5439e927cbc7cf75c6e8a05e4353fd57f9975bd4fcbdbbe297L24-L37) [[3]](diffhunk://#diff-b405bf6fdf096e5439e927cbc7cf75c6e8a05e4353fd57f9975bd4fcbdbbe297L47-R54) [[4]](diffhunk://#diff-b405bf6fdf096e5439e927cbc7cf75c6e8a05e4353fd57f9975bd4fcbdbbe297L59-R70) [[5]](diffhunk://#diff-b405bf6fdf096e5439e927cbc7cf75c6e8a05e4353fd57f9975bd4fcbdbbe297R81)

These changes collectively improve the reliability, scalability, and correctness of changelog row ID generation and distribution in the streaming system.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/e2e_test/streaming/changelog.slt
+++ b/e2e_test/streaming/changelog.slt
@@ -24,13 +24,13 @@ statement ok
 create sink odd into wide (v1, odd) as select v1, v1*2+1 from t4;
 
 statement ok
-create materialized view mv9 as with p as changelog from wide select _changelog_row_id::bigint as changelog_row_id, * from p;
+create materialized view mv9 as with p as changelog from wide select * from p;
 
 statement ok
 insert into t4 select generate_series(1020300, 1020399);
 
 query III rowsort
-with add as (select * from mv9 where changelog_op = 1), update as (select * from mv9 where changelog_op = 3) select add.v1 from add, update where add.v1 = update.v1 and add.changelog_row_id > update.changelog_row_id;
+with add as (select _changelog_row_id,* from mv9 where changelog_op = 1), update as (select _changelog_row_id,* from mv9 where changelog_op = 3) select add.v1 from add, update where add.v1 = update.v1 and add._changelog_row_id > update._changelog_row_id;
 ----
 
 # test public.t1 due to https://github.com/risingwavelabs/risingwave/issues/18747


src/stream/src/from_proto/changelog.rs
```diff
@@ -32,7 +32,17 @@ impl ExecutorBuilder for ChangeLogExecutorBuilder {
     ) -> StreamResult<Executor> {
         let [input]: [_; 1] = params.input.try_into().unwrap();
 
-        let exec = ChangeLogExecutor::new(params.actor_context, input, node.need_op);
+        let vnodes = params
+            .vnode_bitmap
+            .expect("vnodes not set for row id gen executor");
```
hzxa21: nits: the msg mentions row id gen executor instead of changelog

src/stream/src/executor/changelog.rs
```diff
@@ -40,27 +47,48 @@ impl Execute for ChangeLogExecutor {
...
-    pub fn new(ctx: ActorContextRef, input: Executor, need_op: bool) -> Self {
+    pub fn new(
+        ctx: ActorContextRef,
+        input: Executor,
+        need_op: bool,
+        all_vnode_count: usize,
+        vnodes: Bitmap,
+    ) -> Self {
+        let changelog_row_id_generator =
+            ChangelogRowIdGenerator::new(vnodes.iter_vnodes(), vnodes.len());
```
hzxa21: We need to handle scaling as well. When scaling happens, there will be a vnode bitmap changes propagated by barrier mutation.

src/common/src/util/row_id.rs
```diff
@@ -231,6 +232,132 @@ impl RowIdGenerator {
...
+    }
+
+    fn try_update_timestamp(&mut self) {
+        let get_current_timestamp_ms = || self.base.elapsed().unwrap().as_millis() as i64;
+
+        let current_timestamp_ms = get_current_timestamp_ms();
+        let sequence_upper_bound = self.sequence_upper_bound();
+
+        // Check if any vnode has reached the upper bound
+        let any_vnode_at_limit = self
```
hzxa21: For efficiency consideration, I think it is better to avoid iterating `vnodes_sequence` in the critial path here and in L321. We can
1. store a flag `max_sequence_reached` in `ChangelogRowIdGenerator`. Set it to true in `next_changelog_row_id_in_current_timestamp` when the limit is reached and check it here, reset after the timestamp bumps.
2. Init `vnodes_sequence` to a empty hashmap after timestamp update. Insert 0 for the vnode if the vnode is not found.

src/common/src/util/row_id.rs
```diff
@@ -231,6 +232,132 @@ impl RowIdGenerator {
...
+        self.vnodes_sequence.insert(*vnode, current_sequence + 1);
+
+        Some(
+            self.last_timestamp_ms << TIMESTAMP_SHIFT_BITS
+                | (vnode.to_index() << (TIMESTAMP_SHIFT_BITS - self.vnode_bit)) as i64
+                | sequence as i64,
+        )
+    }
+
+    #[allow(clippy::should_implement_trait)]
```
hzxa21: This is unneccessary?

BugenZhao:
Can you fill the PR body first?

src/common/src/util/row_id.rs
```diff
@@ -231,6 +233,117 @@ impl RowIdGenerator {
     }
 }
 
+/// `ChangelogRowIdGenerator` generates unique changelog row ids using snowflake algorithm.
+/// Unlike `RowIdGenerator`, it maintains a separate sequence for each vnode and generates
+/// row ids based on the input vnode.
+#[derive(Debug)]
+pub struct ChangelogRowIdGenerator {
```
BugenZhao, line 264: Is it possible to reuse some code between this and the original one? For example, I believe the majority of `ChangelogRowIdGenerator` is duplicated.

src/frontend/src/planner/changelog.rs
```diff
@@ -11,15 +11,22 @@
...
 use crate::error::Result;
 use crate::optimizer::plan_node::{LogicalChangeLog, LogicalPlanRef as PlanRef};
 
 impl Planner {
     pub(super) fn plan_changelog(&mut self, relation: Relation) -> Result<PlanRef> {
+        let vnode_count = if let Relation::BaseTable(base_table) = &relation {
+            base_table.table_catalog.vnode_count()
+        } else {
+            bail_not_implemented!("changelog only support base table");
+        };
```
BugenZhao: I think there's no need to add the `vnode_count` field, because...

- during execution, it can actually be obtained from `actor_context.vnode_count`
- this requires the changelog to be attched on a base table, but theoretically it can be a table function atop of any stream, we'd better preserve this possibility for future expansion

src/stream/src/executor/changelog.rs
```diff
@@ -40,27 +47,47 @@ impl Execute for ChangeLogExecutor {
...
+                        data_chunk,
+                        &self.pk_indices,
+                        self.all_vnode_count,
+                    );
                     let (ops, mut columns, bitmap) = chunk.into_inner();
                     let new_ops = vec![Op::Insert; ops.len()];
+                    let changelog_row_ids = vnodes
+                        .iter()
+                        .map(|vnode| self.changelog_row_id_generator.next(vnode));
                     // They are all 0, will be add in row id gen executor.
```
BugenZhao: Seems outdated

src/stream/src/executor/changelog.rs
```diff
@@ -40,27 +47,47 @@ impl Execute for ChangeLogExecutor {
...
         let input = self.input.execute();
         #[for_await]
         for msg in input {
             let msg = msg?;
             match msg {
                 Message::Chunk(chunk) => {
+                    let data_chunk = chunk.data_chunk();
+                    let vnodes = VirtualNode::compute_chunk(
+                        data_chunk,
+                        &self.pk_indices,
```
BugenZhao: This is very wrong. Vnode should be calculated based on the distribution key instead of stream key.

src/frontend/src/optimizer/plan_node/logical_changelog.rs
```diff
@@ -145,13 +160,6 @@ impl ToStream for LogicalChangeLog {
 
         let core = self.core.clone_with_input(new_input);
         let plan = StreamChangeLog::new(core).into();
-        let row_id_index = self.schema().fields().len() - 1;
-        let plan = StreamRowIdGen::new_with_dist(
-            plan,
-            row_id_index,
-            Distribution::HashShard(vec![row_id_index]),
-        )
-        .into();
```
BugenZhao, line 154: By removing this, it seems that the distribution key will be the same as the input. Is it your desired behavior?

However, the `stream_key` is not changed accordingly. The problem is that we need to ensure the distribution key is a subset of the stream key.

https://github.com/risingwavelabs/risingwave/blob/4136a60e6905a77a5834550375180be96404f2d3/src/frontend/src/optimizer/plan_node/generic/changelog.rs#L83-L90

BugenZhao (changes requested):

src/frontend/src/planner/changelog.rs
xxhZs: The vnode count here is not for the actor, but rather the maximum value of the overall count. We need this value to calculate the vnode corresponding to the pk.

BugenZhao: `actor_context.vnode_count` is exactly the total count of the fragment.

src/frontend/src/optimizer/plan_node/logical_changelog.rs
```diff
@@ -141,15 +140,29 @@ impl ToBatch for LogicalChangeLog {
...
+        let (distribution_keys, new_input) = match dist {
+            Distribution::HashShard(distribution_keys)
+            | Distribution::UpstreamHashShard(distribution_keys, _) => {
+                (distribution_keys.clone(), input)
+            }
+            _ => {
+                let stream_key = input
+                    .stream_key()
+                    .ok_or_else(|| anyhow::anyhow!("changelog input must have a stream key"))?
+                    .to_vec();
```
BugenZhao: I guess we can assume the stream key must exist for stream plan node? Then we can call `expect_stream_key`. cc @chenzl25

src/frontend/src/optimizer/plan_node/logical_changelog.rs
```diff
@@ -141,15 +140,29 @@ impl ToBatch for LogicalChangeLog {
...
     fn to_stream(&self, ctx: &mut ToStreamContext) -> Result<StreamPlanRef> {
-        let new_input = self.input().to_stream(ctx)?;
-
+        let input = self.input().to_stream(ctx)?;
+        let dist = input.distribution();
+        let (distribution_keys, new_input) = match dist {
+            Distribution::HashShard(distribution_keys)
+            | Distribution::UpstreamHashShard(distribution_keys, _) => {
+                (distribution_keys.clone(), input)
+            }
```
BugenZhao: Can call `dist_column_indices`.

src/frontend/src/optimizer/plan_node/stream_changelog.rs
```diff
@@ -21,20 +21,23 @@ use super::stream::prelude::PhysicalPlanRef;
...
     core: generic::ChangeLog<PlanRef>,
+    distribution_keys: Vec<u32>,
 }
 
 impl StreamChangeLog {
-    pub fn new(core: generic::ChangeLog<PlanRef>) -> Self {
+    pub fn new_with_dist(
+        core: generic::ChangeLog<PlanRef>,
+        dist: Distribution,
+        distribution_keys: Vec<u32>,
```
BugenZhao, line 38: I think we can obtain this from `core.input.distribution().dist_column_indices()`

BugenZhao (commented):
Surprised to see there are no planner tests for this feature. Could you add some in this PR?

Also, can you add the test case mentioned in https://github.com/risingwavelabs/risingwave/issues/23177 as an e2e test?

src/frontend/src/optimizer/plan_node/stream_changelog.rs
xxhZs: According to the comments above, the stream_key of our changelog operator should be changlog_row_id. Therefore, we fix the dist_key here as changlog_row_id instead of using the dist_key from the upstream.

src/frontend/src/optimizer/plan_node/logical_changelog.rs
xxhZs: It seems this method will panic, and it's not certain if there are any special cases here that could cause issues. This should be the same problem as the one above, cc @chenzl25

src/frontend/src/optimizer/plan_node/stream_changelog.rs
BugenZhao: Got it.

I just realized this means that we changed the distribution key within a fragment without introducing an Exchange. Not sure if this will break any assumption... image

src/frontend/src/optimizer/plan_node/logical_changelog.rs
BugenZhao: Oh, I guess you can call `enforce_concrete_distribution` then `dist_column_indices`.

chenzl25:
I will take a look later today.

BugenZhao:
Can you also add some planner test for the new plan? `stream_plan` and `stream_dist_plan` will be nice.

e2e_test/streaming/changelog.slt
```diff
@@ -10,6 +10,29 @@ create table t2 (v1 int, v2 int);
...
+statement ok
+create table wide (v1 int primary key, even int, odd int);
+
+statement ok
+create sink even into wide (v1, even) as select v1, v1*2 from t4;
+
+statement ok
+create sink odd into wide (v1, odd) as select v1, v1*2+1 from t4;
+
+statement ok
```
chenzl25, line 26: columns mismatch

src/frontend/src/optimizer/plan_node/logical_changelog.rs
chenzl25: I think the input of changelog is simply a table scan, so the distribution of scan could be UpstreamHashShard or `Single`. But it seems we don't handle the `Single` distribution here. For example the stream key of Single is empty, but we use it in HashShard.

src/frontend/src/optimizer/plan_node/logical_changelog.rs
```diff
@@ -141,15 +139,21 @@ impl ToBatch for LogicalChangeLog {
...
-        let plan = StreamChangeLog::new(core).into();
+        let input = self.input().to_stream(ctx)?;
+        let dist = input.distribution();
+        let distribution_keys = match dist {
+            Distribution::HashShard(distribution_keys)
+            | Distribution::UpstreamHashShard(distribution_keys, _) => distribution_keys.clone(),
+            _ => {
+                vec![]
+            }
+        };
```
chenzl25, line 157: Please enumerate all the distribution type explicitly and I think `Broadcast` and `SomeShard` should never appear in the input.

src/frontend/src/optimizer/plan_node/logical_changelog.rs
BugenZhao: > I think the input of changelog is simply a table scan, so the distribution of scan could be UpstreamHashShard or Single

True. But theoretically the upstream can be any plan node, so I'd suggest using a more general implementation here. 🤔

hzxa21:
There is a compatibility issue in this PR. I created a MV with `AS CHANGELOG` in commit 5e9eb6baa40b3df5dcbc580c96254b08ceb65370 (one commit before this PR)
```
dev=> create table t(k int primary key, v int);
dev=> create materialized view mv2 as with sub as CHANGELOG FROM t select k,v,changelog_op, _changelog_row_id::bigint AS __row_id from sub;
dev=> insert into t values(1, 3);
INSERT 0 1
dev=> insert into t values(3, 3);
INSERT 0 1

```

Then I upgrade to this PR ad6e6ab4960c7e7c100e215cd35168053e5cbcc4 and insert more rows
```
dev=> insert into t values(5, 3);
INSERT 0 1
```

CN panics:
```
thread 'rw-streaming' panicked at src/common/src/util/row_id.rs:306:13:
vnode VirtualNode(0) not in generator
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/28f1c807911c63f08d98e7b468cfcf15a441e34b/library/std/src/panicking.rs:697:5
   1: core::panicking::panic_fmt
             at /rustc/28f1c807911c63f08d98e7b468cfcf15a441e34b/library/core/src/panicking.rs:75:14
   2: risingwave_common::util::row_id::ChangelogRowIdGenerator::next_changelog_row_id_in_current_timestamp
             at ./src/common/src/util/row_id.rs:306:13
   3: risingwave_common::util::row_id::ChangelogRowIdGenerator::next
             at ./src/common/src/util/row_id.rs:325:36
   4: risingwave_stream::executor::changelog::ChangeLogExecutor::execute_inner::{{closure}}::{{closure}}
             at ./src/stream/src/executor/changelog.rs:87:70
   5: core::iter::adapters::map::map_fold::{{closure}}
             at /home/patrick/.rustup/toolchains/nightly-2025-06-25-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/iter/adapters/map.rs:88:28
   6: <core::slice::iter::Iter<T> as core::iter::traits::iterator::Iterator>::fold
             at /home/patrick/.rustup/toolchains/nightly-2025-06-25-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/slice/iter/macros.rs:255:27
   7: <core::iter::adapters::map::Map<I,F> as core::iter::traits::iterator::Iterator>::fold
             at /home/patrick/.rustup/toolchains/nightly-2025-06-25-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/iter/adapters/map.rs:128:19
   8: <core::iter::adapters::map::Map<I,F> as core::iter::traits::iterator::Iterator>::fold
```

BugenZhao:
> vnode VirtualNode(0) not in generator

True. This is really an issue. Because we are adding a new field into the plan node proto.

PR #23244, opened 2025-09-18
fix(mysql-cdc): avoid NULL values for overflowing BIGINT UNSIGNED during backfill[another way]

wcy-fdu:

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/src/parser/mysql.rs
+++ b/src/connector/src/parser/mysql.rs
@@ -15,6 +15,7 @@
 use std::sync::LazyLock;
 
 use mysql_async::Row as MysqlRow;
+use mysql_common::constants::ColumnFlags;
 use risingwave_common::catalog::Schema;
 use risingwave_common::log::LogSuppresser;
 use risingwave_common::row::OwnedRow;
@@ -58,6 +59,42 @@ macro_rules! handle_data_type {
                     $i,
                     stringify!($rw_type),
                 ))),
+        }
+    }};
+}
+
+macro_rules! handle_data_type_with_signed {
+    (
+        $mysql_row:expr,
+        $mysql_datum_index:expr,
+        $column_name:expr,
+        $signed_type:ty,
+        $unsigned_type:ty
+    ) => {{
+        let column_flags = $mysql_row.columns()[$mysql_datum_index].flags();
+
+        if column_flags.contains(ColumnFlags::UNSIGNED_FLAG) {
+            // UNSIGNED type: use unsigned type conversion, then convert to signed
+            match $mysql_row.take_opt::<Option<$unsigned_type>, _>($mysql_datum_index) {
+                Some(Ok(Some(val))) => Ok(Some(ScalarImpl::from(val as $signed_type))),
+                Some(Ok(None)) => Ok(None),
+                Some(Err(e)) => Err(anyhow::Error::new(e.clone())
+                    .context("failed to deserialize MySQL value into rust value")
+                    .context(format!(
+                        "column: {}, index: {}, rust_type: {}",
+                        $column_name,
+                        $mysql_datum_index,
+                        stringify!($unsigned_type),
+                    ))),
+                None => bail!(
+                    "no value found at column: {}, index: {}",
+                    $column_name,
+                    $mysql_datum_index
+                ),
+            }
+        } else {
+            // SIGNED type: use default signed type conversion
+            handle_data_type!($mysql_row, $mysql_datum_index, $column_name, $signed_type)
         }
     }};
 }
@@ -107,13 +144,13 @@ pub fn mysql_datum_to_rw_datum(
             }
         }
         DataType::Int16 => {
-            handle_data_type!(mysql_row, mysql_datum_index, column_name, i16)
+            handle_data_type_with_signed!(mysql_row, mysql_datum_index, column_name, i16, u16)
         }
         DataType::Int32 => {
-            handle_data_type!(mysql_row, mysql_datum_index, column_name, i32)
+            handle_data_type_with_signed!(mysql_row, mysql_datum_index, column_name, i32, u32)
         }
         DataType::Int64 => {
-            handle_data_type!(mysql_row, mysql_datum_index, column_name, i64)
+            handle_data_type_with_signed!(mysql_row, mysql_datum_index, column_name, i64, u64)
         }
         DataType::Float32 => {
             handle_data_type!(mysql_row, mysql_datum_index, column_name, f32)
--- a/src/connector/src/source/cdc/external/mysql.rs
+++ b/src/connector/src/source/cdc/external/mysql.rs
@@ -534,10 +534,21 @@ impl MySqlExternalTableReader {
 
     /// Quote a column with potential CAST for unsigned types
     fn quote_column_with_cast(&self, column_name: &str) -> String {
+        // 为 unsigned 列添加 CAST 以处理 BIGINT UNSIGNED 溢出问题
         if self.is_unsigned_column(column_name) {
             format!("CAST({} AS UNSIGNED)", Self::quote_column(column_name))
         } else {
             Self::quote_column(column_name)
+        }
+    }
+
+    /// Quote a parameter with potential CAST for unsigned types
+    fn quote_param_with_cast(&self, column_name: &str) -> String {
+        // 为 unsigned 列添加 CAST 以处理 BIGINT UNSIGNED 溢出问题
+        if self.is_unsigned_column(column_name) {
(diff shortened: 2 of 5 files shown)


wcy-fdu:
Had some offline discussion with @hzxa21  and @xiangjinwu , we will ban casts like `u64->i64` on the frontend, so that newly created tables can only be cast up. The backend processing logic remains unchanged.

src/connector/src/parser/mysql.rs
```diff
@@ -106,14 +143,13 @@ pub fn mysql_datum_to_rw_datum(
             }
         }
         DataType::Int16 => {
-            handle_data_type!(mysql_row, mysql_datum_index, column_name, i16)
+            handle_data_type_with_signed!(mysql_row, mysql_datum_index, column_name, i16, u16)
```
xiangjinwu, line 148: Answering my own question here https://github.com/risingwavelabs/risingwave/pull/23278#discussion_r2371770686
> How does upcast work here?

Before this PR, when a MySQL u16 is declared as RisingWave i32, we would get it as `i32`.

After this PR, when a MySQL u16 is declared as RisingWave i32, we would get it as `u32` and then transform into `i32`.

These relies on the fact that `mysql_async` allows getting MySQL u16 as `i32` and `u32`

src/connector/src/source/cdc/external/mysql.rs
```diff
@@ -493,6 +501,64 @@ impl MySqlExternalTableReader {
...
+
+        drop(conn);
+
+        Ok(column_infos)
+    }
+
+    /// Check if a column is unsigned type
+    fn is_unsigned_type(&self, column_name: &str) -> bool {
+        for (col_name, col_type) in &self.upstream_mysql_pk_infos {
+            if col_name == column_name {
```
xiangjinwu: FYI we can also `self.upstream_mysql_pk_infos.iter().find`

src/connector/src/source/cdc/external/mysql.rs
```diff
@@ -543,9 +609,30 @@ impl MySqlExternalTableReader {
                         let ty = field_map.get(pk.as_str()).unwrap();
                         let val = match ty {
                             DataType::Boolean => Value::from(value.into_bool()),
-                            DataType::Int16 => Value::from(value.into_int16()),
-                            DataType::Int32 => Value::from(value.into_int32()),
-                            DataType::Int64 => Value::from(value.into_int64()),
+                            DataType::Int16 => {
+                                let int16_val = value.into_int16();
+                                if int16_val < 0 && self.is_unsigned_type(pk.as_str()) {
```
xiangjinwu, line 613: What if a later refactoring removes `int16_val < 0`? Will existing test cases catch the error, or it is safe to remove?

src/connector/src/source/cdc/external/mysql.rs
```diff
@@ -693,6 +780,8 @@ mod tests {
 
     #[test]
     fn test_mysql_filter_expr() {
+        // This test is commented out because filter_expression now requires &self
+        // and we need a proper MySqlExternalTableReader instance to test it
```
xiangjinwu: I do not get this. How is it commented out?

src/connector/src/parser/mysql.rs
```diff
@@ -62,6 +63,42 @@ macro_rules! handle_data_type {
...
+        $column_name:expr,
+        $signed_type:ty,
+        $unsigned_type:ty
+    ) => {{
+        let column_flags = $mysql_row.columns()[$mysql_datum_index].flags();
+
+        if column_flags.contains(ColumnFlags::UNSIGNED_FLAG) {
+            // UNSIGNED type: use unsigned type conversion, then convert to signed
+            match $mysql_row.take_opt::<Option<$unsigned_type>, _>($mysql_datum_index) {
+                Some(Ok(Some(val))) => Ok(Some(ScalarImpl::from(val as $signed_type))),
```
xiangjinwu, line 81: Add a comment here to acknowledge we are intentionally converting `18446251075179777772u64` into `-492998529773844i64`

e2e_test/source_inline/cdc/mysql/test_unsigned_int.slt
```diff
@@ -124,42 +146,74 @@ sleep 3s
...
-select id, unsigned_bigint, unsigned_int, unsigned_smallint, unsigned_tinyint, info from rw_unsigned_int where id = 53;
+select id, unsigned_bigint, unsigned_int, unsigned_smallint, unsigned_tinyint, info from rw_unsigned_int_upcast where id = 53;
 ----
 53	18446251075179777772	4294967295	65535	255	debezium
 
+# Check id=53 for unsigned bigint -> bigint table
+query TTTTTT
+select id, unsigned_bigint, info from rw_unsigned_int_to_signed where id = 53;
+----
+53	-492998529773844	debezium
```
xiangjinwu, line 202: What about other columns? Are they converted to -1 for both backfill and debezium case?

wcy-fdu: Here, I actually feel conflicted. I tend to prefer that this PR only targets unsigned bigint. The reason is that this PR is a somewhat dirty fix—a workaround that had to be implemented due to "certain special reasons from users." The correct approach here should be to guide users to perform an upcast.

Back to your question
- For unsigned bigint, if an overflow occurs, Debezium will handle it internally, sending a value that has already been converted to a negative number (for example, if the upstream value is 18446251075179777770, Debezium will convert it and send -492998529773846).
- However, for other types like unsigned int or unsigned smallint, Debezium will directly send original value(eg. unsigned int 4294967295, which is i32 overflowed), and we will need to handle this situation ourselves.

>Are they converted to -1 

To convert such overflow value, we need to specially handling [here](https://github.com/risingwavelabs/risingwave/blob/d052368d0f54631b5649dce648b70ac323634954/src/connector/src/parser/unified/json.rs#L289-L351) like 
```
// ---- Int32 -----
            (
                DataType::Int32,
                ValueType::I64 | ValueType::I128 | ValueType::U64 | ValueType::U128,
            ) => {
                // Try signed first, if that fails, try unsigned and convert
                // This handles MySQL unsigned int that overflow i32::MAX
                value.try_as_i32()
                    .or_else(|_| {
                        // Try to get as u32 and convert to i32
                        value.as_u64()
                            .and_then(|v| u32::try_from(v).ok())
                            .map(|v| v as i32)
                            .ok_or(())
                    })
                    .map_err(|_| create_error())?
                    .into()
            }
```

This is where all the JSON processing is done; I find it a bit odd to add this logic here.

In summary, I haven't yet handled other unsigned int, unsigned smallint, etc., meaning they won't be converted to -1, but rather to null. This is feasible, but I don't want to go further down a dirty way.

wcy-fdu: What's your opinion? @xiangjinwu

wcy-fdu: The value will be -1 during the backfill phase and will become null during the debezium phase. This null value should be used to remind the user to perform the correct upcast.

xiangjinwu: I checked the debezium schema with an example, and yes the inconsistency originates from debezium:
```
{"type":"int32","optional":true,"field":"u16"},
{"type":"int64","optional":true,"field":"u32"},
{"type":"int64","optional":true,"field":"u64"}
```

What about only doing the overflow-as-negative for u64, and require upcast for smaller integrals?

wcy-fdu: >What about only doing the overflow-as-negative for u64, and require upcast for smaller integrals?


This is what I want to do, but requiring users to upcast seems a bit difficult. Therefore, this PR only slightly opened a way for `unsigned bigint` to be cast to signed.

wcy-fdu: The best approach would be to check the schema during table creation, and 
- if an `unsigned int->int` map is detected, issue an error and alert the user; 
- if an` unsigned bigint->bigint` map is detected, issue a warning and allow it. However, that might not be easy to implement. Leave it for the later PR.

src/connector/src/source/cdc/external/mysql.rs
wcy-fdu: It's safe to remove those `int16_val < 0` judgement, because the subsequent check of `self.is_unsigned_type` will be successful. This is added here to reduce overhead caused by checking `is_unsigned_type`, based on the principle of prioritizing the left-hand side.

PR #23245, opened 2025-09-18
feat(kinesis): add more metrics for Kinesis source

tabVersion:

## Checklist

## Documentation

<details>

</details>


src/connector/src/source/kinesis/source/reader.rs
```diff
@@ -149,13 +160,19 @@ impl KinesisSplitReader {
...
+                        .with_guarded_label_values(&self.metrics_labels)
+                        .observe(resp.millis_behind_latest().unwrap_or(0) as f64);
+
                     self.shard_iter = resp.next_shard_iterator().map(String::from);
                     let chunk = (resp.records().iter())
                         .map(|r| from_kinesis_record(r, self.split_id.clone()))
                         .collect::<Vec<SourceMessage>>();
                     if let Some(shard) = &resp.child_shards
                         && !shard.is_empty()
+                        && self.shard_iter.is_none()
```
tabVersion, line 175: here we narrow down the condition to early terminate a shard.

tabVersion: reference https://docs.aws.amazon.com/streams/latest/dev/developing-consumers-with-sdk.html

> Your application has already read all the data from one of these shards. This is the shard for which getRecordsResult.getNextShardIterator returned null. If the order of the data is important to your application, ensure that it also reads all the data from the other parent shard before reading any new data from the child shard created by the merge.

tabVersion: may fix too early terminate impl from https://github.com/risingwavelabs/risingwave/pull/18183/

hzxa21 (approved):

PR #23258, opened 2025-09-19
feat: metric indicating barrier_interval_ms

tabVersion:

sometimes we got confusing metrics by user changed `barrier_interval_ms`, which can only accessible from sql params. 

## Checklist

## Documentation

<details>

</details>


src/meta/src/rpc/metrics.rs
```diff
@@ -76,6 +76,8 @@ pub struct MetaMetrics {
     pub in_flight_barrier_nums: LabelGuardedIntGaugeVec,
     /// The timestamp (UNIX epoch seconds) of the last committed barrier's epoch time.
     pub last_committed_barrier_time: IntGaugeVec,
+    /// The barrier interval of each database
+    pub barrier_interval_by_database: HistogramVec,
```
tabVersion: here we not use GuardedHistogramVec because I want to keep the data point as the cluster is running and the interval remains not changed

Chiro11:
I think metrics should also be modified here.
https://github.com/risingwavelabs/risingwave/blob/892026ae33d8e50e94396ca4cdc4107943840b4c/src/meta/src/barrier/schedule.rs#L414-L426

My suggestion is, considering that each time create or modify `barrier_interval_ms`, will use this function, we can add a `database_id` parameter to this function, and then put the modification of metrics inside.
https://github.com/risingwavelabs/risingwave/blob/892026ae33d8e50e94396ca4cdc4107943840b4c/src/meta/src/barrier/schedule.rs#L407-L411

grafana/dashboard/dev/streaming.py
```diff
@@ -316,6 +316,16 @@ def _(outer_panels: Panels):
                         ),
                     ],
                 ),
+                panels.timeseries_latency(
+                    "Barrier Interval",
+                    "Barrier interval of each database",
+                    [
+                        panels.target(
+                            f"rate({metric('meta_barrier_interval_by_database_sum')}[$__rate_interval]) / rate({metric('meta_barrier_interval_by_database_count')}[$__rate_interval]) > 0",
```
kwannoel: Why do we use rate?

tabVersion: following the practice for time-metrics (latency, etc.)
the metric `meta_barrier_interval_by_database` will be derived to `meta_barrier_interval_by_database_sum`, `meta_barrier_interval_by_database_count` and `meta_barrier_interval_by_database_bucket` automatically, but lacking `meta_barrier_interval_by_database` itself. 

for reference

https://github.com/risingwavelabs/risingwave/blob/892026ae33d8e50e94396ca4cdc4107943840b4c/grafana/dashboard/dev/streaming.py#L288-L291

kwannoel (commented):
Rather than report it on each scheduled barrier, why don't we report it at the following events where it changes:
1. On startup.
2. On system configuration change.
3. On database configuration change.

We can maintain separate metrics for system and database metrics.

grafana/dashboard/dev/streaming.py
kwannoel: Is the objective of this PR to observe the exact barrier interval each database at a specific point in time? If so using `rate of change` doesn't seem precise to me. Could you provide a screenshot of how these metrics look on the grafana dashboard to see if it's as expected?

tabVersion:
> Rather than report it on each scheduled barrier, why don't we report it at the following events where it changes:
> 
> 1. On startup.
> 2. On system configuration change.
> 3. On database configuration change.
> 
> We can maintain separate metrics for system and database metrics.

Yes, that's exactly @Chiro11 's suggestion.

tabVersion:
image

metrics preview. 

~now WIP for adding panel to the dashboard~

grafana/dashboard/dev/streaming.py
tabVersion: Now changed to `Gauge` rather than `histogram`.  Now we can observe an accurate time interval as f64.

Issue #23263, opened 2025-09-19
Optimize append only group topN


hzxa21:
Copied from slack discussion:
Assuming we have a append-only table with two column a, b , and I would like to calculate group top 1 b (group by a order by b ), which of the following streaming SQL is more efficient:
1. Use MV: DISTINCT ON (a) .... ORDER BY a, b DESC . I assume this will be optimized to be append only group top 1, right?
2. Use SINK INTO TBABLE with version column: CREATE TABLE (a int primary key, b int) ON CONFLICT OVERWRITE WITH VERSION COLUMN(b).

At the first glance, we thought there is no huge difference between the two. However, we just realized that option 1 is way actually more inefficient than option 2 because:
- option 1 persists all input data in the append-only group topn state table while option 2 only persists one row per PK. (I thought append only group top N only needs to persist N rows per group key but after checking I realized that I am wrong...
- with more states in option 1, the state I/O and operator/storage caching can be much more inefficient because there are more states to be cached and you to skip a lot of unused data for each group key.

Random thoughts:
- Can we convert append only group top1 to sink into table with version column?
- For append only group top N in general, is it possible to only persist N rows at max per group key?

hzxa21:
> (I thought append only group top N only needs to persist N rows per group key but after checking I realized that I am wrong...

Correction: there is deletion in the state table here:
https://github.com/risingwavelabs/risingwave/blob/dce92415d8636d7e9188b2cf58ce99943c26c447/src/stream/src/executor/top_n/top_n_cache.rs#L720

This means append only group top N only maintain N entries in the state store. However, even though the high part of the topN cache [is not used](https://github.com/risingwavelabs/risingwave/blob/545b21e0ca1d458d060f58c63b9203c5080eb248/src/stream/src/executor/top_n/top_n_cache.rs#L726), we still drain the storage iterator to populate it.

This will cause the storage iterator to read through all the unreclaim tombstones and amplifying the data to scan in the storage.

PR #23264, opened 2025-09-19
feat: avoid populating the high part of append-only group topn cache

hzxa21:

See https://github.com/risingwavelabs/risingwave/issues/23263#issuecomment-3311630206 for details.

The high part of the append only group top N cache should always be empty so calling storage iter next is wasteful.

## Checklist

## Documentation

<details>

</details>


hzxa21:
Nexmark Q9/Q18 both uses append only group top n. Though based on the nexrmark perf test grafana, there is no tombstones in the group top n state table because of the workload characteristic, this PR shouldn't cause regression. The perf result proves it.

Q9:
image

Q18:
image

PR #23273, opened 2025-09-22
refactor(executor): Decouple barrier alignment and mutation for merge type executor

### LINKED ISSUE #23064: Unable to add column to table that has both upstream sink and downstream dependency

Chiro11:

Based on https://github.com/risingwavelabs/risingwave/pull/23221#issuecomment-3306657966

Decouple barrier mutation from upstream modifications, to eliminate potential dependencies.

The original implementation may cause deadlocks in the future:
- For the new created table, we must create new executors, then create its corresponding dispatcher for downstream, then we can do `execute()`.
- Creating the new dispatcher requires waiting for the downstream job to register the new output.
- The `Merge` operator in downstream job must receive a barrier from the old table's fragment-graph before it can obtain the mutation, create a new upstream, and register new output.
- This means that the old job must process the barrier and send it to downstream normally before the new job is created.
- This means that the upstream job's dispatcher must also be able to send the barrier normally.
- The upstream job's dispatcher, when receiving the barrier, must wait for the new downstream table to register its output before sending the barrier to both the old and new downstream tables.

So, this becomes:
```
create-new-job -> downstream-receive-barrier -> old-job-process-barrier -> upstream-dispatch-barrier -> create-new-job
```
which will cause deadlocks.

See comments in `DispatchBarrierBuffer` for details.

Close #23064

## Checklist

## Documentation

<details>

</details>


wenym1:
After this PR, do we still need https://github.com/risingwavelabs/risingwave/pull/23221?

Chiro11:
> After this PR, do we still need #23221?

Deadlock will be handled. But consider that `Merge` operator now also builds the channel with upstream when the actor is created (not when it is executed), the changes in `UpstreamSinkUnion` executor can also be retained, for consistency.

src/stream/src/executor/merge.rs
```diff
@@ -227,120 +236,226 @@ impl MergeExecutor {
         SelectReceivers::new(upstreams, None, merge_barrier_align_duration.clone())
     }
 
+    async fn partially_apply_barrier(
```
wenym1: I think the current implementation is too complicated, and introduced too many unnecessary code change in this PR. The new logic to decouple the two things should be similar for MergeExecutor and ReceiverExecutor, but now we have to implement it for twice for each of them.

All we need in this PR is to, while recv from upstream before upstream barrier, we concurrently poll new barrier from `barrier_rx` and then create new inputs if there are any. This `partially_apply_barrier` has included too many extra logic to make the method unnecessarily complicated.

We can have a helper struct to implement the logic. The struct will just repeat in a for-loop to poll barrier from barrier_rx, create new inputs if there is any, and then buffer the `(Barrier, Option<Vec<BoxedActorInput>>)` in a `VecDeque`. When we receive barrier from upstream, if the buffer is not empty, we take the barrier and optional new inputs directly, and if empty, we wait for the next barrier and optional new inputs. The rest logic to handle the barrier will be basically the same as original, except that the inputs have already been created instead of being created after we start handling the barrier.

Such struct can be reused for `MergeExecutor` and `ReceiverExecutor`, and then many changes in the two executors can be avoided. Similar logic can be applied for UpstreamSinkUnion. We don't have to involve the extra logic in `partially_apply_barrier`, but only receive barrier and create new inputs. Other handling logic can be untouched.

Chiro11: Fixed. But I still kept the changes to `UpstreamSinkUnion` because its `partially_apply_barrier` logic is quite complex, involving not only cache but also resending barriers and state modification.

src/stream/src/executor/mod.rs
```diff
@@ -1713,3 +1717,85 @@ impl<InputId: Clone + Ord + Hash + std::fmt::Debug, M> DynamicReceivers<InputId,
         self.blocked.is_empty() && self.active.is_empty()
     }
 }
+
+pub(crate) struct DispatchBarrierBuffer {
+    pub buffer: VecDeque<(Barrier, Option<Vec<BoxedActorInput>>)>,
+    pub curr_fragment_id: FragmentId,
```
wenym1: Should it be `upstream_fragment_id`?

src/stream/src/executor/merge.rs
```diff
@@ -229,6 +225,27 @@ impl MergeExecutor {
 
     #[try_stream(ok = Message, error = StreamExecutorError)]
     pub(crate) async fn execute_inner(mut self: Box<Self>) {
+        // Explanation of the implementation:
```
wenym1: Good comment!

Can move to the doc comment of `DispatchBarrierBuffer`.

src/stream/src/executor/mod.rs
```diff
@@ -1713,3 +1717,85 @@ impl<InputId: Clone + Ord + Hash + std::fmt::Debug, M> DynamicReceivers<InputId,
...
+
+                msg = stream.next() => {
+                    return msg
+                        .ok_or(StreamExecutorError::channel_closed(
+                            "upstream executor closed unexpectedly",
+                        ))
+                        .flatten();
+                }
+
+                barrier = pending_on_none(self.barrier_rx.recv()) => {
```
wenym1: `pre_apply_barrier` may block `stream.next()`. Can we possibly make them concurrent?

The concurrency should be like
```
select!(
    stream.next(), 
    async {
        loop {
            let barrier = self.barrier_rx.recv();
            self.pre_apply_barrier(barrier).await?
            ...
        }
    }
)
```

Note that we should handle cancellation safety properly. As we may be pending on `self.pre_apply_barrier(barrier).await`, if `stream.next()` receives a new message, we will return, and the future in the other select arm will be dropped. If we don't handle cancellation safety properly, we may lose the barrier that we just received but has not finished  `self.pre_apply_barrier(barrier)` yet.

Besides, should return error on None instead of being pending.

src/stream/src/executor/receiver.rs
```diff
@@ -112,31 +111,41 @@ impl Execute for ReceiverExecutor {
...
+                    DispatcherMessage::Barrier(barrier) => {
                         tracing::debug!(
                             target: "events::stream::barrier::path",
                             actor_id = actor_id,
                             "receiver receives barrier from path: {:?}",
                             barrier.passed_actors
                         );
-                        barrier.passed_actors.push(actor_id);
+                        let (mut recv_barrier, new_inputs) =
+                            barrier_buffer.pop_barrier_with_inputs().await?;
```
wenym1: Seems that we always do `pop_barrier_with_inputs` when we receive a barrier from upstream. Shall we pass the `barrier` to `pop_barrier_with_inputs`, and include the `apply_dispatcher_barrier` in `pop_barrier_with_inputs`? After this change, we won't have two barriers in any of the code scope, and the current  `recv_barrier` can be named `barrier` directly.

src/stream/src/executor/mod.rs
```diff
@@ -1713,3 +1717,85 @@ impl<InputId: Clone + Ord + Hash + std::fmt::Debug, M> DynamicReceivers<InputId,
...
+    pub barrier_rx: mpsc::UnboundedReceiver<Barrier>,
+    pub actor_id: ActorId,
+    // cloned/moved from MergeExecutor, just used to create new inputs
+    pub local_barrier_manager: LocalBarrierManager,
+    pub metrics: Arc<StreamingMetrics>,
+    pub fragment_id: FragmentId,
+}
+
+impl DispatchBarrierBuffer {
+    pub async fn select_once(
```
wenym1: The naming is somehow too implementation specific. Better to name it more for its semantic, like `recv`.

src/stream/src/executor/upstream_sink_union.rs
```diff
@@ -421,43 +457,41 @@ impl UpstreamSinkUnionExecutor {
         let upstreams = DynamicReceivers::new(inputs, Some(barrier_align.clone()), None);
         pin_mut!(upstreams);
 
-        let mut current_barrier = None;
+        let mut cached_barriers = VecDeque::new();
```
wenym1: IIUC, the current implementation is to decouple the management of upstream inputs and the barrier tx to the upstreams.

If so, we can have a helper struct to extract to logic of managing barrier tx and barrier forward to make the code more structural.
```
struct UpstreamSinkBarrierManager {
    local_barrier_manager: LocalBarrierManager,
    barrier_tx: HashMap<FragmentId, Sender<Barrier>>,
    pending_barriers: VecDeque<(Barrier, Option<BoxedSinkInput>)>
}
```

The current `partially_apply_barrier` along with `cached_barriers.push_back((barrier, new_sink_input))` will be its core logic. 

Besides, we can make the current `select_once` to be a method of it and rename it properly. This method can only include the logic of concurrently receive message from upstream and receive barrier and then return the message. The logic of handling the message (add/remove upstreams) can be moved to the for-loop body just like other executors.

src/stream/src/executor/merge.rs
```diff
@@ -291,28 +320,12 @@ impl MergeExecutor {
...
-                                }),
-                            )
-                            .await
-                            .context("failed to create upstream receivers")?;
-
+                        if let Some(mut new_upstreams) = new_inputs {
                             // Poll the first barrier from the new upstreams. It must be the same as
                             // the one we polled from original upstreams.
                             for upstream in &mut new_upstreams {
                                 let new_barrier = expect_first_barrier(upstream).await?;
```
wenym1: Here we call `expect_first_barrier` in a for-loop, which may implicitly introduce dependency between different upstreams. Can we possibly change to use `try_join_all` to make them concurrent by the way?

wenym1:
> > After this PR, do we still need #23221?
> 
> Deadlock will be handled. But consider that `Merge` operator now also builds the channel with upstream when the actor is created (not when it is executed), the changes in `UpstreamSinkUnion` executor can also be retained, for consistency.

Code in #23221 is trivial code move. Let's merge its code change to this PR to review together, and also the its tests can be moved here to test whether this PR has already resolved https://github.com/risingwavelabs/risingwave/issues/23064.

src/stream/src/executor/mod.rs
Chiro11: Rename to `await_next_message`

src/stream/src/executor/mod.rs
Chiro11: I'm thinking, essentially, the `pre_apply_barrier` only blocks on `new_inputs` call, which will wait to build stream-rpc with upstream node (for remote input). If the loop is written inside, it could be canceled, and the next time it would have to re-build the connection, forcing all the previous work of to be redone. Is it really necessary, just to prioritize handling upstream messages when we need to add new upstream to `Merge`? Not to mention that this will make the implementation more complex.

Also, I think it should be pending on `barrier_rx` closed, because `barrier_rx` may receive multiple barriers first, and we can't really close it until the real upstream has completely processed all chunks and barriers.

wenym1: > Is it really necessary, just to prioritize handling upstream messages when we need to add new upstream to Merge?

If we don't do `upstream.next()` while creating the new inputs, we may incur back-pressure to upstream and upstream won't be polled. If creating new inputs depends on some progress in upstream, there will be deadlock (though in current implementation there is no yet).

> the next time it would have to re-build the connection, forcing all the previous work of to be redone.

This isn't necessarily true. A proper implementation should not discard the previous progress when the future used in select is dropped and recreate a new future to redo everything from zero, but instead should be able to resume the previous progress. That's what I meant for handling cancellation safety properlly. An implementation in my mind will be having a enum to represent the current state and store the state in the struct.
```
enum BarrierReceiverState {
    ReceivingBarrier,
    CreatingNewInput(future, barrier),
}
```

We only do `self.barrier_rx.recv()` when we are in `BarrierReceiverState::ReceivingBarrier`. When we are in `CreatingNewInput`, we continue await on the future, and when the future finishes, push the barrier and new input to the queue, and switch to `ReceivingBarrier`. We don't need to store the future of `self.barrier_rx.recv()` because it's already cancellation safe.

> Also, I think it should be pending on barrier_rx closed, because barrier_rx may receive multiple barriers first, and we can't really close it until the real upstream has completely processed all chunks and barriers.

Got it. Agree.

Chiro11: I see. My previous idea was to put only the barrier in the state machine, and in each select we may need to redo the entire process. That's really amazing!

src/stream/src/executor/merge.rs
```diff
@@ -291,28 +297,15 @@ impl MergeExecutor {
...
-                            .context("failed to create upstream receivers")?;
-
+                        if let Some(mut new_upstreams) = new_inputs {
                             // Poll the first barrier from the new upstreams. It must be the same as
                             // the one we polled from original upstreams.
-                            for upstream in &mut new_upstreams {
-                                let new_barrier = expect_first_barrier(upstream).await?;
-                                assert_equal_dispatcher_barrier(barrier, &new_barrier);
+                            let new_barriers =
+                                try_join_all(new_upstreams.iter_mut().map(expect_first_barrier))
```
wenym1: I'm even thinking about including `expect_first_barrier` and `assert_equal_dispatcher_barrier` in the `try_join_all` of `DispatchBarrierBuffer`.

src/stream/src/executor/mod.rs
```diff
@@ -1713,3 +1719,177 @@ impl<InputId: Clone + Ord + Hash + std::fmt::Debug, M> DynamicReceivers<InputId,
...
+    buffer: VecDeque<(Barrier, Option<Vec<BoxedActorInput>>)>,
+    barrier_rx: mpsc::UnboundedReceiver<Barrier>,
+    recv_state: BarrierReceiverState,
+    curr_upstream_fragment_id: FragmentId,
+    actor_id: ActorId,
+    // read-only context for building new inputs
+    build_input_ctx: Arc<BuildInputContext>,
+}
+
+#[derive(Clone)]
```
wenym1: Deriving `Clone` seems unnecessary, since we already wrap it with `Arc`.

src/stream/src/executor/mod.rs
```diff
@@ -1713,3 +1719,177 @@ impl<InputId: Clone + Ord + Hash + std::fmt::Debug, M> DynamicReceivers<InputId,
...
+            if let Err(e) = self.try_fetch_barrier_rx(true).await {
+                return e;
+            }
+        }
+    }
+
+    async fn try_fetch_barrier_rx(&mut self, pending_on_end: bool) -> StreamExecutorResult<()> {
+        match &mut self.recv_state {
+            BarrierReceiverState::ReceivingBarrier => {
+                let Some(barrier) = self.barrier_rx.recv().await else {
```
wenym1, line 1836: I think currently we can assume that the corresponding `barrier_tx` is always in the actor state of `LocalBarrierWorker` (at least in current implementation we clean the actor state only after the stop barrier has been collected from the actors to be dropped). Therefore we can simply treat the end of `barrier_rx` an unexpected error to simplify the handling logic here.

src/stream/src/executor/mod.rs
```diff
@@ -1713,3 +1719,177 @@ impl<InputId: Clone + Ord + Hash + std::fmt::Debug, M> DynamicReceivers<InputId,
...
+
+    pub async fn await_next_message(
+        &mut self,
+        stream: &mut (impl Stream<Item = StreamExecutorResult<DispatcherMessage>> + Unpin),
+    ) -> StreamExecutorResult<DispatcherMessage> {
+        tokio::select! {
+            biased;
+
+            msg = stream.next() => {
+                msg.ok_or(StreamExecutorError::channel_closed(
```
wenym1: Should use `ok_or_else`. `ok_or` create an error for every message even the message is not None.

Besides, can use `try_next` instead of `next` so that the error can be handling directly as `msg?.ok_or_else(...)`

src/stream/src/executor/upstream_sink_union.rs
```diff
@@ -204,26 +206,189 @@ impl Execute for UpstreamSinkUnionExecutor {
...
+                msg = pending_on_none(upstreams.next()) => {
+                    return msg;
+                }
+
+                barrier = self.barrier_rx.recv() => {
+                    // Here, if there's no upstream, the cached barrier here will be sent out immediately in the next
+                    // loop. Otherwise, we need to forward the barrier to the current upstreams, cache this barrier, and
+                    // then wait in the first branch until the upstreams have processed the barrier.
+                    let barrier = barrier.context("Failed to receive barrier from barrier_rx")?;
+                    let pending_barrier = self.partially_apply_barrier(barrier, executor).await?;
```
wenym1: Seems that we will have the same cancellation safety issue here. If we gets pending on `partially_apply_barrier`, and then an upstream message arrives, the barrier will be lost.

src/stream/src/executor/upstream_sink_union.rs
```diff
@@ -204,26 +206,189 @@ impl Execute for UpstreamSinkUnionExecutor {
...
+            let upstream_fragment_id = new_upstream_sink
+                .get_info()
+                .unwrap()
+                .get_upstream_fragment_id();
+            let barrier_rx = self.subscribe_local_barrier(upstream_fragment_id);
+            let new_input = executor
+                .new_sink_input(new_upstream_sink, barrier_rx)
+                .await?;
+
+            self.forward_barrier_to_upstream(&barrier, upstream_fragment_id)?;
```
wenym1: We can change from doing `forward_barrier_to_all_upstreams` at the beginning to after adding new inputs before removing input, and then we don't need this separate `forward_barrier_to_upstream`.

src/stream/src/executor/upstream_sink_union.rs
```diff
@@ -204,26 +206,189 @@ impl Execute for UpstreamSinkUnionExecutor {
...
+            barrier.as_new_upstream_sink(executor.actor_context.fragment_id)
+        {
+            // Create new inputs for the newly added upstream sinks.
+            let upstream_fragment_id = new_upstream_sink
+                .get_info()
+                .unwrap()
+                .get_upstream_fragment_id();
+            let barrier_rx = self.subscribe_local_barrier(upstream_fragment_id);
+            let new_input = executor
+                .new_sink_input(new_upstream_sink, barrier_rx)
```
wenym1: Seems that `new_sink_input` is always paired with `subscribe_local_barrier`. Can we possibly include the logic of `subscribe_local_barrier` in `new_sink_input` and return `(new_sink_input, barrier_tx)`?

src/stream/src/executor/mod.rs
Chiro11: Considering that `UpstreamSinkUnion` also holds the `Merge` executor, and `barrier_rx` is actually forwarded from the upstream's `barrier_tx_map`, this maybe not true.

src/stream/src/executor/upstream_sink_union.rs
Chiro11: Here we await in the select handler instead of in the branch condition. Shouldn't the entire handler be guaranteed to be executed once?
Besides, I did not write a state machine like `Merge`, because I think that is too complex and unnecessary. After all, creating a sink is a rare operation. If the sink creation fails, it is normal to cause back pressure on the upstream, we will always have back pressure when align barrier.

wenym1: Indeed. Can we store the dropped barrier_tx in `removed_upstream_fragment_ids`? In this way we can still satisfy this assumption, and it doesn't increase complexity.

src/stream/src/executor/upstream_sink_union.rs
Chiro11: Although they are always paired, they belong to two different classes. According to your suggestion, `UpstreamSinkBarrierManager` is responsible for managing and forwarding barriers to multiple upstreams, while creating `sink_input` depends on the some info with executor itself, so it seems reasonable to separate them.

src/stream/src/executor/upstream_sink_union.rs
Chiro11: It seems to me that meeting end of `barrier_rx` implicitly depends on the upstream destroying `barrier_tx` always being later than the destruction of output channel. This dependency isn't obvious.

src/stream/src/executor/upstream_sink_union.rs
```diff
@@ -204,26 +206,189 @@ impl Execute for UpstreamSinkUnionExecutor {
...
+            barrier_rx,
+            barrier_tx_map: HashMap::new(),
+            pending_barriers: VecDeque::new(),
+        }
+    }
+
+    async fn await_next_message(
+        &mut self,
+        upstreams: &mut DynamicReceivers<FragmentId, BarrierMutationType>,
+        executor: &mut UpstreamSinkUnionExecutor,
```
wenym1: IIUC, we make the field `upstream_sink_barrier_mgr` to be `Option` in `UpstreamSinkUnionExecutor` because we want to take the ownership of `upstream_sink_barrier_mgr`, while keep the ownership of `UpstreamSinkUnionExecutor` intact, so that its mutable reference can be passed here. 

However, it seems unnecessary to pass the `&mut UpstreamSinkUnionExecutor` now. Previously we need the mutable reference because we also modify its `barrier_tx_map` when we create new sink input. However, now we have moved the `barrier_tx_map` to the `UpstreamSinkBarrierManager`, and there is no need to to pass `&mut UpstreamSinkUnionExecutor`. 

We can actually make the methods to create new sink input to be methods of `UpstreamSinkBarrierManager` rather than `UpstreamSinkUnionExecutor`'s, along with some fields moving from `UpstreamSinkUnionExecutor` to `UpstreamSinkBarrierManager`.

Chiro11: I've considered it, but it would involve too many changes, even including testing. I think it's a bit troublesome.

src/stream/src/executor/upstream_sink_union.rs
wenym1: > It seems to me that meeting end of barrier_rx implicitly depends on the upstream destroying barrier_tx always being later than the destruction of output channel.

I didn't get it. Can you elaborate what it means?

To me, storing the `barrier_tx` until collecting the stop barrier from merge is similar to how we handle stop barrier in local barrier manager, which stores the `barrier_tx` in actor state, and only remove the state after we have collected the stop barrier from the actor.

Chiro11: That's true, but it depend on the current implementation. If we do some changes on `LocalBarrierManager` in the future, it might break. More importantly, wouldn't it be better to implement it (barrier_rx handler) with some robustness?

src/stream/src/executor/upstream_sink_union.rs
wenym1: But now the `UpstreamSinkBarrierManager` is already responsible for creating new sink input. The executor itself is only responsible for managing (receiving, adding, dropping) the created sink inputs. Current code structure is wired, where `UpstreamSinkUnionExecutor` calls `UpstreamSinkBarrierManager`, while inside `UpstreamSinkBarrierManager` it calls `UpstreamSinkUnionExecutor` back. A cleaner implementation is that `UpstreamSinkBarrierManager` can inherently handle all its work without other dependenced state.

Chiro11: Yes, but as another comment said, it's a bit complex. I might check back later to see if it's easy to fix.

src/stream/src/executor/upstream_sink_union.rs
wenym1: Wait. Seems that https://github.com/risingwavelabs/risingwave/pull/23273#discussion_r2375167877 
> Indeed. Can we store the dropped barrier_tx in removed_upstream_fragment_ids? In this way we can still satisfy this assumption, and it doesn't increase complexity.

was responding the wrong comment. It was originally for https://github.com/risingwavelabs/risingwave/pull/23273#discussion_r2375137402.
> Considering that UpstreamSinkUnion also holds the Merge executor, and barrier_rx is actually forwarded from the upstream's barrier_tx_map, this maybe not true.

And after reconsideration, that the complexity of `pending_on_end` is acceptable, so we can keep it.

And for
> Here we await in the select handler instead of in the branch condition. Shouldn't the entire handler be guaranteed to be executed once?
Besides, I did not write a state machine like Merge, because I think that is too complex and unnecessary. After all, creating a sink is a rare operation. If the sink creation fails, it is normal to cause back pressure on the upstream, we will always have back pressure when align barrier.

What's the extra complexity compared to `Merge`? I think their state machine should be the same, and only different in the input type.

Chiro11: It is the same, but just too complicated. I think it is not necessary. Creating `sink_input` is an operation that must succeed. If it fails, it will cause barrier-alignment blocking anyway, because logically we still have to wait for all barriers (including the new barrier) to be aligned. So it will cause back pressure (or deadlock, any failure you want to avoid) anyway.

wenym1: >Creating sink_input is an operation that must succeed. If it fails, it will cause barrier-alignment blocking anyway, because logically we still have to wait for all barriers (including the new barrier) to be aligned.

Why it must succeed? If there is any error, we should report error and trigger recovery.

Chiro11: What I mean here is that the system will become abnormal, so it can only rely creating actor side to detect errors or manually recover. In any case, whether we write a state machine or not, we need to ensure that the input is successfully established. Otherwise, we can only recover the whole system.
Besides, we will not wait for the upstream actor to be really created, but only the establishment of RPC at the process level, IIUC.

wenym1: Agree that we shouldn't implement the state machine for twice. But I think there it still a chance to reuse the state machine in both cases. 

The general logic of the state machine is that, we receive a barrier with mutaiton from `barrier_tx`, and then if we see that we should create new input (of any type) from the barrier, we create a future to create it and switch to another state to poll the future to create input, and when the input is created, switch back to the Receiving state, and enqueue the barrier and optional input. The input type can be any general input type, not necessarily the input stream of `Merge`.

The general logic can be represented via a trait like
```
trait HandleBarrier {
    type Input;
    fn apply_barrier(&mut self, barrier: &Barrier) -> Option<BoxFuture<'static, Self::Input>>;
}
```

The state machine becomes
```
enum BarrierReceiverState<Input> {
    ReceivingBarrier,
    CreatingNewInput(Barrier, BoxFuture<'static, Input>),
}
```

And `DispatchBarrierBuffer` becomes a generic struct that can be reused for both cases.

```
pub(crate) struct BarrierBuffer<H: HandleBarrier> {
    buffer: VecDeque<(Barrier, Option<Vec<H::Input>>)>,
    barrier_rx: mpsc::UnboundedReceiver<Barrier>,
    recv_state: BarrierReceiverState<H::Input>,
    handle: H,
}
```

`Merge` and `UpstreamSinkUnion` can have their own `HandleBarrier`.

You can have a try. If there is any unconsidered blocking point that prevents us from reusing state machine, we can then give up and keep the current blocking implementation.

wenym1: > What I mean here is that the system will become abnormal, so it can only rely creating actor side to detect errors or manually recover. In any case, whether we write a state machine or not, we need to ensure that the input is successfully established. Otherwise, we can only recover the whole system. Besides, we will not wait for the upstream actor to be really created, but only the establishment of RPC at the process level, IIUC.

I still don't get why errors in `creating sink_input` can cause `barrier-alignment blocking anyway`? Why not just return with the error to trigger recovery, instead of still waiting for barrier-alignment? `recover the whole system` is the standard way to handle any error on the streaing pipeline.

Chiro11: > I still don't get why errors in creating sink_input can cause barrier-alignment blocking anyway? Why not just return with the error to trigger recovery, instead of still waiting for barrier-alignment? recover the whole system is the standard way to handle any error on the streaing pipeline.

Not a big deal. Anyway, according to the current implementation, if the creating-input fails, it will rely on rpc to return an error, and it should not block on anything.

src/stream/src/executor/upstream_sink_union.rs
wenym1: > I've considered it, but it would involve too many changes, even including testing. I think it's a bit troublesome.

How do these changes come from? I think the core logic is not changed, and these changes are mostly code movement? If we want to change the owner of these methods from `UpstreamSinkUnionExecutor` to `UpstreamSinkBarrierManager` and while minimize the code changes, we can just keep the method body in the original place, and change the method owner by starting a new `impl` block. 

```
impl UpstreamSinkUnionExecutor {
    ...
    #[cfg(test)]
    pub fn for_test(
        ...
    ) -> Self {
        ...
    }
}

impl UpstreamSinkUnionExecutor {
    async fn new_sink_input(
    ...
```

The current logic of `new_sink_input` is also copied from the previous `handle_update`. We can also reduce the code changes in the same way.

As for test, it shouldn't be a consideration.

After code review, before we merge it, we can then reorganize the code.

src/stream/src/executor/upstream_sink_union.rs
Chiro11: This might be a bit difficult.
I think this is because `UpstreamSinkUnion` and `Merge/Receiver` have some fundamental differences.
1. The upstream of UpstreamSinkUnion may be empty (initially empty, and may become empty again after some create/drop operations), while `Merge` always have non-empty upstreams.
2. `Merge` cannot determine receive barriers from either `barrier_rx` or upstream first, while UpstreamSinkUnion always receive barriers from `barrier_rx` first.

Perhaps we could find a way to handle both case correctly, but this would lead to unclear semantics (code reader might not find those differences).
Furthermore, we might not be able to perform certain checks in shared code, or we might have to move them outside the shared code.
Some interfaces might also differ (for `Merge`, we need to support single input, so we use the `Stream` trait as its arg; for `UpstreamSinkUnion`, we need to check whether `DynamicReceivers` is empty. This could theoretically be moved outside, but it may cause redundancy).

Overall, I don't think it's a good idea to unify them.

src/stream/src/executor/upstream_sink_union.rs
```diff
@@ -232,43 +292,62 @@ impl UpstreamSinkUnionExecutor {
...
             upstream_fragment_id,
             upstream_actors,
             merge_schema,
             project_exprs,
         }: UpstreamFragmentInfo,
+        barrier_rx: UnboundedReceiver<Barrier>,
     ) -> StreamExecutorResult<BoxedSinkInput> {
         let merge_executor = self
-            .new_merge_executor(upstream_fragment_id, upstream_actors, merge_schema)
+            .new_merge_executor(
```
wenym1: After the refactor, it seems cleaner to always do `subscribe_local_barrier` here, so that we don't have to do it outside and pass the `barrier_rx` all the way down to here. And then we can even inline `subscribe_local_barrier`, since here will be the only usage of the method.

src/stream/src/executor/upstream_sink_union.rs
```diff
@@ -284,127 +363,85 @@ impl UpstreamSinkUnionExecutor {
...
-                )
-                .boxed_input();
-
-                inputs.push(input);
-            }
-            inputs
-        };
-
+        let inputs = std::mem::take(&mut self.initial_inputs);
         let execution_stream = self.execute_with_inputs(inputs);
```
wenym1: Can access `self.inputs` directly in `execute_with_inputs`, and inside `execute_with_inputs`, we can consume `self.inputs` directly instead of taking it out.  And then we don't even need this extra wrapper method, but instead we can make `execute_with_inputs` to be `execute_inner`.

src/stream/src/executor/upstream_sink_union.rs
```diff
@@ -204,26 +197,93 @@ impl Execute for UpstreamSinkUnionExecutor {
...
+            initial_inputs: Vec::new(),
+            upstream_sink_barrier_mgr,
+        };
+
+        executor.init_with_upstreams(initial_upstream_infos).await?;
+
+        Ok(executor)
+    }
+
+    async fn init_with_upstreams(
```
wenym1: Seems cleaner to make it a method of `UpstreamSinkBarrierManager`, and return `Vec<BoxedSinkInput>`, and then initialize `UpstreamSinkUnionExecutor` directly with the concrete inputs, instead of initializing with a default value and fill in it immediately later.

PR #23275, opened 2025-09-22
feat(streaming): support locality enforcement and locality backfill

chenzl25:

- Previously we have supported [index selection for backfilling](https://github.com/risingwavelabs/risingwave/pull/23092). While index selection for backfilling only improve the locality of the leaves level of the DAG. To extend the idea of locality, we can introduce locality enforcement to the non-leaves nodes of the DAG. That is a new type of backfilling -- locality backfill. With locality backfill, we can provide good locality to the intermediate result during backfilling. In the circumstances of large historical data backfilling and memory-limited, it can significantly improve the performance.
- In the optimizer, we will follow the `try_better_locality` method to enforce locality for an operator if it can't provide. The operator called `LocalityProvider`. Currently, we only generate this operator for LogicalJoin and LogicalScan. With this ability, users don't even need to create index by themselves. But if users know their workload well, they can create indexes to share indexes across different jobs.
- The LocalityProvider has 2 states which one of the used to buffer data during backfilling and provide data locality. The other one is a progress table like normal backfill operator to track the backfilling progress of itself.
- We use the locality columns as the prefix of the state table pk to provide the locality, so implicitly it will sort the table asynchronously by compactors.
- Once we introduce locality backfill, the dependencies between scan backfill and intermediate locality backfill become important. We need to start the scan backfill first, then next to intermediate locality backfill. Among all intermediate locality backfill, we also need to take care of the backfilling dependency, since intermediate `LocalityProvider` operators could depend on each other. We extend these dependencies ordering control based on https://github.com/risingwavelabs/risingwave/pull/20967.
- TODO: The state table could be truncated after backfilling. 

## Performance

TPCH Q18 with scale = 1g 
We limit the compute node memory to 2g

### Backfill throughput

With locality backfill, our backfilling could finish in 7min, while without the locality backfill, it jobs is too slow to finish.

image 

### Cache miss ratio

With locality backfill, our cache miss ratio is much lower than without it.

image

```
set enable_locality_backfill = true;
explain create materialized view q18 as    select
      c_name,
      c_custkey,
      o_orderkey,
      o_orderdate,
      o_totalprice,
      sum(l_quantity) quantity
    from
      customer,
      orders,
      lineitem
    where
      o_orderkey in (
        select
          l_orderkey
        from
          lineitem
        group by
          l_orderkey
        having
          sum(l_quantity) > 1
      )
      and c_custkey = o_custkey
      and o_orderkey = l_orderkey
    group by
      c_name,
      c_custkey,
      o_orderkey,
      o_orderdate,
      o_totalprice
    order by
      o_totalprice desc,
      o_orderdate
    LIMIT 100;

 StreamMaterialize { columns: [c_name, c_custkey, o_orderkey, o_orderdate, o_totalprice, quantity], stream_key: [c_custkey, c_name, o_orderkey, o_totalprice, o_orderdate], pk_columns: [o_totalprice, o_orderdate, c_custkey, c_name, o_orderkey], pk_conflict: NoCheck }
 └─StreamProject { exprs: [customer.c_name, customer.c_custkey, orders.o_orderkey, orders.o_orderdate, orders.o_totalprice, sum(lineitem.l_quantity)] }
   └─StreamTopN { order: [orders.o_totalprice DESC, orders.o_orderdate ASC], limit: 100, offset: 0 }
     └─StreamExchange { dist: Single }
       └─StreamGroupTopN { order: [orders.o_totalprice DESC, orders.o_orderdate ASC], limit: 100, offset: 0, group_key: [$expr1] }
         └─StreamProject { exprs: [customer.c_name, customer.c_custkey, orders.o_orderkey, orders.o_orderdate, orders.o_totalprice, sum(lineitem.l_quantity), Vnode(orders.o_orderkey) as $expr1] }
           └─StreamHashAgg { group_key: [customer.c_custkey, customer.c_name, orders.o_orderkey, orders.o_totalprice, orders.o_orderdate], aggs: [sum(lineitem.l_quantity), count] }
             └─StreamLocalityProvider { locality_columns: [0, 1, 2, 3, 4] }
               └─StreamExchange [no_shuffle] { dist: HashShard(orders.o_orderkey) }
                 └─StreamHashJoin { type: LeftSemi, predicate: orders.o_orderkey = lineitem.l_orderkey }
                   ├─StreamExchange { dist: HashShard(orders.o_orderkey) }
                   │ └─StreamLocalityProvider { locality_columns: [2] }
                   │   └─StreamExchange [no_shuffle] { dist: HashShard(orders.o_orderkey) }
                   │     └─StreamHashJoin { type: Inner, predicate: orders.o_orderkey = lineitem.l_orderkey }
                   │       ├─StreamExchange { dist: HashShard(orders.o_orderkey) }
                   │       │ └─StreamLocalityProvider { locality_columns: [2] }
                   │       │   └─StreamExchange { dist: HashShard(orders.o_orderkey) }
                   │       │     └─StreamHashJoin { type: Inner, predicate: customer.c_custkey = orders.o_custkey }
                   │       │       ├─StreamExchange { dist: HashShard(customer.c_custkey) }
                   │       │       │ └─StreamLocalityProvider { locality_columns: [0] }
                   │       │       │   └─StreamExchange [no_shuffle] { dist: UpstreamHashShard(customer.c_custkey) }
                   │       │       │     └─StreamTableScan { table: customer, columns: [c_custkey, c_name] }
                   │       │       └─StreamExchange { dist: HashShard(orders.o_custkey) }
                   │       │         └─StreamLocalityProvider { locality_columns: [1] }
                   │       │           └─StreamExchange { dist: HashShard(orders.o_custkey) }
                   │       │             └─StreamTableScan { table: orders, columns: [o_orderkey, o_custkey, o_totalprice, o_orderdate] }
                   │       └─StreamExchange { dist: HashShard(lineitem.l_orderkey) }
                   │         └─StreamLocalityProvider { locality_columns: [0] }
                   │           └─StreamExchange { dist: HashShard(lineitem.l_orderkey) }
                   │             └─StreamTableScan { table: lineitem, columns: [l_orderkey, l_quantity, l_linenumber] }
                   └─StreamExchange { dist: HashShard(lineitem.l_orderkey) }
                     └─StreamProject { exprs: [lineitem.l_orderkey] }
                       └─StreamFilter { predicate: (sum(lineitem.l_quantity) > 1:Decimal) }
                         └─StreamProject { exprs: [lineitem.l_orderkey, sum(lineitem.l_quantity)] }
                           └─StreamHashAgg { group_key: [lineitem.l_orderkey], aggs: [sum(lineitem.l_quantity), count] }
                             └─StreamLocalityProvider { locality_columns: [0] }
                               └─StreamExchange { dist: HashShard(lineitem.l_orderkey) }
                                 └─StreamTableScan { table: lineitem, columns: [l_orderkey, l_quantity, l_linenumber] }
(38 rows)
```

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/planner_test/tests/testdata/output/locality_backfill.yaml
+++ b/src/frontend/planner_test/tests/testdata/output/locality_backfill.yaml
@@ -30,3 +30,14 @@
                 └─StreamLocalityProvider { locality_columns: [t2.a] }
                   └─StreamExchange { dist: HashShard(t2.a) }
                     └─StreamTableScan { table: t2, columns: [t2.a, t2._row_id], stream_scan_type: ArrangementBackfill, stream_key: [t2._row_id], pk: [_row_id], dist: UpstreamHashShard(t2._row_id) }
+- sql: |
+    set enable_locality_backfill = true;
+    create table t (a int, b int, c int, primary key (b, a));
+    select count(*) from t group by a, b;
+  stream_plan: |-
+    StreamMaterialize { columns: [count, t.a(hidden), t.b(hidden)], stream_key: [t.a, t.b], pk_columns: [t.a, t.b], pk_conflict: NoCheck }
+    └─StreamProject { exprs: [count, t.a, t.b] }
+      └─StreamHashAgg { group_key: [t.a, t.b], aggs: [count] }
+        └─StreamLocalityProvider { locality_columns: [t.a, t.b] }
+          └─StreamExchange [no_shuffle] { dist: UpstreamHashShard(t.b, t.a) }
+            └─StreamTableScan { table: t, columns: [t.a, t.b], stream_scan_type: ArrangementBackfill, stream_key: [t.b, t.a], pk: [b, a], dist: UpstreamHashShard(t.b, t.a) }
--- a/src/frontend/planner_test/tests/testdata/input/locality_backfill.yaml
+++ b/src/frontend/planner_test/tests/testdata/input/locality_backfill.yaml
@@ -11,3 +11,9 @@
     select count(*) from t1 join t2 on t1.a = t2.a group by t1.b;
   expected_outputs:
     - stream_plan
+- sql: |
+    set enable_locality_backfill = true;
+    create table t (a int, b int, c int, primary key (b, a));
+    select count(*) from t group by a, b;
+  expected_outputs:
+    - stream_plan


chenzl25:
* **#23275** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23275?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#23324](https://github.com/risingwavelabs/risingwave/pull/23324) <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23324?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#23401](https://github.com/risingwavelabs/risingwave/pull/23401) <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23401?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23275?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

chenzl25:
This PR is ready to review.
Basically the PR contains 3 parts:
- optimizer: Introduce `LocalityProvider` and use it together with `try_better_locality` method.
- backfill ordering: Deal with the dependencies among scan and locality backfill fragments.
- executor: Buffer chunk until `StartFragmentBackfill` and backfill the buffered state table.

kwannoel:
> This PR is ready to review.
> Basically the PR contains 3 parts:
> - optimizer: Introduce `LocalityProvider` and use it together with `try_better_locality` method.
> - backfill ordering: Deal with the dependencies among scan and locality backfill fragments.
> - executor: Buffer chunk until `StartFragmentBackfill` and backfill the buffered state table.

Very cool, I'll take a look tomorrow!

kwannoel:
This pull request introduces a new "locality backfill" feature for streaming queries, which enables more efficient backfilling by grouping and buffering input data using specified locality columns. The changes span the SQL test suite, session configuration, planner, protobuf definitions, and catalog metadata to support this feature end-to-end. The most important changes are grouped below:

**Locality Backfill Feature Implementation**

* Added the `LocalityProvider` plan node (`src/frontend/src/optimizer/plan_node/generic/locality_provider.rs`) and integrated it into the planner logic, enabling operators to buffer input data by locality columns during backfill. This includes changes to the planner's aggregation and join logic to use the new node when the feature is enabled. [[1]](diffhunk://#diff-811d8bec7ec8ec648e64cf6d99fa69fb32e8a6209006b59b634f1cfa50f2df5fR1-R85) [[2]](diffhunk://#diff-7fd905eea59d616f93db99b74b7a7eb43328d7f3d0c5266218ab0fd629af42c8L1419-R1426) [[3]](diffhunk://#diff-52d918738da498aa8f4b47cc238acc0ef154048786b74d8460fd773124e78a3dR1544-R1563) [[4]](diffhunk://#diff-52d918738da498aa8f4b47cc238acc0ef154048786b74d8460fd773124e78a3dL1757-R1791) [[5]](diffhunk://#diff-46b99bf01f33d3074c2ff3143b8cff275ecc6fb60e86b6bed6b4cdb8dc47f17fR98-R100) [[6]](diffhunk://#diff-52d918738da498aa8f4b47cc238acc0ef154048786b74d8460fd773124e78a3dL31-R33)
* Added the `LocalityProviderNode` protobuf message and registered it as a fragment type, allowing the streaming engine to recognize and process locality backfill nodes. [[1]](diffhunk://#diff-548bcb845b3db0225b2c6ab9d843234320a813af3a44855b764aaa284eda2dd8R1001-R1009) [[2]](diffhunk://#diff-548bcb845b3db0225b2c6ab9d843234320a813af3a44855b764aaa284eda2dd8R1072) [[3]](diffhunk://#diff-83b3e54f333bae8008f7d35b8b39a223e7e4ad27b0950e00cdc877a693deb934L640-R641) [[4]](diffhunk://#diff-83b3e54f333bae8008f7d35b8b39a223e7e4ad27b0950e00cdc877a693deb934R896-R900)

**Configuration and Test Coverage**

* Introduced the `enable_locality_backfill` session config parameter (default: false), making the feature opt-in and allowing users to control its activation. [[1]](diffhunk://#diff-4e19861816f5cf913890db3abd8088d01fa135d85f43080d5e6ee5c6c8ac40aaR424-R427) [[2]](diffhunk://#diff-596df18be8f0436c450a415331c053c8ce03101401f1787dd4d0f0be9514a681R43)
* Added new SQL logic tests and planner test cases to verify locality backfill behavior, including a dedicated end-to-end test and planner output validation. [[1]](diffhunk://#diff-e562b006f1cc87bc2ef16ff4795b64ebfbc665ffd937c0d6151b77d986ca61d7R1-R35) [[2]](diffhunk://#diff-68e1ffb6f217c47c266f6524c997bb725606fbdfda0e4172d49fe85d5843d960R1-R13) [[3]](diffhunk://#diff-8e3a6c996a19158d8f2e13f4868703dc8d3e5b18978becc2feb0019c21b55903R1-R32)
* Updated the CI test script to include the locality backfill test in the automated pipeline. [[1]](diffhunk://#diff-42e730c90514e28e228e56261d26331feda1026fe2cf2ba3a16e32b546f80db2R425-R434) [[2]](diffhunk://#diff-42e730c90514e28e228e56261d26331feda1026fe2cf2ba3a16e32b546f80db2R446)

**Catalog and Utility Updates**

* Registered the new fragment type in the catalog and updated related utilities and test cases to support the `LocalityProvider` node. [[1]](diffhunk://#diff-83b3e54f333bae8008f7d35b8b39a223e7e4ad27b0950e00cdc877a693deb934L640-R641) [[2]](diffhunk://#diff-83b3e54f333bae8008f7d35b8b39a223e7e4ad27b0950e00cdc877a693deb934R896-R900) [[3]](diffhunk://#diff-844129dcf84eb267a43cb3579ebd25d5b68416baabbb703b657fe7bd06ef1adfR312-R315)

These changes collectively enable more granular and efficient backfill operations in streaming queries, controlled by a session-level configuration and fully integrated into the query planner and execution engine.

BugenZhao:
> This PR is ready to review.

I'm wondering if it's possible to split this PR into smaller ones to make the review process easier?

ci/scripts/run-backfill-tests.sh
```diff
@@ -422,6 +422,16 @@ test_cross_db_snapshot_backfill() {
   kill_cluster
 }
 
+test_locality_backfill() {
```
BugenZhao, line 425: Theoretically, if we push this into a larger scope or even enable this by default, then

- we can run all e2e tests under this mode
- all existing planner tests need to be updated

May I know your plan for this?

src/frontend/src/optimizer/plan_node/generic/locality_provider.rs
```diff
@@ -0,0 +1,85 @@
...
+        Self {
+            input,
+            locality_columns,
+        }
+    }
+
+    pub fn fields_pretty<'a>(&self) -> Vec<(&'a str, pretty_xmlish::Pretty<'a>)> {
+        let locality_columns_str = format!("{:?}", self.locality_columns);
+        vec![("locality_columns", locality_columns_str.into())]
+    }
```
BugenZhao: Shall we show the field name?

src/frontend/src/optimizer/plan_node/logical_join.rs
```diff
@@ -1754,23 +1774,19 @@ impl ToStream for LogicalJoin {
...
+        if let Some(better_plan) = self.try_better_locality_inner(columns) {
+            Some(better_plan)
+        } else if self.ctx().session_ctx().config().enable_locality_backfill() {
+            Some(
+                LogicalLocalityProvider::new(
+                    self.clone_with_left_right(self.left(), self.right()).into(),
+                    columns.to_owned(),
+                )
+                .into(),
+            )
```
BugenZhao, line 1786: Just realized that we generate a `LocalityProvider` only until we encounter an upstream with different locality. If I'm understanding correctly, we will generate a plan like 

```
Join -> LocalityProvider -> Filter -> Project -> Agg
```

instead of

```
Join -> Filter -> Project -> LocalityProvider -> Agg
```

Does it mean that we need to buffer unnecessary data that could have been filtered out by those stateless executors?

src/frontend/src/optimizer/plan_node/logical_scan.rs
```diff
@@ -680,48 +726,12 @@ impl ToStream for LogicalScan {
...
-        }
-        if self.table_indexes().is_empty() {
-            return None;
-        }
-        let orders = if columns.len() <= 3 {
-            OrderType::all()
+        if let Some(better_plan) = self.try_better_locality_inner(columns) {
+            Some(better_plan)
+        } else if self.ctx().session_ctx().config().enable_locality_backfill() {
+            Some(LogicalLocalityProvider::new(self.clone().into(), columns.to_owned()).into())
```
BugenZhao, line 732: This pattern seems to appear multiple times. Is there any way to extract it?

src/frontend/src/optimizer/plan_node/stream_locality_provider.rs
```diff
@@ -0,0 +1,204 @@
...
+}
+
+impl StreamLocalityProvider {
+    pub fn new(core: generic::LocalityProvider<PlanRef>) -> Self {
+        let input = core.input.clone();
+
+        let dist = match input.distribution() {
+            Distribution::HashShard(keys) => {
+                // If the input is hash-distributed, we make it a UpstreamHashShard distribution
+                // just like a normal table scan. It is used to ensure locality provider is in its own fragment.
```
BugenZhao, line 49: Does it mean that this is actually a workaround for the backfill order control to work correctly? I guess we can keep the `HashShard` and omit one shuffle if it's solely for correctness.

src/frontend/src/optimizer/plan_node/stream_locality_provider.rs
```diff
@@ -0,0 +1,204 @@
...
+        for field in &input_schema.fields {
+            catalog_builder.add_column(field);
+        }
+
+        // Set locality columns as primary key.
+        for locality_col_idx in self.locality_columns() {
+            catalog_builder.add_order_column(*locality_col_idx, OrderType::ascending());
+        }
+        // add streaming key of the input as the rest of the primary key
+        if let Some(stream_key) = input.stream_key() {
```
BugenZhao:
```suggestion
        input.expect_stream_key()
```

src/frontend/src/optimizer/plan_node/stream_locality_provider.rs
```diff
@@ -0,0 +1,204 @@
...
+        let input_schema = input.schema();
+
+        // Add all input columns in original order
+        for field in &input_schema.fields {
+            catalog_builder.add_column(field);
+        }
+
+        // Set locality columns as primary key.
+        for locality_col_idx in self.locality_columns() {
+            catalog_builder.add_order_column(*locality_col_idx, OrderType::ascending());
```
BugenZhao, line 149: Just realize that ideally we should also have `Order` as a requirement for "better locality".

src/frontend/src/optimizer/plan_node/stream_locality_provider.rs
```diff
@@ -0,0 +1,204 @@
...
+                self.input().distribution().dist_column_indices().to_vec(),
+                0,
+            )
+            .with_id(state.gen_table_id_wrapped())
+    }
+
+    /// Build the progress table catalog for tracking backfill progress
+    /// Schema: | vnode | pk(locality columns + input stream keys) | `backfill_finished` | `row_count` |
+    /// Key: | vnode | pk(locality columns + input stream keys) |
+    fn build_progress_catalog(&self, state: &mut BuildFragmentGraphState) -> TableCatalog {
```
BugenZhao, line 169: Can we unify this with other backfill nodes?

src/meta/src/stream/stream_graph/fragment.rs
```diff
@@ -1032,8 +1034,153 @@ impl StreamFragmentGraph {
...
+        // run before LocalityProvider fragments.
+        if fragment_ordering.is_empty() {
+            for value in mapping.values() {
+                for &fragment_id in value {
+                    fragment_ordering.entry(fragment_id).or_default();
+                }
+            }
+        }
+
+        // 2. Add dependencies: all backfill fragments should run before LocalityProvider fragments
```
BugenZhao, line 1048: Is it necessary, or just for the sake of simplicity? For the example below, it seems that we can backfill `Locality` and `Fact` simultaneously.

```
Dim -> Locality 
                 -> Join
          Fact

File: src/meta/src/barrier/progress.rs [hunk 10]
```diff
@@ -183,6 +187,7 @@ impl Progress {
                 BackfillUpstreamType::MView => mv_count += 1,
                 BackfillUpstreamType::Source => source_count += 1,
                 BackfillUpstreamType::Values => (),
+                BackfillUpstreamType::LocalityProvider => mv_count += 1, /* Count LocalityProvider as an MView for progress */
```
BugenZhao, line 190: Wondering if there's (or we should add) a way to inspect the progress of all backfill (and locality provider) nodes separately? This seems to benefit observability of backfill order control a lot, especially we will now "generate" or "manipulate" the backfill order to also handle the locality provider nodes.

BugenZhao (commented):
Looks great to me! Reviewed all files except `stream`.

src/meta/src/barrier/progress.rs
kwannoel: We currently have fragment level backfill observability, in the form of [`rw_fragment_backfill_progress`](https://docs.risingwave.com/sql/system-catalogs/rw-catalog#available-risingwave-catalogs). Because `LocalityProvider` is also dedicated to a single fragment, we can reuse this mechanism to track the progress, provided `LocalityProvider` also has a progress state table. We just need to update the system catalog.

src/frontend/planner_test/tests/testdata/output/locality_backfill.yaml
```diff
@@ -0,0 +1,32 @@
+# This file is automatically generated. See `src/frontend/planner_test/README.md` for more information.
```
kwannoel, line 1: Can we add a test for NoShuffleExchange? We should make sure that the LocalityProvider lies on a different Fragment is no shuffle exchange is used.

src/frontend/src/optimizer/plan_node/logical_locality_provider.rs
```diff
@@ -0,0 +1,181 @@
...
+
+        let input = self.input().to_stream(ctx)?;
+        let required_dist =
+            RequiredDist::shard_by_key(self.input().schema().len(), self.locality_columns());
+        let input = required_dist.streaming_enforce_if_not_satisfies(input)?;
+        let input = if input.as_stream_exchange().is_none() {
+            // Force a no shuffle exchange to ensure locality provider is in its own fragment.
+            // This is important to ensure the backfill ordering can recognize and build
+            // the dependency graph among different backfill-needed fragments.
+            StreamExchange::new_no_shuffle(input).into()
```
kwannoel, line 137: We need to test this.

kwannoel (approved):
Executor part looks good to me.

ci/scripts/run-backfill-tests.sh
chenzl25: We have a tracking issue of locality backfill here https://github.com/risingwavelabs/risingwave/issues/23338.
The final step is to enable it by default. So theoretically, we will update the backfill e2e test and planner tests in this final step.

src/frontend/src/optimizer/plan_node/logical_join.rs
chenzl25: Good insight. To resolve this issue, I think we should add the LocalityProvider to the requirement side instead of the provider side.

src/frontend/src/optimizer/plan_node/stream_locality_provider.rs
chenzl25: Not only for backfill order control but also for backfilling itself. Considering the downstream is a HashJoin and the both inputs distribution are `HashShard` and satisfy the join key distribution requirement. Then in the same fragment, we have 2 backfilling node, and it is coupled and not scalable.

src/meta/src/stream/stream_graph/fragment.rs
chenzl25: Just for the sake of simplicity. For your case, I think we can optimize the scheduling issue further.

src/frontend/src/optimizer/plan_node/stream_locality_provider.rs
BugenZhao: > Then in the same fragment, we have 2 backfilling node, and it is coupled and not scalable.

This is true, because we want different tables to be scaled independently regardless of whether there's a downstream job joining them together.

However, for locality backfill nodes, I think it's okay to couple the parallelism and distribution here, as we don't support scaling a specific fragment now (cc @shanicky). If having multiple backfill nodes in a single fragment/actor could be a concern of performance, we can also split it using no-shuffle exchanges, which also maintain the existing assumption that there should be at most one backfill node per fragment.

src/meta/src/stream/stream_graph/fragment.rs
BugenZhao: Got it. Let's track it in code comments or the tracking issue.

src/frontend/planner_test/tests/testdata/output/locality_backfill.yaml
chenzl25: Sure. Will add it.

src/frontend/src/optimizer/plan_node/stream_locality_provider.rs
chenzl25: But it seems the order doesn't matter, any order can provide the locality we want.

chenzl25:
> I'm wondering if it's possible to split this PR into smaller ones to make the review process easier?

I think to split the PR, I need to guarantee after merging everything is still working as expected, considering right now the code is not that large and it is a minimal PR to demonstrate the functionality, I will keep it as is. Any further optimization would be listed in the tracking issue.

src/frontend/src/optimizer/plan_node/stream_locality_provider.rs
chenzl25: The table scan backfill progress table schema is not always the same as the locality backfill, when the `StreamScanType` is SnapshotBackfill, so we can't unify it. https://github.com/risingwavelabs/risingwave/blob/b4de83b164aede716cd6e385240e2a29fa48e334/src/frontend/src/optimizer/plan_node/stream_table_scan.rs#L139-L186

src/frontend/src/optimizer/plan_node/stream_locality_provider.rs
BugenZhao: Just some brainstorm: if the downstream is a `MAX` aggregation, then descending order may provide better locality?

chenzl25: IIUC, the ordering here for `MAX` aggregation is an optimization of CPU (like we have a shortcut here do not need to update the TopN cache) instead of IO.

chenzl25: If we use `Order ` as a requirement for "better locality", we need to enumerate all the possibility combination in the requirement side which will cause the algorithm complexity grows up fast.

src/frontend/src/optimizer/plan_node/logical_scan.rs
chenzl25: Yep. I will refactor this part together with [this optimization](https://github.com/risingwavelabs/risingwave/pull/23275#discussion_r2415459239) in the following PR.

src/frontend/src/optimizer/plan_node/stream_locality_provider.rs
BugenZhao: This also seems to introduce unnecessary `Exchange` nodes, which may harm performance even after the backfilling stage is finished... 🤔

Issue #23214, opened 2025-09-16
Discussion: how to handle unsigned data type in upstream source and downstream sink


hzxa21:
There is only signed data type (int16, int32, int64) in PG's type system, so does RisingWave. However, it is not uncommon in upstream and downstream system to have unsigned data type. Let's discuss on the behavior on how we should handle that.

Some ideas:
1. Cast unsigned data type to string type (e.g. VARCHAR).
2. Cast unsigned data type to signed data type forcibly. This can potentially overflow.
3. Up cast unsigned data type to signed data type (u16 -> i32, u32 -> i64, i64 -> decimal)
4. Cast unsigned data type to signed data type in best effort. Treat it as parse error if it overflows.

Currently we adopt 4 but I think it is not flexible enough because user doesn't have an easy way to workaround the issue when there is a overflow other than fixing their data, which is hard to do, especially for the source cases.

hzxa21:
@xiangjinwu @BugenZhao @wcy-fdu call for discussion. WDYT?

hzxa21:
Workaround that adopts 2 for mysql source: https://github.com/risingwavelabs/risingwave/pull/23207/files

BugenZhao:
Options 2, 3, and 4 all seem reasonable to me.

I believe it would be most beneficial to introduce an option to allow users to specify the desired behavior, which is also a practice the connectors in many other systems are adopting. For example, there are a lot of `handling.mode` properties in Debezium MySQL connector: https://debezium.io/documentation/reference/stable/connectors/mysql.html#mysql-property-binary-handling-mode

On the other side, I think we should avoid introducing native support for unsigned integers unless it's very necessary to do so, even though a lot of systems have done this (like DuckDB). Creating a set of stable and accurate expressions for interaction between signed and unsigned integers can be highly complex and time-consuming.

hzxa21:
> I believe it would be most beneficial to introduce an option to allow users to specify the desired behavior, which is also a practice the connectors in many other systems are adopting.

+1. That is my thought too. We can have different default behavior for different connectors that suits the common use cases the most.

> On the other side, I think we should avoid introducing native support for unsigned integers unless it's very necessary to do so

+1. User can easily do up cast using SQL statement without native support for unsigned types I believe so I don't think it is necessary to introduce unsigned type natively. Example:
```
SELECT
  my_int8_column,
  CASE
    WHEN my_int8_column >= 0 THEN my_int8_column::numeric
[...]

PR #23278, opened 2025-09-22
fix(mysql-cdc): up cast all unsigned int types

wcy-fdu:

Allow up cast unsigned data type to signed data type (u16 -> i32, u32 -> i64, i64 -> decimal).

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/src/parser/unified/json.rs
+++ b/src/connector/src/parser/unified/json.rs
@@ -49,7 +49,7 @@ pub enum TimeHandling {
     Micro,
 }
 
-#[derive(Clone, Debug)]
+#[derive(Clone, Copy, Debug)]
 pub enum BigintUnsignedHandlingMode {
     /// Convert unsigned bigint to signed bigint (default)
     Long,
@@ -427,39 +427,15 @@ impl JsonParseOptions {
                     _ => {}
                 }
 
-                match Decimal::from_str(str_val) {
-                    Ok(decimal) => decimal.into(),
-                    Err(_) => {
-                        // Only attempt base64 decoding in Precise mode
-                        match self.bigint_unsigned_handling {
-                            BigintUnsignedHandlingMode::Precise => {
-                                // When processing CDC data, if the upstream system has unsigned bigint (e.g., MySQL CDC),
-                                // the best practice is to upcast, i.e., convert unsigned bigint to decimal.
-                                // Only when users configure `debezium.bigint.unsigned.handling.mode='precise'`,
-                                // Debezium will convert unsigned bigint to base64-encoded decimal.
-                                // Reference: https://debezium.io/documentation/reference/stable/connectors/mysql.html#mysql-property-bigint-unsigned-handling-mode
-                                //
-                                // Therefore, here we check if the string falls into the above case after Decimal::from_str returns an error.
-
-                                // A better approach would be to get bytes + org.apache.kafka.connect.data.Decimal from schema
-                                // instead of string, as described in https://github.com/risingwavelabs/risingwave/issues/16852.
-                                // However, Rust doesn't have a library to parse Kafka Connect metadata, so we'll refactor this
-                                // after implementing that functionality.
-                                let value = base64::engine::general_purpose::STANDARD
-                                    .decode(str_val)
-                                    .map_err(|_| create_error())?;
-                                let unscaled = num_bigint::BigInt::from_signed_bytes_be(&value);
-                                Decimal::from_str(&unscaled.to_string())
-                                    .map_err(|_| create_error())?
-                                    .into()
-                            }
-                            BigintUnsignedHandlingMode::Long => {
-                                // In Long mode, return the original error directly
-                                return Err(create_error());
-                            }
-                        }
-                    }
-                }
+                Decimal::from_str(str_val)
+                    .or_else(|_err| {
+                        try_base64_decode_decimal(
+                            str_val,
+                            self.bigint_unsigned_handling,
+                            create_error,
+                        )
+                    })?
+                    .into()
             }
 
             (DataType::Decimal, ValueType::F64) => {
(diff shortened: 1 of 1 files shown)


java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/MySqlValidator.java
```diff
@@ -310,7 +310,8 @@ private boolean isDataTypeCompatible(
                 return Data.DataType.TypeName.INT32_VALUE <= val
                         && val <= Data.DataType.TypeName.INT64_VALUE;
             case "bigint":
-                return val == Data.DataType.TypeName.INT64_VALUE;
+                return val == Data.DataType.TypeName.INT64_VALUE
+                        || val == Data.DataType.TypeName.DECIMAL_VALUE;
```
wcy-fdu, line 314: I will try to distinguish signed bigint from unsigned bigint in a later PR to achieve more accurate schema check.

src/connector/src/parser/unified/json.rs
```diff
@@ -393,26 +406,67 @@ impl JsonParseOptions {
...
+                        ))));
+                    }
+                    _ => {}
+                }
+
+                match Decimal::from_str(str_val) {
+                    Ok(decimal) => decimal.into(),
+                    Err(_) => {
+                        // Only attempt base64 decoding in Precise mode
+                        match self.bigint_unsigned_handling {
```
xiangjinwu: Extract this special logic into a dedicated function, then we can do

```diff
-ScalarImpl::Decimal(Decimal::from_str(str).map_err(|_err| create_error())?)
+ScalarImpl::Decimal(Decimal::from_str(str).or_else(|_err| second_chance_base64(..))?)
```

This avoids deep indentation, and keeps the flow of normal handling uninterrupted when reading thru the code.

src/connector/src/parser/mysql.rs
```diff
@@ -107,12 +144,13 @@ pub fn mysql_datum_to_rw_datum(
             }
         }
         DataType::Int16 => {
-            handle_data_type!(mysql_row, mysql_datum_index, column_name, i16)
+            handle_data_type_with_signed!(mysql_row, mysql_datum_index, column_name, i16, u16)
```
xiangjinwu: How does upcast work here?

src/connector/src/source/cdc/external/mysql.rs
```diff
@@ -518,6 +517,7 @@ impl MySqlExternalTableReader {
                             DataType::Date => Value::from(value.into_date().0),
                             DataType::Time => Value::from(value.into_time().0),
                             DataType::Timestamp => Value::from(value.into_timestamp().0),
+                            DataType::Decimal => Value::from(value.into_decimal().to_string()),
```
xiangjinwu, line 520: So we can also query mysql decimal or bigint unsigned with a string?

wcy-fdu: I think so.

src/connector/src/parser/unified/json.rs
```diff
@@ -393,26 +406,67 @@ impl JsonParseOptions {
...
+                        return Ok(DatumCow::Owned(Some(ScalarImpl::Decimal(
+                            Decimal::NegativeInf,
+                        ))));
+                    }
+                    _ => {}
+                }
+
+                match Decimal::from_str(str_val) {
+                    Ok(decimal) => decimal.into(),
+                    Err(_) => {
```
zwang28: When will Err occur?

src/connector/src/parser/mysql.rs
zwang28: Could you explain the relationship between DataType::Int16, i16, and u16?

src/connector/src/parser/unified/json.rs
wcy-fdu: When try to decode a base64 encode string into decimal. The current logic is to try base64 only if Decimal::from_str(str) fails, which will not affect the correct decimal.handling.mode=string part.

src/connector/src/parser/mysql.rs
wcy-fdu: I think I got it mixed up, handle_data_type_with_signed is not actually for upcast, it's for Cast unsigned data type to signed data type forcibly. (2 in https://github.com/risingwavelabs/risingwave/issues/23214)

removed, and let's keep this pr only solve 3, that is, allow the conversion of `bigint unsigned->decimal`.

src/connector/src/parser/unified/json.rs
```diff
@@ -393,26 +406,67 @@ impl JsonParseOptions {
...
+                        // Only attempt base64 decoding in Precise mode
```
zwang28: I think this can be problematic. Decimal::from_str may succeed even when bigint_unsigned_handling is Precise, in which the result of Decimal::from_str is incorrect.

wcy-fdu: IIUC, when `bigint_unsigned_handling` is Precise, debezium will encode decimal value to it's base64 encode bytes, so it will never success to parse a base64 encoded string into decimal directly.

src/connector/src/parser/unified/json.rs
```diff
@@ -393,26 +406,67 @@ impl JsonParseOptions {
                     .into()
             }
             (DataType::Decimal, ValueType::I64 | ValueType::U64) => {
-                Decimal::from(value.try_as_i64().map_err(|_| create_error())?).into()
+                let i64_val = value.try_as_i64().map_err(|_| create_error())?;
```
zwang28, line 409: Will a `value` of `ValueType::U64` overflow and fail on `try_as_i64`? Is this by design?

wcy-fdu: Good cache! I think this may be a issue? cc @xiangjinwu

xiangjinwu: You are right it would be an error and be filled with null.

> Is this by design?

It is by poor design. Decimals in JSON shall be encoded as string rather than number. Support of i64/u64 in json are non-standard extensions and we do not guarantee its end-to-end quality. This is a fundamental problem of json itself. If the user insist on using json number rather than json string, they are at their own risk to have data loss when it is larger than MAX_SAFE_INTEGER (2^53-1), according to RFC 8259

To summarize, a fix is good to have but not required.

wcy-fdu: I see. Thanks for the explanation!

PR #23283, opened 2025-09-23
feat(row): introduce zero-copy row encoding

BugenZhao:

See https://github.com/risingwavelabs/risingwave/issues/20017 for the background.

Introduce a new row type that is more compact and has better locality than `OwnedRow`, while still being able to borrow `DatumRef` from it with no allocation (zero-copy).

Ideally, it could be completely based on a continuous byte slice. However, in this PR, some scalar variants remain unsupported, and will be directly store into an internal `todo: OwnedRow` as a workaround. This allows for the progressive use of the new row format.

----

This PR adopts `musli-zerocopy` for the layout of the encoding. By deriving `ZeroCopy` on a type, its value can then be safely transmuted to and from bytes. As a result, we don't need to write imperative logic for encoding or decoding, but directly define the layout/structure we want.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/common/src/row/zc.rs
+++ b/src/common/src/row/zc.rs
@@ -25,6 +25,10 @@ use crate::types::{
 };
 
 /// The zero-copy representation of `Datum`.
+///
+/// Each datum is currently 16 bytes, including 1 byte for the tag. If the data can be fit in the
+/// rest 15 bytes with alignment, it will be inlined in `ZcDatum`. Otherwise, it will be stored
+/// somewhere else in the buffer, and a pointer (`Ref`) to it will be stored in `ZcDatum`.
 // TODO(zc): variants that are commented out are not supported yet.
 #[derive(Debug, Copy, Clone, ZeroCopy)]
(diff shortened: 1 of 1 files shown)


BugenZhao:
* **#23323** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23323?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#23315** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23315?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#23283** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23283?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23283?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #23288, opened 2025-09-24
feat(vector): support stream vector index lookup

wenym1:

## Checklist

## Documentation

<details>

</details>


wenym1:
* **#23288** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23288?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23288?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#23154** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23154?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#23339** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23339?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#23464](https://github.com/risingwavelabs/risingwave/pull/23464) <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23464?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

src/frontend/src/optimizer/plan_node/stream_vector_index_lookup_join.rs
```diff
@@ -0,0 +1,114 @@
...
+        let base = PlanBase::new_stream_with_core(
+            &core,
+            core.input.distribution().clone(),
+            core.input.stream_kind(),
+            core.input.emit_on_window_close(),
+            core.input.watermark_columns().clone(),
+            core.input.columns_monotonicity().clone(),
+        );
+        Self { base, core }
+    }
```
chenzl25, line 55: I think we need to think about what's the `stream_key` of VectorIndexLookupJoin. Considering it is a join, so we can't use the input's stream_key as the output of the join stream key.

chenzl25: Considering there is no exchange would be used to enforce the join input, the stream key of VectorIndexLookupJoin should be the left stream_key + right stream_key

wenym1: > Considering it is a join, so we can't use the input's stream_key as the output of the join stream key.

It only has the same name `join` as normal join, and all others are different. There is even no right stream_key, because it yields a single row for each input row, with the output row as `(original row + array(vector topn result))` so the stream key of the input row should be the same as the output row.

chenzl25: I see. You are right

PR #23302, opened 2025-09-24
feat: Datafusion engine for Iceberg batch query

cardigan1008:

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/handler/query.rs
+++ b/src/frontend/src/handler/query.rs
@@ -16,6 +16,7 @@ use std::collections::HashSet;
 use std::sync::Arc;
 use std::time::Instant;
 
+use datafusion::arrow::ipc::convert;
 use itertools::Itertools;
 use pgwire::pg_field_descriptor::PgFieldDescriptor;
 use pgwire::pg_response::{PgResponse, StatementType};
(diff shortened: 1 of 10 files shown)


src/frontend/src/handler/query.rs
```diff
@@ -580,3 +642,193 @@ pub async fn local_execute(
...
+    let first_field_format = formats.first().copied().unwrap_or(Format::Text);
+
+    Ok(PgResponse::builder(stmt_type)
+        .row_cnt_format_opt(Some(first_field_format))
+        .values(row_stream, pg_descs)
+        .into())
+}
+
+/// Convert `DataFusion` Arrow `DataType` to RisingWave `DataType`
+fn convert_arrow_type_to_rw_type(arrow_type: &datafusion::arrow::datatypes::DataType) -> DataType {
```
chenzl25: Is it possible to use `IcebergArrowConvert` ?https://github.com/risingwavelabs/risingwave/blob/fa879a183dcf92cb5c47041dddaa1f8e4170718e/src/common/src/array/arrow/arrow_iceberg.rs#L37 instead

src/frontend/src/optimizer/plan_visitor/plan_rw_to_df_converter.rs
```diff
@@ -0,0 +1,314 @@
...
+            table_name: table_name.into(),
+            source: std::sync::Arc::new(DefaultTableSource::new(std::sync::Arc::new(
+                EmptyTable::new(std::sync::Arc::new(df_schema.as_arrow().clone())),
+            ))),
+            projection: None,
+            projected_schema: std::sync::Arc::new(df_schema),
+            filters: vec![],
+            fetch: None,
+        })
+    }
```
chenzl25, line 306: Right now, it is an empty table, right?

src/frontend/src/handler/query.rs
cardigan1008: Yes, we can use `type_from_field`. But it has some unsupported types that have been implemented here.

src/frontend/src/optimizer/plan_visitor/plan_rw_to_df_converter.rs
cardigan1008: Yes.

PR #23305, opened 2025-09-25
fix(file_source): update state table offset and add more error log

wcy-fdu:

Should update offset in state table to ensure exactly once, and add error log.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/src/parser/parquet_parser.rs
+++ b/src/connector/src/parser/parquet_parser.rs
@@ -13,7 +13,6 @@
 // limitations under the License.
 use std::sync::Arc;
 
-use anyhow::Context;
 use futures_async_stream::try_stream;
 use prometheus::core::GenericCounter;
 use risingwave_common::array::arrow::arrow_array_iceberg::RecordBatch;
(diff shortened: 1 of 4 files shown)


src/connector/src/source/iceberg/parquet_file_handler.rs
```diff
@@ -289,6 +289,37 @@ pub async fn read_parquet_file(
...
+                .map(|m| m.len())
+                .unwrap_or(0)
+        );
+        // Log each leaf column's path and types
+        let schema_descr = file_metadata.schema_descr();
+        for col in schema_descr.columns() {
+            let path = col.path().string();
+            let physical = col.physical_type();
+            let logical = col.logical_type();
+            tracing::debug!(
```
wcy-fdu, line 311: Set per column log to debug level first.

src/stream/src/executor/source/fs_fetch_executor.rs
```diff
@@ -413,6 +414,7 @@ impl<S: StateStore, Src: OpendalSource> FsFetchExecutor<S, Src> {
                                 yield Message::Chunk(chunk);
                             }
                             None => {
+                                tracing::debug!("Deleting file: {:?}", reading_file);
```
wcy-fdu, line 430: Keep per file log to debug level.

hzxa21 (commented):
I suggest we also emit a metrics like [this](https://github.com/risingwavelabs/risingwave/blob/e9e6d91959001043c3508397d70551e488b0a05c/src/stream/src/executor/source/source_backfill_executor.rs#L505) and structure the log with source_id field here so that the error can be correctly displayed in cloud UI as well as grafana dashboard: https://github.com/risingwavelabs/risingwave/blob/38d2d7a7b63ce83127e26de6d3210097c941b9b9/src/stream/src/executor/source/fs_fetch_executor.rs#L267

wcy-fdu:
> I suggest we also emit a metrics like [this](https://github.com/risingwavelabs/risingwave/blob/e9e6d91959001043c3508397d70551e488b0a05c/src/stream/src/executor/source/source_backfill_executor.rs#L505) and structure the log with source_id field here so that the error can be correctly displayed in cloud UI as well as grafana dashboard:
> 
> https://github.com/risingwavelabs/risingwave/blob/38d2d7a7b63ce83127e26de6d3210097c941b9b9/src/stream/src/executor/source/fs_fetch_executor.rs#L267

Added.

src/stream/src/executor/source/fs_fetch_executor.rs
tabVersion: do you think we should add one metric here as it is an event show the consumption at file level.

src/connector/src/parser/parquet_parser.rs
```diff
@@ -122,10 +123,24 @@ impl ParquetParser {
...
+                                format!(
+                                    "Parquet parser ERROR: to_arrow_field failed, column='{}', rw_type='{}', offset={}",
+                                    rw_column_name, rw_data_type, self.offset
+                                )
+                            })?;
                         let array_impl = IcebergArrowConvert
-                            .array_from_arrow_array(&arrow_field, parquet_column)?;
+                            .array_from_arrow_array(&arrow_field, parquet_column)
+                            .with_context(|| format!(
+                                "Parquet parser ERROR: array_from_arrow_array failed, column='{}', rw_type='{}', arrow_field='{}', parquet_type='{}', offset={}",
```
tabVersion: how about making `Parquet parser ERROR` a `ConnectorError`, instead of printing the whole string.

wcy-fdu: >do you think we should add one metric here as it is an event show the consumption at file level.

I think this is unnecessary. I'm just adding a debug log here as a defensive measure to help with troubleshooting. In reality, a file source might read tens of millions of files, so per-file logging and monitoring isn't necessary under normal circumstances.

PR #23306, opened 2025-09-25
feat(frontend): support enforce ssl connection

chenzl25:

- Add a new env `RW_SSL_ENFORCE` (default is false) to control the ssl connection handling behavior.
This pull request enhances the PostgreSQL wire protocol implementation by adding support for enforcing SSL-only connections. It introduces new configuration options to control SSL enforcement, tracks SSL usage per connection, and updates the connection handling logic to reject non-SSL clients when required.

* On startup, if SSL enforcement is enabled but the connection is not using SSL, the server now rejects the connection with an appropriate error message.
* Updated connection initialization to store the `TlsConfig` and initialize the new SSL tracking field.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/utils/pgwire/src/pg_protocol.rs
+++ b/src/utils/pgwire/src/pg_protocol.rs
@@ -29,6 +29,7 @@ use itertools::Itertools;
 use openssl::ssl::{SslAcceptor, SslContext, SslContextRef, SslMethod};
 use risingwave_common::types::DataType;
 use risingwave_common::util::deployment::Deployment;
+use risingwave_common::util::env_var::env_var_is_true;
 use risingwave_common::util::panic::FutureCatchUnwindExt;
 use risingwave_common::util::query_log::*;
 use risingwave_common::{PG_VERSION, SERVER_ENCODING, STANDARD_CONFORMING_STRINGS};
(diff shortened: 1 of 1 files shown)


src/utils/pgwire/src/pg_protocol.rs
```diff
@@ -118,14 +124,28 @@ pub struct TlsConfig {
...
 
 impl TlsConfig {
     pub fn new_default() -> Option<Self> {
         let cert = std::env::var("RW_SSL_CERT").ok()?;
         let key = std::env::var("RW_SSL_KEY").ok()?;
-        tracing::info!("RW_SSL_CERT={}, RW_SSL_KEY={}", cert, key);
-        Some(Self { cert, key })
+        let enforce_ssl = std::env::var("RW_SSL_ENFORCE")
+            .map(|v| v.to_lowercase() == "true" || v == "1")
+            .unwrap_or(false);
```
BugenZhao: Can use `env_var_is_true`.

src/utils/pgwire/src/pg_protocol.rs
```diff
@@ -602,6 +624,7 @@ where
             // Construct ssl stream and replace with current one.
             self.stream.write(BeMessage::EncryptionResponseSsl).await?;
             self.stream.upgrade_to_ssl(context).await?;
+            self.is_ssl_connection = true;
```
BugenZhao: I guess there's no need to introduce this variable. We can check the variant of `self.stream`.

BugenZhao (approved):

Issue #11539, opened 2023-08-08
sink message of debezium json without schema field


masterclock:
### Is your feature request related to a problem? Please describe.

when create a sink like:
```
CREATE SINK some_sink FROM some_view WITH (
    connector = 'kafka',
    properties.bootstrap.server = 'localhost:9092',
    type = 'debezium',
    topic = 'cloud_testees_with_station',
    primary_key = 'id'
);
```
the output message looks like:
```
{
  "schema": ...,
  "payload": ""
}
```
as the "schema" part in message is very large and rarely used, I'd like to remove this field, just like configuration in debezium
```
schamas.enable = false
```

### Describe the solution you'd like

a configuration in `WITH` to disable schema generation.

fuyufjh:
I am not sure whether the debezium format definition allows it 🤔 @tabVersion can you please take a look?

tabVersion:
> I am not sure whether the debezium format definition allows it 🤔 @tabVersion can you please take a look?

yes, debezium allows this operation by specifying `key.converter.schemas.enable` and `value.converter.schemas.enable`.
I am not sure if we can do this now because the generation logic is nested into kafka sink & kinesis sink. Maybe some prior refactoring is needed.

devillove084:
@tabVersion Hi, is there any progress on this issue, I'd like to try and help in any way I can if I can.

tabVersion:
> @tabVersion Hi, is there any progress on this issue, I'd like to try and help in any way I can if I can.

Sorry, there's no update yet. And contribution is welcomed

xiangjinwu:
We already introduced `schemas.enable` for upsert json sink in #12113, which is essentially the same option from debezium because both are based on the Java class `org.apache.kafka.connect.json.JsonConverter`. There has also been some recent fixes regarding how RisingWave types maps to these "kafka connect types" by @xuefengze.

Moving forward, I would suggest to:
* Refactor `DebeziumJsonFormatter` to leverage the `kafka_connect` option from `JsonEncoder`.
* Use `schemas.enable` as the user facing name. When absent, default to `false` for `upsert` but `true` for `debezium`. This is necessary for backward compatibility.

cnsky1103:
hi I've submitted a PR to solve this, @tabVersion can you do some code review?

PR #23307, opened 2025-09-25
feat: make 'schema' field of debezium json sink output optional (#11539)

cnsky1103:

When sink to debezium-style json output, the 'schema' field is now optional, which can be configed by 
```sql
CREATE SINK some_sink FROM some_view WITH (connector = 'kafka', ...)
FORMAT DEBEZIUM ENCODE JSON (schemas.enable = false);
```
The default value is true, when set false,  the schema field is ignored

e2e test:
```sql
dev=> CREATE TABLE debezium_schema_optional_e2e (
    id INT PRIMARY KEY,
    name VARCHAR
);

CREATE SINK debezium_schema_optional_e2e_sink
FROM debezium_schema_optional_e2e
WITH (
    connector = 'kafka',
    topic = 'debezium-schema-optional-test-v2',
    properties.bootstrap.server = '127.0.0.1:9092',
    primary_key = 'id'
)
FORMAT DEBEZIUM ENCODE JSON (
    schemas.enable = 'false'
);

INSERT INTO debezium_schema_optional_e2e VALUES (1, 'alice');
```
```bash
kafka-console-consumer \ 
  --bootstrap-server 127.0.0.1:9092 \
  --topic debezium-schema-optional-test-v2 \
  --from-beginning \
  --max-messages 1 \
  --property print.key=true \
  --property key.separator=' | '
{"payload":{"id":1}} | {"payload":{"after":{"id":1,"name":"alice"},"before":null,"op":"c","source":{"db":"dev","table":"debezium_schema_optional_e2e","ts_ms":1783674827096},"ts_ms":1783674827096}}
Processed a total of 1 messages
```

upsert e2e test:
```sql
dev=> CREATE TABLE upsert_schema_optional_e2e (
    id INT PRIMARY KEY,
    name VARCHAR
);

CREATE SINK upsert_schema_optional_e2e_sink
FROM upsert_schema_optional_e2e
WITH (
    connector = 'kafka',
    topic = 'upsert-schema-optional-test-v2',
    properties.bootstrap.server = '127.0.0.1:9092',
    primary_key = 'id'
)
FORMAT UPSERT ENCODE JSON (
    schemas.enable = 'false'
);

INSERT INTO upsert_schema_optional_e2e VALUES (101, 'alice');
INSERT INTO upsert_schema_optional_e2e VALUES (102, 'bob');
UPDATE upsert_schema_optional_e2e SET name = 'alice-v2' WHERE id = 101;
CREATE_TABLE
CREATE_SINK
INSERT 0 1
INSERT 0 1
UPDATE 1
```

```bash
 kafka-console-consumer \  
  --bootstrap-server 127.0.0.1:9092 \
  --topic upsert-schema-optional-test-v2 \
  --from-beginning \
  --max-messages 20 \
  --property print.key=true \
  --property key.separator=' | '
{"id":102} | {"id":102,"name":"bob"}
{"id":101} | {"id":101,"name":"alice"}
{"id":101} | {"id":101,"name":"alice-v2"}
```

## Checklist

## Documentation

<details>
There is a new JSON format config, key.schema.enable, which tells whether the output contains the 'schema' field.

</details>


hzxa21:
@cnsky1103 CI fails because 
```

<html>
<body>
<!--StartFragment-->
error: this `map_or` can be simplified
--
  | --> src/connector/src/sink/formatter/mod.rs:405:27
  | \|
  | 405 \|           let with_schema = b
  | \|  ___________________________^
  | 406 \| \|             .builder
  | 407 \| \|             .format_desc
  | 408 \| \|             .options
  | 409 \| \|             .get(KEY_SCHEMA_ENABLE)
  | 410 \| \|             .map_or(true, \|s\| s.to_lowercase().parse::<bool>().unwrap_or(true));
  | \| \|_______________________________________________________________________________^
  | \|
  | = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_map_or
  | = note: `-D clippy::unnecessary-map-or` implied by `-D warnings`
  | = help: to override `-D warnings` add `#[allow(clippy::unnecessary_map_or)]`
  | help: use is_none_or instead
  | \|
  | 410 -             .map_or(true, \|s\| s.to_lowercase().parse::<bool>().unwrap_or(true));
  | 410 +             .is_none_or(\|s\| s.to_lowercase().parse::<bool>().unwrap_or(true));
  | \|
  |  
  | error: could not compile `risingwave_connector` (lib) due to 1 previous error
  | warning: build failed, waiting for other jobs to finish...
  | error: could not compile `risingwave_connector` (lib test) due to 1 previous erro

<!--EndFragment-->
</body>
</html>```


Can you fix it?

**Comment by cnsky1103** [comment]
@hzxa21 done fixing

File: src/connector/src/sink/formatter/debezium_json.rs [hunk 1]
```diff
@@ -27,23 +27,48 @@ use crate::sink::encoder::{
 use crate::tri;
 
 const DEBEZIUM_NAME_FIELD_PREFIX: &str = "RisingWave";
+pub(crate) const KEY_SCHEMA_ENABLE: &str = "key.schema.enable";
```
xiangjinwu: In original debezium, `key.schemas.enable` only disables it for kafka key, and the payload part is controlled separately by `value.schemas.enable`.

Given both the RisingWave syntax as well as the current implementation affects both key and value, what about just name it `schemas.enable`? (also note the plural form)

cnsky1103: sure, that's more accurate

cnsky1103:
@xiangjinwu hi, could you have a review on this? it's been a while

src/connector/src/sink/formatter/debezium_json.rs
```diff
@@ -27,23 +27,48 @@ use crate::sink::encoder::{
 use crate::tri;
 
 const DEBEZIUM_NAME_FIELD_PREFIX: &str = "RisingWave";
+pub(crate) const KEY_SCHEMA_ENABLE: &str = "schemas.enable";
```
xiangjinwu: nit: it is no longer key-only. And we shall also update the PR description as well.

src/connector/src/sink/formatter/debezium_json.rs
```diff
@@ -27,23 +27,48 @@ use crate::sink::encoder::{
...
+    P: Into<Value>,
+{
+    let payload_value: Value = payload.into();
+    if with_schema {
+        json!({
+            "schema": schema_val,
+            "payload": payload_value,
+        })
+    } else {
+        json!({ "payload": payload_value })
```
xiangjinwu, line 68: Confirm the behavior when schema is off with standard debezium, or the upsert sink of RisingWave.

xiangjinwu (commented):
Could you also add some tests for the new behavior? You can check for existing sink tests

Thank you!

src/connector/src/sink/formatter/debezium_json.rs
cnsky1103: @xiangjinwu 
test added, and the e2e test is in the PR description, please have a review

PR #23321, opened 2025-09-26
fix(sink): fix data loss caused by rate limit reader cancellation unsafety

wenym1:

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/src/sink/log_store.rs
+++ b/src/connector/src/sink/log_store.rs
@@ -15,14 +15,13 @@
 use std::cmp::Ordering;
 use std::collections::VecDeque;
 use std::fmt::Debug;
-use std::future::{Future, poll_fn};
+use std::future::{Future, pending, poll_fn};
 use std::pin::pin;
 use std::sync::Arc;
 use std::task::Poll;
 use std::time::Instant;
 
 use await_tree::InstrumentAwait;
-use either::Either;
 use futures::future::BoxFuture;
 use futures::{TryFuture, TryFutureExt};
 use risingwave_common::array::StreamChunk;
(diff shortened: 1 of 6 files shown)


wenym1:
* **#23321** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23321?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23321?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/connector/src/sink/log_store.rs
```diff
@@ -531,36 +543,43 @@ impl<R: LogReader> LogReader for RateLimitedLogReader<R> {
...
+            }
+            truncate_offset = Some(*upstream_offset);
+            self.core.consumed_offset_queue.pop_back();
+        }
+        if !stop && let Some((_, downstream_offsets, _)) = &mut self.core.consuming_chunk {
+            while let Some(prev_downstream_offset) = downstream_offsets.back() {
+                if *prev_downstream_offset <= downstream_offset {
+                    downstream_offsets.pop_back();
+                } else {
+                    // stop = true;
```
hzxa21, line 580: This can be removed.

wenym1:
> LGTM!
> 
> Is it possible to add a sim test coverage in TestSink
> 
> https://github.com/risingwavelabs/risingwave/blob/4f84de6aa6a5de84223f383787ab471dcb932875/src/tests/simulation/tests/integration_tests/sink/utils.rs#L422
> 
> to test the case when LogReader is interrupible? We can do it in a separate PR though.

Added in https://github.com/risingwavelabs/risingwave/pull/23388.

src/connector/src/sink/log_store.rs
wenym1: The code is commented in purpose. Want to maintain the semantic of `stop`, but if the code is not commented, we will get a linter warning that the written value is not read.

PR #23324, opened 2025-09-26
feat(streaming): support truncating locality backfill state

chenzl25:

- Support truncating the locality backfill state table after backfilling.
- Extend the `StartFragmentBackfill` barrier mutation to carry the truncation messages to the locality backfill.
- Considering we have a backfill ordering of A before B. The truncation of A will be scheduled together the backfill starting of B.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/meta/src/barrier/checkpoint/control.rs
+++ b/src/meta/src/barrier/checkpoint/control.rs
@@ -884,7 +884,7 @@ impl DatabaseCheckpointControl {
                         .extend(refresh_finished_table_ids);
                 }
 
-                let mut finished_jobs = self.create_mview_tracker.apply_collected_command(
+                let staging_commit_info = self.create_mview_tracker.apply_collected_command(
                     node.command_ctx.command.as_ref(),
                     &node.command_ctx.barrier_info,
                     &node.state.resps,
@@ -894,7 +894,7 @@ impl DatabaseCheckpointControl {
                     .cdc_table_backfill_tracker
                     .apply_collected_command(&node.state.resps);
                 if !node.command_ctx.barrier_info.kind.is_checkpoint() {
-                    assert!(finished_jobs.is_empty());
+                    assert!(staging_commit_info.is_none());
                     node.notifiers.into_iter().for_each(|notifier| {
                         notifier.notify_collected();
                     });
@@ -909,12 +909,16 @@ impl DatabaseCheckpointControl {
                     }
                     continue;
                 }
+                let mut staging_commit_info =
+                    staging_commit_info.expect("should be Some for checkpoint");
                 node.state
                     .finished_jobs
                     .drain()
                     .for_each(|(job_id, (resps, source_change))| {
                         node.state.resps.extend(resps);
-                        finished_jobs.push(TrackingJob::new(job_id.table_id as _, source_change));
+                        staging_commit_info
+                            .finished_jobs
+                            .push(TrackingJob::new(job_id.table_id as _, source_change));
                     });
                 let task = task.get_or_insert_default();
                 node.command_ctx.collect_commit_epoch_info(
@@ -923,7 +927,7 @@ impl DatabaseCheckpointControl {
                     self.collect_backfill_pinned_upstream_log_epoch(),
                 );
                 self.completing_barrier = Some(node.command_ctx.barrier_info.prev_epoch());
-                task.finished_jobs.extend(finished_jobs);
+                task.finished_jobs.extend(staging_commit_info.finished_jobs);
                 task.finished_cdc_table_backfill
                     .extend(finished_cdc_backfill);
                 task.notifiers.extend(node.notifiers);
@@ -933,6 +937,9 @@ impl DatabaseCheckpointControl {
                         (Some((node.command_ctx, node.enqueue_time)), vec![]),
                     )
                     .expect("non duplicate");
+                task.commit_info
+                    .truncate_tables
+                    .extend(staging_commit_info.table_ids_to_truncate);
                 break;
             }
         }
--- a/src/meta/src/barrier/progress.rs
+++ b/src/meta/src/barrier/progress.rs
@@ -41,6 +41,15 @@ enum BackfillState {
     Done(ConsumedRows),
 }
 
+/// Represents the backfill nodes that need to be scheduled or cleaned up.
+#[derive(Debug, Default)]
+pub(super) struct PendingBackfillFragments {
+    /// Fragment IDs that should start backfilling in the next checkpoint
+    pub next_backfill_nodes: Vec<FragmentId>,
+    /// State tables of locality provider fragments that should be truncated
+    pub truncate_locality_provider_state_tables: Vec<TableId>,
+}
+
 /// Progress of all actors containing backfill executors while creating mview.
 #[derive(Debug)]
 pub(super) struct Progress {
@@ -101,13 +110,14 @@ impl Progress {
     }
 
     /// Update the progress of `actor`.
+    /// Returns the backfill fragments that need to be scheduled or cleaned up.
     fn update(
         &mut self,
         actor: ActorId,
         new_state: BackfillState,
         upstream_total_key_count: u64,
-    ) -> Vec<FragmentId> {
-        let mut next_backfill_nodes = vec![];
+    ) -> PendingBackfillFragments {
+        let mut result = PendingBackfillFragments::default();
         self.upstream_mvs_total_key_count = upstream_total_key_count;
         let total_actors = self.states.len();
         let backfill_upstream_type = self.backfill_upstream_types.get(&actor).unwrap();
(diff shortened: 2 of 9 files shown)


chenzl25:
* **#23454** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23454?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#23324** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23324?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23324?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#23275** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23275?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#23401](https://github.com/risingwavelabs/risingwave/pull/23401) <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23401?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

BugenZhao:
> * Considering we have a backfill ordering of A before B. The truncation of A will be scheduled together the backfill starting of B.

I'm a bit confusing... 🤔 Why do we need meta service to coordinate truncation here? Once a locality backfill executor completes its backfilling stage, can it directly request to truncate its state table?

chenzl25:
> I'm a bit confusing... 🤔 Why do we need meta service to coordinate truncation here? Once a locality backfill executor completes its backfilling stage, can it directly request to truncate its state table?

The truncation api right now is table level instead of vnode level, however, each backfill actors might complete in different epoch which depends on the data distribution of vnodes of each actor. So we need to coordinate the truncation on the meta. Unless we have supported a vnode level truncation api.  cc @Li0k

BugenZhao:
> Unless we have supported a vnode level truncation api.

Makes sense. I thought truncation can be performed or implemented as simple as `vnodes.for_each(|vnode| table.range_delete(prefix(vnode)))`, but it seems more complicated than I expected. 😕 cc @wenym1 @hzxa21 WDYT?

Another approach might be allowing the meta service to collect the vnode bitmap of a table that needs truncation. Once it identifies the complete set of vnodes, it can proceed with the truncation.

hzxa21:
> > Unless we have supported a vnode level truncation api.
> 
> Makes sense. I thought truncation can be performed or implemented as simple as `vnodes.for_each(|vnode| table.range_delete(prefix(vnode)))`, but it seems more complicated than I expected. 😕 cc @wenym1 @hzxa21 WDYT?
> 
> Another approach might be allowing the meta service to collect the vnode bitmap of a table that needs truncation. Once it identifies the complete set of vnodes, it can proceed with the truncation.

Storage doesn't support arbitrary `range_delete` and only support setting a monotonically increasing/decreasing per vnode watermark for range delete. However, per vnode truncation doesn't imply that we can have a monotonically increasing/decreasing PK prefix watermark. Instead, it implies that we should have an epoch watermark (epoch is encoded in the suffix of the key), which is less efficient and more complicated to implement in the storage side. That is why I think @Li0k choose to implement the truncation using pure metadata operations in hummock manager.

@Li0k is the author of the current approach so he knows more details on this subject. Feel free to provide more contexts.

src/stream/src/task/barrier_worker/managed_state.rs
```diff
@@ -1168,6 +1175,47 @@ impl DatabaseManagedBarrierState {
...
+        };
+        let Some(graph_state) = self.graph_states.get_mut(partial_graph_id) else {
+            warn!(
+                ?epoch,
+                actor_id, table_id, "ignore truncate state table: graph_state not found"
+            );
+            return;
+        };
+        graph_state
+            .truncate_tables
```
wenym1: Should comment in the previous PR.

IIUC, this `truncate_tables` is only for reporting the tables to truncate back to meta node global barrier manager. However, the information of what tables to truncate should already be available in global barrier manager, in the barrier mutation of the `Command`. Therefore, it seems unnecessary to report and maintain it here in local barrier manager on CN. When we construct the `CommitEpochInfo`, we can directly get the tables to truncate from the barrier mutation.

chenzl25: I think in the meta, we don't maintain the relationship of which tables to truncate in the fragment, unless we let the exact executor to report the tables to truncate.

wenym1: Shall we include this information in meta, in the `CreateMviewTracker`? It might not be difficult, because we should have the information of fragment StreamNode when we want to add new tracking items to the tracker (e.g. create new streaming job, or restore backfilling jobs in recovery). 

Current implementation looks wired to collect logical information, which is already available on meta node, from runtime executors.

chenzl25: You are right. I just refactor the code and let it truncate the tables on meta node directly.

src/meta/src/stream/stream_graph/fragment.rs
```diff
@@ -1084,6 +1084,41 @@ impl StreamFragmentGraph {
...
+            // Check if this fragment contains a LocalityProvider node
+            if let Some(node) = fragment.node.as_ref() {
+                let mut state_table_ids = Vec::new();
+
+                visit_stream_node_cont(node, |stream_node| {
+                    if let Some(NodeBody::LocalityProvider(locality_provider)) =
+                        stream_node.node_body.as_ref()
+                    {
+                        // Collect state table ID
+                        if let Some(state_table) = &locality_provider.state_table {
```
wenym1: Can we assume that there must be `state_table` set for `locality_provider`? If so, we can just unwrap.

src/meta/src/rpc/ddl_controller.rs
```diff
@@ -1677,6 +1677,14 @@ impl DdlController {
             cross_db_snapshot_backfill_info
         );
 
+        let locality_fragment_state_table_mapping =
```
wenym1, line 1642: Currently how do we handle the case of recovery for LocalityProvider? Seems that this is the only call to `find_locality_provider_fragment_state_table_mapping`, and in recovery we won't call the method to resolve this information.

src/meta/src/barrier/checkpoint/control.rs
```diff
@@ -859,12 +859,13 @@ impl DatabaseCheckpointControl {
                         .extend(refresh_finished_table_ids);
                 }
 
-                let mut finished_jobs = self.create_mview_tracker.apply_collected_command(
-                    node.command_ctx.command.as_ref(),
-                    &node.command_ctx.barrier_info,
-                    &node.state.resps,
-                    hummock_version_stats,
-                );
+                let (mut finished_jobs, table_ids_to_truncate) =
```
wenym1: When the barrier is not checkpoint barrier, we won't call `commit_epoch` to truncate the state table. Do we handle this corner case? In code below we may add an assertion that `table_ids_to_truncate.is_empty()` when `!barrier.is_checkpoint()`.

src/meta/src/rpc/ddl_controller.rs
chenzl25: cc @kwannoel Is the backfill ordering supports recovery? I think the `locality_fragment_state_table_mapping` is sharing the same code path of backfill order state.

kwannoel: Nope, can see our docs:
> Backfill order recovery is not supported for background ddl. If a backfilling job restarts, the specified order will be lost, and all tables will backfill concurrently.

chenzl25: I see. @wenym1 Locality backfill is not supported running in background now, since it depends on backfill ordering. Let me track this issue.

src/meta/src/barrier/checkpoint/control.rs
chenzl25: Done

src/meta/src/rpc/ddl_controller.rs
chenzl25: ban background ddl for locality backfill here https://github.com/risingwavelabs/risingwave/pull/23460

src/meta/src/barrier/progress.rs
```diff
@@ -510,6 +551,15 @@ impl CreateMviewProgressTracker {
         self.pending_backfill_nodes.extend(backfill_nodes);
     }
 
+    pub(super) fn stash_table_ids_to_truncate(&mut self, table_ids_to_truncate: Vec<TableId>) {
+        self.pending_table_ids_to_truncate
+            .extend(table_ids_to_truncate);
+    }
+
+    pub(super) fn take_table_ids_to_truncate(&mut self) -> Vec<TableId> {
```
wenym1: It seems that this method is always called together with `take_finished_jobs`. Shall we unify them into a single method?

src/meta/src/barrier/progress.rs
```diff
@@ -410,64 +445,70 @@ impl CreateMviewProgressTracker {
...
     /// Return the finished jobs when the barrier kind is `Checkpoint`
+    /// and the table ids whose locality provider state tables need to be truncated.
     pub(super) fn apply_collected_command(
         &mut self,
         command: Option<&Command>,
         barrier_info: &BarrierInfo,
         resps: impl IntoIterator<Item = &PbBarrierCompleteResponse>,
         version_stats: &HummockVersionStats,
-    ) -> Vec<TrackingJob> {
+    ) -> (Vec<TrackingJob>, Vec<TableId>) {
```
wenym1: Now we have two return values. Shall we change to `Option<(Vec<TrackingJob>, Vec<TableId>)>`? We can even introduce a struct named like `StagingCommitInfo` to hold the two values. We can impl Default on it, and then we can return `Some(take(&mut self.staging_commit_info))` directly.

PR #23331, opened 2025-09-28
fix: only check schema usage privilege for show object command

yezizp2012:

As title, aligns the behaviors with `\dt`.

## Checklist

## Documentation

<details>

</details>


yezizp2012:
> Should we add e2e tests for the privilege check?

Search path has been considered in the e2e test already. However switching users is unsupported in SQLLogictest. Therefore, we cannot currently add tests to verify the behavior for users lacking `USAGE` privileges on the specified schema.

PR #23339, opened 2025-09-29
refactor(vector): extract common vector index reader

wenym1:

Similar to `BatchTable`, in this PR, we extract a `VectorIndexReader` to encapsulate the logic to read from vector index, so that the logic can be easily reused in future PR. The `VectorIndexReader` can be built from `VectorIndexReaderDesc`, which is a counterpart of `StorageTableDesc` and contains all logical information required to read from a vector index.

Besides, in this PR, we include the following two more fields in `VectorIndexReaderDesc`.
```
info_output_indices: Arc<Vec<usize>>,
include_distance: bool,
```
`info_output_indices` is used to specified which info columns of vector index will be included in the output, and `include_distance` is used to optionally include the distance column in the output. Previously before this PR, we always include all info columns with the distance column in the output.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/storage/src/table/batch_table/vector_index_reader.rs
+++ b/src/storage/src/table/batch_table/vector_index_reader.rs
@@ -17,7 +17,7 @@ use std::sync::Arc;
 use itertools::Itertools;
 use risingwave_common::array::{
     Array, ArrayBuilder, ArrayImpl, DataChunk, ListArrayBuilder, ListValue, StructArrayBuilder,
-    StructValue, VectorRef,
+    StructValue,
 };
 use risingwave_common::catalog::TableId;
 use risingwave_common::row::RowDeserializer;
@@ -36,6 +36,10 @@ pub struct VectorIndexReader<S> {
     vector_info_struct_type: StructType,
     state_store: S,
     table_id: TableId,
+
+    info_output_indices: Arc<Vec<usize>>,
+    include_distance: bool,
+
     top_n: usize,
     measure: DistanceMeasurement,
     sqrt_distance: bool,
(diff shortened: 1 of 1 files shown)


wenym1:
* **#23339** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23339?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#23154](https://github.com/risingwavelabs/risingwave/pull/23154) <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23154?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#23464](https://github.com/risingwavelabs/risingwave/pull/23464) <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23464?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23339?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/frontend/src/optimizer/plan_node/generic/mod.rs
```diff
@@ -93,11 +93,15 @@ mod postgres_query;
 pub use postgres_query::*;
 
 mod mysql_query;
+
```
xiangjinwu, line 96: unintended?

src/frontend/src/optimizer/plan_node/generic/vector_index_lookup_join.rs
```diff
@@ -0,0 +1,147 @@
...
+                        ),
+                )
+                .into(),
+            ),
+        ));
+        schema
+    }
+
+    fn stream_key(&self) -> Option<Vec<usize>> {
+        self.input.stream_key().map(|key| key.to_vec())
```
xiangjinwu, line 81: Just to highlight this function is filled up by this refactor PR. It was returning None.

xiangjinwu (approved):
LGTM.

Given the proto change is for batch plan, there should be no backward compatibility issue when upgrading a cluster from v2.6 to v2.7. Right?

wenym1:
> LGTM.
> 
> Given the proto change is for batch plan, there should be no backward compatibility issue when upgrading a cluster from v2.6 to v2.7. Right?

Yes. It's only batch plan node changes, and not persisted, so no worry for backward compatibility issue.

PR #23350, opened 2025-09-30
feat(sink): support ADLS for iceberg sink

xxhZs:

as title 
must use it with rest catalog

## Checklist

## Documentation

<details>

</details>

1. in lakekeeper: 

- use azure -> client credentials: 
   You need to create a client in Azure first, and then fill in client-id + client-secret + tenant-id.
- use azure -> shared access key:
   You can set access key into shared access key
- the configuration shared by the two options above.
   account-name and filesystem（It refers to the name of the container corresponding to you in the storage.）The rest can be left as default, or filled in as needed.

2. in rw pgsql 

like

```
CREATE SINK s6 from t1 WITH (
    connector='iceberg',
            type='upsert',
            primary_key = 'v1',
            catalog.type = 'rest',
            catalog.uri = 'http://127.0.0.1:8181/catalog/',
            adlsgen2.account_name = 'xxx',
            adlsgen2.account_key= 'xxx',
            catalog.name='demo',
            database.name='s1',
            table.name='t1',
            warehouse.path = 'test',
            create_table_if_not_exists = 'true',
);
```

### DIFF
--- a/src/connector/with_options_connection.yaml
+++ b/src/connector/with_options_connection.yaml
@@ -126,6 +126,15 @@ IcebergConnection:
       - Each header is a key-value pair, separated by an '='.
       - Multiple headers can be specified, separated by a ';'.
     required: false
+  - name: adlsgen2.account_name
+    field_type: String
+    required: false
+  - name: adlsgen2.account_key
+    field_type: String
+    required: false
+  - name: adlsgen2.endpoint
+    field_type: String
+    required: false
 KafkaConnection:
   fields:
   - name: properties.bootstrap.server


risedev.yml
```diff
@@ -130,10 +130,10 @@ profile:
...
         exporter-port: 1261
-      - use: prometheus
-      - use: grafana
-      - use: kafka
-        persist-data: true
+      # - use: prometheus
+      # - use: grafana
+      # - use: kafka
+      #   persist-data: true
       - use: lakekeeper
```
chenzl25: revert it?

src/connector/src/connector_common/iceberg/mod.rs
```diff
@@ -69,6 +69,13 @@ pub struct IcebergCommon {
     #[serde(rename = "azblob.endpoint_url")]
     pub azblob_endpoint_url: Option<String>,
 
+    #[serde(rename = "adlsgen2.account_name")]
+    pub adlsgen2_account_name: Option<String>,
+    #[serde(rename = "adlsgen2.account_key")]
+    pub adlsgen2_account_key: Option<String>,
+    #[serde(rename = "adlsgen2.endpoint")]
+    pub adlsgen2_endpoint: Option<String>,
```
chenzl25, line 77: Where will we use this parameter?

src/connector/src/connector_common/iceberg/mod.rs
```diff
@@ -69,6 +69,13 @@ pub struct IcebergCommon {
     #[serde(rename = "azblob.endpoint_url")]
     pub azblob_endpoint_url: Option<String>,
 
+    #[serde(rename = "adlsgen2.account_name")]
```
chenzl25, line 72: Shall we add more parameters like `tenant-id`, `client-id` to this configuration? https://github.com/chenzl25/iceberg-rust/blob/2d4407d280b9c49c68ca15a26970738a18039ebc/crates/iceberg/src/io/storage_azdls.rs#L34-L55

Issue #23332, opened 2025-09-28
bug: compaction sees the same epoch multiple times for a key in singleton state table


hzxa21:
### Describe the bug

I suspect there is a bug somewhere to miss deriving the correct singleton distribution for some queries. We have seen several occurrences of the following panics recently:
```
thread 'rw-compaction' panicked at /risingwave/src/storage/hummock_sdk/src/key.rs:1086:21:
key UserKey { 21770, TableKey { 0000 } } epoch EpochWithGap(9237667740385280) >= prev epoch EpochWithGap(9237667740385280)
```

In all occurrences, the panic happens on table with `TableKey { 0000 }`, which should be the state table supposed to be written by operator with singleton distribution.

We saw this panic in the CN shared buffer compaction as well as in the compactor compaction. That means it is highly possible that there are duplicate writes into the same singleton table row.

### Error message/log

```text

```

### To Reproduce
v2.3.4
v2.4.3
v2.5.1
1. Start RisingWave cluster: `risedev ci-start ci-3cn-1fe`.
2. Create tables: `psql -h localhost -p 4566 -d dev -U root -f create_table.sql` 

[create_table.sql](https://github.com/user-attachments/files/22588274/create_table.sql)

3. Create MV and cause panic: `psql -h localhost -p 4566 -d dev -U root -f create_mv_panic.sql`

[create_mv_panic.sql](https://github.com/user-attachments/files/22588276/create_mv_panic.sql)

You'll find error message like `key UserKey { 44, TableKey { 0000 } } epoch EpochWithGap(9298423982456832) >= prev epoch EpochWithGap(9298423982456832)` in all 3 compute nodes.

Li0k:
https://github.com/risingwavelabs/risingwave/pull/22829

sidd-27:
@hzxa21 in src/frontend/src/stream_fragmenter/mod.rs, theres no handling for NodeBody::SimpleAgg, so requires_singleton is never set to true for simple aggregations. which may be causing the duplicate writes into the same singleton table row

PR #23369, opened 2025-10-06
refactor(sqlsmith): support path-based reduction

cardigan1008:

This PR refactors the SQL reduction system from a rule-based approach to a path-based approach, significantly improving its ability to handle complex nested SQL structures and WITH clauses. The new system uses precise AST path navigation to systematically explore and reduce SQL components at any nesting level.

### How does this PR work?

1. Path-Based Navigation System
- Introduced `PathComponent `enum (`Field(String)` and `Index(usize)`) for precise AST node addressing.
- Implemented `get_node_at_path()` and `set_node_at_path()` for traversing complex nested structures.
- Added comprehensive `AstNode` enum covering all SQL components (Query, Select, With, Cte, ExprList, etc.).
2. Systematic Path Enumeration
- `enumerate_reduction_paths()` recursively discovers all reducible paths in the AST.
- Generates reduction candidates for each discoverable node.
- Supports deep nesting including subqueries, CTEs, and complex expressions.
3. Enhanced Reduction Rules
- Converted existing rules to work with the path-based system.
- Added support for previously unsupported constructs (WITH clauses, complex expressions).
- Implemented `RemoveListElement` operations for fine-grained list simplification.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/tests/sqlsmith/src/bin/reducer.rs
+++ b/src/tests/sqlsmith/src/bin/reducer.rs
@@ -59,8 +59,8 @@ async fn main() {
             panic!("Failed to restore RW");
         }
         Err(err) => {
-            tracing::error!("Failed to execute restore command: {}", err);
-            panic!("Failed to execute restore command: {}", err);
+            tracing::error!("Failed to execute restore command: {}", err.as_report());
+            panic!("Failed to execute restore command: {}", err.as_report());
         }
     }
 


cardigan1008:
For #23332 , reducer works as follows: 

```sql
diff --git a/error.sql.log b/reduced.sql.log
index 4007680fff..36d1b475eb 100644
--- a/error.sql.log
+++ b/reduced.sql.log
@@ -10,9 +10,6 @@ FROM
     )
     FULL JOIN (
         SELECT
-            tumble_3.c1 AS col_0,
-            hop_2.c3 AS col_1,
-            hop_2.c14 AS col_2,
             hop_2.c3 AS col_3
         FROM
             (
@@ -28,25 +25,14 @@ FROM
                     FROM
                         (
                             SELECT
-                                tumble_6.category AS col_0,
-                                tumble_5.date_time AS col_1,
-                                TIME '00:18:54' AS col_2
+                                tumble_6.category AS col_0
                             FROM
                                 tumble(bid, bid.date_time, INTERVAL '15') AS tumble_5
                                 FULL JOIN tumble(auction, auction.expires, INTERVAL '4') AS tumble_6
                                     ON tumble_5.extra = tumble_6.description
                                     AND (252 > 1)
                             GROUP BY
-                                tumble_6.seller,
-                                tumble_6.expires,
-                                tumble_6.extra,
-                                tumble_5.price,
-                                tumble_6.category,
-                                tumble_5.date_time,
-                                tumble_5.auction,
                                 tumble_5.extra
-                            HAVING
-                                FALSE
                         ) AS sq_7
                         FULL JOIN hop(alltypes1, alltypes1.c11, INTERVAL '41', INTERVAL '574') AS hop_8
                             ON sq_7.col_0 = hop_8.c4
@@ -78,51 +64,17 @@ FROM
                                 FALSE
                         )
                     GROUP BY
-                        hop_8.c5,
-                        sq_7.col_1,
-                        hop_8.c4,
-                        hop_8.c1,
-                        hop_8.c8,
-                        hop_8.c6,
-                        hop_8.c11,
                         hop_8.c14
                 )
                 SELECT
-                    DATE '2025-09-27' AS col_0,
-                    SMALLINT '807' AS col_1,
                     TRUE AS col_2
-                FROM
-                    with_4
             ) AS sq_11
                 ON hop_2.c2 = sq_11.col_1
                 AND sq_11.col_2
-        WHERE
-            hop_2.c5 > tumble_3.c6
         GROUP BY
-            tumble_3.c5,
-            tumble_3.c16,
-            hop_2.c3,
-            hop_2.c14,
-            tumble_3.c14,
-            tumble_3.c8,
-            tumble_3.c3,
-            hop_2.c5,
-            tumble_3.c6,
-            tumble_3.c1
-        HAVING
             tumble_3.c1
     ) AS sq_12
         ON tumble_1.c14 = sq_12.col_2
         AND sq_12.col_0
 GROUP BY
-    tumble_1.c13,
-    tumble_1.c2,
-    tumble_1.c16,
-    t_0.c_custkey,
-    t_0.c_address,
-    tumble_1.c15,
-    t_0.c_comment,
-    sq_12.col_0,
-    tumble_1.c7,
-    t_0.c_phone,
     tumble_1.c3;
```

cc @kwannoel do you see any opportunities to make the reduction cleaner or more minimal?

src/tests/sqlsmith/src/bin/reducer.rs
```diff
@@ -28,6 +28,12 @@ enum ReductionStrategy {
     Consecutive,
 }
 
+#[derive(Debug, Clone, ValueEnum)]
+enum ReductionModeArg {
+    PathBased,
+    PassBased,
```
kwannoel: I guess this is the old approach? `PassBased`? We can retain it for now, but eventually we should just keep the `PathBased` one, since I assume it's more effective in shrinking to smaller queries.

cardigan1008: Yes, I kept the old one for now. I'll remove it when the latest one is okay to be merged.

src/tests/sqlsmith/src/sqlreduce/reducer.rs
```diff
@@ -205,4 +240,150 @@ impl Reducer {
...
+                tracing::info!(
+                    "  Trying candidate {} of {} (global #{}): {:?}",
+                    i + 1,
+                    candidates.len(),
+                    candidate_index,
+                    candidate
+                );
+
+                if let Some(new_ast) = apply_reduction_operation(&ast_node, candidate) {
+                    if let Some(new_stmt) = ast_node_to_statement(&new_ast) {
```
kwannoel: Instead of nesting, I prefer something like:

```
let Some(new_ast) = ... else {
    tracing::error!("    Failed to convert reduced AST back to statement");
    continue;
}

let Some(new_stmt) = ast_node_to_statement(&new_ast) else {
    tracing::error!("    Failed to convert reduced AST back to statement");
    continue;
}

let new_sql = new_stmt.to_string();
let new_len = new_sql.len();
...

```

With the shorter branching paths first (error paths), the code will be more readable.

Further we should also fix the log level, so it's more friendly to `grep` through if there's unexpected reductions.

kwannoel: Nit: I also find the extra whitespace in front of the log redundant. Is there any specific reason why we do that?

src/tests/sqlsmith/src/sqlreduce/rules.rs
```diff
@@ -0,0 +1,679 @@
...
+
+use std::collections::HashMap;
+
+use risingwave_sqlparser::ast::*;
+
+use crate::sqlreduce::path::{AstNode, AstPath, PathComponent};
+
+/// Defines what actions can be performed on an AST node during reduction.
+#[derive(Debug, Clone, Default)]
+pub struct ReductionRule {
```
kwannoel, line 28: Do we use reduction rules in this module still for `path` based reduction?

src/tests/sqlsmith/src/sqlreduce/path.rs
```diff
@@ -0,0 +1,1035 @@
...
+
+use risingwave_sqlparser::ast::*;
+
+/// Represents a path component in an AST navigation path.
+/// Components that make up a path through the AST.
+/// Enables precise navigation to any AST node.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum PathComponent {
+    /// Field access by name (e.g., "selection", "projection")
+    Field(String),
```
kwannoel: Instead of using `String`, which is error prone, why don't we use an `enum` instead to map out all variants?

src/tests/sqlsmith/src/sqlreduce/rules.rs
kwannoel: Ignore me, yes we do.

src/tests/sqlsmith/src/sqlreduce/rules.rs
```diff
@@ -0,0 +1,679 @@
...
+            if get_node_at_path(root, &attr_path).is_none() {
+                tracing::debug!(
+                    "apply_reduction_operation: Attribute '{}' does not exist at path {}",
+                    attr,
+                    display_ast_path(&attr_path)
+                );
+                return None;
+            }
+
+            let result = set_node_at_path(root, &attr_path, None);
```
kwannoel: Why do we need to check if attribute exists first? Don't we already call `get_node_at_path` inside `set_node_at_path`?

kwannoel:
Thanks for the efforts!

src/tests/sqlsmith/src/sqlreduce/reducer.rs
```diff
@@ -205,4 +240,150 @@ impl Reducer {
...
+                                        || "<failed to convert AST to statement>".to_owned()
+                                    ),
+                                new_sql
+                            );
+
+                            seen_queries.insert(new_sql.clone());
+
+                            // Check if the failure is preserved
+                            tracing::info!("    Checking if failure is preserved...");
+                            if self.checker.is_failure_preserved(sql, &new_sql).await {
```
kwannoel: validation step

src/tests/sqlsmith/src/sqlreduce/reducer.rs
cardigan1008: Fixed. Thanks!

src/tests/sqlsmith/src/sqlreduce/rules.rs
cardigan1008: You're right, I've removed it. Thanks!

src/tests/sqlsmith/src/sqlreduce/path.rs
cardigan1008: I've refactored it with enum. Thanks!

src/tests/sqlsmith/src/sqlreduce/path.rs
```diff
@@ -0,0 +1,1302 @@
...
+            // CTE fields
+            "cte_inner" => Some(AstField::CteInner),
+
+            // TableFactor fields
+            "lateral" => Some(AstField::Lateral),
+
+            _ => None,
+        }
+    }
+}
```
kwannoel: Not sure why we need these. If it's to transform our parser AST to our sqlsmith AST, we should just do a direct mapping.

kwannoel:
Follow-up after this PR:
- Add a reducer binary (or repurpose the existing one).
- It should take in a query, get the initial error.
- Run the reducer loop on it and spit out the shrunk query.

src/tests/sqlsmith/src/sqlreduce/path.rs
kwannoel: Let's address this and merge this PR?

cardigan1008: Fixed. Also remove the previous pass-based method.

PR #13290, opened 2023-11-07
fix: array contains does not support nested types

thexiay:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://gist.github.com/TennyZhuang/f00be7f16996ea48effb049aa7be4d66#file-rw_cla).

## Checklist

- [x] I have written necessary rustdoc comments
- [x] I have added necessary unit tests and integration tests
- [ ] I have added fuzzing tests or opened an issue to track them. (Optional, recommended for new SQL features #7934).
- [ ] My PR contains breaking changes. (If it deprecates some features, please create a tracking issue to remove them in the future).
- [x] All checks passed in `./risedev check` (or alias, `./risedev c`)
- [ ] My PR changes performance-critical code. (Please run macro/micro-benchmarks and show the results.)

- [ ] My PR contains critical fixes that are necessary to be merged into the latest release. (Please check out the [details](https://github.com/risingwavelabs/risingwave/blob/main/CONTRIBUTING.md))

## Documentation

- [ ] My PR needs documentation updates. (Please use the **Release note** section below to summarize the impact on users)

## Release note

If this PR includes changes that directly affect users or other significant modifications relevant to the community, kindly draft a release note to provide a concise summary of these changes. Please prioritize highlighting the impact these changes will have on users.

</details>

PR #23370, opened 2025-10-06
feat: implement array_intersect

PierreNowak:

**Implement array_intersect function**

This PR implements the `array_intersect` function that returns the intersection of two arrays, containing only elements that appear in both input arrays.

**How it works:**
- Takes two array inputs and returns a new array containing elements present in both
- Uses a HashSet-based approach for efficient O(n+m) time complexity
- Preserves element order from the first array
- Handles null values appropriately
- Supports all array element types that implement Hash + Eq

**Implementation details:**
- Added the function signature and registration in the scalar function registry
- Implemented the core logic using HashSet for fast lookups
- Properly handles ListArray/ListRef types and data type propagation
- Returns elements in the order they appear in the first array
- Deduplicates results (each element appears at most once in output)

**Example usage:**
```sql
SELECT array_intersect(ARRAY[1,2,3,4], ARRAY[3,4,5,6]); -- Returns [3,4]
SELECT array_intersect(ARRAY['a','b','c'], ARRAY['b','c','d']); -- Returns ['b','c']
```

## Checklist

- [ ]  I have checked the Release Timeline and Currently Supported Versions to determine which release branches I need to cherry-pick this PR into.

## Documentation

<details>

**New Feature: array_intersect function**

Added support for the `array_intersect(array1, array2)` function that returns the intersection of two arrays. The function returns a new array containing only elements that appear in both input arrays, preserving the order from the first array.

This function is useful for finding common elements between arrays in SQL queries and supports all comparable data types.

</details>

### DIFF
--- a/src/frontend/src/binder/expr/function/builtin_scalar.rs
+++ b/src/frontend/src/binder/expr/function/builtin_scalar.rs
@@ -342,6 +342,7 @@ impl Binder {
                     }
                     Ok(FunctionCall::new_unchecked(ExprType::ArrayFlatten, vec![input], return_type).into())
                 })),
+                ("array_intersect", raw_call(ExprType::ArrayIntersect)),
                 ("trim_array", raw_call(ExprType::TrimArray)),
                 (
                     "array_ndims",
--- a/proto/expr.proto
+++ b/proto/expr.proto
@@ -239,6 +239,7 @@ message ExprNode {
     ARRAY_CONTAINS = 550;
     ARRAY_CONTAINED = 551;
     ARRAY_FLATTEN = 552;
+    ARRAY_INTERSECT = 553;
 
     // Int256 functions
     HEX_TO_INT256 = 560;
--- a/src/expr/impl/src/scalar/array_intersect.rs
+++ b/src/expr/impl/src/scalar/array_intersect.rs
@@ -1,0 +1,126 @@
+// Copyright 2025 RisingWave Labs
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::collections::HashSet;
+
+use risingwave_common::array::{ListRef, ListValue};
+use risingwave_expr::function;
+
+/// Returns the intersection of two arrays, i.e., elements that appear in both arrays.
+/// The result preserves the order of elements from the first array and removes duplicates.
+///
+/// ```sql
+/// array_intersect (array1 anyarray, array2 anyarray) → array
+/// ```
+///
+/// Examples:
+///
+/// ```slt
+/// query T
+/// select array_intersect(array[1,2,3], array[2,3,4]);
+/// ----
+/// {2,3}
+///
+/// query T
+/// select array_intersect(array[1,2,2,3], array[2,3,4]);
+/// ----
+/// {2,3}
+///
+/// query T
+/// select array_intersect(array[1,2,3], array[4,5,6]);
+/// ----
+/// {}
+///
+/// query T
+/// select array_intersect(array[1,2,NULL], array[2,NULL,4]);
+/// ----
+/// {2,NULL}
+///
+/// query T
+/// select array_intersect(null::int[], array[1,2,3]);
+/// ----
+/// NULL
+///
+/// query T
+/// select array_intersect(array[1,2,3], null::int[]);
+/// ----
+/// NULL
+/// ```
+
+#[function("array_intersect(anyarray, anyarray) -> anyarray")]
+fn array_intersect(left: ListRef<'_>, right: ListRef<'_>) -> ListValue {
+    let right_set: HashSet<_> = right.iter().collect();
+    let mut seen = HashSet::new();
+    let mut result = Vec::new();
+
+    for item in left.iter() {
+        if right_set.contains(&item) && !seen.contains(&item) {
+            result.push(item);
+            seen.insert(item);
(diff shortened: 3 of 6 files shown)


src/frontend/src/binder/expr/binary_op.rs
```diff
@@ -137,6 +137,23 @@ impl Binder {
                     }
                 }
             }
+            BinaryOperator::Custom(name) if name == "&&" => {
```
kwannoel, line 140: @xiangjinwu when should we use `Custom` vs directly add the operator to the AST? I suppose when there's no english name for the operator? `&&` for instance doesn't have an english name.

xiangjinwu: > when should we use Custom vs directly add the operator to the AST? I suppose when there's no english name for the operator?

No. The sqlparser only needs to recognize an operator when there are special parsing rules, for example precedence or syntax conflicts, related to it. Everything else is treated equally in sqlparser and dispatched only in binder - this is similar to regular function call `substr(arg0, arg1)` (`Expr::Function`) vs special syntax `substring(arg0 FROM arg1)` (`Expr::Substring`)

And for PostgreSQL compatibility, all operators that need special treatment have already been defined in AST. All other extensions shall use the default rules and we shall never add more special ones.

See #22362 for more

xiangjinwu:
The PR description is off. `array_intersect` is not a function in PostgreSQL 18. The closest is `int[] & int[]` (or `_int_inter`) found in the `intarray` extension. But I agree having this `array_intersect` is beneficial.

PierreNowak:
> The PR description is off. `array_intersect` is not a function in PostgreSQL 18. The closest is `int[] & int[]` (or `_int_inter`) found in the `intarray` extension. But I agree having this `array_intersect` is beneficial.

indeed, I simplified the description

src/frontend/src/binder/expr/binary_op.rs
```diff
@@ -137,6 +137,23 @@ impl Binder {
...
+                let left_type = (!bound_left.is_untyped()).then(|| bound_left.return_type());
+                let right_type = (!bound_right.is_untyped()).then(|| bound_right.return_type());
+                match (left_type, right_type) {
+                    (Some(DataType::List { .. }), Some(DataType::List { .. }))
```
kwannoel: It doesn't seem like we validate the different array types during binding or evaluation. So if the types are different, we will just return false. That differs from postgres' behaviour which returns an error.

In postgres we have:
```
metadata=# select array['a'] && array[1];
ERROR:  operator does not exist: text[] && integer[]
LINE 1: select array['a'] && array[1];
                          ^
HINT:  No operator matches the given name and argument types. You might need to add explicit type casts.
metadata=# select array[1::bigint] && array[1::int];
ERROR:  operator does not exist: bigint[] && integer[]
LINE 1: select array[1::bigint] && array[1::int];
```

We need to add more test coverage to cover these different permutations. You can checkout the `array.*.slt` files in `e2e_test` folder.

src/frontend/src/binder/expr/binary_op.rs
```diff
@@ -137,6 +137,35 @@ impl Binder {
...
+                        // Validate that both arrays have the same element type
+                        if left_elem != right_elem {
+                            return Err(ErrorCode::BindError(format!(
+                                "operator does not exist: {}[] && {}[]\nHint: No operator matches the given name and argument types. You might need to add explicit type casts.",
+                                left_elem,
+                                right_elem,
+                            ))
+                            .into());
+                        }
+                        ExprType::ArrayOverlap
```
xiangjinwu, line 154: Just dispatch to `ExprType::ArrayOverlap` when both sides are array of some elements. The `type_inference` below will be checking the element types, just like `@>` and `<@`.

Furthermore, PostgreSQL allows this:
```
test=# select array[array[1,2,3], array[4,5,6]] && array[9,8,7,6];
 ?column? 
----------
 t
(1 row)
```
That's why the type inference uses `unnest_list` #13290. You may want to update the error message there as it currently says `array_contains`

src/frontend/src/binder/expr/binary_op.rs
PierreNowak: I added some tests

src/frontend/src/binder/expr/function/builtin_scalar.rs
```diff
@@ -342,6 +342,7 @@ impl Binder {
                     }
                     Ok(FunctionCall::new_unchecked(ExprType::ArrayFlatten, vec![input], return_type).into())
                 })),
+                ("array_intersect", raw_call(ExprType::ArrayIntersect)),
```
xiangjinwu, line 345: To avoid conflict with future PostgreSQL function, what about naming this `rw_array_intersect`?

Specifically, we are not sure whether the future function would also:
* consider null element matches a null element
* preserves only 1 copy of duplicate elements
  * preserves the first copy (e.g. 1.00 = 1.0, interval '30' day = interval '1' month)
* preserves the ordering of (first copy in) first array
* checks at the first dimension

<details>
<summary>For reference, existing functions in PostgreSQL have all kinds of behaviors.</summary>

* `&` operator in `intarray` extension:
  * returns error when input array contains null
  * preserves only 1 copy of duplicate elements
    * only supports `int[]`
  * sorts the result elements in ascending order
  * flatten recursively and returns 1d-array
* `<@`/`@>`/`&&`
  * null does NOT match null [#23380](https://github.com/risingwavelabs/risingwave/issues/23380)
  * flatten recursively
* new function `array_sort` in PostgreSQL 18
  * sorts at the first dimension rather than flatten
* `sort` in `intarray` extension:
  * sorts as if flattened but preserves original structure
</details>

src/frontend/src/expr/type_inference/func.rs
```diff
@@ -585,7 +585,7 @@ fn infer_type_for_special(
             }
             Ok(Some(DataType::Varchar))
         }
-        ExprType::ArrayContains | ExprType::ArrayContained => {
+        ExprType::ArrayContains | ExprType::ArrayContained | ExprType::ArrayOverlap => {
```
xiangjinwu, line 588: We shall also ensure both sides of `ArrayIntersect` are of the same array type - either at the first dimension or flatten recursively (see comments above).

PR #23373, opened 2025-10-06
fix(docker): reduce the default memory limit

cyberchen98:

Reduce the memory limit to prevent out of memory failures on machines with limited resources and simplify the first-time `docker-compose up` experience. For example, on a 2c4g EC2 instance:
```
thread 'rw-standalone-compute' panicked at src/compute/src/lib.rs:213:9:
total_memory_bytes 21474836480 is larger than the total memory available bytes 4023242752 that can be acquired.
stack backtrace:
   0: rust_begin_unwind
             at ./rustc/f9e0239a7bc813b4aceffc7f069f4797cde3175c/library/std/src/panicking.rs:695:5
   1: core::panicking::panic_fmt
             at ./rustc/f9e0239a7bc813b4aceffc7f069f4797cde3175c/library/core/src/panicking.rs:75:14
2025-10-06T21:05:57.037424117Z  INFO      rw-standalone-meta sea_orm_migration::migrator: Migration 'm20230908_072257_init' has been applied
2025-10-06T21:05:57.038052345Z  INFO      rw-standalone-meta sea_orm_migration::migrator: Applying migration 'm20231008_020431_hummock'
```

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -22,10 +22,10 @@ services:
                     --advertise-addr 0.0.0.0:5688 \
                     --async-stack-trace verbose \
                     --parallelism 8 \
-                    --total-memory-bytes 21474836480 \
+                    --total-memory-bytes 2147483648 \
                     --role both \
                     --meta-address http://0.0.0.0:5690 \
-                    --memory-manager-target-bytes 22333829939 \" \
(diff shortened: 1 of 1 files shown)


yingjunwu:
actually i think the better way to mention in the doc and let users to configure on their own. cc @hengm3467

PR #20276, opened 2025-01-23
feat(license): invalidate license when cpu exceeds limit instead of rejecting new compute nodes from joining

BugenZhao:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## What's changed and what's your intention?

In #18022, we introduced CPU core limit for license keys, where we would reject new compute nodes from joining the cluster if the limit was exceeded. This led to some confusing behavior like https://github.com/risingwavelabs/risingwave/issues/18401.

In this PR, we modify the behavior so that the license key is temporarily invalidated (thus all `Paid` features unavailable) when the CPU exceeds its limit. This requires the information of "cluster size" is broadcast to all worker nodes. For simplicity, a new notification variant of `compute_node_total_cpu_count` is introduced, instead of maintaining the complete worker node meta data.

The benefit is that, the check for CPU limit is deferred until a paid feature is actually used, and scaling out the cluster won't trigger a hard error any more. If the user isn't utilizing any paid features, they may not even notice this limit. This may enable us to offer a default 'Paid' license key with a small CPU limit in production, helping users better investigate and demonstrate our product.

## Checklist

- [x] I have written necessary rustdoc comments.
- [x]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  My PR contains critical fixes that are necessary to be merged into the latest release. 

## Documentation

- [ ]  My PR needs documentation updates. 

<details>
<summary><b>Release note</b></summary>

</details>

PR #23390, opened 2025-10-09
feat(license):  also count cpu cores for nodes other than compute nodes

BugenZhao:

This PR changes the behavior for CPU core count check in license manager, to include all nodes in the cluster, not just the compute nodes as before. (#18022, #20276)

To avoid double-counting the number of CPU cores, especially for the case of single-node / standalone / playground, we take the **maximum** value of the resources for worker nodes with the same **hostname**, instead of the sum. This should work in most cases.

## Checklist

## Documentation

<details>

Beginning with v2.7, we will count the CPU cores of all nodes in the cluster against the `cpu_core_limit` specified in the license key. Previously, we claimed to count all nodes, but actually we only counted compute nodes. There should be no confusion since this PR.

</details>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4881,7 +4881,7 @@ dependencies = [
  "serde",
  "thiserror 2.0.12 (registry+https://github.com/rust-lang/crates.io-index)",
  "time",
- "winnow 0.7.4",
+ "winnow 0.7.13",
 ]
 
 [[package]]
@@ -7829,9 +7829,9 @@ dependencies = [
 
 [[package]]
 name = "madsim"
-version = "0.2.31"
+version = "0.2.34"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b3c97f34bb19cf6a435a4da2187e90acc6bc59faa730e493b28b6d33e1bb9ccb"
+checksum = "18351aac4194337d6ea9ffbd25b3d1540ecc0754142af1bff5ba7392d1f6f771"
 dependencies = [
  "ahash 0.8.11",
  "async-channel 2.2.1",
@@ -7840,6 +7840,7 @@ dependencies = [
  "bincode 1.3.3",
  "bytes",
  "downcast-rs 1.2.1",
+ "errno",
  "futures-util",
  "lazy_static",
  "libc",
@@ -7853,7 +7854,7 @@ dependencies = [
  "spin 0.9.8",
  "tokio",
  "tokio-util",
- "toml",
+ "toml 0.9.6",
  "tracing",
  "tracing-subscriber",
 ]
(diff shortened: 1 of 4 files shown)


src/meta/src/controller/cluster.rs
```diff
@@ -161,9 +165,7 @@ impl ClusterController {
             )
             .await?;
 
-        if r#type == PbWorkerType::ComputeNode {
-            self.update_compute_node_total_cpu_count().await?;
-        }
+        self.update_cluster_resource_for_license().await?;
```
BugenZhao, line 168: Core changes.

src/meta/src/controller/cluster.rs
```diff
@@ -608,12 +610,28 @@ impl ClusterControllerInner {
...
+        }
+
+        // For different hostnames, we sum up the resources.
+        per_host
+            .into_values()
+            .reduce(|a, b| ClusterResource {
+                total_cpu_cores: a.total_cpu_cores + b.total_cpu_cores,
+                total_memory_bytes: a.total_memory_bytes + b.total_memory_bytes,
+            })
+            .unwrap_or_default()
```
BugenZhao, line 634: Core changes.

src/license/src/manager.rs
```diff
@@ -240,7 +244,7 @@ impl LicenseManager {
         }
 
         // Check the CPU core limit.
```
BugenZhao, line 246: We may extend to check the limit on memory in the future.

src/meta/src/controller/cluster.rs
```diff
@@ -608,12 +610,36 @@ impl ClusterControllerInner {
...
+        let mut per_host = HashMap::new();
+
+        for info in self.worker_extra_info.values() {
+            let r = per_host
+                .entry(info.resource.hostname.as_str())
+                .or_insert_with(ClusterResource::default);
+
+            // Treat nodes with empty hostname as distinct hosts.
+            // TODO: This is currently used as a workaround for madsim.
+            // We should support overriding the hostname in madsim instead.
```
BugenZhao: https://github.com/madsim-rs/madsim/pull/248

BugenZhao:
* **#23403** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23403?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#23390** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23390?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23390?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #23399, opened 2025-10-10
fix(expr): resolve PG incompatibility with NULL handling in array con…

### LINKED ISSUE #23380: PG-incompatibility: NULL in array containment `@>` / `<@` / `&&`
### Describe the bug

_No response_

### Error message/log

```text

```

### To Reproduce

```
select array[1,null,2] @> array[1,null,2];
```

### Expected behavior

PostgreSQL returns false, while RisingWave currently returns true.

_No response_

### The version of RisingWave

_No response_

### Additional context

https://github.com/postgres/postgres/blob/REL_18_0/src/backend/utils/adt/arrayfuncs.c#L4367

Standing-Man:

- Closes #23380.
- Fixed the behavior of array containment (@>, <@) when arrays contain NULL values.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/expr/impl/src/scalar/array_contain.rs
+++ b/src/expr/impl/src/scalar/array_contain.rs
@@ -35,6 +35,26 @@ use risingwave_expr::function;
 /// f
 ///
 /// query I
+/// SELECT array[1,2,3] @> array[3,1];
+/// ----
+/// t
+///
+/// query I
+/// SELECT array[1,2] @> array[1,1];
+/// ----
+/// t
+///
+/// query I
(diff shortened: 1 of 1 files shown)


Standing-Man:
Hi @kwannoel, PTAL. 

Following PostgreSQL behavior: `NULL` elements don’t match any value in `@>` or `<@`. For `left array @> right array`, If the right array contains `NULL`, return false immediately.

https://github.com/postgres/postgres/blob/REL_18_0/src/backend/utils/adt/arrayfuncs.c#L4463C3-L4476C4

PR #23403, opened 2025-10-10
feat(license): change cpu limit to rwu limit, restricting cluster total memory

BugenZhao:

This PR changes the `cpu_core_limit` in license key to `rwu_limit`, meaning that we will not only limit the number of CPU cores, but also the total available memory in the cluster. This makes the license more consistent with our pricing model on Cloud: https://docs.risingwave.com/cloud/pricing#risingwave-unit-rwu

Note that this applies to all existing license keys, because the `cpu_core_limit` field still exists, while it is now interpreted as `rwu_limit` in the code.

## Checklist

## Documentation

The `cpu_core_limit` field in license should be changed to `rwu_limit`. With an `rwu_limit` of X, a cluster may only use X CPU cores **and** up to x * 4 GiB memory in total, otherwise the premium features won't take effect.

### DIFF
--- a/src/license/src/rwu.rs
+++ b/src/license/src/rwu.rs
@@ -93,7 +93,7 @@ mod tests {
             0,
             33 * 4 * GB,
             expect![
-                "feature TestDummy is not available due to license error: a valid license key is set, but it is currently not effective because the memory in the cluster (132 GiB) exceeds the maximum allowed by the license key (128 GiB); consider removing some nodes or acquiring a new license key with a higher limit"
+                "feature TestDummy is not available due to license error: a valid license key is set, but it is currently not effective because the memory in the cluster (132 GiB) exceeds the maximum allowed by the license key (132.00 GiB); consider removing some nodes or acquiring a new license key with a higher limit"
             ],
         );
     }


BugenZhao:
* **#23403** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23403?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23403?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#23390** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23390?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

hzxa21:
IIUC, this will break compatibility and only license issued after this PR will honor the new `rwu_limit`, right?

BugenZhao:
> IIUC, this will break compatibility and only license issued after this PR will honor the new `rwu_limit`, right?

Nope. This will change the semantics of `cpu_core_limit` in license key to additionally enforce memory limit since v2.7, no matter when the license key was generated. More context available here: https://risingwave-labs.slack.com/archives/C07A6KUUB27/p1759447098277369

arkbriar (approved):
LGTM!

PR #23413, opened 2025-10-11
fix(err): Get error score with root cause in anyhow

Chiro11:

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/error.rs
+++ b/src/stream/src/error.rs
@@ -161,14 +161,14 @@ impl StreamError {
 
                 // Uncategorized: try to downcast to known types and reuse their scores. Fallback to 1000.
                 ErrorKind::Uncategorized(e) => {
-                    let root_error = e.root_cause();
-                    if let Some(e) = root_error.downcast_ref::<StreamError>() {
-                        stream_error_score(e)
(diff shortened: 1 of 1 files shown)


src/stream/src/error.rs
```diff
@@ -167,12 +159,20 @@ impl StreamError {
                 // Executor errors first.
                 ErrorKind::Executor(ee) => 2000 + stream_executor_error_score(ee),
 
+                // Uncategorized: try to downcast to known types and reuse their scores. Fallback to 1000.
+                ErrorKind::Uncategorized(e) => {
+                    let root_error = e.root_cause();
```
BugenZhao: IIUC `root_cause` is the inner-most source of the error, meaning that even if an uncategorized error is a wrapper of `StreamExecutorError`, we may still fail to get the `StreamExecutorError` instance here, but the source of it (like `StorageError`).

What about going though the source chain (via `.chain()`) to determine the score based on the first error matching `StreamError` or `StreamExecutorError`?

Chiro11: Good suggestion. Fixed.

PR #23414, opened 2025-10-11
feat: integrated with new iceberg compact interface

Li0k:

Summary
Integrate with new Iceberg compaction interface and improve delete file cleanup strategy.

Changes
Update Iceberg SDK dependency - Integrate with the latest iceberg-rust SDK that includes improved delete file filtering during compaction

Refactor Iceberg compactor runner - Update to use new SDK APIs for better delete file selection and cleanup logic

Fix test expectations - Update iceberg_vacuum.slt to expect equality delete files (content=2) instead of position delete files, aligning with actual RisingWave behavior

Add comprehensive test - New iceberg_compaction_delete_cleanup.slt test to validate delete file cleanup across multiple VACUUM FULL operations

Reduce logging noise - Optimize compactor periodic logging to only print on state changes or every 60s (reduces log volume by ~98%)

Test Coverage
✅ Updated existing vacuum tests
✅ Added new delete cleanup validation test
✅ Verified delete files are properly cleaned up after second VACUUM FULL

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/storage/src/hummock/compactor/iceberg_compaction/mod.rs
+++ b/src/storage/src/hummock/compactor/iceberg_compaction/mod.rs
@@ -23,16 +23,14 @@ use tokio::sync::Notify;
 
 type TaskId = u64;
 
-/// Metadata for a task in the queue.
+/// Task metadata for queue operations.
 ///
-/// The runner (payload) is stored separately in the queue's `HashMap` to allow:
-/// - Tests to create tasks without runners
-/// - Queue operations without moving runner ownership
+/// The actual runner payload is stored separately to support queue operations
+/// without moving ownership, and to allow tests without runner instances.
 #[derive(Debug, Clone)]
 pub struct IcebergTaskMeta {
-    /// Unique task ID assigned by Meta
-    pub task_id: TaskId,
-    /// Parallelism required to execute this task (must be >0 and <= `max_parallelism`)
+    pub task_id: u64,
+    /// Must be in range `1..=max_parallelism`
     pub required_parallelism: u32,
 }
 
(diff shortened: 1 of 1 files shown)


src/storage/src/hummock/compactor/mod.rs
```diff
@@ -396,11 +396,11 @@ pub fn start_iceberg_compactor(
...
-                        tracing::info!(
-                            running_parallelism_count = %running_task_parallelism.load(Ordering::SeqCst),
-                            pull_task_ack = %pull_task_ack,
-                            pending_pull_task_count = %pending_pull_task_count
-                        );
+                        // tracing::info!(
+                        //     running_parallelism_count = %running_task_parallelism.load(Ordering::SeqCst),
+                        //     pull_task_ack = %pull_task_ack,
+                        //     pending_pull_task_count = %pending_pull_task_count
+                        // );
```
chenzl25: Deadcode

src/storage/src/hummock/iceberg_compactor_runner.rs
```diff
@@ -211,185 +403,74 @@ impl IcebergCompactorRunner {
...
-                let _new_table = commit_manager
-                    .overwrite_files(data_files, input_files, MAIN_BRANCH)
-                    .await
-                    .map_err(|e| HummockError::compaction_executor(e.as_report()))?;
+                all_stats.push(plan_stats);
             }
 
-            Ok::<RewriteFilesStat, HummockError>(stats)
+            // Merge all statistics
+            let merged_stats = Self::merge_rewrite_stats(all_stats);
```
chenzl25: It looks like we only merge the stats, but would generate multi snapshots during partial compaction.

Li0k: I plan to keep the logic simple in this PR and first add a Todo doc.

Li0k:
cc @xxhZs @chenzl25

e2e_test/iceberg/test_case/pure_slt/iceberg_vacuum.slt
```diff
@@ -102,11 +102,48 @@ t
 statement ok
 VACUUM FULL t_vacuum_test;
 
+# Check file types remaining after VACUUM FULL
+# content: 0 = data, 1 = position deletes, 2 = equality deletes
```
chenzl25: It seems we can remove this comment now, since we have changed the content type from an int to a string.

src/storage/src/hummock/compactor/mod.rs
```diff
@@ -600,6 +613,11 @@ pub fn start_compactor(
         let mut min_interval = tokio::time::interval(stream_retry_interval);
         let mut periodic_event_interval = tokio::time::interval(periodic_event_update_interval);
 
+        // Track last logged state to avoid duplicate logs
+        let mut last_logged_state: Option<(u32, bool, u32)> = None;
```
chenzl25: The log state looks unmaintainable.

chenzl25: Try to use `LogSuppresser` instead https://github.com/risingwavelabs/risingwave/blob/67cb18b51d3fbf1318f884491d7e5a83495fa7d6/src/common/src/log.rs#L27

chenzl25 (approved):
LGTM

src/storage/src/hummock/compactor/mod.rs
Li0k: A simple rate limit doesn't seem to meet this requirement. Perhaps it can be modified in another PR.

PR #23420, opened 2025-10-13
feat(pg-cdc): allows users to configure whether upstream database is AWS RDS

wcy-fdu:

Allow user to config  `postgres.is.aws.rds = 'true'` to specify that the upstream PG is AWS RDS to strengthen the verification of IsAWSRDS and handle the case of custom DNS. The default value is `false`.

## Checklist

## Documentation

The document needs to add the explanation of this parameter to the PG CDC parameters.  @WanYixian 

<details>

</details>


zwang28:
Do we need [other db].is.aws.rds = 'true' ?

wcy-fdu:
> Do we need [other db].is.aws.rds = 'true' ?

Currently no need.

PR #23425, opened 2025-10-13
fix(frontend): let is_nullable default to true in views

### LINKED ISSUE #22749: View columns are always reported as not nullable
### Describe the bug

_No response_

### Error message/log

```text

```

### To Reproduce

Create tables and views:

```sql
create table has_nulls (x int);
insert into has_nulls values (null);
create view has_nulls_view as select * from has_nulls;
```

Run this select statement:

```sql
select column_name, data_type, is_nullable from information_schema.columns where table_name = 'has_nulls_view';
```

You should get this output:

```
 column_name | data_type | is_nullable
-------------+-----------+-------------
 x           | integer   | NO
(1 row)
```

### Expected behavior

Since these are views, and the computation could be arbitrary (and thus produce nulls), it seems like the output should always be `YES` for the `is_nullable` column.

Docker Compose: https://github.com/ibis-project/ibis/blob/main/compose.yaml#L559

### The version of RisingWave

```
dev=> select version();
                                    version
--------------------------------------------------------------------------------
 PostgreSQL 13.14.0-RisingWave-2.5.0 (e21c37449a133796334467a4910ca60ab0439065)
(1 row)
```

### Additional context

_No response_

howenyap:

Closes #22749. Sets the `is_nullable` field to `true` for views.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/catalog/system_catalog/rw_catalog/rw_columns.rs
+++ b/src/frontend/src/catalog/system_catalog/rw_catalog/rw_columns.rs
@@ -69,7 +69,7 @@ fn read_rw_columns_in_schema(current_user: &UserCatalog, schema: &SchemaCatalog)
                 is_primary_key: false,
                 is_distribution_key: false,
                 is_generated: false,
-                is_nullable: false,
+                is_nullable: true,
                 generation_expression: None,
                 data_type: column.data_type().to_string(),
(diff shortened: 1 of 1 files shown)


howenyap:
@BugenZhao checks are done, could you merge?

PR #23442, opened 2025-10-13
feat(grafana): Add sink and relation lag to developer dashboard

kwannoel:

These metrics are missing from our developer dashboard. They show lag per MV and sink. Using a combination of this and the `"Current Epoch of Materialize Views"` panel, we can identify bottleneck relations.

Screenshot 2025-10-13 at 4 19 31 PM

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/grafana/dashboard/dev/streaming.py
+++ b/grafana/dashboard/dev/streaming.py
@@ -206,6 +206,26 @@ def _(outer_panels: Panels):
                     ],
                 ),
                 panels.timeseries_latency(
+                    "Latency of Materialize Views & Sinks",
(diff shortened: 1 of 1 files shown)


kwannoel:
Tricky parts:
1. This uses global epoch, rather than a local db epoch.
2. This means we likely need to perform a join in prometheus to widen and namespace our relations metrics by db.
3. This means we need to provide db id inside the state table metrics to perform the join.
4. Inside the metrics, we need to namespace by worker id and database id as well.
5. Another idea is to measure the epoch right at the ingress of the mv. This can be done at the stream scan, source scan side. We need to factor in union and joins in such a scenario however.

PR #23446, opened 2025-10-13
fix(binder): report error when defaulting null values to a non-nullable column

### LINKED ISSUE #23378: Insert without `NOT NULL` column should report error
### Describe the bug

When inserting into a table and excluding a `NOT NULL` column, the query succeeds but no data is actually inserted.

### Error message/log

```text

```

### To Reproduce

```
dev=> create table t(id int not null, v varchar not null);
CREATE_TABLE
dev=> insert into t(id,v) values(1,'a');
INSERT 0 1
dev=> insert into t(id) values(1);
INSERT 0 1
dev=> flush;
FLUSH
dev=> select * from t;
 id | v 
----+---
  1 | a
(1 row)
dev=> 
```

### Expected behavior

Expected behavior is to raise an error indicating the missing non-nullable column.

_No response_

### The version of RisingWave

v2.5.1

### Additional context

_No response_

howenyap:

Closes #23378. When insert values default to `null` but the column is not nullable, wrap the expression in a check not null function call, referenced from [here](https://github.com/risingwavelabs/risingwave/blob/main/src/frontend/src/binder/insert.rs#L380-L415).

Attempting to insert an unspecified `null` value into a non-nullable column now reports the error, instead of silently failing.
image

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/binder/insert.rs
+++ b/src/frontend/src/binder/insert.rs
@@ -326,7 +326,7 @@ impl Binder {
                         vec![
                             expr,
                             ExprImpl::literal_varchar(column.name().to_owned()),
-                            ExprImpl::literal_varchar(table_name.to_owned()),
+                            ExprImpl::literal_varchar(table_name.clone()),
                         ],
                         column.data_type().clone(),
                     )


src/frontend/src/binder/insert.rs
```diff
@@ -311,18 +311,31 @@ impl Binder {
...
+                            expr,
+                            ExprImpl::literal_varchar(column.name().to_owned()),
+                            ExprImpl::literal_varchar(table_name.to_owned()),
+                        ],
+                        column.data_type().clone(),
+                    )
+                    .into()
+                };
+
+                Ok((i, expr))
```
kwannoel: Why need to return `Ok(...)` here?

kwannoel: Here we only construct the function call which is pure, it doesn't actually evaluate the function yet. Only later during evaluation of the function call an error will be returned.

howenyap: yeap think I forgot to remove it from an earlier version of the code, will make the change to return a regular tuple instead

howenyap:
not sure why build took so long this time and timed out

BugenZhao:
> not sure why build took so long this time and timed out

I guess it's because we recently updated the toolchain version, causing cache miss during the compilation of many crates. Rebasing the branch and retrying should help.

PR #23447, opened 2025-10-13
refactor: propagate job_id in shared fragment info and related apis

shanicky:

**Summary**
This PR makes the `job_id` field explicit throughout the streaming fragment management flow by propagating it into `SharedFragmentInfo` and all related APIs, eliminating implicit associations between fragments and jobs. This adjustment ensures robust, unambiguous handling of fragment-job relationships across barrier, checkpoint, DDL, and recovery subsystems.

**Details**

- Key change 1: `SharedFragmentInfo` now includes an explicit `job_id` field, and all APIs that construct, mutate, or propagate fragment info require and forward this identifier.
- Key change 2: Conversion patterns and call sites (including upsert, recovery, and mutation functions) are updated to pass `(fragment, job_id)` tuples, standardizing job-aware handling across barrier, checkpoint, and streaming job management logic.
- Key change 3: All internal iteration and notification mechanisms now yield or manage `(fragment, job_id)` pairs, ensuring downstream consumers have direct access to job associations, which is critical for accurate recovery, notification, and DDL operations.

## Checklist

### DIFF
--- a/src/meta/src/barrier/rpc.rs
+++ b/src/meta/src/barrier/rpc.rs
@@ -762,16 +762,15 @@ impl ControlStreamManager {
             database_id,
             database_jobs
                 .values()
+                .chain(
+                    creating_streaming_job_controls
+                        .values()
+                        .map(|job| job.graph_info()),
+                )
                 .flat_map(|info| {
                     info.fragment_infos()
-                        .map(|fragment| (fragment, info.job_id))
-                })
-                .chain(creating_streaming_job_controls.values().flat_map(|job| {
-                    let graph_info = job.graph_info();
-                    graph_info
-                        .fragment_infos()
-                        .map(|fragment| (fragment, graph_info.job_id))
-                })),
+                        .map(move |fragment| (fragment, info.job_id))
+                }),
         );
 
         let committed_epoch = barrier_info.prev_epoch();


src/meta/src/stream/stream_manager.rs
```diff
@@ -204,6 +204,7 @@ impl AutoRefreshSchemaSinkContext {
     pub fn new_fragment_info(&self) -> InflightFragmentInfo {
         InflightFragmentInfo {
             fragment_id: self.new_fragment.fragment_id,
+            job_id: self.tmp_sink_id,
```
shanicky: @wenym1 Can we use this id here as the job_id for the fragment?

yezizp2012: Like `replace job`, here it should be the old id. Or it needs to be updated again in `post_collect`.

wenym1: I think here it should be set with the original job_id. It should be consistent to the job_id in the outer `HashMap<DatabaseId, HashMap<JobId, HashMap<FragmentId, InflightFragmentInfo>>>` that wraps the current fragment.

wenym1: Can we try having this job_id field only in `SharedFragmentInfo` and do not have this field in `InflightFragmentInfo`? After we create a `InflightFragmentInfo`, we must add it to a `InflightStreamingJob`, and we must know the job_id of the fragment. And then the `SharedFragmentInfo` should be created from `(&InflightFragmentInfo, JobId)`, instead of only `&InflightFragmentInfo`.

shanicky: In some cases, job_id might be used since there used to be many requests to find fragments by job_id. However, after several refactors, it seems like only a few less important parts still use job_id. I'll try using the Fragment table instead first.

shanicky: > created from (&InflightFragmentInfo, JobId)

make sense, I'll try it

shanicky: updated

src/meta/src/barrier/rpc.rs
```diff
@@ -760,11 +760,18 @@ impl ControlStreamManager {
 
         self.env.shared_actor_infos().recover_database(
             database_id,
-            database_jobs.values().flatten().chain(
-                creating_streaming_job_controls
-                    .values()
-                    .flat_map(|job| job.graph_info().fragment_infos()),
-            ),
+            database_jobs
```
wenym1, line 763: Can first get an iter of job by `database_jobs.values.chain(creating_streaming_job_controls.values())` and then turn the job iter into fragment iter with a single unified flat_map.

PR #23448, opened 2025-10-13
refactor(batch): store BatchQueryEpoch in batch node

wenym1:

Previously in a batch query, we assign a single `BatchQueryEpoch` for the whole query, with the assumption that the epoch to read different tables should be the same. However, this no longer holds after we introduces time travel query. In batch query nodes that read from a tables (e.g. RowSeqScan, LocalLookupJoin and etc), there is a field `as_of`, and in the batch executors of these nodes, we always first check the `as_of` field, and then the global query_epoch of the query, to determine the epoch used to read from the tables.

In this PR, we will change to specify the query_epoch directly the batch nodes, and then the batch executor don't have to combine with `as_of` to resolve the epoch, but can instead use the query_epoch in the nodes. The query_epoch of the nodes will be resolved in frontend. In the `try_to_batch_prost_body` of `BatchPlanRef`, we will first convert the optional `as_of` into an optional query_epoch. Later before we executor the query, in both local and distributed mode, we will visit all nodes of each stage, to fill in the query_epoch if the query_epoch in the node is None.

The query fragmenter is refactored by the way to remove some unnecessary `Arc` wrappers to allow us more earily in place set the query_epoch of the nodes. We also introduce a strong typed `StageId` to make the code more rigid.

## Checklist

## Documentation

<details>

</details>


wenym1:
* **#23448** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23448?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23448?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

proto/batch_plan.proto
```diff
@@ -28,7 +28,9 @@ message RowSeqScanNode {
 
   // The pushed down `batch_limit`. Max rows needed to return.
   optional uint64 limit = 6;
-  optional plan_common.AsOf as_of = 7;
+  reserved "as_of";
+  reserved 7;
```
BugenZhao: IIUC we never persisted batch plan nodes, so it's safe to remove the `AsOf` directly without reserving field number or name, and reuse the field number for `QueryEpoch`.

src/frontend/src/optimizer/plan_node/utils.rs
```diff
@@ -486,24 +485,31 @@ pub(crate) fn plan_can_use_background_ddl(plan: &StreamPlanRef) -> bool {
     }
 }
 
-pub fn to_pb_time_travel_as_of(a: &Option<AsOf>) -> Result<Option<PbAsOf>> {
+pub fn unix_timestamp_sec_to_epoch(ts: i64) -> risingwave_common::util::epoch::Epoch {
+    let ts = ts.checked_add(1).unwrap();
```
BugenZhao, line 493: Shall we have some comments on why this is necessary? 🤔 @zwang28

zwang28: You mean to clarify time travel query is based on epoch rather than a timestamp?

chenzl25 (approved):
LGTM

src/frontend/src/optimizer/plan_node/utils.rs
BugenZhao: I mean what `+1` is for 🥵

zwang28: I'll add comment in other PR.

PR #23451, opened 2025-10-14
fix(ci): use bitnamilegacy to replayce bitnami

cyliu0:

use bitnamilegacy to replayce bitnami

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/ci/docker-compose.yml
+++ b/ci/docker-compose.yml
@@ -109,7 +109,7 @@ services:
       - KAFKA_CFG_LISTENER_NAME_CONTROLLER_PLAIN_SASL_JAAS_CONFIG=org.apache.kafka.common.security.plain.PlainLoginModule required username="dev" password="rw" user_dev="rw";
 
   message_queue_sasl_2:
-    image: docker.io/bitnami/kafka:4.0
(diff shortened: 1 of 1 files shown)


xiangjinwu:
```
2022-03-17T08:03:54.148353Z ERROR node{id=4 name="frontend-1"}:task{id=710}: pgwire::pg_protocol: error when process message error=Failed to run the query: Failed to bind expression: count((TRIM(TRAILING CAST(NULL AS BYTEA) FROM (CAST(NULL AS BYTEA))))): function rtrim(bytea, bytea) does not exist

thread '<unnamed>' panicked at /risingwave/src/tests/sqlsmith/src/test_runners/fuzzing.rs:37:57:
called `Result::unwrap()` on an `Err` value: Encountered unexpected error: db error: ERROR: Failed to run the query

  1: Failed to bind expression: count((TRIM(TRAILING CAST(NULL AS BYTEA) FROM (CAST(NULL AS BYTEA)))))
  2: function rtrim(bytea, bytea) does not exist
```

`rtrim(bytea, bytea)` is a new function added to main branch. However our fuzzing test suite shares the same set of queries among main and past release branches.

#23452

Issue #23338, opened 2025-09-29
Tracking locality backfill


chenzl25:

chenzl25:
To make it possible to enable the locality backfill by default, we need to support run locality backfill in background. While Backfill order state is currently maintained in memory only and is not persisted, @kwannoel do you have any plan to support persisting the backfill order state? If there were any issues you met previously when you support it, you can share with me as well.

kwannoel:
> To make it possible to enable the locality backfill by default, we need to support run locality backfill in background. While Backfill order state is currently maintained in memory only and is not persisted, [@kwannoel](https://github.com/kwannoel) do you have any plan to support persisting the backfill order state? If there were any issues you met previously when you support it, you can share with me as well.

No blockers. It was previously waiting for background_ddl refactor works to complete. But those should be done now. Let me revisit it.

kwannoel:
> > To make it possible to enable the locality backfill by default, we need to support run locality backfill in background. While Backfill order state is currently maintained in memory only and is not persisted, [@kwannoel](https://github.com/kwannoel) do you have any plan to support persisting the backfill order state? If there were any issues you met previously when you support it, you can share with me as well.
> 
> No blockers. It was previously waiting for background_ddl refactor works to complete. But those should be done now. Let me revisit it.

Tracked here: https://github.com/risingwavelabs/risingwave/issues/24423

PR #23460, opened 2025-10-14
fix(streaming): ban background ddl for locality backfill

chenzl25:

- Since backfill ordering state isn't persisted and it is unrecoverable, so we need to ban background ddl for locality backfill until we persist and can recover all the necessary states.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/proto/stream_plan.proto
+++ b/proto/stream_plan.proto
@@ -1187,6 +1187,9 @@ message BackfillOrder {
 // For the persisted form, see `TableFragments`.
 message StreamFragmentGraph {
   message StreamFragment {
+    reserved 5, 6;
+    reserved "table_ids_cnt", "upstream_table_ids";
+
     // 0-based on frontend, and will be rewritten to global id on meta.
     uint32 fragment_id = 1;
     // root stream node in this fragment.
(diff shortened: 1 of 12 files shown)


BugenZhao:
> * Since backfill ordering state isn't persisted and it is unrecoverable

What will happen if it's recovered from a fresh order state? I suppose there's still no correctness issue but only performance issue?

chenzl25:
> What will happen if it's recovered from a fresh order state? I suppose there's still no correctness issue but only performance issue?

It would be stuck and no progress, since we rely on the `StartFragmentBackfill` to trigger the backfill.

BugenZhao:
> > What will happen if it's recovered from a fresh order state? I suppose there's still no correctness issue but only performance issue?
> 
> It would be stuck and no progress, since we rely on the `StartFragmentBackfill` to trigger the backfill.

What about normal backfill with order control under background DDL? Will it also suffer from this issue?

chenzl25:
> What about normal backfill with order control under background DDL? Will it also suffer from this issue?

For the normal backfill with order control, it will start backfill after recovery of all nodes anyway without guarantee the order.

PR #23467, opened 2025-10-14
refactor: remove source_splits model and related code

shanicky:

The functionality of source_splits has been completely replaced by fragment_splits. Before the new version is released, let's directly remove the models related to source_splits.

**Summary**
This PR removes the entire `source_splits` feature, including its database table, ORM model, controller API methods, and all related update logic in the metadata and source managers. The removal simplifies the system by eliminating persistent tracking and management of source splits.

**Details**
- Deleted the `source_splits` migration and model, fully removing the database schema and ORM entity.
- Removed all controller methods and calls responsible for updating source splits, including `update_source_splits`.
- Cleaned up metadata manager and source manager code that handled source splits persistence and synchronization.
- Removed update logic for source splits in the barrier context, focusing updates on actor and fragment splits only.
- Updated migration registration to exclude the removed source splits migration, preventing it from applying in future deployments.
- Removed all imports and module references to `source_splits` to avoid dead code or compilation errors.
- Reflects a design decision to deprecate persistent source splits storage, likely favoring alternative split management approaches or simplifying the system.

## Checklist

### DIFF
--- a/src/meta/src/stream/source_manager/split_assignment.rs
+++ b/src/meta/src/stream/source_manager/split_assignment.rs
@@ -337,11 +337,6 @@ impl SourceManagerCore {
                     discovered_splits.values().cloned().collect_vec(),
                 );
 
-                source_splits_discovered.insert(
-                    *source_id,
-                    discovered_splits.values().cloned().collect_vec(),
-                );
-
                 if let Some(new_assignment) = reassign_splits(
                     fragment_id,
                     prev_actor_splits,


src/meta/model/migration/src/lib.rs
```diff
@@ -52,7 +52,6 @@ mod m20250819_014448_add_version_column_indices_to_table;
 mod m20250820_120000_add_cdc_table_type;
 mod m20250821_081110_cdc_table_snapshot_splits_add_column;
 mod m20250905_144810_deprecate_table_incoming_sinks;
-mod m20250907_000000_source_splits;
```
tabVersion, line 55: Does this cause inconsistency? Maybe add a new migration to drop the table?
@yezizp2012

tabVersion (approved):
rest LGTM

src/meta/model/migration/src/lib.rs
yezizp2012: Nope, if it's not part of an official release, it won't introduce any inconsistencies.

shanicky: This is not included in any release, so it's safe to delete it as soon as possible.

Issue #21863, opened 2025-05-14
feat: atomically create iceberg engine table 


chenzl25:
Currently, the iceberg table engine creation DDL is not atomic, which might produce a partially created table. We need to create the iceberg table atomically. We can use a dedicated RPC to create these objects in the meta node together.
- Create a hummock table with a connector
- Create an iceberg sink
- Create an iceberg source

chenzl25:
[Iceberg append only table](https://github.com/risingwavelabs/risingwave/pull/21851) depends on this feature.

PR #23468, opened 2025-10-14
feat(iceberg): handle iceberg table DDL creation in meta side

yezizp2012:

Step 1 for transactional iceberg table DDL. Related #21863 .

## Checklist

## Documentation

<details>

</details>


src/frontend/src/session.rs
```diff
@@ -1367,6 +1399,24 @@ impl SessionImpl {
         self.temporary_source_manager.lock().clone()
     }
 
+    pub fn create_staging_table(&self, table: TableCatalog) {
+        self.staging_catalog_manager
+            .lock()
+            .create_table(table.name.clone(), table);
+    }
+
+    pub fn get_staging_table(&self, name: &str) -> Option<TableCatalog> {
```
chenzl25: Where do we call this method? It seems we only create the staging table and drop it.

src/meta/service/src/ddl_service.rs
```diff
@@ -1364,6 +1368,178 @@ impl DdlService for DdlServiceImpl {
...
+                .run_command(DdlCommand::DropStreamingJob {
+                    job_id: StreamingJobId::Table(None, table_id as _),
+                    drop_mode: DropMode::Cascade,
+                })
+                .await
+                .inspect_err(|err| {
+                    tracing::error!(error = %err.as_report(),
+                        "Failed to clean up table after iceberg sink creation failure",
+                    );
+                });
```
chenzl25, line 1488: If the iceberg sink is failed to create, I think we need to drop the iceberg table as well, because in the frontend, we have created the iceberg table.

src/frontend/src/session.rs
yezizp2012: Oh, I cloned the staging catalog manager and used `get_table` directly during binding stage. Let me remove it for now.

src/meta/service/src/ddl_service.rs
yezizp2012: Yes, actually we need to drop the iceberg table at any point the rpc call fails. Let me move this part to the frontend, Although there is still no transaction guarantee, I will try to avoid the logic of pre-create iceberg catalog and migrate it to meta.

chenzl25 (approved):
LGTM!

PR #17568, opened 2024-07-04
feat: compact_noop_update_after_outer_join

st1page:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://gist.github.com/TennyZhuang/f00be7f16996ea48effb049aa7be4d66#file-rw_cla).

## What's changed and what's your intention?

to solve https://github.com/risingwavelabs/risingwave/issues/17450

use the SQL to test if is takes effect.
```SQL
create table t1(v int, k int primary key);
create table d(vv int, k int primary key);
INSERT into t1 VALUES (1,1);
INSERT into d values (1,1);

create materialized view mv as 
select t1.v, t1.k 
from t1 
full outer join d d1 using(k)
full outer join d d2 using(k)
full outer join d d3 using(k)
full outer join d d4 using(k)
full outer join d d5 using(k)
full outer join d d6 using(k)
full outer join d d7 using(k)
full outer join d d8 using(k)
full outer join d d9 using(k)
full outer join d d10 using(k);

UPDATE t1 set v = 2; 
```
before
image
after
image

## Checklist

- [ ] I have written necessary rustdoc comments
- [ ] I have added necessary unit tests and integration tests
- [ ] I have added test labels as necessary. See [details](https://github.com/risingwavelabs/risingwave/blob/main/docs/developer-guide.md#ci-labels-guide).
- [ ] I have added fuzzing tests or opened an issue to track them. (Optional, recommended for new SQL features #7934).
- [ ] My PR contains breaking changes. (If it deprecates some features, please create a tracking issue to remove them in the future).
- [ ] All checks passed in `./risedev check` (or alias, `./risedev c`)
- [ ] My PR changes performance-critical code. (Please run macro/micro-benchmarks and show the results.)

- [ ] My PR contains critical fixes that are necessary to be merged into the latest release. (Please check out the [details](https://github.com/risingwavelabs/risingwave/blob/main/CONTRIBUTING.md))

## Documentation

- [ ] My PR needs documentation updates. (Please use the **Release note** section below to summarize the impact on users)

## Release note

If this PR includes changes that directly affect users or other significant modifications relevant to the community, kindly draft a release note to provide a concise summary of these changes. Please prioritize highlighting the impact these changes will have on users.

</details>

PR #23463, opened 2025-10-14
refactor(streaming): assert stream key consistency for `U-` `U+` pairs

BugenZhao:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## What's changed and what's your intention?

Similar to https://github.com/risingwavelabs/risingwave/pull/12557, we add a debug assertion that `Update` row pairs must have the same stream key value.

It seems that the only violation is outer join, which was fixed in #23472. So all end-to-end tests can pass now.

## Checklist

- [x] I have written necessary rustdoc comments.
- [x]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  I have checked the [Release Timeline](https://github.com/risingwavelabs/rw-commits-history/blob/main/release_timeline.md) and [Currently Supported Versions](https://docs.risingwave.com/changelog/release-support-policy#support-end-dates-for-recent-releases) to determine which release branches I need to cherry-pick this PR into. 

## Documentation

- [ ]  My PR needs documentation updates. 

<details>
<summary><b>Release note</b></summary>

</details>

PR #23472, opened 2025-10-15
fix(streaming): use `Delete`/`Insert` instead of `UpdateDelete`/`UpdateInsert` for outer joins

BugenZhao:

In streaming hash-join when matching a row with zero degree, we need to first retract the `NULL` row before yielding the new matched row. Previously, such changes will be associated with `UpdateDelete` and `UpdateInsert` operations.

However, since the stream key of the join output consists of the stream key of both sides, these 2 rows actually have different stream keys, which violates https://github.com/risingwavelabs/risingwave/issues/12539.

This PR fixes this by using plain `Delete` and `Insert` operations instead. Correspondingly, the no-op update elimination logic introduced in #17568 needs to be updated too. It appears that it can be simply replaced with an enhanced version of `StreamChunk::eliminate_adjacent_noop_update` introduced in #23483.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/executor/asof_join.rs
+++ b/src/stream/src/executor/asof_join.rs
@@ -1286,8 +1286,8 @@ mod tests {
             chunk,
             StreamChunk::from_pretty(
                 " I I I I I I
-                + 3 8 1 . . .
-                - 3 8 1 . . ."
+                + 3 8 1 . . . D
+                - 3 8 1 . . . D"
             )
         );
 
--- a/src/stream/src/executor/hash_join.rs
+++ b/src/stream/src/executor/hash_join.rs
@@ -2395,8 +2395,8 @@ mod tests {
             chunk,
             StreamChunk::from_pretty(
                 "  I I
-                 + 3 8
-                 - 3 8",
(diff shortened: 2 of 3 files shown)


BugenZhao:
* **#23463** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23463?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#23472** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23472?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23472?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#23483** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23483?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

chenzl25 (approved):
LGTM

PR #16087, opened 2024-04-02
perf(memory): use thread-local sequence-based memory eviction policy

MrCroxx:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://gist.github.com/TennyZhuang/f00be7f16996ea48effb049aa7be4d66#file-rw_cla).

### Motivation

Resolves #15305

The previous memory eviction strategy was based on epoch, and in the following scenarios, there may be cases of excessively aggressive eviction:

1. Uneven data volume between epochs caused by sudden increase or decrease in traffic.
2. Large data volume in recent epochs due to frequent access to certain data.
3. Large epoch interval.

This PR introduces a sequence-based memory eviction mechanism. The eviction of the cache is no longer based on epochs but on the sequence of cache access allocation with finer granularity.

Because the sequence needs to be globally shared, although only one atomic variable is needed for the sequence, the overhead of cache invalidation caused by frequent `fetch_add` cannot be ignored. Therefore, this PR introduces a `thread_local` sequence that allows global reordering within a certain range.

When insert/access an entry into/from the managed lru cache, the managed lru cache acquires a sequence from the `Sequencer`. The `Sequencer` use a thread-local variable to grant the sequence. the thread-local variable is synchronized with the global sequence if (a) the pre-allocated local sequences (`step`) are exhausted, or (b) the local sequence lag is higher than the threshold (`lag`). When evicting, the memory controller calculate the memory ratio to evict and normalize it as watermark sequence with the global sequence. The out-of-order threshold is `max(lag, step)`.

CN node memory (before vs after):

image

### Changes

- Remove forked `lru` dependency. Use customized implemention in `risingwave_common::lru`.
- Add thread-local sequencer implementation in `risingwave_common::sequencer`.
- Add `factor` configuration for each eviction policy to control the eviction speed.

### Configurations

```rust
    #[serde(default = "default::developer::memory_controller_eviction_factor_aggressive")]
    pub memory_controller_eviction_factor_aggressive: f64,

    #[serde(default = "default::developer::memory_controller_eviction_factor_graceful")]
    pub memory_controller_eviction_factor_graceful: f64,

    #[serde(default = "default::developer::memory_controller_eviction_factor_stable")]
    pub memory_controller_eviction_factor_stable: f64,

    #[serde(default = "default::developer::memory_controller_sequence_tls_step")]
    pub memory_controller_sequence_tls_step: u64,

    #[serde(default = "default::developer::memory_controller_sequence_tls_lag")]
[...]

PR #23473, opened 2025-10-15
chore: bump toolchain to nightly-20251010

wenym1:

Source of code changes:
1. Our forked `lru-rs` can no longer compiler in the latest toolchain, since it uses `Box::into_raw`, but in the latest toolchain the method has been disabled for `Box` with customized `Allocator` other than `Global`. Though it's easy to make it compile again, but it seems that the forked crate is only used in bench and rpc_client as a container for client, and therefore we can directly use the official `lru-rs`.
2. `hashbrown 0.14.5` can not cleanly compiled under nightly toolchain without enabling the `nightly` feature. Previous the feature is enabled via the forked `lru-rs`. As we are removing the dependency on the forked crate, we should enable the feature in our workspace explicitly.
3. Our previous enabled `string_to_string` has been deprecated, and the compiler indicates that we should instead enable the superset of it `implicit_clone`. Most code changes are caused by enabling this lint rule.
4. Removed some explicitly enabled feature that has newly become stable.
5. supress newly generated warning.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/src/sink/encoder/proto.rs
+++ b/src/connector/src/sink/encoder/proto.rs
@@ -175,7 +175,7 @@ impl MessageIndexes {
 /// * For `validate`, the inputs are (RisingWave type, ProtoBuf type).
 /// * For `encode`, the inputs are (RisingWave type, RisingWave data, ProtoBuf type).
 ///
-/// Thus we impl [`MaybeData`] for both [`()`] and [`ScalarRefImpl`].
+/// Thus we impl [`MaybeData`] for both `()` and [`ScalarRefImpl`].
 trait MaybeData: std::fmt::Debug {
     type Out;
 
--- a/lints/Cargo.lock
+++ b/lints/Cargo.lock
@@ -162,8 +162,8 @@ checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
 
 [[package]]
 name = "clippy_utils"
-version = "0.1.89"
-source = "git+https://github.com/risingwavelabs/clippy?rev=ff5af2fc22b4bb813fc7af028f99f58ecea22cd4#ff5af2fc22b4bb813fc7af028f99f58ecea22cd4"
+version = "0.1.92"
+source = "git+https://github.com/risingwavelabs/clippy?rev=0b9b812ddd1b49a581ae865afde73e7d6365eb8c#0b9b812ddd1b49a581ae865afde73e7d6365eb8c"
 dependencies = [
  "arrayvec",
  "itertools 0.12.0",
--- a/lints/src/utils/format_args_collector.rs
+++ b/lints/src/utils/format_args_collector.rs
@@ -22,7 +22,7 @@ use clippy_utils::source::snippet_opt;
 use itertools::Itertools;
 use rustc_ast::{Crate, Expr, ExprKind, FormatArgs};
 use rustc_data_structures::fx::FxHashMap;
-use rustc_lexer::{TokenKind, tokenize};
+use rustc_lexer::{FrontmatterAllowed, TokenKind, tokenize};
 use rustc_lint::{EarlyContext, EarlyLintPass};
 use rustc_session::impl_lint_pass;
 use rustc_span::{Span, hygiene};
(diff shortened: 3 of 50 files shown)


wenym1:
* **#23490** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23490?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#23473** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23473?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23473?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/connector/src/sink/encoder/proto.rs
```diff
@@ -143,11 +143,11 @@ impl MessageIndexes {
         const TAG_MESSAGE_NESTED: i32 = 3;
 
         let mut indexes = vec![];
-        let mut path = desc.path().array_chunks();
-        let &[tag, idx] = path.next().unwrap();
+        let mut path = desc.path().iter().copied().array_chunks();
```
wenym1: We didn't check remainder.

src/frontend/src/optimizer/optimizer_context.rs
```diff
@@ -91,7 +91,7 @@ impl OptimizerContext {
     /// Create a new [`OptimizerContext`] from the given [`HandlerArgs`] and [`ExplainOptions`].
     pub fn new(mut handler_args: HandlerArgs, explain_options: ExplainOptions) -> Self {
         let session_timezone = RefCell::new(SessionTimezone::new(
-            handler_args.session.config().timezone().to_owned(),
+            handler_args.session.config().timezone().clone(),
```
wenym1: The clone is unnecessary.

src/common/Cargo.toml
```diff
@@ -46,6 +46,7 @@ foyer = { workspace = true }
 futures = { version = "0.3", default-features = false, features = ["alloc"] }
 governor = { workspace = true }
 hashbrown = { workspace = true }
+hashbrown0_14 = { workspace = true }
```
BugenZhao, line 49: What about specifying this in the `workspace-config` crate?

Cargo.toml
```diff
@@ -178,7 +181,7 @@ itertools = "0.14.0"
 jni = { version = "0.21.1", features = ["invocation"] }
 jsonbb = "0.1.4"
 linkme = { version = "0.3.32", features = ["used_linker"] }
-lru = { git = "https://github.com/risingwavelabs/lru-rs.git", rev = "2682b85" }
+lru = "0.12"
```
BugenZhao: It seems we've stopped using the forked crate and have instead vendorred it directly into our repository since version #16087.

wenym1: Yes, we have stopped using it in executor. But as mentioned in the PR description, our rpc_client still use it to hold client, and there is logic of iterating over the lru cache, which is not supported in the one implemented in our repo. This code change is for rpc_client to use the official `lru-rs` crate.

src/common/Cargo.toml
wenym1: Doesn't look like a good idea.

The hashbrown 0.14.5 can only compile in nightly toolchain with its `nightly` feature is enabled. When doing feature unification to resolve the features enabled in each crate, cargo only include the crate usage when a crate include it in the dependency lists and explicitly `use` it. Though we have added the hashbrown 0.14.5 in the `workspace-config` crate, and even have used it with `use hashbrown0_14 as _;` in the `lib.rs`, the `nightly` feature will only be enabled when the crate we are building has included `workspace-config` in the dependency list. However, the `workspace-config` is only depended by `risingwave_cmd`, which means if we build solely on other crates, such as running a test on `risingwave_connector`, the `nightly` feature of hashbrown won't be enabled, and then we cannot pass the compile. 

To resolve this, we will have to directly or indirectly include `workspace-config` in all crates that want to be built independently, and the crate `risingwave_common` is a good common crate that is directly or indirectly depended by most crates. If so, it's better to directly include hashbrown 0.14.5 in `risingwave_common`, like what we do currently in this PR.

BugenZhao: Make sense. This somehow reminds me of the `workspace-hack` crate. I also encountered this in #23453.

Issue #23368, opened 2025-10-06
feat(expr): PostgreSQL 18 new expressions


xiangjinwu:
https://www.postgresql.org/docs/18/release-18.html#RELEASE-18-DATATYPES

* jsonb
  * [x] `jsonb` `null` cast to sql null
  * [ ] `json_strip_nulls` for null array elements
* array
  * [ ] `array_sort ( array anyarray [, descending boolean [, nulls_first boolean ]] ) → anyarray`
  * [x] `array_reverse ( anyarray ) → anyarray`
* bytea
  * [x] `reverse ( bytea ) → bytea`
  * [ ] explicit cast `int2/4/8 <-> bytea`
  * [ ] `crc32 ( bytea ) → bigint`
  * [ ] `crc32c ( bytea ) → bigint`
* unicode text
  * [ ] `casefold ( text ) → text`
* math
  * [ ] `gamma ( double precision ) → double precision`
  * [ ] `lgamma ( double precision ) → double precision`

PR #23478, opened 2025-10-15
feat(expr): add `array_reverse`

howenyap:

Adds `array_reverse` as requested in #23368.

## Checklist

## Documentation

Update to be made [here](https://docs.risingwave.com/sql/functions/array#array-functions-and-operators)

<details>

</details>

### DIFF
--- a/src/expr/impl/src/scalar/array_reverse.rs
+++ b/src/expr/impl/src/scalar/array_reverse.rs
@@ -1,4 +1,4 @@
-// Copyright 2025 RisingWave Labs
+// Copyright 2026 RisingWave Labs
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.


hzxa21 (approved):
Generally LGTM

howenyap:
@hzxa21 are you waiting for a review from @xiangjinwu as well?

e2e_test/batch/aggregate/scalar_as_agg.slt.part
```diff
@@ -7,3 +7,8 @@ query T
 select aggregate:array_sort(v) from (values (3), (2), (1)) as t(v);
 ----
 {1,2,3}
+
+query T
+select aggregate:array_reverse(v) from (values (1), (3), (2)) as t(v);
```
xiangjinwu: This is not a good example of aggregation because the input ordering is unspecified and only deterministic due to the single `values`.

Let's either remove the case here, or update with an explicit `order by`:
```
select aggregate:array_reverse(v order by v) from (values (1), (3), (2)) as t(v);
```

src/expr/impl/src/scalar/array_reverse.rs
```diff
@@ -0,0 +1,21 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use risingwave_common::array::{ListRef, ListValue};
+use risingwave_expr::function;
+
+#[function("array_reverse(anyarray) -> anyarray")]
+fn array_reverse(list: ListRef<'_>) -> ListValue {
+    ListValue::from_datum_iter(&list.elem_type(), list.iter().rev())
```
xiangjinwu: Could you help update to the new writer style #23767? (I know this PR is created before it existed)

howenyap:
thank you for the review @xiangjinwu , fixed! 🙏

xiangjinwu (approved):

Issue #17450, opened 2024-06-25
perf: Outer joins will always multiply their input when join key cardinality is 1.


st1page:
## minimal reproduction case
as title, Due to our current implementation of stream join, which processes each row independently, intermediate NULL results are emitted.
https://github.com/risingwavelabs/risingwave/pull/17449/

## Original case: 
The user wants to use a FULL OUTER JOIN to link together several tables that have the same primary key

```
create table t1(k int primary key, v1 int);
create table t2(k int primary key, v2 int);
create table t3(k int primary key, v3 int);

create materialized view mv as select k, v1, v2, v3 from 
t1
full outer join t2 using(k)
full outer join t3 using(k);
```
Due to this issue, the output throughput of the join operation increases exponentially with the number of joins, leading to performance problems. The user's original query involves dozens or even a hundred tables.

fuyufjh:
Very interesting point! Although I don't have any good idea of solution yet 🤔

- A straight-forward solution is delaying the `NULL` row until the end of epoch. However, if the two join sides come from different sources, there is no guarantee that both sources are synchronized, that is, the joined event may highly likely come in different epochs.
- One step further, just saying, if we allow an option `delay_null_row_for_n_epochs`, it will break the semantic of SQL because some results were deferred. To make it semantic-complete, we should rather do this under `emit on window close` mode.
- So ... it sounds like we should optimize this for `emit on window close`? This sounds good, because it makes OUTER joins to be append-only, too.

lmatz:
In this particular user's case, the events with the same primary key must come in within the same epoch.

The events that get inserted into t1, t2, and t3 respectively originate from the same source event, but via different intermediate MVs.
It's more like
```
create t(k int primary key, v1 int);

create materialized view m1 as select ... from t ...
create materialized view m2 as select ... from t ...
create materialized view m3 as select ... from t ...

create materialized view mv as select k from 
m1
full outer join m2 using(k)
full outer join m3 using(k);
```

Besides, the latency requirement of this user is restricted. 
The expected output must be sent downstream within 60 ms e2e latency, which means output triggered by the arrival of the barrier only is not desirable.

The solution, if possible, is better to offer a knob for the user to tune, similar to the "dirty_heap_size" parameter in the aggregation operator to decide when to output. (I imagine some customized periodic-timer-like stream that can get selected together with the normal data/barrier stream, to avoid the latency due to the barrier not arriving yet and no data coming in either)


Another issue is that:
```
[...]

PR #23483, opened 2025-10-16
feat(streaming): also match `Insert` then `Delete` pairs when eliminating adjacent no-op updates

BugenZhao:

Previously, we only match `Delete` then `Insert` pairs when eliminating adjacent no-op updates. However, it appears that `Insert` then `Delete` pairs are also eligible for elimination.

This is a preparation of https://github.com/risingwavelabs/risingwave/pull/23472.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/common/src/array/stream_chunk.rs
+++ b/src/common/src/array/stream_chunk.rs
@@ -321,7 +321,7 @@ impl StreamChunk {
         }
     }
 
-    /// Remove the adjacent delete-insert if their row value are the same.
+    /// Remove the adjacent delete-insert and insert-deletes if their row value are the same.
     pub fn eliminate_adjacent_noop_update(self) -> Self {
         let len = self.data_chunk().capacity();
         let mut c: StreamChunkMut = self.into();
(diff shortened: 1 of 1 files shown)


BugenZhao:
* **#23463** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23463?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#23472** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23472?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#23483** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23483?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23483?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/common/src/array/stream_chunk.rs
```diff
@@ -330,17 +330,32 @@ impl StreamChunk {
...
-                } else {
-                    prev_r = Some(curr)
-                }
+            if let Some(prev) = prev_r
+                && (
+                    // 1. Delete then Insert
+                    (matches!(c.op(prev), Op::UpdateDelete | Op::Delete)
+                    && matches!(c.op(curr), Op::UpdateInsert | Op::Insert))
+                    ||
+                    // 2. Insert then Delete
```
hzxa21, line 339: This makes sense to me. I wonder why we don't handle this elimination previously.

BugenZhao: I guess by "updating", the first impression comes to our mind is `UpdateDelete` then `UpdateInsert`.

However, there are indeed also cases like https://github.com/risingwavelabs/risingwave/issues/17450 where we first issue an `Insert` of an all-NULL _unmatched_ row in outer join, then immediately retract it. Currently it's handled by another elimination implementation, but it seems that we can make it more general purpose.

PR #23490, opened 2025-10-16
refactor(storage): use new higher rank lifetime assumption to avoid unnecessary clone

wenym1:

Add an experimental rustc flag `"-Zhigher-ranked-assumptions"` to bypass the false positive borrow checker error discussed in https://github.com/rust-lang/rust/issues/100013. The rustc flag is recently enabled after https://github.com/rust-lang/rust/pull/143545 gets merged to the rust toolchain.

After enabling this rustc flag, for the `on_key_value` and `nearest` state store trait method, which pass a closure to observe the slice value, the closure does not have to be static, and in the method caller, we can avoid unnecessary clone to make the closure static.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/ci/scripts/doc.sh
+++ b/ci/scripts/doc.sh
@@ -9,14 +9,14 @@ echo "--- Set openssl static link env vars"
 configure_static_openssl
 
 echo "--- Build documentation"
-RUSTDOCFLAGS="-Dwarnings" cargo doc --document-private-items --no-deps
+RUSTDOCFLAGS="-Dwarnings -Zhigher-ranked-assumptions" cargo doc --document-private-items --no-deps
 
 echo "--- Show sccache stats"
 sccache --show-stats
 sccache --zero-stats
 
 echo "--- Run doctest"
-RUSTDOCFLAGS="-Clink-arg=-fuse-ld=lld" cargo test --doc
+cargo test --doc
 
 echo "--- Show sccache stats"
 sccache --show-stats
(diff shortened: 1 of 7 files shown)


wenym1:
* **#23490** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23490?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23490?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#23473** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23473?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

ci/scripts/doc.sh
```diff
@@ -9,14 +9,14 @@ echo "--- Set openssl static link env vars"
 configure_static_openssl
 
 echo "--- Build documentation"
-RUSTDOCFLAGS="-Dwarnings" cargo doc --document-private-items --no-deps
+RUSTDOCFLAGS="-Dwarnings -Zhigher-ranked-assumptions" cargo doc --document-private-items --no-deps
```
BugenZhao, line 12: Shall we by default to also set `-Dwarnings` in `config.toml`?

wenym1: I don't think this is a good idea. `-Dwarnings` is used here because we want to quickly detect the warning and return with error in CI, but not a general requirement when we build the doc.

PR #23502, opened 2025-10-17
chore: enable `clippy::redundant_clone` in workspace

### LINKED ISSUE #23486: Enable clippy::redundant_clone in workspace

_No response_

See https://rust-lang.github.io/rust-clippy/master/index.html#redundant_clone

_No response_

### Additional context

_No response_

Standing-Man:

- Closes #23486.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/storage/hummock_sdk/src/compaction_group/hummock_version_ext.rs
+++ b/src/storage/hummock_sdk/src/compaction_group/hummock_version_ext.rs
@@ -2275,7 +2275,6 @@ mod tests {
     }
 
     #[test]
-    #[allow(clippy::redundant_clone)]
     fn test_split_sst() {
         let epoch = test_epoch(1);
         let sst = gen_sstable_info(1, vec![1, 2, 3, 5], epoch);
@@ -2349,8 +2348,7 @@ mod tests {
 
         {
             let split_key = group_split::build_split_key(6, VirtualNode::ZERO);
-            let origin_sst = sst.clone();
-            let split_type = group_split::need_to_split(&origin_sst, split_key);
+            let split_type = group_split::need_to_split(&sst, split_key);
             assert_eq!(SstSplitType::Left, split_type);
         }
 
--- a/src/tests/sqlsmith/src/sqlreduce/passes/mod.rs
+++ b/src/tests/sqlsmith/src/sqlreduce/passes/mod.rs
@@ -79,7 +79,6 @@ pub trait Transform: Send + Sync {
     /// - `ast`: The SQL AST to apply the transformation to.
     /// - `idx`: The index of the reduction point to apply the transformation to.
     /// - `strategy`: The strategy to use for applying the transformation.
-    #[allow(clippy::redundant_clone)]
     fn transform(&self, ast: Ast, idx: usize, strategy: Strategy) -> Vec<(Ast, usize)> {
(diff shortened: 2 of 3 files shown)


src/storage/hummock_sdk/src/compaction_group/hummock_version_ext.rs
```diff
@@ -2275,6 +2275,7 @@ mod tests {
     }
 
     #[test]
+    #[allow(clippy::redundant_clone)]
```
wenym1: Any reason to explicitly allow it? Besides, we usually prefer using `expect` to `allow`.

src/tests/sqlsmith/src/sqlreduce/passes/mod.rs
```diff
@@ -79,6 +79,7 @@ pub trait Transform: Send + Sync {
     /// - `ast`: The SQL AST to apply the transformation to.
     /// - `idx`: The index of the reduction point to apply the transformation to.
     /// - `strategy`: The strategy to use for applying the transformation.
+    #[allow(clippy::redundant_clone)]
```
wenym1: Any reason to allow it here?

wenym1: I guess it was for the warning on calling the `apply_on` method?

Actually, I think the trait definition of `fn apply_on(&self, ast: &mut Ast, reduction_points: Vec<usize>) -> Ast` is not good. Can you help to change the parameter `ast` to `ast: &Ast` or `ast: Ast`? You can hava a separate refactor PR before merging the current PR.

wenym1 (commented):
Rest LGTM. Thanks for the contribution!

src/tests/sqlsmith/src/sqlreduce/passes/mod.rs
Standing-Man: yes, i had fixed on PR #23503.

src/storage/hummock_sdk/src/compaction_group/hummock_version_ext.rs
Standing-Man: It’s strange — the compiler reports that line 2352 takes ownership of sst, causing line 2359 to lose ownership of it.

https://github.com/Standing-Man/risingwave/blob/37193111476001e53a41e8844aaa6562a5ac2fe8/src/storage/hummock_sdk/src/compaction_group/hummock_version_ext.rs#L2359

wenym1: I had a try locally. The warning note is not like the common `this value is dropped without further use`. It says `cloned value is neither consumed nor mutated`, which suggests that in the following code,
```
let origin_sst = sst.clone();
let split_type = group_split::need_to_split(&origin_sst, split_key);
```
we can instead directly do 
```
let split_type = group_split::need_to_split(&sst, split_key);
```

Standing-Man: Thanks a lot for the hint! I’ve removed `#[allow(clippy::redundant_clone)]`.

Standing-Man:
Thanks for all careful review and guidance!

PR #23503, opened 2025-10-17
chore: change the parameter `ast` to `ast: Ast` about apply method

Standing-Man:

- Adjusted the parameters of the apply_on function to support the changes introduced in PR #23502.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/tests/sqlsmith/src/sqlreduce/passes/remove.rs
+++ b/src/tests/sqlsmith/src/sqlreduce/passes/remove.rs
@@ -291,7 +291,7 @@ mod tests {
         let reduction_points = OrderByRemove.get_reduction_points(ast[0].clone());
         assert_eq!(reduction_points, vec![2, 1, 0]);
 
-        let new_ast = OrderByRemove.apply_on(ast[0].clone(), &reduction_points[..1].to_vec());
+        let new_ast = OrderByRemove.apply_on(ast[0].clone(), &reduction_points[..1]);
         assert_eq!(
             new_ast,
             parse_sql("SELECT a, b FROM t ORDER BY a, b;")[0].clone()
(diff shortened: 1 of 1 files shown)


Standing-Man:
@wenym1 PTAL when you have time. 😄

src/tests/sqlsmith/src/sqlreduce/passes/pullup.rs
```diff
@@ -54,8 +54,8 @@ impl Transform for BinaryOperatorPullup {
         reduction_points
     }
 
-    fn apply_on(&self, ast: &mut Ast, reduction_points: Vec<usize>) -> Ast {
-        if let Some(query) = extract_query_mut(ast)
+    fn apply_on(&self, ast: Ast, reduction_points: Vec<usize>) -> Ast {
+        if let Some(mut query) = extract_query_mut(ast.clone())
```
wenym1: The change in `extract_query_mut` is incorrect. We want to modify the query of the original ast, however, here the original ast is not changed. Only the intermediate `query` is modified, and then dropped.

The correct should be like
```
fn apply_on(&self, mut ast: Ast, reduction_points: Vec<usize>) -> Ast {
        if let Some(mut query) = extract_query_mut(&mut ast) {
            ...
        }
        ast
```

with `extract_query_mut` unchanged.

src/tests/sqlsmith/src/sqlreduce/passes/mod.rs
```diff
@@ -71,7 +71,7 @@ pub trait Transform: Send + Sync {
     ///
     /// # Returns
     /// - The modified AST.
-    fn apply_on(&self, ast: &mut Ast, reduction_points: Vec<usize>) -> Ast;
+    fn apply_on(&self, ast: Ast, reduction_points: Vec<usize>) -> Ast;
```
wenym1:
```suggestion
    fn apply_on(&self, ast: Ast, reduction_points: &[usize]) -> Ast;
```

We can even pass a slice of `reduction_points` to avoid the unnecessary `.to_vec` in the caller.

Standing-Man: Got it! I’ve fixed it in the new commit.

src/tests/sqlsmith/src/sqlreduce/passes/pullup.rs
Standing-Man: Thanks for the suggestion very much! I learned that `mut ast: Ast` takes ownership and allows mutation, whereas `ast: &mut Ast` only mutates temporarily without ownership.

wenym1 (approved):

PR #23504, opened 2025-10-17
refactor: make batch posix source follow new stream plan

tabVersion:

following https://github.com/risingwavelabs/risingwave/pull/23487

all refreshable source shall follow the new plan (list -> fetch)

**changes the default behavior: the refresh-able table does not ingest at creation**

## Checklist

## Documentation

<details>

</details>


src/connector/src/source/filesystem/opendal_source/batch_posix_fs_source.rs
```diff
@@ -135,10 +130,10 @@ impl SplitEnumerator for BatchPosixFsEnumerator {
...
 #[derive(Debug)]
 pub struct BatchPosixFsReader {
-    properties: BatchPosixFsProperties,
-    splits: Vec<BatchPosixFsSplit>,
-    parser_config: ParserConfig,
-    source_ctx: SourceContextRef,
+    _properties: BatchPosixFsProperties,
+    _splits: Vec<BatchPosixFsSplit>,
+    _parser_config: ParserConfig,
+    _source_ctx: SourceContextRef,
```
chenzl25: dead code?

src/frontend/src/optimizer/plan_node/generic/source.rs
```diff
@@ -134,7 +134,7 @@ impl Source {
 
     /// The output is [`risingwave_connector::source::filesystem::FsPageItem`] / [`iceberg::scan::FileScanTask`]
     pub fn file_list_node(core: Self) -> Self {
-        let column_catalog = if core.is_iceberg_connector() {
+        let column_catalog = if core.is_iceberg_connector() || core.is_batch_connector() {
```
chenzl25: The following columns are iceberg specific, why batch connector needs them?

src/stream/src/executor/source/batch_source/batch_posix_fs_fetch.rs
```diff
@@ -0,0 +1,433 @@
...
+    file_queue: VecDeque<(String, JsonbVal)>,
+}
+
+/// Fetched data from a file, along with file path for logging
+struct FileData {
+    /// The actual data chunks read from the file
+    chunks: Vec<StreamChunk>,
+
+    /// Path to the data file, used for logging
+    _file_path: String,
```
chenzl25: If we use this field, we should change it to `file_path`.

src/frontend/src/optimizer/plan_node/generic/source.rs
tabVersion: Its metadata is like `(batch task id, batch task info)`, I think all batch task can reuse similar columns.

PR #23507, opened 2025-10-17
refactor(streaming): extract `ChangeBuffer` for chunk compaction

BugenZhao:

We have quite a few versions of similar code logic for key-based streaming change compaction.

- `MemTable`
- `MaterializeCache`
- `TopNStaging`
- `OverWindowExecutor`
- `SinkExecutor` for several different purposes, including the handling of
  * unordered updates
  * user-specified primary key
  * no-op updates
- `ChangeLogValue`

Maintaining them is really a burden. Let's refactor!

----

This PR extracts a new struct named `ChangeBuffer`, unifying the implementation of two methods in `StreamChunkCompactor` with it. Also refine the documentation and comments.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9809,12 +9809,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "prehash"
-version = "1.0.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "04bfa62906ce8d9badf8d1764501640ae7f0bcea3437a209315830e0f73564d1"
-
-[[package]]
 name = "pretty-hex"
 version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -12823,7 +12817,6 @@ dependencies = [
  "multimap 0.10.0",
  "parking_lot 0.12.1",
  "pin-project",
- "prehash",
  "pretty_assertions",
  "prometheus",
  "prost 0.13.4",


BugenZhao:
* **#23508** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23508?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#23507** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23507?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23507?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/stream/src/common/compact_chunk.rs
```diff
@@ -447,17 +298,20 @@ mod tests {
...
-                 - 6 6 9
-                 + 4 9 2
+                "  I I I
                 U- 1 1 1
                 U+ 1 1 2
+                 + 4 9 2
+                 + 2 5 5
+                 - 6 6 9
                  + 2 2 2",
-            )
```
BugenZhao, line 461: The records are the same except for the order appearing in the chunk. This is because we switch to use `IndexMap` to aggregate the changes to better preserve the original order in the chunk.

src/stream/src/common/change_buffer.rs
```diff
@@ -0,0 +1,249 @@
...
+    impl<K> Key for K where K: Eq + std::hash::Hash {}
+
+    pub trait Row: Eq + Default {}
+    impl<R> Row for R where R: Eq + Default {}
+}
+
+/// A buffer that accumulates changes and produce compacted changes.
+#[derive(Debug)]
+pub struct ChangeBuffer<K, R> {
+    // We use an `IndexMap` to preserve the original order of the changes as much as possible.
```
hzxa21, line 65: This sounds right intutively. Just out of curiousity, in which case do we need to preserved the order of the changes across different keys?

BugenZhao: Previously, the output order depends on the implementation of the internal `Map`: some use `BTreeMap` to reorder input with key order, while others use `HashMap` which randomize the records at all. I believe we don't have to do this solely for the sake of correctness, but maintaining the order can enhance observability.

PR #23509, opened 2025-10-17
feat(optimizer): support column prunning for iceberg streaming source

chenzl25:

- Support column pruning for iceberg streaming source.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/executor/source/iceberg_list_executor.rs
+++ b/src/stream/src/executor/source/iceberg_list_executor.rs
@@ -124,7 +124,7 @@ impl<S: StateStore> IcebergListExecutor<S> {
                 .collect::<Vec<_>>()
         });
 
-        tracing::info!("downstream_columns: {:?}", downstream_columns);
+        tracing::debug!("downstream_columns: {:?}", downstream_columns);
 
         yield Message::Barrier(first_barrier);
         let barrier_stream = barrier_to_message_stream(barrier_receiver).boxed();


src/stream/src/executor/source/iceberg_list_executor.rs
```diff
@@ -124,7 +124,7 @@ impl<S: StateStore> IcebergListExecutor<S> {
                 .collect::<Vec<_>>()
         });
 
-        println!("downstream_columns: {:?}", downstream_columns);
+        tracing::info!("downstream_columns: {:?}", downstream_columns);
```
tabVersion: change to debug? seems the log message is too vague for info level.

src/frontend/src/optimizer/plan_node/logical_source.rs
```diff
@@ -239,6 +242,95 @@ impl LogicalSource {
...
+        );
+        assert_eq!(
+            self.core.column_catalog[schema_len - 2].name(),
+            ICEBERG_FILE_POS_COLUMN_NAME
+        );
+        assert_eq!(
+            self.core.column_catalog[schema_len - 1].name(),
+            ROW_ID_COLUMN_NAME
+        );
+        assert_eq!(self.output_row_id_index, Some(self.schema().len() - 1));
```
tabVersion, line 273: here we add very strong check for column orders, maybe we can lower it to check these columns exist.

src/frontend/src/optimizer/plan_node/logical_source.rs
```diff
@@ -239,6 +242,95 @@ impl LogicalSource {
...
+            }
+        }
+
+        // Always append the last 4 columns: [_iceberg_sequence_number, _iceberg_file_path, _iceberg_file_pos, _row_id]
+        source_cols.extend([
+            iceberg_start_idx,
+            iceberg_start_idx + 1,
+            iceberg_start_idx + 2,
+            row_id_idx,
+        ]);
```
tabVersion, line 295: why not do it in one loop 🥵
it adds columns [:-4] and then extend [-4:]

tabVersion (approved):
rest LGTM

src/frontend/src/optimizer/plan_node/logical_source.rs
chenzl25: The column orders matter here, since our executor can only generate these system columns in this order, though it is a bit inflexible.

src/frontend/src/optimizer/plan_node/logical_source.rs
chenzl25: The first  "[:-4]" is from required_cols which is not the same thing. The extend here is to persevere the invariant.

PR #23520, opened 2025-10-21
fix: fix prod compile

Li0k:

## Checklist

## Documentation

<details>

</details>


chenzl25 (approved):

src/meta/src/hummock/manager/compaction/compaction_group_schedule.rs
```diff
@@ -118,7 +118,6 @@ impl HummockManager {
 
         let created_tables = if let Some(created_tables) = created_tables {
             // if the created_tables is provided, use it directly, most for test
-            const { assert!(cfg!(debug_assertions)) };
```
wenym1, line 121: Might keep the assertion, and use `expect` to suppress the clippy warning.

Li0k: Of course, you can commit directly to the branch.

PR #23522, opened 2025-10-21
fix: use counter instead of gauge for error metrics

hzxa21:

Prior to this PR, error metrics are implemented in the following ways:
- Use GAUGE for the metric type.
- Implement a customized collector to report the metric on scrape.
- When the metric value is reported, reset the metric value to 0.

This is due to legacy reasons before we have user-facing alert and structured logging for errors. Back then, we put the error message content in the `error_msg` label, which can easily lead to label explosion in the metric system as well as compute node OOM. Therefore, we adopted the approach I mentioned above. Now the `error_msg` label has already been removed so the original motivation no longer holds.

Moreover, the current approach can lead to data loss because it assumes that on each scrape, the metric value can always be successfully persisted back to the metric system after scraping, which is not true. Some examples:
- Network issues or other failures during scraping
- There are more than one scrapers scraping the metric independently

This PR fixes the problem by using a regular COUNTER type for the error metrics, which also simplified the error reporting logics. 
- New metric names are used to ensure backward compatibility.
- To make sure error rate is accurate when the counter value jumps from NULL -> NOT-NULL and from NOT-NULL -> NULL, the promql is written in a bit tricky way because there is no easy way to do gap filling in promql：
```
(
    sum(irate(user_compute_error_cnt{job=~"$job",instance=~"$node"}[$__rate_interval])) by (error_type, executor_name, fragment_id) 
    or
    sum(user_compute_error_cnt{job=~"$job",instance=~"$node"}) by (error_type, executor_name, fragment_id) * 0 + 0.05
)
unless on(instance, job) ((absent_over_time(user_compute_error_cnt{job=~"$job", instance=~"$node"}[20s])) > 0)
```

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/grafana/dashboard/dev/cluster_essential.py
+++ b/grafana/dashboard/dev/cluster_essential.py
@@ -313,7 +313,7 @@ def _(outer_panels: Panels):
                         "remote storage error {{type}}: {{%s}} @ {{%s}}"
                         % (COMPONENT_LABEL, NODE_LABEL),
                     ),
-                    # We add a small constant 0.05 to make sure that the counter jumps from null to not-null, 
+                    # We add a small constant 0.05 to make sure that the counter jumps from null to not-null,
                     # the line will be flat at y=0.05 instead of disappearing.
                      panels.target(
                         f"sum(irate({metric('user_compute_error_cnt')}[$__rate_interval])) by (error_type, executor_name, fragment_id) or "


hzxa21:
image

BugenZhao:
> Back then, we put the error message content in the `error_msg` label, which can easily lead to label explosion in the metric system as well as compute node OOM. Therefore, we adopted the approach I mentioned above. Now the `error_msg` label has already been removed so the original motivation no longer holds.

IIUC, now we use `error_type` and we believe the cardinality won't be too large?

BugenZhao (approved):
LGTM

PR #23527, opened 2025-10-21
feat(iceberg): support refreshable batch iceberg table

tabVersion:

following #23491 , change source catalog accordingly

## Checklist

## Documentation

<details>

### RisingWave Release Notes — Iceberg Manual Refresh (User-facing)

- **On-demand refresh for Iceberg batch tables**
  - New table option: `refresh_mode = 'FULL_RECOMPUTE'` for `connector = 'iceberg'` tables.
  - New SQL: `REFRESH TABLE <table_name>;` to pull the latest Iceberg snapshot on demand.
  - Keeps default behavior unchanged for existing sources/tables (streaming remains default).

- **Accurate delete handling**
  - Refresh applies Iceberg deletes (PositionDeletes and EqualityDeletes) so query results match the latest table state in Iceberg.
  - You can verify discovered delete files via the `rw_iceberg_files` system catalog.

- **Great for**
  - Cost/control sensitive workloads needing explicit sync points.
  - Batch pipelines that should advance only after upstream commits.

### Quick start

```sql
create table iceberg_batch_table (
  primary key (id)
) with (
  connector = 'iceberg',
  s3.region = 'us-west-2',
  catalog.type = 'storage',
  table.name = 't_refresh',
  database.name = 'public',
  refresh_mode = 'FULL_RECOMPUTE'
);

refresh table iceberg_batch_table;
```

- **Compatibility**: Opt-in only; no changes to existing tables/sources unless `refresh_mode` is set.
- **Scope**: Tested with Iceberg storage catalog and S3-compatible object storage.

</details>

### DIFF
--- a/e2e_test/source_inline/cdc/mysql/test_unsigned_int.slt
+++ b/e2e_test/source_inline/cdc/mysql/test_unsigned_int.slt
@@ -120,9 +120,6 @@ statement ok
 create table rw_unsigned_int_to_signed (
   id bigint,
   unsigned_bigint bigint,
-  unsigned_int int,
-  unsigned_smallint SMALLINT,
-  unsigned_tinyint SMALLINT,
   info VARCHAR,
   primary key (id, unsigned_bigint)
 ) with (
@@ -200,6 +197,34 @@ query TTTTTT
 select id, unsigned_bigint, info from rw_unsigned_int_to_signed where id = 53;
 ----
 53	-492998529773844	debezium
+
+# Test schema check: unsigned int -> int32 should fail
+statement error Incompatible data type of column unsigned_int. MySQL type: int unsigned, RisingWave type: INT32
+create table rw_unsigned_int_to_int32 (
+  id DECIMAL,
+  unsigned_bigint BIGINT,
+  unsigned_int INT,
+  unsigned_tinyint SMALLINT,
+  info VARCHAR,
+  primary key (id, unsigned_bigint)
+) with (
+  snapshot = 'true',
+  snapshot.batch_size = 10,
+) from s_unsigned_int_to_signed table 'risedev.test_unsigned_int';
+
+# Test schema check: unsigned smallint -> smallint should fail
+statement error Incompatible data type of column unsigned_smallint. MySQL type: smallint unsigned, RisingWave type: INT16
+create table rw_unsigned_smallint_to_smallint (
+  id DECIMAL,
+  unsigned_bigint BIGINT,
+  unsigned_smallint SMALLINT,
+  unsigned_tinyint SMALLINT,
+  info VARCHAR,
+  primary key (id, unsigned_bigint)
+) with (
+  snapshot = 'true',
+  snapshot.batch_size = 10,
+) from s_unsigned_int_to_signed table 'risedev.test_unsigned_int';
 
 
 # Clean up
--- a/java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/MySqlValidator.java
+++ b/java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/MySqlValidator.java
@@ -217,10 +217,16 @@ public class MySqlValidator extends DatabaseValidator implements AutoCloseable {
             class ColumnInfo {
                 String dataType;
                 long charMaxLength; // Use long to avoid overflow for 4294967295
+                String columnType; // Full column type including UNSIGNED modifier
 
-                ColumnInfo(String dataType, long charMaxLength) {
+                ColumnInfo(String dataType, long charMaxLength, String columnType) {
                     this.dataType = dataType;
                     this.charMaxLength = charMaxLength;
+                    this.columnType = columnType;
+                }
+
+                boolean isUnsigned() {
+                    return columnType != null && columnType.toLowerCase().contains("unsigned");
                 }
             }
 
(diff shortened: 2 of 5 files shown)


tabVersion:
~~https://github.com/risingwavelabs/risingwave/pull/23601~~ https://github.com/risingwavelabs/risingwave/pull/23671 will help resolve multiple triggering load-finish issue

src/connector/src/source/iceberg/mod.rs
```diff
@@ -662,34 +676,425 @@ pub async fn scan_task_to_chunk(
...
+                    // Only include positions that match the current data file
+                    for idx in 0..record_batch.num_rows() {
+                        if !file_paths.is_null(idx) && !positions.is_null(idx) {
+                            let file_path = file_paths.value(idx);
+                            let pos = positions.value(idx);
+
+                            // Match file path - handle both full paths and basenames
+                            let matched = file_path == data_file_path
+                                || data_file_path.ends_with(file_path)
+                                || file_path.ends_with(&data_file_path);
```
chenzl25: I think we just need `file_path == data_file_path`

src/connector/src/source/iceberg/mod.rs
```diff
@@ -635,23 +641,31 @@ pub struct IcebergScanOpts {
...
         need_seq_num,
         need_file_path_and_pos,
+        handle_delete_files,
     }: IcebergScanOpts,
     metrics: Option<Arc<IcebergScanMetrics>>,
 ) {
+    use std::collections::BTreeSet;
+
+    use risingwave_common::array::arrow::arrow_array_iceberg::Array;
+
```
chenzl25: Move these `use` out this method?

src/connector/src/source/iceberg/mod.rs
```diff
@@ -662,34 +676,425 @@ pub async fn scan_task_to_chunk(
...
+                                    );
+                                }
+                            } else if idx < 3 {
+                                tracing::debug!(
+                                    "Position delete NO match [row {}]: file_path='{}' vs data_file='{}'",
+                                    idx,
+                                    file_path,
+                                    data_file_path
+                                );
+                            }
```
chenzl25: What's this?

src/connector/src/source/iceberg/mod.rs
```diff
@@ -662,34 +676,425 @@ pub async fn scan_task_to_chunk(
...
+                                // Log first 5 matched deletions
+                                tracing::debug!(
+                                    "Row {} matched delete key: {:?}",
+                                    row_idx,
+                                    row_key
+                                );
+                            }
+                        }
+                    }
+                }
```
chenzl25: The log is weird 🥵

src/connector/src/source/iceberg/mod.rs
```diff
@@ -662,34 +673,279 @@ pub async fn scan_task_to_chunk(
...
+            if !equality_ids.is_empty() {
+                let mut delete_key_set: HashSet<Vec<Datum>> = HashSet::new();
+
+                let delete_schema = equality_delete_tasks[0].schema();
+                let delete_name_vec = equality_ids
+                    .iter()
+                    .filter_map(|id| delete_schema.name_by_field_id(*id))
+                    .map(|s| s.to_owned())
+                    .collect_vec();
+                let _ = delete_schema;
```
chenzl25: Dead code

src/connector/src/source/iceberg/mod.rs
```diff
@@ -662,34 +673,279 @@ pub async fn scan_task_to_chunk(
...
+        let equality_delete_tasks: Vec<_> = data_file_scan_task
+            .deletes
+            .iter()
+            .filter(|delete| delete.data_file_content == DataContentType::EqualityDeletes)
+            .cloned()
+            .collect();
+
+        if !equality_delete_tasks.is_empty() {
+            // Get equality_ids from the first delete task (should be consistent across all delete files)
+            let equality_ids = equality_delete_tasks[0].equality_ids.clone();
```
chenzl25: Actually the equality ids must be the same only holds for iceberg tables produced by RisingWave. External engine can write multiple different equality deletes, so I think we should leave a TODO here to improve it in the future PR.

chenzl25 (approved):
Rest LGTM, Thanks!

src/connector/src/source/iceberg/mod.rs
tabVersion: I just refined the logic here. For each eq-delete file, it will have a dedicated delete entry.

chenzl25 (approved):
LGTM!

proto/plan_common.proto
```diff
@@ -300,3 +300,13 @@ enum AdditionalColumnType {
...
   ADDITIONAL_COLUMN_TYPE_PAYLOAD = 8;
 }
+
+message SourceRefreshMode {
+  message SourceRefreshModeStreaming {}
+  message SourceRefreshModeManualTrigger {}
+
+  oneof refresh_mode {
+    SourceRefreshModeStreaming streaming = 1;
+    SourceRefreshModeManualTrigger manual_trigger = 2;
```
chenzl25: If we support auto refresh in the future, do we still use this `SourceRefreshModeManualTrigger`?

chenzl25: Maybe `full_recompute` is more precise than `manual_trigger`

tabVersion: done with renaming: MANUAL_TRIGGER -> FULL_RECOMPUTE

tabVersion:
Leftover impl:

clean the tracker entry when dropping a refreshable table. 
But if there is a process already running before the drop, it is still possible to yield refresh mutations in the pipeline. The current impl has strong assumption on receiving the corresponding table_id must exist in the tracker collection.
If not removed is also not a big deal, the state is small and can be remove when meta restarts.

Issue #20475, opened 2025-02-12
unify behavior for `ALTER SOURCE` and `ALTER TABLE` for schema change


BugenZhao:
Since #20203 (which resolves #12199), users are free to call `[ADD | DROP] COLUMN` for a **table** no matter whether its schema is derived from external system (i.e., schema registry).

However, the behavior is inconsistent when it comes to `ALTER SOURCE`, since they are on different code paths.

- [ ] `ALTER SOURCE DROP COLUMN` is generally unsupported.
- [ ] Only `ALTER SOURCE REFRESH SCHEMA` is supported for schema change on a source with schema registry.
- [ ] We may want to deprecate `ALTER SOURCE FORMAT .. ENCODE .. (...)` (introduced in https://github.com/risingwavelabs/risingwave/pull/14057)

PR #23531, opened 2025-10-21
feat: allow adding nested fields when refreshing schema

edvald-garden:

This enables adding new fields to nested struct/message types on events and running `ALTER SOURCE ... REFRESH SCHEMA` (and related ALTER SOURCE statements). Previously this was only possible if fields were added at the root level of the event structure.

To achieve this, the schema compatibility check was re-implemented. An additional benefit is better error messages when a breaking change is made to an individual field.

See this Slack discussion for some more context: https://risingwave-community.slack.com/archives/C03BW71523T/p1760962872810179

**This is my first PR here so I'd be quite surprised if it's good on the first attempt, so any feedback would be much appreciated :)**

## Checklist

- [X]  I have added necessary unit tests and integration tests.
- [X]  I have added test labels as necessary. 

## Documentation

<details>

</details>


src/frontend/src/handler/alter_source_with_sr.rs
```diff
@@ -70,30 +72,189 @@ fn encode_type_to_encode(from: EncodeType) -> Option<Encode> {
...
+    let mut dropped = Vec::new();
+    let mut modified = Vec::new();
+
+    // Find added and modified columns
+    for new_col in &new_relevant {
+        match old_by_name.get(new_col.name()) {
+            Some(old_col) => {
+                // Column exists in both schemas, check compatibility
+                if let Err(_reason) =
+                    is_protobuf_compatible(old_col.data_type(), new_col.data_type())
```
Chiro11: It seems there is a problem here: you only check whether the fields are compatible with the `Protobuf` protocol in this generic method. However, the upstream schema does not guarantee that it is of `Protobuf` type (ie. for struct fields, it could also be generated by `Avro` or other possible nested types).

In addition, I believe that protobuf compatibility should not be checked within RisingWave, but rather ensured by external systems. Even if protobuf is not compatible, as long as the message itself and the schema are updated synchronously, RisingWave can still parse and compute it. This is because the essence of the protobuf compatibility issue arises from the uniqueness of binary serialization/deserialization, while RisingWave merely parses upstream messages into internal types. It does not concern itself with whether the schema is compatible over time, as long as the message itself is compatible with the schema.

Chiro11: cc @BugenZhao

edvald-garden: I actually think the name of the function is misleading here, which is in and of itself a good thing to fix, I've gone ahead and done that to avoid confusion. The function itself basically just checks based on the the internal DataType types and doesn't consider anything protobuf specific, I just had the compatibility semantics of protofuf in mind when first preparing it.

Question is (and I'm really asking, I'm admittedly in some deep waters here) is whether the semantics of the function hold up? Any potential downstream implications that I'm missing?

BugenZhao:
Hi and thanks for your contribution.

Just want to share some context before we proceed with this PR: adding nested fields is actually supported in `ALTER TABLE .. ALTER COLUMN .. TYPE` or `ALTER TABLE .. REFRESH SCHEMA`, but not for `ALTER SOURCE`. So there's already code logic for determining column type compatibility here:

https://github.com/risingwavelabs/risingwave/blob/995c695d62af7eea52b3024634a7db822080f166/src/frontend/src/handler/create_table/col_id_gen.rs#L74-L118

Would you help to check whether we can directly reuse such utility for `ALTER SOURCE`, and furthermore, unifying the handling path for schema change of `SOURCE` and `TABLE` to ensure they behave the same? There's also a tracking issue for this: https://github.com/risingwavelabs/risingwave/issues/20475

edvald-garden:
Thanks for the added info @BugenZhao! Looks like I'll need to get back to this next week, but happy to look into whether the code you linked can be re-used, at least to some degree.

PR #23533, opened 2025-10-21
fix: simplify adaptive split assignment for NATS sources

shanicky:

Refactor adaptive split logic to use actor count instead of actor IDs. This reduces complexity in split assignment code and clarifies APIs.

- Change fill_adaptive_split to take actor count, not HashSet of IDs
- Remove discovered_source_splits field from SplitState
- Update all relevant function signatures to use actor count (usize)
- Eliminate redundant state tracking for discovered splits
- Improve maintainability and reduce risk of actor ID mismatches

## Checklist


tabVersion (approved):
thanks

PR #23545, opened 2025-10-22
refactor(meta): avoid loading StreamJobFragments in recovery

wenym1:

Previously, we have to load `StreamJobFragments` in recover the `CreateMviewProgressTracker` of background ddl job. However, information used in `StreamJobFragments` is already available in `InflightStreamingJobInfo`. Therefore, it's unnecessary to reload `StreamJobFragments` in recovery. 

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/meta/src/barrier/context/recovery.rs
+++ b/src/meta/src/barrier/context/recovery.rs
@@ -502,7 +502,7 @@ impl GlobalBarrierWorkerContextImpl {
                         .await?;
 
                     let background_jobs = {
-                        let unfinished_jobs = HashMap::new();
+                        let mut unfinished_jobs = HashMap::new();
                         let mut background_jobs = self
                             .list_background_job_progress()
                             .await
@@ -518,7 +518,7 @@ impl GlobalBarrierWorkerContextImpl {
                                     .finish_streaming_job(job_id.table_id as _)
                                     .await?;
                             } else {
-                                background_jobs
+                                unfinished_jobs
                                     .try_insert(*job_id, definition)
                                     .expect("non-duplicate");
                             }


wenym1:
* **#23545** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23545?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23545?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/meta/src/barrier/context/recovery.rs
```diff
@@ -508,20 +502,28 @@ impl GlobalBarrierWorkerContextImpl {
...
+                        for (job_id, job_info) in info.values().flatten() {
+                            let Some(definition) = background_jobs.remove(job_id) else {
+                                continue;
+                            };
+                            if job_info.tracking_progress_actor_ids().is_empty() {
+                                // If there's no tracking actor in the job, we can finish the job directly.
+                                self.metadata_manager
+                                    .catalog_controller
+                                    .finish_streaming_job(job_id.table_id as _)
+                                    .await?;
```
kwannoel, line 519: Noticed that we moved `finish_stream_job` path here, from before `pre_apply_drop_cancel`, and actor scaling. Is that intentional? Not sure if there will be any correctness issues.

wenym1: Yes, it's intentional. Before `pre_apply_drop_cancel`, the `job_info` is not loaded yet, and we have no way to figure out the `tracking_progress_actor_ids `. Previously this information is provided in the `StreamJobFragment` loaded before `pre_apply_drop_cancel` but we have avoided loading it, so we have to move this logic to after `pre_apply_drop_cancel`.

PR #23553, opened 2025-10-23
feat(iceberg): enable exactly once for iceberg sink by default

chenzl25:

- Enable exactly once for iceberg sink by default.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/e2e_test/iceberg/test_case/pure_slt/iceberg_sink.slt
+++ b/e2e_test/iceberg/test_case/pure_slt/iceberg_sink.slt
@@ -45,6 +45,24 @@ CREATE SINK s6 AS select mv6.v1 as v1, mv6.v2 as v2, mv6.v3 as v3 from mv6 WITH 
     table.name='e2e_demo_table',
 );
 
+# test is_exactly_once is true with sink_decouple = false
+statement error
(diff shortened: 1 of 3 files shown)


wcy-fdu:
I may loss some context about it, what's the motivation? Enable exactly once bring some overhead.

chenzl25:
> I may loss some context about it, what's the motivation? Enable exactly once bring some overhead.

It does have some overhead, however, we don't commit frequently, so I think it is acceptable in terms of performance. The motivation is that we support upsert mode and primary key in iceberg sink, however, without exactly once, it is easy to confuse users if there are duplicated rows. Enabling exactly once can be helpful to these scenario.

Li0k (approved):

wcy-fdu:
> > I may loss some context about it, what's the motivation? Enable exactly once bring some overhead.
> 
> It does have some overhead, however, we don't commit frequently, so I think it is acceptable in terms of performance. The motivation is that we support upsert mode and primary key in iceberg sink, however, without exactly once, it is easy to confuse users if there are duplicated rows. Enabling exactly once can be helpful to these scenario.

Sound GTM!

wcy-fdu (approved):

PR #23554, opened 2025-10-23
feat(sqlsmith): optimizations to speed up sqlsmith reduction

cardigan1008:

### Problem
The original reducer executed reduction candidates in a random order and treated all SQL operations equally. However, SQL clauses have semantic dependencies (for example, `GROUP BY` depends on `SELECT`, while `WHERE` is independent). This led to:
- High-risk operations being attempted early
- Frequent invalid candidates
- Slow reduction due to repeated validation failures

### Solution
Introduce a field-aware priority strategy based on SQL clause dependencies.

| Priority | Operation Type | Notes |
|----------|----------------|--------|
| 100 | RemoveListElement | Safest, removes one element from a list |
| 95 | Remove(WHERE) | Independent clause |
| 94 | Remove(ORDER BY) | Only affects result ordering |
| 93 | Remove(LIMIT/OFFSET) | Output modifier |
| 92 | Remove(WITH) | Safe if CTE is not referenced |
| 85 | Remove(HAVING) | Depends on GROUP BY |
| 70 | Remove(GROUP BY) | Dependent clause |
| 65 | Remove(Projection/SELECT items) | May depend on GROUP BY |
| 60 | Remove(FROM) | Fundamental dependency |
| 55 | Replace | Medium risk |
| 40 | Pullup | Structural rewrite |
| 20 | TryNull | High risk, may violate constraints |

This prioritization removes independent clauses first to preserve SQL semantics and reduce validation failures.

---

### Problem
Initial batch reduction used global batching with binary search. Due to AST dependencies across paths, batches frequently failed, leading to no speed improvement, and the feature was removed.

### Solution
Implement scoped batch reduction. Batching is only applied to candidates within the same AST path.

- Group consecutive candidates with identical paths
- Batch size reflects operation risk level
- Automatic fallback to single-candidate mode on any batch failure

| Operation Type | Batch Size |
|----------------|------------|
| RemoveListElement | 8 |
| Remove | 3 |
| Replace | 2 |
| Pullup | 2 |

Example:
For removing 8 projection items:
- Original approach: 8 validations × 35 seconds = 4.7 minutes
- New approach: 1 batch validation × 35 seconds
- If batch fails: fallback to 8 sequential checks (no worse than before)

Expected reduction: from 80 validations to 15–25 per iteration, about 3–5× faster.

---

### Problem
The original reducer used depth-first traversal, reducing inner subqueries before outer queries. This caused unnecessary work, because inner structures could be eliminated entirely when an outer query is reduced later.

### Solution
Sort reduction paths by depth, and process outer (shallower) paths before inner (deeper) ones.

**Example**

Given the query:
```sql
SELECT * FROM (SELECT x FROM t WHERE y > 3)
```

Order of traversal:

| Strategy            | Reduction Order                                      |
|---------------------|------------------------------------------------------|
| Original (DFS)      | inner WHERE → inner SELECT → outer SELECT            |
| New (Outer-first)   | outer SELECT → inner SELECT → inner WHERE            |

By reducing outer query structures first, unnecessary reductions on inner subqueries are avoided since they may be eliminated along with the outer structure.

---

### Problem
Previously, all validation failures during reduction triggered the same recovery procedure: full system bootstrap. This approach did not distinguish between harmless SQL frontend errors and severe backend crashes. As a result, the reducer wasted significant time restarting the system even for simple syntax errors.

### Solution
Introduce error classification based on log analysis and apply different recovery strategies according to error severity.

| Error Type | Detection Method | Recovery Strategy |
|-------------|------------------|-------------------|
| Frontend errors (syntax error, type mismatch, invalid reference) | Pattern matching on frontend error messages | Automatic retry with timeout (no restart) |
| Component panic (compute node/meta node failure) | Search for panic keywords in component logs | Immediate bootstrap recovery |
| Unknown errors (connection lost, timeout) | No panic found, backend state uncertain | Conservative bootstrap recovery |

### Log Scanning Optimization
To reduce overhead, log scanning is optimized as follows:
- Only critical component logs are examined
- Search limited to the latest 500 log lines (`tail -n 500`)
- Log access is skipped if the file has not changed since query execution time
- Time window narrowed to query time ± 10 seconds

This strategy shortens recovery time by avoiding unnecessary system restarts, while still ensuring stability when backend crashes occur.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/tests/sqlsmith/src/sqlreduce/path.rs
+++ b/src/tests/sqlsmith/src/sqlreduce/path.rs
@@ -20,6 +20,7 @@
 use std::fmt;
 
 use risingwave_sqlparser::ast::*;
+use strum::EnumDiscriminants;
 
 /// Represents all possible AST field names that can be navigated.
 /// This provides compile-time safety for field access.
@@ -250,7 +251,9 @@ pub fn display_ast_path(path: &AstPath) -> String {
 /// Represents a node in the AST that can be navigated and modified.
 /// This is a simplified representation focusing on the most commonly
 /// reduced SQL constructs.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, EnumDiscriminants)]
+#[strum_discriminants(derive(strum::Display))]
+#[strum_discriminants(name(AstNodeType))]
 pub enum AstNode {
     Statement(Statement),
     Query(Box<Query>),
@@ -909,29 +912,9 @@ impl AstNode {
 
             _ => {
                 // Add debug logging for unmatched cases
-                let node_type = match self {
-                    AstNode::Statement(_) => "Statement",
-                    AstNode::Query(_) => "Query",
-                    AstNode::Select(_) => "Select",
-                    AstNode::Expr(_) => "Expr",
-                    AstNode::SelectItem(_) => "SelectItem",
-                    AstNode::TableWithJoins(_) => "TableWithJoins",
-                    AstNode::Join(_) => "Join",
-                    AstNode::TableFactor(_) => "TableFactor",
-                    AstNode::OrderByExpr(_) => "OrderByExpr",
-                    AstNode::With(_) => "With",
-                    AstNode::Cte(_) => "Cte",
-                    AstNode::ExprList(_) => "ExprList",
-                    AstNode::SelectItemList(_) => "SelectItemList",
-                    AstNode::TableList(_) => "TableList",
-                    AstNode::JoinList(_) => "JoinList",
-                    AstNode::OrderByList(_) => "OrderByList",
-                    AstNode::CteList(_) => "CteList",
-                    AstNode::Option(_) => "Option",
-                };
                 tracing::debug!(
                     "set_child: No match for {} ({:?}) with component {:?}",
-                    node_type,
+                    AstNodeType::from(self),
                     std::mem::discriminant(self),
                     component
                 );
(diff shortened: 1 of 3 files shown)


cardigan1008:
Still under testing. I will update the performance statistics for both versions soon.

src/tests/sqlsmith/src/sqlreduce/path.rs
```diff
@@ -887,8 +897,29 @@ impl AstNode {
...
+                    AstNode::OrderByExpr(_) => "OrderByExpr",
+                    AstNode::With(_) => "With",
+                    AstNode::Cte(_) => "Cte",
+                    AstNode::ExprList(_) => "ExprList",
+                    AstNode::SelectItemList(_) => "SelectItemList",
+                    AstNode::TableList(_) => "TableList",
+                    AstNode::JoinList(_) => "JoinList",
+                    AstNode::OrderByList(_) => "OrderByList",
+                    AstNode::CteList(_) => "CteList",
+                    AstNode::Option(_) => "Option",
```
kwannoel: Can just use this instead: https://docs.rs/strum/latest/strum/derive.EnumDiscriminants.html to generate names.

kwannoel: We should have some examples inside our codebase.

src/tests/sqlsmith/src/sqlreduce/rules.rs
```diff
@@ -458,10 +486,24 @@ pub fn generate_reduction_candidates(
...
     tracing::debug!(
-        "Generated {} total candidates from {} paths",
+        "Generated {} total candidates from {} paths (before sorting)",
         candidates.len(),
         paths.len()
     );
+
+    // Sort candidates by priority (higher priority first)
+    // This ensures we try low-risk, high-success-rate operations first
+    candidates.sort_by(|a, b| b.operation.priority().cmp(&a.operation.priority()));
```
kwannoel, line 534: Hmm I realized this might not be stable sort. But I suppose it doesn't matter, since we batch based on this set of candidates. And won't recompute it each time a batch fails.

kwannoel (approved):

wenym1 (approved):

PR #23556, opened 2025-10-23
feat(expr): implement gamma/lgamma function

Standing-Man:

- Closes Part of #23368.

   - [x] gamma function
   - [x]  lgamma function
 

## Checklist

## Documentation

<details>

support `gamma` and `lgamma` function

</details>


Standing-Man:
Hi @hzxa21, I’m a bit unsure whether the `gamma` function should be wrapped in the `F64`.

https://github.com/risingwavelabs/risingwave/blob/ccea65953af41c2d8b2dc4763616d296c6e7ba69/src/common/src/types/ordered_float.rs#L812

related issue: https://github.com/rust-lang/rust/issues/99842

hzxa21:
> Hi @hzxa21, I’m a bit unsure whether the `gamma` function should be wrapped in the `F64`.
> 
> https://github.com/risingwavelabs/risingwave/blob/ccea65953af41c2d8b2dc4763616d296c6e7ba69/src/common/src/types/ordered_float.rs#L812
> 
> related issue: [rust-lang/rust#99842](https://github.com/rust-lang/rust/issues/99842)

cc the original author of ordered_float to check @BugenZhao

src/expr/impl/Cargo.toml
```diff
@@ -59,6 +59,7 @@ serde_json = "1"
 sha1 = "0.10"
 sha2 = "0.10"
 smallvec = { workspace = true }
+special = "0.11.4"
```
BugenZhao: Shall we consider using the `gamma` and `ln_gamma` functions from the std library? We can enable the nightly feature for them.

- https://doc.rust-lang.org/stable/std/primitive.f64.html#method.gamma

src/expr/impl/src/scalar/arithmetic_op.rs
```diff
@@ -436,6 +437,175 @@ pub fn decimal_trim_scale(d: Decimal) -> Decimal {
...
+/// SELECT gamma('4'::float8);
+/// ----
+/// 6
+///
+/// query R
+/// SELECT gamma('5'::float8);
+/// ----
+/// 24
+///
+/// statement error
```
BugenZhao: We can include part of the error message to test if it matches.

```suggestion
/// statement error underflow
```

src/expr/impl/src/scalar/arithmetic_op.rs
```diff
@@ -436,6 +437,175 @@ pub fn decimal_trim_scale(d: Decimal) -> Decimal {
...
+pub fn gamma_f64(input: F64) -> Result<F64> {
+    let mut result = input;
+    if input.is_nan() {
+        return Ok(result);
+    } else if input.is_infinite() {
+        if input.is_negative() {
+            return Err(ExprError::InvalidParam {
+                name: "gamma",
+                reason: "value out of range: overflow".into(),
+            });
```
BugenZhao: Shall we use `NumericOverflow` and `NumericUnderflow` instead?

src/utils/workspace-config/Cargo.toml
BugenZhao, line 1: Please keep this file untouched. :)

src/expr/impl/Cargo.toml
BugenZhao: > I’m a bit unsure whether the `gamma` function should be wrapped in the `F64`.

Sure. Let's do this.

Standing-Man: Thank you for your suggestions. I've enabled the nightly feature for the `gamma`/`ln_gamma` function of  `f64` type in the new commit.

src/utils/workspace-config/Cargo.toml
BugenZhao, line 1: bump: Let's revert the changes to this file before getting merged.

Standing-Man: Got it, I’ll revert the changes.

Standing-Man:
Hi @xiangjinwu, PTAL when you have time. 😄

Standing-Man:
@hzxa21, ready to review.

Standing-Man:
Different platforms have different levels of precision, which can cause CI failures.

BugenZhao (approved):
LGTM!

> Different platforms have different levels of precision, which can cause CI failures.

I think it's okay to allow some margin of error when comparing the results, for example, using `ROUND` or rewriting it into `SELECT ABS(actual - expected) <= 1e-6`.

PR #23557, opened 2025-10-23
refactor(sink): use `downstream_pk: None` to represent user not specifying sink pk

BugenZhao:

Previously we use an empty slice to represent, which can be confusing, since

- empty primary key has its own semantics of having at most one row, although it's generally not the case for sink, we'd better make it consistent with other concept in optimizer, e.g., `stream_key` is an `Option` and `ShardByKey` does not accept empty dist key input

- we may assume it's not empty, and call `any` or `all` on it to derive some properties, where the result may be unexpected

This PR refactors it to be an `Option`, while minimizing the changes for concrete sink implementation by introducing a method named `downstream_pk_or_empty()` for compatibility.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/optimizer/plan_node/stream_sink.rs
+++ b/src/frontend/src/optimizer/plan_node/stream_sink.rs
@@ -21,7 +21,6 @@ use pretty_xmlish::{Pretty, XmlNode};
 use risingwave_common::catalog::{ColumnCatalog, CreateType, FieldLike};
 use risingwave_common::types::{DataType, StructType};
 use risingwave_common::util::iter_util::ZipEqDebug;
-use risingwave_connector::match_sink_name_str;
 use risingwave_connector::sink::catalog::desc::SinkDesc;
 use risingwave_connector::sink::catalog::{SinkFormat, SinkFormatDesc, SinkId, SinkType};
 use risingwave_connector::sink::file_sink::fs::FsSink;
@@ -31,6 +30,7 @@ use risingwave_connector::sink::{
     CONNECTOR_TYPE_KEY, SINK_TYPE_APPEND_ONLY, SINK_TYPE_DEBEZIUM, SINK_TYPE_OPTION,
     SINK_TYPE_UPSERT, SINK_USER_FORCE_APPEND_ONLY_OPTION,
 };
+use risingwave_connector::{WithPropertiesExt, match_sink_name_str};
 use risingwave_pb::expr::expr_node::Type;
 use risingwave_pb::stream_plan::SinkLogStoreType;
 use risingwave_pb::stream_plan::stream_node::PbNodeBody;
@@ -258,7 +258,8 @@ impl StreamSink {
         let (pk, _) = derive_pk(input.clone(), user_order_by, &columns);
         let derived_pk = pk.iter().map(|k| k.column_index).collect_vec();
 
-        let mut downstream_pk = {
+        // Get downstream pk from user input, override and perform some checks if applicable.
+        let downstream_pk = {
             let downstream_pk = properties
                 .get(DOWNSTREAM_PK_KEY)
                 .map(|v| Self::parse_downstream_pk(v, &columns))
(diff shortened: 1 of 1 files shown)


BugenZhao:
* **#23581** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23581?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#23558** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23558?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#23557** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23557?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23557?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/stream/src/executor/sink.rs
```diff
@@ -239,7 +236,7 @@ impl<F: LogStoreFactory> SinkExecutor<F> {
             self.chunk_size,
             self.input_data_types,
             input_compact_ib,
-            self.sink_param.downstream_pk.clone(),
+            self.sink_param.downstream_pk_or_empty(),
```
wenym1, line 239: Shall we limit the usage of `downstream_pk_or_empty` only in the concrete sink implementation so that in the outer general sink processing logic we can be aware of downstream-not-specified with the more explicit Option rather than empty Vec?

For example, in current code, we may not call `downstream_pk_or_empty` and instead do all the checks directly on the `Option<Vec<usize>>`.

src/frontend/src/optimizer/plan_node/stream_sink.rs
```diff
@@ -279,23 +282,23 @@ impl StreamSink {
...
-                        .try_collect::<_, _, RwError>()?
+                        .try_collect::<_, _, RwError>()?)
                 }
             } else if properties.get(CREATE_TABLE_IF_NOT_EXISTS) == Some(&"true".to_owned())
                 && sink_type == SinkType::Upsert
-                && downstream_pk.is_empty()
+                && downstream_pk.is_none()
             {
-                derived_pk.clone()
+                Some(derived_pk.clone())
```
wenym1, line 302: Shall we add an assertion here that `!derived_pk.is_empty()`?

src/frontend/src/optimizer/plan_node/stream_sink.rs
```diff
@@ -339,20 +343,20 @@ impl StreamSink {
             _ => {
                 match properties.get("connector") {
                     Some(s) if s == "jdbc" && sink_type == SinkType::Upsert => {
-                        if sink_type == SinkType::Upsert && downstream_pk.is_empty() {
+                        let Some(downstream_pk) = &downstream_pk else {
```
wenym1, line 367: Review note: we have ensure `sink_type == SinkType::Upsert` in the outer match arm.

kwannoel (approved):
Looks good after @wenym1's comments, thanks for the refactor.

src/stream/src/executor/sink.rs
BugenZhao: Agree. This will be addressed in the refactor in the next PR.

src/frontend/src/optimizer/plan_node/stream_sink.rs
BugenZhao: I'm not sure if this always hold. For example, the stream key of `SimpleAgg` is an empty slice. 

https://github.com/risingwavelabs/risingwave/blob/308bcad4cfd00e950dfb357c67257e9341d04ba1/src/frontend/src/optimizer/plan_node/generic/agg.rs#L247-L249

This could also be a reason why we should distinguish between "not specified" and "empty key". It can be imagined that there's some bug under current implementation for this specific case.

BugenZhao: Given that the impls now may not be able to handle empty key well, let's still reject empty derived pk here.

https://github.com/risingwavelabs/risingwave/blob/f0cd41a57c2444863de75656e76f9f46ab858866/src/frontend/src/optimizer/plan_node/stream_sink.rs#L314-L326

PR #23558, opened 2025-10-23
refactor(sink): simplify and refine documentation for compaction

BugenZhao:

The code and comments for chunk compaction in sink executor is messy. We refactor to simplify it, and refine the documentation in this PR.

Also, this PR replaces all `downstream_pk_or_empty` calls in sink executor, as a follow-up of #23557.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/common/compact_chunk.rs
+++ b/src/stream/src/common/compact_chunk.rs
@@ -38,10 +38,6 @@ impl StreamChunkCompactor {
     }
 
     /// Remove unnecessary changes in the given chunks, by modifying the visibility and ops in place.
-    ///
-    /// Compared to [`Self::into_compacted_chunks_reconstructed`], this method does not allocate new
-    /// chunks, thus can be more efficient. However, since the rows cannot be reordered, `Update` pairs
-    /// might be broken into inconsecutive `Delete` and `Insert`.
     pub fn into_compacted_chunks_inline(
         self,
         ib: InconsistencyBehavior,
@@ -88,11 +84,6 @@ impl StreamChunkCompactor {
 
     /// Remove unnecessary changes in the given chunks, by filtering them out and constructing new
     /// chunks, with the given chunk size.
-    ///
-    /// Compared to [`Self::into_compacted_chunks_inline`], this method allocates new chunks, thus
-    /// can be less efficient. However, since the chunk is completely rebuilt, we can have `Delete`
-    /// and `Insert` being rewritten as `Update` pairs as much as possible, which is friendlier to
-    /// downstream executors, especially the sink.
     pub fn into_compacted_chunks_reconstructed(
         self,
         chunk_size: usize,


BugenZhao:
* **#23581** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23581?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#23558** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23558?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23558?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#23557** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23557?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/stream/src/executor/sink.rs
```diff
@@ -123,59 +220,21 @@ impl<F: LogStoreFactory> SinkExecutor<F> {
...
-        // So the delete event is not to delete the inserted record in our internal streaming SQL semantic.
-        let need_advance_delete = stream_key_sink_pk_mismatch
-            && !matches!(
-                sink_param.sink_type,
-                SinkType::AppendOnly | SinkType::ForceAppendOnly
-            );
-        // NOTE(st1page): reconstruct with sink pk need extra cost to buffer a barrier's data, so currently we bind it with mismatch case.
-        let re_construct_with_sink_pk = need_advance_delete
-            && sink_param.sink_type == SinkType::Upsert
-            && !downstream_pk.is_empty();
```
BugenZhao, line 167: `!matches!(
                sink_param.sink_type,
                SinkType::AppendOnly | SinkType::ForceAppendOnly
            );` is actually equivalent to  `sink_param.sink_type == SinkType::Upsert`

So `re_construct_with_sink_pk` is basically `need_advance_delete && downstream_pk.is_some()`, which is intuitive that we can compact based on user key only if it's specified.

src/stream/src/executor/sink.rs
```diff
@@ -447,52 +504,44 @@ impl<F: LogStoreFactory> SinkExecutor<F> {
...
-                            let mut insert_chunks = vec![];
-
-                            for c in StreamChunkCompactor::new(stream_key.clone(), chunks)
-                                .into_compacted_chunks_inline(input_compact_ib)
-                            {
-                                // We only enter the branch if need_advance_delete, in which case `sink_type` is not ForceAppendOnly or AppendOnly.
-                                {
-                                    // Force append-only by dropping UPDATE/DELETE messages. We do this when the
-                                    // user forces the sink to be append-only while it is actually not based on
-                                    // the frontend derivation result.
```
BugenZhao, line 466: This comment is completely out-of-topic. It's copied from elsewhere.

src/stream/src/executor/sink.rs
```diff
@@ -428,17 +483,19 @@ impl<F: LogStoreFactory> SinkExecutor<F> {
...
         input_compact_ib: InconsistencyBehavior,
-        down_stream_pk: Vec<usize>,
+        downstream_pk: Option<Vec<usize>>,
+        upsert_behavior: Option<UpsertBehavior>,
         sink_chunk_buffer_size_metrics: LabelGuardedIntGauge,
-        compact_chunk: bool,
+        skip_compact: bool,
     ) {
-        // need to buffer chunks during one barrier
-        if need_advance_delete || re_construct_with_sink_pk {
```
BugenZhao, line 441: `re_construct_with_sink_pk` is `need_advance_delete && ...`, so this is basically `need_advance_delete`

src/stream/src/executor/sink.rs
```diff
@@ -447,52 +504,44 @@ impl<F: LogStoreFactory> SinkExecutor<F> {
...
+                        //    `DELETE` and `INSERT` operations on the same key into `UPDATE` operations, which
+                        //    usually have more efficient implementation.
+                        if let Some(downstream_pk) = &downstream_pk {
+                            let chunks = StreamChunkCompactor::new(downstream_pk.clone(), chunks)
                                 .into_compacted_chunks_reconstructed(
                                     chunk_size,
                                     input_data_types.clone(),
                                     // When compacting based on user provided primary key, we should never panic
                                     // on inconsistency in case the user provided primary key is not unique.
-                                    if sink_type == SinkType::ForceAppendOnly {
```
BugenZhao, line 491: It must be `Upsert` in this arm.

src/stream/src/executor/sink.rs
```diff
@@ -123,59 +220,21 @@ impl<F: LogStoreFactory> SinkExecutor<F> {
...
-                SinkType::AppendOnly | SinkType::ForceAppendOnly
-            );
-        // NOTE(st1page): reconstruct with sink pk need extra cost to buffer a barrier's data, so currently we bind it with mismatch case.
-        let re_construct_with_sink_pk = need_advance_delete
-            && sink_param.sink_type == SinkType::Upsert
-            && !downstream_pk.is_empty();
-        let pk_matched = !stream_key_sink_pk_mismatch;
-        // Don't compact chunk for blackhole sink for better benchmark performance.
-        let compact_chunk = !sink.is_blackhole();
+        let upsert_behavior = if let SinkType::Upsert = sink_param.sink_type {
```
kwannoel:
```suggestion
        let upsert_behavior = if SinkType::Upsert == sink_param.sink_type {
```

src/stream/src/executor/sink.rs
```diff
@@ -447,52 +505,44 @@ impl<F: LogStoreFactory> SinkExecutor<F> {
...
-                                    if chunk.cardinality() > 0 {
-                                        delete_chunks.push(chunk);
-                                    }
-                                }
-                                let chunk = force_append_only(c);
-                                if chunk.cardinality() > 0 {
-                                    insert_chunks.push(chunk);
-                                }
+
+                        // 1. Compact the chunk based on the **stream key**, so that we have at most 2 rows for each
```
wenym1, line 513: Review note: previously this branch will not be hit when `re_construct_with_sink_pk && !need_advance_delete`.

src/stream/src/executor/sink.rs
```diff
@@ -447,52 +505,44 @@ impl<F: LogStoreFactory> SinkExecutor<F> {
...
+                        let chunks = delete_chunks
+                            .into_iter()
+                            .chain(insert_chunks.into_iter())
+                            .collect();
+
+                        // 2. If user specifies a primary key, compact the chunk based on the **downstream pk**
+                        //    to eliminate any unnecessary updates to external systems. This also rewrites the
+                        //    `DELETE` and `INSERT` operations on the same key into `UPDATE` operations, which
+                        //    usually have more efficient implementation.
+                        if let Some(downstream_pk) = &downstream_pk {
```
wenym1, line 538: Review note: previously this branch will not be hit when `need_advance_delete && !re_construct_with_sink_pk`

src/stream/src/executor/sink.rs
```diff
@@ -428,17 +483,20 @@ impl<F: LogStoreFactory> SinkExecutor<F> {
         input: impl MessageStream,
         sink_type: SinkType,
         stream_key: PkIndices,
-        need_advance_delete: bool,
```
wenym1, line 431: Review note: 

`need_advance_delete = stream_key_sink_pk_mismatch && sink_type == Upsert`

`re_construct_with_sink_pk = stream_key_sink_pk_mismatch && sink_type == Upsert && user specified pk`

src/stream/src/executor/sink.rs
wenym1: Unlikely to happen, because `re_construct_with_sink_pk` implies `need_advance_delete`. So the original logic must be hit, and it's correct to remove the if check.

src/stream/src/executor/sink.rs
wenym1: Will not be hit when `need_advance_delete && user not specified pk`.

src/stream/src/executor/sink.rs
```diff
@@ -428,17 +483,20 @@ impl<F: LogStoreFactory> SinkExecutor<F> {
...
-        down_stream_pk: Vec<usize>,
+        downstream_pk: Option<Vec<usize>>,
+        upsert_behavior: Option<UpsertBehavior>,
         sink_chunk_buffer_size_metrics: LabelGuardedIntGauge,
-        compact_chunk: bool,
+        skip_compact: bool,
     ) {
-        // need to buffer chunks during one barrier
-        if need_advance_delete || re_construct_with_sink_pk {
+        // To reorder records, we need to buffer chunks of the entire epoch.
```
wenym1, line 494: Review note: the previous condition can be simplified to `need_advance_delete`, which is `stream_key_sink_pk_mismatch && sink_type == Upsert`, and the same as current logic.

src/stream/src/executor/sink.rs
wenym1: `need_advance_delete` is the original condition of outer branch, so the additional condition here is `user specified pk`, which is the condition in current logic.

src/stream/src/executor/sink.rs
wenym1: Review note: `need_advance_delete` implies `sink_type == Upsert`.

src/stream/src/executor/sink.rs
```diff
@@ -526,9 +577,8 @@ impl<F: LogStoreFactory> SinkExecutor<F> {
                 match msg? {
                     Message::Watermark(w) => yield Message::Watermark(w),
                     Message::Chunk(mut chunk) => {
-                        // Compact the chunk to eliminate any useless intermediate result (e.g. UPDATE
-                        // V->V).
-                        if compact_chunk {
+                        // Compact the chunk to eliminate any unnecessary updates to external systems.
+                        if !skip_compact {
```
wenym1, line 581: Review note: previous `compact_chunk` is the same as `!skip_compact`, which is also `!sink.is_blackhole()`

src/stream/src/executor/sink.rs
```diff
@@ -594,28 +644,16 @@ impl<F: LogStoreFactory> SinkExecutor<F> {
...
-                // downstream PKs are matched. When the upstream and downstream PKs are not matched,
-                // we will buffer the chunks between two barriers, so the compaction is not needed,
-                // since the barriers will preserve chunk boundaries.
-                //
-                // When the sink is not an upsert sink, it is either `force_append_only` or
-                // `append_only`, we should only append to downstream, so there should not be any
-                // overlapping keys.
-                let chunk = if pk_matched && matches!(sink_param.sink_type, SinkType::Upsert) {
-                    compact_chunk_inline(chunk, &downstream_pk, input_compact_ib)
+                let chunk = if let Some(ub) = upsert_behavior
```
wenym1, line 651: Review note: the original condition is `pk_mismatched && sink_type == Upsert`.

wenym1: `pk_specified_and_matched` implies that `downstream_pk.is_some()`, so it's safe to do unwrap below.

src/stream/src/executor/sink.rs
```diff
@@ -123,59 +220,21 @@ impl<F: LogStoreFactory> SinkExecutor<F> {
             assert_eq!(sink_input_schema.data_types(), info.schema.data_types());
         }
 
-        let stream_key = info.pk_indices.clone();
-        let downstream_pk = sink_param.downstream_pk_or_empty();
-        let stream_key_sink_pk_mismatch = stream_key.iter().any(|i| !downstream_pk.contains(i));
```
wenym1, line 128: Review note: originally, when the `stream_key` is not empty, `stream_key_sink_pk_mismatch` is the same as `!pk_specified_and_matched`.

src/stream/src/executor/sink.rs
```diff
@@ -557,7 +607,7 @@ impl<F: LogStoreFactory> SinkExecutor<F> {
         columns: Vec<ColumnCatalog>,
         mut sink_param: SinkParam,
         mut sink_writer_param: SinkWriterParam,
-        pk_matched: bool,
+        upsert_behavior: Option<UpsertBehavior>,
```
wenym1, line 610: In original code, when both `stream_key` and `user specified pk` is empty, `pk_mismatched` will be false and `pk_matched` will be true, while the current `Some(ub) && ub.should_compact_in_log_reader()` will be false, and the logic becomes different.

In our system we can assume that `stream_key` must not be empty. So we may add an assertion in the caller of this method that `stream_key` is non-empty, and then we can reason that the original logic is the same as current logic.

src/stream/src/executor/sink.rs
```diff
@@ -91,6 +88,106 @@ fn force_delete_only(c: StreamChunk) -> StreamChunk {
...
 }
 
+/// When the sink is an upsert sink, we need to do some extra work for correctness and performance.
+#[derive(Clone, Copy, Debug)]
+struct UpsertBehavior {
+    /// Whether the user specifies a primary key for the sink, and it matches the derived stream
+    /// key of the stream.
+    ///
+    /// By matching, we mean that stream key is a subset of the downstream pk.
+    pk_specified_and_matched: bool,
```
wenym1, line 98: Can we instead store `Option<user specified downstream pk>` so that in `should_reorder_records` we can get the user specified downstream pk directly rather doing downstream_pk.unwrap() with some additional assumption?

Issue #23513, opened 2025-10-20
Fail to generate dynamic filter plan with single-value right table with `LIMIT 1`


hzxa21:
### Describe the bug

```sql
 CREATE TABLE t1 (v1 CHARACTER VARYING)
 CREATE TABLE t2 (v2 CHARACTER VARYING)

-- doesn't work
create materialized view test as
select t1.v1, t2.v2
from t1
join (select v2 from t2 limit 1) t2 on t1.v1 > t2.v2;

ERROR:  Failed to run the query

Caused by:
  Not supported: streaming nested-loop join
HINT: The non-equal join in the query requires a nested-loop join executor, which could be very expensive to run. Consider rewriting the query to use dynamic filter as a substitute if possible.
See also: https://docs.risingwave.com/docs/current/sql-pattern-dynamic-filters/

```

### Error message/log

```text

```

BugenZhao:
This seems to be the same issue as https://github.com/risingwavelabs/risingwave/issues/5764: the right hand side appears to be scalar, but there's actually an extra `row_id` column. cc @chenzl25 

https://github.com/risingwavelabs/risingwave/blob/e1b4a21da9b1a0358ba55a7f79073a03d08cbe2b/e2e_test/streaming/dynamic_filter.slt#L14-L16

----

Also note that we require the output does not include right hand side input to convert into dynamic filter:

https://github.com/risingwavelabs/risingwave/blob/e1b4a21da9b1a0358ba55a7f79073a03d08cbe2b/src/frontend/src/optimizer/plan_node/logical_join.rs#L1387-L1394

PR #9513, opened 2023-04-27
feat(frontend): print partial explain traces if there's error

BugenZhao:
Signed-off-by: Bugen Zhao <i@bugenzhao.com>I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://gist.github.com/TennyZhuang/f00be7f16996ea48effb049aa7be4d66#file-rw_cla).

## What's changed and what's your intention?

This PR makes it possible to peek at the partial outputs of `EXPLAIN` if there's an error during plan generation and optimization. This may be helpful for us to develop a new feature or optimizer rule.

```
dev=> create table t (v int);
CREATE_TABLE
dev=> explain(trace) select (select v from t), 1 from t;
                                    QUERY PLAN                                     
-----------------------------------------------------------------------------------
 Begin:
 
 LogicalProject { exprs: [t.v, 1:Int32] }
 └─LogicalApply { type: LeftOuter, on: true, correlated_id: 1, max_one_row: true }
   ├─LogicalScan { table: t, columns: [v, _row_id] }
   └─LogicalProject { exprs: [t.v] }
     └─LogicalScan { table: t, columns: [v, _row_id] }
 
 ERROR: internal error: Scalar subquery might produce more than one row.
(9 rows)
```

For example, if we're working on a logical plan node and are sure that its `to_stream` won't work for now, we can now use `explain(logical)` to get the logical plan. Previously, this can only be achieved with the planner test.

## Checklist For Contributors

- [x] I have written necessary rustdoc comments
- [x] I have added necessary unit tests and integration tests
- [x] I have added fuzzing tests or opened an issue to track them. (Optional, recommended for new SQL features #7934).
- [x] I have demonstrated that backward compatibility is not broken by breaking changes and created issues to track deprecated features to be removed in the future. (Please refer to the issue)
- [x] All checks passed in `./risedev check` (or alias, `./risedev c`)

## Checklist For Reviewers

- [ ] I have requested macro/micro-benchmarks as this PR can affect performance substantially, and the results are shown.

## Documentation

- [x] My PR **DOES NOT** contain user-facing changes.

<details><summary>Click here for Documentation</summary>

### Types of user-facing changes

Please keep the types that apply to your changes, and remove the others.

- Installation and deployment
- Connector (sources & sinks)
- SQL commands, functions, and operators
- RisingWave cluster configuration changes
- Other (please specify in the release note below)

### Release note

</details>

PR #23568, opened 2025-10-24
fix(frontend): keep partial traces even if plan fails

BugenZhao:

This was introduced in #9513, but appeared to be broken by #16284. I find this regression when debugging #23513.

Added a planner test to ensure it works.

## Checklist

## Documentation

<details>

</details>


src/frontend/src/handler/explain.rs
```diff
@@ -197,14 +199,12 @@ pub async fn do_handle_explain(
                     }
 
                     _ => bail_not_implemented!("unsupported statement for EXPLAIN: {stmt}"),
-                }?;
```
BugenZhao, line 200: Do not early-return here.

Issue #23118, opened 2025-09-08
Bug: backfill order control panic when specifying a table not in the query


chenzl25:
### Describe the bug

```
create table car_sales(id int, car_id int, region_id int, price int);

create table car_info(id int, name varchar);

create table car_regions(id int, region varchar);

create table t(a int, b int, c int);
```

If we specify the `t1` inside backfill_order of the following query, it would panic.
```
 create materialized view m1
    with (backfill_order = FIXED(car_regions -> car_sales, t -> car_sales))
    as
      with price_ranges as (
        select
          car_info.name as name,
          car_sales.price as price,
          round(log10(1 + car_sales.price)::numeric, 1) as price_range
        from car_sales join car_info
          on car_sales.car_id = car_info.id
          join car_regions
            on car_sales.region_id = car_regions.id
      )
      select
        name,
        price_range,
        count(*) as sales_count,
        sum(price) as sales_volume,
        avg(price) as sales_avg,
        min(price) as sales_min,
        max(price) as sales_max,
        approx_percentile(0.5) WITHIN GROUP (ORDER BY price) as sales_est_median,
        approx_percentile(0.01) WITHIN GROUP (ORDER BY price) as sales_est_bottom_1_percent,
        approx_percentile(0.99) WITHIN GROUP (ORDER BY price) as sales_est_top_1_percent
      FROM
        price_ranges
GROUP BY name, price_range;
```

### Error message/log

```text

```

PR #23573, opened 2025-10-26
fix: panic in relation not found in backfill order

lyang24:

- Summarize your change
bug when we put table [x] in backfill order but table [x] is not in the backfill query would lead to panic here
https://github.com/risingwavelabs/risingwave/blob/d88378e21c3b71bd055bfadc6b4ab3fe5933db53/src/meta/src/stream/stream_graph/fragment.rs#L1026

- How does this PR work? 
  1. Collects all scanned relation IDs from the query plan (tables and sources)
  2. Validates that every table/source specified in the backfill_order clause is actually used in the query
  3. Returns a clear error message instead of panicking:
  Table or source 't' specified in backfill_order is not used in the query
  
- Refer to a related PR or issue link https://github.com/risingwavelabs/risingwave/issues/23118

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/e2e_test/backfill/test_backfill_order_validation.slt
+++ b/e2e_test/backfill/test_backfill_order_validation.slt
@@ -1,0 +1,99 @@
+statement ok
+SET RW_IMPLICIT_FLUSH TO TRUE;
+
+statement ok
+drop table if exists car_sales cascade;
+
+statement ok
+drop table if exists car_info cascade;
+
+statement ok
+drop table if exists car_regions cascade;
+
+statement ok
+drop table if exists t cascade;
+
+statement ok
+create table car_sales(id int, car_id int, region_id int, price int);
+
+statement ok
+create table car_info(id int, name varchar);
+
+statement ok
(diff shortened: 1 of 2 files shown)


e2e_test/backfill/test_backfill_order_validation.slt
```diff
@@ -0,0 +1,99 @@
...
+drop table car_sales;
+
+statement ok
+drop table car_info;
+
+statement ok
+drop table car_regions;
+
+statement ok
+drop table t;
```
kwannoel: Since the check is done in the frontend, we can make this a `planner_test` instead.

kwannoel:
Rest LGTM, just need to use planner test instead of e2e slt test.

e2e_test/backfill/test_backfill_order_validation.slt
lyang24: per discussion, the change is in the optimizer. planner test is not able to catch this atm. keeping the e2e test.

Issue #23127, opened 2025-09-08
bug: show jobs doesn't display a valid backfill order option


hzxa21:
### Describe the bug

SHOW JOBS shows:
```
CREATE MATERIALIZED VIEW ... WITH (backfill_order = t1 -> t2) AS SELECT ...;
```

It should be:
```
CREATE MATERIALIZED VIEW ... WITH (backfill_order = FIXED(t1 -> t2)) AS SELECT ...;

```

### Error message/log

```text

```

### The version of RisingWave

PostgreSQL 13.14.0-RisingWave-2.5.1 (c9c9ed83656ed4e56759d9f5a0b5cc481e060f7f)

PR #23576, opened 2025-10-27
fix: add display for fixed backfill order

lyang24:

Please explain **IN DETAIL** what the changes are in this PR and why they are needed:

- Summarize your change 
- How does this PR work? 
Display the FIXED backfill order keyword in show jobs by adding `FIXED({})` into the display method 
- Refer to a related PR or issue link https://github.com/risingwavelabs/risingwave/issues/23127

## Checklist

## Documentation

<details>

</details>


src/sqlparser/tests/sqlparser_common.rs
```diff
@@ -4137,3 +4137,26 @@ fn parse_window_clause() {
     let ast = parse_sql_statements(sql).unwrap();
     assert_eq!(ast.len(), 1);
 }
+
+#[test]
+fn parse_create_materialized_view_with_backfill_order() {
```
BugenZhao: Would you make this a test case under `sqlparser/tests/testdata/create.yaml`? You can generate the expected output with `./risedev update-parser-test`

BugenZhao (commented):
Thanks for the fix. Rest LGTM.

----

Seems to be a good example for why we should push https://github.com/risingwavelabs/risingwave/issues/20713. cc @xiangjinwu @kwannoel

src/sqlparser/tests/sqlparser_common.rs
lyang24: done

PR #23579, opened 2025-10-27
fix(binder): fix explain with parameters; simplify IN expr without consts

chenzl25:

- Previously, `select * from t where id in ($1)` will be converted into `select * from t where id in() or id = $1` which is unfriendly to the const folding and it looks like a bug as well.

## Checklist

## Documentation

<details>

</details>


src/frontend/src/binder/expr/mod.rs
```diff
@@ -278,22 +278,36 @@ impl Binder {
...
-            )?
-            .into();
+            if ret.is_none() {
+                ret = Some(FunctionCall::new(ExprType::Equal, vec![left.clone(), expr])?.into());
+            } else {
+                ret = Some(
+                    FunctionCall::new(
+                        ExprType::Or,
+                        vec![
+                            ret.unwrap(),
```
xiangjinwu: nit: `if-let` or `let-else` to avoid this `unwrap()`

chenzl25: >  In which case is the latter code path reachable?

We can simply run `explain select $1` to reproduce it.

chenzl25: Adding 2 test cases here.

xiangjinwu: > We can simply run `explain select $1` to reproduce it.

I see; so it is indeed unrelated to `IN` and just blocks the usage of `explain` when parameter is not provided. `select $1` without `explain` is still rejected (via captured panic) at a later stage: `Parameter should not be serialized to ExprNode`

xiangjinwu: > I agree it is redundant and better removed.

With the redundant part removed, a single equality condition can be pushed down for index selection. I guess this is the real fix (for performance), right?

chenzl25: True, but we shouldn't panic. And explain is useful to debug this prepared statement.

chenzl25: > With the redundant part removed, a single equality condition can be pushed down for index selection. I guess this is the real fix (for performance), right?

Yes. BTW, an in expression without any arguments seems an invalid expression.

xiangjinwu: Agree. Just trying to understand how each part of the PR contribute to fix for panic and performance.

xiangjinwu: > an in expression without any arguments seems an invalid expression

It is syntactically invalid, but semantically well defined - just testing against an empty set, like `IN (subquery)` where the subquery returns zero rows. Or did you find any problem in this definition or our existing implementation?

chenzl25: Indeed.It is syntactically invalid, but semantically well defined .

Issue #22579, opened 2025-07-11
Columns in wide table get overwritten with NULLs when sink uses LEFT JOIN


lgarcia11:
### Describe the bug

When using a LEFT JOIN in a SINK to populate a wide table, columns that are not present are being overwritten with NULL, even if those columns were previously populated with non-null values by other sinks.

### Error message/log

```text
No error message is produced, but the behavior results in data corruption: previously set values are unexpectedly replaced with NULL.
```

### To Reproduce

```
CREATE TABLE d1(v1 INT, k INT PRIMARY KEY);
CREATE TABLE d2(v2 INT, k INT PRIMARY KEY);
CREATE TABLE d3(v3 INT, k INT PRIMARY KEY);

CREATE TABLE wide_d(v1 INT, v2 INT, v3 INT, k INT PRIMARY KEY)  ON CONFLICT DO UPDATE IF NOT NULL;
```
```
INSERT INTO d1(v1, k) SELECT k, k FROM generate_series(1, 1000) AS k;
INSERT INTO d2(v2, k) SELECT k, k FROM generate_series(1, 1000) AS k;
INSERT INTO d3(v3, k) SELECT k, k FROM generate_series(1, 1000) AS k;
```
```
CREATE SINK sink1 INTO wide_d (v1, k)
AS SELECT v1, k FROM d1;

CREATE SINK sink2 INTO wide_d (v2, k)
AS SELECT d2.v2, d1.k FROM d1 LEFT JOIN d2 ON d1.k = d2.k;

CREATE SINK sink3 INTO wide_d (v3, k)
AS SELECT d3.v3, d1.k FROM d1 LEFT JOIN d3 ON d1.k = d3.k;
```

### Expected behavior

Each SINK should only update the columns explicitly selected in its query. Other columns in the target table should remain unchanged, preserving data written by previous sinks.

### How did you deploy RisingWave?

Deployed via Kubernetes using the RisingWave Operator, with PostgreSQL as the meta store and S3 as the state store.

### The version of RisingWave

PostgreSQL 13.14.0-RisingWave-2.4.3 (14409318a2d5c9d615e146f5d222b8c9b5f28574)

pjpringlenom:
https://docs.risingwave.com/processing/maintain-wide-table-with-table-sinks

did you try with force_append_only as per wide table guide

lgarcia11:
> https://docs.risingwave.com/processing/maintain-wide-table-with-table-sinks
> 
[...]

Issue #23540, opened 2025-10-22
Unexpected Op in output of StreamChunkCompactor into_compacted_chunks


qccash:
### Describe the bug

Currently, if there are more 1 pair of UpdateDelete UpdateInsert in the input chunk, the output of into_compacted_chunks would turn to Delete and Insert operations, like:
```rust
    #[test]
    fn test_merge_chunk_row() {
        let pk_indices = [0, 1];
        let chunks = vec![
            StreamChunk::from_pretty(
                "  I I I
                U- 8 8 1
                U+ 8 8 2
                U- 8 8 2
                U+ 8 8 3",
            ),
        ];
        let compactor = StreamChunkCompactor::new(pk_indices.to_vec(), chunks);
        let mut iter = compactor.into_compacted_chunks(InconsistencyBehavior::Panic);
        assert_eq!(
            iter.next().unwrap().compact(),
            StreamChunk::from_pretty(
                " I I I
                - 8 8 1
                + 8 8 3",
            )
        );

        assert_eq!(iter.next(), None);
    }
```

Expected result should be still UpdateDelete and UpdateInsert operations, like:
```rust
            StreamChunk::from_pretty(
                "  I I I
                U- 8 8 1
                U+ 8 8 3",
            )
```

## Why this matters?
For downstream connectors, the UpdateDelete and Delete operations may have different semantics. Taking the JDBC connector as an example, UpdateDelete does not issue an actual DELETE statement, whereas delete does.

## Suggestions
1. Do not normalize_update
2. Handle UpdateDelete and UpdateInsert Op in OpRowMutRefTuple::push

### Error message/log

```text

```

### To Reproduce

Here is an example of unexpected delete statements.
```python
import time

import psycopg2
import pymysql


my_conn = pymysql.connect(host="localhost", port=3306, user="root", password="xxxx", database="test")
my_cursor = my_conn.cursor()
my_cursor.execute("drop table if exists t1;")
my_cursor.execute("""
create table t1 (
    id bigint primary key,
    value bigint,
    mark bigint
);
""")
my_cursor.execute("drop table if exists t2;")
my_cursor.execute("""
create table t2 (
    id bigint primary key,
    value bigint
);
""")
my_cursor.execute("drop table if exists t3;")
my_cursor.execute("""
create table t3 (
    id bigint primary key,
    value bigint
);
""")
my_cursor.execute("drop table if exists t4;")
my_cursor.execute("""
create table t4 (
    id bigint primary key,
[...]

PR #23581, opened 2025-10-27
feat(sink): produce `Upsert` chunk in compaction for `SinkType::Upsert`

BugenZhao:

Extend `StreamChunkCompactor` to allow specifying `OutputKind`, which decides whether to produce chunks in

- `Retract` format, which convert `Update` record to a pair of `U-` and `U+` rows
- or `Upsert` format, which only keep the new row of `Update` record and convert it to an `Insert` row

Dispatch different `OutputKind` based on whether the `SinkType` is `Upsert` or `Retract` (introduced in last PR: #23593).

----

Previously...

- For "inline" compaction, the actual effective `Update` rows might result in inconsecutive, forcing us to rewrite it into 2 operations (`Delete` and `Insert`) (https://github.com/risingwavelabs/risingwave/issues/23540)
- Also, some sink implementation may not handle `Update` operation in the optimal way

This led to unnecessary `Delete` operation to external system, along with I/O overhead and temporary inconsistent state.

Directly produce chunk in upsert format in compaction minimize `Delete` operation as much as possible. This approach also better aligns with the semantics of `SinkType::Upsert`. If the sink explicitly requires old value for updates, typically `DEBEZIUM`, it should be marked `SinkType::Retract` instead.

By using upsert format in sink-into-table, we also fixes https://github.com/risingwavelabs/risingwave/issues/22579, since the update won't be considered as a pair of `Delete` and `Insert` when handling table conflict.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/e2e_test/streaming/sink_into_wide_table.slt
+++ b/e2e_test/streaming/sink_into_wide_table.slt
@@ -1,0 +1,59 @@
+statement ok
+set implicit_flush = true;
+
+statement ok
+create table t (k int primary key, a int, b int) on conflict DO UPDATE IF NOT NULL;
+
+statement ok
+create table sa (k int primary key, a int);
+
+statement ok
+create table sb (k int primary key, b int);
+
+statement ok
+create sink ska into t(k, a) from sa;
+
+statement ok
+create sink skb into t(k, b) from sb;
+
+statement ok
+insert into sa values (1, 101), (2, 102);
+
+query III
+select * from t order by k;
+----
+1 101 NULL
+2 102 NULL
+
+# 1. `Insert` record from a sink can enrich an existing row.
+statement ok
+insert into sb values (1, 201);
+
+query III
+select * from t order by k;
+----
+1 101 201
+2 102 NULL
+
+# 2. `Update` record from a sink can also correctly enrich an existing row.
+# https://github.com/risingwavelabs/risingwave/issues/22579
+statement ok
+update sb set b = 202 where k = 1;
+
+query III
+select * from t order by k;
+----
+1 101 202
+2 102 NULL
+
+# 3. `Delete` record from one sink will remove the entire row.
+statement ok
+delete from sb where k = 1;
+
+query III
+select * from t order by k;
+----
+2 102 NULL
+
+statement ok
+drop table t cascade;


BugenZhao:
* **#23581** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23581?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23581?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#23593** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23593?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

BugenZhao:
I realized that there's an exception: `FORMAT DEBEZIUM` actually requires retractable output.

BugenZhao:
> I realized that there's an exception: `FORMAT DEBEZIUM` actually requires retractable output.

Addressed by introducing `SinkType::Retract` first in #23593, and dispatch different behavior based on it's retractable or upsert.

BugenZhao:
Just realized that this work also fixes https://github.com/risingwavelabs/risingwave/issues/22579.

src/stream/src/common/compact_chunk.rs
```diff
@@ -68,23 +69,32 @@ impl StreamChunkCompactor {
...
+                                new_row.set_op(Op::UpdateInsert);
+                            }
+                        }
                     }
                 }
             }
         }
 
-        chunks.into_iter().map(|c| c.into())
+        chunks.into_iter().map(|c| c.into()).collect()
```
hzxa21, line 92: Just curious, is it possible to keep returning `impl Iterator` instead of `Vec` to avoid one extra buffer? It is not a big deal given that we anyway need to buffer the insert and delete chunks in sink executor when inline compaction happens.

BugenZhao:
> Can we add a e2e test for the [wide table pattern](https://docs.risingwave.com/processing/maintain-wide-table-with-table-sinks#maintain-wide-table-with-table-sinks) with `ON CONFLICT DO UPDATE IF NOT NULL` to demonstrate the upsert semantics for partial update?

True. I've tested locally but forget to add it to the PR.

src/stream/src/common/compact_chunk.rs
BugenZhao: The issue is that with different const generic of `const KIND: OutputKind`, the opaque type of `impl Iterator` will be considered different by the compiler. While we can specify the concrete the return type of `Map<vec::IntoIter<..>, fn ..>` here, I would still prefer collecting a Vec here, which is less ugly. 😂

PR #23582, opened 2025-10-27
feat(pulsar): add support for read_compacted option in consumer properties

tabVersion:

per requested

## Checklist

## Documentation

<details>

add `pulsar.read_compacted` to control the pulsar consumer read from the compacted topic or not. Not allowed to change on the fly. 

</details>

### DIFF
--- a/src/connector/src/source/pulsar/source/reader.rs
+++ b/src/connector/src/source/pulsar/source/reader.rs
@@ -185,18 +185,19 @@ impl SplitReader for PulsarBrokerReader {
 
         let mut already_read_offset = None;
 
-        let mut consumer_option = match split.start_offset.clone() {
+        let mut consumer_options = match split.start_offset.clone() {
             PulsarEnumeratorOffset::Earliest => {
-                if topic.starts_with("non-persistent://") {
+                let initial_position = if topic.starts_with("non-persistent://") {
                     tracing::warn!(
                         "Earliest offset is not supported for non-persistent topic, use Latest instead"
                     );
-                    ConsumerOptions::default().with_initial_position(InitialPosition::Latest)
+                    InitialPosition::Latest
                 } else {
-                    ConsumerOptions::default()
-                        .with_initial_position(InitialPosition::Earliest)
-                        .durable(false)
-                }
+                    InitialPosition::Earliest
+                };
+                ConsumerOptions::default()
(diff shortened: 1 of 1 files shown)


hzxa21:
CI fails
```
<html>
<body>
<!--StartFragment-->
failed to run `e2e_test/source_inline/pulsar/issue-22486.slt.serial`
--
  |  
  | Caused by:
  | statement failed: db error: ERROR: Failed to run the query
  |  
  | Caused by these errors (recent errors listed first):
  | 1: gRPC request to meta service (call `/ddl_service.DdlService/CreateTable`) failed: Internal error
  | 2: failed to create source worker
  | 3: failed to parse json
  | 4: missing field `pulsar.read_compacted`
  |  
  | [SQL] create table t_partitioned (
  | f1 int,
  | f2 varchar)
  | with (
  | connector = 'pulsar',
  | service.url = '${PULSAR_BROKER_URL}',
  | topic = 'persistent://public/default/topic-recovery',
  | ) format plain encode json;
  | at e2e_test/source_inline/pulsar/issue-22486.slt.serial:29

<!--EndFragment-->
</body>
</html>
```

I think there is a backward compatible issue here.

hzxa21 (commented):
Can we have a e2e to verify read_compacted is working when set to true? For example,

1. emit multiple messages with the same key to upstream pulsar
2. create a RW table using pulsar connector and read_compacted set to true
3. create a downstream MV on top of this table using [AS CHNAGE LOG](https://docs.risingwave.com/sql/commands/sql-as-changelog)
4. verify that there is only one row corresponding to the pulsar key instead of many

tabVersion:
> Can we have a e2e to verify read_compacted is working when set to true? For example,
> 
> 1. emit multiple messages with the same key to upstream pulsar
> 2. create a RW table using pulsar connector and read_compacted set to true
> 3. create a downstream MV on top of this table using [AS CHNAGE LOG](https://docs.risingwave.com/sql/commands/sql-as-changelog)
> 4. verify that there is only one row corresponding to the pulsar key instead of many

Sure, we ought to have the test. 

Instead, we can simplify it to:

1. produce to pulsar with schema `(foo, bar)` (say `foo` is the key), value `(1, 1)`, `(1, 2)`, `(1, 3)`
2. create a table with format plain
3. we can only ingest `(1, 3)`

Issue #21668, opened 2025-04-30
Double insert on downstream due to order key mismatch


kwannoel:
### Describe the bug

After https://github.com/risingwavelabs/risingwave/pull/20176, we incorrectly set the _user order key_ to be the dist key.
Because order key does not need to be the same as stream key, it can result in `updateInsert/updateDelete` pairs being split up.

Subsequently, if they are shuffled back to the same node, they can be dis-ordered, instead of `U-/U+` we have `U+/U-` and it leads to inconsistent results.

I think ORDER BY can be part of MV's distribution key. But it must contain the full stream key as well to ensure updates don't become dis-ordered if we re-shuffle back to original distribution.

Consider the following case of StreamTableScan after https://github.com/risingwavelabs/risingwave/commit/27efddf862d4b7969e0057301070c9d6ba3d9dad:
```
       └─StreamExchange { dist: HashShard(token_mint) }
         └─StreamTableScan {
              stream_key: [owner, token_mint],
              pk: [token_mint, balance, owner],
              dist: UpstreamHashShard(token_mint, balance)
```

The dist is the user order key. It gets shuffled to different actors initially, and the update on `0x16, 0x10` becomes disordered...

```
         mint   owner    timestamp                 balance
         v      v        v                         v
A) | + | 0x16 | 0x10 | 1970-01-01 03:31:04+00:00 | 3 | ----------> Actor 38
B) | + | 0x16 | 0x10 | 1970-01-01 03:45:06+00:00 | 1 | ----------> Actor 37 out of order
C) | - | 0x16 | 0x10 | 1970-01-01 03:31:04+00:00 | 3 | ----------> Actor 38 with its update insert
```

... and then shuffled back to the same actor with:
```
       └─StreamExchange { dist: HashShard(token_mint) }
         └─StreamTableScan {

```

So we get:
```
A) | + | 0x16 | 0x10 | 1970-01-01 03:31:04+00:00 | 3 | ----------> Actor 35
B) | + | 0x16 | 0x10 | 1970-01-01 03:45:06+00:00 | 1 | ----------> Actor 35
C) | - | 0x16 | 0x10 | 1970-01-01 03:31:04+00:00 | 3 | ----------> Actor 35
```

Where B should be _after_ C, and we encounter a double-insert error.

### Error message/log

```text

```

Issue #5277, opened 2022-09-12
bug: `Materialize` double insert in tpch q15


BugenZhao:
https://buildkite.com/risingwavelabs/main/builds/1574#01832fbc-fe3e-41ae-8e95-fd685fe09fd1

```
thread 'risingwave-streaming-actor' panicked at 'invalid flush status: double insert [75, 0, 128, 0, 0, 4, 0, 128, 0, 0, 4] -> [1, 4, 0, 0, 0, 1, 18, 0, 0, 0, 83, 117, 112, 112, 108, 105, 101, 114, 35, 48, 48, 48, 48, 48, 48, 48, 48, 52, 1, 25, 0, 0, 0, 66, 107, 55, 97, 104, 52, 67, 75, 56, 83, 89, 81, 84, 101, 112, 69, 109, 118, 77, 107, 107, 103, 77, 119, 103, 1, 15, 0, 0, 0, 50, 53, 45, 56, 52, 51, 45, 55, 56, 55, 45, 55, 52, 55, 57, 1, 0, 0, 4, 0, 98, 25, 56, 25, 2, 0, 0, 0, 0, 0, 0, 0, 1, 4, 0, 0, 0]', src/storage/src/table/streaming_table/mem_table.rs:69:21
stack backtrace:
2022-09-12T03:42:38.947750Z DEBUG risingwave_stream::task::stream_manager: drop actors actors=[6510, 6511, 6512, 6513, 6522, 6523, 6524, 6525, 6534, 6535, 6536, 6537, 6546, 6547, 6548, 6549]
   0: rust_begin_unwind
             at /rustc/9067d5277d10f0f32a49ec9c125a33828e26a32b/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/9067d5277d10f0f32a49ec9c125a33828e26a32b/library/core/src/panicking.rs:142:14
   2: risingwave_storage::table::streaming_table::mem_table::MemTable::insert
             at ./src/storage/src/table/streaming_table/mem_table.rs:69:21
   3: risingwave_storage::table::streaming_table::state_table::StateTable<S>::write_chunk
             at ./src/storage/src/table/streaming_table/state_table.rs:446:58
   4: risingwave_stream::executor::mview::materialize::MaterializeExecutor<S>::execute_inner::{{closure}}
             at ./src/stream/src/executor/mview/materialize.rs:119:21
[...]

PR #23583, opened 2025-10-27
refactor(optimizer): add assertion on dist key being a subset of stream key in exchange

BugenZhao:

This is a key invariant to avoid disordering of the changes on the same source row during parallel processing. We've encountered relevant issues several times:

- https://github.com/risingwavelabs/risingwave/issues/5277
- https://github.com/risingwavelabs/risingwave/issues/21668
- https://github.com/risingwavelabs/risingwave/pull/23592

We should really add an assertion.

Note that we don't require this invariant to hold everywhere on the stream. We just need to check it right before the stream is going to be re-shuffled. Since `StreamExchange` won't change the `stream_key` of the input plan, so it could be common that right after an `Exchange`, the new distribution key will temporarily **not** be a subset of the stream key. Imagine that a stream is shuffled to be distributed by the join key before going into a `HashJoin`, we only add the join key into the stream key **after** outputting from `HashJoin`, which is okay.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/optimizer/plan_node/stream_exchange.rs
+++ b/src/frontend/src/optimizer/plan_node/stream_exchange.rs
@@ -73,7 +73,7 @@ impl StreamExchange {
                 "distribution key must be a subset of stream key before shuffle to a different distribution\n\
                  - dist_key: {input_dist_key:?}\n\
                  - stream_key: {input_stream_key:?}\n\
-                 - schema: {:?}\n\
+                 - schema: {}\n\
                  - plan:\n{}",
                 input.schema().formatted_col_names(),
                 input.explain_to_string()


BugenZhao:
* **#23583** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23583?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23583?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#23687** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23687?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

chenzl25 (approved):
LGTM

hzxa21 (approved):
LGTM

src/frontend/src/optimizer/plan_node/stream_exchange.rs
```diff
@@ -46,6 +48,38 @@ impl StreamExchange {
             "exchange can not be used to enforce such distribution"
         );
 
+        // For non-append-only distributed stream, check that distribution key is a subset of stream key
```
chenzl25, line 51: It seems it is too strict. This assertion means we can't shuffle a plan if its stream key doesn't contain the shuffle key, but for a table/Mv, its stream key is fixed, does it mean we can't shuffle any columns not in the stream key? Clearly not.

BugenZhao: > doesn't contain the shuffle key

Nope. Here we check the distribution key of the upstream fragment, instead of the key to shuffle with.

chenzl25: The invariant should be the exchange's downstream stream key should always contain the shuffle key.

BugenZhao: This is not this assertion is for. 🥵 Can check the PR description: 

> Note that we don't require this invariant to hold everywhere on the stream. We just need to check it right before the stream is going to be re-shuffled. Since `StreamExchange` won't change the `stream_key` of the input plan, so it could be common that right after an `Exchange`, the new distribution key will temporarily **not** be a subset of the stream key. Imagine that a stream is shuffled to be distributed by the join key before going into a `HashJoin`, we only add the join key into the stream key **after** outputting from `HashJoin`, which is okay.

chenzl25: > Nope. Here we check the distribution key of the upstream fragment, instead of the key to shuffle with.

Oh. I got it

PR #23592, opened 2025-10-28
fix(streaming): fix index's stream key to include distribution key

chenzl25:

- Previously, index's stream key is the same as the primary key of the primary table, but the distribution keys of the index are the index columns. In streaming queries, it is important to include the distribution keys to the stream keys, so in this PR, we will add the distribution keys of indexes to the stream keys. And for the backward compatibility issue, we add a method `stream_key()` to the table catalog, so that we can always include the distribution keys to the stream keys. We should never use the stream_key field directly, instead we should use the `stream_key()` method in the future.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/planner_test/tests/testdata/input/index_selection_for_backfilling.yaml
+++ b/src/frontend/planner_test/tests/testdata/input/index_selection_for_backfilling.yaml
@@ -223,3 +223,10 @@
     select sum(unnest) from (select a, b, c, unnest(x) as unnest from t) group by a + 1;
   expected_outputs:
     - stream_plan
+- name: Ban index selection for indexes with distribution key is not the subset of the group by keys
+  sql: |
+    create table t(a int, b int, c int, x int[]);
+    create index idx_a on t(a, b, c) distributed by (a, b, c);
+    select count(*) from t group by a, b;
+  expected_outputs:
+    - stream_plan
--- a/src/frontend/planner_test/tests/testdata/output/index_selection_for_backfilling.yaml
+++ b/src/frontend/planner_test/tests/testdata/output/index_selection_for_backfilling.yaml
@@ -362,3 +362,14 @@
           └─StreamProject { exprs: [(idx_a.a + 1:Int32) as $expr1, Unnest($1), idx_a.t._row_id, projected_row_id] }
             └─StreamProjectSet { select_list: [$0, Unnest($1), $2] }
               └─StreamTableScan { table: idx_a, columns: [idx_a.a, idx_a.x, idx_a.t._row_id], stream_scan_type: ArrangementBackfill, stream_key: [idx_a.a, idx_a.t._row_id], pk: [a, t._row_id], dist: UpstreamHashShard(idx_a.a) }
+- name: Ban index selection for indexes with distribution key is not the subset of the group by keys
+  sql: |
+    create table t(a int, b int, c int, x int[]);
+    create index idx_a on t(a, b, c) distributed by (a, b, c);
+    select count(*) from t group by a, b;
+  stream_plan: |-
+    StreamMaterialize { columns: [count, t.a(hidden), t.b(hidden)], stream_key: [t.a, t.b], pk_columns: [t.a, t.b], pk_conflict: NoCheck }
+    └─StreamProject { exprs: [count, t.a, t.b] }
+      └─StreamHashAgg { group_key: [t.a, t.b], aggs: [count] }
+        └─StreamExchange { dist: HashShard(t.a, t.b) }
+          └─StreamTableScan { table: t, columns: [t.a, t.b, t._row_id], stream_scan_type: ArrangementBackfill, stream_key: [t._row_id], pk: [_row_id], dist: UpstreamHashShard(t._row_id) }
--- a/src/frontend/src/optimizer/plan_node/logical_scan.rs
+++ b/src/frontend/src/optimizer/plan_node/logical_scan.rs
@@ -718,7 +718,16 @@ impl ToStream for LogicalScan {
             let order_satisfied_index = self.indexes_satisfy_order(&required_order);
             for index in order_satisfied_index {
                 if let Some(index_scan) = self.to_index_scan_if_index_covered(index) {
-                    return Some(index_scan.into());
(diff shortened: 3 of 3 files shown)


src/frontend/src/catalog/table_catalog.rs
```diff
@@ -447,6 +449,24 @@ impl TableCatalog {
...
+    pub fn stream_key(&self) -> Vec<usize> {
+        // if distribution key is not in stream key, we need to add them to stream key
+        if self
+            .distribution_key
+            .iter()
+            .any(|dist_key| !self.stream_key.contains(dist_key))
+        {
+            let mut new_stream_key = self.distribution_key.clone();
+            new_stream_key.extend(self.stream_key.iter());
+            new_stream_key
```
BugenZhao, line 469: Do we need to deduplicate the columns?

src/frontend/src/optimizer/plan_node/derive.rs
```diff
@@ -79,16 +79,23 @@ pub(crate) fn derive_columns(
...
+    // Add distribution key columns to stream key
+    let mut stream_key = match user_distributed_by {
+        RequiredDist::PhysicalDist(distribution) => distribution.dist_column_indices().to_vec(),
+        RequiredDist::ShardByKey(_) => {
+            unreachable!("Right now, it is not possible to have ShardByKey here")
+        }
+        RequiredDist::AnyShard | RequiredDist::Any => vec![],
+    };
+
+    stream_key.extend(input.expect_stream_key().iter().copied().unique());
```
BugenZhao: ditto

src/frontend/src/optimizer/plan_node/generic/join.rs
```diff
@@ -213,76 +213,68 @@ impl<PlanRef: GenericPlanRef> GenericPlanNode for Join<PlanRef> {
...
 
-        let mut pk_indices = left_pk
+        // Collect PKs in internal column space (without applying i2o mapping yet)
+        let mut pk_indices_internal = left_pk
             .iter()
             .map(|index| l2i.try_map(*index))
             .chain(right_pk.iter().map(|index| r2i.try_map(*index)))
             .flatten()
-            .map(|index| i2o.try_map(index))
-            .collect::<Option<Vec<_>>>()?;
```
chenzl25, line 222: The derivation of join keys logic is changed. We apply the `i2o` mapping at the end. Otherwise, index selection for streaming join might panic, because index selection will replace a primary table with an index. Before this PR, they will have the same stream keys, but after this PR, they have different stream keys. For other operator like Agg, TopN, we can safely replace it with indexes, because the index stream_key's existence can always be  guaranteed after `logical_rewrite_for_stream`. However, for the join, we only keep one side of the join key as the join's stream_key, and the other side's join key could be removed (even if it is the stream_key of the index).

src/frontend/src/catalog/table_catalog.rs
chenzl25: You are right, done

chenzl25:
> thread 'rw-streaming' (19144930) panicked at src/stream/src/executor/wrapper/update_check.rs:38:21:
assertion `left == right` failed: U- and U+ should have same stream key
U- row: [8852, 107730, 1, 5, 773, NULL]
U- key: [8852, 1, 5, 773, NULL]
U+ row: [6892, 107730, 1, 5, 773, NULL]
U+ key: [6892, 1, 5, 773, NULL]
stream key indices: [0, 2, 3, 4, 5]
executor: Materialize 13D00000003
  left: Project { row: [Some(Decimal(Normalized(8852))), Some(Int32(107730)), Some(Int32(1)), Some(Int32(5)), Some(Int32(773)), None], indices: [0, 2, 3, 4, 5] }
 right: Project { row: [Some(Decimal(Normalized(6892))), Some(Int32(107730)), Some(Int32(1)), Some(Int32(5)), Some(Int32(773)), None], indices: [0, 2, 3, 4, 5] }

The e2e test failed, because it is possible that ExchangeExecutor passes through the U-/U+  to a same actor if the vnode of the distribution key calculation is the same even though the distribution key is changed. But Exchange's distribution key should be part of the stream key of the downstream executor. 

It means we need to change U-/U+ to D-/I+ in ExchangeExecutor when the columns of distribution key are changed instead of doing these changes when vnode are different.

chenzl25:
> > thread 'rw-streaming' (19144930) panicked at src/stream/src/executor/wrapper/update_check.rs:38:21:
> > assertion `left == right` failed: U- and U+ should have same stream key
> > U- row: [8852, 107730, 1, 5, 773, NULL]
> > U- key: [8852, 1, 5, 773, NULL]
> > U+ row: [6892, 107730, 1, 5, 773, NULL]
> > U+ key: [6892, 1, 5, 773, NULL]
> > stream key indices: [0, 2, 3, 4, 5]
> > executor: Materialize 13D00000003
> > left: Project { row: [Some(Decimal(Normalized(8852))), Some(Int32(107730)), Some(Int32(1)), Some(Int32(5)), Some(Int32(773)), None], indices: [0, 2, 3, 4, 5] }
> > right: Project { row: [Some(Decimal(Normalized(6892))), Some(Int32(107730)), Some(Int32(1)), Some(Int32(5)), Some(Int32(773)), None], indices: [0, 2, 3, 4, 5] }
> 
> The e2e test failed, because it is possible that ExchangeExecutor passes through the U-/U+ to a same actor if the vnode of the distribution key calculation is the same even though the distribution key is changed. But Exchange's distribution key should be part of the stream key of the downstream executor.
> 
> It means we need to change U-/U+ to D-/I+ in ExchangeExecutor when the columns of distribution key are changed instead of doing these changes when vnode are different.

waiting for https://github.com/risingwavelabs/risingwave/pull/23603

BugenZhao:
* **#23592** <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23592?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/risingwavelabs/risingwave/23592?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

BugenZhao (approved):
LGTM!

PR #23593, opened 2025-10-28
feat(sink): introduce `Retract` sink type

BugenZhao:

Currently, we only have `AppendOnly` and `Upsert` sink type. However, some sink formats, typically `DEBEZIUM`, actually requires sinking stream with `Retract` kind. We'd better introduce a new `Retract` type for it.

By doing this,

- we get a better-defined `SinkType`
- we can correctly reject creating retract sink on upsert stream (introduced in #22799)
- we can differentiate between `Upsert` and `Retract` in sink executor, allowing us to perform additional optimizations for specific sink types (#23581)

After this PR,

- only sinks with `FORMAT DEBEZIUM` will be derived with retract type
- sink-into-table will keep being upsert, since we always treat the stream as upsert after merging sinks in downstream table
- users can specify `type = 'retract'`, but for most sink connectors there won't be effect, making it generally unnecessary

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/e2e_test/sink/append_only_sink.slt
+++ b/e2e_test/sink/append_only_sink.slt
@@ -7,7 +7,7 @@ create sink s1 from t with (connector = 'blackhole');
 statement ok
 create sink s2 as select avg(v1), v2 from t group by v2 with (connector = 'blackhole');
 
-statement error sink cannot be append-only
+statement error cannot be append-only
 create sink s3 from t with (connector = 'blackhole', type = 'append-only');
 
 statement ok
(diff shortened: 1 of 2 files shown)


BugenZhao:
* **#23581** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23581?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#23593** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23593?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23593?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #23597, opened 2025-10-28
feat: manage all catalogs of the iceberg table in a transactional manner

yezizp2012:

Part of #21863 .

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/meta/src/controller/catalog/drop_op.rs
+++ b/src/meta/src/controller/catalog/drop_op.rs
@@ -249,27 +249,32 @@ impl CatalogController {
 
         let mut removed_state_table_ids: HashSet<_> = removed_table_ids.clone().collect();
 
-        // Add associated sources.
-        let mut removed_source_ids: Vec<SourceId> = Table::find()
-            .select_only()
-            .column(table::Column::OptionalAssociatedSourceId)
-            .filter(
-                table::Column::TableId
-                    .is_in(removed_table_ids)
-                    .and(table::Column::OptionalAssociatedSourceId.is_not_null()),
-            )
-            .into_tuple()
-            .all(&txn)
-            .await?;
-        let removed_source_objs: Vec<PartialObject> = Object::find()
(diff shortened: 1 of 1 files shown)


src/meta/src/controller/catalog/drop_op.rs
```diff
@@ -388,6 +432,7 @@ impl CatalogController {
                 removed_actors,
                 removed_fragments,
                 removed_sink_fragment_by_targets,
+                removed_iceberg_table_sinks,
```
chenzl25, line 440: Do we need to add `removed_iceberg_table_sources` here as well?

src/meta/src/controller/streaming_job.rs
```diff
@@ -507,6 +508,11 @@ impl CatalogController {
         // Add streaming job objects to notification
         if need_notify {
             match creating_streaming_job.unwrap() {
+                StreamingJob::Table(Some(source), ..) => {
+                    objects_to_notify.push(PbObject {
+                        object_info: Some(PbObjectInfo::Source(source.clone())),
+                    });
+                }
```
chenzl25, line 515: Is it for iceberg table or another fix?

chenzl25 (approved):
LGTM!!!

src/meta/src/controller/streaming_job.rs
yezizp2012: It's for iceberg table. Typically, creating a normal table with a connector does not require `background` mode, as there is no backfilling involved. Therefore, rw doesn't support it. The change here is because the iceberg table is set to `background` in meta.

src/meta/src/controller/catalog/drop_op.rs
yezizp2012: I thought iceberg source was already in `removed_source_ids`, but in fact there was something wrong with the collection of `removed_source_ids`. I also fixed it in this PR.

PR #23286, opened 2025-09-23
feat(mysql-cdc):support mysql 8.4

wcy-fdu:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## What's changed and what's your intention?
As title, support mysql 8.4:
- backfill phase, get version and choose the right statement
- debezium phase, refer to changes in https://github.com/debezium/debezium/pull/5765/

## Checklist

- [ ] I have written necessary rustdoc comments.
- [ ]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  I have checked the [Release Timeline](https://github.com/risingwavelabs/rw-commits-history/blob/main/release_timeline.md) and [Currently Supported Versions](https://docs.risingwave.com/changelog/release-support-policy#support-end-dates-for-recent-releases) to determine which release branches I need to cherry-pick this PR into. 

## Documentation

- [ ]  My PR needs documentation updates. 

<details>
<summary><b>Release note</b></summary>

</details>

PR #18760, opened 2024-09-29
feat(postgres-cdc): support replicating Postgres schema change

StrikeW:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://gist.github.com/TennyZhuang/f00be7f16996ea48effb049aa7be4d66#file-rw_cla).

## What's changed and what's your intention?

- Copy and patched the Debezium postgres conenctor (v2.6.2) to our project (patched code enclosing with `/* patched */`)
  - `PgOutputMessageDecoder`: dispatch schema change event when handling `R` message
  - Make `PostgresSchema` act as `HistorizedDatabaseSchema` which is a prerequisite to emit schema change event
  - Extends the `ReplicationStream.ReplicationMessageProcessor` to provide interfaces to emit schema change event

- Modify the jar loading order to ensure `risingwave-source-cdc` to be the first jar in the classpath so that the jvm can load those patched .class instead of the original classes
  
- The schema change event will not be emitted until new data events come in, so if DDL has been executed on upstream but there is no IUD on upstream, the schema change event will not be emitted

- Add `auto.schema.change` option to postgres source

### Test Coverage:
* Added a new end-to-end test script `auto_schema_change_pg.slt` to verify the handling of automatic schema changes in PostgreSQL CDC sources. (`e2e_test/source/cdc_inline/auto_schema_change_pg.slt`)

### Logging Enhancements:
* Updated logging statements in `DbzChangeEventConsumer` to include schema and data context in log messages. (`java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/core/DbzChangeEventConsumer.java`) [[1]](diffhunk://#diff-57d910841f22f98a42586fa0510bf55e550bde2921c4a6df2aeb995652e24ab1L216-R216) [[2]](diffhunk://#diff-57d910841f22f98a42586fa0510bf55e550bde2921c4a6df2aeb995652e24ab1L272-R272)

### Configuration Updates:
* Added a new configuration property `include.schema.changes` to `postgres.properties` to control whether schema changes are included. (`java/connector-node/risingwave-connector-service/src/main/resources/postgres.properties`)

### Miscellaneous:
* Updated `.typos.toml` to exclude certain Java files from typo checks. (`.typos.toml`)

## Checklist

- [x] I have written necessary rustdoc comments
- [x] I have added necessary unit tests and integration tests
[...]

PR #23600, opened 2025-10-29
refactor: upgrade to debezium 3.2.4 and cherry-pick patch

zwang28:

Bump debezium to v3.2.4, which requires Java 17.

Since we're applying copy-and-modify pattern to patch debezium, existing patches need be re-applied after upgrading debezium:
1. #18760 is cherry-picked to v3.2.4 files. A few conflicts have been resolved. refer to [this](https://github.com/zwang28/debezium/tree/v3.2.4.rw).
    - The new interface `HistorizedDatabaseSchema::getSchemaHistory` is implemented via a no-op new class `PostgresSchemaHistory`.
2. #23286 this patch is no longer needed since v3.2.4 debezium already support MySql 8.4.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/src/source/cdc/source/reader.rs
+++ b/src/connector/src/source/cdc/source/reader.rs
@@ -167,7 +167,10 @@ impl<T: CdcSourceTypeTrait> SplitReader for CdcSplitReader<T> {
             };
             if !inited {
                 bail!(
-                    "failed to start cdc connector.\nHINT: increase `cdc_source_wait_streaming_start_timeout` session variable to a large value and retry."
+                    "failed to start cdc connector due to timeout (default 60s).\n\
+                    HINT: You can increase the timeout by:\n\
+                    - Alter the running source: ALTER SOURCE <source_name> SET CONNECTOR WITH ('cdc.source.wait.streaming.start.timeout' = '<larger_value>');\n\
+                    - Set session variable and recreate the source: SET cdc_source_wait_streaming_start_timeout = <larger_value>;"
                 );
             }
         }
--- a/src/meta/service/src/stream_service.rs
+++ b/src/meta/service/src/stream_service.rs
@@ -203,7 +203,7 @@ impl StreamManagerService for StreamServiceImpl {
         let canceled_jobs = self
             .stream_manager
             .cancel_streaming_jobs(table_ids.into_iter().map(JobId::from).collect_vec())
-            .await
+            .await?
             .into_iter()
             .map(|id| id.as_raw_id())
             .collect_vec();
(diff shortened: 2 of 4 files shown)


zwang28:
> e2e_test/source_legacy/cdc/mongodb/mongodb_basic.slt

The mongodb cdc test failed.

hzxa21:
```
query I
select count(*) from normalized_users;
----
55

# historical data
query I
select count(*) from users where commit_ts = '1970-01-01 00:00:00+00:00';
----
55
```

The first query passes but the second query returns 0. I think it is related to the timestamp changes in the meta column:
```
CREATE TABLE users (_id JSONB PRIMARY KEY, payload JSONB)
INCLUDE TIMESTAMP as commit_ts
```

hzxa21:
Related to https://github.com/debezium/debezium/pull/6170

The fix is merged in version >= dbz 3.1

.github/workflows/connector-node-integration.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        java: [ '11', '17' ]
+        java: [ '17' ]
```
xiangjinwu: Shall we enable tests against java 21 as well?

zwang28:
> LGTM. Thanks for the efforts!
> 
> I saw that the `patched` comment are removed for the java code diff against dbz. Have we decided to use this [repo](https://github.com/zwang28/debezium/tree/v3.2.4.rw) to track the diff instead of noting that the diff in the codes?

Yes, we'll have a repo hosted in risingwave account for tracking purpose, not this personal one.

wcy-fdu (approved):
Thanks for the efforts!

.github/workflows/connector-node-integration.yml
zwang28: Added java 21.

xiangjinwu:
Are we still waiting for anything before merge?

zwang28:
> Are we still waiting for anything before merge?

~~We'll merge this PR after the release-2.7 code-freeze.~~

UPDATE: we'll include this PR in release-2.7.

xiangjinwu:
```
[ERROR] COMPILATION ERROR :
[INFO] -------------------------------------------------------------
[ERROR] /risingwave/java/connector-node/risingwave-source-cdc/src/main/java/com/risingwave/connector/cdc/debezium/internal/NoDataRecoverySnapshotter.java:[41,70] cannot find symbol
  symbol:   method historyExists()
  location: interface io.debezium.schema.HistorizedDatabaseSchema
[INFO] 1 error
```
merge failure is due to conflict with #22255

PR #23601, opened 2025-10-29
WIP: feat(refresh): implement refresh progress tracking for refresh table (for state switch)

tabVersion:

Added a new module for tracking the progress of refresh operations across parallel actors. This includes the introduction of a `RefreshProgress` structure to monitor the state of each actor during the refresh process. The changes also integrate this tracking into the existing barrier control flow, allowing for better coordination and reporting of refresh completion states. Additionally, updated related files to accommodate the new refresh progress functionality.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/meta/service/src/stream_service.rs
+++ b/src/meta/service/src/stream_service.rs
@@ -536,7 +536,9 @@ impl StreamManagerService for StreamServiceImpl {
             self.barrier_scheduler.clone(),
         );
 
-        let response = refresh_manager.refresh_table(req).await?;
+        let response = refresh_manager
+            .refresh_table(req, &self.env.shared_actor_info)
+            .await?;
 
         Ok(Response::new(response))
     }
--- a/src/meta/src/stream/refresh_manager.rs
+++ b/src/meta/src/stream/refresh_manager.rs
@@ -12,16 +12,25 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::collections::{HashMap, HashSet};
+use std::sync::LazyLock;
+
 use anyhow::anyhow;
-use risingwave_common::catalog::{DatabaseId, TableId};
+use parking_lot::Mutex;
+use risingwave_common::catalog::{DatabaseId, FragmentTypeFlag, TableId};
+use risingwave_meta_model::ActorId;
 use risingwave_meta_model::table::RefreshState;
 use risingwave_pb::catalog::table::OptionalAssociatedSourceId;
 use risingwave_pb::meta::{RefreshRequest, RefreshResponse};
 use thiserror_ext::AsReport;
 
-use crate::barrier::{BarrierScheduler, Command};
+use crate::barrier::{BarrierScheduler, Command, SharedActorInfos};
 use crate::manager::MetadataManager;
 use crate::{MetaError, MetaResult};
+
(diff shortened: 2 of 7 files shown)


src/meta/src/barrier/refresh_progress.rs
```diff
@@ -0,0 +1,401 @@
...
+pub struct RefreshProgressTracker {
+    /// Map from table_id to refresh progress
+    progress_map: HashMap<TableId, RefreshProgress>,
+    /// Map from actor_id to table_id for quick lookup
+    actor_to_table: HashMap<ActorId, TableId>,
+}
+
+impl RefreshProgressTracker {
+    /// Start tracking a new refresh operation
+    pub fn start_refresh(
```
chenzl25: Have we used this method?

src/meta/src/barrier/refresh_progress.rs
```diff
@@ -0,0 +1,401 @@
...
+/// Tracker for all ongoing refresh operations
+///
+/// This tracker ensures that mutations (ListFinish, LoadFinish) are only injected
+/// when ALL parallel actors for a given phase have completed their work.
+#[derive(Default, Debug)]
+pub struct RefreshProgressTracker {
+    /// Map from table_id to refresh progress
+    progress_map: HashMap<TableId, RefreshProgress>,
+    /// Map from actor_id to table_id for quick lookup
+    actor_to_table: HashMap<ActorId, TableId>,
```
chenzl25: I just realize that RefreshProgressTracker needs to handle scaling as well (no matter online scaling or offline scaling), which means we need to know when to update the actor maps to keep it consistent with the latest table's parallelism.

src/meta/src/barrier/refresh_progress.rs
tabVersion: Plan to merge this function with `RefreshManager`

src/meta/src/barrier/refresh_progress.rs
tabVersion: Might be not? As we will cancel the refresh op across recovery, I think we can assume during the run, the actors and parallelism do not change.

chenzl25: OK. To assume the parallelism do not change during the running time, I think we need to ban online scaling to the batch refreshable table cc @shanicky

shanicky: We can skip or ban specific tables during online scaling and manual scaling.

 During offline scaling, is it confirmed that there are no jobs related to those tables?

chenzl25: > During offline scaling, is it confirmed that there are no jobs related to those tables?

Even though there are jobs related to these tables, I think offline scaling is always safe to perform.

tabVersion: The underlying job is always safe to scale. The refresh process will not persist any states, so it is also safe to scale, at the cost of re-running `refresh`.

hzxa21:
Do we merge the changes altogether into https://github.com/risingwavelabs/risingwave/pull/23527 instead so this PR is no longer needed?

tabVersion:
> Do we merge the changes altogether into #23527 instead so this PR is no longer needed?

I messed up the git history in the pr, made some trouble merging into the original one. Will open a new one instead.

PR #23612, opened 2025-10-30
feat(stream,dashboard): improve actor input buffer blocking metrics and add actor busy time panel

kwannoel:

This PR introduces two new sets of derived metrics: `Actor Busy Rate` and `Actor Busy Time (Relative)`.

An actor runtime is occupied by 2 activities:
```
runtime = busy time + idle time
idle = output blocking time + input blocking time
```

Therefore, to get a busy rate, we can make the following calculation:
```
busy rate = 100% - idle rate
                = 100% - output blocking time - input blocking time
```

There are several limitations with this approach as detailed [here](https://www.notion.so/risingwave-labs/RFC-Identifying-Bottleneck-via-Busy-Rate-28bf0d69cb7680f18274c09cbf3a45b8).

First, the input actors may have rather long blocking duration, where they don't report they are blocking. To workaround this, we concurrently report the input blocking metric if some operation is stuck at a 15s interval. After this fix, the input blocking rate has fewer spikes, but they are still present in extreme scenarios.

Second, the input blocking rate was still updated at the actor level granularity. We applied relabelling to the actor metrics after this PR, so the input blocking metric won't get suppressed.

Third, some actors might not have upstreams (e.g. table, source), and some actors might not have downstreams (mv with no downstream, sink). For both these cases, they will not have a matching set of metrics (i.e. have output blocking rate but no input blocking rate and vice versa). Due to this mismatch, prometheus will not derive the busy rate/time for these actors. This is not a problem for root nodes like table and source, because they are typically not the bottleneck. If they are the bottleneck, we can always reference the base `input blocking` and `output blocking` metrics. For terminal nodes, they are commonly the bottlenecks, because they contain things like aggregates (e.g. array_agg). To ensure they still match, we add an `or ...` clause, and always fill it with `0%`.

Fourth, we apply the same normalization which we did for output blocking rate to input blocking rate. Using `avg` is inaccurate because we can have local pre-aggregation (multiple actors aggregated on the same worker node).

Fifth, we provide relative busy time, and not just busy rate. Busy rate can have jitters at times due to the nature of `input blocking rate`, which in our experiments we see can still encounter bursts and jitter during extreme workloads. On the other hand, busy time provides a rather unambiguous graph. To calculate busy time, we normalize it with the maximum busy time, to make the graph easier to read. We don't use idle time, since that doesn't help us identify the bottleneck node(s).

<details><summary><b>Busy Metrics Screenshots</b></summary>

Screenshot 2025-11-07 at 1 31 01 AM

</details>

The above graphs are from [this workload](https://gist.github.com/kwannoel/e7983b633b5632f4ed988708fec27f7d#file-join-complex-downstreams-slt).

<details><summary><b>Stream Graph Screenshot</b></summary>

image

</details>

This workload demonstrates a relatively complex stream graph, where there's a bottleneck due to array aggregation.
It construct a sink into table downstream of the bottleneck. Sink into table makes it hard to solely use a `output backpressure` approach to derive the bottleneck. This is because sink into table will force alignment on its inputs, such that if one input is slow due to some upstream backpressure, another input will be backpressured. In such a scenario, we can't rely on output backpressure, since the slow input manifests as `input backpressure`. If we consult the screenshot above, we maybe mislead to think `t` is the bottleneck.

By using our busy rate approach, we can identify that fragment 10 is the likely bottleneck (although it might not be the root cause). Fragment 10 belongs to `mv_arr_unnest`, and so we can flag this MV and drop it to mitigate.

We have performed[ performance analysis to confirm no performance regression after this PR](https://www.notion.so/risingwave-labs/RFC-Identifying-Bottleneck-via-Busy-Rate-28bf0d69cb7680f18274c09cbf3a45b8?source=copy_link#2a2f0d69cb7680c48d03f3a1540a66cc).

## What's next?

I refrained from doing more involved tuning of these queries on my local cluster. I'd like to test it production clusters to evaluate its effectiveness in a production environment, before deciding if we need to dive deeper into why there's still jitters in input blocking duration. If it works well enough, we can save our efforts.

There are several other cases like large `actor:worker node` which I haven't tested. But I think it's now useful enough for amplification cases.

We probably also need to set the busy time baseline to be by database, rather than at a global level. To keep this PR simple I opt to skip that first. We will need to join on `state table info` and `relation info` to get the database namespacing.

## Checklist

## Documentation

<details>

</details>


kwannoel:
* **#23612** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23612?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#23675](https://github.com/risingwavelabs/risingwave/pull/23675) <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23675?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#23730](https://github.com/risingwavelabs/risingwave/pull/23730) <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23730?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23612?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/stream/src/executor/monitor/streaming_stats.rs
```diff
@@ -352,7 +352,9 @@ impl StreamingMetrics {
                 &["actor_id", "fragment_id", "upstream_fragment_id"],
                 registry
             )
-            .unwrap();
+            .unwrap()
+            // mask the first label `actor_id` if the level is less verbose than `Debug`
```
tabVersion, line 356: as we are going actorless, maybe better remove the dimension?
It is okay to keep here, it is another topic.

src/stream/src/executor/merge.rs
```diff
@@ -248,11 +246,9 @@ impl MergeExecutor {
         );
 
         loop {
-            let msg = barrier_buffer.await_next_message(&mut select_all).await?;
-            metrics
-                .actor_input_buffer_blocking_duration_ns
-                .inc_by(start_time.elapsed().as_nanos() as u64);
-
+            let msg = barrier_buffer
+                .await_next_message(&mut select_all, &metrics)
```
tabVersion, line 251: double check: here measures the input latency, the bottleneck comes from the upstream actors or alignments.

src/stream/src/executor/mod.rs
```diff
@@ -1805,18 +1806,34 @@ impl DispatchBarrierBuffer {
     pub async fn await_next_message(
         &mut self,
         stream: &mut (impl Stream<Item = StreamExecutorResult<DispatcherMessage>> + Unpin),
+        metrics: &ActorInputMetrics,
     ) -> StreamExecutorResult<DispatcherMessage> {
-        tokio::select! {
-            biased;
+        let mut start_time = Instant::now();
+        let interval_duration = Duration::from_secs(15);
```
tabVersion, line 1741: make it a streaming developer config? I feel we may need to tune the value if needs better sensitivity.

tabVersion (approved):
overall LGTM, please also add some comments on the grafana page to help oncallers better understand the panel.

src/stream/src/executor/mod.rs
kwannoel: The default scrape interval for prod is 15s. For finer granularity, we have to use distributed tracing.

src/stream/src/executor/merge.rs
kwannoel: Yes.

src/stream/src/executor/monitor/streaming_stats.rs
kwannoel: Going actorless just means we don't persist actor metadata in the meta store and dynamically generate it. We will still have actor_ids.

kwannoel:
> overall LGTM, please also add some comments on the grafana page to help oncallers better understand the panel.

https://github.com/risingwavelabs/risingwave/pull/23612/commits/8c7d64d089669d3cd3c37cc18c74ce41434b27e4

BugenZhao (approved):
LGTM!

PR #23622, opened 2025-10-31
feat(streaming): better handling of `Update` on conflict in `Materialize`

BugenZhao:

Skip `UpdateDelete` rows when conflict handling behavior is `Overwrite` or `DoUpdateIfNotNull`.

Note that although this can help https://github.com/risingwavelabs/risingwave/issues/22579, the issue is already fixed by sinking `Upsert` chunk in https://github.com/risingwavelabs/risingwave/pull/23581. As a result, the changes of this PR won't take effect in actual usage, because 

- We don't currently have retractable source, thus source won't emit `Update` rows.
- Sink-into-table won't emit `Update` rows either after #23581.
- `UPDATE` statements always operate on full record, thus typically won't hit conflict handling

The statements in doc

> Delete and update operations on the table cannot break the primary key constraint on the table, so the option will not take effect for those cases.

...should still hold, because it cannot be guaranteed that all updates in our system are represented with `Update` operations, so the actual behavior is still undefined.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/executor/mview/materialize.rs
+++ b/src/stream/src/executor/mview/materialize.rs
@@ -407,15 +407,23 @@ impl<S: StateStore, SD: ValueRowSerde> MaterializeExecutor<S, SD> {
                                 self.metrics
                                     .materialize_input_row_count
                                     .inc_by(chunk.cardinality() as u64);
+
                                 // This is an optimization that handles conflicts only when a particular materialized view downstream has no MV dependencies.
                                 // This optimization is applied only when there is no specified version column and the is_consistent_op flag of the state table is false,
-                                // and the conflict behavior is overwrite.
-                                let do_not_handle_conflict = !self.state_table.is_consistent_op()
+                                // and the conflict behavior is overwrite. We can rely on the state table to overwrite the conflicting rows in the storage,
+                                // while outputting inconsistent changes to downstream which no one will subscribe to.
+                                let optimized_conflict_behavior = if let ConflictBehavior::Overwrite =
+                                    self.conflict_behavior
+                                    && !self.state_table.is_consistent_op()
                                     && self.version_column_indices.is_empty()
-                                    && self.conflict_behavior == ConflictBehavior::Overwrite;
+                                {
+                                    ConflictBehavior::NoCheck
+                                } else {
+                                    self.conflict_behavior
+                                };
 
-                                match self.conflict_behavior {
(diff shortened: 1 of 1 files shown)


src/stream/src/executor/mview/materialize.rs
```diff
@@ -407,15 +407,23 @@ impl<S: StateStore, SD: ValueRowSerde> MaterializeExecutor<S, SD> {
...
                                     .materialize_input_row_count
                                     .inc_by(chunk.cardinality() as u64);
+
                                 // This is an optimization that handles conflicts only when a particular materialized view downstream has no MV dependencies.
                                 // This optimization is applied only when there is no specified version column and the is_consistent_op flag of the state table is false,
-                                // and the conflict behavior is overwrite.
-                                let do_not_handle_conflict = !self.state_table.is_consistent_op()
+                                // and the conflict behavior is overwrite. We can rely on the state table to overwrite the conflicting rows in the storage,
+                                // while outputting inconsistent changes to downstream which no one will subscribe to.
+                                let optimized_conflict_behavior = if let ConflictBehavior::Overwrite =
```
BugenZhao, line 415: No logic change. Just refactor.

hzxa21 (approved):
LGTM

src/stream/src/executor/mview/materialize.rs
```diff
@@ -1389,22 +1394,34 @@ impl<SD: ValueRowSerde> MaterializeCache<SD> {
...
+                {
+                    // For `UpdateDelete`s, we skip processing them but directly handle the following `UpdateInsert`
+                    // instead. This is because...
+                    //
+                    // - For `Overwrite`, we only care about the new row.
+                    // - For `DoUpdateIfNotNull`, we don't want the whole row to be deleted, but instead perform
+                    //   column-wise replacement when handling the `UpdateInsert`.
+                    //
+                    // However, for `IgnoreConflict`, we still need to delete the old row first, otherwise the row
+                    // cannot be updated at all.
```
chenzl25, line 1411: I don't quite get this comment, could you please elaborate more on it?

BugenZhao: For `IgnoreConflict`, if we skip the `UpdateDelete` row here, then the following `UpdateInsert` will also be discarded, because the row is still in the storage and we consider it a conflict. Such behavior is not expected.

Issue #23605, opened 2025-10-29
deprcate structs in actor.rs


shanicky:
Can we move this struct to another place? Since it's no longer a model in the meta store.

_Originally posted by @yezizp2012 in https://github.com/risingwavelabs/risingwave/pull/22932#discussion_r2472112625_

PR #23625, opened 2025-10-31
refactor: remove actorstatus enum and simplify actor state handling

### LINKED ISSUE #23604: remove actor status
IIUC there's no such concept of `ActorStatus` after this PR, can we remove this field directly?

_Originally posted by @yezizp2012 in https://github.com/risingwavelabs/risingwave/pull/22932#discussion_r2472150431_

shanicky:
Remove the ActorStatus enum and all related actor state tracking from the meta model and controller layers to simplify the codebase and metadata handling.

- Delete `actor.rs` model file and remove the `status` field from ActorModel
- Remove ActorState enum and `state` field from proto `ActorStatus` message, reserving the field number for backward compatibility
- Update frontend catalog to hardcode actor state as "RUNNING" instead of reading from proto
- Remove all conversions and references to PbActorState and actor status in service and controller code
- Eliminate code rebuilding fragment mappings based on actor states and related tests

This change reflects a design decision to treat all actors as implicitly running and to externalize or remove ephemeral state tracking from metadata, reducing complexity and maintenance overhead.

close #23604 #23605

**Summary**
This PR removes the `ActorStatus` enum and all associated actor state management logic from the meta and frontend codebases. Actor states are no longer tracked dynamically; instead, actors are assumed to be always `"RUNNING"`, simplifying the system’s handling of actor lifecycle and status.

**Details**
- Removed the `ActorState` enum and the `state` field from the `ActorStatus` protobuf message, reserving the field number for backward compatibility.
- Deleted the Rust `actor.rs` model containing `ActorStatus` and the `status` field from `ActorModel`.
- Updated all code references that previously relied on actor states to either omit status or use a hardcoded `"RUNNING"` string (notably in the frontend catalog query and stream service).
- Removed helper functions and logic that filtered or handled actors based on their state (e.g., `rebuild_fragment_mapping_from_actors`).
- Simplified the frontend system catalog `rw_actors` to report a constant actor state, reducing complexity and potential state inconsistencies.
- These changes reflect a design decision to deprecate explicit actor state tracking within the meta layer, likely to reduce overhead and align with updated runtime semantics.

## Checklist

### DIFF
--- a/src/meta/src/controller/fragment.rs
+++ b/src/meta/src/controller/fragment.rs
@@ -94,7 +94,7 @@ pub struct InflightActorInfo {
 }
 
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
-pub struct ActorModel {
+struct ActorInfo {
     pub actor_id: ActorId,
     pub fragment_id: FragmentId,
     pub splits: ConnectorSplits,
@@ -263,11 +263,11 @@ impl CatalogController {
         Ok(fragment)
     }
 
-    pub fn compose_table_fragments(
+    fn compose_table_fragments(
         job_id: JobId,
         state: PbState,
         ctx: Option<PbStreamContext>,
-        fragments: Vec<(fragment::Model, Vec<ActorModel>)>,
+        fragments: Vec<(fragment::Model, Vec<ActorInfo>)>,
         parallelism: StreamingParallelism,
         max_parallelism: usize,
         job_definition: Option<String>,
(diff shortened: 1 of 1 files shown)


src/meta/src/controller/fragment.rs
```diff
@@ -94,6 +92,16 @@ pub struct InflightActorInfo {
     pub splits: Vec<SplitImpl>,
 }
 
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct ActorModel {
```
yezizp2012: Can we remove the Model key word as well?

yezizp2012 (approved):
Lgtm

src/meta/src/controller/fragment.rs
shanicky: Should we just call it "Actor"? It feels a bit too general.

shanicky: Renamed it to ActorInfo and strictly limited the scope.

PR #23632, opened 2025-11-03
feat(stream): add license check for locality backfill

chenzl25:

- Add license check for locality backfill. If a streaming query uses more than 5 locality provider, it needs a license. 

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/optimizer/plan_visitor/locality_provider_counter.rs
+++ b/src/frontend/src/optimizer/plan_visitor/locality_provider_counter.rs
@@ -1,0 +1,44 @@
+// Copyright 2025 RisingWave Labs
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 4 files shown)


lmatz (approved):
Thanks

src/frontend/src/optimizer/plan_visitor/locality_provider_counter.rs
```diff
@@ -0,0 +1,44 @@
...
+impl StreamPlanVisitor for LocalityProviderCounter {
+    type Result = usize;
+
+    type DefaultBehavior = impl DefaultBehavior<Self::Result>;
+
+    fn default_behavior() -> Self::DefaultBehavior {
+        Merge(|a: usize, b| a + b)
+    }
+
+    fn visit_stream_locality_provider(
```
BugenZhao, line 38: In a graph plan, do we ensure that each plan node is visited at most once?

chenzl25: Actually it would be more than once, if it is a DAG with shared plans, but we can just count it multiple time as it is a "complex" query.

PR #23634, opened 2025-11-03
feat(optimizer): strip unnecessary stream key from the output of `TopN` when the `LIMIT` is 1

BugenZhao:

When user specifies `ORDER BY .. LIMIT 1`, we will directly use the group key (or `[]` if it's global top-1) as the stream key, instead of the union of the group key and the input stream key, because it's guaranteed there's at most one row per group key.

In this PR, we further strip the stream key column added by `input.logical_rewrite_for_stream()` from the output of `TopN`, since it's unnecessary. This ensures that the schema of `Top-1` is exactly what user specifies, so that it can also be used for the inner side of dynamic filter.

Fix https://github.com/risingwavelabs/risingwave/issues/23513

## Checklist

## Documentation

<details>

</details>


BugenZhao:
> How about adding an e2e test as well to ensure it works as expected.

Sounds good. I just realized there are some commented-out lines for using top-1 as the inner side of dynamic filter in `dynamic_filter.slt`. Enabled them.

PR #23643, opened 2025-11-03
chore(executor): unify functions in gap fill and replace / with div_floor

guluguluhhhh:

## Checklist

## Documentation

<details>
unify functions in gap fill and replace / with div_floor

</details>

### DIFF
--- a/src/common/src/gap_fill.rs
+++ b/src/common/src/gap_fill.rs
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use num_integer::Integer;
+
 use crate::types::{Datum, DatumRef, Decimal, ScalarImpl, ScalarRefImpl};
 
 /// Strategy for filling gaps in time series data.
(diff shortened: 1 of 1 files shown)


guluguluhhhh:
Hi @kwannoel @yuhao-su, could you please help review this PR?

src/common/src/gap_fill.rs
```diff
@@ -0,0 +1,95 @@
...
+    };
+    if steps == 0 {
+        return None;
+    }
+    match (s1, s2) {
+        (ScalarRefImpl::Int16(v1), ScalarRefImpl::Int16(v2)) => {
+            // Convert to i64 to avoid overflow, then convert back
+            let diff = (v2 as i64) - (v1 as i64);
+            Some(ScalarImpl::Int16(
+                Integer::div_floor(&diff, &(steps as i64)) as i16,
```
kwannoel, line 52: Why don't we use the native `div_floor`?

src/common/src/gap_fill.rs
```diff
@@ -0,0 +1,95 @@
...
+pub fn apply_interpolation_step(current: &mut Datum, step: &ScalarImpl) {
+    if let Some(curr) = current.as_mut() {
+        match (curr, step) {
+            (ScalarImpl::Int16(v1), ScalarImpl::Int16(v2)) => *v1 += v2,
+            (ScalarImpl::Int32(v1), ScalarImpl::Int32(v2)) => *v1 += v2,
+            (ScalarImpl::Int64(v1), ScalarImpl::Int64(v2)) => *v1 += v2,
+            (ScalarImpl::Float32(v1), ScalarImpl::Float32(v2)) => *v1 += v2,
+            (ScalarImpl::Float64(v1), ScalarImpl::Float64(v2)) => *v1 += v2,
+            (ScalarImpl::Decimal(v1), ScalarImpl::Decimal(v2)) => *v1 = *v1 + *v2,
+            _ => (),
```
kwannoel, line 92: Another issue with this approach for floating points. Consider if our gap has the following parameters:
```
start: 0.1
steps: 10
end: 0.2
```

We can write the following example:
```
fn main() {
  let start = 0.1;
  let end = 0.2;
  let steps: f64 = 10.0;
  let step_size = (end - start) / steps;
  {
      let mut end = start;
      for _ in 0..10 {
          end += step_size;
      }
      println!("end: {end}");     
  }

}
```

Because of binary precision issues, we will get the following output for addition:
```
end: 0.20000000000000007
```

You can see it will overflow the end bounds. For the last step let's make sure to clamp it to the end bound. You can do this in a separate PR.

kwannoel: In the PR, please also add an SLT test to verify this case.

src/common/src/gap_fill.rs
guluguluhhhh:
```
error[E0658]: use of unstable library feature `int_roundings`
  --> src/common/src/gap_fill.rs:49:41
   |
49 |             Some(ScalarImpl::Int16(diff.div_floor(steps as i64) as i16))
   |                                         ^^^^^^^^^
   |
   = note: see issue #88581 <https://github.com/rust-lang/rust/issues/88581> for more information
   = help: add `#![feature(int_roundings)]` to the crate attributes to enable
   = note: this compiler was built on 2025-10-09; consider upgrading it if it is out of date
```
Seems like the native `div_floor()` is currently unstable (requires `#![feature(int_roundings)]`). I used `Integer::div_floor` from the `num-integer` crate to keep the code stable-compatible. 

Happy to switch to the native method if we're okay with adding the feature flag.

yuhao-su (approved):

src/common/src/gap_fill.rs
guluguluhhhh: Thanks! I've understood the binary precision issues. Maybe we should use LERP to interpolate each step, which is like:
```
fn main() {
    let start = 0.1;
    let end = 0.2;
    let steps: f64 = 10.0;
    {
        let mut current = start;
        for i in 1..=10 {
            let t = i as f64 / steps;
            current = start + (end - start) * t;
        }
        println!("end: {current}");
    }
}
```
Each step is computed independently, so precision errors won't accumulate. We will get the output:
```
end: 0.2
```

 What do you think? @kwannoel

kwannoel: Nevermind, I realized it's unlikely to happen. The end bound is exclusive, and we won't gapfill on the end bound.

Issue #23394, opened 2025-10-09
new exactly-once sink design


wenym1:
### Background

Currently, only iceberg-sink supports exactly-once semantic. The logic to support exactly once is embedded inside the implementation of iceberg-sink. It works like the following:

IcebergSinkWriter receives checkpoint barrier from LogReader
-> finish file writer and send CommitRequest to SinkCoordinator
-> SinkCoordinator collects CommitRequests from all parallelisms
-> SinkCoordinator calls IcebergSinkCommitter::commit
-> IcebergSinkCommitter wait for checkpoint of the epoch to commit
-> IcebergSinkCommitter persists the metadata to commit with epoch in a meta store table
-> IcebergSinkCommitter idempotently commit the metadata to external iceberg table
-> IcebergSinkCommitter mark the epoch as committed, and return from IcebergSinkCommitter::commit
-> SinkCoordinator send CommitResponse to IcebergSinkWriter
-> IcebergSinkWriter ack on the checkpoint barrier

In recovery, in IcebergSinkCommitter::init, it checks meta store table for uncommitted metadata, and idempotently commits the pending metadata, and mark the epoch as committed.

The current implementation has the following drawbacks:
1. it requires that the sink has enabled sink_decouple, because it acks the checkpoint barrier after the checkpoint has finished, and without the log store enabled by decoupled sink, there will be deadlock.
2. the finish of IcebergSinkCommitter::commit should wait for the checkpoint, which greatly increases the latency and the IcebergSinkWriter is actually blocked by the checkpoint of upstream streaming job checkpoint.
3. the logic to support exactly-once (e.g., store and update meta store table) is embedded inside the implementation of IcebergSinkCommitter, which makes it hard to be reused for other sinks to support exactly-once.

### Design

In the new design, the 2-phase commit semantic will be inherently expressed in the 

```
#[async_trait]
pub trait SinkCommitCoordinator {
    async fn init(&mut self) -> Result<Option<u64>>;
    async fn pre_commit(
        &mut self,
        epoch: u64,
        metadata: Vec<SinkMetadata>,
        add_columns: Option<Vec<Field>>,
    ) -> Result<Option<Bytes>>;
    async fn commit(&mut self, commit_metadata: Bytes) -> Result<()>;
    async fn abort(&mut self, commit_metadata: Bytes); // no error
}
```

We will have a meta store table to store the metadata of pending uncommitted epoch. The meta store table will have the following schema
```
[...]

PR #23657, opened 2025-11-04
feat(sink): exactly once sink on coordination process

Chiro11:

Related issue: #23394

Based on the issue's design, implement the exactly once sink using coordination framework.
And the logic for the iceberg sink has been migrated to this framework.

Currently, the modifications for integration tests temporarily reused the original Iceberg test code. In the future, more systematic 2PC tests will be added.

From a usage perspective, two traits are introduced here.
```
#[async_trait]
pub trait SinglePhaseCommitCoordinator {
    async fn init(&mut self) -> Result<()>;

    async fn commit(
        &mut self,
        _epoch: u64,
        _metadata: Vec<SinkMetadata>,
        _add_columns: Option<Vec<Field>>,
    ) -> Result<()>;
}

#[async_trait]
pub trait TwoPhaseCommitCoordinator {
    async fn init(&mut self) -> Result<()>;

    async fn pre_commit(
        &mut self,
        _epoch: u64,
        _metadata: Vec<SinkMetadata>,
        _add_columns: Option<Vec<Field>>,
    ) -> Result<Vec<u8>>;

    async fn commit(&mut self, _epoch: u64, _commit_metadata: Vec<u8>) -> Result<()>;

    async fn abort(&mut self, _epoch: u64, _commit_metadata: Vec<u8>);
}
```

When calling `Sink::new_coordinator`,
```
pub trait Sink {
    ...
    async fn new_coordinator(
        &self,
        _iceberg_compact_stat_sender: Option<UnboundedSender<IcebergSinkCompactionUpdate>>,
    ) -> Result<SinkCommitCoordinator>;
}
```
the returned type will be `enum SinkCommitCoordinator`.
```
pub enum SinkCommitCoordinator {
    SinglePhase(Box<dyn SinglePhaseCommitCoordinator + Send + 'static>),
    TwoPhase(Box<dyn TwoPhaseCommitCoordinator + Send + 'static>),
}
```

## Checklist

## Documentation

<details>

</details>


src/connector/src/sink/mod.rs
```diff
@@ -813,20 +811,51 @@ pub type SinkCommittedEpochSubscriber = Arc<
...
+pub enum SinkCommitStrategy {
+    /// for at-least-once semantics
+    SinglePhase,
+    /// for exactly-once semantics
+    TwoPhase,
+}
+
 #[async_trait]
-pub trait SinkCommitCoordinator {
+pub trait SinkCommitCoordinator: Send {
```
wenym1: How about split it into two separate traits, and return an enum in `Sink::new_coordinator`? And then we may remove the `SinkCommitCoordinator` associated type in `Sink`.

src/connector/src/sink/coordinate.rs
```diff
@@ -201,61 +201,56 @@ impl<W: SinkWriter<CommitMetadata = Option<SinkMetadata>>> LogSinker for Coordin
...
                             log_reader.truncate(TruncateOffset::Barrier { epoch })?;
-                        } else {
-                            let metadata = sink_writer.barrier(false).await?;
-                            if let Some(metadata) = metadata {
-                                warn!(?metadata, "get metadata on non-checkpoint barrier");
-                            }
+                            return pending().await;
                         }
+                        log_reader.truncate(TruncateOffset::Barrier { epoch })?;
                     } else {
```
wenym1, line 259: Any reason to remove this branch?

src/meta/src/manager/sink_coordination/manager.rs
```diff
@@ -510,6 +512,7 @@ mod tests {
                             // validate the start request
                             assert_eq!(param, expected_param);
                             CoordinatorWorker::execute_coordinator(
+                                DatabaseConnection::Disconnected,
```
wenym1, line 1115: Can we add some similar unit tests to cover the execution of 2PC? Besides, we may also add simulation integration tests for 2PC, like other sink type.

src/connector/src/sink/coordinate.rs
Chiro11: Never mind, reverted.

src/connector/src/sink/mod.rs
Chiro11: Refined now.

Chiro11:
@yezizp2012 @wcy-fdu Please take a look if the meta-table migration logic is correct.

src/connector/src/sink/mod.rs
```diff
@@ -813,53 +810,43 @@ pub type SinkCommittedEpochSubscriber = Arc<
...
-    /// the set of metadata. The metadata is serialized into bytes, because the metadata is expected
-    /// to be passed between different gRPC node, so in this general trait, the metadata is
-    /// serialized bytes.
-    async fn commit(
+pub trait SinglePhaseCommitCoordinator {
+    /// Initialize the sink committer coordinator.
+    async fn init(&mut self) -> Result<()>;
+
+    /// Commit directly using single-phase strategy.
+    async fn commit_directly(
```
wenym1: Since we have separate into two traits, we can name it `commit` directly. 

nit: no need to add `_` before the var name.

src/connector/src/sink/utils.rs
```diff
@@ -68,12 +67,14 @@ pub(crate) mod dummy {
     #[allow(dead_code)]
     pub struct FeatureNotEnabledCoordinator<S: FeatureNotEnabledSinkMarker>(PhantomData<S>);
     #[async_trait::async_trait]
-    impl<S: FeatureNotEnabledSinkMarker> SinkCommitCoordinator for FeatureNotEnabledCoordinator<S> {
-        async fn init(&mut self, _subscriber: SinkCommittedEpochSubscriber) -> Result<Option<u64>> {
+    impl<S: FeatureNotEnabledSinkMarker> SinglePhaseCommitCoordinator
```
wenym1: Seems unnecessary to have this `FeatureNotEnabledCoordinator` after the associated type of the trait is removed.

src/meta/src/barrier/context/recovery.rs
```diff
@@ -69,6 +71,55 @@ impl GlobalBarrierWorkerContextImpl {
         Ok(())
     }
 
+    async fn resolve_pending_sink_state(&self, database_id: Option<DatabaseId>) -> MetaResult<()> {
```
wenym1:
```suggestion
abort_dirty_pending_sink_state(&self, database_id: Option<DatabaseId>) -> MetaResult<()> {
```

src/meta/src/barrier/context/recovery.rs
```diff
@@ -69,6 +71,55 @@ impl GlobalBarrierWorkerContextImpl {
...
+            .await?;
+
+        let mut to_abort_records: HashMap<SinkId, Vec<u64>> = HashMap::new();
+
+        for (sink_id, table_ids) in sink_with_state_tables {
+            let Some(table_id) = table_ids.first() else {
+                return Err(anyhow!("no state table id in sink: {}", sink_id).into());
+            };
+
+            self.hummock_manager
```
wenym1, line 111: Seems cleaner to only resolve the committed epoch of each sink here, and then pass map<SinkId, committed epoch> to `abort_pending_sink_epochs`.

src/connector/src/sink/iceberg/mod.rs
```diff
@@ -1743,119 +1776,64 @@ impl SinkCommitCoordinator for IcebergSinkCommitter {
             .into());
         }
 
+        // Check snapshot limit before proceeding with commit
```
wenym1: I think it's for more committed snapshot, and therefore better to move it to `commit_inner`.

src/connector/src/sink/iceberg/mod.rs
```diff
@@ -1743,119 +1776,64 @@ impl SinkCommitCoordinator for IcebergSinkCommitter {
...
         let write_results: Vec<IcebergCommitResult> = metadata
             .iter()
             .map(IcebergCommitResult::try_from)
             .collect::<Result<Vec<IcebergCommitResult>>>()?;
 
         // Skip if no data to commit
         if write_results.is_empty() || write_results.iter().all(|r| r.data_files.is_empty()) {
-            tracing::debug!(?epoch, "no data to commit");
-            return Ok(());
+            return Ok(None);
```
wenym1, line 1992: May keep the debug log?

src/connector/src/sink/iceberg/mod.rs
```diff
@@ -1743,119 +1776,64 @@ impl SinkCommitCoordinator for IcebergSinkCommitter {
...
+                .any(|r| r.partition_spec_id != expect_partition_spec_id)
         {
             return Err(SinkError::Iceberg(anyhow!(
                 "schema_id and partition_spec_id should be the same in all write results"
             )));
         }
 
-        // Check snapshot limit before proceeding with commit
-        self.wait_for_snapshot_limit().await?;
+        // Load the latest table to avoid concurrent modification with the best effort.
```
wenym1: Seems copied from previous implementation. Not sure if the reload is necessary, because it seems heavy, but we only reload to generate the unique snapshot id in slightly better efforts. @chenzl25 @Li0k What do you think?

src/meta/src/manager/sink_coordination/coordinator_worker.rs
```diff
@@ -523,24 +535,168 @@ impl CoordinatorWorker {
...
+        db: &DatabaseConnection,
+        sink_id: SinkId,
+        coordinator: &mut SinkCommitCoordinator,
+    ) -> anyhow::Result<Option<u64>> {
+        match coordinator {
+            SinkCommitCoordinator::SinglePhase(coordinator) => {
+                coordinator.init().await?;
+                Ok(None)
+            }
+            SinkCommitCoordinator::TwoPhase(coordinator) => {
```
wenym1: The propagation of the first barrier of a newly create job and its checkpoint is expected not to be blocked. Therefore, when we see that the epoch list is empty, maybe we can get the first committed epoch from the CommittedEpochSubscriber, and then initialize a committed row in meta store table, and then in later code we can always treat like we have an initialized committed epoch.

src/meta/src/manager/sink_coordination/coordinator_worker.rs
```diff
@@ -523,24 +535,168 @@ impl CoordinatorWorker {
...
+                            &db,
+                            sink_id as _,
+                            epoch,
+                            commit_metadata.clone(),
+                        )
+                        .await?;
+                        self.handle_manager
+                            .ack_commit(epoch, commit_requests.handle_ids)?;
+                        // Get the latest hummock_committed_epoch and the receiver
+                        let (hummock_committed_epoch, mut rw_futures_utilrx) =
```
wenym1: It's expected to call `subscriber` for only once and store the rx, and then received newly committed epoch from the rx.

src/meta/src/manager/sink_coordination/coordinator_worker.rs
```diff
@@ -523,24 +535,168 @@ impl CoordinatorWorker {
...
+                            .pre_commit(epoch, metadatas, first_add_columns)
+                            .await?;
+                        persist_pre_commit_metadata(
+                            &db,
+                            sink_id as _,
+                            epoch,
+                            commit_metadata.clone(),
+                        )
+                        .await?;
+                        self.handle_manager
```
wenym1, line 770: Better to separate the handling logic of `pre_commit` and `commit`. We can have two queue to store the pending metadata, one for uncommitted epoch, and one for committed epoch. And then we can introduce two more async events, first is new epoch committed, and second is retry backoff time passed. On new epoch committed, move the corresponding item from uncommitted epoch queue to committed epoch queue. On retry backoff time passed, take the first items in the committed epoch queue and do the commit.

Here after `pre_commit`, we only add pending metadata to the uncommitted epoch queue, and then continue the loop.

src/meta/src/manager/sink_coordination/coordinator_worker.rs
```diff
@@ -523,24 +535,168 @@ impl CoordinatorWorker {
...
+                if ordered_metadata.is_empty() {
+                    Ok(None)
+                } else {
+                    let mut last_committed_epoch = None;
+                    let mut aborted_epochs = vec![];
+                    for (epoch, state, metadata) in ordered_metadata {
+                        assert!(
+                            last_committed_epoch.is_none() || epoch > last_committed_epoch.unwrap()
+                        );
+                        match state {
```
wenym1: Based on the idea of having two queues to store pending metadata mentioned in another commit, here during initialization, we may only call abort on the aborted epoch, and for pending epochs, we may only add the metadata to the queue for committed epoch, and then we do commit in the same place as the processing logic of handling new epoch.

src/connector/src/sink/mod.rs
```diff
@@ -813,53 +810,43 @@ pub type SinkCommittedEpochSubscriber = Arc<
         + 'static,
 >;
 
+pub enum SinkCommitCoordinator {
```
wenym1, line 814: We can implement `From<XxxPhaseCoordinator>` for `SinkCommitCoordinator` so that we can use `.into()` to create it.

wenym1 (commented):
Coordination worker part still needs some improvement. Rest LGTM. Thanks for the PR! @chenzl25 @Li0k  @wcy-fdu can you take a look at iceberg part?

src/connector/src/sink/iceberg/mod.rs
chenzl25: I think we should move this reload from `pre_commit` to the `commit`, since there is a gap between `pre_commit` and `commit`. The reload table here is to avoid concurrent modification with the best effort, so the best place should be right ahead of the actual commit. To generate a unique snapshot id, I think using a not latest table is enough since, internally the snapshot id is an Uuid, table is used to guarantee the uuid is unique(though I don't think it would be duplicated anyway).

src/connector/src/sink/iceberg/mod.rs
Chiro11: Already logged in the external function.

src/connector/src/sink/mod.rs
Chiro11: It seems difficult because a type might implement both traits at the same time.

wenym1: I didn't mean implement it for the coordinator of each specific sink. I meant implementing it for `BoxXxxPhaseCoordinator`, so that we don't have to write the whole `SinkCommitCoordinator::SinglePhase(Box::new(coordinator))`, but only `Box::new(coordinator).into()`.

Chiro11: That's not what I mean. But if you do it this way,
```
let coordinator = SnowflakeSinkCommitter::new(
    self.config.clone(),
    &self.schema,
    &self.pk_indices,
    self.is_append_only,
)?;
Ok(Box::new(coordinator).into())
```
it won't compile because the compiler won't consider `Box<SnowflakeSinkCommitter>` to be a `BoxSinglePhaseCoordinator.` You have to explicitly declare it:
```
let boxed = Box::new(coordinator) as Box<dyn SinglePhaseCommitCoordinator>;
```
If I understand correctly. And this is not any simpler than before.

wenym1: Oh I see. Then let's keep the current way.

src/meta/src/manager/sink_coordination/coordinator_worker.rs
Chiro11: Based on the current code, it doesn't seem easy to know which epoch this **first barrier** corresponds to. This wasn't passed when creating stream handle at the beginning. It seems to involve a lot of code changes, but the benefits are not obvious.

wenym1: What I meant was that, when there is no committed epoch found when the coordinator is initialized, the first epoch we can get from the epoch subscriber must be the epoch of the first barrier.

But now I realize that this is only correct when decouple is disabled. When decouple is enabled, the first epoch from the subscriber can be any later committed epoch. 

Can just skip this comment.

src/meta/src/controller/catalog/mod.rs
```diff
@@ -726,6 +727,96 @@ impl CatalogController {
...
+        let query = Fragment::find()
+            .select_only()
+            .columns([fragment::Column::JobId, fragment::Column::StateTableIds])
+            .filter(
+                fragment::Column::JobId
+                    .is_in(sink_ids)
+                    .and(FragmentTypeMask::intersects(FragmentTypeFlag::Sink)),
+            );
+
+        let rows: Vec<(JobId, I32Array)> = query.into_tuple().all(&inner.db).await?;
```
wenym1: Can use `TableIdArray` to avoid later conversion.

src/meta/src/controller/catalog/mod.rs
```diff
@@ -726,6 +727,96 @@ impl CatalogController {
...
+        for (sink_id, committed_epoch) in sink_committed_epoch {
+            pending_sink_state::Entity::update_many()
+                .col_expr(
+                    pending_sink_state::Column::SinkState,
+                    Expr::value(pending_sink_state::SinkState::Aborted),
+                )
+                .filter(
+                    pending_sink_state::Column::SinkId
+                        .eq(sink_id)
+                        .and(pending_sink_state::Column::Epoch.lt(committed_epoch as i64)),
```
wenym1: Should it be `gt`?

src/meta/src/barrier/context/recovery.rs
```diff
@@ -56,6 +57,8 @@ impl GlobalBarrierWorkerContextImpl {
             .reset_all_refresh_jobs_to_idle()
             .await?;
 
+        self.abort_dirty_pending_sink_state(database_id).await?;
```
wenym1: I think we should ensure that all sink coordinators of the database has stopped before we call this method, and otherwise there can be new rows written concurrently. But in current code path, we call sink_manager.reset() after reload runtime info, which is later than calling this method. Besides, we may need support resetting the coordinators of specific database on database recovery.

src/tests/simulation/tests/integration_tests/sink/exactly_once/mod.rs
```diff
@@ -38,10 +38,10 @@ async fn test_exactly_once_sink_basic() -> Result<()> {
 }
 
 #[tokio::test]
-async fn test_exactly_once_sink_corner_case() {
+async fn test_exactly_once_sink_corner_case() -> Result<()> {
     // If, during a re-commit, a new snapshot_id is generated instead of using the previously persisted snapshot_id, it can lead to duplicate data when consecutive "23" appears in err_events.
-    let err_rate_list = vec![0.0, 0.0, 0.3, 0.3, 0.0];
-    let _ = test_exactly_once_sink_inner(err_rate_list).await;
+    let err_rate_list = vec![0.0, 0.0, 0.0, 0.3, 0.0];
```
wenym1: Any reason to adjust the err rate?

src/meta/src/manager/sink_coordination/coordinator_worker.rs
```diff
@@ -523,24 +713,105 @@ impl CoordinatorWorker {
...
+            }
+            SinkCommitCoordinator::TwoPhase(coordinator) => {
+                let ordered_metadata = list_sink_states_ordered_by_epoch(db, sink_id as _).await?;
+
+                if ordered_metadata.is_empty() {
+                    Ok(None)
+                } else {
+                    let mut last_committed_epoch = None;
+                    let mut aborted_epochs = vec![];
+                    for (epoch, state, metadata) in ordered_metadata {
```
wenym1: I think every time we mark a row as committed, we can delete the previous committed row. So that we can ensure that there is at most one committed epoch for each sink. 

And then here we after reload all metadata, we can use the first epoch as last_committed_epoch if its state is committed.

wenym1: Besides, we should have a stronger sanity check here that all epochs with Aborted state should be greater than pending epochs.

src/meta/src/manager/sink_coordination/coordinator_worker.rs
```diff
@@ -104,11 +110,128 @@ impl<R> AligningRequests<R> {
     }
 }
 
+struct TwoPhaseCommitHandler {
+    curr_hummock_committed_epoch: u64,
+    rw_futures_util_rx: UnboundedReceiver<u64>,
```
wenym1: Please have a meaningful name for it.

src/meta/src/manager/sink_coordination/coordinator_worker.rs
```diff
@@ -212,8 +335,11 @@ impl CoordinationHandleManagerEvent {
 }
 
 impl CoordinationHandleManager {
-    async fn next_event(&mut self) -> anyhow::Result<(HandleId, CoordinationHandleManagerEvent)> {
-        {
+    async fn next_event(
```
wenym1: `CoordinationHandleManager` should only be responsible for communication with the handles between sink writer, and should not handle logic of 2pc here, and therefore better to put the `two_phase_handler` separate to `CoordinationHandleManager`. 

Previously in the outer event loop we do `self.handle_manager.next_event()` because `handle_manager` is the only source of events. However, after this PR, we have more events, and then we should not wait for events only from `self.handle_manager.next_event()`, but also from `two_phase_handler`.

Besides, it's cleaner to decouple the emission of events and the heavy handling logic of the events. Current implementation greatly couple them together, where we pass the coordinator and handle the heavy commit inside this `next_event` method.

We can implement it in this way.

```
enum CoordinatorWorkerEvent {
    HandleManagerEvent(HandleId, CoordinationHandleManagerEvent),
    CommitEpoch(u64, Vec<u8>),
}

fn next_event(
    handle_manager: Xxx,
    two_phase_handler: Xxx,
) -> CoordinatorWorkerEvent {
    select! (
        handle_manager.next_event() => {
            CoordinatorWorkerEvent::HandleManagerEvent
        }
        two_phase_handler.next_epoch_to_commit() => {
            CoordinatorWorkerEvent::CommitEpoch
        }
        ...
    )
}
```

and then in the outer event loop
```
loop {
    match next_event(..) {
         CoordinatorWorkerEvent::HandleManagerEvent => {
             ..
         }
         CoordinatorWorkerEvent::CommitEpoch(epoch, metadata) => {
             match coordinator.commit(epoch, metadata).await {
                 Ok(()) => {
                     two_phase_handler.ack_committed(epoch).await;
                 }
                 Err(err) => {
                     two_phase_handler.fail_committed(epoch).await;
                 }
             }
         }
    }
}
```

src/meta/src/manager/sink_coordination/coordinator_worker.rs
```diff
@@ -104,11 +110,128 @@ impl<R> AligningRequests<R> {
...
+                }
+            }
+        }
+    }
+
+    fn push_pending(&mut self, epoch: u64, metadata: Vec<u8>) {
+        self.pending_epochs.push_back((epoch, metadata));
+    }
+
+    // Only used when initializing from store.
```
wenym1: This is not true. When we receive new job_committed_epoch from the committed epoch rx, we should move from `self.pending_epochs` to `self.prepared_epochs`, and we can only take things in `self.prepared_epochs` to commit.

src/meta/src/manager/sink_coordination/coordinator_worker.rs
```diff
@@ -104,11 +110,128 @@ impl<R> AligningRequests<R> {
     }
 }
 
+struct TwoPhaseCommitHandler {
+    curr_hummock_committed_epoch: u64,
+    rw_futures_util_rx: UnboundedReceiver<u64>,
+    pending_epochs: VecDeque<(u64, Vec<u8>)>,
+    prepared_epochs: VecDeque<(u64, Vec<u8>)>,
+    delay: Duration,
+    max_backoff: Duration,
```
wenym1: Can store an iterator of backoff sleep future like `get_retry_backoff_strategy` instead of manually implementing one. We can store the state of backoff retrying as `Option<(backoff_future, iter of backoff future)>`.

src/meta/src/manager/sink_coordination/coordinator_worker.rs
```diff
@@ -104,11 +110,128 @@ impl<R> AligningRequests<R> {
...
+                        "Hummock committed epoch sender closed unexpectedly"
+                    ));
+                }
+            }
+        }
+
+        Ok(epoch)
+    }
+
+    async fn next_to_commit(&mut self) -> anyhow::Result<(u64, Vec<u8>)> {
```
wenym1, line 157: The implementation can be cleaner.

```
async fn next_to_commit(&mut self) -> anyhow::Result<(u64, Vec<u8>)> {
    let wait_backoff = async {
        if let Some(backoff_future) = &mut self.backoff_future {
            backoff_future.await;
            self.backoff_future = None;
        }
    };
    loop {
        wait_backoff if !self.prepared_epochs.is_empty() => {
            self.prepared_epochs.pop_front()
        }
        job_committed_epoch_rx.recv() => {
            // some check on epoch
            self.prepared_epochs.push_back(self.pending_epochs.pop_front());
        }
    }
}
```

src/tests/simulation/tests/integration_tests/sink/exactly_once/mod.rs
Chiro11: Otherwise, this test will time out. In any case, the purpose of this test is to check the idempotency of snapshot_id.

src/meta/src/manager/sink_coordination/coordinator_worker.rs
Chiro11: It's better to use the previous logic of `next_prepared_epoch`, instead of directly waiting on `job_committed_epoch_rx`, because for new pre-commit epoch, we always add it to the pending queue. The newly added epoch may have already been committed, even if we haven't received a new committed epoch from rx.

Furthermore, if we consider placing a new pre-commit epoch into the pending or prepared queue based on `curr_committed_epoch`, there might still be out-of-order issues. Considering that we always need to commit in sequence, a simpler choice is to ensure all epochs are first added to the pending queue and then to the prepared queue.

wenym1: > It's better to use the previous logic of `next_prepared_epoch`, instead of directly waiting on `job_committed_epoch_rx`, because for new pre-commit epoch, we always add it to the pending queue. The newly added epoch may have already been committed, even if we haven't received a new committed epoch from rx.
> 
> Furthermore, if we consider placing a new pre-commit epoch into the pending or prepared queue based on `curr_committed_epoch`, there might still be out-of-order issues. Considering that we always need to commit in sequence, a simpler choice is to ensure all epochs are first added to the pending queue and then to the prepared queue.

I don't get how there can be out-of-order issues? The code snippet above is over-simplified. To address the case that job epoch has been committed earlier than coordinator pre-commit, I think we can just store the latest job_committed_epoch. 

When a new epoch gets pre-committed, if it's above the latest job_committed_epoch, add it to `self.pending_epochs`, and otherwise add it to `self.prepared_epochs` with an assertion that `self.pending_epochs.is_empty()`. When we receive new later `job_committed_epoch`, move the epochs `>= job_committed_epoch` from `self.pending_epochs` one by one to `self.prepared_epochs`. In this way there won't be out of order issue.

Chiro11: I think you're right, but it's not obvious to someone who isn't familiar with the whole logic. Overall, the complexity of the two methods doesn't differ much. `next_prepared_epoch` is a very clear method name.

Chiro11: I just thought: for such an unbounded-channel, if there are no sink data for a long time, new epochs might keep piling up at the channel's tx side. Besides memory issues, this could mean that the next commit would need to process a large backlog of epochs all at once. Maybe this is an advantage of it. I'll modify.

src/tests/simulation/tests/integration_tests/sink/exactly_once/mod.rs
wenym1: Timeout can be caused by data loss, since the finish condition is to wait for specific row count, and this is one of the purpose of the test, instead of only testing exactly once. Can you confirm that in the original setting the test can eventually finish if we increase the timeout?

Chiro11: Fixed. There was a bug in previous backoff implementation, that would cause future starvation, but in the first basic test, the other error position triggered rewind in sink-executor, which hide this error.

src/meta/src/barrier/context/mod.rs
```diff
@@ -99,6 +100,11 @@ pub(super) trait GlobalBarrierWorkerContext: Send + Sync + 'static {
         &self,
         refresh_finished_table_job_ids: Vec<JobId>,
     ) -> impl Future<Output = MetaResult<()>> + Send + '_;
+
```
wenym1: I think we can call it in the current trait method `reload_xxx_info` before we mark the dirty epochs as aborted instead of introducing a new trait method.

src/meta/src/controller/catalog/list_op.rs
```diff
@@ -336,4 +336,23 @@ impl CatalogController {
...
             .collect())
     }
+
+    pub async fn list_sink_ids(&self, database_id: Option<DatabaseId>) -> MetaResult<Vec<SinkId>> {
+        let inner = self.inner.read().await;
+
+        let sink_ids: Vec<SinkId> = Sink::find()
+            .select_only()
+            .column(sink::Column::SinkId)
+            .join(JoinType::InnerJoin, sink::Relation::Object.def())
```
wenym1: When database_id is none, we don't need to join the object table.

src/meta/src/manager/sink_coordination/coordinator_worker.rs
```diff
@@ -104,6 +112,117 @@ impl<R> AligningRequests<R> {
...
+                }
+
+                recv_epoch = self.job_committed_epoch_rx.recv() => {
+                    let Some(recv_epoch) = recv_epoch else {
+                        return Err(anyhow!(
+                            "Hummock committed epoch sender closed unexpectedly"
+                        ));
+                    };
+                    self.curr_hummock_committed_epoch = recv_epoch;
+                    while let Some((epoch, _)) = self.pending_epochs.front() && *epoch <= recv_epoch {
```
wenym1: Can use `pop_front_if`.

src/meta/src/manager/sink_coordination/coordinator_worker.rs
```diff
@@ -104,6 +112,117 @@ impl<R> AligningRequests<R> {
...
+                        let (epoch, metadata) = self.pending_epochs.pop_front().expect("should have prepared epoch");
+                        self.prepared_epochs.push_back((epoch, metadata));
```
wenym1, line 184: Shall we have sanity check that every time when we push_back to `prepared_epochs` and `pending_epochs`, the pushed epoch is greater than the previous epoch at front?

chenzl25 (approved):
The iceberg part LGTM

src/meta/src/manager/sink_coordination/coordinator_worker.rs
```diff
@@ -434,60 +557,133 @@ impl CoordinatorWorker {
...
 
+    async fn next_event(
+        &mut self,
+        two_phase_handler: Option<&mut TwoPhaseCommitHandler>,
+    ) -> anyhow::Result<CoordinatorWorkerEvent> {
+        // For single-phase coordinator, there is no need to wait.
+        let two_phase_next_fut = match two_phase_handler {
+            Some(handler) => Either::Left(handler.next_to_commit()),
+            None => Either::Right(pending::<_>()),
+        };
```
wenym1:
```suggestion
        async {
            if let Some(handle) = two_phase_handler {
                handler.next_to_commit().await
            } else {
                pending().await
            }
        }
```

src/meta/src/manager/sink_coordination/coordinator_worker.rs
```diff
@@ -523,24 +719,117 @@ impl CoordinatorWorker {
...
+        &mut self,
+        db: &DatabaseConnection,
+        sink_id: SinkId,
+        subscriber: SinkCommittedEpochSubscriber,
+        coordinator: &mut BoxTwoPhaseCoordinator,
+    ) -> anyhow::Result<(Option<u64>, TwoPhaseCommitHandler)> {
+        let ordered_metadata = list_sink_states_ordered_by_epoch(db, sink_id as _).await?;
+        let mut pending_items = vec![];
+
+        let last_committed_epoch = if ordered_metadata.is_empty() {
```
wenym1: I think the handling logic is the same no matter whether `ordered_metadata.is_empty()` or not, so this branch is unnecessary.

src/meta/src/manager/sink_coordination/coordinator_worker.rs
```diff
@@ -523,24 +719,117 @@ impl CoordinatorWorker {
...
+            None
+        } else {
+            let mut metadata_iter = ordered_metadata.into_iter().peekable();
+            let last_committed_epoch = metadata_iter
+                .next_if(|(_, state, _)| matches!(state, SinkState::Committed))
+                .map(|(epoch, _, _)| epoch);
+
+            let mut aborted_epochs = vec![];
```
wenym1: Can first do `pending_items.extend(metadata_iter.take_while(state is pending))`, and then `let aborted_epochs = metadata_iter.inspect(assert state is aborted).collect()`.

src/meta/src/manager/sink_coordination/coordinator_worker.rs
```diff
@@ -104,6 +112,117 @@ impl<R> AligningRequests<R> {
...
+        let (last_epoch, _) = self.prepared_epochs.pop_front().expect("non-empty");
+        assert_eq!(last_epoch, epoch);
+
+        commit_and_prune_epoch(&self.db, self.sink_id, epoch, self.last_committed_epoch).await?;
+        self.last_committed_epoch = Some(epoch);
+        Ok(())
+    }
+
+    fn failed_committed(&mut self, epoch: u64, err: SinkError) {
+        if let Some((prev_fut, strategy)) = &mut self.backoff_state {
```
wenym1, line 224: May also assert that the epoch at the front equals to `epoch`.

src/meta/src/manager/sink_coordination/coordinator_worker.rs
Chiro11: Meet
```
use of unstable library feature `vec_deque_pop_if`
see issue #135889 <https://github.com/rust-lang/rust/issues/135889> for more information
add `#![feature(vec_deque_pop_if)]` to the crate attributes to enable
this compiler was built on 2025-10-09; consider upgrading it if it is out of date`
```

src/meta/src/manager/sink_coordination/coordinator_worker.rs
Chiro11: Considering that abort is an asynchronous method, the second section still remains the loop pattern.

src/meta/src/manager/sink_coordination/coordinator_worker.rs
wenym1: Will it fail to compile after we add `#![feature(vec_deque_pop_if)]` to lib.rs?

src/meta/src/manager/sink_coordination/coordinator_worker.rs
```diff
@@ -188,9 +191,15 @@ impl TwoPhaseCommitHandler {
 
     fn push_new_item(&mut self, epoch: u64, metadata: Vec<u8>) {
         if epoch > self.curr_hummock_committed_epoch {
+            if !self.pending_epochs.is_empty() {
```
wenym1: can use `if let Some() = self.pending_epochs.back()`

src/meta/src/manager/sink_coordination/coordinator_worker.rs
Chiro11: No, it can be compiled, I have added.

src/meta/model/migration/src/m20251124_195858_pending_sink_state.rs
```diff
@@ -0,0 +1,156 @@
...
+                        Index::create()
+                            .col(PendingSinkState::SinkId)
+                            .col(PendingSinkState::Epoch),
+                    )
+                    .to_owned(),
+            )
+            .await?;
+
+        if manager
+            .has_table(ExactlyOnceIcebergSinkMetadata::Table.to_string())
```
wenym1: @wcy-fdu Can you review the migration logic?

Besides, have we tested the backward compatibility of this migration either locally or in the CI backward test?

wenym1 (approved):
Rest LGTM.

We may also need to monitor the number of pending and prepared epochs for each sink. Can do it this PR or in another PR.

src/meta/model/migration/src/m20251124_195858_pending_sink_state.rs
```diff
@@ -0,0 +1,156 @@
...
+                    .primary_key(
+                        Index::create()
+                            .col(PendingSinkState::SinkId)
+                            .col(PendingSinkState::Epoch),
+                    )
+                    .to_owned(),
+            )
+            .await?;
+
+        if manager
```
yezizp2012: Here should assert that table `ExactlyOnceIcebergSinkMetadata` exists.

yezizp2012: If `ExactlyOnceIcebergSinkMetadata` is deprecated, it's best to remove it and clean up the associated code.

PR #23658, opened 2025-11-04
refactor(streaming): separate streaming config in `actor_context` from `stream_env`

BugenZhao:

Currently, when an executor or an executor builder needs to access the `StreamingConfig`, there are two origins which are the clone of the same `StreamingConfig` instance:

- `StreamEnvironment.config`
- `ActorContext.streaming_config`

...where `StreamEnvironment` is global while `ActorContext` is per-actor. 

As a preparation to support per-actor (or per-job) configuration, in this PR, we

- rename `StreamEnvironment.config` to `global_config`
- rename `ActorContext.streaming_config` to `config`
- let (almost) all executors and executor builders access config via `ActorContext` instead of `StreamEnvironment`

This will allow us to provision different configurations for actors through `ActorContext`, with some entries overridden by the user based on the `global_config`. Currently, they are still the same under the hood.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/task/actor_manager.rs
+++ b/src/stream/src/task/actor_manager.rs
@@ -123,7 +123,6 @@ impl StreamActorManager {
         .await
     }
 
-    #[expect(clippy::too_many_arguments)]
     async fn create_snapshot_backfill_node(
         &self,
         stream_node: &StreamNode,
@@ -489,7 +488,6 @@ impl StreamActorManager {
         };
 
         let monitor_handle = if self.streaming_metrics.level >= MetricLevel::Debug
-            // TODO(config): use config from actor context, instead of the global one
             || self
                 .env
                 .global_config()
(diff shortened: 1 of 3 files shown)


kwannoel:
> As a preparation to support per-actor (or per-job) configuration, in this PR, we

    rename StreamEnvironment.config to global_config
    rename ActorContext.streaming_config to config
    let (almost) all executors and executor builders access config via ActorContext instead of StreamEnvironment


Do you mean _per fragment_ configuration? I suppose we won't support per actor, we have removed the actor table.

BugenZhao:
> Do you mean _per fragment_ configuration?

I suppose what's user-facing will actually be per-job. Per-actor is just implementation wise.

src/stream/src/task/actor_manager.rs
kwannoel, line 564: Shall we store it in a structured way instead? 🤔 

And does this make `ExecutorInfo` obsolete?

BugenZhao: `ExecutorParams` only lives in `from_proto`. It's more like the parameters/arguments to the executor builder and won't be held by the executor, while `ExecutorInfo` will.

Executors will use the `config` from `actor_context` after they are constructed. A clone is placed here solely for syntactic convenience.

PR #23671, opened 2025-11-05
feat: implement refresh progress tracking for refresh table

tabVersion:
…ured source events

- Introduced new message types for `LoadFinishedSource` and `ListFinishedSource` to encapsulate source completion events, improving clarity and structure.
- Updated `BarrierCompleteResponse` to use these new message types, replacing previous `uint32` identifiers.
- Modified related handling in the `RefreshManager`, `GlobalBarrierWorkerContext`, and other components to accommodate the new structured events.
- Enhanced logging and error handling for better traceability of source completion events.I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

to replace https://github.com/risingwavelabs/risingwave/pull/23601

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/executor/source/batch_source/batch_posix_fs_fetch.rs
+++ b/src/stream/src/executor/source/batch_source/batch_posix_fs_fetch.rs
@@ -345,6 +345,7 @@ impl<S: StateStore> BatchPosixFsFetchExecutor<S> {
                                 && file_queue.is_empty()
                                 && list_finished
                                 && is_refreshing
+                                && barrier.is_checkpoint()
                             {
                                 tracing::info!(
                                     ?epoch,
--- a/src/stream/src/executor/source/batch_source/batch_posix_fs_list.rs
+++ b/src/stream/src/executor/source/batch_source/batch_posix_fs_list.rs
@@ -290,7 +290,7 @@ impl<S: StateStore> BatchPosixFsListExecutor<S> {
                             }
 
                             // Report list finished after all files are listed
-                            if is_refreshing {
+                            if is_refreshing && barrier.is_checkpoint() {
                                 self.report_list_finished(barrier.epoch);
                                 is_refreshing = false;
                             }


src/meta/src/stream/refresh_manager.rs
```diff
@@ -12,17 +12,55 @@
...
+/// Global, per-table refresh progress tracker.
+///
+/// Lifecycle for each entry (keyed by `TableId`):
+/// - Created at refresh start in `RefreshManager::refresh_table` before any `await`,
+///   populated with the expected source/fetch actor sets for that table.
+/// - Updated on each barrier in checkpoint control when executors report
+///   list/load progress; see `CheckpointControl::handle_refresh_table_info`.
+/// - Removed when the table is reported as refresh-finished by compute on a
+///   barrier (`refresh_finished_table_ids`).
+///
```
chenzl25, line 40: I think we need to handle the case when the refreshable table be dropped.

chenzl25 (approved):
Rest LGTM!

src/meta/src/stream/refresh_manager.rs
```diff
@@ -12,17 +12,55 @@
...
+/// Failure/retry notes:
+/// - If scheduling the refresh fails, the table state is reset to `Idle`
+pub static REFRESH_TABLE_PROGRESS_TRACKER: LazyLock<Mutex<GlobalRefreshTableProgressTracker>> =
```
BugenZhao, line 43: I think we should avoid introducing such global variable.

tabVersion: True, but we need a way to have the same lifespan as RefreshTable Job, and access each job's status each time a barrier arrives. 
Luckily it does not persist any data, so less burden for future improvements.

PR #23675, opened 2025-11-05
fix(grafana): use percentage for hummock read metrics

kwannoel:

Artifacts generated in https://github.com/risingwavelabs/risingwave/pull/23676

These metrics are using _rate_ and so the unit can't be per second, it should be a percentage. They reflect the percentage of time spent per interval reading records, either via iteration or point gets.

```
sum(rate({table_metric('state_store_get_duration_bucket')
sum(rate({table_metric('state_store_iter_init_duration_bucket')
sum(rate({table_metric('state_store_iter_scan_duration_bucket')
```

Further, these metrics are all at the _second level_ granularity. [Since `rate` is](https://prometheus.io/docs/prometheus/latest/querying/functions/#rate):
> rate(v range-vector) calculates the per-second average rate of increase of the time series in the range vector. 

Therefore we don't need to apply normalization as we would for other metrics tracked as sub-second level (e.g. `ns`, `ms`, etc...).

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/grafana/dashboard/dev/hummock_read.py
+++ b/grafana/dashboard/dev/hummock_read.py
@@ -114,7 +114,7 @@ def _(outer_panels: Panels):
                         ),
                     ],
                 ),
-                panels.timeseries_latency(
+                panels.timeseries_percentage(
                     "Read Duration - Get",
                     "Histogram of the latency of Get operations that have been issued to the state store.",
                     [
@@ -134,7 +134,7 @@ def _(outer_panels: Panels):
                         ),
                     ],
                 ),
-                panels.timeseries_latency(
+                panels.timeseries_percentage(
                     "Read Duration - Iter",
                     "Histogram of the time spent on iterator initialization."
                     "Histogram of the time spent on iterator scanning.",


kwannoel:
* **#23675** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23675?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#23607](https://github.com/risingwavelabs/risingwave/pull/23607) <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23607?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#23676](https://github.com/risingwavelabs/risingwave/pull/23676) <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23676?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23675?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#23612** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23612?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#23730](https://github.com/risingwavelabs/risingwave/pull/23730) <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23730?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

hzxa21:
> These metrics are using rate and so the unit can't be per second, it should be a percentage. They reflect the percentage of time spent per interval reading records, either via iteration or point gets.

Can you explain more on this? I think it returns the tail duration in time unit instead of percentage.

For example, this returns the duration (in seconds) that 99% of your API calls completed within, calculated over the 5-minute window
```
histogram_quantile(0.99, sum(rate(api_call_duration_seconds_bucket[5m])) by (le, path))
```

kwannoel:
> > These metrics are using rate and so the unit can't be per second, it should be a percentage. They reflect the percentage of time spent per interval reading records, either via iteration or point gets.
> 
> Can you explain more on this? I think it returns the tail duration in time unit instead of percentage.
> 
> For example, this returns the duration (in seconds) that 99% of your API calls completed within, calculated over the 5-minute window
> 
> ```
> histogram_quantile(0.99, sum(rate(api_call_duration_seconds_bucket[5m])) by (le, path))
> ```

Nevermind ignore me. I misinterpreted the metric: `api_call_duration_seconds_bucket` to be measuring _seconds_, rather than counts in duration buckets.

PR #23681, opened 2025-11-05
feat(sink): allow buffering and compacting when stream key is same as sink primary key

### LINKED ISSUE #23450: feat: sink buffering and compacting even when the stream key is the same as sink primary key

https://docs.risingwave.com/delivery/overview#sink-buffering-behavior

Some use cases can tolerate the latency of a barrier interval before sinking outputs.
By buffering and compacting:
1. the number of output messages could be reduced (largely)
2. more importantly, some users' downstream systems wish to accept only one update per key per barrier interval. It can make their logic simpler.

Add a new config to enforce buffering and compacting even when the stream key and primary key are the same.
It does not hurt correctness.

_No response_

### Additional context

_No response_

silver-ymz:

close #23450

This PR add an option `force_compaction` to allow buffering and compaction when stream key is same as sink primary key.

## Checklist

## Documentation

<details>

Previously, if the stream key matches the sink primary key exactly, no buffering is performed and the sink emits changes immediately as they arrive. Now, users can set the `force_compaction` option to force buffering and compaction to be enabled.

Example:
```sql
create table t_force_compaction (
    id integer primary key,
    v_int int
);

create sink sink_t_force_compaction from t_force_compaction with (
  connector = 'kafka',
  properties.bootstrap.server = 'message_queue:29092',
  topic = 'test-rw-force-compaction-sink',
  force_compaction = 'true',
  primary_key = 'id')
format upsert encode json;
```

</details>

### DIFF
--- a/ci/scripts/e2e-sink-test.sh
+++ b/ci/scripts/e2e-sink-test.sh
@@ -96,6 +96,7 @@ risedev slt './e2e_test/sink/license.slt'
 risedev slt './e2e_test/sink/rate_limit.slt'
 risedev slt './e2e_test/sink/auto_schema_change.slt'
 risedev slt './e2e_test/sink/sink_vector_columns.slt'
+risedev slt './e2e_test/sink/force_compaction_sink.slt'
 sleep 1
 
 echo "--- preparing postgresql"
--- a/e2e_test/sink/force_compaction_sink.slt
+++ b/e2e_test/sink/force_compaction_sink.slt
@@ -1,0 +1,28 @@
+statement ok
+create table t_force_compaction (
+    id integer primary key,
+    v_int int
+);
+
+statement ok
+create sink sink_t_force_compaction from t_force_compaction with (
+  connector = 'blackhole',
+  type = 'upsert',
+  force_compaction = 'true',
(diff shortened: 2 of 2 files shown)


src/stream/src/executor/sink.rs
```diff
@@ -1184,4 +1196,131 @@ mod test {
...
+            Message::Chunk(StreamChunk::from_pretty(
+                " I I I
+                    + 1 2 20
+                    - 1 2 20",
+            )),
+            Message::Chunk(StreamChunk::from_pretty(
+                " I I I
+                    - 1 1 10
+                    + 1 1 40",
+            )),
```
BugenZhao, line 1260: Shall we demonstrate where the `Delete` row and the `Insert` row of an update reside in different chunks?

kwannoel (dismissed):

kwannoel:
Implementation looks good to me. You should add e2e sink tests for this new option as well. Can be done in `e2e_test/sink/*.slt` with blackhole sink for simplicity.

Could you also update your release note with an example of how this option can be used?

```
create sink s1 from t with (connector='kafka', force_compaction='true', ...);
```

PR #23685, opened 2025-11-06
feat(dashboard): add timeseries latency panel for Materialize Views & Sinks to monitor processing epochs

tabVersion:

the metrics exist in User Dashboard but not Dev Dashboard, but seems useful

https://github.com/risingwavelabs/risingwave/blob/7fea82d39639633487eeef44dd4c8ff76f6ed183/grafana/dashboard/user/streaming.py#L100-L119

## Checklist

## Documentation

<details>

</details>


hzxa21 (approved):

kwannoel:
I wrote some evaluation about this metric here: https://www.notion.so/risingwave-labs/RFC-Identifying-Bottleneck-via-Busy-Rate-28bf0d69cb7680f18274c09cbf3a45b8?source=copy_link#296f0d69cb7680c28c11de72b2ec9d89. We should be careful when using it.

PR #23686, opened 2025-11-06
feat(optimizer): add limit for sys queries if meet max_one_row requirement

chenzl25:

- Lots of PG sys queries running against RisingWave failed because of `Scalar subquery might produce more than one row`, which is annoying and not user friendly. Considering sys table can only be used in Batch queries and most of them are able to run in a postgres instance, I think we can put a limit if the plan containing only a SysScan operator which though is not equivalent in terms of plan transformation, it is much more easier than doing derivation to strictly ensure it is max-one-row.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/optimizer/plan_visitor/sole_sys_table_visitor.rs
+++ b/src/frontend/src/optimizer/plan_visitor/sole_sys_table_visitor.rs
@@ -23,7 +23,7 @@ pub struct SoleSysTableVisitor {
 }
 
 impl SoleSysTableVisitor {
-    pub fn has_sys_table(plan: LogicalPlanRef) -> bool {
+    pub fn sys_table_only(plan: LogicalPlanRef) -> bool {
         let mut visitor = SoleSysTableVisitor {
             has_sys_table: false,
         };
--- a/src/frontend/src/optimizer/rule/max_one_row_eliminate_rule.rs
+++ b/src/frontend/src/optimizer/rule/max_one_row_eliminate_rule.rs
@@ -36,7 +36,7 @@ impl Rule<Logical> for MaxOneRowEliminateRule {
         }
 
         if !right.max_one_row() {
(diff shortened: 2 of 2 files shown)


src/frontend/src/optimizer/plan_visitor/sole_sys_table_visitor.rs
```diff
@@ -0,0 +1,52 @@
...
+use crate::optimizer::plan_node::{LogicalSysScan, LogicalValues};
+use crate::optimizer::plan_visitor::PlanVisitor;
+
+#[derive(Debug, Clone, Default)]
+pub struct SoleSysTableVisitor {
+    has_sys_table: bool,
+}
+
+impl SoleSysTableVisitor {
+    pub fn has_sys_table(plan: LogicalPlanRef) -> bool {
```
BugenZhao: Shall we rename the method name to represent "sys table only"?

src/frontend/planner_test/tests/testdata/input/subquery.yaml
```diff
@@ -312,3 +312,38 @@
     SELECT anon_1.f1 AS "IntegerRange(0, -2, 1)" FROM unnest(CASE WHEN (nullif(1, 0) IS NOT NULL AND sign(1) = sign( -2 - 0)) THEN array_remove(array((SELECT generate_series(0,  -2, 1) AS generate_series_1)),  -2) ELSE CAST(ARRAY[] AS SMALLINT[]) END) AS anon_1(f1);
   expected_outputs:
     - binder_error
+- name: sys subquery
+  sql: |
+    SELECT
```
BugenZhao, line 317: May I know what this query is for?

chenzl25: This query is generated by an external tool, probably used to determine relationships between source keys and target keys of foreign constraints.

PR #23687, opened 2025-11-06
fix(optimizer): fix properties of plan nodes for approx percentile

BugenZhao:

The properties of the plan nodes for approx percentile are largely incorrect. It's because the plan is simple and fixed, we don't encounter any issues. However, we should still fix them for robustness.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/optimizer/plan_node/stream_row_merge.rs
+++ b/src/frontend/src/optimizer/plan_node/stream_row_merge.rs
@@ -60,8 +60,8 @@ impl StreamRowMerge {
         assert_eq!(lhs_input.stream_key(), rhs_input.stream_key());
         assert_eq!(lhs_input.stream_kind(), rhs_input.stream_kind());
 
-        // Currently, `RowMerge` only supports and is only used for merging simple agg and approx percentile agg.
-        // We restrict the input's distribution, key, and kind here for simplicity.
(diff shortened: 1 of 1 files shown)


BugenZhao:
* **#23583** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23583?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#23687** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23687?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23687?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

kwannoel:
To provide historical context for this PR, the following properties were incorrect before it:
- Stream key: Should be empty, rather than following upstream.
- Distribution: Should be Single , rather than following upstream.
- Watermark: Should be empty, rather than following upstream.

It's because the initial design wasn't just for merging Distribution::Single upstreams. It was also intended to merge approx percentile + group by. But we realized the design is not suitable as we can only output on barrier to align these upstreams, so it would lead to performance issues.
I didn't change it after that design switch. Thanks for making the necessary changes 🙏

Issue #23680, opened 2025-11-05
bug: cdc heartbeat event's offset field is not used.


zwang28:
### Describe the bug

#12868 introduce CDC heartbeat message.

However, since a heartbeat message by design will be put in a dedicated stream chunk, it could never enter [this line](https://github.com/risingwavelabs/risingwave/blob/664f3e9104656d8c699f6fb3fd19805e56fa0cf5/src/stream/src/executor/source/reader_stream.rs#L210) because `msg.rows().count()` always equal to 0. Consequently, a heartbeat message's offset will never be used by source executor.

### Error message/log

```text

```

hzxa21:
I saw that we will assign offset to any message including the heartbeat message:
https://github.com/risingwavelabs/risingwave/blob/c9c9ed83656ed4e56759d9f5a0b5cc481e060f7f/java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/core/DbzChangeEventConsumer.java#L164


Does it mean that offset in the heartbeat message is not empty but since hb message doesn't have [payload](https://github.com/risingwavelabs/risingwave/blob/c9c9ed83656ed4e56759d9f5a0b5cc481e060f7f/java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/core/DbzChangeEventConsumer.java#L268), it will never update the split offset? In other words, the hearbeat message is currently useless in main.

hzxa21:
I think it is a bug introduce by #20526.

Prior to this PR, it seems that we will update the split offset when meeting any chunk:

<img width="918" height="1050" alt="Image" src="https://github.com/user-attachments/assets/c14d327c-3cbc-437b-add3-0110bb87eba9" />

PR #15678, opened 2024-03-14
fix(source): parse message with empty key and payload

StrikeW:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://gist.github.com/TennyZhuang/f00be7f16996ea48effb049aa7be4d66#file-rw_cla).

## What's changed and what's your intention?
CDC connector could emit heartbeat message to the Source. We rely on the heartbeat to keep up with upstream change log offset, which is important for dedicated source (e.g. mongodb) to reset to a valid offset upon recovery.
Without heartbeat messages, the source offset may become invalid in our side.

This PR fix the issue introduced by [#14524](https://github.com/risingwavelabs/risingwave/pull/14524/files#diff-d20a701c6d209a35e84545b11ccd13ab2f332a68aa175eb4bf73edf67184e8bcL644-L648). We extract the heartbeat message from the source batch message and emit the heartbeat chunk first to executor. The overhead should be negligible.

## Checklist

- [x] I have written necessary rustdoc comments
- [ ] I have added necessary unit tests and integration tests
- [ ] I have added test labels as necessary. See [details](https://github.com/risingwavelabs/risingwave/blob/main/docs/developer-guide.md#ci-labels-guide).
- [ ] I have added fuzzing tests or opened an issue to track them. (Optional, recommended for new SQL features #7934).
- [ ] My PR contains breaking changes. (If it deprecates some features, please create a tracking issue to remove them in the future).
- [ ] All checks passed in `./risedev check` (or alias, `./risedev c`)
- [ ] My PR changes performance-critical code. (Please run macro/micro-benchmarks and show the results.)

- [ ] My PR contains critical fixes that are necessary to be merged into the latest release. (Please check out the [details](https://github.com/risingwavelabs/risingwave/blob/main/CONTRIBUTING.md))

## Documentation

- [ ] My PR needs documentation updates. (Please use the **Release note** section below to summarize the impact on users)

## Release note

If this PR includes changes that directly affect users or other significant modifications relevant to the community, kindly draft a release note to provide a concise summary of these changes. Please prioritize highlighting the impact these changes will have on users.

</details>

PR #23689, opened 2025-11-06
fix(connector): fix cdc heartbeat message issues

zwang28:

This PR fixes 2 cdc heartbeat message issues:
1. By design [All rows (including those visible or invisible) will be used to update the offset](https://github.com/risingwavelabs/risingwave/pull/15678#discussion_r1537102881). However currently only [visible rows](https://github.com/risingwavelabs/risingwave/blob/a0562979f135660e9674197dba22d9917c75fc7c/src/stream/src/executor/source/reader_stream.rs#L208) are used to update the offset. It leads to [this bug](https://github.com/risingwavelabs/risingwave/issues/23680#issuecomment-3490519134).
2. Currenly a message batch will emit heartbeat message only when [all messages are heartbeat messages in this batch](https://github.com/risingwavelabs/risingwave/blob/a0562979f135660e9674197dba22d9917c75fc7c/src/connector/src/parser/mod.rs#L255). However, this design might result in neither heartbeat or any other messages are emitted for a batch. e.g. a batch of [heartbeat, transaction begin, transcation end] will emit no messages to downstream. This is almost 100% reproducible for an idle MySQL CDC, which results in consistently no heartbeat message being emitted at all. Because [heartbeat messages are skipped](https://github.com/risingwavelabs/risingwave/blob/a0562979f135660e9674197dba22d9917c75fc7c/src/connector/src/parser/mod.rs#L288) and transaction control messages never emit.

## Checklist

## Documentation

<details>

</details>


zwang28:
@tabVersion @BugenZhao Could you please review this PR at your convenience?

src/connector/src/parser/mod.rs
```diff
@@ -284,10 +284,21 @@ async fn parse_message_stream<P: ByteStreamSourceParser>(
 
         let mut txn_started_in_last_batch = chunk_builder.is_in_transaction();
         let process_time_ms = chrono::Utc::now().timestamp_millis();
-
+        let mut is_heartbeat_emitted = false;
```
BugenZhao, line 259: Is the short circuit of `batch.iter().all(|msg| msg.is_cdc_heartbeat())` still necessary now? 

> let's just emit the latest heartbeat message

How much will it cost if we don't have such optimization? Will we frequently encounter a large number of heartbeats (only) in a chunk?

zwang28: We can remove this short circuit. Because by default the heartbeat interval is either [1 minute](https://github.com/risingwavelabs/risingwave/blob/887cc81f18361050f2f77a906c6ee600a83aaaa1/java/connector-node/risingwave-connector-service/src/main/resources/mysql.properties#L26) or [5 minutes](https://github.com/risingwavelabs/risingwave/blob/887cc81f18361050f2f77a906c6ee600a83aaaa1/java/connector-node/risingwave-connector-service/src/main/resources/postgres.properties#L21). Although users can override this by setting `debezium.heartbeat.interval.ms`, this is uncommon.

wcy-fdu:
Sorry for the delay of a few days, I will review this PR today.

src/connector/src/parser/mod.rs
```diff
@@ -284,10 +284,21 @@ async fn parse_message_stream<P: ByteStreamSourceParser>(
...
+                if !is_heartbeat_emitted {
+                    tracing::debug!(offset = msg.offset, "handling a heartbeat message");
+                    chunk_builder.heartbeat(MessageMeta {
+                        source_meta: &msg.meta,
+                        split_id: &msg.split_id,
+                        offset: &msg.offset,
+                    });
+                    for chunk in chunk_builder.consume_ready_chunks() {
+                        yield chunk;
+                    }
```
wcy-fdu, line 271: The optimal solution here would be to process the last heartbeat in this batch, not the first. In some corner cases, such as when there are many heartbeat messages in the batch, the recorded offset might not be the latest. However, in practice, the impact is minimal, and I think the current solution is acceptable. 
correct me if I was wrong.

src/connector/src/parser/mod.rs
zwang28: > Is it possible to have some tests to avoid regression in the future?

I'm working on a regression test. It may be included in another PR if I cannot finish it before this release code freeze.

PR #23692, opened 2025-11-06
feat(mysql-cdc): monitor the progress of binlog offsets in the upstream and downstream processes

wcy-fdu:
fI hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

Unlike Postgres, where the offset is the LSN in the WAL and is a comparable u64, MySQL's offset consists of the binlog file and the position (in bytes) within that particular binlog file. Therefore, I used two metrics to jointly record the MySQL offset: file seq, which indicates the current offset in which binlog file, and binlog position, which represents the reading position within that file. During normal operation, file seq will gradually increase, while binlog position will loop back to the beginning and increase again after switching files.

In addition, periodically querying the maximum and minimum file seq of the upstream MySQL binlog can help determine how far the current offset is from the binlog expiration and the lag from the latest binlog. However, this information is not currently displayed on Grafana (i.e., `max_seq - current_seq `and `current_seq - min_seq`). If you think it's necessary, please let me know, and I can add it.

image

## Checklist

## Documentation

<details>

</details>


grafana/dashboard/dev/streaming_cdc.py
```diff
@@ -89,24 +89,46 @@ def _(outer_panels: Panels):
...
+                            "source_id {{source_id}} - State Table Binlog Position (bytes)",
+                        ),
+                        panels.target(
+                            f"{metric('mysql_cdc_binlog_file_seq_min')}",
+                            "{{hostname}}:{{port}} - Upstream Binlog File Min Seq (oldest)",
+                        ),
+                        panels.target(
+                            f"{metric('mysql_cdc_binlog_file_seq_max')}",
+                            "{{hostname}}:{{port}} - Upstream Binlog File Max Seq (newest)",
+                        ),
```
wcy-fdu, line 129: This two seq help determine how far the current offset is from the binlog expiration and the lag from the latest binlog. However, this information is not currently displayed on Grafana, i.e., 
- lag_from_latest: `max_seq - current_seq `
- expiration_countdown: `current_seq - min_seq`
 If you think it's necessary, please let me know, and I can add it.

src/stream/src/executor/source/source_executor.rs
```diff
@@ -435,16 +463,32 @@ impl<S: StateStore> SourceExecutor<S> {
         if !cache.is_empty() {
             tracing::debug!(state = ?cache, "take snapshot");
 
-            // Record LSN metrics for PostgreSQL CDC sources before moving cache
+            // Record metrics for CDC sources before moving cache
             let source_id = core.source_id.to_string();
             for split_impl in &cache {
-                // Extract offset for CDC using type-safe matching
+                // PostgreSQL CDC: Extract and record LSN
                 if let Some(state_table_lsn_value) = extract_split_offset(split_impl) {
```
hzxa21: Related to #22968: the naming `extract_split_offset` here is confusing. It is specific to PG CDC bug the naming suggests that it is for connector. What we do in this PR by checking the SplitImpl enum first is clearer to me, let's change the PG part as well
```
match split {
        SplitImpl::PostgresCdc(pg_split) => {
            if let Some(pg_lsn) = pg_split.pg_lsn() {
              ...
            }
            
        },
        SplitImpl::MysqlCdc(mysql_split) => {
          if let Some((file_seq, position)) = mysql_split.mysql_binlog_offset() {
             ...
          }
        }
                    
        _ => None,
    }
```

src/stream/src/executor/source/source_executor.rs
```diff
@@ -76,6 +76,34 @@ fn extract_pg_cdc_lsn_from_offset(offset_str: &str) -> Option<u64> {
...
+///     "file": "binlog.000123",
+///     "pos": 456789,
+///     ...
+///   }
+/// }
+///
+/// Returns `Some((file_seq, position))` where:
+/// - `file_seq`: the numeric part of binlog filename (e.g., 123 from "binlog.000123")
+/// - `position`: the byte offset within the binlog file
+fn extract_mysql_cdc_binlog_offset(offset_str: &str) -> Option<(u64, u64)> {
```
hzxa21: `source_executor.rs` is already huge. Let's try to avoid putting connector specific logics in this file as much as possible.

Let's put these methods in https://github.com/risingwavelabs/risingwave/blob/d052368d0f54631b5649dce648b70ac323634954/src/connector/src/source/cdc/split.rs#L256

Same for `extract_pg_cdc_lsn_from_offset`

Example:
```
impl DebeziumCdcSplit<Postgres> {
  pub fn pg_lsn(&self) -> Option<u64> {
     ...
  }
}

impl DebeziumCdcSplit<Mysql> {
  pub fn mysql_binlog_offset(&self) -> Option<(u64, u64)> {
     ...
  }
}
```

grafana/dashboard/dev/streaming_cdc.py
```diff
@@ -89,24 +89,46 @@ def _(outer_panels: Panels):
...
+                            f"{metric('stream_mysql_cdc_state_binlog_position')}",
+                            "source_id {{source_id}} - State Table Binlog Position (bytes)",
+                        ),
+                        panels.target(
+                            f"{metric('mysql_cdc_binlog_file_seq_min')}",
+                            "{{hostname}}:{{port}} - Upstream Binlog File Min Seq (oldest)",
+                        ),
+                        panels.target(
+                            f"{metric('mysql_cdc_binlog_file_seq_max')}",
+                            "{{hostname}}:{{port}} - Upstream Binlog File Max Seq (newest)",
```
hzxa21, line 128: Why do we use `hostname` and `port` rather than `source_id` in the label?

src/connector/src/source/cdc/enumerator/mod.rs
```diff
@@ -271,6 +272,140 @@ impl<T: CdcSourceTypeTrait> CdcMonitor for DebeziumSplitEnumerator<T> {
...
 }
 
+impl DebeziumSplitEnumerator<Mysql> {
+    async fn monitor_mysql_binlog_files(&mut self) -> ConnectorResult<()> {
+        // Get hostname and port for metrics labels
+        let hostname = self
+            .properties
+            .get("hostname")
+            .map(|s| s.as_str())
+            .unwrap_or("unknown");
```
hzxa21: should it be an error rather than `unwrap_or("unknown")`?

src/connector/src/source/cdc/enumerator/mod.rs
```diff
@@ -271,6 +272,140 @@ impl<T: CdcSourceTypeTrait> CdcMonitor for DebeziumSplitEnumerator<T> {
...
+        let port = self
+            .properties
+            .get("port")
+            .map(|s| s.as_str())
+            .unwrap_or("unknown");
```
hzxa21: ditto

src/connector/src/source/cdc/enumerator/mod.rs
```diff
@@ -271,6 +272,140 @@ impl<T: CdcSourceTypeTrait> CdcMonitor for DebeziumSplitEnumerator<T> {
...
+            .get("password")
+            .ok_or_else(|| anyhow::anyhow!("password not found in CDC properties"))?;
+        let database = self
+            .properties
+            .get("database.name")
+            .ok_or_else(|| anyhow::anyhow!("database.name not found in CDC properties"))?;
+
+        // Build MySQL connection string
+        let mysql_url = format!(
+            "mysql://{}:{}@{}:{}/{}",
```
hzxa21: How about the SSL related configs? https://github.com/risingwavelabs/risingwave/blob/d052368d0f54631b5649dce648b70ac323634954/src/connector/src/source/cdc/external/mysql.rs#L448

Can we extract the common codes to set up mysql connection based on the with properties? If I search for `mysql_async::Pool::new` in our code bases, there are many occurrences and I think it is easy to miss some configs if we don't unify them.

hzxa21: Btw, I found that we hard code ssl mode for PG rather than honoring the user provided option when querying the LSN. Will this be an issue?
https://github.com/risingwavelabs/risingwave/blob/d052368d0f54631b5649dce648b70ac323634954/src/connector/src/source/cdc/enumerator/mod.rs#L211

grafana/dashboard/dev/streaming_cdc.py
wcy-fdu: Even with multiple sources, the maximum and minimum binlog values ​​within the same upstream MySQL instance are fixed. Therefore, I believe the upstream monitoring is per (IP + host) rather than per source.

src/connector/src/source/cdc/enumerator/mod.rs
wcy-fdu: There won't be any correctness issues using the `preffered` mode because it works regardless of whether SSL is configured. It can also be used with the customer-configured settings.

wcy-fdu: > in our code bases, there are many occurrences and I think it is easy to miss some configs if we don't unify them.

There are many places in our code bases using this connection pool in many places:
- mysql cdc backfill
- mysql tvf
- new StarRocks schema client
- this pr, monitoring upstream mysql binlog

I plan to extract a common method like `build_mysql_connection_pool`  first to unify the CDC backfill and monitoring, and then unify the rest in later separate pr. This will prevent the changes in this pr from being too broad.

PR #23693, opened 2025-11-06
fix: suppress "telemetry event reporting disabled" log lines with rate limit

BugenZhao:

It seems too verbose in main-cron runs.

image

## Checklist

## Documentation

<details>

</details>


src/common/telemetry_event/src/lib.rs
```diff
@@ -77,7 +78,15 @@ pub fn report_event_common(
     if let Some(tracking_id) = TELEMETRY_TRACKING_ID.get() {
         event_tracking_id = tracking_id.to_string();
     } else {
-        tracing::info!("Telemetry tracking_id is not set, event reporting disabled");
+        static LOG_SUPPERSSER: LazyLock<LogSuppresser> =
```
kwannoel: Think you made a typo here.

BugenZhao: You are right. The entire codebase is wrong. 😂

image

PR #23697, opened 2025-11-06
fix(iceberg): enfore iceberg sink type to retract when specified as upsert

chenzl25:

- https://github.com/risingwavelabs/risingwave/pull/23593 introduces an retract sink type, but for iceberg sink, its sink type even specified as upsert should be enforced to retract, because iceberg sink relies on the retractable input to write delete files.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/from_proto/sink.rs
+++ b/src/stream/src/from_proto/sink.rs
@@ -219,6 +219,7 @@ impl ExecutorBuilder for SinkExecutorBuilder {
             SinkType::Retract
         } else {
             let sink_type_from_proto = SinkType::from_proto(sink_desc.get_sink_type().unwrap());
+            // For backward compatibility: Iceberg sink with Upsert type should be treated as Retract type.
             if connector.eq_ignore_ascii_case(ICEBERG_SINK)
                 && matches!(sink_type_from_proto, SinkType::Upsert)
             {


hzxa21 (approved):

chenzl25:
cc @BugenZhao  PTAL

src/frontend/src/optimizer/plan_node/stream_sink.rs
```diff
@@ -591,7 +591,14 @@ impl StreamSink {
         if let Some(sink_type) = properties.get(SINK_TYPE_OPTION) {
             let sink_type = match sink_type.as_str() {
                 SINK_TYPE_APPEND_ONLY => SinkType::AppendOnly,
-                SINK_TYPE_UPSERT => SinkType::Upsert,
+                SINK_TYPE_UPSERT => {
```
BugenZhao, line 594: Does it mean that we want to keep current syntax of using `sink_type = 'upsert'` to represent non-append-only, instead of using `retract` which seems more correct in semantics?

Li0k: +1

chenzl25: Using `retract` is more accurate, but `upsert` has been used in our docs and users, if we change this behavior suddenly, I think it will confuse our users. So I think we need to update our doc about iceberg sink to recommend users to use  `retract` instead of `upsert`, but keep `upsert` semantics as `retract` for iceberg (for a time).

BugenZhao: Also `IcebergConfig` has its own deserializing logic of deciding the `sink_type`:

https://github.com/risingwavelabs/risingwave/blob/7f419753b0df412aa6ca7e561d29e2b8ca2edbfd/src/connector/src/sink/iceberg/mod.rs#L131-L137

This doesn't seem good, because all other sinks will directly read from `SinkParam`. However, this is already being used in multiple places in our system. So changing the literal here may lead to a much larger refactoring.

chenzl25: Even we have some check here https://github.com/risingwavelabs/risingwave/blob/e9df190590bd4b303684e016b37a37ea71146344/src/connector/src/sink/iceberg/mod.rs#L265-L272 I think we need another PR and e2e test to cover the `retract` type for iceberg sink and iceberg table engine.

PR #23699, opened 2025-11-06
feat: strong typed id in generated pb struct

wenym1:

Previously, though we introduced a strong typed id, we still have to handle the logic to convert with the strong typed id and the primitive u32 in the generated pb struct, which involves a lot of `.into()` and `.as_raw_id()`.

In this PR, we will make the generated pb struct to use the strong typed id directly, so that the conversion logic can be avoided, and we can unify the id type in pb struct and our own code.

Implementation magic:
1. we fork the `prost` repo and add a new method `field_wrapper` to prost build config in crate `prost-build` to specify the fields to use the wrapper type. For fields declared to use wrapper type, in the generated struct, we will change to use the declared wrapper type as the field type, and meanwhile we will add an attribute `prost(wrapper = "crate::id::XxxId")` to the field. Currently we only support plain, required, optional and repeated type. `oneof` is not supported.
The Generated code will be like
```
#[derive(prost_helpers::AnyPB)]
#[derive(Clone, PartialEq, ::prost::Message)]
pub struct Table {
    #[prost(uint32, tag = "1", wrapper = "crate::id::XxxId")]
    pub id: u32,
```
2. The `::prost::Message` derive macro will also have to handle the newly introduced `wrapper` attribute. Therefore, we have to add some logic in crate `prost-derive`, to handle the conversion when `wrapper` is present. 

The patched code can be found in https://github.com/risingwavelabs/prost/pull/1.

For easier usage, we can declare the wrapped fields inside the following macro `for_all_wrapped_id_fields`.
```

for_all_wrapped_id_fields! {
    meta {
        TableFragments.Fragment {
            state_table_ids: TableId,
            table_id: JobId,
        }
        RefreshRequest {
            table_id: TableId,
        }
        ListTableFragmentsRequest {
            table_ids: JobId,
        }
        ListStreamingJobStatesResponse.StreamingJobState {
            table_id: JobId,
            database_id: DatabaseId,
            schema_id: SchemaId,
        }
        FragmentDistribution {
            table_id: JobId,
            state_table_ids: TableId,
        }
...
```

## Checklist

## Documentation

<details>

</details>


wenym1:
* **#23699** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23699?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23699?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

BugenZhao:
> 3\. Therefore, we have to fork the original `prost` repo, and added some logic in crate `prost-derive`, to handle the conversion when `wrapper` is present.

I'm not sure if this is a good idea, especially since `prost` hasn't been on a stable version right now. Also, if we eventually decide to fork the crate, can we get rid of another proc macro `#[prost_helpers::primitive_wrapper]`?

Supporting custom types in `prost` is indeed an long-standing feature request: https://github.com/tokio-rs/prost/issues/764

wenym1:
> > 3. Therefore, we have to fork the original `prost` repo, and added some logic in crate `prost-derive`, to handle the conversion when `wrapper` is present.
> 
> I'm not sure if this is a good idea, especially since `prost` hasn't been on a stable version right now. Also, if we eventually decide to fork the crate, can we get rid of another proc macro `#[prost_helpers::primitive_wrapper]`?
> 
> Supporting custom types in `prost` is indeed an long-standing feature request: [tokio-rs/prost#764](https://github.com/tokio-rs/prost/issues/764)

Ideally we shouldn't fork the repo, and we depend on our own macro to hack into the struct definition. I had a try at the beginning, and the `primitive_wrapper` can already successfully rewrite the field type to the desire type. The only incompatibility is caused by the `::prost::Message` derive macro, which generates code that assumes the field type must be the original u32. As a result we have to fork the original repo, to hack into the `::prost::Message` derive macro. 

As mentioned, `prost` has not been stable and there can be potential breaking change in the future, and I think we should try our best to minimize the code changes on the forked repo, so that in the future when we want to upgrade to later version, we can more easily migrate our patch to the later version. The current code change in the forked repo https://github.com/risingwavelabs/prost/pull/1 is limited and trivial. The reason to introduce the  `#[prost_helpers::primitive_wrapper]` macro in our own repo is to avoid too many code changes in the forked repo,  If we don't use a separate macro and instead directly support a more complete feature in prost builder, there will be a lot of code changes.

wenym1:
> > > 3. Therefore, we have to fork the original `prost` repo, and added some logic in crate `prost-derive`, to handle the conversion when `wrapper` is present.
> > 
> > 
> > I'm not sure if this is a good idea, especially since `prost` hasn't been on a stable version right now. Also, if we eventually decide to fork the crate, can we get rid of another proc macro `#[prost_helpers::primitive_wrapper]`?
> > Supporting custom types in `prost` is indeed an long-standing feature request: [tokio-rs/prost#764](https://github.com/tokio-rs/prost/issues/764)
> 
> Ideally we shouldn't fork the repo, and we depend on our own macro to hack into the struct definition. I had a try at the beginning, and the `primitive_wrapper` can already successfully rewrite the field type to the desire type. The only incompatibility is caused by the `::prost::Message` derive macro, which generates code that assumes the field type must be the original u32. As a result we have to fork the original repo, to hack into the `::prost::Message` derive macro.
> 
> As mentioned, `prost` has not been stable and there can be potential breaking change in the future, and I think we should try our best to minimize the code changes on the forked repo, so that in the future when we want to upgrade to later version, we can more easily migrate our patch to the later version. The current code change in the forked repo [risingwavelabs/prost#1](https://github.com/risingwavelabs/prost/pull/1) is limited and trivial. The reason to introduce the `#[prost_helpers::primitive_wrapper]` macro in our own repo is to avoid too many code changes in the forked repo, If we don't use a separate macro and instead directly support a more complete feature in prost builder, there will be a lot of code changes.

Just had a try on moving logic of `#[prost_helpers::primitive_wrapper]` to the forked repo, and it doesn't increase too many code changes, and the logic there is also trivial. Thanks for bringing this up! PTAL in the latest code @BugenZhao

BugenZhao:
> Just had a try on moving logic of `#[prost_helpers::primitive_wrapper]` to the forked repo, and it doesn't increase too many code changes, and the logic there is also trivial.

Sounds good!

src/prost/src/id.rs
```diff
@@ -13,108 +13,142 @@
...
+            }
+        }))
     }
-}
 
-impl<const N: usize> From<u32> for TypedId<N> {
-    fn from(id: u32) -> Self {
-        Self::new(id)
+    pub fn raw_slice(slice: &[Self]) -> &[u32] {
+        // #safety: transparent repr
```
BugenZhao:
```suggestion
        // SAFETY: transparent repr
```

yezizp2012 (approved):
LGTM!

xiangjinwu:
Nightly image build failure:
https://buildkite.com/risingwavelabs/docker/builds/17866#019a6357-9006-486a-bc84-fe48209b929c

wenym1:
> Nightly image build failure: https://buildkite.com/risingwavelabs/docker/builds/17866#019a6357-9006-486a-bc84-fe48209b929c

Fixed in https://github.com/risingwavelabs/risingwave/pull/23725.

PR #23700, opened 2025-11-06
fix(meta): Fix cancel creating job not return error

Chiro11:

As title.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/meta/src/stream/stream_manager.rs
+++ b/src/meta/src/stream/stream_manager.rs
@@ -161,11 +161,10 @@ impl CreatingStreamingJobInfo {
                         Ok(()) => {
                             receivers.insert(job_id, rx);
                         }
-                        Err(e) => {
+                        Err(_) => {
                             return Err(anyhow::anyhow!(
-                                "failed to send shutdown signal for streaming job {}: {:?}",
-                                job_id,
-                                e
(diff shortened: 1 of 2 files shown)


src/meta/src/stream/stream_manager.rs
```diff
@@ -149,26 +149,34 @@ impl CreatingStreamingJobInfo {
...
-                        receivers.insert(job_id, rx);
-                    } else {
-                        tracing::warn!(id=?job_id, "failed to send canceling state");
+                    match shutdown_tx.send(tx) {
+                        Ok(()) => {
+                            receivers.insert(job_id, rx);
+                        }
+                        Err(_) => {
+                            return Err(anyhow::anyhow!(
+                                "failed to send shutdown signal for streaming job {}: receiver dropped",
```
kwannoel, line 166: Why would the receiver for the streaming job be dropped? Is it due to concurrent cancellations?

Chiro11: Checking the status and canceling are not atomic; it is possible that the job has already finished backfilling when the signal is received.

kwannoel (approved):

yezizp2012 (approved):

PR #23704, opened 2025-11-06
chore(.gitignore): reorganize sections

AndreaBozzo:
## Summary

Reorganized `.gitignore` with clear section headers.

## Changes

- Reorganized file into logical sections with clear comments for better maintainability
- Consolidated scattered patterns (logs, compiler crashes) into single sections
- Moved build artifacts to appropriate sections

The reorganization makes it easier to find and maintain patterns while preserving all existing functionality.


.gitignore
```diff
@@ -1,100 +1,94 @@
...
 !src/prost/src/lib.rs
 !src/prost/src/imp.rs
 **/file_descriptor_set.bin
-# runtime log
-src/log/
-# srcy-tags
-**/srcy-tags.vi
 
-# rustc ice
+## Rust - Build Outputs & Binaries
```
wenym1, line 43: I don't think these are general items to be ignored. All build output are in the `target` folder by default. 

From the PR description, the motivation to ignore these items is that the pre-built risingwave binaries are downloaded to the root folder of source repo, which I think is not a common practice. We expect that users either download the pre-built binaries to somewhere clean to run it directly, or download the source and build the binaries from source.

AndreaBozzo:
Thanks for the review !

.gitignore
AndreaBozzo: Removed the added patterns

wenym1:
Please resolve the conflict.

AndreaBozzo:
Solved the merge prob and checks went by

AndreaBozzo:
Thanks, very happy to contribute to this beautiful project!

Issue #23595, opened 2025-10-28
[Bug] Panic "missing U- before U+" in HashDataDispatcher when Update operations span chunk boundaries


jw-itq:
### Describe the bug

RisingWave compute node panics with error missing U- before U+ in dispatch.rs:1011 when processing materialized views with complex JOINs that read from internal tables with primary keys.

Data Flow Example:

```
-- ============================================================================
-- Layer 0: Iceberg Sources
-- ============================================================================

-- Source 1: Product data from Iceberg
CREATE SOURCE iceberg_product_source WITH (
    connector = 'iceberg',
    s3.endpoint = 'https://s3.amazonaws.com',
    s3.region = 'us-west-2',
    database.name = 'my_database',
    table.name = 'product_catalog'
);

-- Source 2: Service data from Iceberg
CREATE SOURCE iceberg_service_source WITH (
    connector = 'iceberg',
    s3.endpoint = 'https://s3.amazonaws.com',
    s3.region = 'us-west-2',
    database.name = 'my_database',
    table.name = 'service_catalog'
);

-- Source 3: Category metadata from Iceberg
CREATE SOURCE iceberg_category_source WITH (
    connector = 'iceberg',
    s3.endpoint = 'https://s3.amazonaws.com',
    s3.region = 'us-west-2',
    database.name = 'my_database',
    table.name = 'categories'
);

-- Source 4: Supplier metadata from Iceberg
CREATE SOURCE iceberg_supplier_source WITH (
    connector = 'iceberg',
    s3.endpoint = 'https://s3.amazonaws.com',
    s3.region = 'us-west-2',
    database.name = 'my_database',
    table.name = 'suppliers'
);

-- ============================================================================
-- Layer 1: Internal Tables with PRIMARY KEY (for deduplication)
-- ============================================================================

-- Table 1: Products
CREATE TABLE products (
    product_id VARCHAR,
    name VARCHAR,
    category VARCHAR,
    price INTEGER,
    supplier_id VARCHAR,
    updated_at TIMESTAMP,
    PRIMARY KEY (product_id)  -- ← Converts duplicate INSERTs to UPDATEs
);

CREATE SINK sink_iceberg_to_products
FROM iceberg_product_source
INTO products;

-- Table 2: Services
CREATE TABLE services (
    service_id VARCHAR,
    service_name VARCHAR,
    category VARCHAR,
    price INTEGER,
    supplier_id VARCHAR,
    updated_at TIMESTAMP,
[...]

Issue #23708, opened 2025-11-07
meta node log anomaly


jw-itq:
### Describe the bug

meta node Log anomaly，But the service and data are normal

### Error message/log

```text
ERROR risingwave_meta_service::ddl_service: failed to migrate legacy table fragments: failed to get table replace plan from frontend: status: Internal, message: "Catalog error: table not found: internal_test", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "risingwave-error-bin": "NAAAAAAAAABDYXRhbG9nIGVycm9yOiB0YWJsZSBub3QgZm91bmQ6IGludGVybmFsX2NiX3N5c191c2VyASUAAAAAAAAAdGFibGUgbm90IGZvdW5kOiBpbnRlcm5hbF9jYl9zeXNfdXNlcgABBwAAAAAAAABSd0Vycm9yAAGsAAAA", "risingwave-grpc-call": "/frontend_service.FrontendService/GetTableReplacePlan", "date": "Tue, 04 Nov 2025 11:58:47 GMT", "content-length": "0"} }, attempt 277, retrying in 5 secs
2025-11-04T11:58:52.502242972Z ERROR risingwave_meta_service::ddl_service: failed to migrate legacy table fragments: failed to get table replace plan from frontend: status: Internal, message: "Catalog error: table not found: internal_test", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "risingwave-error-bin": "NAAAAAAAAABDYXRhbG9nIGVycm9yOiB0YWJsZSBub3QgZm91bmQ6IGludGVybmFsX2NiX3N5c191c2VyASUAAAAAAAAAdGFibGUgbm90IGZvdW5kOiBpbnRlcm5hbF9jYl9zeXNfdXNlcgABBwAAAAAAAABSd0Vycm9yAAGsAAAA", "risingwave-grpc-call": "/frontend_service.FrontendService/GetTableReplacePlan", "date": "Tue, 04 Nov 2025 11:58:52 GMT", "content-length": "0"} }, attempt 278, retrying in 5 secs
```

### The version of RisingWave

2.6.0

fahadullahshah261:
Hi Wanmingshi,

I am Fahad from RisingWave, and I am editing our monthly newsletter. We have recently introduced the Community Corner section that features contributors and community members. Not much space, but goes out to 8k subscribers. 
[...]

PR #23709, opened 2025-11-07
Fix the meta node log issue and panic issue

jw-itq:

The changes in this PR have resolved issues https://github.com/risingwavelabs/risingwave/issues/23595 and https://github.com/risingwavelabs/risingwave/issues/23708 locally. Since issue https://github.com/risingwavelabs/risingwave/issues/23595 has a significant impact, I just hope you can take a look and see if the changes are reasonable. Thanks!

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/meta/service/src/ddl_service.rs
+++ b/src/meta/service/src/ddl_service.rs
@@ -16,7 +16,7 @@ use std::collections::{HashMap, HashSet};
 use std::pin::pin;
 use std::sync::Arc;
 
-use anyhow::anyhow;
+use anyhow::{Context, anyhow};
 use futures::future::select;
 use rand::rng as thread_rng;
 use rand::seq::IndexedRandom;
@@ -151,14 +151,6 @@ impl DdlServiceImpl {
                     return Ok(());
                 }
 
-                let schema_map: HashMap<u32, String> = metadata_manager
-                    .catalog_controller
-                    .list_schemas()
-                    .await?
-                    .into_iter()
-                    .map(|s| (s.id, s.name))
-                    .collect();
-
                 let client = {
                     let workers = metadata_manager
                         .list_worker_node(Some(WorkerType::Frontend), Some(State::Running))
@@ -172,37 +164,16 @@ impl DdlServiceImpl {
 
                 for table in tables {
                     let start = tokio::time::Instant::now();
-
-                    let table_name = schema_map
-                        .get(&table.schema_id)
-                        .map(|s| format!("{}.{}", s, table.name))
-                        .unwrap_or_else(|| table.name.clone());
-
                     let req = GetTableReplacePlanRequest {
                         database_id: table.database_id,
                         owner: table.owner,
-                        table_name,
+                        table_name: table.name.clone(),
                         cdc_table_change: None,
                     };
-                    let resp = match client.get_table_replace_plan(req).await {
-                        Ok(resp) => resp,
-                        Err(e) => {
-                            // Skip tables that no longer exist in catalog (orphaned metadata)
-                            if e.to_string().contains("table not found") {
-                                tracing::warn!(
-                                    table_id = table.id,
-                                    table_name = %table.name,
-                                    "Skipping migration for table not found in catalog (orphaned metadata): {}",
-                                    e.as_report()
-                                );
-                                continue;
-                            }
-                            return Err(MetaError::from(anyhow::anyhow!(
-                                "failed to get table replace plan from frontend: {}",
-                                e
-                            )));
-                        }
-                    };
+                    let resp = client
+                        .get_table_replace_plan(req)
+                        .await
+                        .context("failed to get table replace plan from frontend")?;
 
                     let plan = resp.into_inner().replace_plan.unwrap();
                     let replace_info = DdlServiceImpl::extract_replace_table_info(plan);


src/meta/service/src/ddl_service.rs
```diff
@@ -166,16 +172,37 @@ impl DdlServiceImpl {
 
                 for table in tables {
                     let start = tokio::time::Instant::now();
+
+                    let table_name = schema_map
+                        .get(&table.schema_id)
+                        .map(|s| format!("{}.{}", s, table.name))
+                        .unwrap_or_else(|| table.name.clone());
```
Chiro11: It seems the `unwrap_or_else` here is unnecessary. Normally, all schemas should be found in schema_list. If one isn't found, perhaps it should just log and skip.

src/meta/service/src/ddl_service.rs
```diff
@@ -166,16 +172,37 @@ impl DdlServiceImpl {
...
+                        Err(e) => {
+                            // Skip tables that no longer exist in catalog (orphaned metadata)
+                            if e.to_string().contains("table not found") {
+                                tracing::warn!(
+                                    table_id = table.id,
+                                    table_name = %table.name,
+                                    "Skipping migration for table not found in catalog (orphaned metadata): {}",
+                                    e.as_report()
+                                );
+                                continue;
```
Chiro11: And it's better to directly use the previous logic. Because the catalog pub-sub is asynchronous, there might be tables that exist in the metastore but have not been synchronized to the frontend at startup. In this case, it should just wait and retry.

Chiro11 (commented):
The fix for the meta-log error looks fine to me. Thank you very much.

BugenZhao:
> The changes in this PR have resolved issues #23595 and #23708 locally.

Thanks ahead for your contribution. Could you let me know if the fixes for these two issues are relevant? If not, would you split into two PRs to facilitate the review?

wenym1:
@jw-itq Thanks for the investigation and the fix! For issue of table migration, we will raise another PR to fix it together with another related issue. You may only leave the fix on streaming executor in this PR.

jw-itq:
> @jw-itq Thanks for the investigation and the fix! For issue of table migration, we will raise another PR to fix it together with another related issue. You may only leave the fix on streaming executor in this PR.

ok,thanks!

jw-itq:
> > The changes in this PR have resolved issues #23595 and #23708 locally.
> 
> Thanks ahead for your contribution. Could you let me know if the fixes for these two issues are relevant? If not, would you split into two PRs to facilitate the review?

ok, it's fine. Regarding the fix for issue #23595, I have observed the local cluster for a week and it is functioning normally. This pull request also hopes that you can provide a more professional fix solution, as this issue causing the computing nodes to be unavailable is very serious. thanks.

BugenZhao (commented):
Although we _can_ fix unpaired `UpdateDelete` and `UpdateInsert` in the dispatcher, it's important to note that this invariant is relied upon in many places during execution, for example:

https://github.com/risingwavelabs/risingwave/blob/a76476429cd9cfe849178625130caa647c7d3370/src/common/src/array/stream_chunk_iter.rs#L87-L107

So if some executors yield unpaired updates to another executor, without going through a dispatcher (or `Exchange`), we will still panic.

I would prefer to investigate the root cause of the unpaired update rows rather than apply the fix in this PR, which only delays addressing the underlying issue.

jw-itq:
> Although we _can_ fix unpaired `UpdateDelete` and `UpdateInsert` in the dispatcher, it's important to note that this invariant is relied upon in many places during execution, for example:
> 
> https://github.com/risingwavelabs/risingwave/blob/a76476429cd9cfe849178625130caa647c7d3370/src/common/src/array/stream_chunk_iter.rs#L87-L107
> 
> So if some executors yield unpaired updates to another executor, without going through a dispatcher (or `Exchange`), we will still panic.
> 
> I would prefer to investigate the root cause of the unpaired update rows rather than apply the fix in this PR, which only delays addressing the underlying issue.

@BugenZhao  Thanks. Okay, I know my approach may not be professional enough, but that's why I really need help. The issue causing the entire computing node to fail is unacceptable. I believe others in my situation may encounter the same problem. I hope you can provide some solutions or new PR fixes as soon as possible.I urgently need a solution. Thank you!

BugenZhao:
> @BugenZhao Thanks. Okay, I know my approach may not be professional enough, but that's why I really need help. The issue causing the entire computing node to fail is unacceptable. I believe others in my situation may encounter the same problem. I hope you can provide some solutions or new PR fixes as soon as possible.I urgently need a solution. Thank you!

Sure, we will continue looking into the issue. In the meanwhile, you can still use the patched version in your deployment.

jw-itq:
> > @BugenZhao Thanks. Okay, I know my approach may not be professional enough, but that's why I really need help. The issue causing the entire computing node to fail is unacceptable. I believe others in my situation may encounter the same problem. I hope you can provide some solutions or new PR fixes as soon as possible.I urgently need a solution. Thank you!
> 
> Sure, we will continue looking into the issue. In the meanwhile, you can still use the patched version in your deployment.

@BugenZhao Thank you. I will continue to follow the progress. You have worked hard!

PR #23711, opened 2025-11-07
refactor(mysql-cdc): enhance mysql cdc schema check for unsigned int

wcy-fdu:

For unsigned int types(unsigned bigint, unsigned int, unsigned smallint, etc.), the best way to map it in RisingWave is upcast to higher type to avoid overflow. This PR does this during schema check, only handling unsigned bigint specially, allowing it to be converted to bigint (because Debezium's default behavior is to convert to bigint).

## Checklist

## Documentation

<details>

</details>


e2e_test/source_inline/cdc/mysql/test_unsigned_int.slt
```diff
@@ -201,6 +198,34 @@ select id, unsigned_bigint, info from rw_unsigned_int_to_signed where id = 53;
 ----
 53	-492998529773844	debezium
 
+# Test schema check: unsigned int -> int32 should fail
+statement error
```
xiangjinwu: We can include part of the error message here, or it can falsely pass when any error is reported (eg panic):
```
statement error Incompatible data type of column unsigned_int
```

xiangjinwu (commented):
And clean up the negative -> positive logic in src/connector/src/source/cdc/external/mysql.rs as well?

PR #23715, opened 2025-11-07
fix(test): add compatibility test

zwang28:

Add backwards compatibility test for debezium.
#23600

## Checklist

## Documentation

<details>

</details>


src/risedevtool/src/task/sql_server_service.rs
```diff
@@ -42,7 +42,7 @@ impl DockerServiceConfig for SqlServerConfig {
 
     fn data_path(&self) -> Option<String> {
         self.persist_data
-            .then(|| "/var/lib/sqlserver/data".to_owned())
+            .then(|| "/var/opt/mssql/data".to_owned())
```
hzxa21: just to make sure: is this change expected?

zwang28:
Test fails due to error `service requires docker to be installed`.
Because this PR additiaonally starts postgres/mysql/sqlserver in cluster startup via risedev, which is unnecessary. Because they're already started by docker compose.

kwannoel:
> Test fails due to error `service requires docker to be installed`. Because this PR additiaonally starts postgres/mysql/sqlserver in cluster startup via risedev. Need fix.

Rather than start via risedev, you can follow the way we do it for source / sink test. Inside `pull-request.yml`, you will see:
```
  - label: "Backwards compatibility tests"
    command: "VERSION_OFFSET={{matrix.version_offset}} RW_COMMIT=$BUILDKITE_COMMIT ci/scripts/backwards-compat-test.sh -p ci-dev"
...
          run: source-test-env  // <-------- this line.
```

```

  source-test-env:
    image: public.ecr.aws/w1p7b4n3/rw-build-env:${BUILD_ENV_VERSION:?}
    depends_on:
      - mysql
      - mysql-meta
      - sqlserver-server
      - db
      - message_queue
      - schemaregistry
      - pulsar-server
      - mongodb
      - mongodb-setup
      - mongo_data_generator
      - nats-server
      - vault-server
      - message_queue_sasl_1
      - message_queue_sasl_2
      - rabbitmq-server
    volumes:
      - ..:/risingwave
    stop_grace_period: 30s
```

You can add the docker services for your test services here.

src/risedevtool/src/task/sql_server_service.rs
zwang28: In my local test, the sqlserver started by risedev is persisted to `/var/lib/sqlserver`, which is not a persistent path. Therefore, the SQL Server instance in risedev does not persist data across restarts.
In constrast, `/var/opt/mssql/data` is a path mounted to local disk path `.risingwave/data`, which is pesistent. However, I can't find the code where this mount is specified.

Anyway I revert this change, because this PR doesn't rely on risedev's sqlserver instance.

zwang28:
The test is flaky even with retry. It fails at `validate_original` so it's irrelevant to backward compatibility.
I suspect the barrier latency is high due to other tests prior to this one.

zwang28:
> The test is flaky even with retry. It fails at `validate_original` so it's irrelevant to backward compatibility. I suspect the barrier latency is high due to other tests prior to this one.

After removing the CPU:4 constraint, the test is no longer flaky. However I have to remove sql server part from the test because the license key allows only 4 cores.

PR #23721, opened 2025-11-08
fix(executor): handle large gap in gap_filling

guluguluhhhh:

This PR refactors the gap fill executor to use streaming approaches for generating and scanning filled rows, significantly reducing memory usage when handling large time gaps.

### Stream-based Generation of Filled Rows

When generating filled rows between two original rows, the implementation now uses a streaming approach (`try_stream`) that yields rows incrementally through the `chunk_builder` instead of collecting them all in a `Vec` first. This prevents memory accumulation when there are large time gaps that would require generating many filled rows.

**Before**: All filled rows were generated and collected in a `Vec<OwnedRow>` before being output.  
**After**: Filled rows are generated and immediately appended to chunks via `chunk_builder`, allowing them to be streamed out as they are produced.

### Stream-based Scanning with Borrow Checker Constraints

The scanning of filled rows (for deletion purposes) has been refactored to use streaming. However, when deleting filled rows in the normal gap fill mode when new original rows arrive, we still need to collect all rows before deletion due to **Rust's borrow checker constraints**:

- **Scanning** requires an **immutable borrow** of `cache_manager` and `managed_state`.
- **Deletion** requires a **mutable borrow** of these same structures.
- Rust's borrow checker prevents holding both borrows simultaneously.

## Checklist

## Documentation

<details>
fix(executor): handle large gap

</details>

### DIFF
--- a/src/stream/src/executor/eowc/eowc_gap_fill.rs
+++ b/src/stream/src/executor/eowc/eowc_gap_fill.rs
@@ -80,17 +80,13 @@ impl<S: StateStore> ExecutorInner<S> {
         metrics: &'a GapFillMetrics,
         chunk_builder: &'a mut StreamChunkBuilder,
     ) {
-        let (Some(prev_time_scalar), Some(curr_time_scalar)) = (
-            prev_row.datum_at(time_column_index),
-            curr_row.datum_at(time_column_index),
-        ) else {
-            tracing::error!(
-                "Gap fill skipped: time column contains NULL values. prev_row: {:?}, curr_row: {:?}",
-                prev_row,
-                curr_row
-            );
-            return Ok(());
-        };
+        // Time column should never be NULL due to validation in execute_inner
+        let prev_time_scalar = prev_row
+            .datum_at(time_column_index)
+            .expect("Gap fill time column must be non-NULL");
+        let curr_time_scalar = curr_row
+            .datum_at(time_column_index)
+            .expect("Gap fill time column must be non-NULL");
 
         let prev_time = match prev_time_scalar {
             ScalarRefImpl::Timestamp(ts) => ts,
@@ -352,6 +348,12 @@ impl<S: StateStore> EowcGapFillExecutor<S> {
                         .consume(watermark.val.clone(), &mut this.buffer_table)
                     {
                         let current_row = row?;
+
+                        // Validate that time column is not NULL
+                        let _ = current_row
+                            .datum_at(this.time_column_index)
+                            .expect("Gap fill time column must be non-NULL");
+
                         if let Some(p_row) = &staging_prev_row {
                             #[for_await]
                             for chunk in ExecutorInner::<S>::generate_filled_rows(
--- a/src/stream/src/executor/gap_fill.rs
+++ b/src/stream/src/executor/gap_fill.rs
@@ -684,17 +684,13 @@ impl<S: StateStore> GapFillExecutor<S> {
         config: &'a FillConfig<'a>,
         state: &'a mut FillState<'a, S>,
     ) {
-        let (Some(prev_time_scalar), Some(curr_time_scalar)) = (
-            prev_row.datum_at(config.time_column_index),
-            curr_row.datum_at(config.time_column_index),
-        ) else {
-            tracing::error!(
-                "Gap fill skipped: time column contains NULL values. prev_row: {:?}, curr_row: {:?}",
-                prev_row,
-                curr_row
-            );
-            return Ok(());
-        };
+        // Time column should never be NULL due to validation in execute_inner
+        let prev_time_scalar = prev_row
+            .datum_at(config.time_column_index)
+            .expect("Gap fill time column must be non-NULL");
+        let curr_time_scalar = curr_row
+            .datum_at(config.time_column_index)
+            .expect("Gap fill time column must be non-NULL");
 
         let prev_time = match prev_time_scalar {
             ScalarRefImpl::Timestamp(ts) => ts,
@@ -938,6 +934,11 @@ impl<S: StateStore> GapFillExecutor<S> {
 
                         match op {
                             Op::Insert | Op::UpdateInsert => {
+                                // Validate that time column is not NULL
+                                let _ = row
+                                    .datum_at(time_column_index)
+                                    .expect("Gap fill time column must be non-NULL");
+
                                 let cache_key = managed_state.serialize_time_to_cache_key(&row);
 
                                 // Find previous and next original row neighbors.
@@ -1068,6 +1069,11 @@ impl<S: StateStore> GapFillExecutor<S> {
                                 }
                             }
                             Op::Delete | Op::UpdateDelete => {
+                                // Validate that time column is not NULL
+                                let _ = row
+                                    .datum_at(time_column_index)
+                                    .expect("Gap fill time column must be non-NULL");
+
                                 let cache_key = managed_state.serialize_time_to_cache_key(&row);
 
                                 // Find previous and next original row neighbors before deletion.


guluguluhhhh:
Hi @kwannoel @yuhao-su, could you please help review this PR?

kwannoel:
> Hi @kwannoel @yuhao-su, could you please help review this PR?

Thanks for the PR! Could you help to update the PR description?

guluguluhhhh:
> > Hi @kwannoel @yuhao-su, could you please help review this PR?
> 
> Thanks for the PR! Could you help to update the PR description?

Done!

src/stream/src/executor/eowc/eowc_gap_fill.rs
```diff
@@ -71,20 +71,23 @@ struct ExecutionVars<S: StateStore> {
...
+        chunk_builder: &'a mut StreamChunkBuilder,
+    ) {
+        let Some((prev_time_scalar, curr_time_scalar)) = (|| {
+            Some((
+                prev_row.datum_at(time_column_index)?,
+                curr_row.datum_at(time_column_index)?,
+            ))
+        })() else {
+            return Ok(());
         };
```
kwannoel: This looks complicated. Can you simplify it?

src/stream/src/executor/eowc/eowc_gap_fill.rs
```diff
@@ -71,20 +71,23 @@ struct ExecutionVars<S: StateStore> {
...
+        metrics: &'a GapFillMetrics,
+        chunk_builder: &'a mut StreamChunkBuilder,
+    ) {
+        let Some((prev_time_scalar, curr_time_scalar)) = (|| {
+            Some((
+                prev_row.datum_at(time_column_index)?,
+                curr_row.datum_at(time_column_index)?,
+            ))
+        })() else {
+            return Ok(());
```
kwannoel: We should not return `Ok(())` if the row schema is mismatched. It should return an error.

src/stream/src/executor/eowc/eowc_gap_fill.rs
```diff
@@ -94,7 +97,7 @@ impl<S: StateStore> ExecutorInner<S> {
                     Ok(timestamp) => timestamp,
                     Err(_) => {
                         warn!("Failed to convert timestamptz to timestamp: {:?}", ts);
-                        return Ok(filled_rows);
+                        return Ok(());
```
kwannoel, line 101: It should return an error.

src/stream/src/executor/eowc/eowc_gap_fill.rs
```diff
@@ -103,7 +106,7 @@ impl<S: StateStore> ExecutorInner<S> {
                     "Failed to convert time column to timestamp, got {:?}. Skipping gap fill.",
                     prev_time_scalar
                 );
-                return Ok(filled_rows);
```
kwannoel, line 106: It should return an error.

src/stream/src/executor/eowc/eowc_gap_fill.rs
```diff
@@ -123,24 +126,24 @@ impl<S: StateStore> ExecutorInner<S> {
...
         let mut fill_time = match prev_time.checked_add(interval) {
             Some(t) => t,
             None => {
-                return Ok(filled_rows);
+                return Ok(());
             }
         };
         if fill_time >= curr_time {
-            return Ok(filled_rows);
+            return Ok(());
```
kwannoel, line 155: Please take a check to verify if all these should return `Ok(())`.

src/stream/src/executor/gap_fill.rs
```diff
@@ -555,60 +528,63 @@ impl GapFillCacheManager {
...
-        let range_fully_in_cache = self.contains_time(start_time) && self.contains_time(end_time);
-
-        if range_fully_in_cache {
-            for (_cache_key, (compacted_row, row_type)) in self.cache.range::<GapFillCacheKey, _>((
+    pub fn scan_filled_rows_from_cache<'a, S: StateStore>(
+        &'a self,
+        start_time: &'a GapFillCacheKey,
+        end_time: &'a GapFillCacheKey,
+        managed_state: &'a ManagedGapFillState<S>,
+    ) -> impl Iterator<Item = StreamExecutorResult<(OwnedRow, OwnedRow)>> + 'a {
```
kwannoel: Shouldn't it be this instead:
```
StreamExecutorResult<impl Iterator ...>
```

src/stream/src/executor/gap_fill.rs
```diff
@@ -721,46 +712,44 @@ impl<S: StateStore> GapFillExecutor<S> {
...
                 );
-                return Ok(filled_rows);
+                return Ok(());
             }
         };
 
         // Check if fill_time is already >= curr_time, which means no gap to fill
         if fill_time >= curr_time {
-            return Ok(filled_rows);
+            return Ok(());
```
kwannoel, line 761: As above. Please validate if we should return errors.

guluguluhhhh: I think it’s better to return an error when the time column is NULL or its type is not Timestamp or Timestamptz.
For checked_add overflow, I log a warning and skip the gap-filling.

src/stream/src/executor/eowc/eowc_gap_fill.rs
```diff
@@ -71,76 +70,84 @@ struct ExecutionVars<S: StateStore> {
...
+        chunk_builder: &'a mut StreamChunkBuilder,
+    ) {
         let (Some(prev_time_scalar), Some(curr_time_scalar)) = (
             prev_row.datum_at(time_column_index),
             curr_row.datum_at(time_column_index),
         ) else {
-            return Ok(filled_rows);
+            return Err(ExprError::InvalidParam {
+                name: "time_column",
+                reason: "Time column contains NULL values".into(),
```
kwannoel: Is it possible for the user to somehow provide null values in their `prev_row` and `curr_row`? If so we should just return an error log and just not gapfill. Otherwise we might just have crash loop which is poor UX.

kwannoel: I suppose returning `Ok(())` is fine here.

src/stream/src/executor/eowc/eowc_gap_fill.rs
```diff
@@ -71,76 +70,84 @@ struct ExecutionVars<S: StateStore> {
...
-                    }
-                }
-            }
+            ScalarRefImpl::Timestamptz(ts) => risingwave_common::types::Timestamp::with_micros(
+                ts.timestamp_micros(),
+            )
+            .map_err(|_| ExprError::InvalidParam {
+                name: "time_column",
+                reason: format!("Failed to convert timestamptz to timestamp: {:?}", ts).into(),
+            })?,
```
kwannoel: Returning `Ok(())` and an error log is fine here.

src/stream/src/executor/gap_fill.rs
kwannoel: Makes sense.

src/stream/src/executor/eowc/eowc_gap_fill.rs
guluguluhhhh: > Is it possible for the user to somehow provide null values in their `prev_row` and `curr_row`? If so we should just return an error log and just not gapfill. Otherwise we might just have crash loop which is poor UX.

Yes, it is possible for now. But I've tested this scenario in Tiger Data, and they return an error like `ERROR:  invalid time_bucket_gapfill argument: ts cannot be NULL`.

guluguluhhhh: I think I haven't handled null time-column values properly as well. Maybe it's better to disallow gap fill when the time column contains null values.
I notice that in over_window we panic with `.expect("order key column must be non-NULL")`. Should I panic or return an error when gap fill encounters a null time column? @kwannoel

kwannoel: We can follow the behaviour of over_window and panic.

guluguluhhhh: done!

PR #23729, opened 2025-11-10
fix: handle uninitialized streaming vnode mapping with explicit error variant

shanicky:

**Summary**
This PR introduces explicit handling for the uninitialized state of the streaming vnode mapping by changing its type to an `Option<HashMap>`, adding a dedicated error variant, and updating related code paths accordingly. These changes improve robustness and clarity in worker node management, especially during system startup and recovery.

**Details**
- Added a new `StreamingVnodeMappingNotInitialized` error variant to clearly signal when the vnode mapping is not yet initialized.
- Changed `streaming_fragment_vnode_mapping` from a plain `HashMap` to an `Option<HashMap>`, allowing explicit representation of an uninitialized state (`None`).
- Updated getter methods to return the new error when the vnode mapping is uninitialized, enabling callers to handle this condition explicitly.
- Mutation methods lazily initialize the vnode mapping on first use, ensuring safe and efficient updates.
- Adjusted simulation tests to expect and handle the new error, including making error string comparisons case-insensitive to improve test resilience.
- Added informative logging when vnode mappings are replaced, aiding debugging of scaling and background DDL operations.

## Checklist

### DIFF
--- a/src/batch/src/error.rs
+++ b/src/batch/src/error.rs
@@ -135,6 +135,9 @@ pub enum BatchError {
     #[error("Serving vnode mapping not found for fragment {0}")]
     ServingVnodeMappingNotFound(FragmentId),
 
+    #[error("Streaming vnode mapping has not been initialized")]
+    StreamingVnodeMappingNotInitialized,
+
     #[error("Streaming vnode mapping not found for fragment {0}")]
     StreamingVnodeMappingNotFound(FragmentId),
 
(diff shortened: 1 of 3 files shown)


BugenZhao (approved):
IIUC, this is to differentiate the transient error against the real error, and we believe it's retryable in recovery test?

yezizp2012 (approved):

shanicky:
> IIUC, this is to differentiate the transient error against the real error, and we believe it's retryable in recovery test?

Yes

PR #23735, opened 2025-11-11
feat(streaming): add functionality to override config when building actors

BugenZhao:

Design RFC: [Notion](https://www.notion.so/risingwave-labs/mini-RFC-Per-Job-Configuration-2a8f0d69cb768032b2acf1bb9bfee979?source=copy_link)

This PR adds the functionality to override the actor's local config (introduced in #23658) with given `config_override` string when building actors.

In future work, user will be able to override some specific configuration entry via SQL interface, and we will persist the change with this `config_override` string. Currently, it's always empty.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/task/actor_manager.rs
+++ b/src/stream/src/task/actor_manager.rs
@@ -436,9 +436,7 @@ impl StreamActorManager {
         new_output_request_rx: UnboundedReceiver<(ActorId, NewOutputRequest)>,
     ) -> StreamResult<Actor<DispatchExecutor>> {
         let actor_id = actor.actor_id;
-
-        // TODO(config): use user specified real config override value.
-        let actor_config = self.get_overridden_config("", actor_id);
+        let actor_config = self.get_overridden_config(&actor.config_override, actor_id);
 
         let actor_context = ActorContext::create(
             &actor,


BugenZhao:

- #23752
    - #23735 <-- you are here 🤪
        - #23768

kwannoel:
> This change is part of the following stack:
> 
> - #23735 <-- you are here 🤪
>     - #23752
> 
> <sub>Change managed by [git-spice](https://abhinav.github.io/git-spice/).</sub>
> 
> 

Does git spice generate the commit message?

BugenZhao:
> Does git spice generate the commit message?

Yes. Auto generated.

src/stream/src/task/actor_manager.rs
```diff
@@ -50,6 +50,11 @@ use crate::task::{
     StreamEnvironment, await_tree_key,
 };
 
+/// Default capacity for `ConfigOverrideCache`.
+/// Since we only support per-job config override right now, 256 jobs should be sufficient on a single node.
+pub const CONFIG_OVERRIDE_CACHE_DEFAULT_CAPACITY: u64 = 256;
```
kwannoel, line 55: Is there anyway to warn if we exceed 256?

BugenZhao: I don't think that's necessary. The cache is only used when building actors. Unless we're concurrently building actors for more than 256 jobs, the cache can always reduce the call to `merge_streaming_config_section` from `O(number of actors)` to `O(1)`.

PR #23737, opened 2025-11-11
feat: add periodic refresh table jobs & refactor ProgressTracker

tabVersion:
- Introduced a new migration for periodic refresh jobs, including the creation of the `periodic_refresh_jobs` table.
- Implemented the `GlobalRefreshManager` to manage ongoing refresh processes and periodic refresh jobs.
- Added functionality to initialize and trigger periodic refresh jobs based on defined intervals.
- Updated relevant modules to support the new refresh management system.

following https://github.com/risingwavelabs/risingwave/pull/23527

Key aspects of this change include:

1. Refactored Refresh Mode: The FULL_RECOMPUTE enum and related logic have been renamed and replaced with FULL_RELOAD, better reflecting the operation of reloading data from external sources.
2. Persistent Refresh Job State: A new database model refresh_job is introduced to store the state (e.g., IDLE, REFRESHING), last trigger time, and configured refresh interval for each refreshable table. This state is now persistent and managed by the meta service.
3. Scheduled Refreshes: The FULL_RELOAD mode now supports an optional refresh_interval_sec property, allowing users to configure tables to automatically reload their data at specified intervals.
4. User-facing Observability: A new system catalog, rw_catalog.rw_refresh_table_state, has been added. Users can query this table to view the current status, last trigger time, and configured interval for all refreshable tables.
5. Monitoring: New Grafana dashboards have been added to track metrics related to refresh job durations, finish rates, and cron job triggers/misses, providing deeper insights into the refresh process.
6. Integration: The GlobalRefreshManager is tightly integrated with the meta service's barrier management and catalog operations, ensuring consistent and controlled refreshes.

## Checklist

## Documentation

<details>

---

**PLEASE MARK AS EXPERIMENTAL**

Now, with FULL_RELOAD, you gain:
* Scheduled Refreshes: Configure refreshable tables to automatically reload their data at specified intervals.
* Persistent Job State: RisingWave now keeps track of the refresh status, last trigger time, and interval for each FULL_RELOAD table, even across restarts.
* Enhanced Observability: A new system catalog rw_catalog.rw_refresh_table_state allows you to monitor the status of your refresh jobs.

This means you can set up data sources (like Iceberg tables) to be periodically reloaded into RisingWave, ensuring your views and queries are always up-to-date with the latest batch data, without manual intervention.

## How it Works

When creating a refreshable table, you can now specify a FULL_RELOAD mode with an optional refresh_interval_sec property:

```
CREATE TABLE iceberg_batch_table (
    id int primary key,
    name varchar
) WITH (
    connector = 'iceberg',
    catalog.type = 'storage',
    table.name = 'my_iceberg_table',
    database.name = 'public',
    refresh_mode = 'FULL_RELOAD', -- MUST set to `FULL_RELOAD`
    refresh_interval_sec = '60' -- reload the table every 60s
);
```

still can manually refresh the table

```
REFRESH TABLE iceberg_batch_table;
```

the status can be queried with sql 

```
SELECT table_id, current_status, last_trigger_time, trigger_interval_secs
FROM rw_catalog.rw_refresh_table_state;
```

</details>


tabVersion:
will support sql interface (for with options) in https://github.com/risingwavelabs/risingwave/pull/23780

chenzl25:
@tabVersion This is a user-facing feature, so please add a release note to describe this feature and provide an example to illustrate how to use it, thanks.

hzxa21:
> @tabVersion This is a user-facing feature, so please add a release note to describe this feature and provide an example to illustrate how to use it, thanks.

Please also mention that this is an experimental feature for doc team's reference.

proto/plan_common.proto
```diff
@@ -307,7 +307,9 @@ enum AdditionalColumnType {
 
 message SourceRefreshMode {
   message SourceRefreshModeStreaming {}
-  message SourceRefreshModeFullRecompute {}
+  message SourceRefreshModeFullRecompute {
```
hzxa21: User facing question: after this PR, user specifies `refresh_mode = 'FULL_RECOMPUTE'` in order to use the refreshable batch source table feature. Personally I feel `FULL_RECOMPUTE` can mislead user to think that there will be full recomputation on the MV instead of just on the table to calculate the diff.

I am thinking whether we should call it `refresh_mode = 'SNAPSHOT_DIFF'` instead. WDYT? @tabVersion @chenzl25

chenzl25: We borrow the same name from here https://docs.databricks.com/aws/en/optimizations/incremental-refresh#determine-the-refresh-type-of-an-update I think it is fine, because that's how to refresh this table.snapshot diff is more like something we generate it to the downstream.

tabVersion: had an offline discussion, the name finalized to `FULL_RELOAD`

PR #23739, opened 2025-11-11
fix(mssql): correctly handle cases where `database_name` is omitted in create table

wcy-fdu:

As title, onlt allow user to omit sqlserver  database name when creating table from source. The correct statement are
```
create table from s schema.table_name;
```
or
```
create table from s database_name.schema_name.table_name;
```

Reject other cases.

Unify full_table_name extract logic for all CDCs.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/java/connector-node/risingwave-source-cdc/src/main/java/io/debezium/connector/binlog/BinlogOffsetContext.java
+++ b/java/connector-node/risingwave-source-cdc/src/main/java/io/debezium/connector/binlog/BinlogOffsetContext.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2025 RisingWave Labs
+ * Copyright 2026 RisingWave Labs
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/java/connector-node/risingwave-source-cdc/src/main/java/io/debezium/connector/binlog/history/BinlogHistoryRecordComparator.java
+++ b/java/connector-node/risingwave-source-cdc/src/main/java/io/debezium/connector/binlog/history/BinlogHistoryRecordComparator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2025 RisingWave Labs
+ * Copyright 2026 RisingWave Labs
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/java/connector-node/risingwave-source-cdc/src/main/java/io/debezium/connector/mysql/MySqlOffsetContext.java
+++ b/java/connector-node/risingwave-source-cdc/src/main/java/io/debezium/connector/mysql/MySqlOffsetContext.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2025 RisingWave Labs
+ * Copyright 2026 RisingWave Labs
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/java/connector-node/risingwave-source-cdc/src/main/java/io/debezium/openlineage/DebeziumOpenLineageEmitter.java
+++ b/java/connector-node/risingwave-source-cdc/src/main/java/io/debezium/openlineage/DebeziumOpenLineageEmitter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2025 RisingWave Labs
+ * Copyright 2026 RisingWave Labs
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/java/connector-node/risingwave-source-cdc/src/main/java/io/debezium/pipeline/ChangeEventSourceCoordinator.java
+++ b/java/connector-node/risingwave-source-cdc/src/main/java/io/debezium/pipeline/ChangeEventSourceCoordinator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2025 RisingWave Labs
+ * Copyright 2026 RisingWave Labs
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/src/common/common_service/src/profile_service.rs
+++ b/src/common/common_service/src/profile_service.rs
@@ -1,4 +1,4 @@
-// Copyright 2025 RisingWave Labs
+// Copyright 2026 RisingWave Labs
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/src/expr/impl/src/scalar/array_reverse.rs
+++ b/src/expr/impl/src/scalar/array_reverse.rs
@@ -1,4 +1,4 @@
-// Copyright 2025 RisingWave Labs
+// Copyright 2026 RisingWave Labs
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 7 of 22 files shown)


src/connector/src/source/cdc/source/message.rs
```diff
@@ -69,13 +69,63 @@ impl DebeziumCdcMeta {
...
+    /// The number of characters to skip from the start of `full_table_name` when calling
+    /// `extract_table_name()`. This value is used to strip the database prefix for
+    /// Postgres/SQL Server while keeping the full name for MySQL.
+    fn extract_db_name_prefix_len_from_full_table_name(full_table_name: &str) -> usize {
+        if let Some(first_dot) = full_table_name.find('.') {
+            // Check if there's a second dot after the first one
+            if full_table_name[first_dot + 1..].find('.').is_some() {
+                // Found 2 dots (3 parts): Postgres/SQL Server format "database.schema.table"
+                // Return position after first dot to skip "database." part
+                first_dot + 1
```
hzxa21: Prior to this PR, `db_name_prefix_len` for `mydb.dbo.orders` is 5 while after this PR it is 6. Is this expected?

src/connector/src/parser/mod.rs
```diff
@@ -118,7 +118,8 @@ impl<'a> MessageMeta<'a> {
                     CDC_TABLE_NAME_COLUMN_NAME,
                     "unexpected cdc meta column name"
                 );
-                Some(cdc_meta.full_table_name.as_str().into())
+                // For CDC sources, extract_table_name() strips the database prefix:
+                cdc_meta.extract_table_name()
```
zwang28, line 122: Is this change backward-compatible?
e.g. an existing connector's full_table_name might not contains database_name, so the "strips the database prefix" result in incorrect result?
i.e. is the full_table_name used here speicifed by user when creating connector, or by debezium?

src/connector/src/source/cdc/source/message.rs
```diff
@@ -69,13 +69,63 @@ impl DebeziumCdcMeta {
...
+    /// - **MySQL** has no schema concept, so the "table name" IS `database.table`
+    /// - **Postgres/SQL Server** have schemas, and the database is specified in the source,
+    ///   so the "table name" is just `schema.table`
+    ///
+    /// # Returns
+    ///
+    /// The number of characters to skip from the start of `full_table_name` when calling
+    /// `extract_table_name()`. This value is used to strip the database prefix for
+    /// Postgres/SQL Server while keeping the full name for MySQL.
+    fn extract_db_name_prefix_len_from_full_table_name(full_table_name: &str) -> usize {
```
zwang28: Counting dot it not a reilable method. The period (.) can appear in db/schema/table names.

src/connector/src/source/cdc/source/message.rs
wcy-fdu: This PR only changes the frontend parsing in `create_table.rs` to allow users to write `"dbo.orders"` instead of `"mydb.dbo.orders"`. It doesn't modify the `extract_db_name_prefix_len_from_full_table_name()` function in `message.rs`, which always calculates based on Debezium's full table name (still "mydb.dbo.orders").

For "mydb.dbo.orders": first_dot at index 4 → returns 4+1 = 5 → extracts "dbo.orders" correctly.

src/connector/src/source/cdc/source/message.rs
wcy-fdu: >the period (.) can appear in db/schema/table names.

No, we have implemented strict restrictions: entries in db_name/schema/table containing a dot (.) will be rejected by the frontend to ensure
- postgres: must be db.schema.table
- mysql: must be db.table
- sqlserver: must be db.schema.table or schema.table

src/connector/src/parser/mod.rs
wcy-fdu: The `full_table_name` here is always from Debezium's topic (not user input), which consistently includes the database name regardless of how users create the table. So there's no backward compatibility issue - Debezium always sends the complete format.

zwang28 (approved):

src/connector/src/source/cdc/source/message.rs
hzxa21: Still don't get it. 

Prior to this PR:
```rust
let db_name_prefix_len = full_table_name.as_str().find('.').unwrap_or(0);
```

After this PR:
```rust
let db_name_prefix_len =
            Self::extract_db_name_prefix_len_from_full_table_name(&full_table_name);
```

extract_db_name_prefix_len_from_full_table_name will return 0 if it only contains one dot. That means db_name_prefix_len will be different.

wcy-fdu: I understand your point of confusion.

The previous logic was that regardless of which CDC it is, we obtain the table name based on the dot (.) notation. This can easily lead to errors in SQL Server CDC, because the full table name in SQL Server can be either db.schema.table or schema.table.

Additionally, there is another issue: we used the same `db_name_prefix_len` to [derive the database name and table name from the full_table_name ](https://github.com/risingwavelabs/risingwave/blob/b10151b74fc951d31fa57f5abbf46702cd9219ef/src/connector/src/source/cdc/source/message.rs#L60-L70)(for PG/SQL Server, it is schema.table, and for MySQL, it is just table). **This is incorrect in any case.**

Eg. `mydb.dbo.orders`
- db_name_prefix_len = 4 means `extract_database_name`=`mydb`, `extract_table_name`=`.dbo.orders`
- db_name_prefix_len = 5 means `extract_database_name`=`mydb.`, `extract_table_name`=`dbo.orders`

Both of these(4 or 5) are incorrect, **but the issue with the table name is more serious. A table name mismatch can prevent new rows from being read, while the database name is only used when INCLUDE DATABASE_NAME.** So previously this pr aim to make table name part correct, and make you confuse(why from 4 to 5)

It is **not** sufficient to simply change extract_database_name to [0..self.db_name_prefix_len-1], because in some cases, db_name_prefix_len may be 0 (for example, when SQL Server omits the database name).

**The actual objective here is to correctly parse the `db_name` and `table_name` for various forms of `full_table_name`.** Therefore, it should be changed to store two fields: db_name_end and `table_name_start` and
- extract_database_name = full_table_name[0..db_name_end]
- extract_table_name = full_table_name[table_name_start..].

But above method still unable to distinguish
- mysql: db.table
- sqlserver: schema.table

wcy-fdu: Will bring cdcType in `DebeziumCdcMeta` to resolve.

wcy-fdu: This comment is resolved. cc @hzxa21 can take another look

src/connector/src/parser/utils.rs
```diff
@@ -134,7 +134,9 @@ pub fn extract_cdc_meta_column<'a>(
     match column_type {
         ColumnType::Timestamp(_) => Ok(cdc_meta.extract_timestamp()),
         ColumnType::DatabaseName(_) => Ok(cdc_meta.extract_database_name()),
-        ColumnType::TableName(_) => Ok(cdc_meta.extract_table_name()),
+        // `table_name` in additional columns should be the object name only, not the routing key
+        // (which can be `schema.table` or `db.table` depending on the connector).
+        ColumnType::TableName(_) => Ok(cdc_meta.extract_table_name_only()),
```
zwang28, line 139: Is this extracted table name used anywhere?

hzxa21: I had the same question when reviewing the recent fix PR #25057

wcy-fdu: `extract_table_name_only` is used to populate the CDC `INCLUDE TABLE_NAME` additional column (object name only), not for `_rw_table_name` routing/filtering.

PR #23740, opened 2025-11-11
fix(hummock): backward iterator for imm should return older version first

jrthe42:

Currently, if multiple versions of a key exist in imm, the backward iterator will return values in DESCENDING epoch order. But in sst file, it will return values in ASCENDING epoch order. This will cause conflict in BackwardUserIterator.

This PR fix the issue and make sure backward iterator for imm return values in ASCENDING epoch order.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/storage/src/hummock/iterator/backward_user.rs
+++ b/src/storage/src/hummock/iterator/backward_user.rs
@@ -317,16 +317,22 @@ mod tests {
     use risingwave_hummock_sdk::sstable_info::SstableInfo;
 
     use super::*;
-    use crate::hummock::iterator::MergeIterator;
+    use crate::hummock::compactor::merge_imms_in_memory;
     use crate::hummock::iterator::test_utils::{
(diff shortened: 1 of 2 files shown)


wenym1 (commented):
Thanks for the careful investigation!

Although this is a bug in imm backward iter implementation, it only causes problem when there are multiple versions of value for a key in a single imm. However, we have deprecated the logic of merging multiple imms long time ago, and without imm merge, there won't be multi-version values for a key in a single imm. Therefore, seeing from higher level, there is no bug.

This might lead to a discussion on whether or not to remove the code of merge_imm and  simplify the struct of SharedBufferBatchInner (imm), to flatten the key and value one on one in a single vec, rather than in current way, which is dedicated for storing multi-version values. @hzxa21 @Li0k What do you think?

jrthe42:
Oh, I see, no wonder this is not causing issue in your integration tests and production deployments. Maybe it's better to simplify the struct of imm. I'll close this issue first.

PR #23742, opened 2025-11-11
feat(parser): support `ALTER .. [SET | RESET] CONFIG`

BugenZhao:

Design RFC: [Notion](https://www.notion.so/risingwave-labs/mini-RFC-Per-Job-Configuration-2a8f0d69cb768032b2acf1bb9bfee979?source=copy_link)

Note that an extra keyword `CONFIG` is introduced and used because plain `SET` and `RESET` should be used to alter the entries from `WITH (..)`, according to the convention of Postgres syntax. Although we've decided to use `ALTER CONNECTOR WITH` for this instead.

## Checklist

## Documentation

<details>

</details>


BugenZhao:

- #23742 <-- you are here 🤪
    - #23748

PR #23744, opened 2025-11-11
refactor(meta): maintain per job create mview progress separately

wenym1:

As titled.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/meta/src/barrier/checkpoint/control.rs
+++ b/src/meta/src/barrier/checkpoint/control.rs
@@ -817,11 +817,9 @@ impl DatabaseCheckpointControl {
 
                 self.handle_refresh_table_info(task, &node);
 
-                self.state.inflight_graph_info.apply_create_mview_progress(
-                    node.state
-                        .resps
-                        .iter()
-                        .flat_map(|resp| &resp.create_mview_progress),
+                self.state.inflight_graph_info.apply_collected_command(
+                    node.command_ctx.command.as_ref(),
+                    node.state.resps.iter(),
                     hummock_version_stats,
                 );
                 let finished_cdc_backfill = self
@@ -1160,9 +1158,7 @@ impl DatabaseCheckpointControl {
             table_ids_to_commit,
             jobs_to_wait,
             prev_paused_reason,
-        ) = self
-            .state
-            .apply_command(command.as_ref(), hummock_version_stats);
+        ) = self.state.apply_command(command.as_ref());
 
         // Tracing related stuff
         barrier_info.prev_epoch.span().in_scope(|| {
--- a/src/meta/src/barrier/checkpoint/state.rs
+++ b/src/meta/src/barrier/checkpoint/state.rs
@@ -19,7 +19,6 @@ use std::mem::take;
 use risingwave_common::catalog::{DatabaseId, TableId};
 use risingwave_common::id::JobId;
 use risingwave_common::util::epoch::Epoch;
-use risingwave_pb::hummock::HummockVersionStats;
 use tracing::warn;
 
 use crate::barrier::info::{
@@ -137,7 +136,6 @@ impl BarrierWorkerState {
     pub fn apply_command(
         &mut self,
         command: Option<&Command>,
-        hummock_version_stats: &HummockVersionStats,
     ) -> (
         InflightDatabaseInfo,
         HashMap<TableId, u64>,
@@ -152,11 +150,11 @@ impl BarrierWorkerState {
         }) = command
         {
             None
-        } else if let Some((new_job, fragment_changes)) =
-            command.and_then(|command| command.fragment_changes(hummock_version_stats))
+        } else if let Some((new_job_id, fragment_changes)) =
+            command.and_then(Command::fragment_changes)
         {
             self.inflight_graph_info
-                .pre_apply(new_job, &fragment_changes);
+                .pre_apply(new_job_id, &fragment_changes);
             Some(fragment_changes)
         } else {
             None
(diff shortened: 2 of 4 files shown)


wenym1:
* **#23795** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23795?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#23754** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23754?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#23744** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23744?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23744?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/meta/src/barrier/checkpoint/creating_job/mod.rs
```diff
@@ -627,23 +640,34 @@ impl CreatingStreamingJobControl {
...
+        InflightFragmentInfo::existing_table_ids(self.fragment_infos.values())
     }
 
-    pub fn graph_info(&self) -> &InflightStreamingJobInfo {
-        &self.job_info
+    pub fn graph_info(&self) -> &HashMap<FragmentId, InflightFragmentInfo> {
+        &self.fragment_infos
+    }
+
+    pub fn into_tracking_job(self) -> TrackingJob {
```
wenym1, line 662: review note: has the same semantic as the removed `Self::is_finish()`

src/meta/src/stream/source_manager.rs
```diff
@@ -474,7 +474,7 @@ impl SourceManager {
     }
 }
 
-#[derive(strum::Display, Debug)]
+#[derive(strum::Display, Debug, Clone)]
```
wenym1, line 475: review note: the derived Clone is added temporarily in this PR to make the processing logic focused on related changes. Will remove the derived Clone added in this PR in later PR.

src/meta/src/barrier/checkpoint/creating_job/status.rs
```diff
@@ -115,13 +116,15 @@ pub(super) enum CreatingStreamingJobStatus {
...
+        tracking_job: TrackingJob,
         log_store_progress_tracker: CreateMviewLogStoreProgressTracker,
         barriers_to_inject: Option<Vec<BarrierInfo>>,
     },
     /// All backfill actors have started consuming upstream, and the job
     /// will be finished when all previously injected barriers have been collected
     /// Store the `prev_epoch` that will finish at.
-    Finishing(u64),
+    Finishing(u64, TrackingJob),
+    PlaceHolder,
```
kwannoel, line 127: When do we transition to `PlaceHolder` state?

src/meta/src/barrier/checkpoint/creating_job/status.rs
```diff
@@ -158,13 +158,26 @@ impl CreatingStreamingJobStatus {
...
+                        backfill_epoch,
+                        snapshot_backfill_actors,
+                        ..
+                    } = replace(self, CreatingStreamingJobStatus::PlaceHolder)
+                    else {
+                        unreachable!()
+                    };
+
+                    let tracking_job = create_mview_tracker.into_tracking_job();
+
```
kwannoel, line 172: Not sure why we need to use `PlaceHolder` here, since it will immediately be re-assigned to `Self::ConsumingLogStore` later.

Instead of adding a new internal state to `CreatingStreamingJobStatus`, shall we just:
1. Clone `create_mview_tracker`, `backfill_epoch`, `snapshot_backfill_actors`.
2. reassign to `CreatingStreamingJobStatus::ConsumingLogStore {..}`.

src/meta/src/barrier/checkpoint/creating_job/status.rs
```diff
@@ -193,12 +209,20 @@ impl CreatingStreamingJobStatus {
                 let prev_epoch = barrier_info.prev_epoch();
                 {
                     assert!(barrier_info.kind.is_checkpoint());
-                    *self = CreatingStreamingJobStatus::Finishing(prev_epoch);
+                    let CreatingStreamingJobStatus::ConsumingLogStore { tracking_job, .. } =
+                        replace(self, CreatingStreamingJobStatus::PlaceHolder)
```
kwannoel, line 213: As above, I think we don't need a new placeholder state.

src/meta/src/barrier/checkpoint/creating_job/status.rs
wenym1: > Clone create_mview_tracker, backfill_epoch, snapshot_backfill_actors.

The purpose of introducing place holder is to avoid impl Clone for CreateMviewTracker, so as to
1. avoid unnecessary clone out of performance consideration.
2. refine the ownership management of the tracker, so that we won't implicitly duplicate the tracker for a job, and can ensure that there is only one tracker for a job unless we explicitly create another one.

kwannoel: Okay makes sense. I still think adding some transient state like `PlaceHolder` is kind of dirty. But I don't have better ideas at the moment.

wenym1: I think that's somehow a limitation of rust compiler. For a owned variable, it does allow consume the ownership of the variable and then reassign another value to it before accessing the variable, but this is not yet allowed for a mutable reference.

wenym1:
Thanks @copilot. Will apply your suggestion on format and grammar in the next PR, since I don't want to go through the CI and wait for it again😄

PR #23747, opened 2025-11-12
feat(stream,dashboard): aggregate tokio actor metrics by fragment

kwannoel:

Benchmark results with metrics always on: https://github.com/risingwavelabs/risingwave/pull/23747#issuecomment-3525563117.

## Motivation

Tokio metrics can give us information about:
1. Tokio Runtime overhead, for instance task contention. (scheduling metrics)
2. Fragments bottlenecked by CPU workloads. (poll duration metrics)

## Implementation

This PR improves the actor metrics collection in the streaming engine, so we can continually collect them:

1. Changed actor metrics from gauges to counters to better track cumulative values
2. Renamed `stream_actor_actor_execution_time` to `stream_actor_execution_duration` for clarity
3. Added fragment_id label to execution duration metric so we can aggregate by it and reduce metric cardinality.
5. Modified the metrics collection interval from 1 second to 15 seconds to reduce overhead
6. Updated the Grafana dashboard to use the new metric name

These changes provide more accurate and useful metrics for monitoring actor performance in the streaming engine, while also reducing the overhead of metrics collection.

## Checklist

## Documentation


kwannoel:
* **#23747** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23747?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23747?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/stream/src/executor/monitor/streaming_stats.rs
```diff
@@ -54,17 +54,17 @@ pub struct StreamingMetrics {
...
+    actor_scheduled_duration: RelabeledGuardedIntCounterVec,
+    actor_scheduled_cnt: RelabeledGuardedIntCounterVec,
+    actor_fast_poll_duration: RelabeledGuardedIntCounterVec,
+    actor_fast_poll_cnt: RelabeledGuardedIntCounterVec,
+    actor_slow_poll_duration: RelabeledGuardedIntCounterVec,
+    actor_slow_poll_cnt: RelabeledGuardedIntCounterVec,
+    actor_poll_duration: RelabeledGuardedIntCounterVec,
+    actor_poll_cnt: RelabeledGuardedIntCounterVec,
+    actor_idle_duration: RelabeledGuardedIntCounterVec,
+    actor_idle_cnt: RelabeledGuardedIntCounterVec,
```
kwannoel: We use `RelabeledGuardedIntCounterVec` to aggregate by fragment_id, when the metric level is lower than `debug`, e.g. `info`.

We also change these metrics from recording `Gauge` to `Counter`. This is because these metrics should only increase. Given that we will continually update these metrics, I think it's better to choose a more performant metric type.

src/stream/src/executor/monitor/streaming_stats.rs
```diff
@@ -364,85 +355,104 @@ impl StreamingMetrics {
         )
         .unwrap();
 
-        let actor_fast_poll_duration = register_guarded_gauge_vec_with_registry!(
+        let actor_execution_duration = register_guarded_int_counter_vec_with_registry!(
+            "stream_actor_execution_duration",
+            "Total execution duration (ns) of an actor",
+            &["actor_id", "fragment_id"],
```
zwang28: Relabeled metric takes first N label. So the order should be ["fragment_id", "actor_id"].

src/stream/src/executor/monitor/streaming_stats.rs
```diff
@@ -364,85 +355,104 @@ impl StreamingMetrics {
...
+            registry
+        )
+        .unwrap()
+        .relabel_debug_1(level);
+
+        let actor_fast_poll_duration = register_guarded_int_counter_vec_with_registry!(
             "stream_actor_fast_poll_duration",
             "tokio's metrics",
             &["actor_id"],
```
zwang28: Need add "fragment_id" as well?

kwannoel:
```suggestion
            &["actor_id", "fragment_id"],
```

src/stream/src/executor/monitor/streaming_stats.rs
```diff
@@ -364,85 +355,104 @@ impl StreamingMetrics {
...
         )
-        .unwrap();
+        .unwrap()
+        .relabel_debug_1(level);
 
-        let actor_fast_poll_cnt = register_guarded_int_gauge_vec_with_registry!(
+        let actor_fast_poll_cnt = register_guarded_int_counter_vec_with_registry!(
             "stream_actor_fast_poll_cnt",
             "tokio's metrics",
             &["actor_id"],
```
kwannoel:
```suggestion
            &["actor_id", "fragment_id"],
```

src/stream/src/executor/monitor/streaming_stats.rs
```diff
@@ -364,85 +355,104 @@ impl StreamingMetrics {
...
         )
-        .unwrap();
+        .unwrap()
+        .relabel_debug_1(level);
 
-        let actor_slow_poll_duration = register_guarded_gauge_vec_with_registry!(
+        let actor_slow_poll_duration = register_guarded_int_counter_vec_with_registry!(
             "stream_actor_slow_poll_duration",
             "tokio's metrics",
             &["actor_id"],
```
kwannoel:
```suggestion
            &["actor_id", "fragment_id"],
```

src/stream/src/executor/monitor/streaming_stats.rs
```diff
@@ -364,85 +355,104 @@ impl StreamingMetrics {
...
         )
-        .unwrap();
+        .unwrap()
+        .relabel_debug_1(level);
 
-        let actor_poll_duration = register_guarded_gauge_vec_with_registry!(
+        let actor_poll_duration = register_guarded_int_counter_vec_with_registry!(
             "stream_actor_poll_duration",
             "tokio's metrics",
             &["actor_id"],
```
kwannoel:
```suggestion
            &["actor_id", "fragment_id"],
```

src/stream/src/executor/monitor/streaming_stats.rs
```diff
@@ -364,85 +355,104 @@ impl StreamingMetrics {
...
         )
-        .unwrap();
+        .unwrap()
+        .relabel_debug_1(level);
 
-        let actor_poll_cnt = register_guarded_int_gauge_vec_with_registry!(
+        let actor_poll_cnt = register_guarded_int_counter_vec_with_registry!(
             "stream_actor_poll_cnt",
             "tokio's metrics",
             &["actor_id"],
```
kwannoel:
```suggestion
            &["actor_id", "fragment_id"],
```

src/stream/src/executor/monitor/streaming_stats.rs
```diff
@@ -364,85 +355,104 @@ impl StreamingMetrics {
...
         )
-        .unwrap();
+        .unwrap()
+        .relabel_debug_1(level);
 
-        let actor_scheduled_cnt = register_guarded_int_gauge_vec_with_registry!(
+        let actor_scheduled_cnt = register_guarded_int_counter_vec_with_registry!(
             "stream_actor_scheduled_cnt",
             "tokio's metrics",
             &["actor_id"],
```
kwannoel:
```suggestion
            &["actor_id", "fragment_id"],
```

src/stream/src/executor/monitor/streaming_stats.rs
```diff
@@ -364,85 +355,104 @@ impl StreamingMetrics {
...
         )
-        .unwrap();
+        .unwrap()
+        .relabel_debug_1(level);
 
-        let actor_idle_cnt = register_guarded_int_gauge_vec_with_registry!(
+        let actor_idle_cnt = register_guarded_int_counter_vec_with_registry!(
             "stream_actor_idle_cnt",
             "tokio's metrics",
             &["actor_id"],
```
kwannoel:
```suggestion
            &["actor_id", "fragment_id"],
```

src/stream/src/executor/monitor/streaming_stats.rs
```diff
@@ -364,85 +355,104 @@ impl StreamingMetrics {
...
         )
-        .unwrap();
+        .unwrap()
+        .relabel_debug_1(level);
 
-        let actor_idle_duration = register_guarded_gauge_vec_with_registry!(
+        let actor_idle_duration = register_guarded_int_counter_vec_with_registry!(
             "stream_actor_idle_duration",
             "tokio's metrics",
             &["actor_id"],
```
kwannoel:
```suggestion
            &["actor_id", "fragment_id"],
```

src/stream/src/executor/monitor/streaming_stats.rs
```diff
@@ -364,85 +355,104 @@ impl StreamingMetrics {
...
         )
-        .unwrap();
+        .unwrap()
+        .relabel_debug_1(level);
 
-        let actor_scheduled_duration = register_guarded_gauge_vec_with_registry!(
+        let actor_scheduled_duration = register_guarded_int_counter_vec_with_registry!(
             "stream_actor_scheduled_duration",
             "tokio's metrics",
             &["actor_id"],
```
kwannoel:
```suggestion
            &["actor_id", "fragment_id"],
```

src/stream/src/executor/monitor/streaming_stats.rs
kwannoel: Hmm, I followed the convention of other metrics in this file, where actor id is the first label. Let me confirm it again.

zwang28: Ah I'm wrong.

src/common/src/config/mod.rs
```diff
@@ -342,7 +342,7 @@ pub mod default {
         }
 
         pub fn enable_actor_tokio_metrics() -> bool {
-            false
+            true
```
zwang28, line 349: Since the overhead hasn't been tested, I'm fine with keeping it set to false for now.

kwannoel:
Will add a performance test result later.

src/common/src/config/mod.rs
kwannoel: Set to false.

kwannoel:
https://buildkite.com/risingwavelabs/docker/builds/17878 building image now.

Edit: Building another image also with metrics always enabled to benchmark performance: https://buildkite.com/risingwavelabs/docker/builds/17880

kwannoel:
Benchmarking: https://buildkite.com/risingwave-test/nexmark-benchmark/builds/6098

kwannoel:
Benchmark results look good when the metrics are always enabled: http://metabase.risingwave-cloud.xyz/dashboard/1405-nexmark-8c32g-1cn-rw-source-throughput?start_date=2024-12-01

~ same as previous results.

Query: throughput
q0: 1.3M
q4: 340K
q15: 486K
q20: 440K
q22: 1.3M
q101: 460K

src/common/src/config/mod.rs
kwannoel: Leaving it always enabled has minimal impact to the cluster performance from our nexmark benchmark. https://github.com/risingwavelabs/risingwave/pull/23747#issuecomment-3525563117

grafana/dashboard/dev/streaming_actors_tokio.py
```diff
@@ -9,23 +9,23 @@ def _(outer_panels: Panels):
...
-                panels.timeseries_actor_latency_small(
-                    "Tokio: Actor Fast Poll Time",
+                panels.timeseries_percentage(
+                    "Tokio: Actor Fast Poll Rate",
                     "",
                     [
                         panels.target(
-                            f"rate({metric('stream_actor_fast_poll_duration')}[$__rate_interval]) > 0",
-                            "{{actor_id}}",
+                            f"rate({metric('stream_actor_fast_poll_duration')}[$__rate_interval]) / 1000000000",
```
kwannoel: We use an int counter instead for better performance, and consistency with other latency measures (e.g. backpressure metrics). We increment in `nanoseconds` to maintain precision, so we need to normalize by 1e9 here.

src/stream/src/task/actor_manager.rs
```diff
@@ -487,53 +487,49 @@ impl StreamActorManager {
...
             let actor_monitor_task = self.runtime.spawn(async move {
-                let metrics = streaming_metrics.new_actor_metrics(actor_id);
-                loop {
-                    let task_metrics = monitor.cumulative();
+                let metrics = streaming_metrics.new_actor_metrics(actor_id, fragment_id);
+                for task_metrics in monitor.intervals() {
                     metrics
-                        .actor_execution_time
-                        .set(task_metrics.total_poll_duration.as_secs_f64());
+                        .actor_execution_duration
```
zwang28: actor_execution_duration is duplicate of actor_poll_duration

BugenZhao: +1

src/stream/src/task/actor_manager.rs
```diff
@@ -487,53 +487,49 @@ impl StreamActorManager {
...
                     metrics
                         .actor_scheduled_duration
-                        .set(task_metrics.total_scheduled_duration.as_secs_f64());
+                        .inc_by(task_metrics.total_scheduled_duration.as_nanos() as u64);
                     metrics
                         .actor_scheduled_cnt
-                        .set(task_metrics.total_scheduled_count as i64);
-                    tokio::time::sleep(Duration::from_secs(1)).await;
+                        .inc_by(task_metrics.total_scheduled_count);
+                    tokio::time::sleep(Duration::from_secs(15)).await; // Our scraping interval is at 15s.
```
BugenZhao: Shall we use `tokio::time::interval` instead?

src/stream/src/executor/monitor/streaming_stats.rs
```diff
@@ -364,85 +355,104 @@ impl StreamingMetrics {
...
+            registry
+        )
+        .unwrap()
+        .relabel_debug_1(level);
+
+        let actor_fast_poll_duration = register_guarded_int_counter_vec_with_registry!(
             "stream_actor_fast_poll_duration",
             "tokio's metrics",
-            &["actor_id"],
+            &["actor_id", "fragment_id"],
```
BugenZhao: nit: it confuses me a bit when I realized we relabel from the left hand side instead of the right. 🤔

BugenZhao (approved):
LGTM!

src/stream/src/task/actor_manager.rs
kwannoel: What's the difference? Seems to add more LoC if we iterate over `interval.tick().await`.

BugenZhao: > The difference between `interval` and [`sleep`](https://docs.rs/tokio/latest/tokio/time/fn.sleep.html "fn tokio::time::sleep") is that an [`Interval`](https://docs.rs/tokio/latest/tokio/time/struct.Interval.html "struct tokio::time::Interval") measures the time since the last tick, which means that [`.tick().await`](https://docs.rs/tokio/latest/tokio/time/struct.Interval.html#method.tick "method tokio::time::Interval::tick") may wait for a shorter time than the duration specified for the interval if some time has passed between calls to [`.tick().await`](https://docs.rs/tokio/latest/tokio/time/struct.Interval.html#method.tick "method tokio::time::Interval::tick").
> 
> If the tick in the example below was replaced with [`sleep`](https://docs.rs/tokio/latest/tokio/time/fn.sleep.html "fn tokio::time::sleep"), the task would only be executed once every three seconds, and not every two seconds.
> 
https://docs.rs/tokio/latest/tokio/time/fn.interval.html

src/stream/src/task/actor_manager.rs
kwannoel: Fixed https://github.com/risingwavelabs/risingwave/pull/23747/commits/da37010eed1f80f5def734ff0bd16a84b4062343

src/stream/src/task/actor_manager.rs
kwannoel: Fixed: https://github.com/risingwavelabs/risingwave/pull/23747/commits/32f0eb3f95a60edffb912153bd5d6cf3c85daf92

kwannoel:
I made the following changes:
- Normalized the actor metrics by actor count (0- ~100%).
- Improved the title for the panels.
- Excluded count metrics by default, I think they are less useful and just add noise.
- Removed slow/fast poll, keep only total poll.

kwannoel:
Screenshot 2025-12-03 at 1 28 23 PM

Here's what the metrics look like at the latest. I'm going to open a PR later to also add these metrics at the stream job level.

PR #23752, opened 2025-11-12
feat(meta): add `config_override` in stream context and meta store

BugenZhao:

Introduces a new `config_override` column to the `streaming_job` table of meta store. Adds it into `StreamContext` so that it can be

- constructed by frontend when building the stream graph
- handled by meta ddl service to persist it into the `streaming_job` table
- passed to compute node to build actor with

Previously, the `timezone` is the only field in `StreamContext`. So in many code paths we simply passed the bare `timezone` field instead of the structured `StreamContext`. This PR has refactored that so that it can be easier to introduce extra fields in the future.

There's no actual behavior change in this PR, as the `config_override` field originated from the frontend fragmenter, is currently set to a placeholder value "empty string".

## Checklist

## Documentation

<details>

</details>


BugenZhao:

- #23752 <-- you are here 🤪
    - #23735
        - #23768

shanicky (approved):

yezizp2012 (approved):

PR #23748, opened 2025-11-12
feat(frontend): add handler for `ALTER .. [SET | RESET] CONFIG`

BugenZhao:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## What's changed and what's your intention?

Follow up of #23742, this PR implements the handler functions in frontend. The actual implementation on the meta node remains `TODO`.

Extracted the logic of extracting streaming job id from `ALTER` into a separate function to be reused by all sort of `ALTER` commands that's applicable for any kind of streaming jobs.

## Checklist

- [x] I have written necessary rustdoc comments.
- [ ]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  I have checked the [Release Timeline](https://github.com/risingwavelabs/rw-commits-history/blob/main/release_timeline.md) and [Currently Supported Versions](https://docs.risingwave.com/changelog/release-support-policy#support-end-dates-for-recent-releases) to determine which release branches I need to cherry-pick this PR into. 

## Documentation

- [ ]  My PR needs documentation updates. 

<details>
<summary><b>Release note</b></summary>

</details>

PR #23768, opened 2025-11-14
feat: migrate `JoinEncodingType` from session config to per-job config override

BugenZhao:

Based on https://github.com/risingwavelabs/risingwave/pull/23735 and following the [Design RFC](https://www.notion.so/risingwave-labs/mini-RFC-Per-Job-Configuration-2a8f0d69cb768032b2acf1bb9bfee979?source=copy_link), this PR demonstrates the initial config override by migrating the `streaming_join_encoding: JoinEncodingType` into per-job config override.

The main idea is that:

Previously, when building prost `NodeBody`, we read the value of `streaming_join_encoding` from session config and persist it into the `NodeBody::Join(JoinNode { .. })` for later being read by the executor builder. There's no way to update it without rebuilding the entire job.

With the new design:

- The session config is still there, and the user-facing behavior remains the same, i.e., setting the config value with affect the `Join` of new streaming jobs created in the same session.
- However, we've marked the field `join_encoding_type` in `JoinNode` deprecated and won't persist it anymore. Instead, we dump it from the session config to the "initial config override" for the new job. When the join executors of this job is being built, we will resolve the value from the overridden per-job config.
- As a result, a new TOML configuration entry with the same name has been introduced to enable global settings.
  * if the job is created without explicitly setting the session config `join_encoding_type`, the value from global configuration file will be used
  * after jobs are created
    - for jobs using the value from global configuration: users can modify this global config to tweak the behavior of existing jobs
    - for jobs using the value from per-job config override: users can modify the config independently for specific jobs via `ALTER .. [SET | RESET] CONFIG` https://github.com/risingwavelabs/risingwave/pull/23748

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/error/src/anyhow.rs
+++ b/src/error/src/anyhow.rs
@@ -142,7 +142,7 @@ macro_rules! def_anyhow_newtype {
                         return stringify!($from);
                     }
                 )*
-                return "connector_error";
+                return stringify!($name);
             }
         }
 


BugenZhao:

- #23752
    - #23735
        - #23768 <-- you are here 🤪
            - #23792
                - #23794

PR #23769, opened 2025-11-14
fix(cdc): handle numeric identifiers in connector properties when alter source

wcy-fdu:

Fail to alter source if the definition of source is
```
CREATE SOURCE xxx WITH (
  connector='sqlserver-cdc',
  debezium.column.truncate.to."2000000".chars = 'xxx'
);
```
Will get following error:
```
 Invalid Parameter Value: Invalid parameter: sql parser error: Parsing Error: ContextError { context: [], cause: Some(StrError("expected =, found: .2000000")) }
```

### Root Cause
When ALTER SOURCE rebuilds the SQL definition from catalog:
1. [real_value() strips quotes from identifiers during storage](https://github.com/risingwavelabs/risingwave/blob/0c2e34a2879dd1a1fcc172cf20751be41653eb0f/src/sqlparser/src/ast/mod.rs#L167-L172)
2. SqlOption::try_from() reconstructs SQL by concatenating strings
3. Tokenizer misinterprets 2000000 as a Number token instead of an identifier
4. Parser fails when it expects = but sees .2000000

The issue is that Ident::from_real_value() didn't recognize that identifiers starting with digits require quotes per SQL standard. This pr enhanced Ident::from_real_value() to automatically quote identifiers that start with digits.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/sqlparser/src/ast/mod.rs
+++ b/src/sqlparser/src/ast/mod.rs
@@ -185,7 +185,10 @@ impl Ident {
     pub fn from_real_value(value: &str) -> Self {
         let needs_quotes = value
             .chars()
-            .any(|c| !matches!(c, 'a'..='z' | '0'..='9' | '_'));
+            .any(|c| !matches!(c, 'a'..='z' | '0'..='9' | '_'))
+            // Also need quotes if the identifier starts with a digit, as it would be
+            // tokenized as a number instead of an identifier (e.g., "2000000")
+            || value.chars().next().is_some_and(|c| c.is_ascii_digit());
 
         if needs_quotes {
             Self::with_quote_unchecked('"', value.replace('"', "\"\""))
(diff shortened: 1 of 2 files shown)


src/sqlparser/src/ast/mod.rs
```diff
@@ -176,7 +185,10 @@ impl Ident {
     pub fn from_real_value(value: &str) -> Self {
         let needs_quotes = value
             .chars()
-            .any(|c| !matches!(c, 'a'..='z' | '0'..='9' | '_'));
+            .any(|c| !matches!(c, 'a'..='z' | '0'..='9' | '_'))
+            // Also need quotes if the identifier starts with a digit, as it would be
+            // tokenized as a number instead of an identifier (e.g., "2000000")
+            || value.chars().next().is_some_and(|c| c.is_ascii_digit());
```
xiangjinwu, line 191: Reference: https://github.com/postgres/postgres/blob/REL_18_0/src/backend/utils/adt/ruleutils.c#L13023

PR #23754, opened 2025-11-12
refactor(meta): refine inflight info ownership and avoid unnecessary clone

wenym1:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## What's changed and what's your intention?

Previously we impl `Clone` for `InflightXxxInfo`, so that between `pre_apply` and `post_apply` a command, we can clone the `InflightDatabaseInfo` to take a snapshot of it and pass the snapshot for later handling. However, this intermediate snapshot is only used shortly later, and then gets dropped, and therefore in this PR, we can actually generate the information used in later handling between `pre_apply` and `post_apply` instead of cloning it to take a snapshot. And then we removed the derived `Clone` implementation on `InflightXxxInfo`, so that we can avoid unnecessary large clone on the snapshot, and in the meanwhile the ownership of each inflight actor and fragment get clearer. The naming `inflight_graph_info` is replaced with `database_info` to avoid confusion.

## Checklist

- [ ] I have written necessary rustdoc comments.
- [ ]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  I have checked the [Release Timeline](https://github.com/risingwavelabs/rw-commits-history/blob/main/release_timeline.md) and [Currently Supported Versions](https://docs.risingwave.com/changelog/release-support-policy#support-end-dates-for-recent-releases) to determine which release branches I need to cherry-pick this PR into. 

## Documentation

- [ ]  My PR needs documentation updates. 

<details>
<summary><b>Release note</b></summary>

</details>

PR #23795, opened 2025-11-18
refactor(meta): separate per job cdc backfill tracker

wenym1:

Previously we have a globally shared cdc progress tracker. In this PR, we will refactor to split the tracking of the cdc progress to each job in the `InflightStreamingJob`, just like what we do in https://github.com/risingwavelabs/risingwave/pull/23744 for create mview tracker.

Major changes in this PR
* cdc progress are reloaded in every recovery, instead of storing in the shared tracker to cross multiple recoveries.
* the cdc progresses are maintained in separately in each `InflightStreamingJob` in global barrier worker.
* for ongoing parallel cdc backfill, we also store the physical splits in `InflightStreamingJob`, and we always assign splits to actor inside global barrier worker. Outside global barrier worker:
  *  for a newly create streaming job, we only include the splits instead of split assignments in the create streaming job command
  * for reschedule and replace table, we won't include any information of cdc splits in the command. Instead we reassign the splits in global barrier worker when we handle the reschedule/replace table command.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/meta/src/barrier/cdc_progress.rs
+++ b/src/meta/src/barrier/cdc_progress.rs
@@ -166,6 +166,8 @@ impl CdcTableBackfillTracker {
         self.next_generation += 1;
         let splits = match &mut self.status {
             CdcBackfillStatus::Backfilling(progress) => {
+                progress.split_backfilled_count = 0;
+                progress.split_completed_count = 0;
                 progress.split_assignment_generation = generation;
                 progress.splits.as_slice()
             }
@@ -324,5 +326,6 @@ mod test {
         };
         assert_eq!(progress.splits.len() as u64, split_count);
         assert_eq!(progress.split_completed_count, 0);
+        assert_eq!(progress.split_backfilled_count, 0);
     }
 }


wenym1:
* **#23795** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23795?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23795?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#23754** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23754?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#23744** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23744?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/meta/src/barrier/checkpoint/creating_job/mod.rs
```diff
@@ -140,6 +139,11 @@ impl CreatingStreamingJobControl {
             .flat_map(build_actor_connector_splits)
             .collect();
 
+        assert!(
```
zwang28, line 142: Will this line of code be executed in this scenario?
SET streaming_use_snapshot_backfill to TRUE;
CREATE TABLE cdc_table (...) FROM cdc_source TABLE '...';

src/meta/src/barrier/checkpoint/state.rs
```diff
@@ -229,6 +230,7 @@ impl BarrierWorkerState {
                     fragment_infos: job_fragments,
                     subscribers: Default::default(), // no initial subscribers for newly created snapshot backfill
                     status: CreateStreamingJobStatus::Created,
+                    cdc_table_backfill_tracker: None, // no cdc table backfill for snapshot backfill
```
zwang28, line 233: ditto

src/meta/src/barrier/checkpoint/creating_job/mod.rs
wenym1: No, because snapshot backfill only works for backfilling from RW tables/MVs.

PR #23832, opened 2025-11-25
test(ci): speed up source tests

tabVersion:

as title

* concurrent install dep
* use more retry instead of sleep for checking result

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/e2e_test/source_inline/cdc/mongodb/mongodb_basic.slt
+++ b/e2e_test/source_inline/cdc/mongodb/mongodb_basic.slt
@@ -97,7 +97,7 @@ select _id, database_name, collection_name from all_types_extjson;
 statement ok
 insert into phases values ('03_ingested', now());
 
-query TTT
+query TTT retry 5 backoff 1s
 select * from all_types_access order by _id, field;
 ----
 {"$oid": "6911a33502454e0f774f8803"} f01a_double    2.5
@@ -157,7 +157,7 @@ from all_types_extjson order by commit_ts;
 {"$oid": "6911a33502454e0f774f8803"} test all_types f
 {"$oid": "6911a33502454e0f774f8804"} test all_types t
 
-query TTT
+query TTT retry 3 backoff 1s
 select * from all_types_access order by _id, field;
 ----
 {"$oid": "6911a33502454e0f774f8803"} f01a_double    2.5
--- a/e2e_test/source_inline/kafka/handling_mode.slt
+++ b/e2e_test/source_inline/kafka/handling_mode.slt
@@ -76,9 +76,7 @@ with (
   topic = 'json_timestamptz_handling_mode')
 format debezium encode json (timestamptz.handling.mode = 'milli');
 
-sleep 2s
-
-query TT retry 3 backoff 5s
+query TT retry 6 backoff 3s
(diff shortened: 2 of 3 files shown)


BugenZhao:
> * concurrent install dep

Personally I don't think this is a good practice. Instead, would you consider

- provisioning directly in the CI docker image, instead of installing deps every time when running tests
- cache installed deps for package manager including `pip`, just like what we did for `cargo` and `maven`
  https://github.com/risingwavelabs/risingwave/blob/5cd9d03cc3067c9b069491d36a442ee751a5a856/ci/workflows/pull-request.yml#L11-L39

tabVersion:
> provisioning directly in the CI docker image, instead of installing deps every time when running tests

Good point, now I update CI image and go with this approach. Later we can have a workflow to update deps in the pip requirements.

tabVersion:
found the new image may lead to more retries in SLT 🤡

ci/requirements.txt
BugenZhao, line 1: Can we reuse `e2e_test/requirements.txt` instead of duplicating one?

ci/Dockerfile
```diff
@@ -53,6 +53,25 @@ RUN curl -sSL "https://github.com/bufbuild/buf/releases/download/v1.29.0/buf-$(u
...
+
+# Install mongosh
+RUN wget --no-verbose https://repo.mongodb.org/apt/ubuntu/dists/noble/mongodb-org/8.0/multiverse/binary-amd64/mongodb-mongosh_2.5.8_amd64.deb && \
+    dpkg -i mongodb-mongosh_2.5.8_amd64.deb && \
+    rm mongodb-mongosh_2.5.8_amd64.deb
+
+# Copy and install e2e test Python requirements (merged from all sub-requirements)
+COPY requirements.txt /tmp/requirements.txt
+RUN pip3 install --break-system-packages -r /tmp/requirements.txt && \
+    rm /tmp/requirements.txt
```
BugenZhao, line 73: By installing python packages during CI image build, it means that updating deps requires a rebuild of CI image. I'm not sure if this is a good DX and can be confusing to others.

I would suggest

- either keeping the `pip install` line in CI images, then in most time it's no-op, but we still have the chance to incrementally install new packages in case there's any
- or avoid vendoring them in image, instead, introduce a pip cache

e2e_test/source_inline/cdc/postgres/postgres_schema_check_and_schema_change.slt
```diff
@@ -218,12 +216,8 @@ psql -c "
     );
     "
 
-sleep 5s
-
-
-
 # Test schema change by adding a new column
-system ok
+system ok retry 5 backoff 1s
```
BugenZhao, line 220: I suppose you should use `sleep_secs + 1` as the retry times here?

MrCroxx:
If modified this way, will the test pass have false positives?

For example:

Expectation: Atomic success, no intermediate state  
After modification: Intermediate states will trigger retry, and success is achieved after results are consistent.

BugenZhao:
> If modified this way, will the test pass have false positives?

This is a good point. But I think there's no way we are able to assert "something should never happen". In other words, it's not possible to check "atomic success" in your example. 😕

PR #23836, opened 2025-11-25
fix: update Minio client initialization to use keyword arguments

tabVersion:

seem MinIO has breaking changes recently

got the error on main-cron

```
Traceback (most recent call last):
File "/risingwave/e2e_test/s3/file_source.py", line 300, in <module>
client = Minio(
^^^^^^
TypeError: Minio.__init__() takes 1 positional argument but 4 positional arguments (and 1 keyword-only argument) were given
```

https://buildkite.com/risingwavelabs/main-cron/builds/7727#_

switch to all keyword arguments approach from their example https://pypi.org/project/minio/

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/e2e_test/s3/file_sink.py
+++ b/e2e_test/s3/file_sink.py
@@ -488,7 +488,7 @@ def test_file_sink_batching():
         secret_key="hummockadmin",
         secure=False,
     )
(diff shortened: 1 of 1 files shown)


wcy-fdu (approved):
LGTM if all tests pass.

PR #23837, opened 2025-11-25
fix(sink): fix bigquery sink auto_create

xxhZs:

We fixed this issue because BigQuery has changed the error type.

sql： 

> CREATE SINK bigquery_noreaga_commerce_dispute_funds_movements
> FROM dispute_funds_movements 
> WITH (
>     connector = 'bigquery',
>     type = 'upsert',
>     primary_key = 'id',
>     bigquery.project = 'peek-prod',
>     bigquery.dataset = 'risingwave_noreaga_commerce',
>     bigquery.table = 'dispute_funds_movements',
>     bigquery.credentials = secret gcp_peek_prod_service_account_json,
>     auto_create = 'true'
> );

error:

> ERROR: Failed to run the query Caused by these errors (recent errors listed first): 1: gRPC request to meta service (call `/ddl_service.DdlService/CreateSink`) failed: Internal error 2: failed to validate sink 3: BigQuery error 4: Response error (error: ResponseError { error: NestedResponseError { code: 404, errors: [{"domain": "global", "reason": "notFound", "message": "Not found: Table peek-prod:risingwave_noreaga_commerce.dispute_funds_movements"}], message: "Not found: Table peek-prod:risingwave_noreaga_commerce.dispute_funds_movements", status: "NOT_FOUND" } }) (SQLSTATE XX000)

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/src/sink/big_query.rs
+++ b/src/connector/src/sink/big_query.rs
@@ -539,7 +539,7 @@ impl Sink for BigQuerySink {
                 .get(project_id, dataset_id, table_id, None)
                 .await
             {
-                Err(BQError::ResponseError{..}) => {
+                Err(BQError::ResponseError { error }) if error.error.code == 404 => {
                     // early return: no need to query schema to check column and type
                     return self
                         .create_table(


hzxa21:
How is this PR tested? Let's also include the test steps if it is manually tested so that we know the context in the future.

hzxa21:
> because BigQuery has changed the error type.

Is this behavior change relevant to a specific BigQuery server version or client sdk version? If it is the former, there can be compatibility issue, like the code works on specific version but doesn't work on others. 

I think it is better to check the 404 error code specifically instead of relying on the error type.

PR #23838, opened 2025-11-25
fix(streaming): fix locality backfill state -- write upstream chunk to the state

chenzl25:

- We need to write upstream chunk to the state table always for locality backfill.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/executor/locality_provider.rs
+++ b/src/stream/src/executor/locality_provider.rs
@@ -706,9 +706,7 @@ impl<S: StateStore> LocalityProviderExecutor<S> {
                     state_table.write_chunk(chunk);
                 }
 
-                let post_commit1 = state_table
-                    .commit(barrier.epoch)
-                    .await?;
+                let post_commit1 = state_table.commit(barrier.epoch).await?;
 
                 // Update progress with current epoch and snapshot read count
                 // Report both consumed rows and buffered rows separately for precise progress


BugenZhao:
🥵 Why didn't we encounter any issue before?

src/stream/src/executor/locality_provider.rs
```diff
@@ -740,7 +743,8 @@ impl<S: StateStore> LocalityProviderExecutor<S> {
                     .inc_by(cur_barrier_upstream_processed_rows);
 
                 yield Message::Barrier(barrier);
-                post_commit.post_yield_barrier(None).await?;
+                post_commit1.post_yield_barrier(None).await?;
```
wenym1, line 744: `commit_assert_no_update_vnode_bitmap` doesn't mean no-op, but means that there must no vnode update bitmap happen. 

Since here we always pass `None` to `post_yield_barrier`, we can assert that there is no update vnode bitmap, we therefore we can call `commit_assert_no_update_vnode_bitmap` directly.

hzxa21:
> 🥵 Why didn't we encounter any issue before?

Same question. How did we spot this issue?

wenym1:
> > 🥵 Why didn't we encounter any issue before?
> 
> Same question. How did we spot this issue?

By left-right brain battler in offline discussion 🥊

chenzl25:
> > > 🥵 Why didn't we encounter any issue before?
> > 
> > 
> > Same question. How did we spot this issue?
> 
> By left-right brain battler in offline discussion 🥊

By thinking about the improvement of how to avoid rebuilding the snapshot stream for each epoch 🥵

src/stream/src/executor/locality_provider.rs
chenzl25: cc @shanicky Do you have any idea about how to deal with the online scaling in the locality backfill operator?

chenzl25: It seems like we can't do online scaling during (locality) backfilling, because the CreateMvTracker is with static actors. For locality backfill, after backfilling, there is no need to read and write the state, so there is no need to handle the online scaling 🤔

hzxa21:
> > > 🥵 Why didn't we encounter any issue before?
> > 
> > 
> > Same question. How did we spot this issue?
> 
> By left-right brain battler in offline discussion 🥊

GAN lol

PR #23851, opened 2025-11-26
fix(doc): missing GNU Parallel installation

Standing-Man:

- Add missing GNU Parallel installation 

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -807,23 +807,6 @@ do
   echo
 done
 
-echo "install: $(tput setaf 4)GNU Parallel$(tput sgr0)"
-UNAME=$(uname | tr "[:lower:]" "[:upper:]" )
-case $UNAME in
-  DARWIN)
-    if ! command -v brew >/dev/null 2>&1 ; then
-      echo "Installing Homebrew (https://brew.sh/)"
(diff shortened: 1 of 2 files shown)


Standing-Man:
Hi @BugenZhao, please take a look when you have time.

BugenZhao:
Hi. I don't think we should handle installation of external tools requiring different package manager for different OS in RiseDev. Otherwise, should we also help to install `tmux`, `cmake`, or even `psql`, which can be much more complicated?

Standing-Man:
> Hi. I don't think we should handle installation of external tools requiring different package manager for different OS in RiseDev. Otherwise, should we also help to install `tmux`, `cmake`, or even `psql`, which can be much more complicated?

Should we add a command in `risedev` that set up the development environment for developers? 
Or just add the requirement to install `GNU Parallel` in the environment setup section of the `RisingWave Developer Guide`.

BugenZhao:
> Or just add the requirement to install `GNU Parallel` in the environment setup section of the `RisingWave Developer Guide`.

Sounds good. I would vote for this. I'm afraid automating env setup would lead to a maintenance burden.

Standing-Man:
> Sounds good. I would vote for this. I'm afraid automating env setup would lead to a maintenance burden.

Got it, i've added the requirement on `Developer Guide`.

PR #23858, opened 2025-11-26
refactor(meta): avoid self-generate command in barrier worker

wenym1:

Previously, we have two commands `StartFragmentBackfill` and `MergeSnapshotBackfillStreamingJobs` which are generated inside global barrier worker rather than scheduled from outside. We introduce these two commands because we want to fit in the current procedure of handling, where we split the handling of a command into several steps, and we pass the commands as input parameter of these steps. These steps includes applying changes to inflight info, updating new snapshot backfill jobs, generating mutation, and etc..

In this PR, we merge these separate steps into a single one, to inside the `apply_command` method. After we merge these steps, we don't need the two self-generated commands any more. Inside `apply_command`, in the scenario when we previously generated the `MergeSnapshotBackfillStreamingJobs`, instead of generating the command, we will directly apply the changes of snapshot backfill jobs and generate the mutation, and in the scenario when we previously generated `StartFragmentBackfill`, we will generate the mutation directly. Previously we had to generate the mutation via the `to_mutation` method of the command.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/meta/src/barrier/checkpoint/state.rs
+++ b/src/meta/src/barrier/checkpoint/state.rs
@@ -283,36 +283,38 @@ impl DatabaseCheckpointControl {
             Some(mutation) => Some(mutation),
             None => {
                 let mut subscription_info_to_drop = vec![];
-                for (&job_id, creating_job) in &mut self.creating_streaming_job_controls {
-                    if let Some(upstream_table_ids) = creating_job.should_merge_to_upstream() {
-                        subscription_info_to_drop.extend(upstream_table_ids.iter().map(
-                            |upstream_table_id| PbSubscriptionUpstreamInfo {
-                                subscriber_id: job_id.as_subscriber_id(),
-                                upstream_mv_table_id: *upstream_table_id,
-                            },
-                        ));
-                        for upstream_mv_table_id in upstream_table_ids {
-                            assert_matches!(
-                                self.database_info.unregister_subscriber(
-                                    upstream_mv_table_id.as_job_id(),
-                                    job_id.as_subscriber_id()
-                                ),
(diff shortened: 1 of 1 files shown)


wenym1:
* **#23858** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23858?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23858?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#24140** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24140?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/meta/src/barrier/checkpoint/control.rs
```diff
@@ -1027,10 +1008,11 @@ impl DatabaseCheckpointControl {
             return Ok(());
         }
 
-        let Some(barrier_info) =
```
wenym1, line 1034: review note: previously `next_barrier_info` return `None` when `database_info.is_empty()`, and otherwise `Some(barrier_info)`.

src/meta/src/barrier/checkpoint/control.rs
```diff
@@ -1039,146 +1021,42 @@ impl DatabaseCheckpointControl {
             return Ok(());
         };
 
-        let mut edges = self
```
wenym1, line 1046: Review note: this logic is moved to inside `apply_command`.

yezizp2012 (approved):
lgtm

PR #23859, opened 2025-11-26
fix(sink): use downstream pk to compact when stream key is its subset

### LINKED ISSUE #23857: sink-into-table does not correctly retract data
Found by @chenzl25:

```sql
create table t(ts timestamptz primary key, value int);

create materialized view s as select max(d), 114514 value from generate_series(
  '2025-11-20 17:36:00'::timestamptz,
  now(),
  interval '1 second'
) d;

create sink my_sink into t as select s.max as ts, s.value from s;
```

We got multiple `max(d) as ts` in the table.

```
dev=> select * from t;
            ts             | value
---------------------------+--------
 2025-11-26 06:14:11+00:00 | 114514
 2025-11-26 06:14:12+00:00 | 114514
 2025-11-26 06:14:13+00:00 | 114514
 2025-11-26 06:14:14+00:00 | 114514
(4 rows)
```

Cannot reproduce on release v2.6. Seems to be introduced by #23581.

BugenZhao:

Fix #23857.

#23581 exposes a hidden bug that, for the case when the stream key _matches_ the downstream pk by being a subset of it, we incorrectly use the stream key to compact the chunk, instead of using the downstream key.

This fix is necessary because

- For `SinkType::Retract`, the downstream will require the `U-/U+` pairs to have the same `downstream_pk` (https://github.com/risingwavelabs/risingwave/pull/23463). However, when `stream_key` is the subset of `downstream_pk`, by compacting the chunk with `stream_key`, we cannot enforce the key of `U-/U+` pairs to be the same.

- For `SinkType::Upsert`, we will produce upsert chunk when performing compaction. If a wrong key is specified, we may lose some necessary `Delete` records, leading to bugs like #23857.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/e2e_test/streaming/bug_fixes/issue_23857.slt
+++ b/e2e_test/streaming/bug_fixes/issue_23857.slt
@@ -7,6 +7,8 @@ create table t (k int primary key, v int);
 statement ok
 create table s (k int);
 
+# The stream key is empty ([]) as the output of simple agg.
+# The downstream pk is [k], following the table's definition.
 statement ok
 create sink sk into t as select max(k) as k, 114514 as v from s;
 


hzxa21 (approved):
Good find! LGTM

PR #23674, opened 2025-11-05
refactor(connector):  introduce a macro to easily gate sink impl behind a feature

BugenZhao:
Signed-off-by: Bugen Zhao <i@bugenzhao.com>I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## What's changed and what's your intention?

A follow-up of #21786.

Developers can now easily gate some sink implementation module behind a cargo feature with macro `feature_gated_sink_mod!`. This can help decrease the compilation time and the binary size if the specific feature is not necessary for development (or even distribution).

This PR also adds some more feature gates, which have a relatively clear code boundary, including:
- `sink-bigquery`
- `sink-google_pubsub`
- `sink-clickhouse`
- `sink-doris`
- `sink-starrocks`

## Checklist

- [x] I have written necessary rustdoc comments.
- [ ]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  I have checked the [Release Timeline](https://github.com/risingwavelabs/rw-commits-history/blob/main/release_timeline.md) and [Currently Supported Versions](https://docs.risingwave.com/changelog/release-support-policy#support-end-dates-for-recent-releases) to determine which release branches I need to cherry-pick this PR into. 

## Documentation

- [ ]  My PR needs documentation updates. 

<details>
<summary><b>Release note</b></summary>

</details>
Signed-off-by: Bugen Zhao <i@bugenzhao.com>Signed-off-by: Bugen Zhao <i@bugenzhao.com>

PR #23860, opened 2025-11-26
feat: use datafusion engine for Iceberg batch query

silver-ymz:

follow #23302

This PR adds a minimum support to use datafusion engine for Iceberg batch query.
- Support `IcebergScan`, `Projection`, `Join`, `Value`.
- Add session config `enable_datafusion_engine` to control whether to use datafusion engine

Example:
```sql
set enable_datafusion_engine = true;

create secret my_secret with (
  backend = 'meta'
) as 'hummockadmin';
create connection my_conn
with (
    type = 'iceberg',
    s3.region = 'us-west-2',
    catalog.type = 'storage',
);
set iceberg_engine_connection = 'public.my_conn';

create table t(id int primary key, name varchar) with(commit_checkpoint_interval = 1) engine = iceberg;
FLUSH;

select * from t;
----
1 xxx
```

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/datafusion/iceberg_executor.rs
+++ b/src/frontend/src/datafusion/iceberg_executor.rs
@@ -18,6 +18,7 @@ use std::sync::Arc;
 use datafusion::arrow::array::RecordBatch;
 use datafusion::arrow::compute::concat_batches;
 use datafusion::catalog::TableProvider;
+use datafusion::error::Result as DFResult;
 use datafusion::execution::{SendableRecordBatchStream, TaskContext};
 use datafusion::physical_expr::EquivalenceProperties;
 use datafusion::physical_plan::execution_plan::{Boundedness, EmissionType};
@@ -93,7 +94,11 @@ impl ExecutionPlan for IcebergScan {
         partition: usize,
         context: Arc<TaskContext>,
     ) -> datafusion_common::Result<SendableRecordBatchStream> {
-        assert!(partition == 0, "IcebergScan only supports single partition");
+        if partition != 0 {
+            return Err(DataFusionError::Internal(
+                "IcebergScan only supports single partition".to_owned(),
+            ));
+        }
 
         let chunk_size = context.session_config().batch_size();
         let stream = self.inner.clone().execute_inner(chunk_size);
@@ -111,11 +116,12 @@ impl IcebergScan {
         _projection: Option<&Vec<usize>>,
         _filters: &[Expr],
         _limit: Option<usize>,
-    ) -> Self {
-        assert!(
-            provider.iceberg_scan_type == IcebergScanType::DataScan,
-            "Only DataScan is supported currently"
-        );
+    ) -> DFResult<Self> {
+        if provider.iceberg_scan_type != IcebergScanType::DataScan {
+            return Err(DataFusionError::NotImplemented(
+                "Only DataScan is supported currently".to_owned(),
+            ));
+        }
 
         let plan_properties = PlanProperties::new(
             EquivalenceProperties::new(provider.schema()),
@@ -140,9 +146,9 @@ impl IcebergScan {
             need_file_path_and_pos: false,
             plan_properties,
         };
-        Self {
+        Ok(Self {
             inner: Arc::new(inner),
-        }
+        })
     }
 }
 
--- a/src/frontend/src/handler/query.rs
+++ b/src/frontend/src/handler/query.rs
@@ -59,12 +59,14 @@ pub enum BatchPlanChoice {
 }
 
 impl BatchPlanChoice {
-    pub fn unwrap_rw(self) -> BatchQueryPlanResult {
+    pub fn unwrap_rw(self) -> Result<BatchQueryPlanResult> {
         match self {
-            BatchPlanChoice::Rw(result) => result,
+            BatchPlanChoice::Rw(result) => Ok(result),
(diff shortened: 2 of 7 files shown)


src/frontend/src/datafusion/iceberg_executor.rs
```diff
@@ -0,0 +1,212 @@
...
+impl IcebergScanInner {
+    #[try_stream(ok = RecordBatch, error = DataFusionError)]
+    pub async fn execute_inner(self: Arc<Self>, chunk_size: usize) {
+        let mut buffer = RecordBatchBuffer::new(chunk_size);
+        let table = self
+            .iceberg_properties
+            .load_table()
+            .await
+            .map_err(to_datafusion_error)?;
+        let mut scan_builder = table.scan().with_batch_size(Some(chunk_size));
```
chenzl25, line 164: With the chunk size setting in the scan builder, do we need an additional RecordBatchBuffer?

src/frontend/src/datafusion/iceberg_executor.rs
```diff
@@ -0,0 +1,212 @@
...
+
+impl IcebergScanInner {
+    #[try_stream(ok = RecordBatch, error = DataFusionError)]
+    pub async fn execute_inner(self: Arc<Self>, chunk_size: usize) {
+        let mut buffer = RecordBatchBuffer::new(chunk_size);
+        let table = self
+            .iceberg_properties
+            .load_table()
+            .await
+            .map_err(to_datafusion_error)?;
```
chenzl25, line 163: I think we'd better using a fixed snapshot id in the executor instead of loading the latest iceberg table, because the latest iceberg table could be changed during our execution if others update it.

src/frontend/src/datafusion/iceberg_executor.rs
silver-ymz: This logic is copied from `iceberg-compaction`. My idea is as follows. The batch size setting in `iceberg::TableScan` is set for each `FileScanTask`. For cases with many small `FileScanTask`s, using an additional RecordBatchBuffer can provide benefits.

src/frontend/src/datafusion/iceberg_executor.rs
chenzl25: I will open a PR to add a snapshot id field to the LogicalIcebergScan

chenzl25: https://github.com/risingwavelabs/risingwave/pull/23882

src/frontend/src/datafusion/iceberg_executor.rs
chenzl25: I see.

src/frontend/src/datafusion/iceberg_executor.rs
chenzl25: The PR has been merged. Could you please use the snapshot id here?

src/frontend/src/optimizer/plan_visitor/datafusion_plan_converter.rs
```diff
@@ -0,0 +1,244 @@
...
+
+        Some(LogicalPlan::Join(Join {
+            left: Arc::new(left_plan),
+            right: Arc::new(right_plan),
+            on: join_on_exprs,
+            filter,
+            join_type: df_join_type,
+            join_constraint: JoinConstraint::On,
+            schema: Arc::new(join_schema),
+            null_equals_null: false,
```
chenzl25: The `null_equals_null` means the join eq condition must use `==` as the equal semantics, however, we have `is not distinct from` expr which could be used to treat null equals null. So we'd better take a look at which type of equal we use in `plan.on()`. Notice that RisingWave supports mixed-use of `==` and `is not distinct from` however, datafusion doesn't. We need to reject those mixed-use cases.

silver-ymz:
> I'm wondering if we can make it behind a cargo feature gate, similar to how we did for sink (https://github.com/risingwavelabs/risingwave/pull/21786, https://github.com/risingwavelabs/risingwave/pull/23674). In this way we can have a clearer scope and abstract boundary, and also speed up compilation if it's not necessary for some developers' local testing.

Done. `datafusion` has been set as an optional dependency. Developers need to enable it by `./risedev configure`.

src/frontend/src/optimizer/plan_visitor/datafusion_plan_converter.rs
silver-ymz: Actually, `JOIN` in previous commit doesn't work. The newest commit add supports to `JOIN` and `PROJECTION`. And this case has been handled.

src/frontend/src/handler/query.rs
```diff
@@ -48,21 +48,71 @@ use crate::scheduler::{
...
+    },
+}
+
+impl BatchPlanChoice {
+    pub fn unwrap_rw(self) -> BatchQueryPlanResult {
+        match self {
+            BatchPlanChoice::Rw(result) => result,
+            #[cfg(feature = "datafusion")]
+            BatchPlanChoice::Df { .. } => {
+                panic!("called `BatchPlanChoice::unwrap_rw()` on a `DataFusion` plan")
```
chenzl25: Shall we return an error instead of panic?

silver-ymz: updated in new commit

PR #23877, opened 2025-11-27
refactor(watermark): use clean_watermark_indices to replace clean_watermark_index_in_pk

yuhao-su:

Added support for multiple watermark columns for state cleaning by introducing a new `clean_watermark_indices` field in the `Table` protobuf message. This replaces the existing `clean_watermark_index_in_pk` field which only supports a single watermark column and is now deprecated.

The changes include:
- Added a new `clean_watermark_indices` field to the `Table` protobuf message that stores indices of watermark columns
- Deprecated the old `clean_watermark_index_in_pk` field but kept it for backward compatibility
- Added helper methods to get watermark indices with backward compatibility
- Updated the database schema with a new migration

TODO:
- Storage layer should start to use `get_clean_watermark_indices` to decide all the watermark columns can be used for cleaning.

## Checklist

## Documentation

### DIFF
--- a/src/prost/src/lib.rs
+++ b/src/prost/src/lib.rs
@@ -529,6 +529,50 @@ impl catalog::Sink {
     }
 }
 
+impl catalog::Table {
+    /// Get clean watermark column indices with backward compatibility.
+    /// Returns the new `clean_watermark_indices` if set, otherwise derives it from the old
+    /// `clean_watermark_index_in_pk` by converting PK index to column index.
+    #[expect(deprecated)]
+    pub fn get_clean_watermark_column_indices(&self) -> Vec<u32> {
+        if !self.clean_watermark_indices.is_empty() {
+            // New format: directly return clean_watermark_indices
+            self.clean_watermark_indices.clone()
+        } else if let Some(pk_idx) = self.clean_watermark_index_in_pk {
+            // Old format: convert PK index to column index
+            // The pk_idx is the position in the PK, we need to find the corresponding column index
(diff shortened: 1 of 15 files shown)


yuhao-su:
* **#23877** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23877?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/23877?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

zwang28:
We need `clean_watermark_column_id` for non-pk watermark, because add/drop column will change indice.

yuhao-su:
> clean_watermark_column_id

I agree it should be stored as id. The problem is that we are storing every in indices e.g. `value_indices`

src/prost/src/lib.rs
```diff
@@ -529,6 +529,50 @@ impl catalog::Sink {
...
+        } else {
+            vec![]
+        }
+    }
+
+    /// Convert clean watermark column indices to PK indices and return the minimum.
+    /// Returns None if no clean watermark is configured.
+    /// This is a backward-compatible method to replace the deprecated `clean_watermark_index_in_pk` field.
+    /// TODO: remove this method after totally deprecating `clean_watermark_index_in_pk`.
+    pub fn get_clean_watermark_index_in_pk_compat(&self) -> Option<usize> {
```
zwang28: Should we ban DROP COLUMN for watermark column? Otherwise the returned column might change.

PR #23884, opened 2025-11-27
feat: persist compactor iceberg flag across meta and diagnostics

Li0k:

## Summary
- add meta model schema/migration plus worker registration updates to persist compactor.is_iceberg flag
- propagate the proto/property usage so compute/compactor registration and diagnose output show the new bit

image

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/catalog/system_catalog/rw_catalog/rw_worker_nodes.rs
+++ b/src/frontend/src/catalog/system_catalog/rw_catalog/rw_worker_nodes.rs
@@ -39,6 +39,7 @@ struct RwWorkerNode {
     system_total_cpu_cores: Option<i64>,
     started_at: Option<Timestamptz>,
     resource_group: Option<String>,
+    is_iceberg_compactor: Option<bool>,
 }
 
 #[system_catalog(table, "rw_catalog.rw_worker_nodes")]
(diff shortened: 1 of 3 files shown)


src/meta/src/controller/cluster.rs
```diff
@@ -725,18 +726,49 @@ impl ClusterControllerInner {
...
+                        Set(Some(add_property.internal_rpc_host_addr));
+
+                    WorkerProperty::update(property).exec(&txn).await?;
+                } else {
+                    let property = worker_property::ActiveModel {
+                        worker_id: Set(worker.worker_id),
+                        parallelism: Set(add_property
+                            .parallelism
+                            .try_into()
+                            .expect("invalid parallelism")),
```
hzxa21, line 750: `parallelism` is not used by compactor. Should we let compactor report its parallelism as well? It sounds like a useful information.

src/meta/src/manager/diagnose.rs
```diff
@@ -175,14 +176,35 @@ impl DiagnoseCommand {
...
+                        worker_node.property.as_ref().map(|p| p.is_serving),
+                    )
+                } else {
+                    (None, None)
+                }
+            };
+            try_add_cell(&mut row, is_streaming);
+            try_add_cell(&mut row, is_serving);
+            // is_iceberg_compactor is only meaningful for Compactor worker type
+            let is_iceberg_compactor = {
```
hzxa21, line 195: should we change `rw_worker_nodes` table and show the `is_iceberg_compactor` flag?

https://github.com/risingwavelabs/risingwave/blob/6ebd5e5671551c9f3cbf2f20b3838cd5a976fda5/src/frontend/src/catalog/system_catalog/rw_catalog/rw_worker_nodes.rs#L44

src/meta/src/controller/cluster.rs
Li0k: a brilliant idea

PR #23885, opened 2025-11-27
feat: Expose license status, limits, and features in diagnose output

Li0k:

as title 

image

## Checklist

## Documentation

<details>

</details>


tabVersion (approved):

src/meta/src/manager/diagnose.rs
```diff
@@ -777,6 +780,98 @@ impl DiagnoseCommand {
...
+            Ok(license) => {
+                let fmt_option = |value: Option<u64>| match value {
+                    Some(v) => v.to_string(),
+                    None => "unlimited".to_owned(),
+                };
+
+                let expires_at = if license.exp == u64::MAX {
+                    "never".to_owned()
+                } else if license.exp > i64::MAX as u64 {
+                    format!("invalid ({})", license.exp)
```
BugenZhao: We can also treat this case as "never".

src/meta/src/manager/diagnose.rs
```diff
@@ -777,6 +780,98 @@ impl DiagnoseCommand {
...
+                };
+
+                let mut row = Row::new();
+                row.add_cell("status".into());
+                row.add_cell("valid".into());
+                table.add_row(row);
+
+                let mut row = Row::new();
+                row.add_cell("tier".into());
+                row.add_cell(format!("{:?}", license.tier).into());
```
BugenZhao: Using the debug repr might be verbose. Given that we already have the "feature_summary" row, shall we add a `fn name` to `Tier` to only obtain the tier name, like

- free
- paid
- all
- `<custom name>`

PR #23886, opened 2025-11-27
feat(cdc): allow user alter `debezium.max.queue.size` to avoid JVM OOM

wcy-fdu:

As title, currently `debezium.max.queue.size` is set to 8192 if user do not config it when creating source, and can not be changed when running. This pr make if adjustable during system running, which can reduce the usage of the debezium portion of JVM memory. The statement is
```
ALTER SOURCE s CONNECTOR WITH (debezium.max.queue.size = '999');
```

The next pr will do some similar work on another config `debezium.max.queue.size.in.bytes`.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/src/allow_alter_on_fly_fields.rs
+++ b/src/connector/src/allow_alter_on_fly_fields.rs
@@ -99,20 +99,32 @@ pub static SOURCE_ALLOW_ALTER_ON_FLY_FIELDS: LazyLock<HashMap<String, HashSet<St
     // CDC Properties - added for schema.change.failure.policy
     map.try_insert(
         std::any::type_name::<MysqlCdcProperties>().to_owned(),
-        ["cdc.source.wait.streaming.start.timeout".to_owned()].into_iter().collect(),
+        [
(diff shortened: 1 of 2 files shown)


src/connector/src/allow_alter_on_fly_fields.rs
```diff
@@ -99,20 +99,32 @@ pub static SOURCE_ALLOW_ALTER_ON_FLY_FIELDS: LazyLock<HashMap<String, HashSet<St
     // CDC Properties - added for schema.change.failure.policy
     map.try_insert(
         std::any::type_name::<MysqlCdcProperties>().to_owned(),
-        ["cdc.source.wait.streaming.start.timeout".to_owned()].into_iter().collect(),
+        [
+            "cdc.source.wait.streaming.start.timeout".to_owned(),
+            "debezium.max.queue.size".to_owned(),
```
zwang28, line 104: Declare a global constant string and use its reference.

wcy-fdu: No needed.

PR #23888, opened 2025-11-27
fix(optimizer): use shard_by_exact_key instead of shard_by_key for locality backfill

chenzl25:

- Added RequiredDist::ShardByExactKey to match the distribution key exactly the same. Though we can use `hash_shard` to achieve the same result but `shard_by_exact_key` can be satisfied by `UpstreamHashShard` as the input distribution.
- Use `shard_by_exact_key` instead of `shard_by_key`, because locality provider will change the `stream_key` to include locality columns.
        If we use `shard_by_key`, it is possible that the locality columns are (`a`, `b`), but input stream key is only (`a`).
        In this case, `shard_by_key` will only shuffle by `a`. once `b` is changed, we will meet  U- and U+ should have same stream key error.
        Though we can let locality provider stream key to include its distribution columns only to fix the error,
        using `shard_by_exact_key` for another reason that it can provide better locality by `shard_by_key`.
        For example, if locality columns are (`a`, `b`), and we use `shard_by_key` with key (`b`),
        then all data with same `a` but different b will be shuffled to different nodes, which hurts locality.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/optimizer/plan_node/logical_locality_provider.rs
+++ b/src/frontend/src/optimizer/plan_node/logical_locality_provider.rs
@@ -130,6 +130,10 @@ impl ToStream for LogicalLocalityProvider {
         // Use `shard_by_exact_key` instead of `shard_by_key`, because locality provider will change the `stream_key` to include locality columns.
         // If we use `shard_by_key`, it is possible that the locality columns are (`a`, `b`), but input stream key is only (`a`).
         // In this case, `shard_by_key` will only shuffle by `a`. once `b` is changed, we will meet  U- and U+ should have same stream key error.
+        // Though we can let locality provider stream key to include its distribution columns only to fix the error,
+        // using `shard_by_exact_key` for another reason that it can provide better locality by `shard_by_key`.
+        // For example, if locality columns are (`a`, `b`), and we use `shard_by_key` with key (`b`),
+        // then all data with same `a` but different b will be shuffled to different nodes, which hurts locality.
         let required_dist =
             RequiredDist::shard_by_exact_key(self.input().schema().len(), self.locality_columns());
         let input = required_dist.streaming_enforce_if_not_satisfies(input)?;


src/frontend/planner_test/tests/testdata/output/locality_backfill.yaml
```diff
@@ -134,3 +134,36 @@
...
+        , lead(v2, 2) over (partition by p1, p2 order by v1, v2) as out9
+    from t;
+  stream_plan: |-
+    StreamMaterialize { columns: [id, p1, p2, time, v1, v2, t._row_id(hidden), out5, out6, out7, out8, out9], stream_key: [p1, p2, t._row_id], pk_columns: [p1, p2, t._row_id], pk_conflict: NoCheck }
+    └─StreamOverWindow { window_functions: [first_value(t.v2) OVER(PARTITION BY t.p1, t.p2 ORDER BY t.v1 ASC, t.v2 ASC ROWS BETWEEN 2 FOLLOWING AND 2 FOLLOWING)] }
+      └─StreamLocalityProvider { locality_columns: [t.p1, t.p2] }
+        └─StreamExchange [no_shuffle] { dist: UpstreamHashShard(t.p1, t.p2) }
+          └─StreamOverWindow { window_functions: [first_value(t.v1) OVER(PARTITION BY t.p1, t.p2 ORDER BY t.time ASC, t.id ASC ROWS BETWEEN 1 PRECEDING AND 1 PRECEDING), first_value(t.v2) OVER(PARTITION BY t.p1, t.p2 ORDER BY t.time ASC, t.id ASC ROWS BETWEEN 1 FOLLOWING AND 1 FOLLOWING)] }
+            └─StreamLocalityProvider { locality_columns: [t.p1, t.p2] }
+              └─StreamExchange [no_shuffle] { dist: UpstreamHashShard(t.p1, t.p2) }
```
BugenZhao, line 162: Is `UpstreamHashShard` due to the output of `StreamLocalityProvider`?

chenzl25: Yes. `StreamLocalityProvider` will turn `HashShare` to  `UpstreamHashShard`

PR #23897, opened 2025-11-28
feat(cdc): introduce configurable and adjustable `max.queue.size.in.bytes`

wcy-fdu:

`max.queue.size.in.bytes` is a parameter used to limit the total memory size occupied by Debezium's internal queues, preventing excessive JVM heap memory consumption due to accumulated change records in the queue, which could lead to an OOM. Previously we do not config `max.queue.size.in.bytes`, which means this part of memory is unlimited.

This pr make it configurable and adjustable by introducing a new parameter called `debezium.queue.memory.ratio`, the meaning of this parameter is the maximum bytes of the Debezium queue as a proportion of the entire JVM_HEAP_SIZE. The allowed range is 0.01 to 0.8, and users can configure this within that range and modify it dynamically to adjust the total size of the Debezium internal queue.

`JVM_HEAP_SIZE` is typically 0.07 of the system's remaining memory. If configured, it is determined by the user's settings. Therefore, the allocation for the Debezium internal queue is between 0.01 and 0.8 of this size.

To set it in create source
```
create source s with (
  username = 'postgres',
  connector='postgres-cdc',
  hostname='localhost',
  port='8432',
  password = '',
  database.name = 'postgres',
  schema.name = 'public',
  slot.name = 'rw_slot',
  auto.schema.change = 'true',
  debezium.queue.memory.ratio = '0.3',
);
```

To modify it for a running source
```
ALTER SOURCE s CONNECTOR WITH (debezium.queue.memory.ratio = 0.4);
```

If user do not set `debezium.queue.memory.ratio`, fallback to original behavior, which is DEFAULT_MAX_QUEUE_SIZE_IN_BYTES=0, meaning unlimited memory.
### About how Debezium avoiding deadlock

refer to [debezium source code](https://github.com/debezium/debezium/blob/a04fc51857dbefddeb990dc5a7265f9f7e001a5f/debezium-core/src/main/java/io/debezium/connector/base/ChangeEventQueue.java#L244-L264)

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/src/allow_alter_on_fly_fields.rs
+++ b/src/connector/src/allow_alter_on_fly_fields.rs
@@ -119,6 +119,7 @@ pub static SOURCE_ALLOW_ALTER_ON_FLY_FIELDS: LazyLock<HashMap<String, HashSet<St
             "cdc.source.wait.streaming.start.timeout".to_owned(),
             "debezium.max.queue.size".to_owned(),
             "debezium.queue.memory.ratio".to_owned(),
+
         ].into_iter().collect(),
     ).unwrap();
 


java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/DbzConnectorConfig.java
```diff
@@ -354,4 +369,76 @@ private Properties initiateDbConfig(String fileName, StringSubstitutor substitut
...
+            // Validate ratio is in reasonable range (1% - 80%)
+            if (queueMemoryRatio < MIN_QUEUE_MAX_MEMORY_RATIO
+                    || queueMemoryRatio > MAX_QUEUE_MAX_MEMORY_RATIO) {
+                LOG.warn(
+                        "Invalid {}: {}, must be between {} and {}, ignoring",
+                        QUEUE_MAX_MEMORY_RATIO,
+                        queueMemoryRatio,
+                        MIN_QUEUE_MAX_MEMORY_RATIO,
+                        MAX_QUEUE_MAX_MEMORY_RATIO);
+                return;
```
hzxa21: Can we validate this during `CREATE SOURCE` and `ALTER SOURCE s CONNECTOR WITH` and throw an SQL error if user inputs a bad ratio?

wcy-fdu: Sounds good. Modified.

java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/DbzConnectorConfig.java
```diff
@@ -332,7 +336,18 @@ public DbzConnectorConfig(
             dbzProps.putIfAbsent(entry.getKey(), entry.getValue());
         }
 
+        // Calculate max.queue.size.in.bytes based on JVM heap size and ratio
+        calculateAndSetMaxQueueSizeInBytes(dbzProps);
+
         LOG.info("Final Debezium properties: {}", dbzProps);
+        LOG.info(
```
zwang28, line 345: Isn't max.queue.size.in.bytes already logged in previous line?

src/connector/src/allow_alter_on_fly_fields.rs
```diff
@@ -99,20 +99,36 @@ pub static SOURCE_ALLOW_ALTER_ON_FLY_FIELDS: LazyLock<HashMap<String, HashSet<St
     // CDC Properties - added for schema.change.failure.policy
     map.try_insert(
         std::any::type_name::<MysqlCdcProperties>().to_owned(),
-        ["cdc.source.wait.streaming.start.timeout".to_owned()].into_iter().collect(),
+        [
+            "cdc.source.wait.streaming.start.timeout".to_owned(),
+            "debezium.max.queue.size".to_owned(),
+            "debezium.queue.memory.ratio".to_owned(),
```
zwang28, line 105: Why is a ratio used instead of bytes? I think bytes is more straightforward.

wcy-fdu: I think it'e better to use ratio:
1. `JVM_HEAP_SIZE`, the memory allocated to all JVMs, can be set in bytes. When not set, it's 0.07 * reserved memory. However, when set to bytes, it's not reset when the cluster scales up or down (this is a known issue). In this case, setting `debezium.queue.memory.bytes` instead of a percentage of JVM_HEAP_SIZE can easily lead to errors.
2. Setting this parameter is highly likely to cause an JVM OOM during operation and maintenance, as frequent scaling up and down operations are inevitable. As mentioned above, after scaling the cluster, the size needs to be reset, which is troublesome for both users and our operations team.

java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/DbzConnectorConfig.java
wcy-fdu: No, it's not a debezium config, but a parameter we introduced.

src/connector/src/allow_alter_on_fly_fields.rs
wcy-fdu: cc @zwang28 😁

PR #23854, opened 2025-11-26
fix(ci): update connector node integration test pipeline to java 17 & 21

tabVersion:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## What's changed and what's your intention?

following #23600 

the workflow succeeds https://buildkite.com/risingwavelabs/pull-request/builds/86116 

## Checklist

- [ ] I have written necessary rustdoc comments.
- [ ]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  I have checked the [Release Timeline](https://github.com/risingwavelabs/rw-commits-history/blob/main/release_timeline.md) and [Currently Supported Versions](https://docs.risingwave.com/changelog/release-support-policy#support-end-dates-for-recent-releases) to determine which release branches I need to cherry-pick this PR into. 

## Documentation

- [ ]  My PR needs documentation updates. 

<details>
<summary><b>Release note</b></summary>

</details>

PR #23900, opened 2025-11-28
fix(ci): fix several failures in main-cron

BugenZhao:

1. We may not exactly match the "topic not found" error for pulsar source, because the `available topics` clause is not stable due to parallelization.
2. Increase timeout for `e2e-source-test` and `unit-test`. The previous value is even shorter than pull-request's.
3. Remove Java 11 and replace it with Java 21 in main-cron as well #23854 

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/ci/workflows/main-cron.yml
+++ b/ci/workflows/main-cron.yml
@@ -958,8 +958,8 @@ steps:
     matrix:
       setup:
         java_version:
-          - "11"
           - "17"
+          - "21"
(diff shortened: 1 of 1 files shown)


BugenZhao:
There's no timeout anymore. The remaining failure is #23901.

https://buildkite.com/risingwavelabs/main-cron/builds/7916

tabVersion (approved):

PR #23902, opened 2025-11-28
chore: upgrade parking_lot to 0.12.5

Li0k:

https://github.com/Amanieu/parking_lot/issues/465

In `src/storage/src/hummock/event_handler/hummock_event_handler.rs`

we spin over pending and retry try_write_for with a fixed 1 ms timeout:

```rust
while let Some(instance_id) = pending.pop_front() {
    let (_, read_version) = self.local_read_version_mapping.get(&instance_id).expect(...);
    match read_version.try_write_for(TRY_LOCK_TIMEOUT) {
        Some(mut write_guard) => f(instance_id, &mut write_guard),
        _ => {
            warn!(instance_id, "failed to get lock again for instance");
            pending.push_back(instance_id);
        }
    }
}
```

parking_lot issue  causes **try_write_for** to keep returning None once there are waiting writers, even after the lock becomes free. With the retry loop above, that turns into an infinite retry on the same instance_id, effectively wedging the handler.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5348,9 +5348,9 @@ dependencies = [
 
 [[package]]
 name = "fastant"
-version = "0.1.10"
+version = "0.1.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "62bf7fa928ce0c4a43bd6e7d1235318fc32ac3a3dea06a2208c44e729449471a"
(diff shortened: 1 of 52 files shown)


hzxa21 (approved):
LGTM!

Issue #17904, opened 2024-08-02
Set `pause_on_next_bootstrap` on an offline cluster


kwannoel:
Support changing it in `risingwave.toml`. In certain cases the cluster is crash looping, and because in standalone modes the faults between different nodes are not isolated as they are all embedded, the cluster can't be easily recovered.

Edit: Workaround for now: `RW_COMPUTE_NODE_ROLE=serving`. This pauses all streaming jobs.

BugenZhao:
> In certain cases the cluster is crash looping

I suppose you are referring to process panicking. Regular soft failures in streaming jobs should not affect the capability of setting system parameters with SQL interfaces.

kwannoel:
> > In certain cases the cluster is crash looping
> 
> I suppose you are referring to process panicking. Regular soft failures in streaming jobs should not affect the capability of setting system parameters with SQL interfaces.

Yeah panic, regular soft failure will just trigger recovery, and meta + frontend should still be responsive.

kwannoel:
Already supported.
```
[system]
pause_on_next_bootstrap = true
```

BugenZhao:
> Already supported.
> 
> ```
> [system]
> pause_on_next_bootstrap = true
> ```

Ah. Specifying system parameters through config file may only take effect on fresh cluster.

kwannoel:
@EdwinaZhu will be working on this, by providing a new `pause_on_next_boostrap_offline` configuration to the meta node.

PR #23910, opened 2025-11-28
feat: support setting pause_on_next_bootstrap_offline

EdwinaZhu:

Added a `pause_on_next_bootstrap_offline` switch under `[meta]` so developers can flip the bootstrap pause behavior directly from `risingwave.toml`; 
meta startup now copies that flag into the `pause_on_next_bootstrap` system parameter before bootstrapping.
Solved the issue #17904 .

## Checklist

## Documentation

<details>

A new configuration flag, `[meta].pause_on_next_bootstrap_offline`, is now available in `risingwave.toml`. This flag is **especially useful in scenarios where the cluster is running in standalone mode and enters a crash loop**—for example, due to an unresolvable source or sink.

Previously, users could only set the `pause_on_next_bootstrap` system parameter, either using .toml that only works on a fresh cluster or by using `ALTER SYSTEM SET` while the cluster was already running stably. If a cluster was unable to start (i.e., caught in a crash loop), there was no easy way to pause sources, as manual commands couldn’t be issued.

With this new setting:
- Users can **directly pause all sources by adding this flag to `risingwave.toml`**—even if the cluster is stuck and cannot be recovered via SQL or `risectl`.
- The meta node will automatically pause sources on the next bootstrap, making it much easier to perform troubleshooting, recovery, or maintenance on developer or offline clusters without extra manual intervention.
- This approach solves [#17904](https://github.com/risingwavelabs/risingwave/issues/17904) by providing a straightforward and reproducible recovery option.

**How to use:**
Add or update the following section in your `risingwave.toml`:
```toml
[meta]
pause_on_next_bootstrap_offline = true  # or false
```

</details>

### DIFF
--- a/src/meta/node/src/lib.rs
+++ b/src/meta/node/src/lib.rs
@@ -552,7 +552,7 @@ pub fn start(
                     .cdc_table_split_init_insert_batch_size,
 
                 enable_legacy_table_migration: config.meta.enable_legacy_table_migration,
-                pause_on_next_bootstrap: config.meta.pause_on_next_bootstrap_offline,
+                pause_on_next_bootstrap_offline: config.meta.pause_on_next_bootstrap_offline,
             },
             config.system.into_init_system_params(),
             Default::default(),
--- a/src/meta/src/barrier/worker.rs
+++ b/src/meta/src/barrier/worker.rs
@@ -176,7 +176,7 @@ impl GlobalBarrierWorker<GlobalBarrierWorkerContextImpl> {
             .system_params_reader()
             .await
             .pause_on_next_bootstrap()
-            || self.env.opts.pause_on_next_bootstrap;
+            || self.env.opts.pause_on_next_bootstrap_offline;
 
         if paused {
             warn!(
--- a/src/meta/src/manager/env.rs
+++ b/src/meta/src/manager/env.rs
@@ -293,7 +293,7 @@ pub struct MetaOpts {
     pub cdc_table_split_init_insert_batch_size: u64,
 
     pub enable_legacy_table_migration: bool,
-    pub pause_on_next_bootstrap: bool,
+    pub pause_on_next_bootstrap_offline: bool,
 }
 
 impl MetaOpts {
@@ -388,7 +388,7 @@ impl MetaOpts {
             cdc_table_split_init_insert_batch_size: 1000,
             enable_legacy_table_migration: true,
             refresh_scheduler_interval_sec: 60,
-            pause_on_next_bootstrap: false,
+            pause_on_next_bootstrap_offline: false,
         }
     }
 }


kwannoel (approved):
Looks good, thank you!

src/common/src/config/meta.rs
```diff
@@ -158,6 +158,12 @@ pub struct MetaConfig {
     #[serde(default)]
     pub disable_recovery: bool,
 
+    /// Whether meta should request pausing all data sources on the next bootstrap.
+    /// This is a convenience switch that sets the `pause_on_next_bootstrap` system parameter when
+    /// the meta service starts.
```
kwannoel:
```suggestion
    /// This allows us to pause the cluster on next bootstrap in an offline way.
    /// It's important for standalone or single node deployments.
    /// In those cases, meta node, frontend and compute may all be co-located.
    /// If the compute node enters an inconsistent state, and continously crashloops, 
    /// we may not be able to connect to the cluster to run `alter system set pause_on_next_bootstrap = true;`.
    /// By providing it in the static config, we can have an offline way to trigger the pause on bootstrap.
```

You can refine this comment.

BugenZhao:
I don't think this works. We're still operating on the **`init_`**`system_param`, which will be ignored unless it's a fresh cluster.

https://github.com/risingwavelabs/risingwave/blob/0132a8646be07b13f0edcfa51f81776cbf69ff79/src/meta/src/controller/system_param.rs#L100-L144

kwannoel:
> I don't think this works. We're still operating on the **`init_`**`system_param`, which will be ignored unless it's a fresh cluster.
> 
> https://github.com/risingwavelabs/risingwave/blob/0132a8646be07b13f0edcfa51f81776cbf69ff79/src/meta/src/controller/system_param.rs#L100-L144

@EdwinaZhu we can set it here instead: https://github.com/risingwavelabs/risingwave/blob/526a9a73794f3c6166687b824d9753378a7a3817/src/meta/src/barrier/worker.rs#L176-L194

src/meta/node/src/lib.rs
```diff
@@ -552,6 +552,7 @@ pub fn start(
                     .cdc_table_split_init_insert_batch_size,
 
                 enable_legacy_table_migration: config.meta.enable_legacy_table_migration,
+                pause_on_next_bootstrap: config.meta.pause_on_next_bootstrap_offline,
```
kwannoel: Can we make the naming consistent, the field name should match

kwannoel: `pause_on_next_bootstrap_offline`

kwannoel: `pause_on_next_bootstrap_offline`

kwannoel (approved):

EdwinaZhu:
@copilot please expand the "Release note" section in this PR description.
Goal: explain `[meta].pause_on_next_bootstrap_offline`, what it toggles, when to use it, and how it helps testing recovery on dev clusters.

PR #23945, opened 2025-12-02
feat(cdc): introduce reset source command

wcy-fdu:

This pr introduce reset source command, which is `alter source s reset;`. This command is actually an operational method, primarily used for recovery when the CDC offset recorded in the state table of the source executor no longer exists upstream.

In simpler terms, using MySQL as an example, when the binlog offset has expired in the upstream binlog, this source can no longer consume data from upstream. The command `ALTER SOURCE s RESET;` sets the offset in the source state table from an expired value to null. Therefore, the next time Debezium starts, it will automatically set the offset to the latest value.  At this point, the source will start consuming from the current latest offset.

### Importane Note
1. This command is only meant to be used when the offset no longer exists upstream, because if the offset is still present, the source will continue to consume normally. Even if it is set to null, it will be reset the next time new data comes in. The most likely databases to experience this issue with CDC are TTL-based ones, such as MySQL and MongoDB. However, it is also possible for offset expiration to occur with PostgreSQL and MSSQL CDC.

2. After executing this command, a recovery or restart needs to be triggered. At this point, regardless of whether there is new data coming in from upstream, the offset will be set to the latest value.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/expr/impl/src/scalar/array_reverse.rs
+++ b/src/expr/impl/src/scalar/array_reverse.rs
@@ -1,4 +1,4 @@
-// Copyright 2025 RisingWave Labs
+// Copyright 2026 RisingWave Labs
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


tabVersion:
@cursoragent help review the pr plz

wcy-fdu:
PTAL ❤️cc @hzxa21 @zwang28

wcy-fdu:
>TL, DR cursor comment

I browsed the comments on cursor, and I found most of them to be rather worthless. I modified some of the code comments to make it more rigorous.

Issue #23977, opened 2025-12-04
Review usage of the executor_id


hzxa21:
`executor_id` is constructed via `actor_id << 32 | operator_id`. There are 2 indications:
1. `executor_id` is unique within each RW cluster. However, it is not necessarily unique across different RW clusters.
2. After https://github.com/risingwavelabs/risingwave/pull/23528, it is possible that the same actor id can assign to different jobs after cluster recovery/restart. This means `executor_id` is not unique across cluster recovery/restart.

We need to review the usage of `executor_id` and make sure we use it correctly. For example, in starrocks sink, we are using `rw-txn-{executor_id}-{ts}` as the txn id, which means if there are two RW cluster sinking data to the same SR cluster, even they sink to different target table, the txn id can conflict with each other. This looks risky to me.

Also, we are using `executor_id` in many sinks including file, monogodb, redshift, snowflake, starrrocks, mqtt sinks. We should also review the usage outside of sink in other components as well.

hzxa21:
@shanicky @wcy-fdu @xxhZs @wenym1

wcy-fdu:
IIRC, file sink with(sink_decouple = false) using executor_id as part of the output file name, but currently sink_decouple is by default enabled. I will take a look to double confirm.

wcy-fdu:
> IIRC, file sink with(sink_decouple = false) using executor_id as part of the output file name, but currently sink_decouple is by default enabled. I will take a look to double confirm.

I remembered it wrong. Regardless of whether it's decoupled or not, it always includes executor_id...

hzxa21:
@xxhZs reminder to review the usage of executor_id in various sinks.

PR #23981, opened 2025-12-04
refactor(file_sink): do not use executor id for file sink

wcy-fdu:

detail in https://github.com/risingwavelabs/risingwave/issues/23977

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/src/sink/file_sink/opendal_sink.rs
+++ b/src/connector/src/sink/file_sink/opendal_sink.rs
@@ -154,12 +154,13 @@ impl<S: OpendalSinkBackend> Sink for FileSink<S> {
 
     async fn new_log_sinker(
         &self,
-        _writer_param: crate::sink::SinkWriterParam,
+        writer_param: crate::sink::SinkWriterParam,
     ) -> Result<Self::LogSinker> {
         let writer = OpenDalSinkWriter::new(
             self.op.clone(),
             &self.path,
             self.schema.clone(),
+            writer_param.executor_id,
             &self.format_desc,
             self.engine_type.clone(),
             self.batching_strategy.clone(),
@@ -206,6 +207,7 @@ pub struct OpenDalSinkWriter {
     operator: Operator,
     sink_writer: Option<FileWriterEnum>,
     write_path: String,
+    executor_id: u64,
     unique_writer_id: Uuid,
     encode_type: SinkEncode,
     row_encoder: JsonEncoder,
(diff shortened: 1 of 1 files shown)


src/connector/src/sink/file_sink/opendal_sink.rs
```diff
@@ -251,12 +251,8 @@ impl OpenDalSinkWriter {
...
                         let metadata = w.close().await?;
                         tracing::info!(
-                            "writer {:?}_{:?}finish write file, metadata: {:?}",
-                            self.executor_id,
-                            self.created_time
-                                .duration_since(UNIX_EPOCH)
-                                .expect("Time went backwards")
-                                .as_secs(),
+                            "writer {} finish write file, metadata: {:?}",
+                            self.unique_writer_id,
```
hzxa21, line 257: I think it is better for debugging to include `executor_id` and the `created_time` as part of the path.

Also, is this a breaking change? Do we explicitly mention the sink file name format somewhere to our user?

wcy-fdu: > think it is better for debugging to include executor_id and the created_time as part of the path.

Agree.

>Also, is this a breaking change? Do we explicitly mention the sink file name format somewhere to our user?

No, the change is safe. We don't use `executor_id`, we only use `created_at` to split files when specifying `partition_by`.

hzxa21: I mean from existing file sink created prior to this PR, files before upgrading will have old naming while files after upgrading will have new naming. Will this be an issue?

If we don't communicate the actual file naming to the user, it is fine. Otherwise, this can be a breaking change for user's downstream application if they somehow rely on the file naming to do processing.

wcy-fdu: >If we don't communicate the actual file naming to the user, it is fine.

To my recollection, the specific meaning of this filename was never explained to the user.🤔 Do you have any recommendations for better ways to modify the naming rules?

wcy-fdu: We can highlight this change in doc and release note, to indicates that further operations cannot be performed based on the executor_id in the filename. Besides that, it seems there's nothing else that can be done.

hzxa21: No. I am just thinking whether we need to keep the old naming for already created sink and only use the new naming for new sink. If the naming is not exposed to the user, that is fine.

hzxa21: oh... The naming is exposed in our doc: https://docs.risingwave.com/delivery/overview#file-organization

hzxa21: That is fine. You are still using the `/Option<path_partition_prefix>/executor_id + timestamp.suffix` format in this PR. You just changed the value of `executor_id`.

wcy-fdu: yes, however, the length of the current UUID is different from that of the previous executor_id: 
before this pr
```
data/4294967296_1733299200.parquet
```
after this pr
```
data/01935d2c-8f12-7890-abcd-ef1234567890_1733299200.parquet
```

Only have issue when user parse the executor_id and use it in a special way(For example, comparing lengths). But I really don't think any user would use it like that.

hzxa21: I think that is fine. We need to mention that in the doc site and release note though.

PR #23988, opened 2025-12-04
feat: snowflake source support

tabVersion:

## Checklist

## Documentation

<details>

---

## Type Mapping

|Snowflake Type|Risingwave Type|
|--|--|
|string|varchar|
|number|decimal / bigint|
|float|double precision|
|decimal|decimal|
|char|varchar|
|text|varchar|
|date|date|
|time| NA(report error) |
|timestamp_ntz|timestamp without time zone|
|timestamp_ltz|timestamp with time zone|
|timestamp_tz|timestamp with time zone|
|bool|boolean|
|binary|bytea|
|variant|varchar|
|object|varchar|
|array|varchar|

Based on the PR diff, here's a user-facing release notes entry:

---

### Overview

RisingWave now supports loading data from Snowflake tables using the ADBC (Arrow Database Connectivity) protocol. This enables you to create refreshable tables that periodically pull data from Snowflake.

### Usage

Create a refreshable table from a Snowflake table:

```sql
CREATE TABLE my_snowflake_table (
primary key ("some_key")) WITH (
    connector = 'adbc_snowflake',
    refresh_mode = 'FULL_RELOAD',
    refresh_interval_sec = '3600',  -- Refresh every hour
    
    -- Snowflake connection parameters
    adbc_snowflake.account = 'myaccount.us-east-1',
    adbc_snowflake.username = 'myuser',
    adbc_snowflake.password = 'mypassword',
    adbc_snowflake.database = 'MYDB',
    adbc_snowflake.schema = 'MYSCHEMA',
    adbc_snowflake.warehouse = 'MYWH',
    adbc_snowflake.table = 'MYTABLE'
);
```

### Key Features

- **Automatic Schema Inference**: No need to manually specify column definitions - the schema is automatically detected from the Snowflake table
- **Consistent Snapshots**: Uses Snowflake's time travel feature to ensure all data is read from the same point in time
- **Multiple Authentication Methods**: Supports various authentication types including OAuth, JWT, SSO, and more

### Supported Authentication Options

- `auth_snowflake` (default): Username/password authentication
- `auth_oauth`: OAuth token authentication
- `auth_jwt`: JWT private key authentication  
- `auth_ext_browser`: External browser SSO
- `auth_okta`, `auth_mfa`, `auth_pat`, `auth_wif`: Other authentication methods

### Requirements & Limitations

- **Feature Flag**: Requires the `source-adbc_snowflake` feature to be enabled at compile time (enabled by default in official builds)
- **Refresh Mode**: Only `FULL_RELOAD` mode is supported - the entire table is re-read on each refresh
- **Table Schema**: Do not specify column definitions in the `CREATE TABLE` statement - they are automatically inferred

### Example with JWT Authentication

```sql
CREATE TABLE snowflake_data WITH (
    connector = 'adbc_snowflake',
    refresh_mode = 'FULL_RELOAD',
    refresh_interval_sec = '7200',
    
    adbc_snowflake.account = 'myaccount',
    adbc_snowflake.username = 'myuser',
    adbc_snowflake.database = 'MYDB',
    adbc_snowflake.schema = 'PUBLIC',
    adbc_snowflake.warehouse = 'COMPUTE_WH',
    adbc_snowflake.table = 'ORDERS',
    adbc_snowflake.auth_type = 'auth_jwt',
    adbc_snowflake.jwt_private_key_path = '/path/to/key.pem'
);
```

### Future Enhancements

Parallel fetching support is planned to improve performance for large tables by partitioning data across multiple workers based on primary key ranges.

---

</details>

### DIFF
--- a/e2e_test/source_inline/refresh/refresh_table.slt
+++ b/e2e_test/source_inline/refresh/refresh_table.slt
@@ -75,9 +75,7 @@ CREATE SOURCE refresh_fs_t (id int PRIMARY KEY, name varchar) WITH (
 db error: ERROR: Failed to run the query
 
 Caused by:
-  Bind error: can't CREATE SOURCE with refreshable batch connector
-
-Hint: use CREATE TABLE instead
+  Protocol error: `refresh_mode` only supported for CREATE TABLE
 
 
 statement ok


src/stream/src/executor/source/batch_source/batch_adbc_snowflake_list.rs
```diff
@@ -0,0 +1,553 @@
...
+        {
+            let min_val: String = min_array.value(0).into();
+            let max_val: String = max_array.value(0).into();
+            return Some((min_val, max_val));
+        }
+        None
+    }
+
+    /// Generate WHERE clause for a split based on PK range
+    fn generate_where_clause(
```
chenzl25: What if the first pk column type is not a number?

src/stream/src/executor/source/batch_source/batch_adbc_snowflake_list.rs
```diff
@@ -0,0 +1,553 @@
...
+        connection: &mut risingwave_connector::source::adbc_snowflake::Connection,
+        pk_column: &str,
+        snapshot_timestamp: Option<&str>,
+    ) -> StreamExecutorResult<(String, String)> {
+        // Try with snapshot first if available
+        if let Some(ts) = snapshot_timestamp {
+            let table_expr_with_snapshot =
+                format!("{} AT(TIMESTAMP => '{}')", properties.table_ref(), ts);
+            let range_query = format!(
+                "SELECT MIN({})::STRING, MAX({})::STRING FROM {}",
```
chenzl25: This query looks like a bit heavy, do we split it in the same way in parallel cdc backfill? @zwang28

src/stream/src/executor/source/batch_source/batch_adbc_snowflake_list.rs
tabVersion: Actually planning to remove the parallel loading part, cuz PK is not necessarily the distribution key in snowflake. Scan over pk does not ensure best performance.

src/stream/src/executor/source/batch_source/batch_adbc_snowflake_list.rs
```diff
@@ -0,0 +1,553 @@
...
+                snapshot_timestamp: snapshot_timestamp.clone(),
+            };
+        } else {
+            // Generate splits based on primary key ranges
+            // For simplicity, we'll use the first primary key column
+            let pk_column = &pk_columns[0];
+
+            // Determine number of splits based on estimated row count
+            // Use roughly 100k rows per split as a heuristic
+            let rows_per_split = 100_000;
```
chenzl25: I feel 100k is a bit small for a split. How about 1M?

chenzl25:
cc @xxhZs  Please take a look and get familiar with ADBC snowflake, so that we can migrate snowflake sink from jdbc to adbc in the future if possible.

scripts/check/check-trailing-spaces.sh
tabVersion, line 1: Made the change because of meeting this error on CI but not reproducible on local side. 

The script checks `.cargo/` dir, which usually do not have issue. 

Image

cc @BugenZhao

PR #23993, opened 2025-12-04
fix: fix auto cdc schema change with schema binding and memory leak in fe

yezizp2012:

This PR fixes two bugs:
1. replace table's catalog schema will always be `public`, fortunately the schema field has not been updated in meta.
2. The dummy session was never explicitly ended, resulting in a memory leak in the FE.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/handler/create_table.rs
+++ b/src/frontend/src/handler/create_table.rs
@@ -2465,8 +2465,7 @@ fn get_source_and_resolved_table_name(
     table_name: ObjectName,
 ) -> Result<(Arc<SourceCatalog>, String)> {
     let db_name = &session.database();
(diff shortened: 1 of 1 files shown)


src/frontend/src/handler/create_table.rs
```diff
@@ -1158,13 +1158,13 @@ pub(super) async fn handle_create_table_plan(
                 session.get_database_and_schema_id_for_create(schema_name.clone())?;
 
             // cdc table cannot be append-only
-            let (format_encode, source_name) =
+            let (source_schema, source_name) =
```
yezizp2012, line 1161: It seems like someone previously made a bulk "search and replace" mistake.

PR #24002, opened 2025-12-04
feat: support risingwave scalar function in datafusion

silver-ymz:

This PR adds support for RisingWave scalar functions in DataFusion. Some expressions will be converted to datafusion's native implementation, e.x. `eq`. For the remaining expressions, DataFusion will call the corresponding methods in RisingWave's dynamic Expression trait. However, this latter approach incurs extra conversion overhead.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/datafusion/function.rs
+++ b/src/frontend/src/datafusion/function.rs
@@ -292,8 +292,10 @@ impl ScalarUDFImpl for RwScalarFunction {
         let chunk = IcebergArrowConvert
             .chunk_from_record_batch(&record_batch)
             .map_err(boxed)?;
-        // TODO: better way to run async eval in sync context
-        let value = futures::executor::block_on(self.expr.eval_v2(&chunk)).map_err(boxed)?;
+        let value = tokio::task::block_in_place(|| {
+            tokio::runtime::Handle::current().block_on(self.expr.eval_v2(&chunk))
+        })
+        .map_err(boxed)?;
         let res = match value {
             ValueImpl::Array(array_impl) => {
                 let array = IcebergArrowConvert


chenzl25:
FYI, I am bumping the iceberg-rust together with datafusion from (v45 to v50). Hope the bumping won't cause lots of conflicts. https://github.com/risingwavelabs/risingwave/pull/24004

src/frontend/src/datafusion/function.rs
```diff
@@ -0,0 +1,319 @@
...
+    let func = RwScalarFunction {
+        name: format!(
+            "RisingWave Scalar Function (type: {})",
+            func_call.func_type().as_str_name()
+        ),
+        expr: boxed_expr,
+        input_columns: column_vec,
+        signature: Signature {
+            type_signature: TypeSignature::Any(input_columns.len()),
+            volatility: Volatility::Volatile,
```
BugenZhao, line 236: nit: We may consider `is_pure`.

src/frontend/src/datafusion/function.rs
```diff
@@ -0,0 +1,319 @@
...
+            .zip_eq_fast(self.input_columns.iter())
+            .map(|(array, col)| Field::new(col.name(), array.data_type().clone(), true))
+            .collect();
+        let input_schema = Arc::new(Schema::new(input_fields));
+        let record_batch = RecordBatch::try_new_with_options(input_schema, columns, &opts)?;
+        let chunk = IcebergArrowConvert
+            .chunk_from_record_batch(&record_batch)
+            .map_err(boxed)?;
+        // TODO: better way to run async eval in sync context
+        let value = futures::executor::block_on(self.expr.eval_v2(&chunk)).map_err(boxed)?;
```
BugenZhao: nit: I guess using `tokio` could be better here, as some function impls may use `tokio` IO functions and then depend on `tokio` executor?

silver-ymz: updated now. It uses `tokio::task::block_in_place(|| tokio::runtime::Handle::current().block_on(...))`

src/frontend/src/datafusion/function.rs
```diff
@@ -0,0 +1,321 @@
...
+        .data_type()
+        .clone();
+    Some(DFExpr::Cast(Cast {
+        expr: Box::new(arg),
+        data_type: target_type,
+    }))
+}
+
+// TODO: verify more cast safety cases
+// case 1: IcebergArrowConvert just converts Jsonb to varchar. So any cast from or to Jsonb should not be handled by DataFusion
```
chenzl25, line 202: We convert jonsb to varchar for iceberg, because Iceberg V2 spec doesn't have a json type, while the V3 spec has introduce a new data type called variant which is more suitable for jsonb data type. However, during the datafusion expr evaluation, it seems we can convert a jsonb to datafusion jsonb(if they have) **using a new jsonb converter**, because the iceberg scan will recover a jsonb type from a varchar type from the storage and during execution, the upper operators like project and filters can always treat jsonb as jsonb.  Anyway, this can be implemented in the future as an improvement.

src/frontend/src/datafusion/execute.rs
```diff
@@ -0,0 +1,184 @@
...
+        .query_planner()
+        .create_physical_plan(&plan.plan, &state)
+        .await?;
+    let data_stream = execute_stream(physical_plan, ctx.task_ctx())?;
+
+    let compute_runtime = session.env().compute_runtime();
+    let (sender1, receiver) = mpsc::channel(10);
+    let shutdown_rx = session.reset_cancel_query_flag();
+    let sender2 = sender1.clone();
+    let cast_executor = build_cast_executor(&plan.schema)?;
```
chenzl25, line 70: Why do we need this cast executor? Could you please use an example to illustrate?

silver-ymz: `IcebergArrowConvert` currently converts Arrow `RecordBatch` values into `DataChunk` without applying the target RisingWave logical types. This means it never performs the necessary down-casts. For example, Iceberg reads `smallint` (`INT16`) columns as Arrow `Int32`, and `IcebergArrowConvert` preserves that `Int32` instead of converting it back to the expected `Int16`.

Example:
```sql
CREATE TABLE t2 (
    v_small_int smallint,
) WITH (commit_checkpoint_interval = 1) ENGINE = iceberg;
INSERT INTO t2 VALUES (1);
SELECT v_small_int FROM t2;
```

The scan produces a DataChunk with schema [Int32], while the table schema requires [Int16], causing:
```
Panicked when handling the request: chunk_data_types: [Int32], column_types: [Int16].
```

src/frontend/src/datafusion/function.rs
silver-ymz: Jsonb is supported, but not here.

This cast check just determines whether a cast can safely be delegated to DataFusion. If the cast isn’t supported, we fall back to `fallback_rw_expr_builder`, which executes the expression using RisingWave’s boxed expression framework instead.

DataFusion currently has no native `jsonb` type, so any cast involving jsonb cannot be evaluated inside the DataFusion engine. In those cases, we route the cast through RisingWave’s own expression executor. For example:

```
CAST(varchar AS jsonb)
```

will be executed as:

- DataFusion identifies that the cast involves jsonb.
- The cast is not supported, so we fall back to RisingWave’s boxed expression.
- RisingWave performs the actual JSONB cast.
- The result is converted back into a Arrow array (`varchar` here) for downstream operators.

src/frontend/src/datafusion/function.rs
```diff
@@ -0,0 +1,319 @@
...
+                ColumnarValue::Scalar(value)
+            }
+        };
+        Ok(res)
+    }
+}
+
+fn boxed<E: std::error::Error + Send + Sync + 'static>(e: E) -> BoxedError {
+    // Convert to anyhow::Error first to capture backtrace information.
+    anyhow::Error::new(e).into()
```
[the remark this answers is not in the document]
silver-ymz: There will be another PR to improve error handling.

src/frontend/src/datafusion/execute.rs
chenzl25: I see previously I thought it is handled by the Risingwave cast already, but it turns out not enough.

src/frontend/src/datafusion/function.rs
chenzl25: If we convert the jsonb back to varchar again in datafusion, how do we differentiate it from the actual varchar type?

silver-ymz: In risingwave schema, it's `jsonb`; in datafusion schema, it's `varchar`. And we will always use risingwave schema info to execute expressions. Only risingwave side will do type inference.

chenzl25: This query will panic.
```
select jsonb_array_length(v_json), v_json from full_type_t;
```

chenzl25: Actually, this case jsonb_array_length will expect a jsonb column type as input, but datafusion returns an utf8 array.

silver-ymz: We can add a cast before `Expression::eval_v2`. Could we complete it in the next PR? Because this PR is too large.

chenzl25: sure

chenzl25 (approved):
LGTM！

PR #24004, opened 2025-12-04
feat(iceberg): bump iceberg rust to 20251111 version

chenzl25:

- use arrow 56
- use datafusion 50
- use new iceberg-rs writer (need to check partition transformation again)
- - new `IcebergWriterDispatch`
- use new iceberg-rs reader (need to check the position delete filter again)
- use new iceberg-rs commit interface

Also bumped related packages in `arrow` and `datafusion` ecosystem, like `google-cloud` or `parquet` to avoid introducing too many duplicated versions.

TODOs after this PR:
- Bump deltalake to eliminate arrow 55
- Bump prost / tonic to latest version

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/src/sink/iceberg/mod.rs
+++ b/src/connector/src/sink/iceberg/mod.rs
@@ -1086,7 +1086,7 @@ impl IcebergSinkWriter {
 }
 
 impl IcebergSinkWriterInner {
-    async fn build_append_only(table: Table, writer_param: &SinkWriterParam) -> Result<Self> {
+    fn build_append_only(table: Table, writer_param: &SinkWriterParam) -> Result<Self> {
         let SinkWriterParam {
             extra_partition_col_idx,
             actor_id,
@@ -1189,7 +1189,7 @@ impl IcebergSinkWriterInner {
         })
     }
 
-    async fn build_upsert(
+    fn build_upsert(
         table: Table,
         unique_column_ids: Vec<usize>,
         writer_param: &SinkWriterParam,
@@ -1282,7 +1282,7 @@ impl IcebergSinkWriterInner {
             )
             .map_err(|err| SinkError::Iceberg(anyhow!(err)))?;
             let parquet_writer_builder = ParquetWriterBuilder::new(
-                parquet_writer_properties.clone(),
+                parquet_writer_properties,
                 Arc::new(
                     arrow_schema_to_schema(config.projected_arrow_schema_ref())
                         .map_err(|err| SinkError::Iceberg(anyhow!(err)))?,
@@ -1306,7 +1306,7 @@ impl IcebergSinkWriterInner {
             data_file_builder,
             position_delete_builder,
             equality_delete_builder,
-            unique_column_ids.clone(),
+            unique_column_ids,
             schema.clone(),
         );
         let original_arrow_schema = Arc::new(
@@ -1375,15 +1375,12 @@ impl SinkWriter for IcebergSinkWriter {
 
         let table = create_and_validate_table_impl(&args.config, &args.sink_param).await?;
         let inner = match &args.unique_column_ids {
-            Some(unique_column_ids) => {
-                IcebergSinkWriterInner::build_upsert(
-                    table,
-                    unique_column_ids.clone(),
-                    &args.writer_param,
-                )
-                .await?
-            }
-            None => IcebergSinkWriterInner::build_append_only(table, &args.writer_param).await?,
+            Some(unique_column_ids) => IcebergSinkWriterInner::build_upsert(
+                table,
+                unique_column_ids.clone(),
+                &args.writer_param,
+            )?,
+            None => IcebergSinkWriterInner::build_append_only(table, &args.writer_param)?,
         };
 
         *self = IcebergSinkWriter::Initialized(inner);


chenzl25:
To fix the ci image build. Need https://github.com/risingwavelabs/risingwave/pull/24035

silver-ymz:
There is a non-trival update for datafusion.

DataFusion now requires `ScalarUDFImpl` to implement `Eq + Hash` for expression deduplication. Since physical expressions are not comparable, RwScalarFunction defines equality purely on logical identity.

To make this correct, the function name now embeds the full FunctionCall: `format!("RisingWave Scalar Function ({:?})", func_call)` instead of only func_type. This guarantees that two calls with different logical inputs (e.g. add($1, $2) vs add($1, $3)) produce different names, while repeated uses of the same logical expression share the same name.

DataFusion’s deduplication (e.g. extracting join keys from projections) only compares the logical shape of the expression, so this name-based `Eq + Hash` is sufficient and correct: identical logical expressions map to the same UDF, and distinct ones never collide.

src/meta/src/manager/iceberg_compaction.rs
```diff
@@ -826,27 +826,17 @@ impl IcebergCompactionManager {
...
-            expired_snapshots = expired_snapshots.retain_last(retain_last);
-        }
-
-        expired_snapshots = expired_snapshots
-            .clear_expired_files(iceberg_config.snapshot_expiration_clear_expired_files);
-
-        expired_snapshots = expired_snapshots
+        let expired_snapshots = txn
+            .expire_snapshot()
+            .expire_older_than(snapshot_expiration_timestamp_ms)
```
Li0k, line 833: We must handle `snapshot_expiration_retain_last`

Li0k (changes requested):

ci/Dockerfile
```diff
@@ -41,6 +41,15 @@ RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- --no-mo
...
 ENV PATH=/root/.cargo/bin/:$PATH
 
+# Install newer zstd (Ubuntu 24.04 ships 1.5.5 without ZSTD_c_targetCBlockSize)
+RUN curl -L "https://github.com/facebook/zstd/releases/download/v1.5.7/zstd-1.5.7.tar.gz" \
+    | tar -xz && \
+    cd zstd-1.5.7 && \
+    make -j"$(nproc)" && \
+    make install && \
+    ldconfig && \
+    cd / && rm -rf /zstd-1.5.7
```
BugenZhao: Which component is this needed for? Is it possible to remove it after we minimizing the version bumps of this PR? Otherwise I'm afraid we have to update `release.sh` or `docker/Dockerfile` to include such manual installation scripts, which doesn't seem friendly to maintenance.

src/meta/src/manager/iceberg_compaction.rs
chenzl25: Fixed

ci/Dockerfile
chenzl25: Let me try to remove it. I think after the downgrading of some dependencies, we don't need the zstd-1.5.7 now

chenzl25:
* **#24089** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24089?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24004** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24004?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24004?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/connector/src/sink/iceberg/mod.rs
```diff
@@ -938,6 +940,88 @@ enum ProjectIdxVec {
...
+            )?)
+        } else {
+            Some(RecordBatchPartitionSplitter::new_with_precomputed_values(
+                self.schema.clone(),
+                self.partition_spec.clone(),
+            )?)
+        };
+
+        // For unpartitioned tables, ignore any provided partition key.
+        if self.partition_spec.is_unpartitioned() {
```
Li0k: This logic is really weird. We aren't using the **partition_key**, and we use the same piece of build code regardless of whether the partition_spec is empty or not.

could we simplify it?

src/connector/src/sink/iceberg/mod.rs
```diff
@@ -1103,9 +1148,10 @@ impl IcebergSinkWriterInner {
             )
             .build();
 
-        let parquet_writer_builder = ParquetWriterBuilder::new(
-            parquet_writer_properties,
-            schema.clone(),
+        let parquet_writer_builder =
+            ParquetWriterBuilder::new(parquet_writer_properties, schema.clone());
+        let rolling_builder = RollingFileWriterBuilder::new_with_default_file_size(
```
Li0k, line 1137: Is using **default_file_size** the expected behavior?

src/connector/src/sink/iceberg/mod.rs
```diff
@@ -1252,9 +1259,10 @@ impl IcebergSinkWriterInner {
             .build();
 
         let data_file_builder = {
-            let parquet_writer_builder = ParquetWriterBuilder::new(
-                parquet_writer_properties.clone(),
-                schema.clone(),
+            let parquet_writer_builder =
+                ParquetWriterBuilder::new(parquet_writer_properties.clone(), schema.clone());
+            let rolling_writer_builder = RollingFileWriterBuilder::new_with_default_file_size(
```
Li0k, line 1247: ditto

src/connector/src/source/iceberg/mod.rs
```diff
@@ -791,9 +793,10 @@ pub async fn scan_task_to_chunk_with_deletes(
             for delete_task in equality_delete_tasks {
                 let equality_ids = delete_task.equality_ids.clone();
 
-                if equality_ids.is_empty() {
+                if equality_ids.is_none() {
```
Li0k: if let Some(equality_ids) = equality_ids else {continue}?

Li0k (approved):
Rest LGTM

src/connector/src/sink/iceberg/mod.rs
chenzl25: Good point. We don't need to impl IcebergWriterBuilder for TaskWriterBuilderWrapper

src/connector/src/sink/iceberg/mod.rs
chenzl25: Yes. Currently, we don't have such a parameter to control the target file size for writer, but we have a similar parameter for iceberg compaction.

PR #24021, opened 2025-12-05
fix: notify local subscribers when streaming job backfill finishes

shanicky:

Add a new local notification variant to signal streaming job backfill completion within the meta service. This enables interested components to react to backfill completion events in a decoupled manner using existing notification infrastructure.

- Introduce `StreamingJobBackfillFinished(JobId)` in `LocalNotification` enum
- Expose `job_id()` getter on `TrackingJob` for cleaner access
- Emit local notification in `finish_creating_job` after job completion
- Add async unit test to verify notification dispatch and reception

## Checklist

### DIFF
--- a/src/meta/src/barrier/context/context_impl.rs
+++ b/src/meta/src/barrier/context/context_impl.rs
@@ -39,6 +39,7 @@ use crate::barrier::{
     Scheduled,
 };
 use crate::hummock::CommitEpochInfo;
+use crate::manager::LocalNotification;
 use crate::model::FragmentDownstreamRelation;
 use crate::stream::{SourceChange, SplitState};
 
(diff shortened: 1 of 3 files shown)


shanicky:
> This also means that we can more easily implement `wait <job_id>` to synchronize on background ddl right?

It seems so. I'm wondering if this can be used to implement different parallelism levels for the backfill phase and the normal phase of a job.

kwannoel:
> > This also means that we can more easily implement `wait <job_id>` to synchronize on background ddl right?
> 
> It seems so. I'm wondering if this can be used to implement different parallelism levels for the backfill phase and the normal phase of a job.
> 

Elaborate?

PR #24040, opened 2025-12-08
feat(iceberg): support schema change(add column) for iceberg sink

wcy-fdu:

This pr support schema change(currently only `add column`) for iceberg sink, whih completes one of the most common pipelines in our end-to-end schema change process(Postgres->RisingWave->Iceberg).

In short, schema change behavior will passed through barrier, and when receive such barrier, iceberg sink will commit schema change by `catalog.update_table()`.

For exactly once, we will pre commit the added column in metastore, and compare it with iceberg's current schema when real commit. When failure recovery, we should check both whether data committed and meta change committed successfully.

## Checklist

## Documentation

<details>

</details>


wcy-fdu:
test passed
image

src/connector/src/sink/iceberg/mod.rs
```diff
@@ -1963,18 +2080,32 @@ impl SinglePhaseCommitCoordinator for IcebergSinkCommitter {
...
-
+        // Commit data if present
         let (write_results, snapshot_id) =
-            match self.pre_commit_inner(epoch, metadata, add_columns)? {
+            match self.pre_commit_inner(epoch, metadata, add_columns.clone())? {
                 Some((write_results, snapshot_id)) => (write_results, snapshot_id),
                 None => {
-                    tracing::debug!(?epoch, "no data to commit");
+                    // No data to commit, but may have schema change
+                    if let Some(add_columns) = add_columns {
```
wcy-fdu: Also apply those change for non exactly once situation, and it's much simpler than exactly once.

src/connector/src/sink/iceberg/mod.rs
```diff
@@ -2254,6 +2590,138 @@ impl IcebergSinkCommitter {
...
+            "All {} columns already exist in iceberg table",
+            add_columns.len()
+        );
+        Ok(true)
+    }
+
+    /// Commit schema changes (e.g., add columns) to the iceberg table.
+    /// This function uses `catalog.update_table()` API to atomically update the table schema
+    /// with optimistic locking to prevent concurrent conflicts.
+    async fn commit_schema_change(&mut self, add_columns: Vec<Field>) -> Result<()> {
```
wcy-fdu, line 2534: This function should be reviewed carefully.

Cargo.toml
```diff
@@ -171,14 +171,14 @@ hashbrown0_15 = { package = "hashbrown", version = "0.15", features = [
 ] }
 hytra = "0.1"
 # branch dev_rebase_main_20251111
-iceberg = { git = "https://github.com/risingwavelabs/iceberg-rust.git", rev = "fc3867d81aa279bdfecd8703b56144cfb80e9c12", features = [
+iceberg = { git = "https://github.com/risingwavelabs/iceberg-rust.git", rev = "15602441f017b5a36d88dfedc52e3c21116a63e1", features = [
```
chenzl25: Why do we need to change the sdk?

src/connector/src/sink/iceberg/mod.rs
```diff
@@ -2173,6 +2509,138 @@ impl IcebergSinkCommitter {
...
+
+        tracing::info!(
+            "Starting to commit schema change with {} columns",
+            add_columns.len()
+        );
+
+        // Step 1: Get current table metadata
+        let metadata = self.table.metadata();
+        let current_schema = metadata.current_schema().as_ref().clone();
+        let mut next_field_id = metadata.last_column_id + 1;
```
chenzl25: use `metadata.last_column_id()` instead ?

Cargo.toml
wcy-fdu: Minor [change](https://github.com/risingwavelabs/iceberg-rust/pull/104) in sdk: public some schema change needed field and api

chenzl25: I think we don't need to change the sdk. We can use public method instead.

src/connector/src/sink/iceberg/mod.rs
```diff
@@ -97,6 +97,103 @@ use crate::sink::{
 use crate::{deserialize_bool_from_string, deserialize_optional_string_seq_from_string};
 
 pub const ICEBERG_SINK: &str = "iceberg";
+
+// ============ Helper functions for schema change ============
+
+/// Convert Arrow `DataType` to Iceberg `Type`
+fn arrow_type_to_iceberg_type(arrow_type: &ArrowDataType) -> Result<iceberg::spec::Type> {
```
chenzl25: Can we use these method in the sdk instead? https://github.com/risingwavelabs/iceberg-rust/blob/4eafd2ca22e032c3ee4d9e6abe6ff3eae93e0189/crates/iceberg/src/arrow/schema.rs#L214-L228

src/connector/src/sink/iceberg/mod.rs
```diff
@@ -1942,45 +2116,213 @@ impl TwoPhaseCommitCoordinator for IcebergSinkCommitter {
...
+            let snapshot_id =
+                snapshot_id.ok_or_else(|| SinkError::Iceberg(anyhow!("Snapshot id missing")))?;
+            let add_columns = add_columns
+                .ok_or_else(|| SinkError::Iceberg(anyhow!("Schema metadata missing")))?;
+            return self
+                .commit_data_and_schema(epoch, write_results, snapshot_id, add_columns)
+                .await;
+        }
+
+        Ok(())
```
chenzl25: Use a `match` instead?

wcy-fdu: I thought it's clearer than match😂. NVM I have changed to `match`

src/connector/src/sink/iceberg/mod.rs
```diff
@@ -1942,45 +2059,217 @@ impl TwoPhaseCommitCoordinator for IcebergSinkCommitter {
...
+                let add_columns = add_columns
+                    .ok_or_else(|| SinkError::Iceberg(anyhow!("Schema metadata missing")))?;
+                return self
+                    .commit_data_and_schema(epoch, write_results, snapshot_id, add_columns)
+                    .await;
+            }
+            (false, false) => {}
+        }
+
+        Ok(())
```
chenzl25: we can remove the `return` and unify all the return values for each arm. 
```
  ...
  match xxx {
       case1 =>  Ok()
       case2 =>  Ok()
       case3 =>  Ok()
       case4 =>  Ok()     
  }
}
```

src/connector/src/sink/iceberg/mod.rs
```diff
@@ -1915,25 +1989,68 @@ impl TwoPhaseCommitCoordinator for IcebergSinkCommitter {
...
+                .map(|cols| cols.iter().map(|c| &c.name).collect::<Vec<_>>())
+        );
+
+        let has_schema_change = add_columns.is_some();
 
+        // Try to get write_results and snapshot_id
         let (write_results, snapshot_id) =
-            match self.pre_commit_inner(epoch, metadata, add_columns)? {
-                Some((write_results, snapshot_id)) => (write_results, snapshot_id),
+            match self.pre_commit_inner(epoch, metadata, add_columns.clone())? {
```
Chiro11: If add_columns is not needed by this function, consider removing the parameter and renaming it to `pre_commit_data` or something similar, thus avoid `clone`.

src/connector/src/sink/iceberg/mod.rs
```diff
@@ -1942,45 +2057,213 @@ impl TwoPhaseCommitCoordinator for IcebergSinkCommitter {
...
+    }
+
+    async fn abort(&mut self, _epoch: u64, _commit_metadata: Vec<u8>) {
+        // TODO: Files that have been written but not committed should be deleted.
+        tracing::debug!("Abort not implemented yet");
+    }
+}
+
+impl IcebergSinkCommitter {
+    fn parse_commit_metadata(commit_metadata: Vec<u8>) -> Result<ParsedCommitMetadata> {
```
chenzl25, line 2117: We need to better design the commit metadata serialization, because it would be persisted  in the metastore, so backward compatibility is needed. The current implementation breaks the compatibility which is undesired.

wcy-fdu: had some offline discussion, changed commit metadata to protobuf for backward compatibility

proto/iceberg_commit_metadata.proto
```diff
@@ -0,0 +1,58 @@
...
+
+package iceberg_commit_metadata;
+
+import "plan_common.proto";
+
+option java_package = "com.risingwave.proto";
+option optimize_for = SPEED;
+
+// Schema change operation for sink
+message SinkSchemaChange {
```
Chiro11: Consider placing this message in [stream_plan.proto](https://github.com/risingwavelabs/risingwave/blob/main/proto/stream_plan.proto), as it is used not only by iceberg.

wcy-fdu: Resolved.

PR #24048, opened 2025-12-09
feat(sink): Support schema change in decouple sink

Chiro11:

For barriers containing schema change message, we need to synchronize the log-store at this barrier (which means log-writer must wait for the reader to consume all data in the log-store and truncate).

Reason: Schema change cannot be persisted into the log-store, so we have to wait, either to complete the schema change in a single commit, or to ensure this message is persisted to the meta store in two-phase commit.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/common/log_store_impl/kv_log_store/writer.rs
+++ b/src/stream/src/common/log_store_impl/kv_log_store/writer.rs
@@ -181,7 +181,7 @@ impl<LS: LocalStateStore> LogWriter for KvLogStoreWriter<LS> {
             options.add_columns,
         );
         if has_schema_change {
-            let truncate_offset = self.tx.wait_for_barrier_truncation(epoch).await;
+            let truncate_offset = self.tx.wait_for_barrier_truncation(epoch).await?;
             assert_eq!(truncate_offset, (epoch, None));
             self.last_truncate_offset = Some(truncate_offset);
         }
--- a/src/stream/src/common/log_store_impl/kv_log_store/buffer.rs
+++ b/src/stream/src/common/log_store_impl/kv_log_store/buffer.rs
@@ -325,21 +325,23 @@ impl LogStoreBufferSender {
     pub(crate) async fn wait_for_barrier_truncation(
         &self,
         curr_epoch: u64,
-    ) -> ReaderTruncationOffsetType {
+    ) -> LogStoreResult<ReaderTruncationOffsetType> {
         loop {
             let notified = self.truncate_notify.notified();
 
             {
                 let mut inner = self.buffer.inner();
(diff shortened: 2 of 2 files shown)


src/stream/src/common/log_store_impl/kv_log_store/reader.rs
```diff
@@ -605,7 +606,7 @@ impl<S: StateStoreRead> LogReader for KvLogStoreReader<S> {
                         is_checkpoint,
                         new_vnode_bitmap: None,
                         is_stop: false,
-                        add_columns: None,
+                        add_columns,
```
BugenZhao, line 609: Just realize we have such a specialized field solely for the adding column case of schema change. 😯

src/stream/src/common/log_store_impl/kv_log_store/writer.rs
```diff
@@ -165,16 +162,29 @@ impl<LS: LocalStateStore> LogWriter for KvLogStoreWriter<LS> {
...
+            epoch,
+            options.is_checkpoint,
+            next_epoch,
+            options.add_columns,
+        );
+        if has_schema_change {
+            let truncate_offset = self.tx.wait_for_barrier_truncation(epoch).await?;
+            assert_eq!(truncate_offset, (epoch, None));
+            self.last_truncate_offset = Some(truncate_offset);
+        }
```
chenzl25, line 187: Just mark here, if the sink log store lag is too large e.g. 1 days, the schema change would potentially block the whole checkpoint procedure.

chenzl25 (approved):
LGTM, thanks!

src/stream/src/common/log_store_impl/kv_log_store/writer.rs
```diff
@@ -165,16 +162,29 @@ impl<LS: LocalStateStore> LogWriter for KvLogStoreWriter<LS> {
         }
         flush_info.report(&self.metrics);
 
-        let truncate_offset = self.tx.pop_truncation(epoch);
+        if let Some(truncate_offset) = self.tx.pop_truncation(epoch) {
```
wenym1, line 165: Previously we only write vnode watermark to storage when we pop some new progress. After this PR, we will write a same table watermark to storage when there is no new progress, though the impact may be small.

wenym1:
> It seems that the schema of the log store state table is not changed in this PR. It doesn't seem to work if we have some data flushed to log store storage.

Discussed offline. The schema has been updated, but existing tests haven't covered the case of having data flushed to log store storage. Should at least test locally by manually setting the size of in-memory buffer between writer and reader to 0.

PR #24054, opened 2025-12-09
feat(iceberg): do not commit large statistical to avoid persistent metadata becoming too large

wcy-fdu:

For iceberg exactly once sink, when pre commit metadata, we persist DataFile in meta store. The DataFile contain two statistical info, `lower_bounds` and `upper_bounds`.
```
struct DataFile {
    file_path: String,            
    file_format: FileFormat,       
    file_size_in_bytes: i64,       
    record_count: i64,             
    

    partition: HashMap<String, Literal>, 
    
    column_sizes: HashMap<i32, i64>,          
    value_counts: HashMap<i32, i64>,           
    null_value_counts: HashMap<i32, i64>,     
    nan_value_counts: HashMap<i32, i64>,    
    lower_bounds: HashMap<i32, Vec<u8>>,       // ⚠️ Minimum value (binary) for each column
    upper_bounds: HashMap<i32, Vec<u8>>,       // ⚠️ Maximum value (binary) for each column
    
    ...
}
```

For large fields, such as JSON, these two fields may be very large, resulting in excessively large metadata.

This pr first workaround this in RisingWave side to mitigate online issue by truncating these two lager field, and the corresponding modifications will be made to the Iceberg Rust SDK later.

## Checklist

## Documentation

<details>

</details>


src/connector/src/sink/iceberg/mod.rs
```diff
@@ -1688,6 +1688,64 @@ const SCHEMA_ID: &str = "schema_id";
...
+///
+/// This function removes `lower_bounds` and `upper_bounds` that exceed
+/// `MAX_COLUMN_STAT_SIZE` to prevent metadata bloat and OOM errors.
+///
+/// For large fields (JSONB, TEXT, BINARY), storing full min/max values provides
+/// little query optimization benefit but can cause metadata to balloon to gigabytes.
+///
+/// # Arguments
+/// * `json_value` - Mutable reference to the JSON representation of a `DataFile`
+fn truncate_large_column_stats(json_value: &mut serde_json::Value) {
```
chenzl25: I suspect this method fail to truncate the large column, because iceberg datafile doesn't serialize statistics in this struct.

src/connector/src/sink/iceberg/mod.rs
```diff
@@ -1688,6 +1690,73 @@ const SCHEMA_ID: &str = "schema_id";
...
+        }
+        true
+    });
+
+    // Process upper_bounds - remove entries with large values
+    data_file.upper_bounds.retain(|field_id, datum| {
+        // Get the size of the datum when serialized
+        let size = match serde_json::to_vec(datum) {
+            Ok(bytes) => bytes.len(),
+            Err(_) => 0,
```
chenzl25: Use `to_bytes` method of datum to avoid additional serialization overhead.

ci/docker-compose.yml
```diff
@@ -1,7 +1,7 @@
 services:
   # TODO: Rename this to `postgres`
   db:
-    image: postgres:15-alpine
+    image: postgres:18-alpine
```
chenzl25: Any reason we need to upgrade the postgres?

wcy-fdu: oh I accidentally brought over a commit from another branch.

PR #24079, opened 2025-12-10
chore(pgwire): add start of execution logs for all queries

kwannoel:

If `PGWIRE_QUERY_LOG` is enabled, we will also log at the start of query execution, not just when the query finishes execution. This allows us to review queries which encounter `OOM` or `abort` which can't be caught by the runtime. The result is that we will lose these queries that were executed.

A note on overhead: Currently, if `PGWIRE_QUERY_LOG=info` is set, we will _already_ emit at least one log for the query on its completion. We add one more log line for start of execution. If the overhead is too high for the user, they can always switch off query logging.

If the overhead is unacceptable, we could downgrade the log level to debug. But I prefer to keep it enabled, or else we have to wait for the event to occur.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/utils/pgwire/src/pg_protocol.rs
+++ b/src/utils/pgwire/src/pg_protocol.rs
@@ -391,6 +391,13 @@ where
 
         // Query log.
         let fut = async move {
+            if !tracing::Span::current().is_none() {
+                tracing::info!(
+                    target: PGWIRE_QUERY_LOG,
+                    status = "started",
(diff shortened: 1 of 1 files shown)


kwannoel:
Currently only extended query mode will log their parsing step. We miss some local query executions, which might still be heavy.

BugenZhao (approved):

PR #24082, opened 2025-12-10
[PATCH] feat: store per vnode max key in materialized cache for pruning

hzxa21:

Tested locally. I will open a PR for main later.

~~I will probably put the logic inside state table in the main PR.~~

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/executor/mview/materialize.rs
+++ b/src/stream/src/executor/mview/materialize.rs
@@ -323,7 +323,6 @@ impl<S: StateStore, SD: ValueRowSerde> MaterializeExecutor<S, SD> {
                         && b.has_more_downstream_fragments(self.actor_context.id)
                     {
                         self.may_have_downstream = true;
-                        self.materialize_cache.vnode_max_keys.clear();
                     }
                     Self::may_update_depended_subscriptions(
                         &mut self.depended_subscription_ids,


src/stream/src/executor/mview/materialize.rs
```diff
@@ -321,6 +323,7 @@ impl<S: StateStore, SD: ValueRowSerde> MaterializeExecutor<S, SD> {
                         && b.has_more_downstream_fragments(self.actor_context.id)
                     {
                         self.may_have_downstream = true;
+                        self.materialize_cache.vnode_max_keys.clear();
```
chenzl25: Why do we need to clear the cache here?

hzxa21: Originally I put `assert!(self.materialize_cache.vnode_max_keys.is_empty)` here but later I realized that there are cases that the conflict check will still be enabled if there is no downstream on the table so I change it to clear here.

After a second thought, I think we can just remove this line because
- When conflict check is enabled, `materialize_cache.handle` will be called and the `vnode_max_keys` will be set correctly.
- When conflict check is disabled, `vnode_max_keys` will be empty and get init in future `materialize_cache.handle`

Already push a new commit to remove it.

src/stream/src/executor/mview/materialize.rs
```diff
@@ -699,10 +703,43 @@ impl<SD: ValueRowSerde> MaterializeCache<SD> {
...
+        &mut self,
+        state_table: &StateTableInner<S, SD>,
+    ) -> StreamExecutorResult<()> {
+        assert!(self.vnode_max_keys.is_empty());
+        'vnode_loop: for vnode in state_table.vnodes().iter_vnodes() {
+            let table_iter = state_table
+                .rev_iter_keyed_row_with_vnode(
+                    vnode,
+                    &(Bound::<OwnedRow>::Unbounded, Bound::<OwnedRow>::Unbounded),
+                    PrefetchOptions::new(false, false),
```
Li0k, line 721: nits: This is not the subject of this PR. **PrefetchOptions** should be an Optional type. PrefetchOptions::new(false, false) is not intuitive, as it means prefetching is disabled. Alternatively, we could construct it via a dedicated function, such as Prefetch::without_prefetch()....

hzxa21: Good point. I will add `PrefetchOptions::disable()` in the main branch PR.

BugenZhao:
> I will probably put the logic inside state table in the main PR.

I guess this will be a better idea, as per-vnode min/max seems also to benefit other use cases of state accessing. Is it possible to implement it?

hzxa21:
> > I will probably put the logic inside state table in the main PR.
> 
> I guess this will be a better idea, as per-vnode min/max seems also to benefit other use cases of state accessing. Is it possible to implement it?

I would like to try in materialized first. The main reason I didn't implement it in state table is because there are some corner cases to consider for the full set of state table interfaces including iter, reverse iter. Also, I would like the executor to control whether this feature should be enabled. For example, there is no need to maintain vnode min/max in materialized executor that doesn't perform conflict check.

BugenZhao:
@codex review

PR #24113, opened 2025-12-15
fix(frontend): reject defining watermarks on upsert source

BugenZhao:

There is no clear definition of watermarks for upsert/retract streams, and our current impl doesn't expect this to happen as well.

Previously, a `SOURCE` had to be append-only, allowing us to apply the `WatermarkFilter` directly once a user defined a watermark, without any checks. However, since we introduced support for upsert sources in #22856, this assumption is no longer valid, and we need to implement this check.

This PR also add tests for the error handling of defining watermarks on non-append-only tables.

## Checklist

## Documentation

<details>

</details>


BugenZhao:

- #24113 <-- you are here 🤪
    - #24121

yuhao-su (approved):

PR #24121, opened 2025-12-15
feat(frontend): support planning table with watermark-ttl

BugenZhao:

This PR adds support for

- parsing `WATERMARK FOR .. WITH TTL`
- planning a table with watermark-ttl by specifying `clean_watermark_indices` and adding the watermark column to the stream key.

The executor part will be supported in subsequent PRs.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/e2e_test/batch/describe_fragments.slt
+++ b/e2e_test/batch/describe_fragments.slt
@@ -81,7 +81,7 @@ psql_validate.py --db $__DATABASE__ --sql "DESCRIBE FRAGMENTS describe_plan_test
 StreamMaterialize { columns: [id, name, age, created_at, data], stream_key: [id], pk_columns: [id], pk_conflict: Overwrite }
 ├── output: [ id, name, age, created_at, data ]
 ├── stream key: [ id ]
-└── StreamFilter { predicate: IsNotNull(name) } { output: [ id, name, age, created_at, data ], stream key: [] }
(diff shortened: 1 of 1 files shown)


BugenZhao:

- #24164
    - #24152
        - #24232
            - #24121 <-- you are here 🤪
                - #24186

src/frontend/src/optimizer/plan_node/stream_materialize.rs
```diff
@@ -341,7 +345,16 @@ impl StreamMaterialize {
...
+
+        // Add TTL watermark column to stream key.
+        // When a row comes in to a TTL-ed table and we cannot find it in the table, we still cannot tell
+        // whether it is a new row or an update to an expired row. Adding the TTL watermark column to the stream key
+        // can ensure there's no double-insert from the view of the downstream jobs. See RFC for more details.
+        for idx in ttl_watermark_indices.iter().copied() {
+            if !stream_key.contains(&idx) {
+                stream_key.push(idx);
+            }
+        }
```
chenzl25, line 357: We don't need to add the watermark column as the first column of stream_key, because we would support value-based watermark compaction, right?

chenzl25 (approved):
LGTM

src/frontend/src/optimizer/plan_node/stream_materialize.rs
BugenZhao: Yes. Also the storage won't see the stream key of an operator but only the primary key of the table.

PR #24041, opened 2025-12-08
feat(storage): support state clean for watermark in either pk or value column

zwang28:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## What's changed and what's your intention?

This PR add a new ValueSkipWatermarkIter to support state cleanup based on watermark in non-pk column.
- The exisiting PkPrefixSkipWatermarkIter and NonPkPrefixSkipWatermarkIter only handle watermark specified in PK column.
- ValueSkipWatermarkIter  can handle watermark-based state clean for watermark specified in either PK or non-PK column.
- However, ValueSkipWatermarkIter will only be used to handle non-pk column watermark, supplementing existing PkPrefixSkipWatermarkIter and NonPkPrefixSkipWatermarkIter. The justification is ValueSkipWatermarkIter uses basic value encoding for internal table, which deserializes all columns. It brings larger overhead than deserializing PK columns in the other 2 iterators.

## Checklist

- [ ] I have written necessary rustdoc comments.
- [ ]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  I have checked the [Release Timeline](https://github.com/risingwavelabs/rw-commits-history/blob/main/release_timeline.md) and [Currently Supported Versions](https://docs.risingwave.com/changelog/release-support-policy#support-end-dates-for-recent-releases) to determine which release branches I need to cherry-pick this PR into. 

## Documentation

- [ ]  My PR needs documentation updates. 

<details>
<summary><b>Release note</b></summary>

</details>

PR #24134, opened 2025-12-15
refactor(streaming): fix and refactor the watermark based state clean for band join

yuhao-su:

Refactor and fix band join with watermark-based state cleaning. 

### fix
Before this PR, hash join plan include output watermark from both side of the inequal condition. But it should only yield the watermark from the larger side.

### refactor

The new implementation assumed an inequality condition of a join should only include `InputRef` and a comparison condition. The offset (aka the arithmetic operation) will be pushed down by `PushCalculationOfJoinRule`. This PR uses InequalityInputPairV2 to simplify this implementation.

## Checklist

### DIFF
--- a/src/frontend/src/utils/condition.rs
+++ b/src/frontend/src/utils/condition.rs
@@ -30,9 +30,9 @@ use crate::TableCatalog;
 use crate::error::Result;
 use crate::expr::{
     ExprDisplay, ExprImpl, ExprMutator, ExprRewriter, ExprType, ExprVisitor, FunctionCall,
-    InequalityInputPair, InequalityInputPairV2, InputRef, collect_input_refs,
-    column_self_eq_eliminate, factorization_expr, fold_boolean_constant, push_down_not,
-    to_conjunctions, try_get_bool_constant,
+    InequalityInputPair, InputRef, collect_input_refs, column_self_eq_eliminate,
+    factorization_expr, fold_boolean_constant, push_down_not, to_conjunctions,
+    try_get_bool_constant,
 };
 use crate::utils::condition::cast_compare::{ResultForCmp, ResultForEq};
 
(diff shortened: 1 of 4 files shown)


yuhao-su:
* **#24144** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24144?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24134** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24134?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24134?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/stream/src/executor/hash_join.rs
```diff
@@ -776,38 +805,48 @@ impl<K: HashKey, S: StateStore, const T: JoinTypePrimitive, E: JoinEncoding>
...
-                }
-                self.inequality_watermarks[*inequality_index] = Some(selected_watermark);
+            }
+            // Do state cleaning by update_watermark on the larger side (after the loop to avoid borrow issues)
+            // TODO(yuhao-su): implement state cleaning
+            if let Some(_val) = update_left_watermark {
+                // self.side_l.ht.update_watermark(val);
+            }
+            if let Some(_val) = update_right_watermark {
+                // self.side_r.ht.update_watermark(val);
```
chenzl25, line 850: Do we need to resolve it in this PR?

yuhao-su: This should be resolved after #24041

Also we are facing another issue about this. The degree table does not contains the inequality column watermark. If we only update the watermark for join state table but not for degree table, when fetching remote state we will zip the state table and degree table, in current implementation it can cause a panic.

src/frontend/src/utils/condition.rs
```diff
@@ -258,7 +258,10 @@ impl Condition {
     ///
     /// The inequality conditions are transformed into `(left_col_id, right_col_id, offset)` pairs.
     ///
+    /// Deprecated: Use `extract_inequality_keys_v2` instead.
+    ///
     /// [`EqJoinPredicate`]: crate::optimizer::plan_node::EqJoinPredicate
+    #[allow(dead_code)]
     pub(crate) fn extract_inequality_keys(
```
chenzl25: Since we don't use `extract_inequality_keys` any more, do we need to remove these dead codes?

yuhao-su: Removed all the deadcode

PR #24152, opened 2025-12-16
feat(dml): add an option for emitting upsert records

BugenZhao:

Introduced a session config named `upsert_dml`. When enabled, `UPDATE` and `DELETE` will emit records in upsert format, behaving like a source with `FORMAT UPSERT`. This can be useful for testing how our systems handle upsert data without actually involving an external source with upsert format.

- `UPDATE` will only emit `Insert` records for new rows, instead of `Update` records which also contain the old row.
- `DELETE` will only include key columns and pad the rest with NULL, instead of emitting complete rows.

## Checklist

## Documentation

<details>

</details>


src/stream/src/executor/row_id_gen.rs
```diff
@@ -63,17 +65,38 @@ impl RowIdGenExecutor {
...
 
         for ((datum, op), vis) in column.iter().zip_eq_fast(ops).zip_eq_fast(vis.iter()) {
-            // Only refill row_id for insert operation.
-            match op {
-                Op::Insert => builder.append(Some(self.row_id_generator.next().into())),
-                _ => {
-                    if vis {
-                        builder.append(Some(Serial::try_from(datum.unwrap()).unwrap()))
-                    } else {
-                        builder.append(None)
```
BugenZhao: We only want to generate row id for `INSERT`s instead of `UPDATE`s and `DELETE`s. To achieve this, we determine it by `Op`. However, with upsert DML `UPDATE`s will also yield `Op::Insert`.

Thus, this PR changed the logic to skipping generating once there's a non-NULL value for row id column, which also seems more reasonable.

src/batch/executors/src/executor/update.rs
```diff
@@ -187,21 +181,31 @@ impl UpdateExecutor {
...
+                }
+                let chunk = if self.upsert {
+                    // In upsert mode, we only write the new row.
+                    builder.append_record(Record::Insert {
+                        new_row: row_insert,
+                    })
+                } else {
+                    // Note: we've banned updating the primary key when binding `UPDATE` statement.
+                    // So we can safely use `Update` op.
+                    builder.append_record(Record::Update {
```
BugenZhao, line 196: Use `append_record` to avoid constructing an interlaced `-` `+` op ourselves.

src/stream/src/executor/row_id_gen.rs
```diff
@@ -63,17 +65,38 @@ impl RowIdGenExecutor {
...
+            let row_id = if vis {
+                match datum {
+                    // If the row id is already present, e.g., from `UPDATE` or `DELETE`, keep it.
+                    Some(scalar) => Some(scalar.into_serial()),
+                    // Otherwise, generate a new row id only if it's an `Insert` operation.
+                    None => match op {
+                        Op::Insert => Some(self.row_id_generator.next().into()),
+                        Op::Delete => warn_absent_row_id!(op),
+                        Op::UpdateDelete => warn_absent_row_id!(op),
+                        Op::UpdateInsert => warn_absent_row_id!(op),
```
BugenZhao, line 106: Make them different arms to use different `LogSuppressor` instances.

BugenZhao:

- #24164
    - #24152 <-- you are here 🤪
        - #24121
            - #24186

src/stream/src/executor/row_id_gen.rs
BugenZhao: Split into #24164

chenzl25 (approved):
LGTM!

Issue #23923, opened 2025-11-28
refactor: unify calling render_actors and assign_split inside global barrier worker


wenym1:
### Is your feature request related to a problem? Please describe.

Currently, in some `Command`s scheduled to global barrier worker, including `CreateStreamingJob`, `ReplaceTable`, `RescheduleFragment`, `SourceSplitChange` and maybe others, before we schedule the command, we do `render_actors` and `assign_splits`, and include these actors related information in the command.

This current implementation has the following drawbacks. 

First, the information we used to generate these actor information might be stale. For example, we have many logics of loading runtime actor information from `SharedActorInfo`, generate command based on the loaded information, and then schedule the command. However, before the command is handled, the runtime actor information may have been changed by commands before the current command, and then the generated information in the command might be outdated, and there can be potential bug in some implicit corner case.

Second, inside global barrier worker, we already have these logic in recovery for recomposing actors. Besides, as we are supporting batch refresh running jobs, when the job is triggered to run, we will have to call `render_actors` and `assign_splits` inside global barrier worker. If we also have separate similar logics outside global barrier worker, this may increase the maintenance burden. 

In conclusion, we can reach a cleaner implementation if we can limit the call of `render_actors` and `assign_splits` only in global barrier worker.

### Describe the solution you'd like

In `Command` scheduled to global barrier worker, we don't include any runtime information related actors. Instead, we only include the logical information, such as fragment, fragment relations (including no shuffle fragment relations), and the list of splits.

Actors should not be aware outside global barrier worker, unless for displaying runtime information, such as dashboard, or system table. Our goal will be gradually deprecating the usage of `SharedActorInfo`, unless the usage is for displaying.

General workflow:
1. render actor location by a) the list of ground truth connected workers stored in `control_stream_manager` of global barrier worker, and b) no shuffle fragment relations either maintained in `InflightFragmentInfo` or loaded from outside global barrier worker, and c) vnode bitmap of existing actors storing in `InflightFragmentInfo`.
2. assign splits to render actors by a) the list of source splits and b) no shuffle fragment relations provided in the way as mentioned above in 1).
3. generate dispatcher and merge update by current `FragmentEdgeBuilder` and fragment relations provided as in the way as mentioned above in 1).

TODO:
- [ ] CreateStreamingJob: **highest priority** for logic to be reused for batch refresh running jobs.
- [ ] RescheduleFragment
- [ ] ReplaceTable
- [ ] SourceSplitChange
[...]

PR #24155, opened 2025-12-16
refactor: separate fragment metadata loading and actor assignment rendering

shanicky:

**Summary**
This PR refactors the actor assignment process during barrier recovery and scaling by introducing a clear two-stage approach: asynchronous loading of fragment metadata followed by synchronous rendering of actor-to-worker assignments. It replaces the previous monolithic function with well-defined, modular steps, improving clarity, maintainability, and extensibility.

**Details**
- Replaced the old `resolve_database_info` function with two separate functions: `load_fragment_context` (async metadata loading) and `render_actor_assignments` (sync assignment computation).
- Modularized the metadata retrieval and actor assignment logic to decouple data loading from computation, enabling better separation of concerns.
- Introduced new domain-specific types such as `LoadedFragmentContext`, `FragmentRenderMap`, `WorkerInfo`, and `RenderedGraph` to improve type clarity and reduce reliance on nested HashMaps.
- Integrated adaptive parallelism strategy support by reading system parameters, allowing more flexible and dynamic scaling decisions.
- Refactored related components (`recovery.rs`, `fragment.rs`, `scale.rs`) to consistently use the new two-stage approach, eliminating duplicated logic.
- Improved error handling and logging by isolating loading and rendering stages, making failures easier to diagnose.
- Used deterministic data structures (`BTreeMap`) and safe integer types (`NonZeroUsize`) to enhance robustness and predictability.

## Checklist

### DIFF
--- a/src/meta/src/barrier/context/recovery.rs
+++ b/src/meta/src/barrier/context/recovery.rs
@@ -38,7 +38,7 @@ use crate::barrier::DatabaseRuntimeInfoSnapshot;
 use crate::barrier::context::GlobalBarrierWorkerContextImpl;
 use crate::controller::fragment::{InflightActorInfo, InflightFragmentInfo};
 use crate::controller::scale::{
-    FragmentRenderMap, PreparedRenderContext, RenderedGraph, WorkerInfo, render_prepared,
+    FragmentRenderMap, LoadedFragmentContext, RenderedGraph, WorkerInfo, render_actor_assignments,
 };
 use crate::manager::ActiveStreamingWorkerNodes;
 use crate::model::{ActorId, FragmentId, StreamActor};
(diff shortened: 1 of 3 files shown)


src/meta/src/barrier/context/recovery.rs
```diff
@@ -635,14 +679,35 @@ impl GlobalBarrierWorkerContextImpl {
...
+            let system_params_reader = self
+                .metadata_manager
+                .catalog_controller
+                .env
+                .system_params_reader()
+                .await;
+            system_params_reader.adaptive_parallelism_strategy()
+        };
+
         let active_streaming_nodes =
```
wenym1, line 692: Can we try further refactor the code of database recovery, to only reload all necessary metadata in async context of `reload_database_runtime_info`, and after `reload_database_runtime_info` finishes we then render the actors and split assignments? In current implementation, there are still async method calls after `render_actor_assignments`, which are `recovery_table_with_upstream_sinks` and `load_stream_actors`. 

Though `recovery_table_with_upstream_sinks` takes `InflightDatabaseInfo` as input, it doesn't actually depend on having actor rendered, and it only modifies the fragment stream node of fragments in `InflightDatabaseInfo`.  `load_stream_actors` can be non-async if we already load `job_extra_info` when we reload it along with other metadata under async contexts.

shanicky: This is a basic change—just replacing the original render method with async load + non-async render. It doesn't involve any logic changes, and more updates will follow.

yezizp2012 (approved):
LGTM

hzxa21:
relevant #23923

PR #24161, opened 2025-12-17
refactor(iceberg): move iceberg countstar and pushdown from batch to logical

xxhZs:

The optimization plan we previously made for Iceberg was implemented at the batch phase, which prevented our rules from being used by DataFusion. With this PR, we have moved the optimization to the logical phase.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/src/source/iceberg/mod.rs
+++ b/src/connector/src/source/iceberg/mod.rs
@@ -146,141 +146,143 @@ impl IcebergFileScanTaskJsonStr {
     }
 }
 
-#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
-pub enum IcebergFileScanTask {
-    Data(Vec<FileScanTask>),
-    EqualityDelete(Vec<FileScanTask>),
-    PositionDelete(Vec<FileScanTask>),
-    CountStar(u64),
-}
+// #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+// pub enum IcebergFileScanTask {
+//     Data(Vec<FileScanTask>),
+//     EqualityDelete(Vec<FileScanTask>),
+//     PositionDelete(Vec<FileScanTask>),
+//     CountStar(u64),
+// }
 
-impl Hash for IcebergFileScanTask {
-    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
-        match self {
-            IcebergFileScanTask::Data(file_scan_tasks) => {
-                file_scan_tasks.iter().for_each(|task| {
-                    task.data_file_path.hash(state);
-                });
-            }
-            IcebergFileScanTask::EqualityDelete(file_scan_tasks) => {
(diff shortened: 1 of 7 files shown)


src/frontend/src/optimizer/plan_node/logical_iceberg_scan.rs
```diff
@@ -34,14 +37,46 @@ use crate::optimizer::plan_node::{
 use crate::utils::{ColIndexMapping, Condition};
 
 /// `LogicalIcebergScan` is only used by batch queries. At the beginning of the batch query optimization, `LogicalSource` with a iceberg property would be converted into a `LogicalIcebergScan`.
-#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[derive(Debug, Clone)]
 pub struct LogicalIcebergScan {
     pub base: PlanBase<Logical>,
     pub core: generic::Source,
     pub iceberg_scan_type: IcebergScanType,
+    pub predicate: IcebergPredicate,
```
chenzl25: I think we should use `predicate: Condition` instead of `predicate: IcebergPredicate` in LogicalIcebergScan. Only convert it to IcebergPredicate when we need to convert LogicalIcebergScan to BatchIcebergScan.

src/frontend/src/optimizer/plan_node/logical_iceberg_scan.rs
```diff
@@ -25,23 +30,48 @@ use super::{
...
+impl std::hash::Hash for LogicalIcebergScan {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        self.base.hash(state);
+        self.core.hash(state);
+        self.iceberg_scan_type.hash(state);
+        self.snapshot_id.hash(state);
+        self.predicate.hash(state);
+    }
+}
+
```
chenzl25: I think we shouldn't impl PartialEq and Hash by ourselves.

src/frontend/src/optimizer/plan_node/logical_iceberg_scan.rs
```diff
@@ -141,9 +211,15 @@ impl PredicatePushdown for LogicalIcebergScan {
 
 impl ToBatch for LogicalIcebergScan {
     fn to_batch(&self) -> Result<crate::optimizer::plan_node::BatchPlanRef> {
-        let plan =
-            BatchIcebergScan::new(self.core.clone(), self.iceberg_scan_type, self.snapshot_id)
-                .into();
+        let iceberg_predicate =
+            rw_predicate_to_iceberg_predicate(self.predicate.clone(), self.schema().fields())?;
```
chenzl25: If failed to convert it to iceberg predicate, we should keep it as a filter on top of iceberg scan instead of throwing error.

chenzl25: This logic should be the same as previous predicate pushdown rule.

PR #24163, opened 2025-12-17
feat(streaming): always enable merge barrier align metrics

kwannoel:

In high parallelism cases, exchange can become an overhead, and lead to poor performance. What we're interested in is the correlation between this metric and high barrier latency.

Currently we use histogram buckets to measure this, but it leads to imprecise readings, and doesn't lend itself well to `rate` computations (i.e. we can't answer the question precisely: what % of time is merge processing barriers?).

We can just use a counter + rate instead to measure the alignment duration, and be consistent with other alignment metrics, e.g. executor alignment.

Further, now that we have label pruning, we will also remove actor id from this metric label. It's safe to keep it always on, since the cardinality is at the fragment level.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/.github/workflows/bump-main.yml
+++ b/.github/workflows/bump-main.yml
@@ -12,7 +12,7 @@ jobs:
     permissions:
       contents: write
       pull-requests: write
-    
+
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


kwannoel:
https://buildkite.com/risingwavelabs/docker/builds/18012

kwannoel:
Looks good to me. 64 x 10 parallelism, 64 fragments. Largest number of channels between fragments would be `(64 x 10) ^ 2`.
Screenshot 2025-12-18 at 1 32 32 PM

kwannoel:
Can be used as a bottom-up signal for merge bottleneck.

BugenZhao (commented):
LGTM

BugenZhao (approved):

PR #24164, opened 2025-12-17
refactor(streaming): keep & validate row id if specified regardless of op in `RowIdGen`

BugenZhao:
Signed-off-by: Bugen Zhao <i@bugenzhao.com>

We only want to generate row id for `INSERT`s instead of `UPDATE`s and `DELETE`s. To achieve this, we determine it by `Op`. However, with upsert DML (in next PR, #24152) `UPDATE`s will also yield `Op::Insert`.

Thus, this PR changed the logic to skipping generating once there's a non-NULL value for row id column, which also seems more reasonable. Also, we will check if it contains a correct vnode value.

## Checklist

## Documentation

<details>

</details>

## Checklist

## Documentation

<details>

</details>


BugenZhao:

- #24164 <-- you are here 🤪
    - #24152
        - #24121
            - #24186

src/tests/simulation/tests/integration_tests/sink/utils.rs
```diff
@@ -699,13 +697,14 @@ pub fn build_stream_chunk(
         100000,
     );
     for (id, name, split_id, offset) in row_iter {
-        let row_id = ROW_ID_GEN.fetch_add(1, Relaxed);
```
BugenZhao, line 702: The previously generated row id here does not conform to the expected distribution.

shanicky (approved):

PR #24176, opened 2025-12-18
feat: vnode key range info

Li0k:

### Summary
Implements vnode-based SST filtering to reduce unnecessary SST accesses during GET/ITER operations in Hummock storage engine.

### Changes
1. **SST Metadata Enhancement**
   - Added `vnode_key_ranges` field to `SstableInfo` for storing per-vnode key range hints
   - Enabled by config: `meta.compaction_config.max_vnode_key_range_bytes` (default: None)

2. **Query-Time Pruning**
   - GET operations: Skip SSTs where queried vnode is out of range
   - ITER operations: Skip SSTs where query range doesn't overlap with vnode range

3. **Observability**
   - Added metrics: `state_store_vnode_pruning_counts{table_id, operation, result}` (Debug level)
   - Added Grafana panels: "Vnode Pruning Ops" and "Vnode Pruning Rate"

image

image

### Configuration
```toml
[meta.compaction_config]
max_vnode_key_range_bytes = 33554432 
```

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/storage/src/compaction_catalog_manager.rs
+++ b/src/storage/src/compaction_catalog_manager.rs
@@ -221,7 +221,7 @@ impl FilterKeyExtractor for MultiFilterKeyExtractor {
             return user_key;
         }
 
-        let table_id = get_table_id(user_key).into();
+        let table_id = get_table_id(user_key);
         self.id_to_filter_key_extractor
             .get(&table_id)
             .unwrap()


hzxa21:
> Failed: 1
Passed:1
FailedExecutions:
https://buildkite.com/risingwave-test/sysbench-cdc/builds/1660

Please check the failed test.

proto/hummock.proto
```diff
@@ -15,6 +15,20 @@ enum BloomFilterType {
...
+message VnodeKeyRange {
+  uint32 vnode = 1;
+  bytes min_key = 2;
+  bytes max_key = 3;
+}
+
+// Optional per-vnode statistics for SSTs with few vnodes.
+// Enables more precise query pruning by providing vnode-level key ranges.
+message VnodeStatistics {
+  repeated VnodeKeyRange vnode_key_ranges = 1;
```
hzxa21: How about using `map` instead of `repeated`?

src/ctl/src/cmd_impl/hummock/sst_dump.rs
```diff
@@ -226,6 +226,7 @@ pub async fn sst_dump_via_sstable_store(
         range_tombstone_count: 0,
         bloom_filter_kind: Default::default(),
         sst_size: 0,
+        vnode_statistics: None,
```
hzxa21, line 238: It will be helpful to print out the vnode stats in `print_level`

src/storage/hummock_sdk/src/sstable_info.rs
```diff
@@ -67,6 +87,53 @@ impl SstableInfoInner {
...
 
+impl From<&PbVnodeStatistics> for VnodeStatistics {
+    fn from(info: &PbVnodeStatistics) -> Self {
+        Self {
+            vnode_key_ranges: info
+                .vnode_key_ranges
+                .iter()
+                .map(|range| {
+                    let min_key = FullKey::decode(&range.min_key).copy_into();
+                    let max_key = FullKey::decode(&range.max_key).copy_into();
```
hzxa21: Better add an assertion either here or in meta or both to make sure
1. min_key and max_key share the same table id
2. min_key and max_key share the same vnode

src/storage/src/hummock/compactor/compaction_utils.rs
```diff
@@ -139,6 +139,7 @@ pub struct TaskConfig {
     pub table_schemas: HashMap<TableId, PbTableSchema>,
     /// `disable_drop_column_optimization` should only be set in benchmark.
     pub disable_drop_column_optimization: bool,
+    pub max_vnode_key_range_bytes: Option<usize>,
```
hzxa21: This field is unused?

src/storage/src/hummock/sstable/builder.rs
```diff
@@ -676,6 +709,120 @@ impl<W: SstableWriter, F: FilterBuilder> SstableBuilder<W, F> {
...
+    ///
+    /// # Arguments
+    /// * `vnode` - The vnode of the current key
+    /// * `key` - The current key (used as left boundary when a new vnode starts)
+    /// * `prev_key` - The previous key (used as right boundary when vnode switches)
+    ///
+    /// # Semantics
+    /// Implements "allow over-limit write": inserting a range may cause `current_size` to exceed
+    /// `max_bytes`, but stops before starting the next range if it would exceed the limit.
+    fn observe_key(&mut self, vnode: VirtualNode, key: &[u8], prev_key: &[u8]) {
```
hzxa21, line 764: This is a good place to add the assertion to make sure
1. vnode is monotonically increasing
2. keys shared the same vnode
3. keys shared the same table_id

Li0k: I have added a sanity-check for vnode and table-id. However, regarding point 1, I would prefer to rely on the upper-layer **FullKeyTracker** instead of implementing another similar set of logic. Given that the collector is only used in the current scenario, could we allow this "dependency" to minimize the Track overhead as much as possible?

Li0k:
cc @hzxa21

proto/hummock.proto
```diff
@@ -496,6 +513,8 @@ message CompactTask {
   // The maximum key-value count threshold for using Xor16 filter.
   // When the number of keys exceeds this threshold, a BlockedXor16 filter will be used.
   optional uint64 max_kv_count_for_xor16 = 28;
+  // Max bytes for vnode-level key range hints to embed into output SSTs.
+  optional uint64 max_vnode_key_range_bytes = 29;
```
hzxa21, line 517: Will this break serverless compaction controller? or will it just make this feature disabled in serverless compaction controller?

src/config/ci.toml
```diff
@@ -6,6 +6,7 @@ max_heartbeat_interval_secs = 60
 level0_tier_compact_file_number = 6
 level0_overlapping_sub_level_compact_level_count = 3
 level0_max_compact_file_number = 96
+max_vnode_key_range_bytes = 33554432
```
hzxa21: I suggest we set this as large as possible to make sure this feature is triggered in CI. Also, we should also at least manually trigger longevity and compaction test using this PR with the largest possible config for this feature to make sure we have cover this optimization before merging this PR.

src/storage/hummock_sdk/src/sstable_info.rs
```diff
@@ -12,17 +12,31 @@
...
 use crate::key_range::KeyRange;
 use crate::version::{ObjectIdReader, SstableIdReader};
 use crate::{HummockSstableId, HummockSstableObjectId};
 
+#[derive(Debug, PartialEq, Clone, Default)]
+pub struct VnodeStatistics {
+    /// Map from vnode to its full key range (`min_full_key`, `max_full_key`) in this SST.
+    /// All ranges use closed intervals [`min_key`, `max_key`].
+    /// Keys are pre-decoded `FullKey` to allow direct access to `user_key` field.
+    vnode_key_ranges: BTreeMap<VirtualNode, (FullKey<Bytes>, FullKey<Bytes>)>,
```
hzxa21: If only user key is needed, we can reduce the memory footprint by only persisting the user key instead of full key.

src/storage/hummock_sdk/src/compact_task.rs
```diff
@@ -237,6 +243,14 @@ impl CompactTask {
 
         crate::filter_utils::is_kv_count_too_large_for_xor16(kv_count, self.max_kv_count_for_xor16)
     }
+
+    /// Returns the effective vnode key-range hint limit (in bytes) for this compaction task.
+    ///
+    /// The hint is only meaningful when all input SSTs belong to a single table.
+    pub fn effective_vnode_key_range_limit(&self) -> Option<usize> {
```
hzxa21: ditto: rename `effective_vnode_key_range_limit` to `effective_max_vnode_key_range_bytes` to align with the variable naming so that it is more friendly to code search.

src/storage/src/hummock/sstable/builder.rs
```diff
@@ -439,6 +457,14 @@ impl<W: SstableWriter, F: FilterBuilder> SstableBuilder<W, F> {
         let largest_key = self.last_full_key.clone();
         self.finalize_last_table_stats();
 
+        // Vnode key-range hints are only supported for single-table SSTs.
+        // Multi-table SST scenarios should not enable max_vnode_key_range_bytes in config.
+        debug_assert!(
```
hzxa21: Let's make this an assert instead of debug_assert because if the assertion fails, incorrect metadata will be generated

src/storage/src/hummock/store/version.rs
```diff
@@ -761,19 +762,43 @@ impl HummockVersionReader {
...
+                            .key_range
+                            .compare_right_with_user_key(full_key.user_key.as_ref());
+                        // the case that the key falls into the gap between two ssts
+                        if ord == Ordering::Less {
+                            sync_point!("HUMMOCK_V2::GET::SKIP_BY_NO_FILE");
+                            continue;
+                        }
+
+                        // filter vnode-key range that is definitely not containing the key
+                        if let Some(vnode_statistics) = &sstable_info.vnode_statistics {
```
hzxa21: Why don't we apply similar logic for overlapping level?

src/storage/src/hummock/store/version.rs
```diff
@@ -1034,9 +1059,58 @@ impl HummockVersionReader {
                     );
                     local_stats.non_overlapping_iter_count += 1;
                 } else {
+                    let sstable_info = &sstable_infos[0];
+
+                    // Filter by vnode key range: skip if query range does not overlap with vnode range
+                    if let Some(vnode_statistics) = &sstable_info.vnode_statistics
+                        && let (Included(start_key) | Excluded(start_key), _) = &user_key_range_ref
```
hzxa21: Need to make sure the user key range belongs to one vnode instead of assuming it and extracting vnode only from the start key

Li0k: ~This seems a bit redundant? We’ve added an assert to the deserialization code, which makes things more consistent.~

Li0k: remove the filter logic of iter first

proto/hummock.proto
Li0k: This is a break change, but we can address this issue by relying on proto compatibility as before. I will also be pushing for updates on the cloud side.

cc @maxCodeVector

src/storage/src/hummock/sstable/builder.rs
```diff
@@ -392,9 +402,9 @@ impl<W: SstableWriter, F: FilterBuilder> SstableBuilder<W, F> {
             });
         }
 
-        let table_id = full_key.user_key.table_id;
-        let mut extract_key = user_key(&self.raw_key);
-        extract_key = self.compaction_catalog_agent_ref.extract(extract_key);
+        let extract_key = self
+            .compaction_catalog_agent_ref
+            .extract(user_key(&self.raw_key));
```
hzxa21, line 407: This is probably just a naming issue instead a real issue.

I saw that the many `extract` methods in `compaction_catalog_manager.rs` uses `full_key` as its parameter name while we are passing `user_key` here. Can you help check whether it is a real issue? If not, let's fixing the parameter names.

Example:
https://github.com/risingwavelabs/risingwave/blob/cb6bcae9fc14e5a8da250455285c698285446e22/src/storage/src/compaction_catalog_manager.rs#L500

src/storage/src/hummock/store/version.rs
```diff
@@ -715,6 +716,21 @@ impl HummockVersionReader {
...
+                                vnode_statistics.get_vnode_user_key_range(vnode)
+                            {
+                                local_stats.vnode_checked_get_count += 1;
+                                let user_key = full_key.user_key.as_ref();
+                                if user_key < vnode_min.as_ref() || user_key > vnode_max.as_ref() {
+                                    local_stats.vnode_pruned_get_count += 1;
+                                    continue;
+                                }
+                            }
+                        }
```
hzxa21: nits: the logic here is duplicated twice. Let's extract it into a method reduce code duplication.

hzxa21 (approved):

PR #24186, opened 2025-12-18
feat(streaming): executor part for watermark TTL

BugenZhao:

This PR is the final step for the user-facing part (in contrast of the storage state cleaning part) of watermark TTL. Please find the comprehensive test case added for the desired behavior.

Major changes:

- Filter out expired row values when maintaining `MaterializeCache` for conflict handling
- Transform `Update` into `Delete` and `Insert` pairs when outputting from `MaterializeCache`, as we add the extra watermark column to the stream key
- Enable upsert mode for `WatermarkFilter` executor, basically forward impl to `UpsertFilter` executor to achieve the behavior described in RFC

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/meta/src/stream/test_fragmenter.rs
+++ b/src/meta/src/stream/test_fragmenter.rs
@@ -366,6 +366,7 @@ fn make_stream_fragments() -> Vec<StreamFragment> {
             column_orders: vec![make_column_order(1), make_column_order(2)],
             staging_table: None,
             refresh_progress_table: None,
+            cleaned_by_ttl_watermark: false,
         }))),
         fields: vec![], // TODO: fill this later
         operator_id: 7.into(),


BugenZhao:

- #24164
    - #24152
        - #24232
            - #24266
                - #24186 <-- you are here 🤪

e2e_test/streaming/watermark_ttl.slt
```diff
@@ -11,18 +11,287 @@ SET streaming_parallelism TO 1;
...
+2 20 b 2
+4 30 d 1
+
+query IIV
+select * from ttl_mv order by k, ts;
+----
+1 5 a
+1 18 aa
+4 30 d
+
```
chenzl25, line 294: What's the expected behavior, if I run `delete from ttl` at the end. Shall we see rows in the `ttl`?

BugenZhao: I think we won't see any rows. No matter whether expired rows are cleaned by the storage, we issue `DELETE`s on rows that are visible from a `BatchScan`, so all visible rows will be deleted.

chenzl25: Then, it seems like we have a bug, because right now, I still saw rows below watermark. I think it is caused by the materialized cache, if we put a None to the `lru_cache` if the row is below watermark.

BugenZhao: Ah. `DELETE`s on rows below watermark will be filtered out in `WatermarkFilter` so it won't actually come to `Materialize`.

I think this will be resolved after #24041, where we will filter out expired rows when scanning a table, then everything will become consistent.

chenzl25: Not yet. We use upsert filter, so the delete should pass to the materialize executor.

chenzl25: I guess we hit this branch. Because the lru_cache result of the expired row is None. https://github.com/risingwavelabs/risingwave/blob/0f2b732c09f70a8931743ee7d9ecf8b95a26073c/src/stream/src/executor/mview/cache.rs#L334-L335

BugenZhao: > I guess we hit this branch. Because the lru_cache result of the expired row is None.

True. I think it's okay and we can let #24041 handle it.

chenzl25: Emm. But if we use the upsert filter, then we expect the delete to take effect? IIUC, https://github.com/risingwavelabs/risingwave/pull/24041 cleanup is asynchronous.

BugenZhao: The cleanup is indeed async, but I suppose there will also be a filtering process when scanning the table, so that no matter whether the cleanup has been done, expired data won't be visible? 

cc @zwang28

chenzl25: Does it mean we can just use a simple filter in the upsert watermark filter instead of upsert filter for the table with ttl watermark? If we don't care about the actual deletion of the rows below watermark.

BugenZhao: I don't think so. :( If the input is from an upsert source where columns other than key (including the event time column) will always be `NULL`, using a simple filter will filter out all `DELETE`s, even if they should be applied to rows above the watermark.

chenzl25: Another possible implementation is always using delete to overwrite the table's rows below watermark.

BugenZhao: True but personally I don't find it very necessary. Basically TTL is a best-effort method to free up storage by removing outdated data, rather than a reliable way to ensure data management. Not to mention that when there's multiple parallelism and scaling, the watermark value on each parallelism is indeterministic as well. Users should treat all operations on expired data as undefined behavior, and should not rely on whether it can be cleaned up or not.

chenzl25 (approved):
LGTM! As a user facing feature, we'd better describe the delete operation behavior against the rows below watermark for the watermark TTL feature in the doc on the next release.

e2e_test/streaming/watermark_ttl.slt
zwang28: > The cleanup is indeed async, but I suppose there will also be a filtering process when scanning the table, so that no matter whether the cleanup has been done, expired data won't be visible?

No. There's pruning for watermark defined on [first column of PK](https://github.com/risingwavelabs/risingwave/blob/643f80bf54d744e7d1e2b17ca936c528d1c9429d/src/storage/src/hummock/store/version.rs#L546). There's no such pruning for watermark defined on other columns.

cc @Li0k 

https://risingwave-labs.slack.com/archives/C0A1A6HT7UL/p1765945824684999?thread_ts=1765874320.135629&cid=C0A1A6HT7UL

PR #24189, opened 2025-12-18
feat(meta): create new actors for finished snapshot backfill jobs

wenym1:

Previously, after a snapshot backfill finishes, we move the existing running actors to the database partial graph. Under this implementation, the partial graph that an actor belongs to might be changed, which increases the complexity to handle actors. 

Besides, there is a bug in current implementation. We should ensure that for a state table, we should call `start_epoch` on the its `table_id` before we send barrier to actors that own the state table, so as to ensure that `start_epoch` won't be called later than `local_seal_epoch`. However, for the barrier that merge the snapshot backfill job back to the database partial graph, the barrier is injected to the partial graphs of database and snapshot backfill job separately. The `start_epoch` is called when handling the barrier of database partial graph, while the barrier is sent to actors when handling the barrier of snapshot backfill job partial graph. But in current implementation, the barrier is injected to snapshot backfill job partial graph before database partial graph, and then `local_seal_epoch` can be called before `start_epoch`. This bug is discovered by https://github.com/risingwavelabs/risingwave/pull/24116 under a streaming job where barrier passes through very quickly.

In this PR, we change to create new actors on database partial graph and stop the running actors in the snapshot backfill partial graph. In this way, the two problems above can be both resolved.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/meta/src/barrier/checkpoint/creating_job/mod.rs
+++ b/src/meta/src/barrier/checkpoint/creating_job/mod.rs
@@ -407,6 +407,9 @@ impl CreatingStreamingJobControl {
 
         let mut upstream_fragment_downstreams: FragmentDownstreamRelation = Default::default();
         for (upstream_fragment_id, downstreams) in fragment_relations {
+            if fragment_infos.contains_key(upstream_fragment_id) {
+                continue;
+            }
             for downstream in downstreams {
                 if fragment_infos.contains_key(&downstream.downstream_fragment_id) {
                     upstream_fragment_downstreams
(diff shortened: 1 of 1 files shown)


wenym1:
* **#24116** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24116?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24302** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24302?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24286** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24286?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24239** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24239?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24189** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24189?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24189?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

yezizp2012 (approved):
LGTM

PR #24213, opened 2025-12-22
fix(sink): fix snowfalke starrocks redshift executor_id to uuid

xxhZs:

https://github.com/risingwavelabs/risingwave/issues/23977

As stated in the issue, our exec IDs are prone to duplication, which may lead to data overwriting in StarRocks, Snowflake and Redshift. Therefore, this PR uses UUIDs to ensure the uniqueness of these IDs.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/src/sink/snowflake_redshift/mod.rs
+++ b/src/connector/src/sink/snowflake_redshift/mod.rs
@@ -21,7 +21,6 @@ use opendal::Operator;
 use risingwave_common::array::{ArrayImpl, DataChunk, Op, PrimitiveArray, StreamChunk, Utf8Array};
 use risingwave_common::catalog::Schema;
 use risingwave_common::row::Row;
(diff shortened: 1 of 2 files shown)


hzxa21:
Also, please provide more contexts in the PR description.

xxhZs:
> Is this change backward compatible? I know it is backward compatible for starrocks because it only affects the new transaction but I am not sure whether it is backward compatible for snowflake/redshift sink.
> 
> IMO, for existing sinks, the intermediate S3 location can contain both files with the old name and the files with the new name. Will it be a problem?

This will have no impact. Because in Snowflake, we fetch files based on the incremental data of the pipe, which is irrelevant to the file names. In Redshift, the commit operation only fetches a checkpoint file, which ensures that all these files use either the new or old file names.

PR #24214, opened 2025-12-22
feat: maintain per vnode min/max in state table for pruning

hzxa21:

Re-implement #24082 in main by maintaining the vnode min/max key in the state table layer.

The main motivation of this PR is to avoid unnecessary accesses to hummock when the workload has some increasing/decreasing property in their state table key.

## Checklist

## Documentation

<details>

</details>


BugenZhao:
May I know the relationship between this PR and #24176 ?

src/stream/src/common/table/state_table.rs
```diff
@@ -265,9 +334,74 @@ struct StateTableRowStore<LS: LocalStateStore, SD: ValueRowSerde> {
...
+    vnode_stats: Option<HashMap<VirtualNode, VnodeStatistics>>,
+    // Optional metrics for state table operations
+    pub metrics: Option<StateTableMetrics>,
 }
 
 impl<LS: LocalStateStore, SD: ValueRowSerde> StateTableRowStore<LS, SD> {
+    async fn may_load_vnode_stats(&mut self, vnode_bitmap: &Bitmap) -> StreamExecutorResult<()> {
+        if self.vnode_stats.is_none() {
+            return Ok(());
+        }
```
wenym1, line 360:
```suggestion
        let Some(stats_map) =  &mut self.vnode_stats {
            return Ok(());
        }
```

src/stream/src/common/table/state_table.rs
```diff
@@ -958,10 +1122,25 @@ impl<LS: LocalStateStore, SD: ValueRowSerde> StateTableRowStore<LS, SD> {
         key_bytes: TableKey<Bytes>,
         prefix_hint: Option<Bytes>,
     ) -> StreamExecutorResult<Option<OwnedRow>> {
+        if let Some(m) = &self.metrics {
+            m.get_count.inc();
+        }
+        let (vnode, key) = key_bytes.split_vnode_bytes();
```
wenym1: This `split_vnode_bytes` seems unnecessary when both `all_rows` and `vnode_stats` are `None`. 

Can use `LazyCell` to lazily call `split_vnode_bytes`.
```
let split_key_bytes = LazyCell::new(|| key_bytes.split_vnode_bytes());

// later when using it
let (vnode, key) = &*split_key_bytes;
or 
if let Some(stats) = &self.vnode_stats
    && let (vnode, key) = &*split_key_bytes
    && let Some(vnode_stat) = stats.get(&vnode)
    && vnode_stat.can_prune(&key)
```

src/stream/src/common/table/state_table.rs
```diff
@@ -265,9 +334,74 @@ struct StateTableRowStore<LS: LocalStateStore, SD: ValueRowSerde> {
...
+    pub metrics: Option<StateTableMetrics>,
 }
 
 impl<LS: LocalStateStore, SD: ValueRowSerde> StateTableRowStore<LS, SD> {
+    async fn may_load_vnode_stats(&mut self, vnode_bitmap: &Bitmap) -> StreamExecutorResult<()> {
+        if self.vnode_stats.is_none() {
+            return Ok(());
+        }
+
+        // No need to maintain vnode_stats if all_rows are present
```
wenym1: We may explicitly disable `enable_vnode_stats` in state table builder when `preload_all_rows` is enabled when we build the state table, and then we don't need to check `all_rows` here and make the two concepts independent to each other.

Li0k: If these two concepts are independent of each other, we should ensure that they cannot be enabled simultaneously, rather than implicitly ignoring one of them.

src/stream/src/common/table/state_table.rs
```diff
@@ -903,10 +1126,25 @@ impl<LS: LocalStateStore, SD: ValueRowSerde> StateTableRowStore<LS, SD> {
...
+        if let Some(m) = &self.metrics {
+            m.get_count.inc();
+        }
+        let (vnode, key) = key_bytes.split_vnode_bytes();
         if let Some(rows) = &self.all_rows {
-            let (vnode, key) = key_bytes.split_vnode();
-            return Ok(rows.get(&vnode).expect("covered vnode").get(key).cloned());
+            return Ok(rows.get(&vnode).expect("covered vnode").get(&key).cloned());
+        }
+
```
Li0k, line 1142: If get_count represents the number of get method calls, we seem to lack a metric to determine whether the vnode has been hit. The decision is up to you.

src/stream/src/common/table/state_table.rs
```diff
@@ -469,12 +657,27 @@ impl<'a, S: StateStore, SD: ValueRowSerde, const IS_REPLICATED: bool, PreloadAll
         self.output_column_ids = Some(output_column_ids);
         self
     }
+
+    pub fn enable_vnode_stats(mut self, enable: bool) -> Self {
```
BugenZhao: nit: `stats` sounds like an observability feature to me. 🤣

hzxa21:
> May I know the relationship between this PR and #24176 ?

The two PRs complement each other.

Let's say the user workload is:
- Data are coming in logical batch of `N`.
- Within each batch, there is no monotonicity in the keys.
- Across different batch, there is monotonicity in the keys.

1. When N = 1, all incoming keys are monotonically increasing/decreasing
2. When N = +inf, the workload is random,
3. N is between 1 and 2. This is more practical when append-only table is not used because if the keys are monotonically increasing/decreasing. Defining PK is unnecessary.

This PR simply maintain a single min/max key for each vnode on the state table layer. The stat is mostly useful only once for each logical batch. In other words, in most cases, only very few keys inside the `N` batch can be pruned by the state table per vnode stat.

However, combining with #24176, which maintains per vnode min/max in SSTs in a layered LSM, keys that cannot be pruned by the state table vnode stat can potentially be pruned by the SST vnode stat because data is organized in layers with time order. For example, the vnode stat in L6's SSTs roughly represents the per min/max key in a past snapshot.

src/stream/src/common/table/state_table.rs
```diff
@@ -77,10 +87,82 @@ macro_rules! insane_mode_discard_point {
...
+        }
+        if let Some(max) = &self.max_key
+            && key > max
+        {
+            return true;
+        }
+        false
+    }
+
+    fn can_prune_range(&self, start: &Bound<Bytes>, end: &Bound<Bytes>) -> bool {
```
BugenZhao, line 129: I think besides pruning the entire range, we can also "optimize" the given range by clamping the bounds?

src/stream/src/common/table/state_table.rs
hzxa21: I think it will be too finer grain if we have a metric representing per vnode count. We didn't have such metric before either. Do you think it is useful?

src/stream/src/common/table/state_table.rs
hzxa21: We will create a new stats_map instead of clearing and reusing the original ones on `may_load_vnode_stat`

src/stream/src/common/table/state_table.rs
hzxa21: I have changed the builder method to `enable_vnode_key_pruning` while keeping the `vnode_stats` name inside state table

src/stream/src/common/table/state_table.rs
hzxa21: Given that the `split_vnode_bytes` is just a one-line call, I think it is more straight-forward to just duplicate the call.

Fix that in the latest commit.

src/stream/src/common/table/state_table.rs
hzxa21: Good idea. Done.

hzxa21:
The `intergration-test-deterministic-simulation` always failed with
```
panicked at src/tests/simulation/tests/integration_tests/scale/nexmark_source.rs:101:5:[0m
    [31;1mexpected event num in 99000..=100000, got 98964
```
when vnode pruning is enabled.

Need some time to investigate and make sure this is caused by this flaky test instead of a bug in this PR before merging.

src/stream/src/common/table/state_table.rs
```diff
@@ -534,13 +732,16 @@ where
...
 
     /// Create state table from table catalog and store.
+    #[allow(clippy::too_many_arguments)]
     async fn from_table_catalog_inner(
         table_catalog: &Table,
         store: S,
         vnodes: Option<Arc<Bitmap>>,
         op_consistency_level: StateTableOpConsistencyLevel,
         output_column_ids: Vec<ColumnId>,
         preload_all_rows: bool,
```
Li0k, line 744: Consider encapsulating the functional/optimization parameters for future extension, while avoiding the too_many_arguments issue.

```rust
 let should_enable_vnode_key_pruning = if preload_all_rows
            && let Some(enable_vnode_key_pruning) = self.enable_vnode_key_pruning
            && enable_vnode_key_pruning
        {
            false
        } else {
            self.enable_vnode_key_pruning.unwrap_or(false)
        };
```
Further to this, the judgment logic can be encapsulated into a function. However, this depends on your coding style, I have no preference either way.

src/stream/src/common/table/state_table.rs
Li0k: make sense

src/stream/src/executor/mview/materialize.rs
```diff
@@ -257,10 +257,17 @@ impl<S: StateStore, SD: ValueRowSerde> MaterializeExecutor<S, SD> {
...
+        let state_table_metrics = metrics.new_state_table_metrics(
+            table_catalog.id,
+            actor_context.id,
+            actor_context.fragment_id,
+        );
         // Note: The current implementation could potentially trigger a switch on the inconsistent_op flag. If the storage relies on this flag to perform optimizations, it would be advisable to maintain consistency with it throughout the lifecycle.
         let state_table = StateTableBuilder::new(table_catalog, store, vnodes)
             .with_op_consistency_level(op_consistency_level)
             .enable_preload_all_rows_by_config(&actor_context.config)
+            .enable_vnode_key_pruning(true)
```
Li0k, line 269: I haven’t found any obvious errors in the review, but I still prefer to adopt enable_vnode_key_pruning_by_config rather than enabling it by default.

src/stream/src/common/table/state_table.rs
```diff
@@ -77,6 +78,105 @@ macro_rules! insane_mode_discard_point {
...
+        }
+        false
+    }
+
+    fn pruned_key_range(
+        &self,
+        start: &Bound<Bytes>,
+        end: &Bound<Bytes>,
+    ) -> Option<(Bound<Bytes>, Bound<Bytes>)> {
+        if self.can_prune_range(start, end) {
```
wenym1, line 153: Here when nothing can be pruned, we will redundantly compare the key range for twice, once in `can_prune_range` and once in here. I think we can merge all comparison logic to here in `pruned_key_range`, and `can_prune_range` is `pruned_key_range().is_none()`.

src/stream/src/common/table/state_table.rs
```diff
@@ -1064,11 +1304,21 @@ impl<LS: LocalStateStore, SD: ValueRowSerde> StateTableRowStore<LS, SD> {
     fn insert(&mut self, key: TableKey<Bytes>, value: impl Row) {
         insane_mode_discard_point!();
         let value_bytes = self.row_serde.serialize(&value).into();
+
+        let (vnode, key_without_vnode) = key.split_vnode_bytes();
+
+        // Update vnode statistics (skip if all_rows is present)
+        if self.all_rows.is_none()
```
wenym1, line 1332: The check on `self.all_rows.is_none()` seems redundant.

src/stream/src/common/table/state_table.rs
```diff
@@ -1078,9 +1328,22 @@ impl<LS: LocalStateStore, SD: ValueRowSerde> StateTableRowStore<LS, SD> {
     fn delete(&mut self, key: TableKey<Bytes>, value: impl Row) {
         insane_mode_discard_point!();
         let value_bytes = self.row_serde.serialize(value).into();
+
+        let (vnode, key_without_vnode) = key.split_vnode_bytes();
+
+        // Clear vnode statistics on delete (conservative approach, skip if all_rows is present)
+        // The deleted key might be the min or max, so we invalidate the stats
+        if self.all_rows.is_none()
```
wenym1, line 1357: ditto

src/stream/src/common/table/state_table.rs
```diff
@@ -1091,11 +1354,22 @@ impl<LS: LocalStateStore, SD: ValueRowSerde> StateTableRowStore<LS, SD> {
         insane_mode_discard_point!();
         let new_value_bytes = self.row_serde.serialize(&new_value).into();
         let old_value_bytes = self.row_serde.serialize(old_value).into();
+
+        let (vnode, key_without_vnode) = key_bytes.split_vnode_bytes();
+
+        // Update does not change the key, so statistics remain valid (skip if all_rows is present)
+        // But we update to ensure consistency
+        if self.all_rows.is_none()
```
wenym1, line 1383: ditto

src/stream/src/common/table/state_table.rs
```diff
@@ -1064,11 +1304,21 @@ impl<LS: LocalStateStore, SD: ValueRowSerde> StateTableRowStore<LS, SD> {
     fn insert(&mut self, key: TableKey<Bytes>, value: impl Row) {
         insane_mode_discard_point!();
         let value_bytes = self.row_serde.serialize(&value).into();
+
+        let (vnode, key_without_vnode) = key.split_vnode_bytes();
+
+        // Update vnode statistics (skip if all_rows is present)
+        if self.all_rows.is_none()
+            && let Some(stats) = &mut self.vnode_stats
+            && let Some(vnode_stat) = stats.get_mut(&vnode)
```
wenym1, line 1334: I think we can assume that the vnode should exist in the map and panic if not present as a sanity check.

Issue #23760, opened 2025-11-13
Add intermediate.schema.name to redshift sink parameters


zach-all:
### Is your feature request related to a problem? Please describe.

We want to be able to provide final redshift tables in a different schema than the intermediate/staging tables. To easier control access and to organize. 

### Describe the solution you'd like

Adding intermediate.schema.name to parameters of existing create redshift sink query.

PR #24223, opened 2025-12-22
feat(sink): add intermediate.schema.name for redshift sink 

xxhZs:

Resolves https://github.com/risingwavelabs/risingwave/issues/23760
`intermediate.schema.name` has been added for Redshift sink, which allows writing CDC tables and target tables to different schemas via this parameter.
If this parameter is not added, the tables will be written to the default `schema`.

Test: 
Create a Redshift cluster, then create a Redshift sink using an RW user. 
Note: Ensure network connectivity between the Redshift cluster and the RW cluster.

## Checklist

## Documentation

<details>

</details>

Add a parameter in with option: 
`intermediate.schema.name`: the schema name for creating the intermediate table. If this parameter is not configured, the schema specified in the `schema` setting will be used as its default schema.

### DIFF
--- a/src/connector/src/sink/snowflake_redshift/redshift.rs
+++ b/src/connector/src/sink/snowflake_redshift/redshift.rs
@@ -87,6 +87,9 @@ pub struct RedShiftConfig {
     #[serde(rename = "schema")]
     pub schema: Option<String>,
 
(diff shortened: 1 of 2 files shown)


hzxa21 (approved):
LGTM

PR #24229, opened 2025-12-22
refactor(optimizer): refactor plan ref from rc to arc

chenzl25:

- Refactor `PlanRef` to use `Arc` instead of `Rc`.
- Refactor `OptimizerContext` and `Share` to use `Atomic` and `RwLock` instead of `Cell` and `RefCell`.
- Make binder, planner and optimizer method async, so that we can get rid of error-prone `block_on` task with dedicated tokio runtime.
- Remove `FRONTEND_RUNTIME` from table_function and source_to_iceberg_scan_rule.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/tests/sqlsmith/tests/frontend/mod.rs
+++ b/src/tests/sqlsmith/tests/frontend/mod.rs
@@ -210,7 +210,7 @@ async fn run_batch_query(
             e.as_report()
         ))
     })?;
-    batch_plan.gen_batch_distributed_plan().map_err(|e| {
+    batch_plan.gen_batch_distributed_plan().await.map_err(|e| {
         Failed::from(format!(
             "Failed to generate batch distributed plan:\nReason:\n{}",
             e.as_report()


src/frontend/src/expr/table_function.rs
```diff
@@ -377,68 +367,63 @@ impl TableFunction {
 
         #[cfg(not(madsim))]
         {
-            let schema = tokio::task::block_in_place(|| {
-                FRONTEND_RUNTIME.block_on(async {
```
chenzl25, line 381: Ultimate goal is to remove these `FRONTEND_RUNTIME.block_on` calls.

src/frontend/src/optimizer/rule/source_to_iceberg_scan_rule.rs
```diff
@@ -64,9 +66,7 @@ impl FallibleRule<Logical> for SourceToIcebergScanRule {
                 let timezone = plan.ctx().get_session_timezone();
                 let time_travel_info = to_iceberg_time_travel_as_of(&source.core.as_of, &timezone)?;
                 let delete_parameters: IcebergDeleteParameters =
-                    tokio::task::block_in_place(|| {
-                        FRONTEND_RUNTIME.block_on(s.get_delete_parameters(time_travel_info))
-                    })?;
```
chenzl25, line 69: Ultimate goal is to remove these `FRONTEND_RUNTIME.block_on` calls.

src/frontend/src/optimizer/plan_node/mod.rs
```diff
@@ -246,7 +246,7 @@ impl_downcast!(StreamPlanNode);
 // and we currently need a manual implementation of `PartialEq` for `PlanRef`.
 #[allow(clippy::derived_hash_with_manual_eq)]
 #[derive(Debug, Eq, Hash)]
-pub struct PlanRef<C: ConventionMarker>(Rc<C::PlanRefDyn>);
```
chenzl25, line 249: Critical changes

src/frontend/src/optimizer/plan_node/generic/share.rs
```diff
@@ -12,52 +12,61 @@
...
 use risingwave_common::catalog::Schema;
 
 use super::{GenericPlanNode, GenericPlanRef};
 use crate::OptimizerContextRef;
 use crate::optimizer::property::FunctionalDependencySet;
 
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone)]
 pub struct Share<PlanRef> {
-    pub input: RefCell<PlanRef>,
```
chenzl25, line 26: Critical changes

src/frontend/src/optimizer/optimizer_context.rs
```diff
@@ -44,32 +40,30 @@ pub struct OptimizerContext {
...
+    rcte_cache: RwLock<HashMap<ShareId, PlanRef>>,
 
     /// Last assigned plan node ID.
-    last_plan_node_id: Cell<i32>,
+    last_plan_node_id: AtomicI32,
     /// Last assigned correlated ID.
-    last_correlated_id: Cell<u32>,
+    last_correlated_id: AtomicU32,
     /// Last assigned expr display ID.
-    last_expr_display_id: Cell<usize>,
```
chenzl25, line 68: Critical changes.

hzxa21:
What is the motivation of this refactoring? Is there other work blocked by this refactoring?

chenzl25:
> What is the motivation of this refactoring? Is there other work blocked by this refactoring?

No other works are blocked by this refactoring. The motivation is to remove all `block_on` method in the optimizer & binder via changing sync method to async method. Because I remember some users previously mentioned they run iceberg query and found it never returns (but it is probably an old deadlock issue in the sdk).

chenzl25:
> Personally I would -1 for this.
> 
> 1. Calling async code during optimizer is rare and lightweight, and we don't expect any high concurrency here, so using ad-hoc `block_on` seems pretty appropriate.
> 2. All-in `async` can lead to abuse that mess the code even further, and I don't think this is more beneficial than some `block_on` regarding maintenance.
> 3. There's a lot of recursion in planner and optimizer, and using `async_recursion` can be very heavy in both runtime and compile time.
> 4. I believe it should always be possible to extract the current async logic into a separate async step, in which approach we have a clear boundary for async, better performance, and less error at the same time.
> 
> Previous discussion: https://risingwave-labs.slack.com/archives/C036F5Z3EMD/p1728796955033259

Somehow agree. Previously I didn't refactor it because I think it is not an easy job, but with the assistance of current ai agent, I found it is doable now. At least, we know which part of the codebase is needed to change to achieve such a goal.

src/frontend/src/optimizer/rule/mod.rs
```diff
@@ -85,21 +86,23 @@ use InfallibleRule as Rule;
 /// unrecoverable error that stops further optimization.
 ///
 /// An [`InfallibleRule`] is always a [`FallibleRule`].
+#[async_trait]
```
wenym1, line 89: We can try removing the `async_trait`, if we don't have things like `Box<dyn Rule>`.

src/frontend/src/optimizer/rule/project_join_separate_rule.rs
```diff
@@ -16,11 +16,7 @@ use super::prelude::{PlanRef, *};
 use crate::optimizer::plan_node::*;
 pub struct ProjectJoinSeparateRule {}
```
wenym1, line 17: We can even remove the `{}`, so that when we use it we can directly do `ProjectJoinSeparateRule.apply(..)`

src/frontend/src/optimizer/rule/table_function_to_internal_backfill_progress.rs
```diff
@@ -58,7 +58,7 @@ impl FallibleRule<Logical> for TableFunctionToInternalBackfillProgressRule {
 
 impl TableFunctionToInternalBackfillProgressRule {
     fn build_plan(
-        ctx: Rc<OptimizerContext>,
+        ctx: Arc<OptimizerContext>,
```
wenym1, line 61: Can use `OptimizerContextRef` in case we want to change the type again in the future. Can simply replace keyword `Arc<OptimizerContext>` with `OptimizerContextRef.`

src/frontend/src/optimizer/rule/index_selection_rule.rs
```diff
@@ -989,7 +989,7 @@ impl ExprRewriter for ShiftInputRefRewriter {
 }
 
 impl IndexSelectionRule {
-    pub fn create() -> BoxedRule {
-        Box::new(IndexSelectionRule {})
+    pub fn apply_to_plan(&self, plan: PlanRef) -> Option<PlanRef> {
```
wenym1, line 992: We can remove this method and call `IndexSelectionRule.apply` directly.

src/frontend/src/optimizer/plan_node/generic/share.rs
```diff
@@ -12,52 +12,61 @@
...
+    pub input: Arc<RwLock<PlanRef>>,
```
wenym1, line 27: `ArcSwap` can be better, to avoid deadlock caused by accidental lock entrance.

src/frontend/src/optimizer/plan_node/generic/share.rs
```diff
@@ -12,52 +12,61 @@
...
 }
 
+impl<P: PartialEq> PartialEq for Share<P> {
+    fn eq(&self, other: &Self) -> bool {
+        *self.input.read() == *other.input.read()
```
wenym1, line 32: Here might be the cause of the deadlock mentioned in our offline discussion, when the two input refers to the same lock.

src/frontend/src/optimizer/optimizer_context.rs
```diff
@@ -44,32 +40,30 @@ pub struct OptimizerContext {
...
-    logical_explain: RefCell<Option<String>>,
+    logical_explain: RwLock<Option<String>>,
     /// Store options or properties from the `with` clause
     with_options: WithOptions,
     /// Store the Session Timezone and whether it was used.
-    session_timezone: RefCell<SessionTimezone>,
+    session_timezone: RwLock<SessionTimezone>,
     /// Total number of optimization rules have been applied.
-    total_rule_applied: RefCell<usize>,
+    total_rule_applied: RwLock<usize>,
```
wenym1, line 51: May use `AtomicUsize`?

src/frontend/src/optimizer/optimizer_context.rs
```diff
@@ -13,11 +13,9 @@
 // limitations under the License.
 
 use core::fmt::Formatter;
-use std::cell::{Cell, RefCell, RefMut};
 use std::collections::HashMap;
-use std::marker::PhantomData;
-use std::rc::Rc;
-use std::sync::Arc;
+use std::sync::atomic::{AtomicI32, AtomicU32, AtomicUsize, Ordering};
+use std::sync::{Arc, RwLock, RwLockWriteGuard};
```
wenym1, line 18: If we use `parking_lot:RwLock` we can avoid the `.unwrap()` after every `.read()` and `.write()`.

src/frontend/src/binder/expr/function/window.rs
```diff
@@ -228,79 +230,123 @@ impl Binder {
         Ok(WindowFunction::new(kind, args, ignore_nulls, partition_by, order_by, frame)?.into())
     }
 
-    fn bind_window_frame_usize_bounds(
+    async fn bind_window_frame_usize_bounds(
         &mut self,
         start: &WindowFrameBound,
         end: Option<&WindowFrameBound>,
     ) -> Result<(FrameBound<usize>, FrameBound<usize>)> {
-        let mut convert_offset = |offset: &ast::Expr| -> Result<usize> {
```
wenym1, line 236: Can use rust's new async-closure feature to avoid having duplicated code for `start` and `end` like current code.

chenzl25:
According to the discussion, the refactoring is unnecessary at the current moment, or its benefits are much less than the drawbacks it brings together. 

As for the purpose of removing the dedicated tokio runtime, we can consider the following way mentioned by @BugenZhao 

> I believe it should always be possible to extract the current async logic into a separate async step, in which approach we have a clear boundary for async, better performance, and less error at the same time.

So let's close the PR right now.

PR #24230, opened 2025-12-22
feat: add global shared object cache for iceberg table

### LINKED ISSUE #23916: feat: Use manifest cache to speed up iceberg query planning time.

API in iceberg-rust sdk. (Changes are needed)
https://github.com/risingwavelabs/iceberg-rust/blob/833739eacf094fed8e33849ab68df0a0937e16da/crates/iceberg/src/io/object_cache.rs#L42-L47

Maybe we can maintain an manifest cache in the frontend/serving node to speed up the iceberg query planning time. We need to pass the cache to the plan context of iceberg-rust to utilize the cache across table meta. Currently, the iceberg-rust sdk manifest cache implementation is limited to a single table/snapshot and can't be shared across different snapshots and queries, so we need to change the api and pass the cache if necessary.

_No response_

_No response_

### Additional context

_No response_

yezizp2012:

Resolves #23916 .

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/e2e_test/streaming/aggregate/two_phase_approx_percentile_merge.slt
+++ b/e2e_test/streaming/aggregate/two_phase_approx_percentile_merge.slt
@@ -1,4 +1,36 @@
 # Single phase approx percentile
+
+onlyif can-use-recover
+statement ok
+create table t(p_col double, grp_col int);
+
+onlyif can-use-recover
+statement ok
+create materialized view m1 as
+ select
+     approx_percentile(0.01, 0.01) within group (order by p_col) as p01,
+     approx_percentile(0.1, 0.01) within group (order by p_col) as p10,
+     approx_percentile(0.5, 0.01) within group (order by p_col) as p50,
+     approx_percentile(0.9, 0.01) within group (order by p_col) as p90,
+     approx_percentile(0.99, 0.01) within group (order by p_col) as p99
+ from t;
+
+onlyif can-use-recover
+statement ok
+recover;
+
+onlyif can-use-recover
+statement ok
+insert into t values (1.0, 1);
+
+onlyif can-use-recover
+statement ok
+flush;
+
+onlyif can-use-recover
+statement ok
+drop table t cascade;
+
 statement ok
 create table t(p_col double, grp_col int);
 
--- a/src/stream/src/executor/approx_percentile/global_state.rs
+++ b/src/stream/src/executor/approx_percentile/global_state.rs
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 use std::collections::{BTreeMap, Bound};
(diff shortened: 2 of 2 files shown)


src/frontend/src/catalog/system_catalog/rw_catalog/rw_iceberg_all_files.rs
```diff
@@ -82,15 +81,16 @@ async fn read(reader: &SysCatalogReaderImpl) -> Result<Vec<RwIcebergFiles>> {
...
             let table: Table = iceberg_properties.load_table().await?;
-            let metadata = table.metadata();
+            let metadata = table.metadata_ref();
             let snapshots = metadata.snapshots();
             let mut reachable_manifests = HashSet::new();
             for snapshot in snapshots {
-                let manifest_list: ManifestList = snapshot
-                    .load_manifest_list(table.file_io(), table.metadata())
+                let manifest_list = table
+                    .object_cache()
```
chenzl25, line 89: Shall we change the `rw_iceberg_files` system table as well?

src/connector/src/connector_common/iceberg/mod.rs
```diff
@@ -789,6 +799,30 @@ impl IcebergCommon {
...
+/// Get a globally shared object cache keyed by table UUID to avoid reuse after drop & recreate.
+pub(crate) async fn shared_object_cache(
+    init_object_cache: Arc<ObjectCache>,
+    table_uuid: Uuid,
+) -> Arc<ObjectCache> {
+    static CACHE: LazyLock<MokaCache<Uuid, Arc<ObjectCache>>> = LazyLock::new(|| {
+        MokaCache::builder()
+            .max_capacity(SHARED_OBJECT_CACHE_MAX_TABLES)
+            .build()
+    });
```
chenzl25, line 816: It seems this cache could stay in different components like frontend, meta, compute, iceberg-compactor wherever the load_table is called. 🤔

chenzl25: At the beginning, actually we want to introduce this cache to the frontend only.

yezizp2012: Yes, if other components can benefit from the cache, I think it's acceptable? 🤔 If they are only built during initialization and not accessed afterwards, the cache will also be very small.

yezizp2012: Moving the rebuild logic of cache to various parts of the frontend, such as ddl, plan stage will make the code look a little dirty.

chenzl25: Got it.

PR #24232, opened 2025-12-22
refactor(storage): use `clean_watermark_index` in `StateTable`

BugenZhao:

`clean_watermark_index_in_pk` should be deprecated in favor of `clean_watermark_index` after introducing in #23877. This PR refactored the `StateTable` struct to replace `clean_watermark_index_in_pk` with `clean_watermark_index`, as a preparation for #24186.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/prost/src/lib.rs
+++ b/src/prost/src/lib.rs
@@ -540,9 +540,8 @@ impl catalog::Table {
             self.clean_watermark_indices.clone()
         } else if let Some(pk_idx) = self
             .clean_watermark_index_in_pk
-            // At the very beginning, we didn't have `clean_watermark_index_in_pk` as well.
-            // When `cleaned_by_watermark` is set, we assume the watermark is the first column.
-            .or_else(|| self.cleaned_by_watermark.then_some(0))
+            // At the very beginning, the watermark index was hard-coded to the first column of the pk.
+            .or_else(|| (!self.pk.is_empty()).then_some(0))
         {
             // Old format: convert PK index to column index
             // The pk_idx is the position in the PK, we need to find the corresponding column index


BugenZhao:

- #24164
    - #24152
        - #24232 <-- you are here 🤪
            - #24121
                - #24186
            - #24266

src/stream/src/common/table/state_table.rs
```diff
@@ -698,25 +700,43 @@ where
...
+                    ),
+                    // TODO(ttl): may introduce a new type for watermark not in pk.
+                    WatermarkSerdeType::NonPkPrefix,
+                ),
+            };
+            (watermark_serde, watermark_serde_type)
+        });
 
         // Restore persisted table watermark.
-        let watermark_serde = if pk_indices.is_empty() {
```
zwang28, line 707: I did once encounter `pk_indices.is_empty()` in e2e. Do you know what could lead to it?

zwang28 (approved):
rest LGTM

src/stream/src/common/table/state_table.rs
BugenZhao: I think it's pretty normal. Perhaps your table has nothing to do with watermarks.

Based on my investigation, it appears that before we supported watermark on pk non-prefix column, we always presumed the watermark is the first column of the pk. This applied only to tables that **actually** have watermarks, which are determined at runtime by whether the executor commits a watermark, instead of any other field in `Table` catalog.

So if we later hit `commit_pending_watermark`, it means that the `clean_watermark_index[_in_pk]` must be set, or, the primary key is non-empty and the watermark column is the first one.

PR #24242, opened 2025-12-23
fix(optimizer): fix topn index selection for scan with output col idx

chenzl25:

- Introduce top_n_project_transpose_rule to transpose the topn and project to let TopNOnIndexRule can apply to more plans.
- Improve TopNOnIndexRule to handle output column idx of the scan.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/optimizer/rule/project_top_n_transpose_rule.rs
+++ b/src/frontend/src/optimizer/rule/project_top_n_transpose_rule.rs
@@ -1,4 +1,4 @@
-// Copyright 2025 RisingWave Labs
+// Copyright 2026 RisingWave Labs
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/src/frontend/src/optimizer/rule/top_n_project_transpose_rule.rs
+++ b/src/frontend/src/optimizer/rule/top_n_project_transpose_rule.rs
@@ -1,4 +1,4 @@
-// Copyright 2025 RisingWave Labs
+// Copyright 2026 RisingWave Labs
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.


src/frontend/src/optimizer/rule/top_n_on_index_rule.rs
```diff
@@ -54,20 +54,66 @@ impl TopNOnIndexRule {
...
+                            ColumnOrder {
+                                column_index: output_idx,
+                                order_type: OrderType::descending_nulls_first(),
+                            },
+                            ColumnOrder {
+                                column_index: output_idx,
+                                order_type: OrderType::descending_nulls_last(),
+                            },
+                        ]
+                    })
```
BugenZhao: Can we use `all()`?

src/frontend/planner_test/tests/testdata/output/lateral_subquery.yaml
```diff
@@ -72,9 +72,9 @@
     └─BatchHashJoin { type: Inner, predicate: salesperson.id IS NOT DISTINCT FROM all_sales.salesperson_id, output: [salesperson.name, all_sales.amount, all_sales.customer_name] }
       ├─BatchExchange { order: [], dist: HashShard(salesperson.id) }
       │ └─BatchScan { table: salesperson, columns: [salesperson.id, salesperson.name], distribution: SomeShard }
-      └─BatchGroupTopN { order: [all_sales.amount DESC], limit: 1, offset: 0, group_key: [all_sales.salesperson_id] }
-        └─BatchExchange { order: [], dist: HashShard(all_sales.salesperson_id) }
-          └─BatchProject { exprs: [all_sales.salesperson_id, all_sales.amount, all_sales.customer_name] }
+      └─BatchProject { exprs: [all_sales.salesperson_id, all_sales.amount, all_sales.customer_name] }
+        └─BatchGroupTopN { order: [all_sales.amount DESC], limit: 1, offset: 0, group_key: [all_sales.salesperson_id] }
+          └─BatchExchange { order: [], dist: HashShard(all_sales.salesperson_id) }
```
BugenZhao: Is this a side effect of regression on column pruning?

chenzl25: Yes. it is caused by the TopNProjectTransposeRule.

chenzl25: I am thinking if we should add a ProjectTopNTransposeRule at the end of this stage

src/frontend/planner_test/tests/testdata/input/index_selection.yaml
```diff
@@ -568,5 +568,18 @@
...
+  sql: |
+    create table t (c int, b int, a int);
+    create index idx on t(a, b desc, c desc);
+    select b from t where a = 1 order by b desc, c desc limit 1;
+  expected_outputs:
+    - batch_plan
+- name: topn index selection for scan with output column idx
+  sql: |
+    create table t (c int, b int, a int, primary key (a, b, c));
+    select b from t where a = 1 order by b, c limit 1;
```
BugenZhao, line 595: Where's the "index" mentioned in the name? 🤔

BugenZhao (commented):
Rest LGTM

src/frontend/planner_test/tests/testdata/input/index_selection.yaml
chenzl25: The index here is the primary index.

PR #24246, opened 2025-12-23
feat(iceberg): support schema change(add column) for non exactly once iceberg sink 

wcy-fdu:

as title, part of https://github.com/risingwavelabs/risingwave/pull/24040

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/prost/src/lib.rs
+++ b/src/prost/src/lib.rs
@@ -118,9 +118,6 @@ pub mod telemetry;
 #[rustfmt::skip]
 #[cfg_attr(madsim, path = "sim/iceberg_compaction.rs")]
 pub mod iceberg_compaction;
-#[rustfmt::skip]
-#[cfg_attr(madsim, path = "sim/iceberg_commit_metadata.rs")]
-pub mod iceberg_commit_metadata;
 
 #[rustfmt::skip]
 #[path = "sim/secret.rs"]


ci/scripts/e2e-iceberg-sink-cdc-test.sh
```diff
@@ -0,0 +1,246 @@
...
+    s3.endpoint = 'http://127.0.0.1:9301',
+    s3.region = 'custom',
+    s3.access.key = 'hummockadmin',
+    s3.secret.key = 'hummockadmin',
+    create_table_if_not_exists = 'true',
+    commit_checkpoint_interval = 2,
+    primary_key = 'id',
+    force_append_only='true',
+    auto.schema.change = 'true',
+    is_exactly_once = 'false',
```
wcy-fdu, line 139: set` is_exactly_once = false`, after the 2pc framework is complete, we need reset it to true. cc @Chiro11

src/connector/src/sink/iceberg/mod.rs
```diff
@@ -2031,7 +2049,7 @@ impl IcebergSinkCommitter {
         Ok(Some((write_results, snapshot_id)))
     }
 
-    async fn commit_iceberg_inner(
+    async fn commit_datafile(
```
wcy-fdu, line 2052: These renamings will make the implementation of 2PC clearer later, so I included them in this pull request.

src/prost/build.rs
```diff
@@ -802,6 +802,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         "health",
         "hummock",
         "iceberg_compaction",
+        "iceberg_commit_metadata",
```
wenym1: Changes in the pb files seems not related to this PR. May remove the related changes.

src/connector/src/sink/iceberg/mod.rs
```diff
@@ -1882,18 +1887,31 @@ impl SinglePhaseCommitCoordinator for IcebergSinkCommitter {
         add_columns: Option<Vec<Field>>,
     ) -> Result<()> {
         tracing::info!("Starting iceberg direct commit in epoch {epoch}");
-
-        let (write_results, snapshot_id) =
-            match self.pre_commit_inner(epoch, metadata, add_columns)? {
-                Some((write_results, snapshot_id)) => (write_results, snapshot_id),
-                None => {
+        // Commit data if present
```
wenym1: May temporarily add a check in 2pc coordinator when we try to add columns in `pre_commit`.

src/connector/src/sink/iceberg/mod.rs
```diff
@@ -1882,18 +1887,31 @@ impl SinglePhaseCommitCoordinator for IcebergSinkCommitter {
...
+        let (write_results, snapshot_id) = match self.pre_commit_inner(epoch, metadata, None)? {
+            Some((write_results, snapshot_id)) => (write_results, snapshot_id),
+            None => {
+                // No data to commit, but may have schema change
+                if let Some(add_columns) = add_columns {
```
wenym1: We can do 
```
if let Some((write_results, snapshot_id)) = self.pre_commit_inner(epoch, metadata, None)? {
    self.commit_datafile(epoch, write_results, snapshot_id)
            .await?;
}

// Commit schema change if present
if let Some(add_columns) = add_columns {
    tracing::info!(?epoch, "Committing schema change after data commit");
    self.commit_schema_change(add_columns).await?;
}
```

And then we don't have to check for schema change for twice.

proto/iceberg_commit_metadata.proto
Chiro11, line 1: It seems this file is no longer needed.

Chiro11: Compilation failed in ci.

PR #24252, opened 2025-12-24
refactor(meta): store table change log independently

### LINKED ISSUE #23924: refactor: separate table_change_log from HummockVersion struct

When we create a subscription with large retention time, the table_change_log` of `HummockVersion` gets large.  which increases the size of hummock version, and may have the following problem:
1. the size of hummock version checkpoint gets large.  
2. every time when we do `apply_version_update`, we will first clone the hummock version, which doubles the memory usage of hummock version, and can easily OOM

Besides, the `table_change_log` is not frequently used, and it's not worthy of putting in inside the `HummockVersion`, where data is actively accessed by streaming executor.

We may separate the `table_change_log` from `HummockVersion` struct. Instead, we store it in a separate meta store table. This meta store table gets updated in a single transaction as the hummock version delta table, so that the current atomicity can be pertained.

The meta store table may have the following columns
```
CREATE TABLE hummock_table_change_log {
    table_id uint32,
    checkpoint_epoch uint64,
    non_checkpoint_epochs uint64[],
    new_value_sst jsonb, // json!(Vec<SstableInfo>)
    old_value_sst jsonb, // json!(Vec<SstableInfo>)
}
```

Summary of current usage of `table_change_log`
* `iter_log` of hummock
* `next_epoch` of hummock
* in frontend, subscription cursor get the list of epochs
* in recovery of snapshot backfill to resolve the list of lagging epochs

Previously the `table_change_log` gets propagated together in `HummockVersion` so that we can get the required information by accessing the `HummockVersion` maintained via notification from meta. After we separate `table_change_log` from `HummockVersion`, we need to obtain these information in a new way. We may support some RPCs to meta node to fetch the information stored there. We can have some local cache to avoid frequent duplicated requests.

Note that we should also consider the SSTs pinned by this meta store table.

_No response_

### Additional context

_No response_

zwang28:

This PR resolves #23924.
- `table_change_log` field in `HummockVersion` is deprecated.
- `table_change_log` are now stored in meta store.
- The meta node still maintain a in-mem cache for `table_change_log`, independent of `HummockVersion`.
- The frontend node and compute node are modified to fetch table change log on demand. They no longer maintain local table change logs cache.

## Checklist

## Documentation

<details>

</details>


src/meta/src/hummock/manager/transaction.rs
```diff
@@ -258,19 +281,93 @@ impl InMemValTransaction for HummockVersionTransaction<'_> {
...
         if self.disable_apply_to_txn {
             return Ok(());
         }
-        for delta in self
-            .pre_applied_version
-            .iter()
-            .flat_map(|(_, deltas)| deltas.iter())
-        {
-            delta.upsert_in_transaction(txn).await?;
+        if let Some((_, deltas, modified_table_change_log)) = &self.pre_applied_version {
```
wenym1: Can we do the migration during meta bootstrap like where we call `may_fill_backward_compatible_state_table_info_delta`, and then we can change to only read the separate table_change_log.

zwang28: I can resolve this later after we confirm [this PR is still needed](https://github.com/risingwavelabs/risingwave/issues/23924#issuecomment-3692200181).

zwang28:
@wenym1 This PR is ready for review.

src/meta/src/hummock/manager/transaction.rs
zwang28: Resolved.

src/frontend/src/scheduler/snapshot.rs
```diff
@@ -270,22 +252,8 @@ impl HummockSnapshotManager {
...
-            .iter()
-            .filter_map(|(table_id, change_log)| {
-                if change_log.get_non_empty_epochs(0, usize::MAX).is_empty() {
-                    None
-                } else {
-                    Some(*table_id)
-                }
-            })
-            .collect();
         self.table_change_log_notification_sender
```
wenym1: May rename it to `version_update_notification_sender` with item as `()`.

src/storage/backup/src/meta_snapshot_v2.rs
```diff
@@ -202,6 +206,47 @@ impl Metadata for MetadataV2 {
...
+        self.hummock_table_change_logs
+            .iter()
+            .flat_map(|m| {
+                let change_log = to_table_change_log(m);
+                change_log
+                    .new_value
+                    .iter()
+                    .chain(change_log.old_value.iter())
+                    .map(|s| s.object_id.as_raw())
+                    .collect::<HashSet<_>>()
```
wenym1: The inner `collect` is unnecessary

src/storage/backup/src/meta_snapshot_v2.rs
```diff
@@ -202,6 +206,47 @@ impl Metadata for MetadataV2 {
...
+
+    fn table_change_log_object_ids(&self) -> HashSet<HummockRawObjectId> {
+        self.hummock_table_change_logs
+            .iter()
+            .flat_map(|m| {
+                let change_log = to_table_change_log(m);
+                change_log
+                    .new_value
+                    .iter()
+                    .chain(change_log.old_value.iter())
```
wenym1: We can have a helper method in `TableChangeLog` for this new_value_ssts.chain(old_value_ssts)

src/storage/hummock_sdk/src/change_log.rs
```diff
@@ -19,12 +19,14 @@ use risingwave_pb::hummock::hummock_version_delta::PbChangeLogDelta;
...
 
+use crate::HummockObjectId;
 use crate::sstable_info::SstableInfo;
+use crate::version::ObjectIdReader;
 
 #[derive(Debug, Clone, PartialEq)]
 pub struct TableChangeLogCommon<T>(
     // older log at the front
-    VecDeque<EpochNewChangeLogCommon<T>>,
+    pub VecDeque<EpochNewChangeLogCommon<T>>,
```
wenym1: Better to avoid making it pub and provide helper method to access it, unless it will become too complicated.

src/storage/hummock_sdk/src/compaction_group/hummock_version_ext.rs
```diff
@@ -1492,7 +1453,7 @@ fn level_insert_ssts(operand: &mut Level, insert_table_infos: &Vec<SstableInfo>)
     }
 }
 
-pub fn object_size_map(version: &HummockVersion) -> HashMap<HummockObjectId, u64> {
+pub fn version_object_size_map(version: &HummockVersion) -> HashMap<HummockObjectId, u64> {
```
wenym1, line 1446: I think we can always pass `current_table_change_log: &HashMap<TableId, TableChangeLog>,` as a parameter to be consist with the original logic.

proto/hummock.proto
```diff
@@ -381,6 +381,26 @@ message GetCurrentVersionResponse {
   HummockVersion current_version = 2;
 }
 
+message GetTableChangeLogsRequest {
+  message TableFilter {
+    repeated uint32 table_ids = 1;
```
wenym1, line 442: Can declare the type of `table_ids` to be strong-typed `TableId` in `build.rs` like other proto there.

src/meta/service/src/hummock_service.rs
```diff
@@ -668,6 +669,39 @@ impl HummockManagerService for HummockServiceImpl {
...
+            limit,
+        } = request.into_inner();
+        let table_change_logs = self
+            .hummock_manager
+            .get_table_change_logs(
+                epoch_only,
+                start_epoch_inclusive,
+                end_epoch_inclusive,
+                table_ids
+                    .map(|t| t.table_ids.iter().map(|id| TableId::new(*id)).collect())
```
wenym1: Can avoid type conversion between u32 and `TableId` if we declare the field type as the strong typed `TableId`

src/meta/src/hummock/manager/checkpoint.rs
```diff
@@ -140,17 +142,27 @@ impl HummockManager {
...
+                &self.metrics,
+                &versioning.checkpoint,
+                &versioning.checkpoint_table_change_log_object_size,
+                min_pinned_version_id,
+                &versioning.table_change_log,
+            );
             return Ok(0);
         }
-        assert!(new_checkpoint_id > old_checkpoint_id);
+        assert!(new_checkpoint_id >= old_checkpoint_id);
```
wenym1: Any reason to allow `==`?

src/meta/src/hummock/manager/versioning.rs
```diff
@@ -58,12 +61,15 @@ pub struct Versioning {
     pub time_travel_snapshot_interval_counter: u64,
     /// Used to avoid the attempts to rewrite the same SST to meta store
     pub last_time_travel_snapshot_sst_ids: HashSet<HummockSstableId>,
+    /// Used by object GC and metric update.
```
wenym1: Seems unnecessary to maintain this map. Can we just rebuild it every time we want to trigger sst stats like how we do for objects in version?

src/storage/src/hummock/store/hummock_storage.rs
```diff
@@ -219,7 +223,9 @@ impl HummockStorage {
         );
 
         let event_tx = hummock_event_handler.event_sender();
-
+        // TODO(ZW): configurable
+        let table_change_log_manager =
```
wenym1: Can we fix the TODO together in this PR?

src/frontend/src/session/cursor_manager.rs
```diff
@@ -645,14 +647,20 @@ impl SubscriptionCursor {
                     if cur > 0 || timeout_seconds == 0 {
                         break;
                     }
+                    let State::InitLogStoreQuery { seek_timestamp, .. } = &self.state else {
```
wenym1, line 650: @xxhZs Can you help confirm whether this assertion on state is correct?

wenym1 (commented):
Leave some early comments. Will continue review the core logic.

src/frontend/src/scheduler/snapshot.rs
zwang28: refactored.

src/storage/backup/src/meta_snapshot_v2.rs
zwang28: Refactored.

src/storage/hummock_sdk/src/change_log.rs
zwang28: refactored.

src/storage/hummock_sdk/src/compaction_group/hummock_version_ext.rs
zwang28: Which methods do you mean?

proto/hummock.proto
zwang28: refactored.

src/meta/service/src/hummock_service.rs
zwang28: refactored.

src/meta/src/hummock/manager/checkpoint.rs
zwang28: A checkpoint with 0 delta may be committed after this:
```
let need_checkpoint = instance.may_fill_backward_table_change_logs().await?;
instance.load_table_change_log().await?;
if need_checkpoint {
    instance.create_version_checkpoint(0).await?;
}
```

src/meta/src/hummock/manager/versioning.rs
zwang28: It's necessary.

It tracks the object size at the last checkpoint and replays deltas up to the current state to determine the size of any objects removed between checkpoints. The same approach is used for object tracked in version.

```
let mut object_sizes = version_object_size_map(&old_checkpoint.version);
object_sizes.extend(old_checkpoint_table_change_log_object_size.iter());
```

src/storage/src/hummock/store/hummock_storage.rs
zwang28: Resolved.

src/storage/backup/src/meta_snapshot_v2.rs
zwang28: Resolved.

src/storage/hummock_sdk/src/change_log.rs
```diff
@@ -56,6 +56,24 @@ impl<T> TableChangeLogCommon<T> {
...
+    }
+
+    pub fn binary_search_by_epoch(&self, epoch: u64) -> Result<usize, usize> {
+        self.0
+            .binary_search_by_key(&epoch, |log| log.checkpoint_epoch)
+    }
+}
+
+impl<T> IntoIterator for TableChangeLogCommon<T> {
+    type IntoIter = std::collections::vec_deque::IntoIter<EpochNewChangeLogCommon<T>>;
```
wenym1:
```
type IntoIter = impl Iterator<Item = EpochNewChangeLogCommon<T>>;
```

src/meta/src/hummock/manager/versioning.rs
wenym1: Oh, I see, so it's a mapping created from the previous checkpoint version. But how is this field initialized when meta node is launched? Is it an empty map, or we reconstruct this map from the latest checkpoint version?

src/frontend/src/session.rs
```diff
@@ -1312,17 +1313,43 @@ impl SessionImpl {
...
+                true,
+                Some(max_count),
+            )
+            .await?;
+        let Some(e) = ret.get(&table_id) else {
+            return Ok(vec![]);
+        };
+        Ok(e.iter()
+            .flat_map(|l| l.epochs())
+            .filter(|e| *e >= min_epoch)
```
wenym1: May also filter by `*e <= max_epoch` to make it more robust.

src/meta/src/hummock/manager/checkpoint.rs
zwang28: Refactored. Resolved

src/meta/src/hummock/manager/transaction.rs
```diff
@@ -56,30 +59,42 @@ fn trigger_version_stat(metrics: &MetaMetrics, current_version: &HummockVersion)
...
     notification_manager: &'a NotificationManager,
     table_committed_epoch_notifiers: Option<&'a Mutex<TableCommittedEpochNotifiers>>,
     meta_metrics: &'a MetaMetrics,
 
-    pre_applied_version: Option<(HummockVersion, Vec<HummockVersionDelta>)>,
+    #[expect(clippy::type_complexity)]
+    pre_applied_version: Option<(
+        HummockVersion,
+        Vec<HummockVersionDelta>,
+        HashMap<TableId, TableChangeLog>,
```
wenym1: Does it mean we will clone the whole table change log? I am worried that this may increase the memory usage and OOM

src/meta/src/hummock/manager/versioning.rs
```diff
@@ -277,6 +298,129 @@ impl HummockManager {
         }
         Ok(())
     }
+
+    pub async fn may_fill_backward_table_change_logs(&self) -> Result<()> {
+        let mut versioning = self.versioning.write().await;
```
wenym1, line 302: Shall we change the condition to do backfill to `!version.table_change_log.is_empty() && meta store table is empty` and in this way we can avoid re-overwriting the existing table change log if the meta node gets restarted before we make the first checkpoint before the migration. And then we can also remove the `.on_conflict_do_nothing()` below.

src/storage/src/hummock/store/hummock_storage.rs
```diff
@@ -777,28 +785,56 @@ impl StateStoreReadLog for HummockStorage {
...
                     Ok(true)
                 } else {
                     Ok(false)
                 }
-            },
-            || format!("wait next_epoch: epoch: {} {}", epoch, options.table_id),
-        )
+            }
+        };
+        wait_for_update(&self.version_update_notifier_tx, inspect_fn, || {
```
wenym1: We change the `inspect_fn` to a closure that generate a future because we want to do `fetch_table_change_logs` inside the `inspect_fn`. But I think a cleaner way to implement this is that in `inspect_fn` we only wait for the committed epoch of the table to be greater than epoch. After `wait_for_update` returns, we then do `fetch_table_change_logs` over `(epoch, table committed epoch)`, and get the result.

src/meta/src/hummock/manager/checkpoint.rs
```diff
@@ -140,17 +142,27 @@ impl HummockManager {
...
         let mut archive: Option<PbHummockVersionArchive> = None;
         let mut stale_objects = old_checkpoint.stale_objects.clone();
         // `object_sizes` is used to calculate size of stale objects.
-        let mut object_sizes = object_size_map(&old_checkpoint.version);
+        let mut object_sizes = version_object_size_map(&old_checkpoint.version);
+        object_sizes.extend(old_checkpoint_table_change_log_object_size.iter());
         // The set of object ids that once exist in any hummock version
-        let mut versions_object_ids: HashSet<_> =
-            old_checkpoint.version.get_object_ids(false).collect();
+        let mut versions_object_ids: HashSet<_> = old_checkpoint
```
wenym1: I am thinking that, since table change log does not need to support MVCC, which is different to LSM version, maybe here when calculating the `versions_object_ids` and the later `current_version_object_ids` and `removed_object_ids`, we don't have to consider the SSTs in table change log.

And then the GC of SSTs in table change log can be handled separately to LSM version. SSTs are only tracked by table change log if existing in 1. the latest version, and 2. in the snapshot. Therefore, we can add SSTs in table change log directly to `may_delete_object_ids` when the log SSTs are truncated in the latest version.

wenym1 (commented):
Rest LGTM.

src/meta/src/hummock/manager/transaction.rs
zwang28: The clone of table change log also happens even before this PR: They're cloned along with version.

zwang28: [self.orig_version.clone()](https://github.com/risingwavelabs/risingwave/blob/30301dc965a6f30c08de859e2be0e6cb1b66f6b0/src/meta/src/hummock/manager/transaction.rs#L111)

src/storage/hummock_sdk/src/change_log.rs
zwang28: Refactored.

src/meta/src/hummock/manager/versioning.rs
zwang28: > so it's a mapping created from the previous checkpoint version

Yes

> how is this field initialized when meta node is launched? Is it an empty map

Yes. Because Neither meta store nor object store keeps the table change logs of **previous** checkpoint verison. It resulted the size metric calculated from the first version checkpoint after meta is lauched is inaccurate. It's fine since subsequent ones are accurate.

src/frontend/src/session.rs
zwang28: Refactored.

src/meta/src/hummock/manager/versioning.rs
zwang28: Refactored.

src/storage/src/hummock/store/hummock_storage.rs
zwang28: Refactored.

src/meta/src/hummock/manager/checkpoint.rs
zwang28: Does it mean 
1. checkpoint_table_change_log_object_size is not needed.
2. table change log's object size are only tracked in current version, rather than 3 places current_version_object_size/old_version_object_size/stale_object_size like the LSM version?

src/frontend/src/session/cursor_manager.rs
zwang28: Actually No.
SubscriptionCursor::next_row may return None and self.state = State::Fetch.

zwang28: Fixed.

src/meta/src/hummock/manager/checkpoint.rs
wenym1: Yes. It means that table change log SSTs are only tracked by the latest table change log, or in any snapshot.

We might be careful about that an SST can be tracked by both table change log and LSM. So we can't safely delete a SST when the log entry of it gets truncated.

src/meta/src/hummock/manager/transaction.rs
wenym1: I think only holding the mutable reference of the field can also works. We only apply the change on it after the metadata has been committed to meta store table successfully.

zwang28: Thanks for pointing this out.
Refactored in this commit: https://github.com/risingwavelabs/risingwave/pull/24252/changes/c172836a51426d86228b61ee7ff55bfca7db3f0f

src/meta/src/hummock/manager/checkpoint.rs
zwang28: Refactored in this commit: https://github.com/risingwavelabs/risingwave/pull/24252/changes/7e0c030a9049823e9763ec936f2f220bc31e7ce6

> We might be careful about that an SST can be tracked by both table change log and LSM. So we can't safely delete a SST when the log entry of it gets truncated.

Guarded by `get_tracked_object_ids`before deletion.

src/meta/src/hummock/manager/versioning.rs
zwang28: Refactored.

src/meta/src/hummock/manager/transaction.rs
```diff
@@ -123,23 +117,27 @@ impl<'a> HummockVersionTransaction<'a> {
...
-                    self.orig_table_change_log.clone(),
-                    HashMap::new(),
+                    HashSet::new(),
                 )
             });
         let changed_table_info = version.apply_version_delta(&delta);
-        *table_change_log_delta = HummockVersion::apply_change_log_delta(
-            table_change_log,
+        // Ideally, the first parameter should be the cumulative state (orig_table_change_log + all applied deltas).
+        // However, currently, we use orig_table_change_log directly because deltas are only applied after a successful metastore write in the end of the transaction.
```
wenym1, line 130: Any reason why splitting the original `apply_change_log_delta` into the current two steps `collect_gc_change_log_delta` + new `apply_change_log_delta`? Shall we just keep the original `apply_change_log_delta` and change to apply `delta` + `changed_table_info` by `apply_change_log_delta` later in `commit`? We can change `deltas` to store `Vec<(delta, changed_table_info)>`.

src/storage/hummock_sdk/src/compaction_group/hummock_version_ext.rs
```diff
@@ -729,18 +697,12 @@ impl<L: Clone> HummockVersionCommon<SstableInfo, L> {
         changed_table_info
     }
 
-    /// Returns the enriched table change log delta. It's used to apply the change log delta to the meta store.
     pub fn apply_change_log_delta<T: Clone>(
```
wenym1, line 700: may not change the logic here with https://github.com/risingwavelabs/risingwave/pull/24252/changes/2efcb0ab544cad591075d34247d83922a3b9eebe..7e0c030a9049823e9763ec936f2f220bc31e7ce6#r2909343361

src/meta/src/hummock/manager/versioning.rs
```diff
@@ -277,6 +296,112 @@ impl HummockManager {
...
+        let version = &mut versioning.current_version;
+
+        use sea_orm::PaginatorTrait;
+        let meta_store_table_change_log_count =
```
wenym1: select with limit 1 seems to be lighter than count.

src/meta/src/hummock/manager/transaction.rs
zwang28: The collected change log delta are used by **firstly** `HummockVersionTransaction::apply_to_txn` and then `InMemValTransaction::commit`.

wenym1 (approved):
LGTM!

Li0k (commented):
LGTM, thanks for the efforts

PR #24256, opened 2025-12-24
perf: Reduce block-meta copies in the compactor .

Li0k:

### Summary
Reduce block-meta copies in the compactor iterator pipeline.

### Changes
- Change filter_block_metas to return Range<usize> instead of Vec<BlockMeta>, avoiding unnecessary block meta copies
- Store TableHolder and block_metas_range in SstableStreamIterator instead of cloned Vec<BlockMeta>
- Add helper methods block_metas() and block_count() for accessing block metadata via range slicing
###Motivation
Previously, SstableStreamIterator owned a cloned Vec<BlockMeta> which caused unnecessary memory allocations and copies during compaction. Now it holds a reference to the original SST metadata via TableHolder and uses a Range<usize> to track the relevant block range.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/storage/src/hummock/compactor/iterator.rs
+++ b/src/storage/src/hummock/compactor/iterator.rs
@@ -14,6 +14,7 @@
 
 use std::cmp::Ordering;
 use std::collections::HashSet;
+use std::ops::Range;
 use std::sync::atomic::AtomicU64;
 use std::sync::{Arc, atomic};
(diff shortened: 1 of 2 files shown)


hzxa21:
Is this ready for review?

Li0k:
> Is this ready for review?

PTAL @zwang28 @hzxa21

Issue #24199, opened 2025-12-19
panicked at: new database graph info can only be created for normal creating streaming job, but get command


BugenZhao:
### Buildkite URL

https://buildkite.com/risingwavelabs/pull-request/builds/87998#019b349e-a720-48ae-8d4e-cc1fb8629887/L2759

### Description and Insights

```
thread 'rw-main' (13957) panicked at src/meta/src/barrier/checkpoint/control.rs:243:25:
new database graph info can only be created for normal creating streaming job, but get command: 1 DropStreamingJobs { streaming_job_ids: {6, 7}, actors: [14, 21, 15, 27, 31, 2, 28, 6, 16, 34, 24, 17, 3, 30, 19, 20, 0, 23, 36, 35, 7, 12, 18, 32, 11, 33, 26, 8, 13, 4, 5, 9, 1, 25, 29, 10, 22], unregistered_state_table_ids: {7, 9, 8, 10, 6}, unregistered_fragment_ids: {2, 1, 4, 3}, dropped_sink_fragment_by_targets: {} }
```

cc @wenym1

BugenZhao:
One more occurrence, still in `e2e_test/error_ui/simple/recovery.slt`

https://buildkite.com/risingwavelabs/pull-request/builds/89614#019bc11a-1aec-46c4-a3d8-033a4c9e8963

cc @Chiro11 I roughly remember that we have some discussion and conclusion on this?

PR #24261, opened 2025-12-25
feat(meta): pause/resume cluster checkpoint on command

wenym1:

Remove the handling of pause/resume in barrier command mutation, and instead pause/resume the cluster by triggering recovery to rebuild the cluster, which simplifies the management of running job graphs to let it avoid handling the logic of pause/resume in low-level actors and barriers.

resolve https://github.com/risingwavelabs/risingwave/issues/24199 by the way.

Below is generated by AI.

This pull request removes the "Pause" and "Resume" mutation types and associated logic from the streaming cluster's barrier management system. Instead, pausing and resuming are now handled at a higher level in the barrier manager, simplifying the protocol buffers, command infrastructure, and state management. The changes also clean up related code paths and remove unnecessary checks and fields.

Protocol and mutation changes:
* Removed `PauseMutation` and `ResumeMutation` messages and fields from the `stream_plan.proto` protobuf definitions, and marked their field numbers and names as reserved to prevent reuse. [[1]](diffhunk://#diff-548bcb845b3db0225b2c6ab9d843234320a813af3a44855b764aaa284eda2dd8L39-R40) [[2]](diffhunk://#diff-548bcb845b3db0225b2c6ab9d843234320a813af3a44855b764aaa284eda2dd8L113-L116) [[3]](diffhunk://#diff-548bcb845b3db0225b2c6ab9d843234320a813af3a44855b764aaa284eda2dd8L172-L175) [[4]](diffhunk://#diff-548bcb845b3db0225b2c6ab9d843234320a813af3a44855b764aaa284eda2dd8R186-R187)
* Updated the `AddMutation` and related mutation structures to remove the `pause` field and all logic around pausing/resuming at the mutation level. [[1]](diffhunk://#diff-36d1b059149fc82438a9db6a5cfac2fd4c436264fbceb947544d9ae598b5bcb4L159-L160) [[2]](diffhunk://#diff-7b07624ead7f8be9a6b4094f7944fc2148c8e913eafc9470c87030b80863e51fL457) [[3]](diffhunk://#diff-d1502364216dc3b2ed3e8a355337f1b1c91755516ee2156fcb04317672af1130L634) [[4]](diffhunk://#diff-d1502364216dc3b2ed3e8a355337f1b1c91755516ee2156fcb04317672af1130L644)

Barrier command and state management:
* Removed the `Pause` and `Resume` variants from the `Command` enum, as well as all related logic, display formatting, and helper methods. [[1]](diffhunk://#diff-30b031080778ca36f6e37c4769dd8941f9555643e0bab4f357c417a5093efc19L294-L302) [[2]](diffhunk://#diff-30b031080778ca36f6e37c4769dd8941f9555643e0bab4f357c417a5093efc19L400-L401) [[3]](diffhunk://#diff-30b031080778ca36f6e37c4769dd8941f9555643e0bab4f357c417a5093efc19L456-L463) [[4]](diffhunk://#diff-30b031080778ca36f6e37c4769dd8941f9555643e0bab4f357c417a5093efc19L473-L474) [[5]](diffhunk://#diff-30b031080778ca36f6e37c4769dd8941f9555643e0bab4f357c417a5093efc19L617-L621) [[6]](diffhunk://#diff-30b031080778ca36f6e37c4769dd8941f9555643e0bab4f357c417a5093efc19L846-L872) [[7]](diffhunk://#diff-30b031080778ca36f6e37c4769dd8941f9555643e0bab4f357c417a5093efc19L1011-L1012) [[8]](diffhunk://#diff-30b031080778ca36f6e37c4769dd8941f9555643e0bab4f357c417a5093efc19L1244) [[9]](diffhunk://#diff-de9637b343bd5dc38b424a066bc5a13cb98b1a2c26038309829e4c7447ab8607L268-L271) [[10]](diffhunk://#diff-e11afa43355e6c4dd8488b3c725ffd7db527882de703c2ca2c53acc79b9054d2L1001-L1002)
* Removed the `is_paused` field and related methods from `BarrierWorkerState` and `DatabaseCheckpointControl`, and eliminated pause state checks throughout the codebase. [[1]](diffhunk://#diff-44229b137399d20ef70c07c7f928527ec494356b08b98ac0e68c6ccf7b3371b0L53-L86) [[2]](diffhunk://#diff-44229b137399d20ef70c07c7f928527ec494356b08b98ac0e68c6ccf7b3371b0L164) [[3]](diffhunk://#diff-44229b137399d20ef70c07c7f928527ec494356b08b98ac0e68c6ccf7b3371b0L266-L276) [[4]](diffhunk://#diff-5f6b313dd47a23ea5663db1b68de3ed3938e8e5829e484c4b76c79afbfdfcec9L1020-L1035)

Pause/resume handling at the barrier manager level:
* Added `pause` and `resume` methods to `GlobalBarrierManager`, and new `BarrierManagerRequest` variants, so pausing and resuming are now managed at the barrier manager level rather than through mutations. Updated the service layer to use these new methods. [[1]](diffhunk://#diff-6ae613e0313025924be82ea866d0ff355c3b5c8825b61f487227225c908b2939R99-R116) [[2]](diffhunk://#diff-c7d2fe56d7d1503b969b97b4e7503738a8778d351960e87745c32bbc164b217dR117-R118) [[3]](diffhunk://#diff-66d82e7ca80b658c018a0a5f6407e235d890ba46ecc677cdf83a626f4af7dde7L120-R125)

General code cleanup:
* Removed unused imports and parameters related to the old pause/resume logic. [[1]](diffhunk://#diff-30b031080778ca36f6e37c4769dd8941f9555643e0bab4f357c417a5093efc19L48-R50) [[2]](diffhunk://#diff-d1502364216dc3b2ed3e8a355337f1b1c91755516ee2156fcb04317672af1130L29) [[3]](diffhunk://#diff-d1502364216dc3b2ed3e8a355337f1b1c91755516ee2156fcb04317672af1130L612)

These changes streamline the barrier and mutation system by removing the need for pause/resume mutations, reducing complexity, and centralizing pause/resume control.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/meta/src/barrier/checkpoint/control.rs
+++ b/src/meta/src/barrier/checkpoint/control.rs
@@ -231,7 +231,9 @@ impl CheckpointControl {
                             .insert(DatabaseCheckpointControlStatus::Running(new_database))
                             .expect_running("just initialized as running")
                     }
-                    Command::Flush | Command::Pause | Command::Resume => {
+                    Command::Flush
+                    | Command::DropStreamingJobs { .. }
+                    | Command::DropSubscription { .. } => {
                         for mut notifier in notifiers {
                             notifier.notify_started();
                             notifier.notify_collected();
@@ -1016,22 +1018,6 @@ impl DatabaseCheckpointControl {
             }
             return Ok(());
         };
-
-        if let Some(Command::CreateStreamingJob {
-            job_type: CreateStreamingJobType::SnapshotBackfill(_),
-            ..
-        }) = &command
-            && self.state.is_paused()
-        {
-            warn!("cannot create streaming job with snapshot backfill when paused");
-            for notifier in notifiers {
-                notifier.notify_start_failed(
-                    anyhow!("cannot create streaming job with snapshot backfill when paused",)
-                        .into(),
-                );
-            }
-            return Ok(());
-        }
 
         let barrier_info = self.state.next_barrier_info(checkpoint, curr_epoch);
         // Tracing related stuff
--- a/src/meta/src/barrier/worker.rs
+++ b/src/meta/src/barrier/worker.rs
@@ -91,6 +91,8 @@ pub(super) struct GlobalBarrierWorker<C> {
     control_stream_manager: ControlStreamManager,
 
     term_id: String,
+
+    paused: bool,
 }
 
 impl<C: GlobalBarrierWorkerContext> GlobalBarrierWorker<C> {
@@ -111,6 +113,9 @@ impl<C: GlobalBarrierWorkerContext> GlobalBarrierWorker<C> {
         let periodic_barriers = PeriodicBarriers::default();
 
         let checkpoint_control = CheckpointControl::new(env.clone());
+
+        let paused = Self::take_pause_on_bootstrap(&env).await;
+
         Self {
             enable_recovery,
             periodic_barriers,
@@ -123,6 +128,7 @@ impl<C: GlobalBarrierWorkerContext> GlobalBarrierWorker<C> {
             active_streaming_nodes,
             control_stream_manager,
             term_id: "uninitialized".into(),
+            paused,
         }
     }
 }
@@ -168,15 +174,13 @@ impl GlobalBarrierWorker<GlobalBarrierWorkerContextImpl> {
 
         (join_handle, shutdown_tx)
     }
+}
 
+impl<C> GlobalBarrierWorker<C> {
     /// Check whether we should pause on bootstrap from the system parameter and reset it.
-    async fn take_pause_on_bootstrap(&mut self) -> MetaResult<bool> {
-        let paused = self
-            .env
-            .system_params_reader()
-            .await
-            .pause_on_next_bootstrap()
-            || self.env.opts.pause_on_next_bootstrap_offline;
+    async fn take_pause_on_bootstrap(env: &MetaSrvEnv) -> bool {
+        let paused = env.system_params_reader().await.pause_on_next_bootstrap()
+            || env.opts.pause_on_next_bootstrap_offline;
 
         if paused {
             warn!(
@@ -185,14 +189,19 @@ impl GlobalBarrierWorker<GlobalBarrierWorkerContextImpl> {
                  To resume the data sources, either restart the cluster again or use `risectl meta resume`.",
                 PAUSE_ON_NEXT_BOOTSTRAP_KEY
             );
-            self.env
+            if let Err(e) = env
                 .system_params_manager_impl_ref()
                 .set_param(PAUSE_ON_NEXT_BOOTSTRAP_KEY, Some("false".to_owned()))
-                .await?;
+                .await
+            {
+                warn!(e = %e.as_report(), "failed to reset PAUSE_ON_NEXT_BOOTSTRAP_KEY");
+            }
         }
-        Ok(paused)
+        paused
     }
+}
 
+impl GlobalBarrierWorker<GlobalBarrierWorkerContextImpl> {
     /// Start an infinite loop to take scheduled barriers and send them.
     async fn run(mut self, shutdown_rx: Receiver<()>) {
         tracing::info!(
@@ -232,9 +241,7 @@ impl GlobalBarrierWorker<GlobalBarrierWorkerContextImpl> {
                 None,
             );
 
-            let paused = self.take_pause_on_bootstrap().await.unwrap_or(false);
-
-            self.recovery(paused, RecoveryReason::Bootstrap)
+            self.recovery(RecoveryReason::Bootstrap)
                 .instrument(span)
                 .await;
         }
@@ -313,6 +320,20 @@ impl<C: GlobalBarrierWorkerContext> GlobalBarrierWorker<C> {
                                     warn!("failed to notify finish of update database barrier");
                                 }
                             }
+                            BarrierManagerRequest::Pause(tx) => {
+                                if !self.paused {
+                                    self.paused = true;
+                                    self.recovery(RecoveryReason::Adhoc).await;
+                                }
+                                let _ = tx.send(());
+                            }
+                            BarrierManagerRequest::Resume(tx) => {
+                                if self.paused {
+                                    self.paused = false;
+                                    self.recovery(RecoveryReason::Adhoc).await;
+                                }
+                                let _ = tx.send(());
+                            }
                         }
                     } else {
                         tracing::info!("end of request stream. meta node may be shutting down. Stop global barrier manager");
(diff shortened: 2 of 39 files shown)


wenym1:
* **#24261** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24261?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24261?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

wenym1:
resolve https://github.com/risingwavelabs/risingwave/issues/24199 by the way.

src/meta/src/barrier/checkpoint/control.rs
```diff
@@ -231,7 +231,9 @@ impl CheckpointControl {
                             .insert(DatabaseCheckpointControlStatus::Running(new_database))
                             .expect_running("just initialized as running")
                     }
-                    Command::Flush | Command::Pause | Command::Resume => {
+                    Command::Flush
+                    | Command::DropStreamingJobs { .. }
```
wenym1, line 235: newly added logic here resolves https://github.com/risingwavelabs/risingwave/issues/24199

src/meta/src/barrier/worker.rs
```diff
@@ -313,14 +320,28 @@ impl<C: GlobalBarrierWorkerContext> GlobalBarrierWorker<C> {
                                     warn!("failed to notify finish of update database barrier");
                                 }
                             }
+                            BarrierManagerRequest::Pause(tx) => {
+                                if !self.paused {
+                                    self.paused = true;
+                                    self.recovery(RecoveryReason::Adhoc).await;
```
BugenZhao, line 326: I think it would be better to introduce a new kind of `RecoveryReason`. Currently `Adhoc` is specific to `RECOVER` command from users.

src/meta/src/barrier/worker.rs
```diff
@@ -185,14 +189,19 @@ impl GlobalBarrierWorker<GlobalBarrierWorkerContextImpl> {
...
                 .system_params_manager_impl_ref()
                 .set_param(PAUSE_ON_NEXT_BOOTSTRAP_KEY, Some("false".to_owned()))
-                .await?;
+                .await
+            {
+                warn!(e = %e.as_report(), "failed to reset PAUSE_ON_NEXT_BOOTSTRAP_KEY");
+            }
         }
-        Ok(paused)
+        paused
```
BugenZhao, line 200: Why don't we return an error here any more?

src/meta/src/barrier/worker.rs
```diff
@@ -185,14 +189,19 @@ impl GlobalBarrierWorker<GlobalBarrierWorkerContextImpl> {
...
                 PAUSE_ON_NEXT_BOOTSTRAP_KEY
             );
-            self.env
+            if let Err(e) = env
                 .system_params_manager_impl_ref()
                 .set_param(PAUSE_ON_NEXT_BOOTSTRAP_KEY, Some("false".to_owned()))
-                .await?;
+                .await
+            {
+                warn!(e = %e.as_report(), "failed to reset PAUSE_ON_NEXT_BOOTSTRAP_KEY");
```
BugenZhao, line 197:
```suggestion
                warn!(e = %e.as_report(), "failed to reset {PAUSE_ON_NEXT_BOOTSTRAP_KEY}");
```

src/meta/src/barrier/worker.rs
```diff
@@ -735,14 +754,18 @@ impl<C: GlobalBarrierWorkerContext> GlobalBarrierWorker<C> {
...
         });
     }
 
     #[await_tree::instrument("recovery({recovery_reason})")]
-    async fn recovery_inner(&mut self, is_paused: bool, recovery_reason: String) {
+    async fn recovery_inner(&mut self, recovery_reason: String) {
+        if self.paused {
+            info!("skip recovery when paused");
+            return;
+        }
```
BugenZhao, line 768: Wouldn't this skip the recovery right after setting `paused` to `true`?

BugenZhao (commented):
So basically we're deprecating the capability of "pausing the data stream but preserving the control stream" on active executors, but achieve the similar effect by dropping all actors on compute nodes and leave them idle?

I'm not sure if this works in all cases, i.e., if some other functionalities or use pattern relies on an active-yet-paused pipeline. For example, when paused with the new approach, is it still possible to

- scale jobs
- create new jobs atop of existing jobs

cc @hzxa21 @shanicky 

Also, if we're going to proceed such approach, I would suggest completely removing the test-only `Pause` and `Resume` mutation on compute node to avoid confusion.

src/meta/src/barrier/worker.rs
wenym1: The only usage on this method is `take_pause_on_bootstrap().unwrap_or(false)`, which means even `pause_on_next_bootstrap` is enabled, if we failed to reset the system param to false in meta store table, the pause will be disabled. The change here can still enable the pause in this scenario.

src/meta/src/barrier/worker.rs
wenym1: This is as expected, because I don't want to go through a recovery when we receive a pause request, so that we can avoid triggering panic in recovery, which happened sometimes and caused meta node crash loop during which we are not able to handle meta request.

Nevertheless, the recovery triggered on pause request seems unnecessary in current implementation, and I will remove it.

wenym1:
> For example, when paused with the new approach, is it still possible to
> 
> scale jobs
> create new jobs atop of existing jobs

For `scale jobs`, our current scale implementation does not depend on pause anymore after https://github.com/risingwavelabs/risingwave/issues/18312 is supported.

For `create new jobs atop of existing jobs`, we can pause the source consumption by setting source rate limit to 1 as an alternative.

hzxa21:
> For `scale jobs`, our current scale implementation does not depend on pause anymore after #18312 is supported.
> 
> For `create new jobs atop of existing jobs`, we can pause the source consumption by setting source rate limit to 1 as an alternative.

For `scale jobs`, I think what @BugenZhao means is to pause all traffic in the cluster to make sure barrier can flow through, then scale the jobs without triggering recovery.

For `create new jobs atop of existing jobs`, I think you mean setting source rate limit to 0, right? A typical use case for this is to do manual replace sink without data loss:
1. Pause all traffic to the cluster.
2. Call flush to make sure ongoing data are processed by the sink.
3. Drop the old sink.
4. Create new sink with `snapshot = false`

Some followup questions:
- Other than pausing/resuming source traffic, did the pre-existing pause/resume mutation have other side effects?
- Do we support pausing all sources using one command?

BugenZhao:
> For `scale jobs`, our current scale implementation does not depend on pause anymore after #18312 is supported.

I mean from users' perspective, is it still possible to run `ALTER PARALLELISM` at such state?

src/meta/src/barrier/worker.rs
BugenZhao: > Nevertheless, the recovery triggered on pause request seems unnecessary in current implementation, and I will remove it.

🤯 I guess I may not get the point. So how exactly will we "pause"?

wenym1:
Suspend this PR for the time being, because the original pause is still in use case like replace sink, and the current PR cannot fully replace it yet.

PR #24263, opened 2025-12-25
fix: check column name duplication in select

### LINKED ISSUE #24095: SELECT statement may return ambiguous column names without error
### Describe the bug

A `SELECT` statement can contain duplicate column names.  This is normally forbidden by PostgreSQL, but RisingWave seems to allow it.

### Error message/log

```text
I don't see any error message from RisingWave, even though I would expect one.
```

### To Reproduce

Try running the following statement in both RisingWave and PostgreSQL.  Observe that the CTE contains an ambiguous column name of `col`.

```
with cte as (
  select
    1 as col,
    2 as col
)
 
select cte.col from cte
```

### Expected behavior

If you run this statement in PostgreSQL, you will get an error: `column reference "col" is ambiguous`

I would expect this error to happen in RisingWave.  Instead, the query runs in RisingWave without error, and returns a row with value of `1`.

(For comparison with non-PG database engines, I tried running this statement in BigQuery.  BigQuery also returns an error about ambiguous column name.)

RisingWave Cloud, with image `v2.6.3--unlabeled--ccd1dfa--release-2.6`

### The version of RisingWave

`PostgreSQL 13.14.0-RisingWave-2.6.3 (ccd1dfa8c28a6b63beff320eb6d68c6e8cd79163)`

### Additional context

_No response_

hzxa21:

Resolves #24095.

The fix is generated by pair programming with copilot.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/binder/expr/column.rs
+++ b/src/frontend/src/binder/expr/column.rs
@@ -15,7 +15,7 @@
 use risingwave_common::types::DataType;
 use risingwave_sqlparser::ast::Ident;
 
-use crate::binder::{Binder, Clause};
+use crate::binder::{Binder, Clause, COLUMN_GROUP_PREFIX};
 use crate::error::{ErrorCode, Result};
 use crate::expr::{CorrelatedInputRef, ExprImpl, ExprType, FunctionCall, InputRef, Literal};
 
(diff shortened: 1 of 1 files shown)


e2e_test/batch/basic/duplicate_column_name.slt
```diff
@@ -0,0 +1,73 @@
...
+with cte as (
+  select
+    1 as col,
+    2 as col
+)
+select cte.col from cte;
+
+# Test 2: Duplicate column names in simple SELECT
+statement error
+select 1 as col, 2 as col;
```
chenzl25, line 21: duplicate column names in simple select is allowed in Postgres. But it is only not allowed when we want to refer to one of them.

src/frontend/src/binder/expr/column.rs
```diff
@@ -44,38 +44,66 @@ impl Binder {
...
-                        );
+        // For qualified references (with table name), check if it's a column group reference.
+        // Column groups (from natural joins) can have duplicate column names and use COALESCE.
+        // For regular qualified references, error on ambiguous columns.
+        let is_column_group = table_name
+            .as_ref()
+            .map(|t| t.starts_with(COLUMN_GROUP_PREFIX))
+            .unwrap_or(false);
+
+        if table_name.is_some() && !is_column_group {
```
chenzl25, line 55: @xiangjinwu I am not 100% sure whether this handling is correct, WDYT?

PR #24269, opened 2025-12-26
fix(meta): fix race condition during metadata backup

zwang28:

This PR resolves a race condition causing concurrent metadata snapshot manifest operations to overwrite one another.
The impact of the race condition is, when **concurrent** create/delete backup CTL commands are executed and all return successfully,
- The newly created metadata snapshot might not be included in the latest manifest, i.e. the create command doesn't succeed.
- The newly deleted metadata snapshots might still be included in the latest manifest, i.e. the delete command doesn't succeed.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/storage/backup/src/storage.rs
+++ b/src/storage/backup/src/storage.rs
@@ -71,12 +71,18 @@ impl ObjectStoreMetaSnapshotStorage {
         Ok(instance)
     }
 
-    async fn update_manifest(&self, new_manifest: &MetaSnapshotManifest) -> BackupResult<()> {
+    async fn update_manifest(
+        &self,
+        update: impl FnOnce(MetaSnapshotManifest) -> MetaSnapshotManifest,
+    ) -> BackupResult<()> {
+        let mut guard = self.manifest.write().await;
+        let new_manifest = update((**guard).clone());
         let bytes =
-            serde_json::to_vec(new_manifest).map_err(|e| BackupError::Encoding(e.into()))?;
+            serde_json::to_vec(&new_manifest).map_err(|e| BackupError::Encoding(e.into()))?;
         self.store
             .upload(&self.get_manifest_path(), bytes.into())
             .await?;
+        *guard = Arc::new(new_manifest);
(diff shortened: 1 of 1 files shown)


src/storage/backup/src/storage.rs
```diff
@@ -126,7 +126,8 @@ impl MetaSnapshotStorage for ObjectStoreMetaSnapshotStorage {
         self.store.upload(&path, snapshot.encode()?.into()).await?;
 
         // update manifest last
-        let mut new_manifest = (**self.manifest.read()).clone();
+        let mut gurad = self.manifest.write().await;
```
BugenZhao:
```suggestion
        let mut guard = self.manifest.write().await;
```

src/storage/backup/src/storage.rs
```diff
@@ -163,13 +165,16 @@ impl MetaSnapshotStorage for ObjectStoreMetaSnapshotStorage {
...
         let to_delete: HashSet<MetaSnapshotId> = HashSet::from_iter(ids.iter().cloned());
-        let mut new_manifest = (**self.manifest.read()).clone();
-        new_manifest.manifest_id += 1;
-        new_manifest
-            .snapshot_metadata
-            .retain(|m| !to_delete.contains(&m.id));
-        self.update_manifest(new_manifest).await?;
-
+        {
+            let mut gurad = self.manifest.write().await;
```
BugenZhao:
```suggestion
            let mut guard = self.manifest.write().await;
```

src/storage/backup/src/storage.rs
zwang28: resolved.

src/storage/backup/src/storage.rs
zwang28: resolved.

zwang28:
@BugenZhao @hzxa21 @yezizp2012 Could you help review this PR when available? Thank you.

BugenZhao (approved):
LGTM

Issue #23341, opened 2025-09-29
apply downstream sink schema change after checkpoint


wenym1:
### Background

Currently, our sink schema change works like

trigger sink schema change
-> sink executor receives barrier from upstream and passes barrier to concrete sink
-> concrete sink applies downstream schema change
-> concrete sink ack on finish processing the schema change
-> sink executor yield the barrier
-> collect barrier from all actors, commit to state store, and update the sink catalog with schema change, and finish

The current mechanism has a limitation. Let's say originally we have two columns `(col1, col2)` in the sink and its upstream table in `epoch1`, and we are going to add a new column `col3` after `epoch2`. If any failure happens after applying schema change, we will revert to the progress to `epoch1`, and replay the old data from `epoch1` to `epoch2` with schema `(col1, col2)` to write the external sink. However, the external sink has been applied with the schema change to add a new `col3`, and then there will be unrecoverable error if the sink does not support writing a subset of its columns (write data of `(col1, col2)`, while its current schema is already `(col1, col2, col3)`).

### Design

The current limitation is caused by the incorrect order of downstream schema change and internal checkpoint. If we want to ensure that data of old schema won't be written to the new schema, we should ensure that when we apply downstream schema change, all previous data won't be replayed again, which means we should first finish internal checkpoint to persist the progress, and then apply the schema change.

However, if we successfully finish the internal checkpoint, but failed to apply the schema change and trigger the recovery, the information of schema change won't be replayed, and we will lose the schema change, and then we will on the contrary write data of new schema on external sink of old schema. As a result, we should also persist the information of schema before the internal checkpoint.

The new procedure of sink schema change will be like

trigger sink schema change
-> sink executor receives barrier from upstream and passes barrier to concrete sink
-> concrete sink persists the schema change information in meta store table
-> concrete sink ack on finish processing the barrier
-> sink executor yield the barrier
-> collect barrier from all actors, commit to state store, and update the sink catalog with schema change
-> concrete sink waits for state store commit, and once committed, apply the schema change to downstream, and then clear the persisted record

We should also properly handle recovery. The persisted schema change information should be associated to the epoch of the barrier. During recovery, we should clean up the entries of dirty epochs in the meta store table. After sink executors restart, we should load the pending entries from meta store table, and reapply the schema change.  The schema change should be applied in an idenpotent way.

The design follows a similar workflow to a new design and implementation of exactly-once sink described in https://github.com/risingwavelabs/risingwave/issues/23394.

PR #24271, opened 2025-12-26
feat(sink): Sink schema change from barrier to persistent

Chiro11:

This PR refactors the sink schema change mechanism to support more flexible schema change operations beyond just adding columns. The changes include:

1. Replaced the `SinkAddColumns` message with a more general `SinkSchemaChange` message that can support different schema change operations (add columns, drop columns, etc.)
2. Updated the protocol buffers to include operation-specific messages like `SinkAddColumnsOp` and `SinkDropColumnsOp`
3. Added a database migration to store schema change information in the `PendingSinkState` table
4. Modified the two-phase commit protocol to properly handle schema changes during commit operations
5. Updated sink implementations (Iceberg, Redshift, Snowflake) to work with the new schema change format
6. Added proper error handling for unsupported schema change operations in various sink types

These changes provide a more robust foundation for implementing different types of schema changes in sinks while maintaining backward compatibility with existing functionality.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/proto/stream_plan.proto
+++ b/proto/stream_plan.proto
@@ -1285,6 +1285,8 @@ message StreamFragmentGraph {
 
 // Schema change operation for sink
 message SinkSchemaChange {
+  // Original schema before this change.
+  // Used for validation and conflict detection.
   repeated plan_common.Field original_schema = 1;
 
   // Schema change operation (mutually exclusive)


Chiro11:
* **#24277** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24277?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24273** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24273?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24271** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24271?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24271?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

proto/stream_plan.proto
```diff
@@ -1289,18 +1285,15 @@ message StreamFragmentGraph {
 
 // Schema change operation for sink
 message SinkSchemaChange {
-  // Original schema before this change (optional, can be empty)
```
wenym1, line 1292: Can keep this comment, and only remove the last line.

src/connector/src/sink/mod.rs
```diff
@@ -827,7 +828,7 @@ pub trait SinglePhaseCommitCoordinator {
         &mut self,
         epoch: u64,
         metadata: Vec<SinkMetadata>,
-        add_columns: Option<Vec<Field>>,
+        schema_change: Option<PbSinkSchemaChange>,
```
wenym1, line 831: How are we going to handle the sink schema change of `SinglePhaseCommitCoordinator` in the same mechanism? Under current implementation, schema change is still handled before yielding barrier and the later checkpoint.

Chiro11: Yes, single-phase cannot guarantee consistency. Maybe we need to check on ddl? When creating a sink with `auto_schema_change=true`, ensure that the sink has enabled the `exactly_once` option (currently only supported by iceberg).

wenym1: Discussed offline. As described in https://github.com/risingwavelabs/risingwave/issues/23341, this mechanism may not depend on exactly-once semantic, and also works for single phase coordinator.

Chiro11: The api will be split in another PR.

PR #24274, opened 2025-12-26
feat: shorten_block_smallest_key

Li0k:

### Summary

This PR introduces an optimization to reduce SSTable metadata size by shortening `block_meta.smallest_key` using LCP (Longest Common Prefix) based compression.

### Changes

- Add `try_shorten_block_smallest_key()` function that finds the shortest key satisfying `prev_block_last < shortened_key <= block_smallest`
- Add `shorten_block_meta_key_threshold` configuration option (`Option<usize>`):
  - `None`: disabled (default)
  - `Some(0)`: always shorten
  - `Some(n)`: only shorten keys with length >= n bytes
- Configuration flows through: `StorageConfig` → `StorageOpts` → `SstableBuilderOptions`
- Add unit tests for the shortening algorithm

### Usage

```toml
[storage]
# Shorten keys >= 64 bytes
shorten_block_meta_key_threshold = 64
```

## Checklist

## Documentation

<details>

</details>


src/storage/src/hummock/sstable/block.rs
```diff
@@ -811,6 +811,55 @@ impl BlockBuilder {
...
+        for i in 0..min_len {
+            if a[i] != b[i] {
+                return i;
+            }
+        }
+        min_len
+    }
+
+    let prev_table_key = prev_block_last.user_key.table_key.as_ref();
+    let next_table_key = block_smallest.user_key.table_key.as_ref();
```
hzxa21, line 854: We assume the input keys share the same table id here. Let's add an assertion.

src/storage/src/hummock/sstable/block.rs
```diff
@@ -811,6 +811,55 @@ impl BlockBuilder {
...
+
+    // Build candidate key: take LCP + 1 bytes from block_smallest
+    let cand = FullKey::new_with_gap_epoch(
+        block_smallest.user_key.table_id,
+        TableKey(&next_table_key[..shortened_len]),
+        block_smallest.epoch_with_gap,
+    );
+
+    // Verify invariant: prev_block_last < cand <= block_smallest
+    if prev_block_last.cmp(&cand).is_ge() || cand.cmp(block_smallest).is_gt() {
```
hzxa21: Shouldn't this be an assertion? If we hit the condition here, that means the algorithm is wrongly implemented.

src/storage/src/hummock/sstable/builder.rs
```diff
@@ -353,6 +358,21 @@ impl<W: SstableWriter, F: FilterBuilder> SstableBuilder<W, F> {
 
         // Rotate block builder if the previous one has been built.
         if self.block_builder.is_empty() {
+            let smallest_key = if let Some(threshold) =
+                self.options.shorten_block_meta_key_threshold
+                && !self.last_full_key.is_empty()
```
hzxa21, line 363: I think the algorithm can still wotk if last_full_key is empty (i.e. on the 1st block)

hzxa21 (approved):

src/storage/src/hummock/sstable/builder.rs
Li0k: Yes, your observation is very astute. However, the **smallest-key** of the first block-meta will be used to determine the left-key of sstable-info. I do not want to modify the definition of the key-range of sstable-info in this PR, because I cannot assess how much impact the change in key-range precision will have on the compaction picker. Therefore, keeping the current implementation is my choice, as it is more concise. Of course, as you mentioned, we can generate this first-key using a virtual-key-with-table-id.

WDYT @hzxa21

hzxa21: LGTM. We can keep it as it is for this PR. We can worry about it when considering how to shorten the sst key range.

PR #24276, opened 2025-12-26
refactor: extract clients and server impls of `MonitorService` & impl for frontend and compactors

BugenZhao:

We have a `MonitorService`, of which many calls are about general CPU or memory profiling that is not specific to any node types.

https://github.com/risingwavelabs/risingwave/blob/f27143a88f16aa8d2f95c0e443f67608e389a290/proto/monitor_service.proto#L136-L145

However, previously the client for `MonitorService` is a part of `ComputeClient`, and the server part is only implemented in compute node.

This PR extracts these components into `MonitorClient` and `ProfileServiceImpl` so that they are not specific to compute node. Based on that, we also implement the profiling calls in `MonitorService` for frontend nodes and compactor nodes.

## Checklist

## Documentation

<details>

</details>


hzxa21:
Will we have multiple monitor service running for a standalone process? If yes, will they conflict with one another?

hzxa21:
Will we have multiple monitor service running for a standalone process? If yes, will they conflict with one another?

BugenZhao:
> Will we have multiple monitor service running for a standalone process? If yes, will they conflict with one another?

We will have multiple `tonic` services operating behind various ports for each worker node role. However, since these services are essentially handlers and stateless, there won't be any conflicts.

The only long-running service is the `HeapProfiler` that automatically dumps heap when the memory usage reaches some threshold. It has already been gated behind `Once` so there's also no issue.

BugenZhao:
Bump. This PR is necessary for #24278 to work. Most changes are simply code move.

src/rpc_client/src/monitor_client.rs
```diff
@@ -0,0 +1,148 @@
...
+        Ok(self
+            .monitor_client
+            .clone()
+            .heap_profiling(HeapProfilingRequest { dir })
+            .await
+            .map_err(RpcError::from_monitor_status)?
+            .into_inner())
+    }
+
+    /// Available on compute node, compactor node, and frontend node.
```
kwannoel, line 117: How come not on the meta-node? Or do you mean to support it incrementally?

src/rpc_client/src/monitor_client.rs
```diff
@@ -0,0 +1,148 @@
...
+        Ok(self
+            .monitor_client
+            .clone()
+            .get_profile_stats(request)
+            .await
+            .map_err(RpcError::from_monitor_status)?
+            .into_inner())
+    }
+
+    /// Available on compute node, compactor node, and frontend node.
```
kwannoel, line 95: As below, how come not on the meta node as well

src/rpc_client/src/monitor_client.rs
```diff
@@ -0,0 +1,148 @@
...
+            .monitor_client
+            .clone()
+            .stack_trace(req)
+            .await
+            .map_err(RpcError::from_monitor_status)?
+            .into_inner())
+    }
+
+    /// Available on compute node.
+    pub async fn get_streaming_stats(&self) -> Result<GetStreamingStatsResponse> {
```
kwannoel, line 71: What happens if we call this API on the meta / compactor / frontend nodes?

kwannoel: Oh I guess it will just query the CN

kwannoel (approved):
LGTM.

src/rpc_client/src/monitor_client.rs
BugenZhao: Because for all other worker nodes types, the functionality is done via RPC, while for meta node it's the same node so we may need a slightly different entrance for that. Let's get this done in following PRs.

src/common/common_service/src/profile_service.rs
```diff
@@ -0,0 +1,183 @@
+// Copyright 2025 RisingWave Labs
```
BugenZhao:
```suggestion
// Copyright 2026 RisingWave Labs
```

src/frontend/src/rpc/monitor_service.rs
```diff
@@ -0,0 +1,97 @@
+// Copyright 2025 RisingWave Labs
```
BugenZhao:
```suggestion
// Copyright 2026 RisingWave Labs
```

src/rpc_client/src/monitor_client.rs
```diff
@@ -0,0 +1,148 @@
+// Copyright 2025 RisingWave Labs
```
BugenZhao:
```suggestion
// Copyright 2026 RisingWave Labs
```

PR #24288, opened 2025-12-29
refactor(sink): Split sink schema change to use two-phase commit

Chiro11:

Split
```
async fn commit(
    &mut self,
    epoch: u64,
    commit_metadata: Vec<u8>,
    schema_change: Option<PbSinkSchemaChange>,
) -> Result<()>;
```
into:
```
async fn commit_data(&mut self, epoch: u64, metadata: Vec<SinkMetadata>) -> Result<()>;

async fn commit_schema_change(
        &mut self,
        epoch: u64,
        schema_change: PbSinkSchemaChange,
    ) -> Result<()>;
```
for both `SinglePhaseCommitCoordinator` and `TwoPhaseCommitCoordinator`.

Migrate schema change for **Snowflake**, **Redshift**, **Iceberg** to this new interface.

For `SinglePhaseCommitCoordinator`, if there is a schema change, we will have the schema change info go through a `two-phase` commit process separately after `commit_data`, ensuring consistency of schema change even in single-phase scenario.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/.github/workflows/license_check.yml
+++ b/.github/workflows/license_check.yml
@@ -23,11 +23,12 @@ jobs:
       with:
         tool: hawkeye@6.3
     - name: Check License Header
-      run: hawkeye check
-    - name: Hint to fix license headers
-      if: failure()
       run: |
-        echo "::error::License header check failed."
-        echo "Hint: rebase latest main, then auto-fix:"
-        echo "  git fetch origin main && git rebase origin/main"
-        echo "  ./risedev fix-license-header"
+        hawkeye check || {
+          echo "::error::License header check failed."
+          echo "Note: the copyright year is now required to match the file's creation date."
+          echo "Hint: rebase latest main, then auto-fix:"
+          echo "  git fetch origin main && git rebase origin/main"
+          echo "  ./risedev fix-license-header"
+          exit 1
+        }
--- a/src/expr/impl/src/scalar/array_reverse.rs
+++ b/src/expr/impl/src/scalar/array_reverse.rs
@@ -1,4 +1,4 @@
-// Copyright 2025 RisingWave Labs
+// Copyright 2026 RisingWave Labs
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/src/frontend/src/datafusion/window_function.rs
+++ b/src/frontend/src/datafusion/window_function.rs
@@ -1,4 +1,4 @@
-// Copyright 2025 RisingWave Labs
+// Copyright 2026 RisingWave Labs
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/src/frontend/src/optimizer/plan_visitor/datafusion_execute_checker.rs
+++ b/src/frontend/src/optimizer/plan_visitor/datafusion_execute_checker.rs
@@ -1,4 +1,4 @@
-// Copyright 2025 RisingWave Labs
+// Copyright 2026 RisingWave Labs
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.


src/connector/src/sink/mod.rs
```diff
@@ -845,12 +848,15 @@ pub trait TwoPhaseCommitCoordinator {
...
     ) -> Result<Vec<u8>>;
 
-    /// Idempotent implementation is required, because `commit` in the same epoch could be called multiple times.
-    async fn commit(
+    /// Idempotent implementation is required, because `commit_data` in the same epoch could be called multiple times.
+    async fn commit_data(&mut self, epoch: u64, commit_metadata: Vec<u8>) -> Result<()>;
+
+    /// Idempotent implementation is required, because `commit_schema_change` in the same epoch could be called multiple
+    /// times.
+    async fn commit_schema_change(
```
wenym1, line 867: We can have a default implementation in the trait definition so that we don't have to impl it for every coordinator.

src/meta/src/manager/sink_coordination/coordinator_worker.rs
```diff
@@ -724,20 +704,27 @@ impl CoordinatorWorker {
                     let SinkCommitCoordinator::TwoPhase(coordinator) = &mut coordinator else {
```
wenym1: We will enter unreachable code for `SinkCommitCoordinator::SinglePhase`, and this is the root cause of the failure in e2e test.

src/meta/src/manager/sink_coordination/coordinator_worker.rs
```diff
@@ -724,20 +704,27 @@ impl CoordinatorWorker {
...
                         unreachable!("should be two-phase commit coordinator");
                     };
-                    let two_phase_handler = two_phase_handler.as_mut().expect("should exist");
                     let start_time = Instant::now();
-                    let commit_res = run_future_with_periodic_fn(
-                        coordinator.commit(epoch, metadata, schema_change),
-                        Duration::from_secs(5),
-                        || {
+                    let commit_fut = async {
+                        if !metadata.is_empty() {
```
wenym1: Instead of checking on empty `Vec<u8>`, shall we use `Option<Vec<u8>>` to represent whether there is metadata to commit? Our meta store table should support storing and retrieving `None`.

src/meta/src/manager/sink_coordination/coordinator_worker.rs
wenym1: It can be
```
match &mut coordinator {
    SinkCommitCoordinator::SinglePhase(coordinator) => {
        assert_eq!(metadata, None);
        coordinator.commit_schema_change(...).await;
    }
    SinkCommitCoordinator::TwoPhase(coordinator) => {
        coordinator.commit_data(...).await;
        coordinator.commit_schema_change(...).await;
    }
}
```

ci/scripts/e2e-iceberg-sink-cdc-test.sh
```diff
@@ -139,6 +139,7 @@ CREATE SINK s1 from t WITH (
     is_exactly_once = 'false',
 );"
 
+sleep 5
```
wenym1: It seems that we change here to try to fix e2e. But the root cause is actually the false unreachable mentioned in other comment.

wenym1 (commented):
Rest LGTM

ci/scripts/e2e-iceberg-sink-cdc-test.sh
Chiro11: Sorry, I have already fixed this error yesterday, but didn't push this commit.

src/meta/src/manager/sink_coordination/coordinator_worker.rs
Chiro11: The declaration in the migration file now is
```
ColumnDef::new(PendingSinkState::Metadata)
    .rw_binary(manager)
    .not_null(),
```
Refactoring also requires modifying the migration file, which is a bit troublesome.

wenym1: Fortunately, the migration file is not included in any release yet. We can remove the `not_null` in the migration file directly.

src/connector/src/sink/mod.rs
```diff
@@ -855,9 +859,13 @@ pub trait TwoPhaseCommitCoordinator {
...
         &mut self,
-        epoch: u64,
-        schema_change: PbSinkSchemaChange,
-    ) -> Result<()>;
+        _epoch: u64,
+        _schema_change: PbSinkSchemaChange,
+    ) -> Result<()> {
+        Err(SinkError::Coordinator(anyhow!(
+            "Schema change is not implemented for two-phase commit coordinator"
+        )))
```
wenym1, line 875:
```suggestion
        Err(SinkError::Coordinator(anyhow!(
            "Schema change is not implemented for two-phase commit coordinator {}", std::types::any::<Self>()
        )))
```

can include the type name in the error message.

src/connector/src/sink/iceberg/mod.rs
```diff
@@ -1876,32 +1876,38 @@ impl IcebergSinkCommitter {
...
-        }
+        Ok(())
+    }
+
+    async fn commit_schema_change(
+        &mut self,
+        epoch: u64,
+        schema_change: PbSinkSchemaChange,
+    ) -> Result<()> {
+        tracing::info!("Committing schema change in epoch {}", epoch);
```
wenym1: May also include the schema change information in the log.

src/meta/src/manager/sink_coordination/coordinator_worker.rs
```diff
@@ -788,47 +784,63 @@ impl CoordinatorWorker {
...
+                                first_schema_change.as_ref(),
+                            )
+                            .await?;
+                            two_phase_handler.push_new_item(
+                                epoch,
+                                commit_metadata,
+                                first_schema_change,
+                            );
+                        } else {
+                            // No data to commit and no schema change.
```
wenym1, line 836: `self.prev_committed_epoch` was not set previously. Was it a bug?

src/meta/src/manager/sink_coordination/coordinator_worker.rs
```diff
@@ -854,25 +866,29 @@ impl CoordinatorWorker {
             .map(|(epoch, _, metadata, schema_change)| (epoch, metadata, schema_change))
             .collect_vec();
 
-        let mut aborted_epochs = vec![];
+        if let SinkCommitCoordinator::TwoPhase(coordinator) = coordinator {
```
wenym1: The schema change epoch of single phase coordinator should also be aborted? Though single phase coordinator does not have `abort` method, we should call `clean_aborted_records` for these epochs.

src/meta/src/manager/sink_coordination/coordinator_worker.rs
Chiro11: Yes, the maintenance of `prev_committed_epoch` was inaccurate before, but using a smaller value may not affect correctness.

src/meta/src/manager/sink_coordination/coordinator_worker.rs
Chiro11: Good find.

PR #24286, opened 2025-12-29
feat(meta): support drop/cancel creating snapshot backfill job

wenym1:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## What's changed and what's your intention?

Previously we don't support dropping snapshot backfilling jobs because we drop jobs by issuing a barrier in the database streaming graph. However, the snapshot backfilling jobs have their separate streaming graphs, and therefore cannot follow the same mechanism. 

In this PR, we support dropping a snapshot backfill jobs by directly resetting the partial graph of the job, following the code path of resetting partial graph during database recovery. After the partial graph is reset, we notify the completion of the command.

### AI generated
This pull request introduces improvements and fixes to the snapshot backfill streaming job cancellation logic, adds better handling of job state transitions (especially for resetting jobs), and refines related test cases and CI configurations. The main focus is to ensure that snapshot backfill jobs can only be cancelled individually, and to robustly handle the "resetting" state for creating streaming jobs.

**Key changes include:**

### Streaming job cancellation and state handling

* Updated the logic in `DatabaseCheckpointControl` and `DatabaseCheckpointControlStatus` to only allow cancelling (dropping) a snapshot backfill streaming job if it is the only job being cancelled, and to provide clear error messages if a multi-job cancel is attempted. This prevents issues with partial graph resets during concurrent job cancellations.
[...]

Issue #13396, opened 2023-11-13
Discussion: Decouple cancel from barrier


kwannoel:
Support cancel of stream job without barrier. 
- Usually when high barrier latency, some stream job can be stuck, but due to high barrier latency, it cannot be cancelled either.
- Another case: When high barrier latency for first barrier, stream job cannot be created, but cannot be cancelled either, until first barrier pass. If we can decouple cancel from barrier, we can avoid this.

Just noting this down first, didn't think of any ideas on how to support it yet.

kwannoel:
We solve part of the problem, where on cancel, we will always remove the table fragment, such that if there's a recovery triggered, the cluster will clean the stream job still.

kwannoel:
IMO this will lead to a lot of friction for PoC users.

Having to trigger recovery is still high effort.

kwannoel:
Offline discussion with @yezizp2012:
- https://risingwave-labs.slack.com/archives/C0340PNL68P/p1701664338832329.

Highlight some parts:
- Q: What about returning immediately e.g. background_cancel / background_drop mv.
   A: The job itself can be causing high latency, we want to immediately drop actors w/o waiting.
       It will could also consume lots of cpu / mem still.
- Remark from @yezizp2012 :
  > The only way to solve this kind case is to force a recovery to clean and rebuild the whole streaming graph.
But in a normal environment, neither DROP nor CANCEL requests are due to the existence of a problematic job in most cases, and I don’t think we should break the design of modifying the streaming graph via barriers just to be compatible with such one. Maybe we can add a force recovery interface for such cases?

In my experience, it is common for users to experience high latency when they're prototyping queries. It's very common to create a query with high join amplification. (Also we should migrate the join matched metrics to the user dashboard perhaps).

Adding a recovery interface could be OK. But users may feel uncomfortable having to trigger a recovery each time barrier latency becomes high. But this is lower effort it seems. So we could support it first.

I would still like to explore the feasibility of decouple drop / cancel from barrier.

BugenZhao:
Also had a discussion with @fuyufjh earlier.

An intuitive approach could be directly dropping the tokio tasks of the involved actors when dropping a streaming job. However, from the view of the upstream actors, the closing of connections or channels during this procedure can be confused with a network failure, in which case the exception should be propagated instead. Thus, it's significant to exploit a mechanism that allows for graceful notification from the downstream to the upstream of the dropping. We are unsure yet if this will be easy to implement.

----

[...]

PR #24289, opened 2025-12-29
feat(stream): remove dispatcher on error to avoid recovery

wenym1:

Previously, we trigger recovery on dispatcher error. However, when we drop a streaming job, we may want to kill the downstream actors without barrier and not causing upstream errors, such as in https://github.com/risingwavelabs/risingwave/issues/13396#issuecomment-3323165971 and https://github.com/risingwavelabs/risingwave/pull/24286. Therefore, in this PR, we changed to ignore dispatcher error and remove the failed dispatcher in DispatchExecutor. 

For scenarios of unexpected channel error, we should receive error in downstream actor, and the downstream actors will trigger the recovery.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/storage/src/hummock/sstable/block.rs
+++ b/src/storage/src/hummock/sstable/block.rs
@@ -21,7 +21,7 @@ use std::ops::Range;
 use bytes::{Buf, BufMut, Bytes, BytesMut};
 use risingwave_common::catalog::TableId;
 use risingwave_hummock_sdk::KeyComparator;
-use risingwave_hummock_sdk::key::{FullKey, TABLE_PREFIX_LEN, TableKey};
+use risingwave_hummock_sdk::key::{FullKey, TableKey};
 use serde::{Deserialize, Serialize};
 
 use super::utils::{CompressionAlgorithm, bytes_diff_below_max_key_length, xxhash64_verify};
@@ -533,6 +533,7 @@ impl BlockBuilder {
 
     #[cfg(any(test, feature = "test"))]
     pub fn add_for_test(&mut self, full_key: FullKey<&[u8]>, value: &[u8]) {
+        use risingwave_hummock_sdk::key::TABLE_PREFIX_LEN;
         let input_table_id = full_key.user_key.table_id;
         match self.table_id {
             Some(current_table_id) => assert_eq!(current_table_id, input_table_id),


wenym1:
* **#24289** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24289?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24289?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

BugenZhao:
> For scenarios of unexpected channel error, we should receive error in downstream actor, and the downstream actors will trigger the recovery.

What if a network partition occurs between the downstream compute node and the other worker nodes (including the rest of compute nodes and the meta node)?

wenym1:
> > For scenarios of unexpected channel error, we should receive error in downstream actor, and the downstream actors will trigger the recovery.
> 
> What if a network partition occurs between the downstream compute node and the other worker nodes (including the rest of compute nodes and the meta node)?

The downstream compute node communicate with meta node with a long connected bidi-stream. On usual actor error, it reports error by sending a message in the bidi-stream to meta node to trigger the recover. When network partition occurs, the bidi-stream should be broken, and the meta node will be aware of the disconnection and trigger the recovery.

The term `unexpected channel error` in the PR description might be confusing. It actually means the exchange connection between upstream and downstream actors, no matter whether the connection is a remote one or a local one. 

This PR is based on the assumption that when a connection gets broken, both sides of the connection should be aware of the error. I think this assumption should hold in our system.

wenym1:
@BugenZhao The PR has passed all tests in main-cron. PTAL

src/stream/src/executor/dispatch.rs
```diff
@@ -207,38 +211,43 @@ impl DispatchExecutorInner {
...
+                    .await;
 
                 self.metrics
                     .actor_out_record_cnt
                     .inc_by(chunk.cardinality() as _);
             }
         }
+        self.dispatchers.retain(|dispatcher| match &**dispatcher {
+            DispatcherImpl::Failed(dispatcher_id, e) => {
+                warn!(%dispatcher_id, e = %e.as_report(), actor_id = %self.actor_id, "dispatch fail");
```
BugenZhao:
```suggestion
                warn!(%dispatcher_id, error = %e.as_report(), actor_id = %self.actor_id, "dispatch fail");
```

src/stream/src/executor/dispatch.rs
```diff
@@ -539,6 +554,7 @@ impl StreamConsumer for DispatchExecutor {
                             .await?;
                     }
                 }
+                consume_budget().await;
```
BugenZhao: `Output::send` will always `await` on semaphore primitive from `tokio` (unless it's a `Watermark` message, which is uncommon), is it still necessary to manually handle coop scheduling here?

src/stream/src/executor/dispatch.rs
```diff
@@ -602,11 +618,13 @@ async fn try_batch_barriers(
...
 
 #[derive(Debug)]
-pub enum DispatcherImpl {
+pub(crate) enum DispatcherImpl {
     Hash(HashDataDispatcher),
     Broadcast(BroadcastDispatcher),
     Simple(SimpleDispatcher),
+    #[cfg_attr(not(test), expect(dead_code))]
     RoundRobin(RoundRobinDataDispatcher),
+    Failed(DispatcherId, StreamError),
```
BugenZhao, line 635: What about adding some docs?

src/stream/src/executor/dispatch.rs
wenym1: Without it, the [buildkite/pull-request/slow-end-to-end-test](https://buildkite.com/risingwavelabs/pull-request/builds/88901#019b91d9-ca93-455a-ad5e-e1c5f8bf4aa9) will fail, because in recovery, when all downstream actors are killed, all dispatchers will be cleaned up because of dispatch error, and then there will be no call on `Output::send`. When we try to kill the actor via `join_handle.abort();join_handle.await;`, it won't work until the actor gets pending. In a test case of the slow e2e tests, it won't get pending until all messages have been consumed, which makes the recovery stuck for long time. 

I can change to only do `consume_budget().await;` when there is no dispatcher.

BugenZhao: It would be better if we can mark down these notes as comments 😄

PR #24284, opened 2025-12-29
feat: make decimal as datafusion native type

silver-ymz:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## What's changed and what's your intention?

- Enable fast-path execution for decimal types using DataFusion and add comprehensive tests.
- Implement function routing to leverage DataFusion’s math and Unicode functions.
- Truncate decimal values when converting from Arrow to RisingWave to account for Arrow’s higher precision.
- Remove `BinaryTypeCoercer` in `convert_binary_func`, because datafusion will do type analyzing and type cast after generating logical plan. `BinaryTypeCoercer` will limit the optimizations.

## Checklist

- [x] I have written necessary rustdoc comments.
- [ ]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  I have checked the [Release Timeline](https://github.com/risingwavelabs/rw-commits-history/blob/main/release_timeline.md) and [Currently Supported Versions](https://docs.risingwave.com/changelog/release-support-policy#support-end-dates-for-recent-releases) to determine which release branches I need to cherry-pick this PR into. 

## Documentation

- [ ]  My PR needs documentation updates. 

<details>
<summary><b>Release note</b></summary>

</details>

PR #24297, opened 2025-12-30
feat: support window function in datafusion

silver-ymz:

This PR adds support for window functions in DataFusion-based batch query execution for Iceberg tables. The implementation enables window functions like ROW_NUMBER(), RANK(), DENSE_RANK(), LAG(), LEAD(), and aggregate-based window functions to be converted and executed through DataFusion.

Extra updates:
- update `IcebergScanDetector` to `DatafusionExecuteChecker`. Previously, we detect `LogicalIcebergScan` on optimized logical plan. Now we detect iceberg `LogicalSource` with and `LogicalIcebergScan` on unoptimized plan.
- `Datafusion` will apply different optimize rules, because some rules make it difficult to convert the plan into datafusion.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/datafusion/window_function.rs
+++ b/src/frontend/src/datafusion/window_function.rs
@@ -1,4 +1,4 @@
-// Copyright 2025 RisingWave Labs
+// Copyright 2026 RisingWave Labs
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.


silver-ymz:
* **#24341** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24341?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24334** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24334?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24300** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24300?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24297** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24297?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24297?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#24284** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24284?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/frontend/src/optimizer/logical_optimization.rs
```diff
@@ -419,6 +419,21 @@ static CONVERT_OVER_WINDOW: LazyLock<OptimizationStage> = LazyLock::new(|| {
...
+        vec![
+            ProjectMergeRule::create(),
+            ProjectEliminateRule::create(),
+            TrivialProjectToValuesRule::create(),
+            UnionInputValuesMergeRule::create(),
+            OverWindowToAggAndJoinRule::create(),
+        ],
+        ApplyOrder::TopDown,
+    )
+});
```
chenzl25: It seems `OverWindowToTopNRule` is the rule which hard to be handled by datafusion? Can we use a new rule set called `CONVERT_OVER_WINDOW_FOR_BATCH` (removing the `OverWindowToTopNRule`), so that we can unify the `gen_optimized_logical_plan_for_datafusion_batch` and `gen_optimized_logical_plan_for_batch`? I don't really want to differentiate the batch optimized plan for RW and Datafusion.

silver-ymz: Updated. 

And let me explain why `OverWindowToTopNRule` is hard to be handled by datafusion. The main challenge is that we currently cannot translate [GroupTopN](https://github.com/risingwavelabs/risingwave/blob/1ec39bbc40ca2824a136f56e4b77318b7c9dac8f/src/frontend/src/optimizer/plan_visitor/datafusion_plan_converter.rs#L287) into an equivalent DataFusion plan without introducing additional performance overhead. Therefore, proper support likely depends on the implementation of https://github.com/apache/datafusion/issues/6899.

src/frontend/src/optimizer/plan_node/logical_over_window.rs
```diff
@@ -631,17 +631,6 @@ impl ToBatch for LogicalOverWindow {
...
-        // empty PARTITION BY by simply removing the following check.
-        let partition_key_indices = self.window_functions()[0]
-            .partition_by
-            .iter()
-            .map(|e| e.index())
-            .collect_vec();
-        if partition_key_indices.is_empty() {
-            empty_partition_by_not_implemented!();
-        }
-
```
chenzl25, line 644: Any reason to remove these codes?

src/frontend/src/optimizer/plan_visitor/datafusion_execute_checker.rs
```diff
@@ -0,0 +1,96 @@
...
+use crate::optimizer::plan_visitor::{LogicalPlanVisitor, Merge};
+use crate::optimizer::{LogicalPlanRef, PlanVisitor};
+
+/// Visitor to check if this plan can be executed by Datafusion.
+#[derive(Debug, Clone, Copy)]
+struct DatafusionExecuteChecker;
+
+#[derive(Debug, Clone, Copy, Default)]
+struct CheckResult {
+    have_update_node: bool,
```
chenzl25: In RisingWave, we call them `dml`, so `have_dml` is a better name.

src/frontend/src/optimizer/plan_visitor/datafusion_execute_checker.rs
```diff
@@ -0,0 +1,96 @@
...
+impl LogicalPlanVisitor for DatafusionExecuteChecker {
+    type DefaultBehavior = Merge<fn(CheckResult, CheckResult) -> CheckResult>;
+    type Result = CheckResult;
+
+    fn default_behavior() -> Self::DefaultBehavior {
+        Merge(|left, right| CheckResult {
+            have_update_node: left.have_update_node || right.have_update_node,
+            have_iceberg_scan: left.have_iceberg_scan || right.have_iceberg_scan,
+        })
+    }
```
chenzl25, line 38: It seems if we have a plan consists iceberg scan and table(hummock) scan, we still consider it as could datafusion runnable, but actually not. This could be improved later in another PR.

src/frontend/src/handler/query.rs
```diff
@@ -343,7 +342,7 @@ fn gen_batch_query_plan(
         }
     }
 
-    let batch_plan = optimized_logical.gen_batch_plan()?;
```
chenzl25, line 346: If right now, we use the same set of rule with risingwave batch engine, it seems we don't need to change this part?

src/frontend/src/optimizer/plan_node/logical_over_window.rs
silver-ymz: Window functions with an empty `PARTITION BY` clause previously triggered `OverWindowToTopNRule`. After removing `OverWindowToTopNRule`, such queries now fall into `empty_partition_by_not_implemented!()`, causing them to fail at runtime. To restore correct behavior, we should remove these codes.

For example, `SELECT RANK() OVER (ORDER BY y)`

src/frontend/src/optimizer/plan_visitor/datafusion_execute_checker.rs
silver-ymz: I think there’s a better approach. 

We can introduce a `supported` state to track all relevant nodes in `datafusion_plan_converter.rs`. With this mechanism, any query that passes `datafusion_execute_checker` is guaranteed to also pass `datafusion_plan_converter`, eliminating unsupported plan paths at runtime.

chenzl25 (approved):
Rest LGTM

src/frontend/src/handler/query.rs
silver-ymz: updated in new commit

src/frontend/src/optimizer/plan_visitor/datafusion_execute_checker.rs
silver-ymz: updated in new commit

Issue #23953, opened 2025-12-03
Isolation improvement for sync log store


kwannoel:
- Another requirement is to **improve** the current decoupled log store. Currently, synced log store executor is also driven by the downstream polling. This means if there’s downstream backpressure, we simply won’t drive the future.
    
    If downstream is backpressured, we’d like to continue driving the logstore writer, to ensure the upstream of the logstore is isolated from downstream slowness.
    
    We can modify the structure of synced log store executor, and colocate it with the dispatcher at the tail of each fragment.
    
    Dispatchers are `StreamConsumers` which yield a barrier stream: https://github.com/risingwavelabs/risingwave/blob/f771fd0b3f77343c59d12a018866a904faa376b1/src/stream/src/executor/dispatch.rs#L533
    
    They will continually be polled by the stream runtime, to drive them.
    
    They will try to `send` messages downstream. If they are unable to do so due to downstream backpressure, they will block.
    
    Rather than block, we can instead `select` both the `send` and `sync log store write future` and let them continue to write to the logstore.

- We need to construct this inside the actor builder.

PR #24303, opened 2025-12-30
feat(streaming): decouple synced log store progress from downstream backpressure

EdwinaZhu:

A dedicated dispatcher is introduced when a fragment terminates with a synced log store. The synced log store executor is decoupled and colocated with the dispatcher, allowing it to receive and process messages from upstream executors independently of downstream polling.

This PR is raised to solve #23953.

I added support for colocating a synchronous KV log store with the dispatch executor by introducing SyncLogStoreDispatchExecutor. The actor manager instantiates this executor when the tail of a fragment is a SyncKvLogStore. 

WriteFuture is responsible for receiving barriers and chunks from upstream. ConsumeFuture is responsible for switching state between dispatching messages and reading chunks from log store.

The implementation reuses the core read/write futures from sync_kv_log_store.rs.

Todo: 
1. Implemente a frontend optimization that inserts a StreamExchange plan node after each StreamSyncKvLogStore plan node.

2. Add an optimization to eliminate consecutive StreamSyncKvLogStore plan nodes by merging them into a single StreamSyncKvLogStore node.
## Checklist

- [x]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 

## Documentation

<details>

</details>


kwannoel:
I triggered a copilot review first. Can you post the benchmark results here:
1. Benchmark methodology
2. Environment (cores, release mode enabled?, etc..)
3. results

EdwinaZhu:
> I triggered a copilot review first. Can you post the benchmark results here:
> 
> 1. Benchmark methodology
> 2. Environment (cores, release mode enabled?, etc..)
> 3. results

@kwannoel @wenym1 
1. Benchmark methodology:
This benchmark evaluates how well RisingWave completes barrier propagation under downstream backpressure when unaligned join is enabled, focusing on the sink’s ability (with log-store decoupling) to keep consuming upstream even when downstream dispatch is slow.

- **Setup (run once, not timed):**
Enable unaligned join: `set streaming_enable_unaligned_join = true`.
Create a small dimension table `dim`, a large fact table `fact`, and a target table `t`.
Populate `dim` with two keys and populate `fact` with **4,000,000 rows** split across the two keys.
Create a table sink `s_join` that continuously writes the result of `fact JOIN dim` into `t`.
Create a downstream materialized view `mv_t_agg` on top of `t` to introduce **amplification and heavy state updates**, intentionally generating strong backpressure.
Run `flush` to ensure the pipeline is fully caught up and reaches a steady state before measurement.

- **Benchmark action**:
Execute `DELETE FROM` dim; followed by `flush;`.
Deleting `dim` invalidates the join, triggering a large number of retractions/updates that must propagate through s_join into t and then through mv_t_agg.
The timed flush captures the end-to-end completion time until the corresponding barrier is fully processed by the entire streaming pipeline.

2. Environment
I ran the benchmark with release mode, on M1 pro with 10 cores available and 32G mem. 

3. Results
unoptimized version:
unop
optimized version:
op
When running benchmark action, without log store dispatcher, the peak of barrier latency would achieve nearly 30 secs. With log store dispatcher, the peak of barrier latency would achieve only 9 secs. This improves the stability of the system a lot.

kwannoel (commented):
Initial comment: See that `dispatch_sync_log_store.rs` is actually the main part we are reviewing. The rest are just refactoring to change function / struct visibilities.

Can add a review note next time to highlight important files to review.

Taking a look at this file now.

EdwinaZhu:
Compared with sync log store executor, the log store dispatcher did the following modifications.
The dispatcher split the consumer side into two stages:
`ConsumerFuture::ReadingChunk`: include `ReadFuture` reads chunks from the log store/ in-memory buffer, which is nearly same as that in the log store executor.
`ConsumerFuture::Dispatching`: turns a read-out chunk/watermark or a queued barrier into an `inner.dispatch(...)` future and drives that future to completion. Here dispatching becomes an explicit async state `ConsumerFuture::Dispatching { future, ... }`. 

About when handling the barrier passed from upstream, in the dispatcher, the write side only enqueues the barrier into `barriers`; the executor only yield the barrier after it has been successfully dispatched to downstream outputs. Inside `ConsumerFuture::next_event`, if that queue is non-empty and the consumer is in `ConsumerFuture::ReadingChunk`, it will stop reading, save the reading future and switch to dispatching the barrier, achieving barrier prioritization.

Besides, the log store dispatch executor treat the end of stream as a normal event. When it encounters the end of upstream, `next_event` of `WriteFuture` would return `WriteFutureEvent:EndOfStream`. And the outer loop uses `write_done` plus "barrier queue empty + buffer empty" to decide when to exit, making sure it drain any remaining buffered data before terminating.

About waiting strategy, in the dispatcher, `ReadFuture::next_chunk` returns `Option<StreamChunk>` and returns `Ok(None)` when there is nothing to read. The outer loop then decides whether to `pending()`: when `write_done` set to true, the write side would turn into pending. When the consumer side enters Idle state, it would turn into pending. When `write_done` is true and the consumer side enters idle state, the outer loop of `StreamConsumer::execute()` would end.

In `actor_manager.rs`,  `spawn_actor` now branches on `node_body`: if the tail of the fragment is a sync log store node, actors go through a dedicated creation path, while all other actors keep using the generic path.

src/stream/src/executor/dispatch_sync_log_store.rs
```diff
@@ -0,0 +1,1216 @@
...
+        tracing::trace!(
+            end_seq_id,
+            "read flushed chunk of size: {}",
+            chunk.cardinality()
+        );
+        *self = ReadFuture::Idle;
+        Ok(Some(chunk))
+    }
+}
+enum ConsumerFuture<S: StateStoreRead> {
```
EdwinaZhu: The dispatcher split the consumer side into two stages:
ConsumerFuture::ReadingChunk: include ReadFuture reads chunks from the log store/ in-memory buffer, which is nearly same as that in the log store executor.
ConsumerFuture::Dispatching: turns a read-out chunk/watermark or a queued barrier into an inner.dispatch(...) future and drives that future to completion. Here dispatching becomes an explicit async state ConsumerFuture::Dispatching { future, ... }.

src/stream/src/executor/dispatch_sync_log_store.rs
```diff
@@ -0,0 +1,1216 @@
...
+            write_state,
+        }
+    }
+
+    async fn next_event(
+        &mut self,
+        metrics: &SyncedKvLogStoreMetrics,
+    ) -> StreamResult<(BoxedMessageStream, LogStoreWriteState<S>, WriteFutureEvent)> {
+        match self {
+            WriteFuture::ReceiveFromUpstream { future, .. } => {
```
EdwinaZhu: Here it would return `WriteFutureEvent::EndOfStream` when the upstream ends.

src/stream/src/executor/dispatch_sync_log_store.rs
```diff
@@ -0,0 +1,1216 @@
...
+                                consumer_future_state =
+                                    ConsumerFuture::read_chunk(inner, read_future);
+                            }
+                            ConsumerFutureEvent::WaitingForChunks => {
+                                if write_done {
+                                    break;
+                                };
+                                consumer_future_state =
+                                    ConsumerFuture::read_chunk(inner, read_future);
+                            }
```
EdwinaZhu: Consumer side would enter `Dispatching` state once it reads some chunks out from log store.

src/stream/src/executor/dispatch_sync_log_store.rs
```diff
@@ -0,0 +1,1216 @@
...
+                                        )
+                                        .await?;
+                                        seq_id = FIRST_SEQ_ID;
+                                        let update_vnode_bitmap =
+                                            barrier.as_update_vnode_bitmap(actor_id);
+                                        if update_vnode_bitmap.is_some() {
+                                            return Err(anyhow!("vnode bitmap update during dispatch is not supported.").into());
+                                        }
+
+                                        barriers.push_front(Message::Barrier(barrier));
```
EdwinaZhu: Here write side would add barriers into queue when receiving barriers from upstream. Other part is nearly same as that in sync log store executor.

src/stream/src/executor/dispatch.rs
```diff
@@ -50,6 +49,9 @@ use crate::task::{DispatcherId, NewOutputRequest};
 
 mod output_mapping;
 pub use output_mapping::DispatchOutputMapping;
+#[path = "dispatch_sync_log_store.rs"]
```
wenym1: Seems that `path` is not needed?

src/stream/src/task/actor_manager.rs
```diff
@@ -469,6 +475,111 @@ impl StreamActorManager {
...
+
+        let serde = LogStoreRowSerde::new(
+            &table,
+            vnode_bitmap.clone().map(|b: Bitmap| b.into()),
+            &KV_LOG_STORE_V2_INFO,
+        );
+
+        let [input] = node.input.as_slice() else {
+            bail!("SyncLogStoreNode should have exactly one input");
+        };
```
wenym1: These parts can be moved to `SyncLogStoreDispatchExecutor::new`

src/stream/src/executor/dispatch_sync_log_store.rs
```diff
@@ -0,0 +1,1216 @@
...
+        future: ReadFlushedChunkFuture,
+        end_seq_id: SeqId,
+    },
+    Idle,
+}
+
+impl<S: StateStoreRead> ReadFuture<S> {
+    pub async fn next_chunk(
+        &mut self,
+        progress: &mut LogStoreVnodeProgress,
```
wenym1: Is it just a copy of code of the original sync log store? Is there a chance to reuse the original code?

src/stream/src/executor/dispatch_sync_log_store.rs
```diff
@@ -0,0 +1,1216 @@
...
+
+        Ok(Self {
+            input,
+            inner,
+            log_store_config,
+        })
+    }
+}
+
+enum WriteFuture<S: LocalStateStore> {
```
wenym1: Same for WriteFuture. What's the difference between them and the code in the original sync log store?

wenym1 (commented):
Leave some early comments. Will continue review tomorrow.

src/stream/src/executor/dispatch_sync_log_store.rs
```diff
@@ -0,0 +1,1216 @@
...
+            }
+        }
+    }
+}
+
+impl<S: StateStore> StreamConsumer for SyncLogStoreDispatchExecutor<S> {
+    type BarrierStream = impl Stream<Item = StreamResult<Barrier>> + Send;
+
+    fn execute(mut self: Box<Self>) -> Self::BarrierStream {
+        let _max_barrier_count_per_batch = self.inner.actor_config.developer.max_barrier_batch_size;
```
wenym1: May remove the deadcode?

src/stream/src/executor/dispatch_sync_log_store.rs
```diff
@@ -0,0 +1,1216 @@
...
+                log_store_config.chunk_size,
+            );
+            initial_write_state.init(first_write_epoch).await?;
+
+            let initial_write_epoch = first_write_epoch;
+            let mut pause_stream = first_barrier.is_pause_on_startup();
+
+            if log_store_config.aligned {
+                tracing::info!("aligned mode");
+                let log_store_stream = read_state
```
wenym1: We also have a duplicated code in the sync log store. Can we possibly have a method to extract the same logic to generate a stream when `log_store_config.aligned`, and then here we only keep consume the stream and yield the items yielded from that stream.

src/stream/src/executor/dispatch/dispatch_sync_log_store.rs
```diff
@@ -0,0 +1,1216 @@
...
+                                .instrument(tracing::info_span!("dispatch_watermark"))
+                                .instrument_await("dispatch_watermark")
+                                .await?;
+                        }
+                    }
+                }
+                return Ok(());
+            }
+
+            let mut seq_id = FIRST_SEQ_ID;
```
wenym1, line 342: review note: core logic starts here

src/stream/src/executor/dispatch_sync_log_store.rs
EdwinaZhu: The difference locates in ReadFuture::next_chunk() and WriteFuture::next_event(). In original sync log store, when the executor cannot fetch more data from upstream, it would return error directly in next_chunk() and next_event(). Now in next_chunk() when the executor cannot fetch more data from upstream, in ReadFuture::next_chunk() it would return None to main loop, indicating the clean_state of read side of log store.

src/stream/src/executor/dispatch_sync_log_store.rs
EdwinaZhu: In original sync log store, WriteFuture::next_event() would return error when upstream ended. About handling the case where upstream ended, WriteFuture::next_event() would return WriteFutureEvent::EndOfStream, indicating write side is clean. Now in StreamConsumer::execute(), we use the returned WriteFutureEvent to decide whether to keep WriteFuture or ConsumerFuture pending.

wenym1 (commented):
Further comment. Will continue review the core logic

src/stream/src/executor/dispatch.rs
```diff
@@ -514,6 +516,45 @@ impl DispatchExecutor {
     }
 }
 
+async fn dispatch_message_batch(
```
EdwinaZhu, line 521: Helper extracted from `DispatchExecutor`'s execute loop so that `SyncLogStoreDispatchExecutor` can reuse the same dispatch logic.

Dispatches a `MessageBatch` to all downstream dispatchers. Returns `Some(barriers)` if the batch contained barriers (so the caller can yield them to the barrier stream), or `None` for chunk/watermark batches.

src/stream/src/executor/dispatch/dispatch_sync_log_store.rs
```diff
@@ -0,0 +1,699 @@
...
+        })
+    }
+}
+
+type DispatchingFuture = BoxFuture<'static, (DispatchExecutorInner, StreamResult<()>)>;
+enum DispatchType {
+    ChunkOrWatermark,
+    Barrier(Barrier),
+}
+enum ConsumerFuture<S: StateStoreRead> {
```
EdwinaZhu: State machine for the **consumer side** (log store read + downstream dispatch).

The consumer alternates between two phases:

- **`ReadingChunk`**: Polling the log store (persisted stream or in-memory buffer) for the next chunk to dispatch to downstream. While in this state, the `WriteFuture` (write side) continues running in parallel, so upstream progress is **not** blocked by downstream.
- **`Dispatching`**: Actively sending a chunk/barrier to downstream dispatchers. The log store read future is kept alive but not polled. Downstream backpressure only blocks the executor here, not the write side.
- **`Empty`**: A placeholder used during state transitions via `std::mem::replace`. This variant must never be observed in `next_event`.

src/stream/src/executor/dispatch/dispatch_sync_log_store.rs
```diff
@@ -0,0 +1,699 @@
...
+
+    async fn next_event(
+        &mut self,
+        barriers: &mut VecDeque<Message>,
+        progress: &mut LogStoreVnodeProgress,
+        read_state: &LogStoreReadState<S>,
+        buffer: &mut SyncedLogStoreBuffer,
+        metrics: &SyncedKvLogStoreMetrics,
+    ) -> StreamResult<(DispatchExecutorInner, ConsumerFutureEvent, ReadFuture<S>)> {
+        if !barriers.is_empty()
```
EdwinaZhu: **Barrier priority over chunk reads.**

When the write side has enqueued one or more barriers (via `barrers.push_front`), the consumer switches from `ReadingChunk` to dispatching the barrier immediately, before reading any more chunks from the log store. This ensures barriers are forwarded to downstream in-order, without waiting for all buffered chunks in the current epoch to be drained first.

src/stream/src/executor/dispatch/dispatch_sync_log_store.rs
```diff
@@ -0,0 +1,699 @@
...
+                unreachable!("ConsumerFuture::Empty should be handled!")
+            }
+        }
+    }
+}
+
+impl<S: StateStore> StreamConsumer for SyncLogStoreDispatchExecutor<S> {
+    type BarrierStream = impl Stream<Item = StreamResult<Barrier>> + Send;
+
+    fn execute(mut self: Box<Self>) -> Self::BarrierStream {
```
EdwinaZhu, line 282: **Core execute loop — two concurrent futures running in a `select` loop:**

**Write side (`WriteFuture`):** receives messages from upstream and writes them into the log store. On receiving a barrier, it would be enqueued into `barrier` (a `VecDeque`) and continues. It can also be briefly paused to let the consumer side catch up (see `WriteFuture::Paused`).

**Consumer side (`ConsumerFuture`):** reads chunks out of the log store (persisted hitory first, then in-memory buffer) and dispatches them to downstream via `DispatchExecutorInner`. It dispatches pending barriers with higher priority than chunk reads.

**Key decoupling property:** the write side advances the log store and processes ustream messages independently of downstream polling speed. Downstream backpressure only blocks `ConsumerFuture` while it is in the `Dispatching` state — the `WriteFuture` continues running in parallel.

src/stream/src/executor/dispatch/dispatch_sync_log_store.rs
```diff
@@ -0,0 +1,699 @@
...
+                                            barrier.as_update_vnode_bitmap(actor_id);
+                                        if update_vnode_bitmap.is_some() {
+                                            return Err(anyhow!("vnode bitmap update during dispatch is not supported.").into());
+                                        }
+
+                                        write_state_post_write_barrier
+                                            .post_yield_barrier(None)
+                                            .await?;
+
+                                        let is_stop_barrier = barrier.is_stop(actor_id);
```
EdwinaZhu, line 454: After receiving stop barrier, it would set the flag `end_of_stream` as `true`, and the writeside would stop polling upstream. the writefuture would keep pending.

src/stream/src/executor/sync_kv_log_store.rs
```diff
@@ -540,360 +545,398 @@ impl<S: LocalStateStore> WriteFuture<S> {
...
-    async fn execute_inner(self) {
-        let mut input = self.upstream.execute();
-
-        // init first epoch + local state store
-        let first_barrier = expect_first_barrier(&mut input).await?;
-        let first_write_epoch = first_barrier.epoch;
-        yield Message::Barrier(first_barrier.clone());
-
-        let local_state_store = self
+    pub(crate) async fn init_local_log_store_state(
```
EdwinaZhu, line 554: Add helpers which could be reused in both `SyncedKvLogStoreExecutor` and `SyncedKvLogStoreDispatchExecutor`.

src/stream/src/executor/sync_kv_log_store.rs
```diff
@@ -73,8 +73,10 @@ use risingwave_common::array::StreamChunk;
 use risingwave_common::bitmap::Bitmap;
```
wenym1, line 73: Clean and elegant refactor in this file👍

src/stream/src/task/actor_manager.rs
```diff
@@ -469,6 +471,67 @@ impl StreamActorManager {
         Ok(actor)
     }
 
+    #[allow(clippy::too_many_arguments)]
```
wenym1: use `#[expect(...)]`. Same for all similar usages in this PR.

src/stream/src/executor/sync_kv_log_store.rs
```diff
@@ -540,360 +545,398 @@ impl<S: LocalStateStore> WriteFuture<S> {
...
+            match message? {
+                Message::Barrier(barrier) => {
+                    let is_checkpoint = barrier.is_checkpoint();
+                    let mut progress = LogStoreVnodeProgress::None;
+                    progress.apply_aligned(read_state.vnodes().clone(), barrier.epoch.prev, None);
+                    // Truncate the logstore.
+                    let post_seal =
+                        initial_write_state.seal_current_epoch(barrier.epoch.curr, progress.take());
+                    let update_vnode_bitmap = barrier.as_update_vnode_bitmap(actor_id);
+                    if update_vnode_bitmap.is_some() {
```
wenym1: can introduce a common method in `barrier` like `assume_no_update_vnode_bitmap(actor_id)?`.

src/stream/src/executor/dispatch/dispatch_sync_log_store.rs
```diff
@@ -0,0 +1,699 @@
...
+enum DispatchType {
+    ChunkOrWatermark,
+    Barrier(Barrier),
+}
+enum ConsumerFuture<S: StateStoreRead> {
+    ReadingChunk {
+        read_future: ReadFuture<S>,
+        inner: DispatchExecutorInner,
+    },
+    Dispatching {
```
wenym1, line 165: This `ConsumerFuture` can be simplified.
1. `Empty` is actually an unreachable temporary place holder, we may rename it to explicit `PlaceHolder`
2. Both of the two really meaning variant have the `ReadFuture`, and therefore we won't have to include it in this `ConsumerFuture`. Instead of, we can store it outside in the executor stream body, and only pass its mutable reference as param to the `next_event`
3. We can reason that barrier queuing only exists on the `Dispatching` variant, because in current implementation, when the queue is not empty, we always overwrite the `ReadingChunk` variant with the `Dispatching`. Therefore, we only need to have the barrier queue inside the `Dispatching` variant.
```
enum ConsumerFuture {
    ReadingChunk {
        inner: DispatchExecutorInner,
    },
    Dispatching {
        future: DispatchingFuture, // Output is Result<(DispatchExecutorInner, Option<Barrier>)>
    }
}
```
4. For the three variant of `ConsumerFutureEvent`, only the event handler of `BarrierDispatched` have work to do outside the consumer future (which is yield barrier), and for the other two variants, their event handler simpler re-enter the consumer future event. Therefore, we can actually only have a single `BarrierDispatched` event, and then we can further rename this `next_event` to `next_dispatched_barrier` that returns the dispatched barrier. The method will be in a loop, and keep handling different internal events, until a barrier is dispatched and then return this barrier.

src/stream/src/executor/dispatch/dispatch_sync_log_store.rs
```diff
@@ -0,0 +1,699 @@
...
+
+        Ok(Self {
+            input,
+            inner,
+            log_store_context,
+        })
+    }
+}
+
+type DispatchingFuture = BoxFuture<'static, (DispatchExecutorInner, StreamResult<()>)>;
```
wenym1: Can use define_opaque to define a static opaque future type for `DispatchingFuture` to avoid creating a box future for every dispatch. May refer to other usage of `define_opaque` in the repo.

wenym1 (commented):
Rest LGTM. Thanks for the great work!

src/stream/src/executor/sync_kv_log_store.rs
```diff
@@ -540,360 +545,398 @@ impl<S: LocalStateStore> WriteFuture<S> {
...
-                                    .inc_by(flush_info.flush_count as _);
-                                self.metrics
-                                    .storage_write_size
-                                    .inc_by(flush_info.flush_size as _);
+                                seq_id = new_seq_id;
+                                write_future_state = next_write_future;
+                            }
+                            // FIXME(kwannoel): This should truncate the logstore,
+                            // it will not bypass like barrier.
+                            Message::Watermark(_watermark) => {
```
wenym1, line 878: should leave as a TODO to handle upstream watermark

src/stream/src/executor/dispatch/dispatch_sync_log_store.rs
```diff
@@ -0,0 +1,699 @@
...
+                                        } else {
+                                            write_future_state = WriteFuture::receive_from_upstream(
+                                                stream,
+                                                write_state,
+                                            );
+                                        }
+                                        barriers.push_front(Message::Barrier(barrier));
+                                    }
+                                }
+                                Message::Watermark(_watermark) => {
```
wenym1, line 470: should leave as a TODO to handle upstream watermark

src/stream/src/executor/dispatch/dispatch_sync_log_store.rs
```diff
@@ -0,0 +1,699 @@
...
+        async move {
+            let actor_id = self.inner.actor_id;
+            let log_store_config = self.log_store_context;
+
+            let mut barriers = VecDeque::<Message>::new();
+            let mut input = self.input.execute();
+
+            let first_barrier = expect_first_barrier(&mut input).await?;
+            let first_write_epoch = first_barrier.epoch;
+            yield first_barrier.clone();
```
wenym1: better to first dispatch the barrier and then yield the barrier.

src/stream/src/executor/dispatch/dispatch_sync_log_store.rs
EdwinaZhu: Addressed. I changed DispatchingFuture from BoxFuture to an opaque future type with #[define_opaque]. The enum still stores Pin<Box<DispatchingFuture>>.

src/stream/src/executor/dispatch/dispatch_sync_log_store.rs
```diff
@@ -0,0 +1,709 @@
...
+                        .expect("barrier batch should contain one barrier")
+                })
+            });
+        (inner, r)
+    }
+}
+
+/// State machine for the consumer side, which reads chunks from the log store and dispatches
+/// chunks or barriers downstream.
+enum ConsumerFuture {
```
EdwinaZhu, line 159: ConsumerFuture now only has the two meaningful states: ReadingChunk and Dispatching. ReadFuture is stored outside the enum and passed as a mutable reference into next_dispatched_barrier. The barrier queue is now only kept in Dispatching, matching the current invariant that queued barriers always drive the state into dispatching.

Aside from _BarrierDispatched_, I added _CleanStateReached_ as a ConsumerFutureEvent. When the ConsumerFutureSide finds out that the read side is in clean state, it would return _CleanStateReached_, so the outer stream body can stop pausing the write side when the consumer reaches a clean state.

src/stream/src/executor/mod.rs
```diff
@@ -1286,6 +1293,14 @@ impl<M> MessageInner<M> {
             MessageInner::Watermark(watermark) => MessageInner::Watermark(watermark),
         }
     }
+
+    pub fn into_batch(self) -> MessageBatchInner<M> {
```
wenym1: any reason to extract this `into_batch` method, instead of using `msg.into()`?

src/stream/src/executor/dispatch/dispatch_sync_log_store.rs
```diff
@@ -0,0 +1,709 @@
...
+            "sync_log_store_dispatch",
+            "sync_log_store_dispatch",
+        );
+
+        let table = sync
+            .log_store_table
+            .as_ref()
+            .ok_or_else(|| anyhow!("missing log_store_table in SyncLogStoreNode"))?;
+
+        #[allow(deprecated)]
```
wenym1: what's the `#[allow(deprecated)]` for?

src/stream/src/executor/dispatch/dispatch_sync_log_store.rs
```diff
@@ -0,0 +1,709 @@
...
+            *clean_state = true;
+            metrics.clean_state.inc();
+            true
+        } else {
+            false
+        }
+    }
+
+    #[expect(clippy::too_many_arguments)]
+    async fn next_dispatched_barrier<S: StateStoreRead>(
```
wenym1: since we still have two variants in the `ConsumerFutureEvent`, let's keep the name `next_event`

src/stream/src/executor/dispatch/dispatch_sync_log_store.rs
```diff
@@ -0,0 +1,709 @@
...
+                ConsumerFuture::Dispatching { future, .. } => {
+                    let (inner, result) = future.await;
+                    let barrier = result?;
+                    let mut barrier_queue = must_match!(
+                        replace(self, ConsumerFuture::PlaceHolder),
+                        ConsumerFuture::Dispatching { barrier_queue, .. } => barrier_queue
+                    );
+
+                    if let Some(next_barrier) = barrier_queue.pop_back() {
+                        *self = Self::dispatch(inner, next_barrier, barrier_queue);
```
wenym1: We can actually in place set the future as `dispatching_future(inner, message)` directly, instead of re-composing the whole consumer future.

After we avoid calling `Self::dispatch` here, all remaining call to `Self::dispatch` passes the empty vec queue as param, and then we can actually not pass `barrier_queue` as param, indicating that when we enter `Dispatching` state from other state, the `barrier_queue` is always empty.

src/stream/src/executor/dispatch/dispatch_sync_log_store.rs
```diff
@@ -0,0 +1,709 @@
...
+        clean_state: &mut bool,
+        progress: &mut LogStoreVnodeProgress,
+        read_state: &LogStoreReadState<S>,
+        buffer: &mut SyncedLogStoreBuffer,
+        metrics: &SyncedKvLogStoreMetrics,
+    ) -> StreamResult<ConsumerFutureEvent> {
+        loop {
+            match self {
+                ConsumerFuture::ReadingChunk { .. } => {
+                    if Self::mark_clean_state(clean_state, read_future, buffer, metrics) {
```
wenym1: I think `mark_clean_state` below after `next_chunk` is sufficient. Any specific reason to do it again here at the beginning?

src/stream/src/executor/dispatch/dispatch_sync_log_store.rs
```diff
@@ -0,0 +1,709 @@
...
+/// State machine for the consumer side, which reads chunks from the log store and dispatches
+/// chunks or barriers downstream.
+enum ConsumerFuture {
+    /// Polls the log store for the next chunk. The read future is kept outside this state machine
+    /// so both meaningful states can share it.
+    ReadingChunk { inner: DispatchExecutorInner },
+    /// Dispatches the current message downstream. Any barrier received while dispatching is queued
+    /// here and dispatched before reading another chunk.
+    Dispatching {
+        future: Pin<Box<DispatchingFuture>>,
```
wenym1: can try to use `#[pin_project]` to avoid creating a box for every `DispatchingFuture`.

src/stream/src/executor/dispatch/dispatch_sync_log_store.rs
```diff
@@ -0,0 +1,709 @@
...
+            ConsumerFuture::Dispatching { barrier_queue, .. } => {
+                barrier_queue.push_front(message);
+            }
+            ConsumerFuture::PlaceHolder => {
+                unreachable!("ConsumerFuture::PlaceHolder should be handled!")
+            }
+        }
+    }
+
+    fn mark_clean_state<S: StateStoreRead>(
```
wenym1: Does the original sync log store executor share the same `mark_clean_state` logic? If so, we can extract the logic in the original executor, and then call the method here.

wenym1 (commented):
Rest LGTM

src/stream/src/executor/dispatch/dispatch_sync_log_store.rs
EdwinaZhu: The mark_clean_state logistics is extracted into `impl ReadFuture`.

src/stream/src/executor/mod.rs
EdwinaZhu: I rewrited `impl From<MessageInner<M>> for <MessageBatchInner<M>>`. Now it could use `msg.into()`.

src/stream/src/executor/dispatch/dispatch_sync_log_store.rs
EdwinaZhu: It has been deleted. Now these params would be loaded directly from `actor_context.config`.

src/stream/src/executor/dispatch/dispatch_sync_log_store.rs
EdwinaZhu: It has been deleted.

wenym1:
Just pushed a small patch to avoid matching on enum variant for multiple times in some methods.

wenym1 (approved):
LGTM! Thanks for the great work!

We may have a later PR in frontend or meta node to always let the sync log store node to be the latest node in a fragment, so that this optimization can be applied. Besides, we may also add a new compute node streaming config to allow disabling this feature, in case the current implementation has something wrong.

PR #24310, opened 2025-12-31
refactor(sink): replace `ForceAppendOnly` with a separate `ignore_delete` field

BugenZhao:

Previously we have `ForceAppendOnly` as a `SinkType`, which allows sinking non-append-only stream to append-only sink by dropping `Delete` records.

However, whether to drop `Delete` records can actually be orthogonal to whether the sink is append-only. For example, some users want an upsert sink to process `Insert` and `Update` operations as usual while ignoring `Delete` operations to prevent accidental data deletion in downstream systems.

This PR refactors `SinkType::ForceAppendOnly` to be `SinkType::AppendOnly` with a newly introduced `ignore_delete` field set to `true` while maintaining backward compatibility, without actual logic changes. In future PRs, we may allow a combination of `SinkType::Upsert` with `ignore_delete` to unlock the mentioned use case.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/proto/catalog.proto
+++ b/proto/catalog.proto
@@ -163,6 +163,7 @@ message Source {
 enum SinkType {
   SINK_TYPE_UNSPECIFIED = 0;
   SINK_TYPE_APPEND_ONLY = 1;
+  // Deprecated. Use `append_only` + `ignore_delete` instead.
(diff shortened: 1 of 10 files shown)


BugenZhao:

- #24310 <-- you are here 🤪
    - #24311

PR #24323, opened 2026-01-03
feat: add support to alter backfill parallelism for streaming jobs

shanicky:
Introduce a new mechanism to control backfill parallelism for streaming jobs, allowing separate tuning of resource usage during backfill phases.

- Add AlterBackfillParallelism RPC and proto definitions in ddl_service.proto
- Extend CatalogWriter trait and frontend handler to support ALTER ... SET BACKFILL_PARALLELISM
- Implement server-side handling in Meta service and DdlController for backfill parallelism changes
- Add rescheduling logic in GlobalStreamManager and ScaleController to update backfill parallelism safely
- Validate unsupported cases (e.g., Iceberg tables) and support deferred execution mode
- Add integration test verifying backfill parallelism switches after backfill completion
- Extend SQL parser and keywords to recognize BACKFILL_PARALLELISM syntax

**Summary**
This PR introduces support for altering the backfill parallelism of streaming jobs through the DDL gRPC API and the corresponding frontend handlers. It enables users to adjust the parallelism used during backfilling operations, including support for deferred execution and resetting the parallelism setting.

**Details**
- Added new RPC messages (`AlterBackfillParallelismRequest` and `AlterBackfillParallelismResponse`) and the `AlterBackfillParallelism` method to the DDL gRPC service definition.
- Extended the `CatalogWriter` trait and its implementation with an asynchronous method `alter_backfill_parallelism` to communicate with the meta service.
- Implemented a new handler function `handle_alter_backfill_parallelism` to process alter backfill parallelism requests, including extracting parallelism values and validating input.
- Integrated the new handler into the main dispatch logic, supporting the operation on tables, indexes, sinks, and sources, while returning a not implemented error for materialized views and disallowing the operation on iceberg engine tables.
- Added stubs for the new method in testing utilities (`MockCatalogWriter`) to facilitate unit testing.
- Began partial implementation of the meta service handler for backfill parallelism alterations.

## Checklist

## Documentation

<details>
Introduce a new mechanism to control backfill parallelism for streaming jobs,

</details>

### DIFF
--- a/src/meta/src/stream/scale.rs
+++ b/src/meta/src/stream/scale.rs
@@ -180,7 +180,6 @@ impl ScaleController {
     pub async fn reschedule_backfill_parallelism_inplace(
         &self,
         policy: HashMap<JobId, Option<StreamingParallelism>>,
-        workers: HashMap<WorkerId, PbWorkerNode>,
     ) -> MetaResult<HashMap<DatabaseId, Command>> {
         if policy.is_empty() {
             return Ok(HashMap::new());
@@ -213,7 +212,7 @@ impl ScaleController {
 
         let jobs = policy.keys().copied().collect();
 
-        let command = self.rerender_inner(&txn, jobs, &workers).await?;
+        let command = build_reschedule_intent_for_jobs(&txn, jobs).await?;
 
         txn.commit().await?;
 
--- a/src/meta/src/stream/stream_manager.rs
+++ b/src/meta/src/stream/stream_manager.rs
@@ -797,21 +797,9 @@ impl GlobalStreamManager {
             }
         }
 
-        let worker_nodes = self
-            .metadata_manager
-            .list_active_streaming_compute_nodes()
-            .await?
-            .into_iter()
-            .filter(|w| w.is_streaming_schedulable())
-            .collect_vec();
-        let workers = worker_nodes.into_iter().map(|x| (x.id, x)).collect();
-
         let commands = self
             .scale_controller
-            .reschedule_backfill_parallelism_inplace(
-                HashMap::from([(job_id, parallelism)]),
-                workers,
-            )
+            .reschedule_backfill_parallelism_inplace(HashMap::from([(job_id, parallelism)]))
             .await?;
 
         if !deferred {


src/meta/src/stream/scale.rs
```diff
@@ -325,6 +325,62 @@ impl ScaleController {
...
+    pub async fn reschedule_backfill_parallelism_inplace(
+        &self,
+        policy: HashMap<JobId, Option<StreamingParallelism>>,
+        workers: HashMap<WorkerId, PbWorkerNode>,
+    ) -> MetaResult<HashMap<DatabaseId, Command>> {
+        if policy.is_empty() {
+            return Ok(HashMap::new());
+        }
+
+        let inner = self.metadata_manager.catalog_controller.inner.read().await;
```
yezizp2012: Should be `write` lock. Please fix it together with `reschedule_inplace`, `reschedule_fragment_inplace`.

src/meta/src/stream/scale.rs
```diff
@@ -325,6 +325,62 @@ impl ScaleController {
...
+        let txn = inner.db.begin().await?;
+
+        for (table_id, parallelism) in &policy {
```
yezizp2012, line 191: Can we query and update them together for better performance?

src/meta/src/stream/stream_manager.rs
```diff
@@ -705,6 +705,63 @@ impl GlobalStreamManager {
...
+        &self,
+        job_id: JobId,
+        parallelism: Option<StreamingParallelism>,
+        deferred: bool,
+    ) -> MetaResult<()> {
+        let _reschedule_job_lock = self.reschedule_lock_write_guard().await;
+
+        let background_jobs = self
+            .metadata_manager
+            .list_background_creating_jobs()
```
yezizp2012, line 783: Can we add a single function to query directly using job_id?

yezizp2012:
Btw, do we have a plan to expose backfill parallelism of the streaming job in the system catalog?

shanicky:
> Btw, do we have a plan to expose backfill parallelism of the streaming job in the system catalog?

I’m still hesitating because this thing becomes completely useless right after the backfill is finished.

src/meta/src/stream/stream_manager.rs
shanicky: I previously tried it in reschedule_inplace, but the code didn’t look very clean. In the actual scenario, there is only one request ID here.

shanicky: There might be a situation here where this job itself is fine, but the no-shuffle depends on a job that has a no-shuffle backfill. In that case, this job cannot be modified because the upstream job is not allowed to be changed.

src/meta/src/stream/scale.rs
shanicky: I previously tried it in reschedule_inplace, but the code didn’t look very clean. In the actual scenario, there is only one request ID here.

src/meta/src/stream/scale.rs
shanicky: fixed

kwannoel (commented):
Can the usecase be supported by simply altering the streaming parallelism, then we don't need to introduce `backfill_parallelism`

shanicky:
> Can the usecase be supported by simply altering the streaming parallelism, then we don't need to introduce `backfill_parallelism`

Doing it this way adds too much unpredictability. It’s best to match the backfill parallelism of the 'alter' with the 'create' for consistency. We asked about this in a weekly meeting, but it seems everyone is still just handling it individually.

shanicky:
> I guess the main difference is without backfill parallleism, user has to do something like: Alter streaming parallelism during backfillAlter streaming parallelism back to normal after backfil
> 
> vs Alter backfill parallelism during backfillNo need for another step[7:59 AM]Is that the main gain?
> 
> And for backfill parallelism, does it have feature parity with streaming parallelism?
> 
> Meaning it supports:
> 
> * Adaptive (with bounded, full and other configurations)
> * Fixed
> 
> Further, where's the place where we transition from backfill paralleism -> streaming parallelism? I didn't see it in the code

Before we supported backfill parallelism, performing an alter parallelism operation on a job that was undergoing backfilling would only change the parallelism for the regular stage.
Later, we added backfill parallelism, allowing the parallelism for the backfill stage to be specified.
The current behavior, according to previous discussions and expectations, is that during the backfilling stage:

alter parallelism changes the parallelism of the regular stage,
alter backfill parallelism changes the parallelism of the backfill stage.

src/frontend/src/handler/alter_parallelism.rs
```diff
@@ -126,6 +162,13 @@ fn extract_table_parallelism(parallelism: SetVariableValue) -> Result<TableParal
     Ok(target_parallelism)
 }
 
+fn extract_backfill_parallelism(parallelism: SetVariableValue) -> Result<Option<TableParallelism>> {
+    match parallelism {
+        SetVariableValue::Default => Ok(None),
+        other => extract_table_parallelism(other).map(Some),
```
kwannoel, line 168: This seems weird, why is backfill parallelism a "TableParallelism"? It's out of scope of this PR, but perhaps we can refine the struct we used here.

src/meta/src/stream/scale.rs
```diff
@@ -264,7 +266,7 @@ impl ScaleController {
         policy: HashMap<JobId, ReschedulePolicy>,
         workers: &HashMap<WorkerId, WorkerNode>,
     ) -> MetaResult<HashMap<DatabaseId, Command>> {
-        let inner = self.metadata_manager.catalog_controller.inner.read().await;
+        let inner = self.metadata_manager.catalog_controller.inner.write().await;
```
kwannoel, line 134: Why do we need to change this to `write`

src/meta/src/stream/scale.rs
```diff
@@ -320,7 +365,7 @@ impl ScaleController {
             return Ok(HashMap::new());
         }
 
-        let inner = self.metadata_manager.catalog_controller.inner.read().await;
+        let inner = self.metadata_manager.catalog_controller.inner.write().await;
```
kwannoel, line 230: Why we change this to write?

src/meta/src/rpc/ddl_controller.rs
```diff
@@ -520,6 +520,25 @@ impl DdlController {
...
+    pub async fn reschedule_streaming_job_backfill_parallelism(
+        &self,
+        job_id: JobId,
+        parallelism: Option<StreamingParallelism>,
+        mut deferred: bool,
+    ) -> MetaResult<()> {
+        tracing::info!("altering backfill parallelism for job {}", job_id);
+        if self.barrier_manager.check_status_running().is_err() {
+            tracing::info!(
+                "alter backfill parallelism is set to deferred mode because the system is in recovery state"
```
kwannoel, line 538: What's the semantic of deferred here? Will it take place after recovery is triggered? Or another restart needs to be added?

Feel like there's two possible behaviours here:
1. Either we error, and tell the user to try again after recovery finishes.
2. We accept the backfill alter, but make sure to also apply it after recovery completes. We should not require another restart after recovery to make it take effect.

src/meta/src/stream/scale.rs
kwannoel: >  In the actual scenario, there is only one request ID here.

Do you mean because in most cases we only backfill one job at a time?

src/meta/src/stream/stream_manager.rs
kwannoel: In that case I'm not sure how the code logic below applies.

Let's say background_jobs = [1, 2], job_id = 1.
Unreschedulable jobs = 2.

Below we bail only if:
```
                bail!(
                    "Cannot alter the job {} because it is a non-reschedulable background backfill job",
                    job_id,
                );
```

We do not check the upstream of `1` at all. We only check if the job itself is unreschedulable.

To me we can simplify it, just do:
```
collect_unrescheduable_backfill_jobs(vec![job_id].into)
```

kwannoel (approved):
Rest LGTM

src/frontend/src/handler/alter_parallelism.rs
shanicky: This is a legacy issue. We used to use `table` to represent streaming jobs, so back then it was called table parallelism. I will refactor this later.

src/meta/src/stream/scale.rs
shanicky: Because this function performs an update operation, all our current write operations to the database use a write lock.

src/meta/src/rpc/ddl_controller.rs
shanicky: Because our regular alter operations are performed through a barrier, and this barrier cannot be used during the recovery phase, any user attempts to execute it at this time will never succeed. Therefore, we automatically switch to deferred mode, directly modify the database, and wait for the next recovery to take effect.

PR #18015, opened 2024-08-13
feat(cdc): support for verify-ca and verify-full of Postgres SSL

StrikeW:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://gist.github.com/TennyZhuang/f00be7f16996ea48effb049aa7be4d66#file-rw_cla).

## What's changed and what's your intention?

- Support `verify-ca` and `verify-full` for `ssl.mode` option for PG cdc source 
- Support `ssl.root.cert` to specify the root certificate secret

Tested with AWS RDS: https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.SSL.html
Example:
```
CREATE SECRET pgcert WITH (backend='meta') AS '-----BEGIN CERTIFICATE-----
...
-----END CERTIFICATE-----';


create source pg_source with (
 connector = 'postgres-cdc',
 hostname = 'pg-ssl-test.cvea8ce4mlos.ap-southeast-1.rds.amazonaws.com',
 port = '5432',
 username = 'siyuan',
 password = 'wsy_940123',
 database.name = 'mydb',
 slot.name = 'pg_slot',
 ssl.mode = 'verify-full',
 ssl.root.cert = SECRET pgcert AS FILE
);
```

close https://github.com/risingwavelabs/risingwave/issues/17785

## Checklist

- [ ] I have written necessary rustdoc comments
- [ ] I have added necessary unit tests and integration tests
- [ ] I have added test labels as necessary. See [details](https://github.com/risingwavelabs/risingwave/blob/main/docs/developer-guide.md#ci-labels-guide).
- [ ] I have added fuzzing tests or opened an issue to track them. (Optional, recommended for new SQL features #7934).
- [ ] My PR contains breaking changes. (If it deprecates some features, please create a tracking issue to remove them in the future).
- [ ] All checks passed in `./risedev check` (or alias, `./risedev c`)
- [ ] My PR changes performance-critical code. (Please run macro/micro-benchmarks and show the results.)

- [ ] My PR contains critical fixes that are necessary to be merged into the latest release. (Please check out the [details](https://github.com/risingwavelabs/risingwave/blob/main/CONTRIBUTING.md))

## Documentation

- [x] My PR needs documentation updates. (Please use the **Release note** section below to summarize the impact on users)

## Release note
- Support `verify-ca` and `verify-full` for `ssl.mode` option for PG cdc source 
- Support `ssl.root.cert` to specify the root certificate secret

</details>

PR #16579, opened 2024-05-04
feat(mysql-cdc): support `ssl.mode` to allow configure the ssl behavior

StrikeW:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://gist.github.com/TennyZhuang/f00be7f16996ea48effb049aa7be4d66#file-rw_cla).

## What's changed and what's your intention?

- Add `ssl.mode` to allow user configure the ssl behavior ([ref](https://dev.mysql.com/doc/connector-j/en/connector-j-connp-props-security.html#cj-conn-prop_useSSL))

related: #15690 

## Checklist

- [ ] I have written necessary rustdoc comments
- [ ] I have added necessary unit tests and integration tests
- [ ] I have added test labels as necessary. See [details](https://github.com/risingwavelabs/risingwave/blob/main/docs/developer-guide.md#ci-labels-guide).
- [ ] I have added fuzzing tests or opened an issue to track them. (Optional, recommended for new SQL features #7934).
- [ ] My PR contains breaking changes. (If it deprecates some features, please create a tracking issue to remove them in the future).
- [ ] All checks passed in `./risedev check` (or alias, `./risedev c`)
- [ ] My PR changes performance-critical code. (Please run macro/micro-benchmarks and show the results.)

- [ ] My PR contains critical fixes that are necessary to be merged into the latest release. (Please check out the [details](https://github.com/risingwavelabs/risingwave/blob/main/CONTRIBUTING.md))

## Documentation

- [x] My PR needs documentation updates. (Please use the **Release note** section below to summarize the impact on users)

Updated by @neverchanje 

This PR involves a change on Postgres connector. It rename the values of [`ssl.mode`](https://docs.risingwave.com/docs/dev/ingest-from-postgres-cdc/) from "disable, prefer, and require" to "disabled, preferred, and required".
On the other hand, the `ssl.mode` option is also supported for the MySQL connector, enabling a SSL/TLS encrypted connection to the upstream MySQL.

## Release note

If this PR includes changes that directly affect users or other significant modifications relevant to the community, kindly draft a release note to provide a concise summary of these changes. Please prioritize highlighting the impact these changes will have on users.

</details>

PR #24340, opened 2026-01-05
feat(expr): `postgres_query` (aka PostgreSQL batch source) SSL support

xiangjinwu:

The `postgres_query` function has two forms:
* (hostname, port, username, password, database_name, query)
* (cdc_source_name, query)

Both forms were hardcoded to pass `tokio_postgres::NoTls`. This PR enhances the second form to pass SSL params from the existing CDC source, as it is more elegant than having 7th or 8th argument in the first form.

Note the first form also defaults to `SslMode::Prefer` now. This is not a breaking change as it still allows connection without ssl.

Related:
* [#16579](https://github.com/risingwavelabs/risingwave/pull/16579) and [#18015](https://github.com/risingwavelabs/risingwave/pull/18015) for SSL in PostgreSQL CDC source

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/e2e_test/sink/postgres_sink.slt
+++ b/e2e_test/sink/postgres_sink.slt
@@ -495,7 +495,7 @@ statement ok
 flush;
 
 # NOTE(kwannoel): postgres_query doesn't support struct etc... yet.
-system ok
+system ok retry 4 backoff 1s
 PGDATABASE=sink_test psql --tuples-only -c "select * from pg_types_table order by id;"
 ----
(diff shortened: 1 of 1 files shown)


src/frontend/src/expr/table_function.rs
```diff
@@ -379,29 +393,45 @@ impl TableFunction {
...
+
+                    // Connection is spawned in create_pg_client, so we don't need to spawn it here.
+                    // But create_pg_client in the connector code spawns it.
+                    // Wait, create_pg_client implementation I saw in `connector_common/postgres.rs` spawns it.
+                    // So we just take the client.
+                    // However, we need to check if `create_pg_client` definition I saw earlier matches what I thought.
+                    // The one in `src/connector/src/connector_common/postgres.rs` is:
+                    // pub async fn create_pg_client(...) -> anyhow::Result<PgClient>
+                    // And it does `tokio::spawn`.
+                    // So I can remove the tokio::spawn block here.
```
tabVersion: maybe rewrite to more formal comments?

tabVersion (approved):
the impl LGTM, better to have one slt test

src/frontend/src/expr/table_function.rs
xiangjinwu: Removed

Issue #24337, opened 2026-01-05
feat: explain datafusion plan to show physical plan instead of logical plan


chenzl25:
Is it possible to update the explain datafusion plan to show the physical plan instead of the logical plan, so that we can know the final execution plan after the join swapping. Let me improve it later.

_Originally posted by @chenzl25 in https://github.com/risingwavelabs/risingwave/issues/24334#issuecomment-3709207031_

PR #24345, opened 2026-01-05
feat(iceberg): support explain datafusion physical plan

chenzl25:

- Support explain datafusion physical plan instead of logical plan. 
- Resolve https://github.com/risingwavelabs/risingwave/issues/24337

### Example: 

```
dev=> explain select count(*) from lineitem;
                                        QUERY PLAN
-------------------------------------------------------------------------------------------
 AggregateExec: mode=Final, gby=[], aggr=[count(Int32(1))]
   CoalescePartitionsExec
     AggregateExec: mode=Partial, gby=[], aggr=[count(Int32(1))]
       CooperativeExec
         IcebergScan: iceberg_scan_type=ICEBERG_SCAN_TYPE_DATA_SCAN, table=public.lineitem
```

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/datafusion/execute.rs
+++ b/src/frontend/src/datafusion/execute.rs
@@ -15,7 +15,7 @@
 use std::sync::Arc;
 
 use datafusion::config::ConfigOptions;
-use datafusion::physical_plan::execute_stream;
+use datafusion::physical_plan::{ExecutionPlan, execute_stream};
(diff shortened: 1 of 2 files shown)


silver-ymz:
Now we can only print physical plan. Maybe we can use `OptimizerContext::trace` to print datafusion's logical plan in trace mode.

PR #24346, opened 2026-01-05
feat: support alter compaction group config via sql

Li0k:

This PR adds SQL support for updating Hummock compaction group configs with `ALTER COMPACTION GROUP ... SET ...`.

It introduces parser/AST support, a frontend handler that validates superuser access and converts SQL config assignments into Hummock mutable config updates, and meta-side handling for compression algorithm reset/default behavior. It also adds SLT coverage for numeric, boolean, string, multi-group, DEFAULT, and invalid-config cases.

## Checklist

- [x]  Fuzzing tests are not needed for this targeted admin SQL handler. 
- [x]  This PR contains no breaking changes. 
- [x]  This PR does not change performance-critical code, so benchmarks are not needed. 

## Documentation

- [x]  This PR needs no documentation updates. 

<details>

N/A

</details>

### DIFF
--- a/e2e_test/ddl/alter_compaction_group.slt
+++ b/e2e_test/ddl/alter_compaction_group.slt
@@ -82,6 +82,24 @@ SELECT (compaction_config::jsonb)->>'enableEmergencyPicker' FROM rw_hummock_comp
 ----
 true
 
+# Test boolean config using unquoted identifier literal
+statement ok
+ALTER COMPACTION GROUP 2 SET enable_emergency_picker = off;
+
+query T
+SELECT (compaction_config::jsonb)->>'enableEmergencyPicker' FROM rw_hummock_compaction_group_configs WHERE id = 2;
+----
+false
+
+# Test boolean config using numeric literal
+statement ok
+ALTER COMPACTION GROUP 2 SET enable_emergency_picker = 1;
+
+query T
+SELECT (compaction_config::jsonb)->>'enableEmergencyPicker' FROM rw_hummock_compaction_group_configs WHERE id = 2;
+----
+true
+
 # Reject comma-separated list RHS (comma is reserved for separating assignments)
 statement error
 ALTER COMPACTION GROUP 2 SET max_bytes_for_level_base = 1,2;
--- a/e2e_test/ddl/alter_compaction_group.slt
+++ b/e2e_test/ddl/alter_compaction_group.slt
@@ -52,6 +52,67 @@ SELECT id, (compaction_config::jsonb)->>'maxBytesForLevelBase' FROM rw_hummock_c
 2 268435456
 3 268435456
 
+# Test ALTER COMPACTION GROUP using TO syntax (config-param style)
+statement ok
+ALTER COMPACTION GROUP 2 SET target_file_size_base TO 134217728;
+
+# Verify TO syntax updated the config
+query I
+SELECT (compaction_config::jsonb)->>'targetFileSizeBase' FROM rw_hummock_compaction_group_configs WHERE id = 2;
+----
+134217728
+
+# Test ALTER COMPACTION GROUP with mixed TO/= in multiple configs
+statement ok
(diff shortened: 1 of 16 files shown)


src/sqlparser/src/parser.rs
```diff
@@ -3152,11 +3152,13 @@ impl Parser<'_> {
             self.parse_alter_secret()
         } else if self.parse_word("FRAGMENT") {
             self.parse_alter_fragment()
+        } else if self.parse_word("COMPACTION") {
```
hzxa21: I think it is better to make `COMPACTION` a keyword and do `self.parse_keywords(&[Keyword::COMPACTION, Keyword::GROUP])` here.

src/sqlparser/src/parser.rs
```diff
@@ -3919,6 +3921,25 @@ impl Parser<'_> {
...
+            return self.expected("GROUP after ALTER COMPACTION");
+        }
+        let mut group_ids = vec![self.parse_literal_u64()?];
+        while self.consume_token(&Token::Comma) {
+            group_ids.push(self.parse_literal_u64()?);
+        }
+        if !self.parse_keyword(Keyword::SET) {
+            return self.expected("SET after ALTER COMPACTION GROUP <id>");
+        }
+        let configs = self.parse_comma_separated(Parser::parse_sql_option)?;
```
hzxa21: The compaction group config is a config rather than a sql option. sql option is something that is in the with clause.

I suggest using `parse_config_param` instead.

src/frontend/src/handler/alter_compaction_group.rs
```diff
@@ -0,0 +1,273 @@
...
+}
+
+fn build_compaction_config_from_options(options: &[SqlOption]) -> Result<Vec<MutableConfig>> {
+    let mut configs = vec![];
+
+    for option in options {
+        let name = option.name.real_value().to_lowercase();
+        let value = &option.value;
+
+        let config = match name.as_str() {
```
hzxa21, line 87: I can see some of the mutable config missing here, like `split_weight_by_vnode`. Is it expected?

src/frontend/src/handler/alter_compaction_group.rs
```diff
@@ -0,0 +1,273 @@
...
+        )
+        .into());
+    }
+
+    let meta_client = handler_args.session.env().meta_client();
+    meta_client
+        .update_compaction_config(group_ids.clone(), configs)
+        .await?;
+
+    Ok(RwPgResponse::builder(StatementType::ALTER_SYSTEM)
```
hzxa21: Technically this is not the standard system parameter. I suggest adding a new state type for this.

e2e_test/ddl/alter_compaction_group.slt
```diff
@@ -0,0 +1,69 @@
...
+SELECT (compaction_config::jsonb)->'compressionAlgorithm'->>6 FROM rw_hummock_compaction_group_configs WHERE id = 2;
+----
+Lz4
+
+# Test ALTER COMPACTION GROUP with multiple groups
+statement ok
+ALTER COMPACTION GROUP 2, 3 SET max_bytes_for_level_base = 268435456;
+
+# Verify both groups were updated
+query II
```
hzxa21: The second column is stirng. so query IT?

zwang28 (approved):
LGTM

src/frontend/src/handler/alter_compaction_group.rs
Li0k: Expected. `split_weight_by_vnode` is managed by the system and should not be user-tunable through SQL, so the handler now rejects it explicitly with an InvalidInputSyntax error instead of silently omitting it.

src/frontend/src/handler/alter_compaction_group.rs
Li0k: This is addressed by modeling it as a dedicated `Statement::AlterCompactionGroup` with its own pgwire `StatementType::ALTER_COMPACTION_GROUP`, instead of routing it through the standard system-parameter path.

Issue #19084, opened 2024-10-24
Support querying metadata of creating stream jobs


kwannoel:
- [ ] Return mv catalog to FE on creating. (Done)
- [ ] Return sink catalog to FE on creating.
- [ ] Return index catalog to FE on creating.
- [ ] rw_catalog: provide a check if mv is background_ddl.
- [ ] rw_catalog: provide a check if sink is background_ddl.
- [ ] rw_catalog: provide a check if index is background_ddl.
- [ ] rw_catalog: provide a check for mv's backfill_rate_limit.
- [ ] rw_catalog: provide a check for sink's backfill_rate_limit.
- [ ] rw_catalog: provide a check for index's backfill_rate_limit.
- [ ] rw_catalog: provide a check for source's backfill_rate_limit.

Standing-Man:
Hi @kwannoel, could you please update the tracking issues? 🙏🏻  It looks like most of them have already been resolved. If my understanding is correct, the following issues appear to have been addressed. 
- [issue 2](https://github.com/risingwavelabs/risingwave/blob/11222b5db87852c84aa7fa571908724b4520267e/src/meta/src/controller/catalog/mod.rs#L1033-L1071)
- [issue 3](https://github.com/risingwavelabs/risingwave/blob/11222b5db87852c84aa7fa571908724b4520267e/src/meta/src/controller/catalog/mod.rs#L1096-L1124)
- [issue 5](https://github.com/risingwavelabs/risingwave/blob/11222b5db87852c84aa7fa571908724b4520267e/src/meta/src/controller/catalog/drop_op.rs#L231-L249)
- [issue 6](https://github.com/risingwavelabs/risingwave/blob/11222b5db87852c84aa7fa571908724b4520267e/src/frontend/src/handler/drop_index.rs#L90-L98)

PR #24349, opened 2026-01-06
feat(frontend): update `IndexCatalog` to support background ddl

Standing-Man:

- Closes Part of #19084.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/e2e_test/background_ddl/check_background_index.slt
+++ b/e2e_test/background_ddl/check_background_index.slt
@@ -2,7 +2,7 @@ statement ok
 create table t(v1 int);
 
 statement ok
-insert into t select * from generate_series(1, 10000);
+insert into t select * from generate_series(1, 500000);
 
 statement ok
 set background_ddl=false;


Standing-Man:
Hi @kwannoel, PTAL when you have time.

e2e_test/background_ddl/check_background_index.slt
```diff
@@ -0,0 +1,36 @@
...
+statement ok
+set background_ddl=true;
+
+statement ok
+CREATE INDEX index2 ON t (v1);
+
+query I
+select background_ddl from rw_catalog.rw_indexes where name='index2';
+----
+t
```
kwannoel, line 27: If the `CREATE INDEX` statement finishes running in the background before we run the check on the catalog, the check will fail.

How about setting the following parameters and increasing the workload size:
```
statement ok
insert into t select * from generate_series(1, 500000);

...

statement ok
set background_ddl=true;

statement ok
set streaming_parallelism = 1;

statement ok
CREATE INDEX index2 ON t (v1);

query I
select background_ddl from rw_catalog.rw_indexes where name='index2';
----
t
```

I wanted to suggest `alter index set backfill_rate_limit`, but realized it's not supported at the moment.

kwannoel:
Rest looks good, thanks for your contribution!

e2e_test/background_ddl/check_background_index.slt
Standing-Man: Thanks for the suggestion. That’s true — this check does have a flaw. I’ve increased the workload size in a new commit, and I’ll take a look to see whether I can add support for it.

PR #24352, opened 2026-01-06
fix(privilege): Use super user for dummy session to avoid permission checks

Chiro11:

As title.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/rpc/mod.rs
+++ b/src/frontend/src/rpc/mod.rs
@@ -133,7 +133,7 @@ async fn get_new_table_plan(
         table_catalog.create_sql_ast_purified()?
     };
 
-    let (source, mut table, graph, job_type) = get_replace_table_plan(
+    let (mut source, mut table, graph, job_type) = get_replace_table_plan(
         &session,
         table_name,
         definition,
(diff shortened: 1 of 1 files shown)


src/frontend/src/session.rs
```diff
@@ -1496,22 +1496,19 @@ impl SessionManager for SessionManagerImpl {
...
-            self.connect_inner(database_id, user_name.as_str(), Arc::new(dummy_addr))
-        } else {
-            bail_catalog_error!("Role id {} does not exist", user_id)
-        }
+
+        // Always use the built-in super user for dummy sessions.
+        // This avoids permission checks tied to a specific user_id.
+        self.connect_inner(
+            database_id,
+            risingwave_common::catalog::DEFAULT_SUPER_USER,
```
yezizp2012, line 1509: I remember that the default super user is able to be dropped. So how about get a random super user from the FE catalog?
Another thing is, will the table owner be changed to the super user?

Chiro11: For 1, the default super user cannot be dropped, refer to https://github.com/risingwavelabs/risingwave/blob/main/src/meta/src/controller/user.rs#L173-L178
For 2, that's right, I'll fix.

yezizp2012 (approved):
LGTM!

Issue #22297, opened 2025-06-19
Materialize UDTF calls in `ProjectSet` with `MaterializedExprs`


stdrc:
Similar to #21029, we need also to apply `MaterializedExprs`-like executor to UDTF calls.

Issue #22296, opened 2025-06-19
Materialize UDF calls in filter conditions with `MaterializedExprs`


stdrc:
#21029 applies `MaterializedExprsExecutor` to materialize the result of UDF calls in `Project`, to avoid stream inconsistency caused by UDF result inconsistency. We need also to apply it on UDF calls in filter conditions.

hzxa21:
@xiangjinwu why do we pin this issue? Is it a blocker?

xiangjinwu:
> [@xiangjinwu](https://github.com/xiangjinwu) why do we pin this issue? Is it a blocker?

To emphasize its incompleteness and shall not be dependent by other features https://github.com/risingwavelabs/risingwave/pull/22835#discussion_r2266040911

PR #21029, opened 2025-03-20
feat(udf): use `StreamMaterializedExprs` for scalar UDF calls

stdrc:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## What's changed and what's your intention?

> AI disclaimer: 90% of lines of Rust code are written by Cursor `claude-3.7-sonnet`, under the guidance and review of me.

This PR creates `StreamMaterializedExprs` for expressions containing UDF calls in `LogicalProject::to_stream`, so that results of indeterministic UDF calls can be materialized 
and retrieved later when updating or deleting the row.

Resolves #12201.

Future work:

1. UDF calls in filter conditions (`WHERE` and `ON`).
2. UDTF calls in `LogicalProjectSet`.
3. Other indeterministic functions like `rand()`.

## Checklist

- [ ] I have written necessary rustdoc comments.
- [ ]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  My PR contains critical fixes that are necessary to be merged into the latest release. 

## Documentation

- [ ]  My PR needs documentation updates. 

<details>
<summary><b>Release note</b></summary>

</details>

PR #24353, opened 2026-01-06
fix(planner): reject impure expressions on retract stream on `to_proto`

BugenZhao:

Calling impure expressions (also including UDF/UDTF, which we cannot tell the pureness) on retract streams will break the consistency. Previously, we introduced `StreamMaterializedExprs` in this case to materialize the evaluation results (#21029) for later retraction.

While it's essential for correctness, `StreamMaterializedExprs` currently doesn't apply to impure expression in all contexts. As a result, we may still encounter inconsistency panic. I think we should check and prevent these unmaterialized impure expression from being planned to execute on retract streams, before we extending  `StreamMaterializedExprs` to be more general.

This PR
- removes session config `streaming_enable_materialized_expressions`, as it should always be enabled for correctness
- adds checks to reject impure expressions on retract stream when calling `to_proto`, unless it's inside `StreamMaterializedExprs` plan node
- introduces session config `streaming_unsafe_allow_unmaterialized_impure_expr` to bypass the check, since currently we don't support extracting impure expressions and materializing them in all places, like UDTF or filter clause in agg call
  * #22296
  * #22297

```
dev=> create materialized view mv as select sum(v) filter (where v > random()) from t;

  1: Scheduler error
  2: Not supported: using an impure expression (RANDOM) in AGGREGATE FILTER condition on a retract stream may lead to inconsistent results
HINT: rewrite the query to extract the impure expression into the select list, or setting `streaming_unsafe_allow_unmaterialized_impure_expr` to allow the behavior at the risk of inconsistent results or panics during execution

dev=> set streaming_unsafe_allow_unmaterialized_impure_expr to true;
SET_VARIABLE
dev=> create materialized view mv as select sum(v) filter (where v > random()) from t;
NOTICE:  using an impure expression (RANDOM) in AGGREGATE FILTER condition on a retract stream may lead to inconsistent results
CREATE_MATERIALIZED_VIEW
```

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/expr/mod.rs
+++ b/src/frontend/src/expr/mod.rs
@@ -91,7 +91,7 @@ pub(crate) fn reject_impure(expr: impl Into<ExprImpl>, context: &str) -> RwResul
             return Err(ErrorCode::NotSupported(
                 msg,
                 "rewrite the query to extract the impure expression into the select list, \
-                 or setting `streaming_unsafe_allow_unmaterialized_impure_expr` to allow \
+                 or set `streaming_unsafe_allow_unmaterialized_impure_expr` to allow \
                  the behavior at the risk of inconsistent results or panics during execution"
                     .into(),
             )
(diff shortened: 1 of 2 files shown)


chenzl25 (approved):
LGTM. You are RisingWave's safeguarder!

BugenZhao:
@codex review

PR #24361, opened 2026-01-06
feat(meta): avoid sending inject barrier request to all workers

wenym1:

Previously we had to send `InjectBarrierRequest` to all worker nodes, even if there is no actors on the nodes. We implement it in this way because: 1. we used to broadcast some global information via the request, and 2. there was a mysterious panic in CI.

For 1, we don't broadcast anything in the request anymore. For 2, the mysterious panic was because of the following. We have two important fields in `InjectBarrierRequest`, `actor_ids_to_collect` and `table_ids_to_sync`. When handling a command, we have two steps, `pre_apply` and `post_apply`. `actor_ids_to_collect` should include the remaining actors between `pre_apply` and `post_apply`, while `table_ids_to_sync` should only include the table_ids after `post_apply`. Therefore, for example, when handling a drop command, the dropped fragments still exist after `pre_apply`, and have been removed after `post_apply`, and then the actors of the fragments should be included in the `actor_ids_to_collect`, while the table_ids should not be included in `table_ids_to_sync`.

To handle this, in this PR, we additionally pass a param `nodes_to_sync_table` in `inject_barrier`. For a node that has actors on it, but not in `nodes_to_sync_table`, we will set an empty `table_ids_to_sync` in `InjectBarrierRequest`.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/ci/scripts/sql/backfill/basic/insert.sql
+++ b/ci/scripts/sql/backfill/basic/insert.sql
@@ -2,5 +2,5 @@ insert into t1
 SELECT
     generate_series,
     '{"orders": {"id": 1, "price": "2.30", "customer_id": 2}}'::jsonb
-FROM generate_series(1, 50000);
+FROM generate_series(1, 10000);
 FLUSH;


wenym1:
* **#24361** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24361?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24361?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ci/scripts/sql/backfill/basic/insert.sql
```diff
@@ -2,5 +2,5 @@ insert into t1
 SELECT
     generate_series,
     '{"orders": {"id": 1, "price": "2.30", "customer_id": 2}}'::jsonb
-FROM generate_series(1, 50000);
+FROM generate_series(1, 10000);
```
yezizp2012, line 5: What is the reason for the modification here? To decrease the CI time?

yezizp2012 (approved):
LGTM

ci/scripts/sql/backfill/basic/insert.sql
wenym1: Yes, without it, the ci backfill test will be timed out.

Issue #24336, opened 2026-01-05
failed to run e2e tests: entered unreachable code: Merge should always have upstream inputs


BugenZhao:
### Buildkite URL

https://buildkite.com/risingwavelabs/main-cron/builds/8311#019b7c01-e638-4a59-8d50-b5f49bfa05b6

### Description and Insights

```
*** await tree context of current task ***


thread 'rw-streaming' (25142) panicked at src/stream/src/executor/merge.rs:455:21:
internal error: entered unreachable code: Merge should always have upstream inputs
Actor 61: `CREATE MATERIALIZED VIEW mv AS SELECT sum(coalesce(pg_sleep(1), v)) FROM t` [190.000ms]
  Epoch <initial> [190.000ms]
    Materialize 3D00000007 [190.000ms]
stack backtrace:
      expect_first_barrier [190.000ms]
        Project 3D00000006 [190.000ms]
          expect_first_barrier [190.000ms]
            SimpleAgg 3D00000005 [190.000ms]
              expect_first_barrier [190.000ms]
                Merge 3D00000004 [190.000ms]  <== current


   0: __rustc::rust_begin_unwind
             at /rustc/b925a865e2c9a0aefe5a2877863cb4df796f2eaf/library/std/src/panicking.rs:698:5
   1: core::panicking::panic_fmt
             at /rustc/b925a865e2c9a0aefe5a2877863cb4df796f2eaf/library/core/src/panicking.rs:80:14
   2: poll_next<risingwave_stream::executor::DynamicReceivers<risingwave_pb::id::TypedId<5, u32>, ()>>
             at ./Fri Jan  2 00:40:39 UTC 2026 [risedev]: Program exited with 134
```

See compute node's logs.

Chiro11:
It seems that `Merge` executor was created without any upstream actor input (when receiving the first barrier).
After discussing with @shanicky, our system should ensure that `Merge` node initially always has at least one upstream actor, so this behavior is unexpected.
BTW: This assumption is not new; it has existed since the early version of our system. So this might be related to some changes on actor scheduler in the last few months.

wenym1:
Encountered this error again in the CI of my PR that doesn't seem related.
https://buildkite.com/risingwavelabs/pull-request/builds/88837#019b8dc6-7a08-4633-a240-adb08fafc6c9

wenym1:
An intersting finding.

CN panicked log
```
*** await tree context of current task ***


[...]

PR #24366, opened 2026-01-07
fix(stream): handle empty upstream inputs in Merge executor gracefully

silver-ymz:

fix failed CI https://buildkite.com/risingwavelabs/pull-request/builds/88963#019b9326-f62f-451a-ac2d-711b9aa350f2

It seems that it's a regression introduced in #24289. I'm not sure about it.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/executor/merge.rs
+++ b/src/stream/src/executor/merge.rs
@@ -451,8 +451,12 @@ where
                 }
 
                 Poll::Ready(None) => {
-                    // See also the comments in `DynamicReceivers::poll_next`.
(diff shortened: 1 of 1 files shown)


BugenZhao:
Is this the same issue as https://github.com/risingwavelabs/risingwave/issues/24336?

Perhaps we should identify the root cause and fix it instead.
Some more discussions here: https://risingwave-labs.slack.com/archives/C094S2N254M/p1767691876158269

Issue #22181, opened 2025-06-10
refactor: handle alter sink connector props with secret


tabVersion:
I think supporting secrets is must as a delivered feature.

_Originally posted by @tabVersion in https://github.com/risingwavelabs/risingwave/pull/20691#discussion_r1998149639_

Standing-Man:
Hi @tabVersion, is this issue still active and in need of a fix?

tabVersion:
Hi @Standing-Man , it is not done yet. And aligned with @xxhZs , he will take this part.

yuhao-su:
Status on `main` as of today, for anyone picking this up (found while fixing #27182):

`ALTER SINK ... CONNECTOR WITH (<key> = SECRET <name>)` is currently a **silent no-op**, not an error.

- The frontend used to reject it with "ALTER SINK does not support SECRET or CONNECTION now". #20780 removed that check and now sends `changed_secret_refs` to meta.
- Meta's `AlterConnectorPropsObject::Sink` branch ([stream_service.rs](https://github.com/risingwavelabs/risingwave/blob/dec0dcb62bc040a35c39b083b3f5fd617c90810b/src/meta/service/src/stream_service.rs#L732-L740)) forwards only `changed_props` and drops `changed_secret_refs`.
- The alter-on-the-fly allowlist is checked against the plaintext keys only, so the secret-ref keys are never checked either.

Observed on a Kafka sink (build of `main` at dec0dcb62b plus the #27182 fix, which does not touch this path):

```sql
CREATE SECRET probe_secret WITH (backend = 'meta') AS 'probe-value';
CREATE SINK s_probe FROM t_probe WITH (connector = 'kafka', properties.bootstrap.server = '127.0.0.1:29092', topic = 't', type = 'upsert', primary_key = 'id');

-- `topic` is not alterable on the fly, yet no error:
ALTER SINK s_probe CONNECTOR WITH (topic = SECRET probe_secret);                    -- ALTER_SINK
-- alterable key, "succeeds" but nothing changes:
ALTER SINK s_probe CONNECTOR WITH (properties.sasl.password = SECRET probe_secret);  -- ALTER_SINK

SELECT definition, connector_props FROM rw_sinks WHERE name = 's_probe';
-- definition and connector_props are identical to before; no `properties.sasl.password` anywhere

-- no dependency was recorded either:
DROP SECRET probe_secret;                                                            -- DROP_SECRET
```

[...]

PR #24369, opened 2026-01-07
feat(sink): support alter secret for sink

xxhZs:
https://github.com/risingwavelabs/risingwave/issues/22181

This pull request expands and refines the support for "alter on the fly" (runtime updatable) configuration fields for a wide range of sink connectors. It does this by marking sensitive or frequently changed fields (such as credentials and user/passwords) as alterable, updating the configuration mapping, and ensuring the backend logic correctly handles secrets during updates. This enhances operational flexibility and security for managing connector configurations.

Key changes include:

**1. Expanded "alter on the fly" field support for sink connectors**  
* Added the `#[with_option(allow_alter_on_fly)]` attribute to credential and sensitive fields in config structs for connectors such as ClickHouse, DeltaLake, Doris, Iceberg, StarRocks, GooglePubSub, Postgres, ElasticSearch/OpenSearch, BigQuery, RedShift, Snowflake, and SqlServer, allowing these fields to be updated at runtime. 

**2. Centralized configuration mapping updates**  
* Updated the static map `SINK_ALLOW_ALTER_ON_FLY_FIELDS` in `allow_alter_on_fly_fields.rs` to include new alterable fields for each relevant connector, ensuring the system knows which fields can be changed without a full redeploy. [[1]](diffhunk://#diff-a365fc0d179182dc053c563a1b8d16f283b155ca2152d47859c9e5a1c40f9652R171-R189) [[2]](diffhunk://#diff-a365fc0d179182dc053c563a1b8d16f283b155ca2152d47859c9e5a1c40f9652R210-R218) [[3]](diffhunk://#diff-a365fc0d179182dc053c563a1b8d16f283b155ca2152d47859c9e5a1c40f9652R255-R258) [[4]](diffhunk://#diff-a365fc0d179182dc053c563a1b8d16f283b155ca2152d47859c9e5a1c40f9652R267-R313)

**3. Backend logic enhancements for secret handling**  
* Modified the sink and iceberg table property update logic in `stream_service.rs` to correctly process and fill secrets when alterable fields are updated, improving security and correctness during runtime changes.

**4. Trait and derive improvements for config structs**  
* Added `WithOptions` derive and trait implementations for config structs and enums (e.g., `SslMode`), supporting the new field attributes and ensuring consistent behavior across connectors. [[1]](diffhunk://#diff-4ab41fb683dd4d2bebab44ea23b282744e9768828cef3f381db3b38be6c4f069L45-R52) [[2]](diffhunk://#diff-8452e501aa49e64da3d4e0650e43b839ae927548d46cafccec1b34b412756c98R69-R72)

**5. Minor codebase cleanup**  
* Removed unused dependencies and imports, such as `indexmap` from `meta/Cargo.toml` and `streaming_job.rs`, to keep the codebase tidy. [[1]](diffhunk://#diff-70b47595d20570400ebd4c680a4ea950f9d03b538ce682617b1333183cfd8bacL36) [[2]](diffhunk://#diff-9a5dc3bbf08e45fc3db11e16f25f50ef94d92abf57e71835ff09074fc0ae829cL19)I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## Checklist

## Documentation

<details>

</details>


tabVersion (commented):
can you add slt to cover the code path change?

hzxa21:
Ping for update

tabVersion (commented):
## Code Review for PR #24369

This PR significantly expands alter-on-fly support for sink secrets across many connectors. The connector-level annotations, static map updates, and SLT tests are well done. However, there are several issues in the meta layer code that need attention.

---

### 🔴 Bug: Double-quoting in `update_stmt_with_props_and_secrets`

**File:** `src/meta/src/controller/streaming_job.rs` — `update_stmt_with_props_and_secrets`

```rust
let sql_option = SqlOption::try_from((k, &format!("'{}'", v)))
```

`SqlOption::try_from` already wraps the value in single quotes internally (see `src/sqlparser/src/ast/mod.rs:3296-3297`: `let query = format!("{} = '{}'", object_name, escaped_value)`). Pre-wrapping with `format!("'{}'", v)` causes double-quoting. The stored SQL definition will contain `key = '''value'''` instead of `key = 'value'`.

This is a **pre-existing bug** in the current `format_with_option_secret_resolved` on main (line 2340), but this PR copies it into the new `update_stmt_with_props_and_secrets` function.

**Fix:** Change to `SqlOption::try_from((k, &v.to_owned()))` or just `SqlOption::try_from((k, v))`.

---

### 🔴 Bug: Iceberg source `properties` and `secret_ref` not updated

**File:** `src/meta/src/controller/streaming_job.rs` — `update_iceberg_table_props_by_table_id`

The iceberg table ALTER path updates:
- **Sink**: `properties` + `secret_ref` + `definition` ✅
- **Source**: only `definition` ❌
- **Table**: only `definition` (acceptable — tables don't have separate properties)

The `source::Model` has `with_properties: Property` (line 34) and `secret_ref: Option<SecretRef>` (line 42) in `src/meta/model/src/source.rs`. After ALTER, the source's runtime properties will be stale.

**Fix:**
```rust
let active_source = source::ActiveModel {
    source_id: Set(source_id),
    with_properties: Set(risingwave_meta_model::Property(
        options_with_secret.as_plaintext().clone(),
    )),
    secret_ref: Set((!options_with_secret.as_secret().is_empty())
        .then(|| SecretRef::from(options_with_secret.as_secret().clone()))),
    definition: Set(definition.clone()),
    ..Default::default()
};
```

---

### 🟡 `validate_alter_config` only sees plaintext props

**File:** `src/meta/src/controller/streaming_job.rs` — `validate_sink_props`

While the PR correctly validates both plaintext and secret ref **keys** against the allowlist (good!), the subsequent `SinkType::validate_alter_config(&new_props)` only merges plaintext props:

```rust
let mut new_props = sink.properties.0.clone();
new_props.extend(props.clone());  // only plaintext
SinkType::validate_alter_config(&new_props)
```

If a required field (e.g., `password`) is changed from plaintext to a secret, the merged config won't have the `password` key, potentially causing validation to fail or skip checking that field. Consider using a placeholder value for secret-backed keys during validation.

---

### 🟡 Code duplication: Secret dependency management

The secret dependency insert/delete block (~20 lines) is duplicated between `update_sink_props_by_sink_id` and `update_iceberg_table_props_by_table_id`. Consider extracting into a helper.

---

### 🟡 Unused imports in `metadata.rs`

```rust
use risingwave_pb::catalog::PbSink;
use risingwave_pb::stream_plan::PbStreamNode;
use crate::barrier::SharedFragmentInfo;
```

These are imported but appear unused in the diff. This will cause compiler warnings (and CI failure with `-Dwarnings`).

---

### 🟢 Good: SLT test covers secret lifecycle

The updated `jdbc.alter_connector_props.slt` is well-structured:
- Creates secrets, creates sink with secret, alters to a different secret
- Validates `connector_props` JSONB structure after alter
- Tests dependency enforcement (can't drop secret while in use)
- Proper cleanup order

One suggestion: add a test case for altering **from secret to plaintext** (e.g., `ALTER SINK ... CONNECTOR WITH (password = 'newpass')`) to cover the `to_remove_secret_dep` path.

---

### 🟢 Good: `validate_sink_props` now validates secret ref keys

This PR correctly chains `props.keys().chain(secret_refs.keys())` to check both plaintext and secret ref keys against the allowlist. This prevents bypassing the allowlist by passing a disallowed field as a secret.

---

### 🔵 Nit: Missing connector field consistency

Copilot's review flagged this: the `IcebergCommon` struct annotations include `s3.iam_role_arn` and `glue.iam_role_arn`, but verify these are included in all three static maps (source, sink, connection). Also, some connectors only allow altering password but not username (e.g., SqlServer allows `sqlserver.password` but not `sqlserver.user`; Postgres allows `password` but not `user`). This is likely intentional but worth a comment explaining the rationale.

---

### 🔵 Nit: `panic!` on unexpected statement type

Pre-existing, but `panic!("definition is not a create sink statement")` will crash the meta node. Consider returning `MetaError` instead.

---

**Summary:** The main blockers are the double-quoting bug and the missing iceberg source properties/secret_ref update. The expanded connector annotations and SLT tests are good work.

src/meta/src/controller/streaming_job.rs
```diff
@@ -3235,23 +3225,41 @@ impl CatalogController {
     }
 }
 
-fn validate_sink_props(sink: &sink::Model, props: &BTreeMap<String, String>) -> MetaResult<()> {
+fn validate_sink_props(
+    sink: &sink::Model,
+    props: &BTreeMap<String, String>,
+    secret_refs: &BTreeMap<String, PbSecretRef>,
+) -> MetaResult<()> {
```
Standing-Man, line 3232: We should add a restriction here to prevent modifying the `connector` field.

tabVersion: I think you should follow https://github.com/risingwavelabs/risingwave/blob/ae6ec4b27aea97df22f6abbfbaa0d628da2162b9/src/connector/src/allow_alter_on_fly_fields.rs#L379

The allow alter props are tagged with `#[with_option(allow_alter_on_fly)]` and use `./risedev generate-with-options` to generate them.

tabVersion:
close as no updates, replace with https://github.com/risingwavelabs/risingwave/pull/25148

Issue #24350, opened 2026-01-06
failed to run unit test `common::table::test_state_table::test_state_table_with_vnode_stats`


BugenZhao:
### Buildkite URL

https://buildkite.com/risingwavelabs/main-cron/builds/8346#019b90b1-9860-43a0-a3fb-94d942b392d1/L5304

### Description and Insights

```
2026-01-06 00:37:52 UTC | running 1 test
-- | --
2026-01-06 00:37:52 UTC | test common::table::test_state_table::test_state_table_with_vnode_stats ... FAILED
2026-01-06 00:37:52 UTC |  
2026-01-06 00:37:52 UTC | failures:
2026-01-06 00:37:52 UTC |  
2026-01-06 00:37:52 UTC | failures:
2026-01-06 00:37:52 UTC | common::table::test_state_table::test_state_table_with_vnode_stats
2026-01-06 00:37:52 UTC |  
2026-01-06 00:37:52 UTC | test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 250 filtered out; finished in 2.07s
2026-01-06 00:37:52 UTC |  
2026-01-06 00:37:52 UTC | stderr ───
2026-01-06 00:37:52 UTC |  
2026-01-06 00:37:52 UTC | thread '<unnamed>' (99306) panicked at src/stream/src/common/table/test_state_table.rs:1966:5:
2026-01-06 00:37:52 UTC | assertion `left == right` failed
2026-01-06 00:37:52 UTC | left: 3
2026-01-06 00:37:52 UTC | right: 1
```

cc @hzxa21

BugenZhao:
Also failing in madsim:

https://buildkite.com/risingwavelabs/main-cron/builds/8376#019b95c1-329a-42de-b640-312a703ce2b6/L6297-L6298

```

2026-01-07 00:38:34 UTC | FAIL [   2.069s] risingwave_stream common::table::test_state_table::test_state_table_pruned_key_range_with_two_pk_columns
-- | --
2026-01-07 00:38:34 UTC | FAIL [   2.068s] risingwave_stream common::table::test_state_table::test_state_table_with_vnode_stats


```

PR #24373, opened 2026-01-07
fix(sink): fix force-append-only sink panicked at inconsistent changes when specified pk does not guarantee uniqueness (#24368)

BugenZhao:

Cherry-pick #24368. Issue #24350.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/executor/sink.rs
+++ b/src/stream/src/executor/sink.rs
@@ -392,7 +392,6 @@ impl<F: LogStoreFactory> SinkExecutor<F> {
                             self.sink_param,
                             self.sink_writer_param,
                             self.non_append_only_behavior,
-                            input_compact_ib,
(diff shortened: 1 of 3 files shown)


src/stream/src/executor/sink.rs
```diff
@@ -630,7 +631,7 @@ impl<F: LogStoreFactory> SinkExecutor<F> {
                         // Compact the chunk to eliminate any unnecessary updates to external systems.
                         // This should be performed against the downstream pk, not the stream key, to
                         // ensure correct retract/upsert semantics from the downstream's perspective.
-                        if sink_type != SinkType::AppendOnly
+                        if !sink_type.is_append_only()
```
BugenZhao, line 634: This also covers `ForceAppendOnly`

PR #24380, opened 2026-01-08
fix(cdc): excluded generated column from postgres publication validation

wcy-fdu:

When creating a PostgreSQL CDC source, if the publication already exists, the validation logic checks whether all columns in the table schema are covered by the publication. However, generated columns are automatically excluded from PostgreSQL logical replication publications, causing false validation errors like:
```
INVALID_ARGUMENT: The publication 'rw_publication' does not cover all columns of the table
```

This issue occurs when:
1. A source is created with a table containing generated columns
2. The source is dropped, leaving the publication behind
3. The source is recreated with the same table, triggering validation against the existing publication

## Solution

Exclude generated columns from publication validation by:
1. Querying generated columns from the upstream PostgreSQL table
2. Skipping validation for these columns when checking publication coverage

This aligns with PostgreSQL's behavior where generated columns are not included in logical replication publications.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/PostgresValidator.java
+++ b/java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/PostgresValidator.java
@@ -507,6 +507,28 @@ public class PostgresValidator extends DatabaseValidator implements AutoCloseabl
             }
         }
 
+        // Query generated columns - they should be excluded from publication validation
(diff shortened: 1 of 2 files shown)


wcy-fdu:
Based on some offline discussion with @hzxa21 , we will make publication (name) unique for per source, to avoid using a outdated publication for a new source. Leave this change in follow up pr.

zwang28:
Could you add a test for PG CDC with generated column?

zwang28 (approved):

PR #24384, opened 2026-01-08
fix(streaming): some executors should always clear the cache on scaling regardless of `cache_may_stale`

BugenZhao:

Recall: When updating the vnode bitmap of an executor during scaling, we will check whether the new vnode bitmap is a **subset** of the original one:

1. If `true`, then we can keep all of the current cache entries. Some keys won't be accessed by this executor any more, but we can lazily evict them.
2. If `false`, then we have to reset/clear the entire cache. Due to lazy eviction in 1), som keys that are not evicted yet might be owned again by this executor, whose content might be stale.

Almost of our executors followed this pattern to determine whether to clear the cache during scaling. However, this is actually only applicable to keyed-cache. If the cache will be consumed entirely or iterated (instead of point-gets), then we should clear the cache eagerly as long as the vnode assignment changes. 

Note: After the actorless refactor (#22932), we now drop and rebuild all actors (then executors) on compute nodes after each scaling operation, eliminating the issue. However, since fully rebuilding actors might not be the most efficient solution, we may reintroduce incremental scaling in the future, which will still depend on this functionality to operate correctly. cc @shanicky 

---

In this PR, I reviewed the executor implementations and discovered that some executors are affected by this issue:

- Executors leveraging `SortBuffer`
	- EOWC Sort (affecting downstream EOWC Over-window)
	- EOWC Agg
	- EOWC Gap Fill
- FS-like executors
    - File-system Fetch
    - Iceberg Fetch

This PR fixes them and also clarify the semantics of `cache_may_stale` by renaming it to `keyed_cache_may_stale`.

Also added nexmark chaos scaling tests to cover EOWC executors.
- q5/q7 eowc (agg, sort)
- q107 eowc (over-window)

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/tests/simulation/tests/integration_tests/scale/nexmark_chaos.rs
+++ b/src/tests/simulation/tests/integration_tests/scale/nexmark_chaos.rs
@@ -90,7 +90,17 @@ async fn nexmark_chaos_common_inner(
 
     sleep(Duration::from_secs(50)).await;
 
-    session.run(select).await?.assert_result_eq(&final_result);
+    let result = session.run(select).await?;
+    if watermark {
+        if result.trim() != final_result.trim() {
+            println!(
+                "Warn: results mismatch, which might be expected since watermark is used.\nDiff:\n{}",
+                pretty_assertions::StrComparison::new(&final_result, &result)
+            )
+        }
+    } else {
+        result.assert_result_eq(&final_result);
+    }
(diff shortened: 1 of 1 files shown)


hzxa21:
Need to cherry-pick to 2.6 and 2.7.

PR #24388, opened 2026-01-08
feat(sink): add keeplive with option for pg rust sink and jdbc sink

xxhZs:
This pull request adds support for configuring TCP keep-alive parameters for JDBC and Postgres sinks, improving connection reliability and allowing fine-tuning of network behavior. The changes introduce new configuration options, propagate them through the codebase, and ensure they are applied when creating database connections. The most important changes are grouped as follows:

**Configuration Additions:**

* Added TCP keep-alive configuration fields (`tcp.keepalive.enable`, `tcp.keepalive.idle`, `tcp.keepalive.interval`, `tcp.keepalive.count`) to `JDBCSinkConfig` in `JDBCSinkConfig.java`, with corresponding getter methods. [[1]](diffhunk://#diff-794c367584a9a349fa4288a45950757f53aa68538241170efd8472bf85c0088cR54-R66) [[2]](diffhunk://#diff-794c367584a9a349fa4288a45950757f53aa68538241170efd8472bf85c0088cR122-R137)
* Added TCP keep-alive configuration fields to `PostgresConfig` in `postgres.rs`, including default values and serde attributes for deserialization. [[1]](diffhunk://#diff-4ab41fb683dd4d2bebab44ea23b282744e9768828cef3f381db3b38be6c4f069R66-R83) [[2]](diffhunk://#diff-4ab41fb683dd4d2bebab44ea23b282744e9768828cef3f381db3b38be6c4f069R100-R111)

**Connection Logic Updates:**

* Updated `JdbcUtils.getConnectionDefault` to use a custom socket factory for TCP keep-alive when enabled, and log the configuration.
* Introduced `TcpKeepaliveConfig` struct and updated `create_pg_client` to accept and apply TCP keep-alive settings when provided, including platform-specific handling. [[1]](diffhunk://#diff-8452e501aa49e64da3d4e0650e43b839ae927548d46cafccec1b34b412756c98R330-R336) [[2]](diffhunk://#diff-8452e501aa49e64da3d4e0650e43b839ae927548d46cafccec1b34b412756c98R345) [[3]](diffhunk://#diff-8452e501aa49e64da3d4e0650e43b839ae927548d46cafccec1b34b412756c98R355-R373)

**Integration and Propagation:**

* Modified `PostgresSinkWriter` to construct and pass `TcpKeepaliveConfig` to `create_pg_client` based on configuration. [[1]](diffhunk://#diff-4ab41fb683dd4d2bebab44ea23b282744e9768828cef3f381db3b38be6c4f069R309-R318) [[2]](diffhunk://#diff-4ab41fb683dd4d2bebab44ea23b282744e9768828cef3f381db3b38be6c4f069R327)
* Updated module exports and imports to include `TcpKeepaliveConfig` for use in relevant components. [[1]](diffhunk://#diff-4cd07579969007131706e6a28d5b568ac4bf7ee6f9749552919b950abbb614d0L40-R40) [[2]](diffhunk://#diff-4ab41fb683dd4d2bebab44ea23b282744e9768828cef3f381db3b38be6c4f069L36-R38)

**CDC Source Handling:**

* Explicitly set TCP keep-alive to `None` for CDC enumerator and CDC external Postgres source to avoid applying keep-alive settings in these contexts. [[1]](diffhunk://#diff-f423ea62859296c02e09d4256f07c2aad9ec49d37583960fe5ae28859e085a50R236) [[2]](diffhunk://#diff-34e4fd0a94c2ca4b44cd560cf12480171d3e31aee2a105cd76e55be8e32b17ceR249)I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## Checklist

## Documentation

<details>

</details>

Test Method
1. Create a new RW cluster.
2. Deploy the local PostgreSQL instance.
3. Create tables and the corresponding sink.
4. Write a piece of data, then monitor the port with the command: sudo tcpdump -i lo0 port 5432
It is observed that there will be heartbeats.

### DIFF
--- a/java/connector-node/risingwave-sink-jdbc/src/main/java/com/risingwave/connector/JDBCKeepaliveSocketFactory.java
+++ b/java/connector-node/risingwave-sink-jdbc/src/main/java/com/risingwave/connector/JDBCKeepaliveSocketFactory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2023 RisingWave Labs
+ * Copyright 2026 RisingWave Labs
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.


java/connector-node/risingwave-sink-jdbc/src/main/java/com/risingwave/connector/JDBCKeepaliveSocketFactory.java
```diff
@@ -0,0 +1,107 @@
+/*
+ * Copyright 2023 RisingWave Labs
```
tabVersion: seems we need also apply license check on java code, cc @BugenZhao

java/connector-node/risingwave-sink-jdbc/src/main/java/com/risingwave/connector/JDBCKeepaliveSocketFactory.java
```diff
@@ -0,0 +1,107 @@
...
+import javax.net.SocketFactory;
+import jdk.net.ExtendedSocketOptions;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Custom SocketFactory that enables TCP keep-alive with configurable parameters. This factory
+ * configures TCP_KEEPIDLE, TCP_KEEPINTVL, and TCP_KEEPCNT at the socket level.
+ */
+public class JDBCKeepaliveSocketFactory extends SocketFactory {
```
tabVersion, line 31: can you share what's the prev prams? Does the pr change default behavior?

src/connector/src/sink/postgres.rs
```diff
@@ -61,6 +63,24 @@ pub struct PostgresConfig {
...
+        rename = "tcp.keepalive.interval"
+    )]
+    #[serde_as(as = "DisplayFromStr")]
+    pub tcp_keepalive_interval: u32,
+    #[serde(
+        default = "default_tcp_keepalive_count",
+        rename = "tcp.keepalive.count"
+    )]
+    #[serde_as(as = "DisplayFromStr")]
+    pub tcp_keepalive_count: u32,
```
tabVersion, line 83: shall we reuse `TcpKeepaliveConfig` defined above?

java/connector-node/risingwave-sink-jdbc/src/main/java/com/risingwave/connector/JDBCKeepaliveSocketFactory.java
tabVersion: ignore me, the path already in HawkEye check, but the CI does not report an error.

xxhZs:
It’s not feasible to add automated tests for this PR, so I verified its validity by manually freezing the cluster and then monitoring TCP with Linux commands.

java/connector-node/risingwave-sink-jdbc/src/main/java/com/risingwave/connector/JDBCKeepaliveSocketFactory.java
xxhZs: This will have no impact. This code is only invoked when isKeepaliveEnabled  is true, and isKeepaliveEnabled  is false by default.

xxhZs: This will have no impact. This code is only invoked when tcp.keepalive.enable  is true, and tcp.keepalive.enable  is false by default.\

tabVersion:
it is a user-facing change please fill the release note.

src/connector/src/sink/postgres.rs
```diff
@@ -61,6 +61,35 @@ pub struct PostgresConfig {
...
+    #[serde(default, rename = "tcp.keepalive.enable")]
+    #[serde_as(as = "DisplayFromStr")]
+    pub tcp_keepalive_enable: bool,
+
+    pub tcp_keepalive: Option<TcpKeepaliveConfig>,
+}
+
+#[serde_as]
+#[derive(Debug, Clone, Deserialize)]
+pub struct TcpKeepaliveConfig {
```
tabVersion, line 74: have you tested the rust impl?

tcp_keepalive should be marked as `#[serde(flatten)]`. Otherwise, the field will always be `None`. 
The props passed here is flattened, `tcp.keepalive.enable` and `tcp.keepalive.idle` will be at the same level.

java/connector-node/risingwave-sink-jdbc/src/main/java/com/risingwave/connector/JDBCKeepaliveSocketFactory.java
```diff
@@ -0,0 +1,107 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(JDBCKeepaliveSocketFactory.class);
+
+    private static int keepaliveIdleSeconds = 60;
+    private static int keepaliveIntervalSeconds = 10;
+    private static int keepaliveCount = 3;
```
tabVersion: I think `static` may lead to concurrency issue. There can be multiple JDBC SINK running on ConnectorNode, and if they do not share the same Keepalive params, they can impact each other.

PR #24392, opened 2026-01-08
feat: risectl command to force change Source Connector Props

tabVersion:

Add an unsafe way (bypass allowance checks) to alter source/table connector props for operational purposes. 

See details in release notes. 

## Checklist

## Documentation

<details>

# 🛠️ Dynamic Source Property Updates - Technical Documentation

## Overview

RisingWave now supports **dynamic updates to source connector properties** without dropping and recreating sources. This document provides comprehensive technical details for operators and administrators.

### 1. `alter-source-properties-safe` - Safe Property Updates

**Purpose**: Orchestrated property update with automatic pause/resume workflow

**Command Syntax**:
```bash
./risedev ctl meta alter-source-properties-safe \
    --source-id <SOURCE_ID> \
    --props <JSON_PROPERTIES> \
    [--reset-splits]
```

**Parameters**:

| Parameter | Type | Required | Description |
|-----------|------|----------|-------------|
| `--source-id` | `u32` | ✅ | Source ID from `rw_sources` table |
| `--props` | `JSON string` | ✅ | Properties to update (key-value pairs) |
| `--reset-splits` | `bool` | ❌ | Reset split assignments after update (default: `false`) |

**Example Usage**:
```bash
# Update Kafka broker address (most common use case)
./risedev ctl meta alter-source-properties-safe \
    --source-id 123 \
    --props '{"properties.bootstrap.server": "new-kafka-cluster:9092"}' \
    --reset-splits

# Update security settings
./risedev ctl meta alter-source-properties-safe \
    --source-id 123 \
    --props '{
        "properties.security.protocol": "SASL_SSL",
        "properties.sasl.mechanism": "PLAIN",
        "properties.sasl.username": "new-user"
    }' 

# Performance tuning
./risedev ctl meta alter-source-properties-safe \
    --source-id 123 \
    --props '{"properties.receive.message.max.bytes": "5000000"}'
```

**Workflow Details**:
2. **Pause**: Injects pause barrier to stop source consumption
3. **Catalog Update**: Updates `rw_sources` table and secret dependencies
4. **Property Propagation**: Sends `ConnectorPropsChange` mutation to executors
5. **Split Reset**: Clears cached splits if `--reset-splits=true`
6. **Resume**: Injects resume barrier to restart consumption

**When to Use**:
- Kafka cluster migrations
- Security credential updates
- Performance tuning
- Any connector property changes

---

### 2. `reset-source-splits` - Split Assignment Reset

**Purpose**: Force re-discovery of source splits by clearing cached state

**Command Syntax**:
```bash
./risedev ctl meta reset-source-splits \
    --source-id <SOURCE_ID>
```

**Parameters**:

| Parameter | Type | Required | Description |
|-----------|------|----------|-------------|
| `--source-id` | `u32` | ✅ | Source ID to reset splits for |

**Example Usage**:
```bash
# Reset splits for major upstream changes
./risedev ctl meta reset-source-splits --source-id 123
```

**Technical Details**:
- Clears `splits` field in `ConnectorSourceWorkerHandle`
- Triggers `force_tick()` to re-discover splits from upstream
- New splits assigned on next SourceManager tick cycle
- **Warning**: May cause data duplication depending on connector behavior

**When to Use**:
- After major upstream topology changes
- When splits appear stale or incorrect
- Following broker address updates with different partition counts

---

### 3. `inject-source-offsets` - Manual Offset Injection

**Purpose**: Inject specific offsets into source splits (UNSAFE - admin only)

**Command Syntax**:
```bash
./risedev ctl meta inject-source-offsets \
    --source-id <SOURCE_ID> \
    --offsets <JSON_OFFSETS>
```

**Parameters**:

| Parameter | Type | Required | Description |
|-----------|------|----------|-------------|
| `--source-id` | `u32` | ✅ | Source ID to inject offsets for |
| `--offsets` | `JSON string` | ✅ | Split ID → offset mapping |

**Example Usage**:
```bash
# Inject specific offsets (format varies by connector)
./risedev ctl meta inject-source-offsets \
    --source-id 123 \
    --offsets '{"0": "1000", "1": "2000", "2": "1500"}'

# For CDC sources with LSN-based offsets
./risedev ctl meta inject-source-offsets \
    --source-id 123 \
    --offsets '{"mysql-bin.000001": "12345", "mysql-bin.000002": "67890"}'
```

**Technical Implementation**:
- Creates `InjectSourceOffsetsMutation` barrier
- Stores offsets as JSON in source executor state tables
- Overwrites existing split state completely
- **Critical**: Incorrect offsets WILL cause data issues

**Offset Format by Connector**:
- **Kafka**: `{"partition_id": "offset_value"}`
- **CDC**: `{"binlog_file": "position"}` or `{"lsn": "value"}`
- **Pulsar**: `{"partition": "message_id"}`

**When to Use**:
- Disaster recovery scenarios
- Skipping corrupted data sections
- Advanced operational troubleshooting
- **Only when you know exact required offsets**

---

### 1. **Safe Property Updates**
```bash
./risedev ctl meta alter-source-properties-safe \
    --source-id $SOURCE_ID \
    --props '{"properties.bootstrap.server": "new-broker:9092"}' \
    --reset-splits  # For major changes
```

### 2. **Getting Source ID**
```bash
# Query source ID from catalog
SOURCE_ID=$(psql -h localhost -p 4566 -d dev -U root -t -A \
    -c "SELECT id FROM rw_sources WHERE name = 'my_source'")
```

### 3. **Monitoring Updates**
```sql
-- Check if properties were updated
SELECT definition FROM rw_sources WHERE name = 'my_source';

-- Monitor source status during update
SELECT * FROM rw_fragment_parallelism WHERE fragment_id IN (
    SELECT fragment_id FROM rw_fragments WHERE table_id = (
        SELECT id FROM rw_sources WHERE name = 'my_source'
    )
);
```

### 4. **Rollback Strategy**
Properties can be reverted using the same command:
```bash
# Rollback to previous settings
./risedev ctl meta alter-source-properties-safe \
    --source-id $SOURCE_ID \
    --props '{"properties.bootstrap.server": "old-broker:9092"}' 
```

---

### Common Issues

1. **Invalid JSON Format**
```bash
# ❌ Wrong: Single quotes
--props '{"key": "value"}'

# ✅ Correct: Escape double quotes
--props "{\"key\": \"value\"}"
```

2. **Connector Type Change**
```bash
# ❌ Not allowed - will fail
./risedev ctl meta alter-source-properties-safe \
    --source-id $SOURCE_ID \
    --props '{"connector": "pulsar"}'  # ERROR!
```

3. **Source Not Found**
```bash
# Verify source exists
psql -c "SELECT name FROM rw_sources WHERE id = 123"
```

### Safety Checks

The system enforces these validations:
- ✅ Connector type cannot be changed
- ✅ Properties must be valid for the connector
- ✅ Secret dependencies are properly managed
- ✅ Catalog updates are atomic within transactions

---

### Mutation Flow
```
Meta Service → GlobalBarrierWorker → Stream Executors → SourceManager
     ↓              ↓                    ↓              ↓
Command → BarrierMutation → Mutation → UpdateProps
```

### State Management
- **Catalog**: `rw_sources.definition` updated atomically
- **Executors**: `ConnectorPropsChange` mutation applied
- **SourceManager**: `UpdateProps` message to enumerator workers
- **State Tables**: Split metadata persisted with new properties

### Recovery Support
All property updates are:
- ✅ Persisted in catalog (survive restarts)
- ✅ Replicated across meta nodes
- ✅ Applied consistently during recovery
- ✅ Idempotent operations

</details>

### DIFF
--- a/java/connector-node/risingwave-source-cdc/src/main/java/io/debezium/connector/binlog/BinlogOffsetContext.java
+++ b/java/connector-node/risingwave-source-cdc/src/main/java/io/debezium/connector/binlog/BinlogOffsetContext.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2025 RisingWave Labs
+ * Copyright 2026 RisingWave Labs
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/java/connector-node/risingwave-source-cdc/src/main/java/io/debezium/connector/binlog/history/BinlogHistoryRecordComparator.java
+++ b/java/connector-node/risingwave-source-cdc/src/main/java/io/debezium/connector/binlog/history/BinlogHistoryRecordComparator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2025 RisingWave Labs
+ * Copyright 2026 RisingWave Labs
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/java/connector-node/risingwave-source-cdc/src/main/java/io/debezium/connector/mysql/MySqlOffsetContext.java
+++ b/java/connector-node/risingwave-source-cdc/src/main/java/io/debezium/connector/mysql/MySqlOffsetContext.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2025 RisingWave Labs
+ * Copyright 2026 RisingWave Labs
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/java/connector-node/risingwave-source-cdc/src/main/java/io/debezium/openlineage/DebeziumOpenLineageEmitter.java
+++ b/java/connector-node/risingwave-source-cdc/src/main/java/io/debezium/openlineage/DebeziumOpenLineageEmitter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2025 RisingWave Labs
+ * Copyright 2026 RisingWave Labs
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/java/connector-node/risingwave-source-cdc/src/main/java/io/debezium/pipeline/ChangeEventSourceCoordinator.java
+++ b/java/connector-node/risingwave-source-cdc/src/main/java/io/debezium/pipeline/ChangeEventSourceCoordinator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2025 RisingWave Labs
+ * Copyright 2026 RisingWave Labs
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/src/expr/impl/src/scalar/array_reverse.rs
+++ b/src/expr/impl/src/scalar/array_reverse.rs
@@ -1,4 +1,4 @@
-// Copyright 2025 RisingWave Labs
+// Copyright 2026 RisingWave Labs
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/src/frontend/src/datafusion/window_function.rs
+++ b/src/frontend/src/datafusion/window_function.rs
@@ -1,4 +1,4 @@
-// Copyright 2025 RisingWave Labs
+// Copyright 2026 RisingWave Labs
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/src/frontend/src/handler/reset_source.rs
+++ b/src/frontend/src/handler/reset_source.rs
@@ -1,4 +1,4 @@
-// Copyright 2025 RisingWave Labs
+// Copyright 2026 RisingWave Labs
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/src/frontend/src/optimizer/plan_node/logical_iceberg_intermediate_scan.rs
+++ b/src/frontend/src/optimizer/plan_node/logical_iceberg_intermediate_scan.rs
@@ -1,4 +1,4 @@
-// Copyright 2025 RisingWave Labs
+// Copyright 2026 RisingWave Labs
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/src/frontend/src/optimizer/plan_visitor/datafusion_execute_checker.rs
+++ b/src/frontend/src/optimizer/plan_visitor/datafusion_execute_checker.rs
@@ -1,4 +1,4 @@
-// Copyright 2025 RisingWave Labs
+// Copyright 2026 RisingWave Labs
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/src/frontend/src/optimizer/rule/apply_table_function_to_project_set_rule.rs
+++ b/src/frontend/src/optimizer/rule/apply_table_function_to_project_set_rule.rs
@@ -1,4 +1,4 @@
-// Copyright 2025 RisingWave Labs
+// Copyright 2026 RisingWave Labs
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/src/frontend/src/optimizer/rule/iceberg_count_star_rule.rs
+++ b/src/frontend/src/optimizer/rule/iceberg_count_star_rule.rs
@@ -1,4 +1,4 @@
-// Copyright 2025 RisingWave Labs
+// Copyright 2026 RisingWave Labs
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/src/frontend/src/optimizer/rule/iceberg_intermediate_scan_rule.rs
+++ b/src/frontend/src/optimizer/rule/iceberg_intermediate_scan_rule.rs
@@ -1,4 +1,4 @@
-// Copyright 2025 RisingWave Labs
+// Copyright 2026 RisingWave Labs
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/src/frontend/src/optimizer/rule/project_top_n_transpose_rule.rs
+++ b/src/frontend/src/optimizer/rule/project_top_n_transpose_rule.rs
@@ -1,4 +1,4 @@
-// Copyright 2025 RisingWave Labs
+// Copyright 2026 RisingWave Labs
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/src/frontend/src/optimizer/rule/source_to_iceberg_intermediate_scan_rule.rs
+++ b/src/frontend/src/optimizer/rule/source_to_iceberg_intermediate_scan_rule.rs
@@ -1,4 +1,4 @@
-// Copyright 2025 RisingWave Labs
+// Copyright 2026 RisingWave Labs
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/src/frontend/src/optimizer/rule/top_n_project_transpose_rule.rs
+++ b/src/frontend/src/optimizer/rule/top_n_project_transpose_rule.rs
@@ -1,4 +1,4 @@
-// Copyright 2025 RisingWave Labs
+// Copyright 2026 RisingWave Labs
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/src/frontend/src/utils/iceberg_predicate.rs
+++ b/src/frontend/src/utils/iceberg_predicate.rs
@@ -1,4 +1,4 @@
-// Copyright 2025 RisingWave Labs
+// Copyright 2026 RisingWave Labs
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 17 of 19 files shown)


tabVersion:
> PR body please? 🙏

Added, details in release notes.

src/stream/src/executor/source/source_executor.rs
```diff
@@ -823,6 +824,80 @@ impl<S: StateStore> SourceExecutor<S> {
                                     );
                                 }
                             }
+
+                            Mutation::InjectSourceOffsets {
```
hzxa21, line 978: I think the reset can only be triggered via risectl as an maintenance operation instead of a online reset. In this case, adding a new mutation seems to be an overkill.

Can we instead just trigger a recovery after reseting the state on meta and leverage the existing mechanism to propagate the new offset to the source executor so that it can use the new offset during initialization after recovery?

@BugenZhao WDYT

hzxa21:
@BugenZhao @shanicky PTAL

src/stream/src/executor/source/source_executor.rs
tabVersion: > I think the reset can only be triggered via risectl as an maintenance operation instead of a online reset. In this case, adding a new mutation seems to be an overkill.

Agree, and this is what I'd designed in the very beginning. 
However, after going through the concrete implementation details, there are a few constraints that made this approach less safe in practice:

1. Offsets live in state tables, not in meta
Source offsets / split progress are persisted by source executors into state tables, not the meta store.
From the meta node side, overwriting these offsets would require using low-level hummock interfaces. Currently, there is no high-level, meta-facing API designed for safe overwrite semantics, and directly writing via internal interfaces felt risky.

2. We need to guarantee the split being overwritten is actually assigned and consumable
If we inject offsets purely from meta without going through split assignment, we may end up writing offsets for splits that:

* are no longer assigned, or
* will never be consumed by any actor after recovery
In that case, the injected offset would silently have no effect, which is hard to reason about operationally.

3. Mutation keeps the ownership and semantics aligned
By using a mutation:

* the overwrite happens on the actor that currently owns the split
* split assignment is already known and validated
* the executor updates its own state in a way consistent with the existing source execution model

In the current workflow, this also happens while the stream is paused, so:

* there is no concurrent consumption
* the overwrite is fast and deterministic

Overall, I agree that a recovery-based approach is conceptually cleaner. But given the current boundary between meta and state tables, the mutation approach felt like the safest way to ensure correctness and observability without introducing risky state manipulation from meta.

src/stream/src/executor/source/source_executor.rs
```diff
@@ -872,6 +873,80 @@ impl<S: StateStore> SourceExecutor<S> {
...
+                                source_id,
+                                split_offsets,
```
hzxa21, line 980: I don't see a place we check whether the split_offsets json values are valid and aligned with the concrete format for a particular connector in both CN and meta. Give that the effect will be persistent, I think we better add some validation logics.

src/stream/src/executor/source/source_executor.rs
```diff
@@ -872,6 +873,80 @@ impl<S: StateStore> SourceExecutor<S> {
...
+                                                    |_| serde_json::json!({ "offset": offset }),
+                                                );
+                                            (split_id.clone(), JsonbVal::from(json_value))
+                                        })
+                                        .collect();
+
+                                    let num_injected = json_states.len();
+                                    if num_injected > 0 {
+                                        self.stream_source_core
+                                            .split_state_store
```
hzxa21: When will the new state get picked up in source reader? Do we need to explicitly trigger a recovery to pick it up?

src/stream/src/executor/source/source_executor.rs
hzxa21: > From the meta node side, overwriting these offsets would require using low-level hummock interfaces

I don't mean letting meta node operate on hummock state table directly. I mean meta updating the [FragmentSplit](https://github.com/risingwavelabs/risingwave/blob/c0655221f3d1a4ab28f733781f820cfeec207527/src/meta/src/controller/scale.rs#L875) state and forcing source executor to pick up the initial split on recovery.

src/stream/src/executor/source/source_executor.rs
tabVersion: I add check for: 1) input json is valid; 2) the split id specified is known by source manager (discovered splits)

The format in state table is quite simple `{"split_id": <offset in string>}` so I did not add check here.

src/stream/src/executor/source/source_executor.rs
tabVersion: fixed in the new commit, update the `split_change` var to trigger reader stream rebuild.

src/stream/src/executor/source/source_executor.rs
tabVersion: > I mean meta updating the [FragmentSplit](https://github.com/risingwavelabs/risingwave/blob/c0655221f3d1a4ab28f733781f820cfeec207527/src/meta/src/controller/scale.rs#L875) state and forcing source executor to pick up the initial split on recovery.

I think that might not work. When startup, source exec always looks up state table for records in state table. It relies fragment_splits for not waiting for meta split assignments, ie. care about split_id rather than exact offset. 

`reset-splits` and `inject-offset` are two different things.
* `reset-splits` cleans all splits records in the source, it will stop and wait for splits being discovered. And apply via SourceChangeSplit. 
* `inject-offset` does not necessarily work together with reset command. It allows changing loc for each split.

src/stream/src/executor/source/source_executor.rs
hzxa21: Why can the newly added test passed with `split_change` not updated? Is there any implicit path to trigger the stream rebuild prior to this fix?

tabVersion: > Why can the newly added test passed with split_change not updated?

It did not. The slt test does not cover this part well.
I have enhanced the test in commit c8aa0ce2c2a3a807eedf8c15a4e08f8daa8c97e3 . Now it checks the total number of messages and make sure it continue consuming new messages.

ci/workflows/pull-request.yml
```diff
@@ -227,7 +227,7 @@ steps:
           <<: *docker-compose
           run: source-test-env
       - ./ci/plugins/upload-failure-logs
-    timeout_in_minutes: 25
+    timeout_in_minutes: 29
```
hzxa21: If one slt test takes 4mins to run, I think it is better to put it in main-corn instead of in PR CI.

java/connector-node/risingwave-source-cdc/src/main/java/io/debezium/connector/binlog/history/BinlogHistoryRecordComparator.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2026 RisingWave Labs
+ * Copyright 2025 RisingWave Labs
```
hzxa21: irrelevant change

java/connector-node/risingwave-source-cdc/src/main/java/io/debezium/connector/binlog/BinlogOffsetContext.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2026 RisingWave Labs
+ * Copyright 2025 RisingWave Labs
```
hzxa21: ditto

java/connector-node/risingwave-source-cdc/src/main/java/io/debezium/connector/mysql/MySqlOffsetContext.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2026 RisingWave Labs
+ * Copyright 2025 RisingWave Labs
```
hzxa21: ditto

java/connector-node/risingwave-source-cdc/src/main/java/io/debezium/openlineage/DebeziumOpenLineageEmitter.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2026 RisingWave Labs
+ * Copyright 2025 RisingWave Labs
```
hzxa21: ditto

java/connector-node/risingwave-source-cdc/src/main/java/io/debezium/pipeline/ChangeEventSourceCoordinator.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2026 RisingWave Labs
+ * Copyright 2025 RisingWave Labs
```
hzxa21: ditto

src/expr/impl/src/scalar/array_reverse.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2026 RisingWave Labs
+// Copyright 2025 RisingWave Labs
```
hzxa21: ditto for all the header changes below.

src/stream/src/executor/source/source_executor.rs
```diff
@@ -396,6 +397,107 @@ impl<S: StateStore> SourceExecutor<S> {
...
+                failed_splits
+            )));
+        }
+
+        let num_injected = json_states.len();
+        if num_injected > 0 {
+            // Store the injected offsets as JSON in the state table
+            self.stream_source_core
+                .split_state_store
+                .set_states_json(json_states)
```
hzxa21, line 481: `json_states` is populated via `offsets_to_inject` only rather than a union of pre-existing splits and the modified splits.

That being said, do we assume that the `offsets_to_inject` will always contain all the pre-existing splits? Do we validate this assumption somewhere? I see we only validate the modified splits are a subset of the pre-existing splits in source_manager rather than a identical set.

java/connector-node/risingwave-source-cdc/src/main/java/io/debezium/connector/binlog/history/BinlogHistoryRecordComparator.java
tabVersion: These change are made by `./risedev fix-license-header`, the license header is now aligned to the file creation date.
I think it is related to https://github.com/risingwavelabs/risingwave/pull/24343

ci/workflows/pull-request.yml
tabVersion: sure!

src/stream/src/executor/source/source_executor.rs
tabVersion: No, we don't assume or validate that offsets_to_inject contains all known splits.
The implementation intentionally supports partial offset injection, you can check the set_states_json impl that it only overwrites specific keys and leave other untouched.

PR #24404, opened 2026-01-09
feat(cdc): add `RW_JAVA_LOG` to control Java connector logging granularity

wcy-fdu:

~~As title, recently we found some noisy log in both postgres cdc and sqlserver cdc, whose numbers equals the number of upstream tables. Log explosion can occur when there are many upstream tables.~~

~~This pr filter out harmless Debezium WARN logs that occur during CDC schema discovery to reduce log noise.~~

~~Debezium emits frequent WARN logs during schema discovery that are harmless and don't affect CDC functionality:~~
~~1. PostgreSQL: `Type [...] is already mapped` (duplicate type registrations)~~
~~2. PostgreSQL: `Cannot parse column default value` for function expressions like `schema.function()`~~
~~3. SQL Server: `Cannot parse column default value` for `sysdatetime()`, `getdate()`, `CONVERT()`, etc.~~
~~4. SQL Server: `Mapper for type 'uniqueidentifier' not found`~~

~~These warnings confuse users but have no actual impact - default values are correctly generated by the database.~~

This PR introduces `RW_JAVA_LOG`, an environment variable to control Java-side connector logging by logger-name prefix.
It helps suppress noisy `Debezium` logs even though all Java logs are bridged to Rust under the single target `risingwave_connector_node`.

### Configuration examples
- Tested: suppress TypeRegistry WARN (keep ERROR)
`RW_JAVA_LOG="info,io.debezium.connector.postgresql.TypeRegistry=error"`
- Suppress all Debezium WARN (keep ERROR)
`RW_JAVA_LOG="info,io.debezium=error"`
- More granular: suppress Postgres Debezium subtree WARN (keep ERROR)
`RW_JAVA_LOG="info,io.debezium.connector.postgresql=error"`
- Disable a prefix completely (drop everything including ERROR)
```
RW_JAVA_LOG="info,io.debezium.connector.postgresql.TypeRegistry=off"
```

## Performance

Minimal overhead - most logs fail the first check (`level != WARN`) immediately. For filtered logs, we avoid expensive message formatting and JNI calls.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/java/connector-node/tracing/src/main/java/com/risingwave/tracing/TracingSlf4jImpl.java
+++ b/java/connector-node/tracing/src/main/java/com/risingwave/tracing/TracingSlf4jImpl.java
@@ -25,6 +25,7 @@ import java.util.Comparator;
 import java.util.List;
 import java.util.Locale;
 import java.util.Optional;
+import java.util.concurrent.atomic.AtomicBoolean;
 import org.slf4j.event.Level;
 
 public class TracingSlf4jImpl {
@@ -125,6 +126,7 @@ public class TracingSlf4jImpl {
     }
 
     private static class JavaLoggerLevelOverrides {
+        private static final AtomicBoolean warnedInvalidRwJavaLog = new AtomicBoolean(false);
         private final int defaultMaxBindingLevel;
         private final List<Rule> rules;
 
@@ -161,6 +163,7 @@ public class TracingSlf4jImpl {
         private static JavaLoggerLevelOverrides parse(String raw) {
             int defaultMax = BINDING_TRACE;
             List<Rule> rules = new ArrayList<>();
+            List<String> invalidTokens = new ArrayList<>();
 
             for (String part : raw.split(",")) {
                 String token = part.trim();
@@ -174,6 +177,8 @@ public class TracingSlf4jImpl {
                     Optional<Integer> lvl = parseMaxBindingLevel(token);
                     if (lvl.isPresent()) {
                         defaultMax = lvl.get();
+                    } else {
+                        invalidTokens.add(token);
                     }
                     continue;
                 }
@@ -181,6 +186,7 @@ public class TracingSlf4jImpl {
                 String prefix = token.substring(0, eq).trim();
                 String levelStr = token.substring(eq + 1).trim();
                 if (prefix.isEmpty()) {
+                    invalidTokens.add(token);
                     continue;
                 }
 
@@ -189,6 +195,8 @@ public class TracingSlf4jImpl {
                     Optional<Integer> lvl = parseMaxBindingLevel(levelStr);
                     if (lvl.isPresent()) {
                         defaultMax = lvl.get();
+                    } else {
+                        invalidTokens.add(token);
                     }
                     continue;
                 }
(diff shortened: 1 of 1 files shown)


wcy-fdu:
It's a bit hard to use RUST_LOG to directly raise the log level of these two paths to ERROR because：
The Java logs, when passed to Rust through JNI, all have their target changed to risingwave_connector_node. The original Java logger name is only appended as a field in the log content, so the Rust side cannot differentiate between different Java loggers.

zwang28:
I’m concerned about permanently disabling these logs via hardcoding. While these logs may seem like harmless noise in some scenarios, they could be critical in others.

zwang28:
> It's a bit hard to use RUST_LOG to directly raise the log level of these two paths to ERROR because： The Java logs, when passed to Rust through JNI, all have their target changed to risingwave_connector_node. The original Java logger name is only appended as a field in the log content, so the Rust side cannot differentiate between different Java loggers.

Could we implement more granular control over the Java logging levels?

wcy-fdu:
> Could we implement more granular control over the Java logging levels?

Yes that's the best way...

wcy-fdu:
> > It's a bit hard to use RUST_LOG to directly raise the log level of these two paths to ERROR because： The Java logs, when passed to Rust through JNI, all have their target changed to risingwave_connector_node. The original Java logger name is only appended as a field in the log content, so the Rust side cannot differentiate between different Java loggers.
> 
> Could we implement more granular control over the Java logging levels?

I changed the original hardcoding to match certain specific logs to filtering with different granularities controlled by environment variables. cc @zwang28 @hzxa21

wcy-fdu:
PTAL❤️ cc @zwang28 @hzxa21

java/connector-node/tracing/src/main/java/com/risingwave/tracing/TracingSlf4jImpl.java
```diff
@@ -74,4 +123,119 @@ public static boolean isEnabled(Level level) {
...
+        }
+
+        boolean isEnabled(String loggerName, int bindingLevel) {
+            if (bindingLevel < 0) {
+                return false;
+            }
+            int maxAllowed = defaultMaxBindingLevel;
+            if (loggerName != null && !loggerName.isEmpty()) {
+                for (Rule r : rules) {
+                    if (loggerName.startsWith(r.prefix)) {
```
zwang28, line 154: Should we use `eq`?

java/connector-node/tracing/src/main/java/com/risingwave/tracing/TracingSlf4jImpl.java
```diff
@@ -74,4 +123,119 @@ public static boolean isEnabled(Level level) {
...
+                String token = part.trim();
+                if (token.isEmpty()) {
+                    continue;
+                }
+
+                int eq = token.indexOf('=');
+                if (eq < 0) {
+                    // Bare level => default.
+                    Optional<Integer> lvl = parseMaxBindingLevel(token);
+                    if (lvl.isPresent()) {
```
zwang28, line 178: Should we raise error if parse failed?

zwang28: Or write log.

java/connector-node/tracing/src/main/java/com/risingwave/tracing/TracingSlf4jImpl.java
```diff
@@ -74,4 +123,119 @@ public static boolean isEnabled(Level level) {
...
+                String prefix = token.substring(0, eq).trim();
+                String levelStr = token.substring(eq + 1).trim();
+                if (prefix.isEmpty()) {
+                    continue;
+                }
+
+                // Special prefixes for default.
+                if (prefix.equals("*") || prefix.equalsIgnoreCase("root")) {
+                    Optional<Integer> lvl = parseMaxBindingLevel(levelStr);
+                    if (lvl.isPresent()) {
```
zwang28, line 196: ditto

java/connector-node/tracing/src/main/java/com/risingwave/tracing/TracingSlf4jImpl.java
```diff
@@ -74,4 +123,119 @@ public static boolean isEnabled(Level level) {
...
+                        defaultMax = lvl.get();
+                    }
+                    continue;
+                }
+
+                Optional<Integer> lvl = parseMaxBindingLevel(levelStr);
+                if (lvl.isPresent()) {
```
zwang28, line 205: ditto

zwang28 (approved):
rest LGTM

java/connector-node/tracing/src/main/java/com/risingwave/tracing/TracingSlf4jImpl.java
wcy-fdu: Here, granular control needs to be done by package/sub-logger, so using prefix matching with startsWith (combined with longest prefix priority) is more in line with expectations, while equals is too strict and basically has no effect.

java/connector-node/tracing/src/main/java/com/risingwave/tracing/TracingSlf4jImpl.java
wcy-fdu: Added a warn log.

PR #24406, opened 2026-01-09
fix(test): avoid cleaning up reducer input log file

wenym1:

In #23745, we run `risedev clean-data` before we run the fuzz test reducer. However, the log file to be analyzed will also be deleted by the `clean-data`, and we will fail to run the reducer.

In this PR, we will copy the log file to a temp folder, and pass log file in the temp folder for reducer to analyze it. After running the reducer, we move the original log file to the original log folder, with a prefix in the file name to avoid overwriting the log file generated in the reducer check, and then we can see both the original log files and the newly generated log files in the test output file. See https://buildkite.com/risingwavelabs/main-cron/builds/8412#019ba1ec-b894-47a9-bba1-6b7d9b8fad31 as a successful example. https://buildkite.com/risingwavelabs/main-cron/builds/8396#019b9cd2-3ee5-42ab-ac05-7a4067314200 is the current failure, with no error log and error output.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/ci/scripts/run-fuzz-test.sh
+++ b/ci/scripts/run-fuzz-test.sh
@@ -6,6 +6,40 @@ set -euo pipefail
 export LOGDIR=.risingwave/log
 export RUST_LOG=info
 
+backup_existing_logs() {
+    local dest="$1"
+    echo "backing up ${dest}"
+    if [[ ! -d "${LOGDIR}" ]]; then
+        return
+    fi
+    mkdir -p "${dest}"
+    while IFS= read -r -d '' entry; do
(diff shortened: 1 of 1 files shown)


wenym1:
* **#24406** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24406?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24406?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #24417, opened 2026-01-11
perf: improve l0 seclection

Li0k:

This PR extracts NonOverlapSubLevelPicker into its own module and adds focused unit coverage for select_intervals. The selection logic now consistently filters non‑overlapping sub‑levels and applies the non‑pending‑first policy under optimization.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/meta/src/hummock/compaction/picker/intra_compaction_picker.rs
+++ b/src/meta/src/hummock/compaction/picker/intra_compaction_picker.rs
@@ -199,14 +199,14 @@ impl IntraCompactionPicker {
                         target_sub_level_id = Some(sub_level_id);
                     }
 
+                    task_input_size += level_select_sst.iter().map(|sst| sst.sst_size).sum::<u64>();
+                    task_file_count += level_select_sst.len();
+
                     select_level_inputs.push(InputLevel {
                         level_idx: 0,
                         level_type: LevelType::Nonoverlapping,
                         table_infos: level_select_sst,
                     });
-
-                    task_input_size += input.total_file_size;
-                    task_file_count += input.total_file_count;
                 }
                 select_level_inputs.reverse();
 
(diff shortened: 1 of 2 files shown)


src/meta/src/hummock/compaction/picker/non_overlap_sub_level_picker.rs
```diff
@@ -0,0 +1,992 @@
+// Copyright 2026 RisingWave Labs
```
hzxa21, line 1: Are there any real changes introduced in `non_overlap_sub_level_picker.rs` or is it a pure move from codes in min overlap picker?

src/meta/src/hummock/compaction/picker/intra_compaction_picker.rs
```diff
@@ -201,22 +199,22 @@ impl IntraCompactionPicker {
                         target_sub_level_id = Some(sub_level_id);
                     }
 
+                    task_input_size += level_select_sst.iter().map(|sst| sst.sst_size).sum::<u64>();
+                    task_file_count += level_select_sst.len();
```
hzxa21, line 203: What is the implication for the bug prior to this PR? I know it will cause the candidate tasks contain larger `select_input_size` and `total_file_count` than expected but what is the side effect of this?

src/meta/src/hummock/compaction/picker/non_overlap_sub_level_picker.rs
Li0k: This is a copy/paste operation, with the only modification made to select_intervals.

src/meta/src/hummock/compaction/picker/intra_compaction_picker.rs
Li0k: In short, This will impact the validator's decision-making process, resulting in the rejection of the task.

PR #24421, opened 2026-01-12
refactor(pg-cdc): change the automatically created publication to per source

wcy-fdu:

### issue
Multiple sources sharing the same publication name causes conflicts.

When a source is dropped, the publication remains in PostgreSQL. If a new source reuses the same publication name, it inherits the old publication's configuration (table list, column list, partition strategy), which may be incompatible with the new source's requirements.

**Example scenario:**
1. Create source S1 subscribing to 3 columns → creates `rw_publication` with 3 columns
2. Drop source S1 → `rw_publication` still exists in PostgreSQL
3. Create source S2 subscribing to 5 columns, reusing `rw_publication` →  Fails because publication only has 3 columns

### Solution

Generate unique publication name per source, similar to how `slot.name` is already per-source:
- Before: `publication.name = "rw_publication"` (fixed)
- After: `publication.name = "rw_publication_{uuid}"` (unique per source)

This ensures each source has independent publication configuration without interference.

### Test

Added e2e test `postgres_publication_reuse.slt` demonstrating:
- Fixed publication name causes column mismatch after source drop/recreate
- Per-source publication works correctly

## Checklist

## Documentation

<details>

</details>


wcy-fdu:
Many e2e source tests failed ...

e2e_test/source_legacy/cdc/cdc.share_stream.slt
```diff
@@ -410,12 +409,12 @@ WITH (
 FROM pg_source TABLE 'parallelized_cdc.parallelized_cdc_backfill_test';
 
 
-query I retry 6 backoff 1s
+query I retry 15 backoff 1s
 SELECT COUNT(*) FROM pg_parallelized_cdc_backfill_test_table_1
```
hzxa21, line 413: why will this PR cause the test to run longer?

wcy-fdu: The CI occasionally had some issues that were not synchronized in a timely manner, so I increased the timeout to make it pass.

wcy-fdu: Passed😁

PR #24443, opened 2026-01-13
feat(streaming): add dev config `materialize_force_overwrite_on_no_check` and rename `unsafe_enable_strict_consistency` to `unsafe_disable`

BugenZhao:

Add a streaming developer flag to force `NoCheck` materialize conflict behavior to `Overwrite` at plan build, preventing inconsistent changelog propagation downstream. Rename `unsafe_enable_strict_consistency` to `unsafe_disable_strict_consistency` (default false) to clarify semantics while keeping strict consistency enabled by default.

## Checklist

## Documentation

<details>

</details>


BugenZhao:
> LGTM.
> 
> When should we advise user to enable it?

I think it can be used in combination with per-job `unsafe_enable_strict_consistency = false`.

BTW, I feel like the name of `unsafe_enable_strict_consistency` is misleading. It gives an impression that "enabling strict consistency" is unsafe. Perhaps we can fix it considering there's no production usage yet.

PR #24448, opened 2026-01-14
fix: union all will panic because a hidden column

### LINKED ISSUE #17880: Union `unnest` will panic
### Describe the bug

The might be caused by `project_row_id`. It exist in `BoundSelect` but will be eliminate from output in  `BoundQuery`

```
dev=> select unnest(Array[1,2,3]) union all select unnest(Array[1,2,3]);
 unnest 
--------
      1
      2
      3
      1
      2
      3
(6 rows)


dev=> select unnest(Array[1,2,3]) union all (select unnest(Array[1,2,3]));
ERROR:  Panicked when handling the request: assertion failed: Schema::all_type_eq(inputs.iter().map(|x| x.schema()))
This is a bug. We would appreciate a bug report at:
  https://github.com/risingwavelabs/risingwave/issues/new?labels=type%2Fbug&template=bug_report.yml
server closed the connection unexpectedly
        This probably means the server terminated abnormally
        before or while processing the request.
The connection to the server was lost. Attempting reset: Succeeded.
```

### Error message/log

_No response_

### To Reproduce

_No response_

### Expected behavior

_No response_

_No response_

### The version of RisingWave

_No response_

### Additional context

_No response_

yihong0618:

This patch fix #17880 and some related
union all issues cause there's a hidden project_id some case handle but in union all set_options not.

help with CLAUDE and this is the thrid time I try about this case, first time in 2024 LLM can do nothing
and I found the root case but I do not know rust so keep it that

second is I use VSCode agents when claude 4.0 sonnet almost close but failed

this is the third time Now I know some rust and with 4.5 opus with the issue root cause I found in 
the first time It worked...

amazing about the LLM in the past year and a half

in the blog I post in 2024 LLM can do nothing in risingwave
more: https://github.com/yihong0618/gitblog/issues/296

cc @MrCroxx 

what an amazing year

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/planner/set_operation.rs
+++ b/src/frontend/src/planner/set_operation.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use risingwave_common::catalog::PROJECTED_ROW_ID_COLUMN_NAME;
 use risingwave_common::util::column_index_mapping::ColIndexMapping;
 
 use crate::binder::{BoundSetExpr, BoundSetOperation};
@@ -29,7 +30,7 @@ impl Planner {
     fn strip_projected_row_id(plan: PlanRef) -> PlanRef {
         let schema = plan.schema();
         if let Some(field) = schema.fields.first()
-            && field.name == "projected_row_id"
+            && field.name == PROJECTED_ROW_ID_COLUMN_NAME
         {
             let len = schema.len();
             LogicalProject::with_out_col_idx(plan, 1..len).into()
--- a/src/batch/executors/src/executor/project_set.rs
+++ b/src/batch/executors/src/executor/project_set.rs
@@ -16,7 +16,7 @@ use either::Either;
 use futures_async_stream::try_stream;
 use itertools::Itertools;
 use risingwave_common::array::{ArrayRef, DataChunk};
-use risingwave_common::catalog::{Field, Schema};
+use risingwave_common::catalog::{Field, PROJECTED_ROW_ID_COLUMN_NAME, Schema};
 use risingwave_common::types::{DataType, DatumRef};
 use risingwave_common::util::chunk_coalesce::DataChunkBuilder;
 use risingwave_common::util::iter_util::ZipEqFast;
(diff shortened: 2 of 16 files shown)


src/frontend/planner_test/tests/testdata/input/union.yaml
```diff
@@ -122,4 +122,35 @@
...
-    - binder_error
\ No newline at end of file
+    - binder_error
+
+# issue #17880: UNION with set-returning functions (unnest) should not panic
+- name: test union all with unnest
+  sql: |
+    select unnest(Array[1,2,3]) union all (select unnest(Array[1,2,3]));
+  expected_outputs:
+    - batch_plan
```
BugenZhao, line 132: Would you also add `stream_plan` for these new test cases? Since the hidden column `projected_row_id` may be included in the stream key, I'm not sure if directly stripping them can still result in a correct plan.

yihong0618: added

src/frontend/src/planner/set_operation.rs
```diff
@@ -16,12 +16,28 @@ use risingwave_common::util::column_index_mapping::ColIndexMapping;
...
 use crate::planner::Planner;
 
 impl Planner {
+    /// Strip the hidden `projected_row_id` column if present as the first column.
+    /// This is necessary because `LogicalProjectSet` adds this hidden column for set-returning
+    /// functions like `unnest()`, but it should not participate in set operations.
+    fn strip_projected_row_id(plan: PlanRef) -> PlanRef {
+        let schema = plan.schema();
+        if let Some(field) = schema.fields.first()
+            && field.name == "projected_row_id"
```
BugenZhao: Shall we extract all occurrences of the literal `"projected_row_id"` to a constant, and rename it to something like `_rw_projected_row_id`? In this way, it becomes a reserved identifier in binder and this won't accidentally touch a user column with the same name.

yihong0618: done and I am not sure ci failed is releated

BugenZhao: > I am not sure ci failed is releated

Rebasing the branch should resolve that. Let me fix it.

MrCroxx:
Thank you, yihong! ❤️

PR #24452, opened 2026-01-14
feat(hummock): support table refill runtime state refresh

MrCroxx:

- Keep streaming table-refill runtime state updated through the existing streaming vnode/read-version path, while supporting online targeted updates for serving table-refill vnode mappings so serving refiller runtime state follows serving worker changes without requiring a full recovery.
- Refresh meta-owned table-refill runtime state after Hummock subscribe/recovery: persisted table cache refill policy overrides are rebuilt from catalog config and delivered to CN refiller runtime state.
- Keep the policy override boundary explicit: `ALTER ... SET config(streaming.developer.cache_refill_policy=...)` persists the catalog config, but this PR does not claim ALTER-time online policy update. The running CN refiller observes policy overrides through Hummock subscribe snapshot / recovery refresh.
- Keep storage refiller apply semantics field-presence based, so policy refresh and serving mapping refresh can update independently.

## Checklist

## Documentation

<details>

</details>


MrCroxx:
cc @BugenZhao PTAL at the alter streaming job config part. 🙏

src/meta/src/controller/catalog/alter_op.rs
```diff
@@ -1055,8 +1060,42 @@ impl CatalogController {
...
+            .filter(table::Column::BelongsToJobId.eq(job_id))
+            .into_tuple()
+            .all(&txn)
+            .await?;
+
         txn.commit().await?;
 
+        // Notify compute nodes cache refill policy at once.
+        let cache_refill_policy = table
+            .get("cache_refill_policy")
```
BugenZhao: What about accessing new config `merged`?

src/meta/src/controller/catalog/alter_op.rs
```diff
@@ -1055,8 +1060,42 @@ impl CatalogController {
...
+            .select_only()
+            .column(table::Column::TableId)
+            .filter(table::Column::BelongsToJobId.eq(job_id))
+            .into_tuple()
+            .all(&txn)
+            .await?;
+
         txn.commit().await?;
 
+        // Notify compute nodes cache refill policy at once.
```
BugenZhao: Shall we extract these code to a separate function, i.e., some config entries can be incrementally updated by notification.

BugenZhao (commented):
> cc @BugenZhao PTAL at the alter streaming job config part. 🙏

Rest LGTM!

MrCroxx:
cc @codex PTAL

src/meta/src/controller/catalog/alter_op.rs
MrCroxx: Sure. Let me make sure this PR functions correctly first.

MrCroxx:
cc @codex PTAL again for a detailed review please.

src/storage/src/hummock/event_handler/refiller.rs
```diff
@@ -241,6 +260,66 @@ pub(crate) type SpawnRefillTask = Arc<
         + 'static,
 >;
 
+pub struct TableCacheRefillContext {
+    pub streaming: HashMap<TableId, Bitmap>,
```
Li0k: If I understand correctly, the event-handler already maintains the vnodes in **read_version_mapping**. I’m wondering whether we should maintain the streaming/serving vnode map in the event handler and rely on the existing update mechanism to ensure its consistency. cc @wenym1

Li0k:
Rest LGTM on refiller side

src/storage/src/hummock/event_handler/refiller.rs
MrCroxx: It doesn't maintain its own, just an cloneable ref for reading.

Li0k (approved):
LGTM, thanks. PTAL @hzxa21 @wenym1

src/frontend/src/optimizer/rule/iceberg_intermediate_scan_rule.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2026 RisingWave Labs
```
hzxa21: ditto

src/frontend/src/optimizer/rule/top_n_project_transpose_rule.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2026 RisingWave Labs
```
hzxa21: ditto

src/storage/src/hummock/observer_manager.rs
```diff
@@ -93,7 +99,52 @@ impl ObserverState for HummockObserverNode {
             Info::ClusterResource(resource) => {
                 LicenseManager::get().update_cluster_resource(resource);
             }
+            Info::TableCacheRefillPolicies(policies) => {
```
hzxa21: I don't see a place we propagate the per table policies and the vnode mappings on CN startup. How do we make sure the refill policy is effective across CN restarts?

src/meta/src/controller/catalog/alter_op.rs
```diff
@@ -1176,6 +1181,42 @@ impl CatalogController {
...
+                })
+                .collect(),
+        };
+
+        self.env
+            .notification_manager()
+            .notify_hummock_without_version(
+                Operation::Update,
+                Info::TableCacheRefillPolicies(table_refill_policies),
+            );
```
hzxa21: I think it is okay to not support updating the refill policies on the fly. In other words, just like other overwritable streaming configs, a recovery is required to make sure the config is effective. I think this can simplify this PR.

src/meta/src/serving/mod.rs
```diff
@@ -273,3 +327,129 @@ pub fn start_serving_vnode_mapping_worker(
...
+    active_serving_workers: &[PbWorkerNode],
+) -> HashMap<WorkerId, HashMap<TableId, Bitmap>> {
+    active_serving_workers
+        .iter()
+        .map(|worker| (worker.id, HashMap::new()))
+        .collect()
+}
+
+/// Append vnode bitmap mapping to each worker for each state table id.
+fn append_worker_table_vnode_mapping(
```
zwang28: IIUC table -> bitmap can always be derived from WorkerSlotMapping directly.
If so, can we get rid of ServingTableVnodeMappings maintainence and notification completely, since we can use the existing WorkerSlotMapping directly?

zwang28: i.e. just make serving compute handle `Info::ServingWorkerSlotMappings`.

src/tests/simulation/tests/integration_tests/scale/isolation.rs
```diff
@@ -367,6 +499,175 @@ async fn wait_until(session: &mut Session, sql: &str, target: &str) -> Result<()
...
+
+fn serving_refill_matches(
+    serving_vnodes: &serde_json::Map<String, Value>,
+    table_ids: &[u32],
+) -> bool {
+    table_ids.iter().all(|table_id| {
+        serving_vnodes
+            .get(&table_id.to_string())
+            .and_then(Value::as_array)
+            .is_some_and(|vnodes| !vnodes.is_empty())
```
hzxa21: This only checks that each target table has a non-empty serving vnode array. That proves the runtime state was populated, but it doesn't verify the worker-specific mapping is correct, that the vnodes match meta's serving fragment mapping, or that stale mappings were removed.

Can we make this assert the expected vnode ownership instead? For example, derive the expected serving vnode distribution from meta and compare it with the per-compute `serving` stats for the target internal tables. Otherwise this simulation test can pass even if targeted serving mappings are sent to the wrong worker or contain stale vnodes.

src/storage/src/hummock/event_handler/hummock_event_handler.rs
```diff
@@ -528,6 +718,21 @@ impl HummockEventHandler {
         info!("clear finished");
     }
 
+    fn handle_observer_event(&mut self, event: HummockObserverEvent) {
+        match event {
+            HummockObserverEvent::VersionUpdate(version_update) => {
+                self.handle_version_update(version_update);
+            }
+            HummockObserverEvent::TableRefillRuntimeConfig(_, config) => {
+                Self::apply_table_refill_runtime_config(&mut self.refiller, config);
```
hzxa21, line 703: Let's add a metric in https://github.com/risingwavelabs/risingwave/blob/fee5cd451f57b45e284bd1500015044828d522c6/src/storage/src/hummock/event_handler/hummock_event_handler.rs#L192

to measure the event handle latency for this new event so that we know whether it can slow down the single-threaded event handler

src/storage/src/hummock/event_handler/refiller.rs
```diff
@@ -350,99 +622,138 @@ pub(crate) struct CacheRefillContext {
...
+        if self.delta.insert_sst_level != 0 {
+            // Filter by inheritance filter, also requires recent filter enabled.
+            if !self.context.config.skip_recent_filter
+                && !self.context.config.skip_inheritance_filter
+            {
+                tasks = self.filter_by_inheritance_filter(tasks).await;
+            }
+
+            // Filter by table cache refill vnodes.
+            tasks = self.filter_by_table_cache_refill_policy(tasks);
```
hzxa21: Should we apply the table cache refill policy to L0 (`insert_sst_level == 0`) as well? Any downside of doing so?

src/storage/src/hummock/event_handler/refiller.rs
```diff
@@ -454,264 +765,280 @@ impl CacheRefillTask {
...
+                };
+                futures.push(future);
+            }
         }
 
-        try_join_all(tasks).await?;
+        let handles = futures
+            .into_iter()
+            .map(|future| {
+                tokio::spawn(async move {
```
hzxa21: Spawning each data refill future here changes the timeout/cancellation semantics. The outer `CacheRefillTask::run` still wraps `join_all(tasks)` in `tokio::time::timeout`, but once these per-unit futures are spawned, dropping the `join_all(handles)` future on timeout only stops waiting for the handles; it does not abort the spawned tasks.

Before this PR, the data refill unit futures were awaited directly, so timeout cancellation would drop them. With this change, stale refill work can continue reading object store, decoding blocks, and writing the cache after the timeout has fired, which weakens the timeout/backpressure behavior and can overlap with later version updates. Could we avoid spawning here, or explicitly abort the handles when the outer timeout expires?

hzxa21 (commented):
Haven't finished the full review. Left some early comments so that you can take a look first.   I will continue the review in the meantime.

src/storage/src/hummock/event_handler/refiller.rs
```diff
@@ -245,38 +262,141 @@ pub(crate) type SpawnRefillTask = Arc<
...
+    }
+    (start..end).any(|vnode| bitmap.is_set(vnode))
+}
+
+impl TableCacheRefillContext {
+    /// Check whether the given sstable block should be refilled according to the vnode mapping.
+    fn check_table_refill_vnodes(&self, sstable: &Sstable, block_index: usize) -> bool {
+        let block_smallest_key =
+            FullKey::decode(&sstable.meta.block_metas[block_index].smallest_key)
+                .to_vec()
```
hzxa21: The to_vec can be avoided given that we are not mutating the key.

src/storage/src/hummock/event_handler/refiller.rs
```diff
@@ -245,38 +262,141 @@ pub(crate) type SpawnRefillTask = Arc<
...
+                .into_bytes();
+        let block_largest_key = FullKey::decode(
+            sstable
+                .meta
+                .block_metas
+                .get(block_index + 1)
+                .map(|b| &b.smallest_key)
+                .unwrap_or_else(|| &sstable.meta.largest_key),
+        )
+        .to_vec()
```
hzxa21: ditto

src/storage/src/hummock/event_handler/refiller.rs
```diff
@@ -245,38 +262,141 @@ pub(crate) type SpawnRefillTask = Arc<
...
+        .into_bytes();
+
+        let table_key_range = (
+            Bound::Included(block_smallest_key.user_key.table_key),
+            Bound::Excluded(block_largest_key.user_key.table_key),
+        );
+        let vnode_range = vnode_range(&table_key_range);
```
hzxa21: Just need to modify `vnode_range` to be more generic and accept u8 slice to avoid the to_vec above.

src/storage/src/hummock/event_handler/refiller.rs
```diff
@@ -245,38 +262,141 @@ pub(crate) type SpawnRefillTask = Arc<
...
+
+        let num_bits = self
+            .streaming_vnode_bitmap
+            .as_ref()
+            .map(|b| b.len())
+            .or(self.serving_vnode_bitmap.as_ref().map(|b| b.len()))
+            .unwrap_or(0);
+        if num_bits == 0 {
+            return false;
+        }
```
hzxa21: This seems be be unneccessary because `vnode_range_overlaps_bitmap` naturally checks the bitmap len with same time complexity

src/storage/src/hummock/event_handler/refiller.rs
```diff
@@ -350,99 +622,138 @@ pub(crate) struct CacheRefillContext {
...
-        delta: &SstDeltaInfo,
-    ) -> HummockResult<Vec<TableHolder>> {
-        let tasks = delta
-            .insert_sst_infos
-            .iter()
-            .map(|info| async {
-                let mut stats = StoreLocalStatistic::default();
-                GLOBAL_CACHE_REFILL_METRICS.meta_refill_attempts_total.inc();
+        // Return if no data to refill.
+        if self.delta.insert_sst_infos.is_empty() || self.delta.delete_sst_object_ids.is_empty() {
```
hzxa21: CN written L0 SSTs will have non-empty `insert_sst_infos` but empty `delete_sst_object_ids`.
- It may make sense to skip refilling such SSTs for streaming because uploader will refill newly written SSTs already.
- However, for serving, I think we should consider such SSTs because otherwise the refill is incomplete.

src/storage/src/hummock/event_handler/refiller.rs
```diff
@@ -350,99 +622,138 @@ pub(crate) struct CacheRefillContext {
...
-                    .meta_refill_success_duration
-                    .observe(now.elapsed().as_secs_f64());
-                drop(permit);
+                    .data_refill_block_unfiltered_total
+                    .inc_by(total_block_count);
+            } else {
+                GLOBAL_CACHE_REFILL_METRICS
+                    .data_refill_filtered_total
+                    .inc_by(total_block_count);
+                return tasks;
```
hzxa21: The early return here means the refill policy will be ignored if recent filter is enabled (`skip_recent_filter = false`) and recent filter is not hit. Is this what we want? I thought the refill poilcy should have the highest priority.

src/storage/src/hummock/event_handler/refiller.rs
```diff
@@ -350,99 +622,137 @@ pub(crate) struct CacheRefillContext {
...
+                let task = DataCacheRefillTask {
+                    sst: sst.clone(),
+                    blks: blk_start..blk_end,
+                };
+                tasks.push(task);
+            }
+        }
+
+        // Filter by inheritance filter, also requires recent filter enabled.
+        if self.delta.insert_sst_level != 0
```
hzxa21: Let's mention this requirement in the config description of `skip_inheritance_filter`

src/storage/src/hummock/event_handler/refiller.rs
```diff
@@ -350,99 +622,137 @@ pub(crate) struct CacheRefillContext {
...
+            && !self.context.config.skip_recent_filter
+            && !self.context.config.skip_inheritance_filter
```
hzxa21: How about making `skip_recent_filter` and `skip_inheritance_filter` overridable in per table cache refill policy?

src/storage/src/hummock/event_handler/refiller.rs
```diff
@@ -454,264 +764,275 @@ impl CacheRefillTask {
...
+                for task in originals.iter().take(uright).skip(uleft) {
+                    if filtered.contains(task) {
                         continue;
                     }
-                    if context.config.skip_recent_filter || recent_filter.contains(&(psst.id, pblk))
-                    {
-                        res.insert(unit);
+                    if recent_filter.contains(&(psst.id, pblk)) {
+                        filtered.insert(task.clone());
                     }
```
hzxa21, line 807: The clone is not free because it contains the HybridCacheEntry. We can save this clone by changing from `iter` to `into_iter`

hzxa21 (commented):
Finish review for refiller.rs

src/storage/src/hummock/event_handler/hummock_event_handler.rs
```diff
@@ -338,16 +504,36 @@ impl HummockEventHandler {
             spawn_upload_task,
             buffer_tracker,
         );
-        let refiller = CacheRefiller::new(refill_config, sstable_store, spawn_refill_task);
+        let mut refiller =
+            CacheRefiller::new(role, refill_config, sstable_store, spawn_refill_task);
+        let mut pending_observer_events = VecDeque::new();
+        while let Ok(event) = observer_event_rx.try_recv() {
```
hzxa21: Do we need `new_inner` to drain `observer_event_rx` and replay the remaining events through `pending_observer_events`?

This makes the constructor consume runtime events before the event handler worker starts, and gives those drained observer events a separate dispatch path from the normal `tokio::select!` loop. I'm not sure what ordering/initialization guarantee requires this extra queue.

hzxa21: I feel that the logic here is added just for ease of unit testing, which seems unnecessary. We can always add a method used by test to wait for the initial snapshot explicitly. Please confirm.

src/meta/src/serving/mod.rs
```diff
@@ -46,16 +54,19 @@ impl ServingVnodeMapping {
...
         max_serving_parallelism: Option<u64>,
-    ) -> (HashMap<FragmentId, WorkerSlotMapping>, Vec<FragmentId>) {
+    ) -> (
+        HashMap<FragmentId, WorkerSlotMapping>,
+        HashMap<FragmentId, FragmentParallelismInfo>,
+    ) {
         let mut serving_vnode_mappings = self.serving_vnode_mappings.write();
         let mut upserted: HashMap<FragmentId, WorkerSlotMapping> = HashMap::default();
-        let mut failed: Vec<FragmentId> = vec![];
+        let mut failed: HashMap<FragmentId, FragmentParallelismInfo> = HashMap::default();
```
hzxa21: Do we need `upsert` to return the full `FragmentParallelismInfo` for failed fragments?

It looks like the current callers only check whether `failed` is empty, log the failed fragments, and use the fragment ids to build delete notifications. Returning a `HashMap<FragmentId, FragmentParallelismInfo>` makes the API heavier and requires cloning the parallelism info, but I don't see where the extra info is used.

src/storage/src/hummock/event_handler/refiller.rs
Li0k: I would keep `skip_recent_filter` / `skip_inheritance_filter` as global cache-refill debug knobs in this PR instead of adding per-table overrides.

The per-table policy surface here is intentionally limited to refill admission/scope (`Enabled`, `Disabled`, `Streaming`, `Serving`, `Both`). Making the two filter bypasses table-scoped would expand the config model and control-plane semantics beyond the current runtime-state refresh work, and it would need separate design/validation for how table-level filter bypass interacts with recent-filter and inheritance-filter ordering.

For now I clarified the existing dependency in the global config description: `skip_inheritance_filter` only matters after recent-filter admission, so it has no effect when `skip_recent_filter` is enabled.

src/meta/src/barrier/worker.rs
```diff
@@ -1062,6 +1077,46 @@ use crate::barrier::partial_graph::{
 };
 
 impl<C: GlobalBarrierWorkerContext> GlobalBarrierWorker<C> {
+    async fn refresh_table_refill_runtime_state_after_recovery(context: Arc<C>, env: MetaSrvEnv) {
+        // Table refill is a best-effort runtime optimization. Recovery uses an
```
hzxa21: Should this refresh really be best-effort? `ALTER ... SET CONFIG` explicitly tells the user that `RECOVER` is the activation boundary, and the refill e2e test validates the new policy/mapping immediately after `RECOVER`. However, both snapshot/fetch failures below are only warned and swallowed, and the per-database path calls `mark_ready` before reaching this refresh. The notification API also returns after enqueueing; a missing targeted subscriber is only warned about asynchronously. Therefore recovery can report success while Hummock keeps the previous policy or serving vnode mapping, potentially until another recovery, mapping event, or resubscription. I think building and delivering a coherent runtime snapshot should be a fallible/retryable prerequisite of recovery completion (before `mark_ready`/success reporting), with acknowledgement or reconciliation if completion is meant to guarantee application rather than just enqueueing.

e2e_test/commands/psql_refill_validate.py
```diff
@@ -0,0 +1,607 @@
...
+        conn,
+        """
+        SELECT
+            frag.state_table_id,
+            frag.fragment_id,
+            a.actor_id,
+            a.worker_id,
+            rw_actor_vnodes(a.actor_id)::varchar
+        FROM (
+            SELECT unnest(f.state_table_ids) AS state_table_id, f.fragment_id
```
hzxa21: `state_table_ids` is broader than the set of tables that have a live streaming local read version. In this `table_refill.slt` topology:

- the generated sink-log table is not instantiated because sink-into-table bypasses the log store and directly forwards `processed_input`;
- the two hash-join degree tables are unused for this inner join (`need_degree_table_l/r = false`), so their temporary state tables are dropped and their read versions are unregistered.

The actor-catalog query nevertheless assigns vnode ownership to these three tables. I reproduced the result with `ci-3streaming-2serving-3fe`: `stats.streaming` correctly omitted them, while this oracle consistently reported 9 false mismatches (3 tables × 3 streaming CNs).

Please derive/filter the expected streaming table set from actual live local read-version ownership instead of treating every fragment state table as a streaming read table. This comment only covers the streaming oracle; the serving table-selection semantics have separate issues: https://github.com/risingwavelabs/risingwave/pull/24452#discussion_r3585902708 and https://github.com/risingwavelabs/risingwave/pull/24452#discussion_r3585906188.

src/meta/src/serving/mod.rs
```diff
@@ -273,3 +334,147 @@ pub fn start_serving_vnode_mapping_worker(
...
+    table_vnode_mapping: &mut HashMap<TableId, Bitmap>,
+    worker_id: WorkerId,
+    state_table_ids: &HashSet<TableId>,
+    worker_slot_vnode_mapping: &WorkerSlotMapping,
+) {
+    for (worker_slot_id, bitmap) in &worker_slot_vnode_mapping.to_bitmaps() {
+        if worker_slot_id.worker_id() != worker_id {
+            continue;
+        }
+        for table_id in state_table_ids {
```
hzxa21: **Correctness:** `state_table_ids` is not the set of tables that a serving CN can query. Expanding each fragment mapping to every internal state table makes serving refill operator-private state that no batch query will read.

I reproduced this with the PR e2e topology after `ALTER SINK s3 ... cache_refill_policy = both` and `RECOVER`: all seven `s3` internal tables received serving vnode mappings on both serving CNs, including the sink table and the two hash-join degree tables. The query-visible destination `t3` belongs to a different streaming job, so this ALTER does not apply its policy to `t3` at all. The result is that serving refill warms unrelated internal tables while not controlling the actual destination table.

Please derive serving refill mappings only for query-visible result/destination tables, rather than mechanically copying fragment ownership to all `state_table_ids`. For a sink job with no query-visible result table of its own, the serving-scoped part should not refill its internal operator state.

e2e_test/commands/psql_refill_validate.py
```diff
@@ -0,0 +1,607 @@
...
+    job_id: int,
+    fragments: dict[int, list[int]],
+    serving_rows: list[tuple[int, int, int, list[int]]],
+    internal_table_ids: set[int],
+) -> dict[int, dict[int, list[int]]]:
+    result: dict[int, dict[int, set[int]]] = defaultdict(lambda: defaultdict(set))
+    for row_job_id, fragment_id, worker_id, vnodes in serving_rows:
+        if row_job_id != job_id:
+            continue
+        for table_id in fragments.get(fragment_id, []):
```
hzxa21: This serving oracle encodes the same over-broad behavior as the implementation: it assigns a fragment serving mapping to every internal state table in that fragment. Consequently, the e2e passes precisely when serving CNs refill non-queryable hash-join/sink state, and it would reject a correct implementation that limits refill to query-visible tables.

Please build the serving expectation from the destination/result tables that batch serving can actually scan. In this fixture, `ALTER SINK s3` changes only `s3` internal-table policies; the destination table `t3` is a separate job. The test should either assert that `s3` contributes no serving-refill tables, or ALTER a query-visible table/MV and validate that output table

src/meta/src/serving/mod.rs
hzxa21: Serving should not inherit ownership for every internal state table of every fragment. Serving CNs execute batch queries only against query-visible relations such as tables, materialized views, and indexes; join, stream-scan, sink-log, and other operator state belongs to streaming execution. A sink itself is not queryable, so a sink job should not participate in serving refill at all.

I reproduced this E2E with `CREATE SINK s3 INTO t3`: this loop produced serving mappings for all seven `s3` internal tables, including the unused sink-log table and two unused hash-join degree tables. None of these tables is a serving target. The destination `t3` is independently queryable as a table and should be handled through the table job, not by translating the sink policy.

Consequently `serving`/`both` on a sink currently prewarms SSTs that serving will never query. Please build serving mappings only for the query-visible result table of table/MV/index jobs and exclude sink jobs and internal operator state.

src/meta/src/controller/catalog/mod.rs
```diff
@@ -157,7 +160,85 @@ pub(crate) struct CleanedDirtyStreamingJobs {
...
+
+        if policies_by_job.is_empty() {
+            return Ok(PbTableCacheRefillPolicies::default());
+        }
+
+        let internal_tables: Vec<(TableId, Option<JobId>)> = Table::find()
+            .select_only()
+            .column(table::Column::TableId)
+            .column(table::Column::BelongsToJobId)
+            .filter(table::Column::BelongsToJobId.is_in(policies_by_job.keys().copied()))
```
hzxa21: This snapshot applies every explicit job policy only to rows whose `belongs_to_job_id` matches the job, i.e. internal state tables, without considering whether the job is query-visible.

Serving scope needs different, type-aware target selection:

- table/MV/index jobs should apply it to their query-visible result table;
- sink jobs should not produce serving refill targets because a sink cannot be queried;
- internal operator state should remain relevant only to streaming scope when it has live local read-version ownership.

In the refill E2E, `ALTER SINK s3 ... policy = both` currently assigns `both` to internal ids 32-38. None is a valid serving target. The destination `t3` is a separately queryable table and should be configured through its own table job, not through `s3`. Please avoid materializing the serving half of a sink policy onto its internal tables; consider rejecting/ignoring unsupported serving scope for non-queryable job types.

e2e_test/commands/psql_refill_validate.py
hzxa21: This serving oracle repeats the implementation assumption by projecting each fragment mapping onto every `state_table_id`. Therefore expected and actual agree even when the selected table set is wrong.

More fundamentally, this scenario uses a sink job for `--mode both`, but a sink is not query-visible and should have no serving refill target. In the local `s3 INTO t3` run, both expected and actual contained all seven `s3` internal tables, while every one of those tables is irrelevant to serving. The independently queryable destination `t3` should be tested and configured as a table job, not inferred from the sink.

Please split the coverage:

- keep a sink case for streaming scope, deriving expected tables from live local read-version ownership;
- use a table/MV/index job for serving scope, assert that its query-visible result table is present with exact worker vnode ownership, and assert that internal operator tables and sink jobs are absent.

As written, the test validates internal consistency of the wrong mapping and cannot detect serving target-selection regressions.

src/meta/src/serving/mod.rs
hzxa21: To make the job-type boundary explicit: serving refill should ignore **sink jobs entirely**, because a sink is not a batch-query target. The candidates are `table`, `materialized view`, and `index` jobs (more generally, catalog relations with a query-visible Hummock result table).

For sink-into-table, the queryable destination belongs to the **table job**, not the sink job. Its serving policy/mapping must therefore be derived and configured through that table; `ALTER SINK` must not create serving mappings for the sink internal state.

e2e_test/refill/table_refill.slt
```diff
@@ -0,0 +1,45 @@
...
+create sink s3 into t3 as select t1.v1, t1.v2, t2.v3 from t1 join t2 on t1.v1 = t2.v1;
+
+
+query TTT
+select name, key, value from rw_streaming_job_config;
+---
+
+
+statement ok
+alter sink s3 set config(streaming.developer.cache_refill_policy = 'both');
```
hzxa21: A sink is not query-visible, so the `serving` half of `both` has no valid refill target here. This test currently gives `s3` a serving policy and then treats its internal operator tables as serving targets, which encodes the wrong semantics.

Please use `streaming` for the sink case. Add a separate `serving`/`both` case for a query-visible table, MV, or index and validate only its result table. If `both` remains syntactically accepted for sinks, this case should at least assert that no serving mapping is produced for the sink or its internal state tables.

src/storage/src/hummock/event_handler/refiller.rs
```diff
@@ -454,264 +771,283 @@ impl CacheRefillTask {
...
+                let context = self.context.clone();
+                async move {
+                    let holders = match Self::meta_cache_refill(&context, delta).await {
+                        Ok(holders) => holders,
+                        Err(e) => {
+                            tracing::warn!(error = %e.as_report(), "meta cache refill error");
+                            return;
+                        }
+                    };
+                    Self::data_cache_refill(&context, delta, holders).await;
```
hzxa21: **Local test finding (Medium): the same physical L0 object range is refilled once per compaction-group projection.**

I found this with a local black-box, SST-level E2E test on commit `e578232d46` using SQL/`FLUSH`, `sst-dump` as the physical-block oracle, live refill state, and Prometheus counter deltas.

In a fixture with one base table and six MVs across seven compaction groups, one flush created only three physical objects (`632`, `642`, and `651`), but the version delta exposed 21 logical `SstableInfo` entries: seven per physical object. The projections have different `sst_id`/`table_ids` but the same `object_id`. Because each `SstDeltaInfo` is handled independently here, each projection loads the full physical SST metadata and the L0 insert-only path generates the same admitted `(object_id, block range)` tasks again.

This causes actual repeated object-store range reads, not only redundant cache checks. Every admitted task reaches `ObjectStore::read(path(object_id), range)` before decoding and force-inserting the same block keys. The local measurements were:

- `unit=1`: unique oracle = 12 blocks / 2,420 bytes; observed = 84 started tasks / 84 successful blocks / 16,940 bytes (**7x**).
- `unit=8`: unique oracle = 6 tasks / 48 blocks / 2,072,045 bytes; observed = 42 started tasks / 336 successful blocks / 14,504,315 bytes (**7x**).

This primarily affects Serving-CN L0 insert-only refill after a multi-compaction-group flush. A separate local normal L0-to-L6 compaction probe produced one physical/logical SST and showed no amplification. The general trigger is the same physical `object_id` appearing in multiple deltas.

Could we deduplicate/merge admitted tasks across all deltas by `(object_id, block range)` before starting reads, and add a regression test with one physical object projected into multiple `SstDeltaInfo` values?

hzxa21 (approved):
There are CI failures. Rest LGTM.

Thanks a lot for the efforts to dig into the details, address the comments and conduct rounds of tests to make this feature stable! @Li0k

src/storage/src/hummock/event_handler/refiller.rs
```diff
@@ -350,99 +604,177 @@ pub(crate) struct CacheRefillContext {
...
+        // Skip data cache refill if data disk cache is not enabled.
+        if !self.context.sstable_store.block_cache().is_hybrid() {
+            return tasks;
+        }
+
+        let has_parent_ssts = !self.delta.delete_sst_object_ids.is_empty();
+        // CN-written SSTs are appended to L0 without replacing parent SSTs. Other inserted SSTs
+        // need delete-side evidence for recent and inheritance filtering.
+        if self.delta.insert_sst_infos.is_empty()
+            || (!has_parent_ssts && self.delta.insert_sst_level != 0)
```
hzxa21: Just to confirm. `!has_parent_ssts && self.delta.insert_sst_level != 0` should always be false, right? I am okay with keeping this condition for defensive coding.

Li0k: This is definitely defensive programming. That said, I haven’t designed trivial insert for non-L0 levels right now.

src/storage/src/hummock/event_handler/refiller.rs
Li0k: I ended up splitting the unit at the upper layer. For this PR, I’m not going to break the semantics just to save IOPS for now. Since different CGs map to separate LSM trees, I don’t want their IOs mixed together.

PR #24459, opened 2026-01-14
fix(streaming): fix some bugs for tables with watermark TTL & add tests for append-only table

BugenZhao:

It seems that there are some bugs in current impl for `WATERMARK TTL` tables.

1. `Materialize` with watermark ttl will have inconsistent op to the state table. In this case, we will hit an assertion on conflict behavior to be `Overwrite`, which may not hold.
2. The `StreamScan` used to create downstream jobs should also include the `stream_key` (with the extra watermark column in addition to `pk`), to ensure that the stream key can be correctly derived.

Also add tests for defining watermark TTLs on append-only tables, which exposes the bugs above.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/executor/mview/materialize.rs
+++ b/src/stream/src/executor/mview/materialize.rs
@@ -853,9 +853,6 @@ impl<S: StateStore, SD: ValueRowSerde> MaterializeExecutor<S, SD> {
                         .state_table
                         .commit_may_switch_consistent_op(barrier.epoch, op_consistency_level)
                         .await?;
-                    if !post_commit.inner().is_consistent_op() && !self.cleaned_by_ttl_watermark {
-                        assert_eq!(self.conflict_behavior, ConflictBehavior::Overwrite);
-                    }
 
                     let update_vnode_bitmap = barrier.as_update_vnode_bitmap(self.actor_context.id);
 
--- a/e2e_test/streaming/watermark_ttl_append_only_rowid.slt
+++ b/e2e_test/streaming/watermark_ttl_append_only_rowid.slt
@@ -10,9 +10,6 @@ SET streaming_parallelism TO 1;
 # ------------------------------------------
 
 statement ok
-drop table if exists ttl_ao_rowid_ev cascade;
-
-statement ok
 create table ttl_ao_rowid_ev (
     ts int,
     v varchar,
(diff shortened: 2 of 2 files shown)


src/frontend/planner_test/tests/testdata/output/watermark_ttl.yaml
```diff
@@ -14,3 +14,81 @@
...
+      proc_time timestamptz as proctime(),
+      watermark for proc_time as proc_time with ttl
+    ) append only;
+    select v from t;
+  batch_plan: |-
+    BatchExchange { order: [], dist: Single }
+    └─BatchScan { table: t, columns: [t.v], distribution: SomeShard }
+  stream_plan: |-
+    StreamMaterialize { columns: [v, t._row_id(hidden), t.proc_time(hidden)], stream_key: [t._row_id, t.proc_time], pk_columns: [t._row_id, t.proc_time], pk_conflict: NoCheck, watermark_columns: [t.proc_time(hidden)] }
+    └─StreamTableScan { table: t, columns: [t.v, t._row_id, t.proc_time], stream_scan_type: ArrangementBackfill, stream_key: [t._row_id, t.proc_time], pk: [_row_id], dist: UpstreamHashShard(t._row_id) }
```
BugenZhao, line 94: Highlight: `t.proc_time` is included.

src/stream/src/executor/mview/materialize.rs
```diff
@@ -853,9 +853,6 @@ impl<S: StateStore, SD: ValueRowSerde> MaterializeExecutor<S, SD> {
                         .state_table
                         .commit_may_switch_consistent_op(barrier.epoch, op_consistency_level)
                         .await?;
-                    if !post_commit.inner().is_consistent_op() {
-                        assert_eq!(self.conflict_behavior, ConflictBehavior::Overwrite);
-                    }
```
chenzl25, line 858: Append only table with ttl watermark will hit this branch, right?

BugenZhao: Yes, because append-only table will use `IgnoreDelete` conflict behavior. However, non-append-only tables that are manually specified conflict behavior may also hit this branch.

PR #24461, opened 2026-01-14
fix(sqlparser): also quote keyword identifiers

BugenZhao:

Previously, when unparsing an `Ident` or using the `quote_ident` function for SQL expressions, we did not check if it was a SQL keyword that required quoting.

This PR fixes it to also quote keyword identifiers. Also moved the wrapper `QuoteIdent` into `risingwave_sqlparser` and reuse it in `Ident::from_real_value`

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/sqlparser/src/quote_ident.rs
+++ b/src/sqlparser/src/quote_ident.rs
@@ -79,6 +79,7 @@ fn is_keyword(ident: &str) -> bool {
 mod tests {
     use super::QuoteIdent;
 
+    #[track_caller]
     fn check(input: &str, expect: &str) {
         assert_eq!(QuoteIdent(input).to_string(), expect);
     }


yezizp2012 (approved):
LGTM

src/sqlparser/src/quote_ident.rs
```diff
@@ -0,0 +1,105 @@
...
+            return false;
+        }
+    }
+
+    true
+}
+
+fn is_keyword(ident: &str) -> bool {
+    let ident_upper = ident.to_ascii_uppercase();
+    // NOTE: PostgreSQL only quotes non-UNRESERVED keywords. We stay conservative here
```
xiangjinwu: `non-UNRESERVED` 🤯

src/sqlparser/src/quote_ident.rs
```diff
@@ -0,0 +1,105 @@
...
+        .is_ok()
+}
+
+#[cfg(test)]
+mod tests {
+    use super::QuoteIdent;
+
+    fn check(input: &str, expect: &str) {
+        assert_eq!(QuoteIdent(input).to_string(), expect);
+    }
```
xiangjinwu, line 85: nit: `[track_caller]`

xiangjinwu (approved):

chenzl25:
need to cherry pick to 2.7

PR #24466, opened 2026-01-14
feat(meta): system audit log for operational changes

tabVersion:

Adds a system audit log for operational changes and exposes it via rw_catalog.rw_audit_logs, so admins can query recent ALTER/maintenance actions and their details.

added to audits

```
ALTER SYSTEM
ALTER DATABASE
ALTER CONFIG
ALTER PARALLELISM
ALTER FRAGMENT PARALLELISM
ALTER RATE LIMIT (source/mv/table/sink/fragment)
ALTER RESOURCE GROUP
ALTER SOURCE CONNECTOR
ALTER TABLE CONNECTOR
ALTER CONNECTION CONNECTOR
RECOVER
FLUSH
WAIT
REFRESH
VACUUM
CANCEL JOBS
KILL
```

## Checklist

## Documentation

<details>

</details>


yezizp2012:
Do we need to periodically clean up the data in the audit log table? I'm not sure if the data in this table will keep growing to a large scale. Since system catalog reads take the full data, I'm worried about performance issues.

yezizp2012:
Is there another way to record audit logs other than explicitly calling it at the entrance. Not sure if `macro` can help. It also seems that these requests will be sent to meta, perhaps we can record them on meta side for better consistency.

tabVersion:
> Is there another way to record audit logs other than explicitly calling it at the entrance. Not sure if `macro` can help. It also seems that these requests will be sent to meta, perhaps we can record them on meta side for better consistency.

I have the same feeling impl-wise.
The most important thing for audit log is someone does something. But we only keep the connection info/user info inside SessionImpl and does not pass to meta. 

I am not sure if all rpc requests contains the user info, since we do most privilege checks on the frontend. So introduce a new rpc for recording is the only option. Also, because we do not want to record every rpc, macro might not help. 🤔

tabVersion:
I am finding a new way to impl a meta only and transparent to api logic approach. Mark as draft for now.

PR #24470, opened 2026-01-15
feat(iceberg): support iceberg v3 delete vector (read and write) 

chenzl25:

- This is an experimental feature.
- Minimal support iceberg v3 delete vector. Replace the parquet format position delete with puffin format.
- Support delete vector write(via iceberg sink)
- Support delete vector read (via iceberg-rust) instead of constructing anti join (for position) in RisingWave query plan, so the file scan tasks containing (position) delete files to pass to iceberg-rust directly in v3, while equality deletes still use anti join to process.
- Support iceberg compaction for the iceberg v3 format. The main change is supporting read iceberg v3 table with delete vector, we also delegate the read to the iceberg-rust.
- The v3 table test example could be found in `e2e_test/iceberg/test_case/pure_slt/iceberg_v3.slt`
- Related iceberg-compaction PR https://github.com/nimtable/iceberg-compaction/pull/121
- Related iceberg-rust PR https://github.com/risingwavelabs/iceberg-rust/pull/113

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/src/sink/iceberg/mod.rs
+++ b/src/connector/src/sink/iceberg/mod.rs
@@ -27,17 +27,22 @@ use iceberg::arrow::{
     RecordBatchPartitionSplitter, arrow_schema_to_schema, schema_to_arrow_schema,
 };
 use iceberg::spec::{
-    DataFile, MAIN_BRANCH, Operation, PartitionSpecRef, SchemaRef as IcebergSchemaRef,
-    SerializedDataFile, Transform, UnboundPartitionField, UnboundPartitionSpec,
+    DataFile, FormatVersion, MAIN_BRANCH, Operation, PartitionSpecRef,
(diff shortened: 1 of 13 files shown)


src/connector/src/sink/iceberg/mod.rs
```diff
@@ -340,6 +350,11 @@ pub struct IcebergConfig {
     #[serde(rename = "write_mode", default = "default_iceberg_write_mode")]
     pub write_mode: IcebergWriteMode,
 
+    /// Iceberg format version for table creation.
+    #[serde(rename = "format_version", default = "default_iceberg_format_version")]
+    #[serde_as(as = "DisplayFromStr")]
+    pub format_version: u8,
```
Li0k: refactot u8 to strong type as we do in the iceberg-compaction repository.

Li0k (approved):
Rest LGTM

PR #24473, opened 2026-01-15
fix(test): relax nexmark watermark lower bound in madsim scaling test

### LINKED ISSUE #24351: failed to run madsim e2e test `scale::nexmark_source::nexmark_source_with_watermark`
### Buildkite URL

https://buildkite.com/risingwavelabs/main-cron/builds/8346#019b909b-8645-4869-8af7-2aefb28470bc/L666

### Description and Insights

```
test scale::nexmark_source::nexmark_source_with_watermark ... FAILED

    failures:

    failures:
        scale::nexmark_source::nexmark_source_with_watermark

    test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 152 filtered out; finished in 106.13s
    [0m
[31;1m [0m [31;1mstderr[0m [31;1mâ”€â”€â”€[0m

    [0m[31;1mthread '<unnamed>' (811) panicked at src/tests/simulation/tests/integration_tests/scale/nexmark_source.rs:101:5:[0m
    [31;1mexpected event num in 99000..=100000, got 98991[0m
```

Seems to be related to https://github.com/risingwavelabs/risingwave/pull/24214#issuecomment-3695578315. cc @hzxa21

BugenZhao:

- Lower the watermark test lower bound from 99% to 95% to reduce flakes during scale alignment.
- Document that the 99% bound was heuristic; recent failures in #24351 seem to be expected behavior rather than a bug.
- Closes #24351.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/tests/simulation/tests/integration_tests/scale/nexmark_source.rs
+++ b/src/tests/simulation/tests/integration_tests/scale/nexmark_source.rs
@@ -33,8 +33,9 @@ async fn nexmark_source_with_watermark() -> Result<()> {
 async fn nexmark_source_inner(watermark: bool) -> Result<()> {
     let expected_events = 20 * THROUGHPUT;
     let expected_events_range = if watermark {
-        // If there's watermark, we'll possibly get fewer events.
-        (0.99 * expected_events as f64) as usize..=expected_events
+        // If there's watermark, we'll possibly get fewer events due to alignment to the global
+        // max watermark during scaling. The lower bound is a heuristic to reduce flakes.
+        (0.95 * expected_events as f64) as usize..=expected_events
     } else {
         // If there's no watermark, we'll get exactly the expected number of events.
         expected_events..=expected_events


BugenZhao:
> Does that mean there's still a chance to encounter flaky test failures?

True, but the probability will be significantly reduced. The root cause of missing events is due to skew in different parallelism of `WatermarkFilter` operator. As long as madsim uses a fair scheduling algorithm, the chance of dropping more than 5% events can be very small.

kwannoel:
> > Does that mean there's still a chance to encounter flaky test failures?
> 
> True, but the probability will be significantly reduced. The root cause of missing events is due to skew in different parallelism of `WatermarkFilter` operator. As long as madsim uses a fair scheduling algorithm, the chance of dropping more than 5% events can be very small.

Hmm ok. In my experience with SQLSmith, even very small cases could compound because we run these tests in various pipelines (per PR, release), and each run has 64 parallelism. We can take a try first.

BugenZhao:
> Hmm ok. In my experience with SQLSmith, even very small cases could compound because we run these tests in various pipelines (per PR, release), and each run has 64 parallelism. We can take a try first.

The test has been added for 3y+ and a lower bound of 99% almost always worked. Empirically 95% should never fail. 🤣 If it fails again in the future then I guess there must be something wrong that's worth investigation.

PR #24474, opened 2026-01-15
refactor: introduce LogicalIcebergIntermediateScan to reduce iceberg list count

silver-ymz:

follow by #24161

This PR refactors the Iceberg scan implementation to reduce the number of Iceberg metadata list operations. It introduces an intermediate scan node that defers metadata reads until predicate pushdown and column pruning have been applied, avoiding unnecessary work.

### Iceberg scan pipeline

1. `LogicalSource` → `LogicalIcebergIntermediateScan`
   * Resolve `snapshot_id`
2. Apply predicate pushdown and column pruning on `LogicalIcebergIntermediateScan`
3. `LogicalIcebergIntermediateScan` → `LogicalIcebergScan`
   * Generate scan tasks
   * Generate anti-join for delete files
4. `LogicalIcebergScan` → `BatchIcebergScan`
5. Create partitions for all scan tasks in the plan fragmenter
   * Applies only to the non-DataFusion path

### Additional notes

* The optimizer context maintains a map to ensure that all references to the same Iceberg table within a query use the same `snapshot_id`.
* If the intermediate scan determines that there is no data to read, `LogicalIcebergIntermediateScan` is rewritten to an empty `LogicalValue`.
* For `COUNT(*)` queries without filters, the row count is computed directly from scan task metadata instead of scanning data files.
  * This optimization is currently implemented only in the non-DataFusion path.
  * The DataFusion engine could support this optimization as well, but it has not yet been implemented.

### Benchmark

```
# ignore flush, connect from macbook to aws s3 us-east-1
create table t(id int primary key, name varchar) with(commit_checkpoint_interval = 1) engine = iceberg;
insert into t select i, repeat('x', i + 2) from generate_series(2,5) as s(i);
delete from t where id = 1;
select count(*) from t;
```

Before: `5.299 s ±  0.040 s`
After:  `3.557 s ±  0.044 s`

## Checklist

- [x]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 

## Documentation

<details>

</details>

### DIFF
--- a/src/batch/executors/src/executor/values.rs
+++ b/src/batch/executors/src/executor/values.rs
@@ -35,6 +35,23 @@ pub struct ValuesExecutor {
     chunk_size: usize,
 }
 
+impl ValuesExecutor {
+    #[cfg(test)]
+    pub(crate) fn new(
+        rows: Vec<Vec<BoxedExpression>>,
+        schema: Schema,
+        identity: String,
+        chunk_size: usize,
+    ) -> Self {
+        Self {
+            rows: rows.into_iter(),
+            schema,
+            identity,
+            chunk_size,
+        }
+    }
+}
+
 impl Executor for ValuesExecutor {
     fn schema(&self) -> &Schema {
         &self.schema


src/frontend/src/optimizer/rule/batch/batch_iceberg_count_star.rs
```diff
@@ -30,8 +30,11 @@ impl Rule<Batch> for BatchIcebergCountStar {
             if batch_iceberg.iceberg_scan_type() != IcebergScanType::DataScan {
                 return None;
             }
+            if batch_iceberg.predicate().is_some() {
+                return None;
+            }
             return Some(
-                BatchIcebergScan::new_count_star_with_batch_iceberg_scan(batch_iceberg).into(),
+                BatchIcebergScan::new_count_star_with_batch_iceberg_scan(batch_iceberg)?.into(),
```
chenzl25: I think we can make BatchIcebergCountStar a logical rule, so that datafusion can benefit from this rule. BTW, I think we don't need a special IcebergFileScanTask called `CountStar`, because we can always use a `LogicalValue` instead in the optimizer.

src/frontend/src/optimizer/plan_node/logical_iceberg_intermediate_scan.rs
```diff
@@ -0,0 +1,211 @@
...
+
+        let new_output_column = required_cols
+            .iter()
+            .map(|&i| self.output_columns[i].clone())
+            .collect();
+
+        LogicalIcebergIntermediateScan {
+            base,
+            core,
+            predicate: self.predicate.clone(),
```
chenzl25, line 123: It looks like the iceberg-rust can handle a FileScanTask with predicate referring to columns which an't included in the `project_field_ids`.

silver-ymz: Yes. I’ve reviewed the iceberg-rust implementation. A `FileScanTask` can evaluate predicates on columns that are not in `project_field_ids`. It first reads predicate columns to compute the filter, then reads projected columns.

But this also brings a new trade-off. If a column appears in both predicate and projection, it may be read twice, so pushing all predicates into iceberg-rust is not always optimal. This is also noted in Parquet RowFilter docs [link](https://docs.rs/parquet/latest/parquet/arrow/arrow_reader/struct.RowFilter.html#:~:text=As%20a%20counter,are%20not%20contiguous.). Improving this could be a future optimization.

src/frontend/src/optimizer/rule/batch/batch_iceberg_count_star.rs
silver-ymz: done in new commit

PR #24477, opened 2026-01-15
feat(streaming): state clean for eowc over window numbering functions

yuhao-su:

Added support for persisting window function intermediate states in EOWC to enable state cleaning for numbering functions (row_number, rank, dense_rank).

Previously, window functions like row_number would keep all data in the state table, which is inefficient. This PR adds:

1. A new `intermediate_state_table` field to the EOWC over window node to store intermediate window function states
2. Recovery logic that ensures window functions continue from where they left off

The implementation enables persistence only for numbering functions that need it, while maintaining backward compatibility with existing window functions.

## Checklist

### DIFF
--- a/e2e_test/streaming/eowc/eowc_over_window_numbering_functions.slt
+++ b/e2e_test/streaming/eowc/eowc_over_window_numbering_functions.slt
@@ -172,6 +172,7 @@ DROP TABLE rank_input;
 
 statement ok
 CREATE TABLE mixed_input (
+    id INT PRIMARY KEY,
     tm TIMESTAMPTZ,
     foo INT,
     bar INT,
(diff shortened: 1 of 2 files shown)


yuhao-su:
* **#24477** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24477?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24477?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

yuhao-su:
> LGTM! please cover the case of overwindow without partition key like `row_number() over (order by tm)`.

`row_number() over (order by tm)` was not supported. The error was cover by a test before.  I added `partition by 1` test instead.

PR #24479, opened 2026-01-15
chore(license): polish the license error message

hzxa21:

Simple change. Just added the contact sales info on the error message.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/license/src/feature.rs
+++ b/src/license/src/feature.rs
@@ -76,7 +76,7 @@ pub enum FeatureNotAvailable {
     // TODO(license): refine error message to include tier name & better instructions
     #[error(
         "feature {feature:?} is not available based on your license\n\n\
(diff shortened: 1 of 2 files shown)


src/license/src/feature.rs
```diff
@@ -76,7 +76,7 @@ pub enum FeatureNotAvailable {
     // TODO(license): refine error message to include tier name & better instructions
     #[error(
         "feature {feature:?} is not available based on your license\n\n\
-        Hint: You may want to set a license key with `ALTER SYSTEM SET license_key = '...';` command."
+        Hint: You may want to set a license key with `ALTER SYSTEM SET license_key = '...';` command. Please contact sales at sales@risingwave-labs.com for more information."
```
BugenZhao: Actually I've generated a short link `https://go.risingwave.com/premium`. Shall we use it instead? The contact info is also at the bottom of the page.

hzxa21: Let's use that. Addressed in the latest commit.

Issue #17952, opened 2024-08-07
bug: group top-N cannot propagate watermark other than the first group column


stdrc:
### Describe the bug

https://github.com/risingwavelabs/risingwave/blob/b671bc7d65c38abc3595479f08490b018a086984/src/stream/src/executor/top_n/group_top_n.rs#L240-L247

As we can see in the code, we can only propagate watermark messages on the first group column. However in optimizer, we derive watermark for all group columns (and even all input columns for append only case):

https://github.com/risingwavelabs/risingwave/blob/b671bc7d65c38abc3595479f08490b018a086984/src/frontend/src/optimizer/plan_node/stream_group_topn.rs#L45-L57

I tried to change the optimizer derivation of watermark to only propagate the first group column, then many nexmark queries in EOWC mode will failed to generate stream plan.

So the bug is that, we allow such queries to generate EOWC plans, but actually no watermark can drive the emission of any data.

st1page:
> we can only propagate watermark messages on the first group column

I think the optimizer is right? 

> we derive watermark for all group columns (and even all input columns for append only case):

Maybe the executor should follow it.

PR #24481, opened 2026-01-15
fix(optimizer): align watermark propagation in `StreamGroupTopN` with executor

BugenZhao:

As described in https://github.com/risingwavelabs/risingwave/issues/17952, there are some inconsistency between the derivation and execution on watermark propagation. As a result, queries like `q6_group_top_1` can create in EOWC mode, but the window never closes, resulting in no output forever.

This PR updates the optimizer to align with the actual executor behavior. As a result, such queries will be rejected when creating, which seems more reasonable.

## Checklist

## Documentation

<details>

</details>


chenzl25:
Related PR: https://github.com/risingwavelabs/risingwave/pull/24484

Issue #24468, opened 2026-01-15
bug(connector): bug in debezium's AbstractHistoryRecordComparator


zwang28:
### Describe the bug

When Debezium replays MySQL schema history, it uses AbstractHistoryRecordComparator::isPositionAtOrBefore to compare the offset provided by the OffsetBackingStore (RisingWave) against the offset from records in the Schema History (OpenDAL). It compares several fields from the offsets in order, e.g. server_id, timestamp, binlog file name, binlog file position, etc. 

However, two bugs in how Debezium [constructs the new offset instance from OffsetBackingStore's offset](https://github.com/debezium/debezium/blob/4159397bd40dd5d0c8efe2b97ba61ba8eb220b58/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlOffsetContext.java#L84) cause this method to consistently return true, regardless of the actual position.
- [Set server_id from sourceInfo](https://github.com/debezium/debezium/blob/4159397bd40dd5d0c8efe2b97ba61ba8eb220b58/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlOffsetContext.java#L87). However, it's always 0.
- [Set timestamp field as ts_sec](https://github.com/debezium/debezium/blob/4159397bd40dd5d0c8efe2b97ba61ba8eb220b58/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlOffsetContext.java#L102), However, [ts_ms is read when comparing offsets](https://github.com/debezium/debezium/blob/4159397bd40dd5d0c8efe2b97ba61ba8eb220b58/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/strategy/AbstractHistoryRecordComparator.java#L104), which is always 0.

The issue is that when Debezium replays the MySQL schema history, it processes every record instead of skipping those that occur after the current start offset. A schema mismatch with the incoming change event data will result in errors such as::
`Error processing after of row in sk_transaction_record because it's different column size with internal schema size 18, but after size 19,`

### Error message/log

```text

```

hzxa21:
Does it mean that his bug exists in debezium upstream as well?

wcy-fdu:
>Does it mean that his bug exists in debezium upstream as well?

Yes

zwang28:
> Does it mean that his bug exists in debezium upstream as well?

[...]

PR #24482, opened 2026-01-15
fix(debezium): fix schema history record comparator

zwang28:

Resolve https://github.com/risingwavelabs/risingwave/issues/24468.
Please refer to https://github.com/risingwavelabs/debezium/pull/6 for changes made to the Debezium source files.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/java/connector-node/risingwave-source-cdc/src/main/java/io/debezium/connector/binlog/BinlogOffsetContext.java
+++ b/java/connector-node/risingwave-source-cdc/src/main/java/io/debezium/connector/binlog/BinlogOffsetContext.java
@@ -1,0 +1,430 @@
+/*
+ * Copyright 2026 RisingWave Labs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


wcy-fdu:
Let me provide some further explanation for this PR (correct me if wrong):
Assume the schema history file contains two records:
- Record A (older):
`{ "server_id": 7, "ts_sec": 10, "file": "mysql-bin.000001", "pos": 100, "event": 0, "row": 0 }`
- Record B (newer)
`{ "server_id": 7, "ts_sec": 20, "file": "mysql-bin.000001", "pos": 200, "event": 0, "row": 0 }`

RisingWave restores an offset and passes Debezium this desired position:
`{ "file": "mysql-bin.000001", "pos": 150, "event": 0, "row": 0 }`

### before the fix
- During schema history recovery, Debezium may reconstruct the “desired” position without server_id and ts_sec.
- The comparator sees recorded.server_id != desired.server_id (7 vs 0), assumes they are from different servers, and falls back to comparing timestamps.
- Due to using the wrong timestamp key (ts_ms instead of ts_sec), both timestamps often read as 0, so 0 <= 0 becomes true.

Result: both Record A (pos=100) and Record B (pos=200) can be treated as “<= desired”, so Debezium may replay too many history records (it can apply the schema at pos=200 even though desired is only pos=150).
### After the fix
- Debezium keeps server_id and ts_sec from the provided offset during recovery (so the reconstructed “desired” position includes them correctly).
- The comparator uses the correct timestamp field (ts_sec) when it needs timestamps.
- Since server_id matches, it compares file/pos/event/row and correctly decides:
  - Record A (100) is <= 150 → replay
  - Record B (200) is > 150 → do not replay

Result: Debezium replays schema history only up to the correct boundary.

PR #24483, opened 2026-01-15
chore: add AGENTS.md to gitignore list

yezizp2012:

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -95,3 +95,6 @@ book/
 
 # Build caches
 .m2/
+
+# Agent instructions
+AGENTS.md


BugenZhao:
I think `AGENTS.md` is by design to be version controlled, just as we have `.cursor/rules` in our repo.

If your motivation is to avoid personalized instructions to be committed to the repo, I guess it should be put under local directory or named as `*.override.md`. Take `codex` as an example:

> https://developers.openai.com/codex/guides/agents-md/#how-codex-discovers-guidance

Ideally we should migrate `.cursor/rules` to `AGENTS.md`, which is more general to different agents. However, it seems that `AGENTS.md` has not yet provided guidelines for modularized rules, so I personally prefer maintaining the current approach, given that agents are actually able to search all of the well-known places for custom instructions.

PR #24484, opened 2026-01-15
fix(optimizer): revert 23180 of dynamic filter watermark derivation

chenzl25:
This reverts commit a46933d12aa8c0f4afbddb3082688ed94b05c895.

- We can't derive watermark from the left input of the dynamic filter because temporal filter will retract stream, so let's revert it. BTW, the optimizer and the executor is unmatched as well, so the EOWC relies on the LHS will wait forever.

## Checklist

## Documentation

<details>

</details>


BugenZhao:
Makes sense. I think we can also propagate watermark from LHS if `condition_always_relax`?

chenzl25:
> Makes sense. I think we can also propagate watermark from LHS if condition_always_relax?

No. It seems if the temporal join could retract data according to the RHS, then we can't propagate watermark from LHS. @yuhao-su found this issue previously.

BugenZhao:
> It seems if the temporal join could retract data according to the RHS,

Do you mean temporal filter? 🤔

Also I mean when it won't retract data, i.e., `StreamKind::AppendOnly if condition_always_relax` here:

https://github.com/risingwavelabs/risingwave/blob/6f0911fd94dcf9010b3b350e457eedae684d9617/src/frontend/src/optimizer/plan_node/stream_dynamic_filter.rs#L43-L56

BugenZhao (approved):

chenzl25:
> Do you mean temporal filter? 🤔 Also I mean when it won't retract data, i.e., StreamKind::AppendOnly if condition_always_relax here:

Yes I mean temporal filter whose condition_always_relax should be false.   You are right, if condition_always_relax = true, we can derive watermark from the left side, but sadly temporal filter with ttl semantic is not the case.

PR #24491, opened 2026-01-16
feat(pg-cdc): support geometry type for postgres cdc

wcy-fdu:

This PR adds PostgreSQL CDC support for PostGIS geometry columns. RisingWave maps upstream geometry to RW bytea and stores the value as raw EWKB bytes, keeping backfill (snapshot) and incremental (Debezium) paths consistent.

### Backfill (snapshot read via tokio-postgres)
- Treat PostgreSQL geometry as raw bytes and decode it directly from pgwire binary format (EWKB).
- Map upstream geometry to RW bytea so snapshot rows can be read and written without special geospatial parsing.
### Incremental (Debezium JSON)
- Debezium encodes PostGIS geometry as an object like {"srid": <int>, "wkb": <base64>}.
- RW parser recognizes this shape (requires both srid and wkb) and base64-decodes wkb into bytes, storing it as bytea.
### Validation (connector service)
- Update PostgreSQL schema/type validation so upstream geometry is considered compatible with RW bytea when creating CDC tables/sources.
### Behavior / Compatibility
- RW does not interpret geometry semantics; it stores exact EWKB bytes as bytea.
- Backfill and incremental results are consistent: the bytes in RW match the EWKB bytes in PostgreSQL.
### Testing
Added a PostGIS geometry CDC test that verifies:
- backfill ingestion
- incremental INSERT/UPDATE/DELETE
- stored bytes match PostgreSQL EWKB (hex comparison)

## Checklist

## Documentation

<details>

</details>


wcy-fdu:
New added geometry test passed:
image
The current CI error is an unrelated issue and can be resolved by retrying later.

src/connector/src/parser/unified/json.rs
```diff
@@ -654,6 +654,33 @@ impl JsonParseOptions {
...
+                    // Validate srid is numeric and wkb is string
+                    if srid.as_i64().is_some() && wkb.as_str().is_some() {
+                        // Extract and decode the wkb field (base64-encoded EWKB)
+                        let wkb_str = wkb.as_str().unwrap();
+                        base64::engine::general_purpose::STANDARD
+                            .decode(wkb_str)
+                            .map_err(|_| create_error())?
+                            .into_boxed_slice()
+                            .into()
+                    } else {
```
xiangjinwu: Extract the actual parsing logic into a dedicated function, for 2 reasons:
* Easier extension for MySQL in the future as we may need to combine srid and wkb to get ewkb. (This [SMT](https://debezium.io/documentation/reference/stable/transformations/geometry-format-transformer.html) may also be useful in the future for unification).
* Keep the caller `match`/`if` focused on dispatching. There may be other `object` that maps to `bytea` but not a `geometry(numeric srid+string wkb)`

e2e_test/source_inline/cdc/postgres/postgres_postgis_geometry.slt
```diff
@@ -0,0 +1,164 @@
...
+# PostGIS geometry type CDC test
+
+system ok
+psql -c "
+    CREATE EXTENSION IF NOT EXISTS postgis;
+    DROP TABLE IF EXISTS test_geo CASCADE;
+    CREATE TABLE test_geo (
+        id int PRIMARY KEY,
+        name varchar,
+        location geometry(Point, 4326)
```
xiangjinwu, line 12: I did not even know we could add `(Point, 4326)` 😂

e2e_test/source_inline/cdc/postgres/postgres_postgis_geometry.slt
```diff
@@ -0,0 +1,164 @@
...
+);
+
+statement ok
+create table test_geo (
+    id int primary key,
+    name varchar,
+    location bytea
+) from geo_source table 'public.test_geo';
+
+sleep 3s
```
xiangjinwu: Prefer `retry 3 backoff 1s` over static sleep - it may succeed sooner.

e2e_test/source_inline/cdc/postgres/postgres_postgis_geometry.slt
```diff
@@ -0,0 +1,164 @@
...
+query IT
+select id, substring(encode(location, 'hex') from 1 for 12) from test_geo where id = 1;
+----
+1 0101000020e6
+
+# Verify SRID value is 4326 (0xe6100000 in little-endian = 4326)
+query IT
+select id, substring(encode(location, 'hex') from 11 for 8) from test_geo where id = 1;
+----
+1 e6100000
```
xiangjinwu: We could just display the whole EWKB from upstream using `psql` or `postgres_query` alongside with the bytes in RisingWave, to confirm they are the same.

e2e_test/source_inline/cdc/postgres/postgres_postgis_geometry.slt
```diff
@@ -0,0 +1,164 @@
...
+
+query II
+select count(*) as total, count(*) filter (where id = 2) as id_2_count from test_geo;
+----
+2 0
+
+# Test with NULL geometry
+system ok
+psql -c "
+    ALTER TABLE test_geo ALTER COLUMN location DROP NOT NULL;
```
xiangjinwu: Was it `not null`?

src/connector/src/parser/unified/json.rs
```diff
@@ -37,6 +37,40 @@ use super::{Access, AccessError, AccessResult};
...
+/// Debezium represents PostGIS `geometry` as an object: `{"srid": <int>, "wkb": <base64_string>}`.
+/// For our current Postgres CDC ingestion, the `wkb` field is expected to be EWKB bytes (base64-encoded),
+/// and `srid` is redundant. We decode `wkb` into raw bytes and store it as `bytea`.
+///
+/// Note: Debezium provides an SMT to convert between WKB and EWKB, which may be useful for future
+/// unification across connectors (e.g., MySQL): see `GeometryFormatTransformer`.
+fn try_parse_debezium_geometry_as_bytea(
+    value: &BorrowedValue<'_>,
+    create_error: impl Fn() -> AccessError,
+) -> AccessResult<Option<Box<[u8]>>> {
```
xiangjinwu, line 60: It is better to document the difference between `Err` and `Ok(None)` clearly.

e2e_test/source_inline/cdc/postgres/postgres_postgis_geometry.slt
```diff
@@ -0,0 +1,149 @@
...
+1 Chicago 25
+2 NewYork 25
+
+# Display and assert the full EWKB hex from upstream and RisingWave are identical.
+system ok
+psql -t -A -c "select encode(location, 'hex') from test_geo where id = 1;"
+----
+0101000020e610000055c1a8a44ee855c00e4faf9465f04440
+
+query T retry 3 backoff 1s
```
xiangjinwu: CI error here is related to a sqllogictest-rs detail: `system ok` requires TWO empty lines at the end when matching full text results.

e2e_test/source_inline/cdc/postgres/postgres_postgis_geometry.slt
```diff
@@ -0,0 +1,149 @@
...
+query ITI retry 3 backoff 1s
+select id, name, length(location) from test_geo where id = 1;
+----
+1 Chicago 25
+
+# Verify the coordinates actually changed (different hex than before)
+query IT retry 3 backoff 1s
+select id, substring(encode(location, 'hex') from 1 for 50) from test_geo where id = 1;
+----
+1 0101000020e610000000000000000000000000000000000000
```
xiangjinwu, line 119: I meant, we could verify the EKWB matches upstream exactly for all these cases, without relying on indirect ways like `length` or `substring`

PR #24480, opened 2026-01-15
fix(stream): fix inconsistent global approx percentile initial state

wenym1:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## What's changed and what's your intention?

In initialization, `approx_percentile` send `None` as its initial output, in the meantime, it initializes its row_count state as `Some(0)`. However, when there is no data before a recovery, after recovery, in the init of `approx_percentile`, when it sees row_count state as `Some(0)`, it treat its last output as `Some(0.0)`, and then when update happens, the old value in the row sent to downstream will be inconsistent.

See error like
```
sanity check fail first=[None, None, None, None] second=[Some(Float64(0.0)), None, Some(Float64(0.0)), Some(Float64(0.0))]
```

To reproduce
```
create table t(p_col double, grp_col int);

create materialized view m1 as
 select
     approx_percentile(0.01, 0.01) within group (order by p_col) as p01,
     approx_percentile(0.1, 0.01) within group (order by p_col) as p10,
     approx_percentile(0.5, 0.01) within group (order by p_col) as p50,
     approx_percentile(0.9, 0.01) within group (order by p_col) as p90,
     approx_percentile(0.99, 0.01) within group (order by p_col) as p99
 from t;

recover;

insert into t values (1.0, 1); FLUSH; // see error
```

## Checklist

- [ ] I have written necessary rustdoc comments.
- [ ]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  I have checked the [Release Timeline](https://github.com/risingwavelabs/rw-commits-history/blob/main/release_timeline.md) and [Currently Supported Versions](https://docs.risingwave.com/changelog/release-support-policy#support-end-dates-for-recent-releases) to determine which release branches I need to cherry-pick this PR into. 

## Documentation

- [ ]  My PR needs documentation updates. 

<details>
<summary><b>Release note</b></summary>

</details>

PR #24500, opened 2026-01-16
fix(stream): fix inconsistent global approx percentile initial state again

wenym1:

The root cause of the issue fixed by https://github.com/risingwavelabs/risingwave/pull/24480 is that we send `None` as the initial output, while in recovery we treat the initial output as `Some(0.0)`. The original solution of https://github.com/risingwavelabs/risingwave/pull/24480 break some assumption of the following `StreamRowMerge`, and the later patch to align with the assumption is problematic.

In this PR, we revert the fix in https://github.com/risingwavelabs/risingwave/pull/24480 and fix it in another way. Instead of sending `None` as the initial output, we send `Some(0.0)` if there is no row in the first barrier. There was error in https://buildkite.com/risingwavelabs/main-cron/builds/8479, and with the fix logic, there is no error anymore in https://buildkite.com/risingwavelabs/main-cron/builds/8484.

See diff between the fix in the main branch. 
https://github.com/risingwavelabs/risingwave/compare/54a308f..18e562f268

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/executor/approx_percentile/global_state.rs
+++ b/src/stream/src/executor/approx_percentile/global_state.rs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 use std::collections::{BTreeMap, Bound};
+use std::mem;
 
 use risingwave_common::array::Op;
 use risingwave_common::bail;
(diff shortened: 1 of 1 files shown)


wenym1:
* **#24500** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24500?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24500?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Issue #23304, opened 2025-09-25
MySQL CDC record missing


zwang28:
### Describe the bug

- The RW MySQL CDC source only sets its offset after processing the first binlog event.
- Debezium obtains start_offset from RW CDC source.
- If offset is NULL, Debezium appears to start from latest binlog position. The relevant code is [here](https://github.com/debezium/debezium/blob/v2.6.2.Final/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlSnapshotChangeEventSource.java#L269), which reads the latest offset from MySQL.

So if any data change is made when RW is offline, and RW starts from latest binlog position, it will miss those data change.

### Error message/log

```text

```

### To Reproduce

1. Create a MySQL table and INSERT row1.
2. Create a MySQL CDC SOURCE1 and downstream TABLE1 from this MySQL table.
3. Verify the SOURCE1 offset is NULL in state table.
4. Verify row1 is in RW TABLE1.
5. Shutdown RisingWave.
6. INSERT row2 in MySQL.
7. Start RisingWave.
8. Verify row2 is **missing** in RW TABLE1.
9. INSERT row3 in MySQL.
10. Verify row3 is missing in RW TABLE1.

### Expected behavior

8. row2 should be fou in RW TABLE1.

### The version of RisingWave

v2.6

zwang28:
SQL server CDC is affected.
PG CDC is not affected.

zwang28:
I thought we could resolve this issue by initializing the start_offset before finalizing the cdc source creation. However, it turned out that the `start_offset` is a debezium json which cannot be obtained from the database directly. e.g. `{"snapshot_done": false, "split_id": 25, "start_offset": "{\"sourcePartition\":{\"server\":\"RW_CDC_25\"},\"sourceOffset\":{\"transaction_id\":\"7657\",\"lsn_proc\":32485736,\"messageType\":\"MESSAGE\",\"lsn_commit\":32485736,\"lsn\":32912192,\"txId\":7657,\"ts_usec\":1761033069901569},\"isHeartbeat\":false}"}`

We need another solution. e.g. add an interface in dbz for querying `start_offset`.

zwang28:
Need consider this corner case #23544

zwang28:
#12868

wcy-fdu:
[...]

PR #24515, opened 2026-01-19
refactor(cdc): refactor cdc source creation logic

zwang28:

Must be merged after https://github.com/risingwavelabs/risingwave/pull/24523.
Second part of resolving https://github.com/risingwavelabs/risingwave/issues/23304.

This PR refactors the CDC source creation logic to ensure that CDC sources are only marked as "created" after they have actually started reading data and updated their offset at least once, rather than immediately upon creation.

### Problem

Previously, CDC sources were marked as "finished" immediately when created, even before they started reading data from the upstream system. This could lead to situations where the source appears ready but hasn't actually begun processing, causing potential issues with downstream dependencies or status reporting.

### Solution

The refactoring introduces a new initialization state for CDC sources and adds a mechanism to track when CDC sources have successfully updated their offset, indicating they've started reading data.

#### 1. Protocol Buffer Updates (`proto/stream_service.proto`)
- Added `CdcSourceOffsetUpdated` message to report when a CDC source has updated its offset at least once
- Includes `reporter_actor_id` and `source_id` fields

#### 2. Progress Tracking (`src/meta/src/barrier/progress.rs`)
- Added new `CdcSourceInit` state to `CreateMviewStatus` enum for CDC sources in initialization
- Modified `new()` method to detect CDC sources and mark them as `CdcSourceInit` instead of `Finished` when there are no actors to track
- Added `mark_cdc_source_finished()` method to transition CDC sources from `CdcSourceInit` to `Finished` state
- Updated progress reporting to show "Initializing CDC source..." for sources in `CdcSourceInit` state

#### 3. Source Executor (`src/stream/src/executor/source/source_executor.rs`)
- Added `cdc_offset_reported` flag to track whether offset update has been reported (reports only once)
- Modified barrier handling to detect when CDC splits have non-empty offsets
- Reports CDC source offset update through `barrier_manager` when first non-empty offset is detected
- Changed `_barrier_manager` to `barrier_manager` to enable reporting

#### 4. Barrier Manager (`src/stream/src/task/barrier_manager/mod.rs`)
- Added `ReportCdcSourceOffsetUpdated` event type
- Added `report_cdc_source_offset_updated()` method to send offset update events

#### 5. Barrier Worker (`src/stream/src/task/barrier_worker/`)
- Added `cdc_source_offset_updated` field to track offset update reports per epoch
- Implemented `report_cdc_source_offset_updated()` method to handle offset update events
- Updated barrier completion logic to include CDC offset update information in responses

#### 6. Meta Node (`src/meta/src/barrier/info.rs`)
- Added handling for `cdc_source_offset_updated` events in barrier responses
- Calls `mark_cdc_source_finished()` on the progress tracker when offset update is received

## Checklist

## Documentation

<details>

</details>


wcy-fdu:
For MySQL/SQL Server CDC, `debezium.heartbeat.interval.ms` should not be allowed to be 0 (or at least enforce a minimum value / emit a warning). Otherwise, a misconfiguration can cause CREATE SOURCE to hang if their is no traffic in upstream.

Shall we handle this(Users are not allowed to disable heartbeat)？

zwang28:
> For MySQL/SQL Server CDC, `debezium.heartbeat.interval.ms` should not be allowed to be 0 (or at least enforce a minimum value / emit a warning). Otherwise, a misconfiguration can cause CREATE SOURCE to hang if their is no traffic in upstream.
> 
> Shall we handle this(Users are not allowed to disable heartbeat)？

Good catch. Resolved.

wcy-fdu (approved):
LGTM!

PR #24527, opened 2026-01-20
fix(frontend): fix incorrect row_number value with non-zero offset

wenym1:

Found and explained by codex
```
 Example MV Showing the Bug

  CREATE TABLE rn_mv_source(x int, y int);

  CREATE MATERIALIZED VIEW mv_row_number_equals_two AS
  SELECT x, y, rank FROM (
      SELECT *, ROW_NUMBER() OVER (PARTITION BY x ORDER BY y) AS rank
      FROM rn_mv_source
  ) WHERE rank = 2;

  INSERT INTO rn_mv_source VALUES (1,1), (1,2), (1,3);
  SELECT * FROM mv_row_number_equals_two;

  What Should Happen

  - ROW_NUMBER() is 1-based, so within each x partition (1,1) → rank 1, (1,2) → rank 2, (1,3) → rank 3.
  - After filtering rank = 2, the output row should be (1, 2, 2).

  What Actually Happened Before the Fix

  1. The optimizer saw ROW_NUMBER() ... WHERE rank = 2 and rewrote it to a TopN with limit = 1 and offset = 1.
  2. That TopN sits below the OverWindow, so the window function is recomputed on the truncated input {(1,2), (1,3)} instead of the original {(1,1),(1,2),(1,3)}.
  3. When the OverWindow runs on only two rows, the new ROW_NUMBER starts at 1 again (because the window operator has no knowledge of the discarded row), so (1,2) now carries rank = 1.
  4. Because the MV still projects rank, the output becomes (1, 2, 1) even though it logically represents the second row of the original partition.

```

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/e2e_test/streaming/row_number_mv.slt
+++ b/e2e_test/streaming/row_number_mv.slt
@@ -21,6 +21,15 @@ SELECT * FROM mv_row_number_equals_two;
 1  2  2
 2  20 2
 
+query III rowsort
+SELECT x, y, rank FROM (
+    SELECT *, ROW_NUMBER() OVER (PARTITION BY x ORDER BY y) AS rank
+    FROM rn_mv_source
+) WHERE rank = 2;
+----
+1  2  2
+2  20 2
+
 statement ok
 DROP MATERIALIZED VIEW mv_row_number_equals_two;
 
--- a/src/frontend/planner_test/tests/testdata/input/over_window_function.yaml
+++ b/src/frontend/planner_test/tests/testdata/input/over_window_function.yaml
@@ -213,6 +213,16 @@
     - logical_plan
     - stream_plan
     - batch_plan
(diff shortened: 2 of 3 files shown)


wenym1:
* **#24527** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24527?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24527?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

BugenZhao:
Interesting findings. 😄 Can you also demonstrate by adding some planner tests (also for batch queries)?

Issue #23802, opened 2025-11-19
Allow subset of struct fields for file source of parquet encoding


xiangjinwu:
If the upstream Parquet contains a column of `struct<foo, bar, baz>` but user only need `struct<foo, bar>` in RisingWave source, the subset of fields shall be properly filled. Currently we set the whole struct column as `null::struct<foo, bar>` due to data type mismatch.

For context, this is already supported for protobuf/avro/json: #18419

May be helpful to start here:
[parquet_parser](https://github.com/risingwavelabs/risingwave/blob/887cc81f18361050f2f77a906c6ee600a83aaaa1/src/connector/src/parser/parquet_parser.rs#L116) calling [`is_parquet_schema_match_source_schema`](https://github.com/risingwavelabs/risingwave/blob/887cc81f18361050f2f77a906c6ee600a83aaaa1/src/common/src/array/arrow/arrow_impl.rs#L1489) and [`array_from_arrow_array`](https://github.com/risingwavelabs/risingwave/blob/887cc81f18361050f2f77a906c6ee600a83aaaa1/src/common/src/array/arrow/arrow_impl.rs#L610)

hzxa21:
Done in https://github.com/risingwavelabs/risingwave/pull/24530

PR #24530, opened 2026-01-20
feat(source): allow subset of struct fields for file source of parquet encoding

wcy-fdu:

As title, background in https://github.com/risingwavelabs/risingwave/issues/23802.
- Enabled Parquet file source to read a subset of fields from an upstream struct by matching struct fields by name and allowing extra Arrow/Parquet fields. 
- Added struct field projection in Arrow→RisingWave conversion plus tests (and a simple S3 e2e script check) to ensure struct<foo,bar,baz> can be read as struct<foo,bar> correctly.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/e2e_test/s3/file_source.py
+++ b/e2e_test/s3/file_source.py
@@ -119,6 +119,112 @@ CREATE TABLE {table_name}(
     except OSError:
         pass
 
+
+def _test_parquet_struct_missing_field_in_file(minio_client: Minio):
+    """Generate a Parquet file whose struct is missing some fields declared in RW schema.
+
+    Upstream Parquet schema:
+      - id: int32
+      - s: struct<foo:int32>
+
+    RisingWave table schema (superset):
+      - id: int
+      - s: struct<foo:int, bar:varchar>
+
+    Expectation:
+      - Rows are still ingested (count(*) == 2).
+      - The mismatched struct field(s) are returned as NULL. Depending on the current
+        behavior, the whole struct may become NULL.
+    """
+
+    # 1) Generate Parquet locally.
+    run_id = str(random.randint(1000, 9999))
+    local_path = f"parquet_struct_missing_field_{run_id}.parquet"
+    s3_key = f"test_parquet_struct_missing_field/{run_id}.parquet"
+
+    id_arr = pa.array([1, 2], type=pa.int32())
+    foo = pa.array([10, 20], type=pa.int32())
+    s_arr = pa.StructArray.from_arrays([foo], names=["foo"])
+
+    schema = pa.schema(
+        [
+            pa.field("id", pa.int32(), nullable=False),
+            pa.field(
+                "s",
+                pa.struct([pa.field("foo", pa.int32(), nullable=True)]),
+                nullable=True,
+            ),
+        ]
+    )
+    table = pa.Table.from_arrays([id_arr, s_arr], schema=schema)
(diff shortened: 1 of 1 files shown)


src/common/src/array/arrow/arrow_impl.rs
```diff
@@ -907,6 +921,57 @@ pub trait FromArrow {
...
+                    let child = array.columns()[idx].clone();
+                    self.from_array(expected_field, &child).map(Arc::new)
+                } else {
+                    // Field missing in Arrow struct. Fill SQL NULL with the expected RW type.
+                    let rw_ty = self.from_field(expected_field)?;
+                    let mut builder = ArrayBuilderImpl::with_type(len, rw_ty);
+                    for _ in 0..len {
+                        let datum: Datum = None;
+                        builder.append(datum);
+                    }
```
xiangjinwu: nit: `append_n`

xiangjinwu (commented):
Looks like udf test failed:
```
query failed: connection closed
[SQL] select * from split('rising wave');
at e2e_test/udf/tests/language_specific/remote/./test.slt.part:228
at e2e_test/udf/tests/language_specific/remote/main.slt:9
```
risedev-logs/frontend-4566.log:
```
thread 'rw-batch-local' (4652) panicked at src/common/src/array/mod.rs:512:29:
type mismatch, array builder type: Utf8, scalar type: Int32
```
e2e_test/udf/tests/language_specific/remote/test.slt.part:
```
create function split(varchar) returns table (word varchar, length int) ...
```

wcy-fdu:
> Looks like udf test failed:
> 
> ```
> query failed: connection closed
> [SQL] select * from split('rising wave');
> at e2e_test/udf/tests/language_specific/remote/./test.slt.part:228
> at e2e_test/udf/tests/language_specific/remote/main.slt:9
> ```
> 
> risedev-logs/frontend-4566.log:
> 
> ```
> thread 'rw-batch-local' (4652) panicked at src/common/src/array/mod.rs:512:29:
> type mismatch, array builder type: Utf8, scalar type: Int32
> ```
> 
> e2e_test/udf/tests/language_specific/remote/test.slt.part:
> 
> ```
> create function split(varchar) returns table (word varchar, length int) ...
> ```

Fixed

wcy-fdu:
@xiangjinwu PTAL❤️

e2e_test/s3/file_source.py
```diff
@@ -12,6 +12,112 @@
...
+            f"[retry {retry_no}] Now got {result[0]} rows in table, 2 expected, wait 5s"
+        )
+        sleep(5)
+
+    # Verify we can read subset fields from struct.
+    # RisingWave struct field access uses the Postgres-compatible syntax: (s).foo
+    cur.execute(f"select id, (s).foo, (s).bar from {table_name} order by id")
+    rows = cur.fetchall()
+    assert rows == [(1, 10, "a"), (2, 20, "b")], f"unexpected rows: {rows}"
+    print("Test parquet struct subset read pass")
```
tabVersion, line 109: can you also cover the case when the desired field not exist in the struct?
like in rw schema `s STRUCT<foo INT, bar VARCHAR>`, but in parquet only have `s STRUCT<foo INT>`

wcy-fdu: Added.

Issue #24506, opened 2026-01-16
unable to read table data for table with ttl retention in MV executor


wenym1:
### Describe the bug

The `get` of `LocalHummockStorage` use `MAX_EPOCH` as the read epoch to represent reading the latest written data. However, when the retention second of the table is set, the `min_epoch` calculated by the `MAX_EPOCH` will be very large, the all data will be filtered out, and the MV executor will fail to read the data via `LocalHummockStorage`.

Solution: avoid using `MAX_EPOCH`, but instead use self.epoch.curr

### Error message/log

```text

```

wenym1:
Need to cover table with ttl in test.

PR #24540, opened 2026-01-20
fix(storage): fix incorrect min_epoch derived from MAX_EPOCH and ttl

wenym1:

resolve https://github.com/risingwavelabs/risingwave/issues/24506

Besides, remove `retention_seconds` from `ReadOptions`, and always use the one of the table itself.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/storage/src/hummock/store/hummock_storage.rs
+++ b/src/storage/src/hummock/store/hummock_storage.rs
@@ -370,6 +370,7 @@ impl HummockStorageReadSnapshot {
             }
             HummockReadEpoch::NoWait(epoch) => {
                 self.build_read_version_tuple_from_all(epoch, self.table_id, key_range)
+                    .await
             }
         }
     }
@@ -427,7 +428,7 @@ impl HummockStorageReadSnapshot {
         ))
     }
 
-    fn build_read_version_tuple_from_all(
+    async fn build_read_version_tuple_from_all(
         &self,
         epoch: u64,
         table_id: TableId,
@@ -440,11 +441,11 @@ impl HummockStorageReadSnapshot {
         let ret = if let Some(info) = info
             && epoch <= info.committed_epoch
         {
-            if epoch < info.committed_epoch {
-                return Err(
-                    HummockError::expired_epoch(table_id, info.committed_epoch, epoch).into(),
-                );
-            }
+            let pinned_version = if epoch < info.committed_epoch {
+                pinned_version
+            } else {
+                self.get_epoch_hummock_version(epoch, table_id).await?
+            };
             // read committed_version directly without build snapshot
             get_committed_read_version_tuple(pinned_version, table_id, key_range, epoch)
         } else {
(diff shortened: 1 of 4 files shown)


wenym1:
* **#24540** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24540?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24540?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

hzxa21:
CI fails due to
```
thread 'rw-streaming' (3885) panicked at src/storage/src/hummock/store/local_hummock_storage.rs:633:9:
assertion `left == right` failed: lushed snapshot reader should not work with table 1280 with ttl
  left: Some(5)
 right: None
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/b925a865e2c9a0aefe5a2877863cb4df796f2eaf/library/std/src/panicking.rs:698:5
```

wenym1:
> CI fails due to
> 
> ```
> thread 'rw-streaming' (3885) panicked at src/storage/src/hummock/store/local_hummock_storage.rs:633:9:
> assertion `left == right` failed: lushed snapshot reader should not work with table 1280 with ttl
>   left: Some(5)
>  right: None
> stack backtrace:
>    0: __rustc::rust_begin_unwind
>              at /rustc/b925a865e2c9a0aefe5a2877863cb4df796f2eaf/library/std/src/panicking.rs:698:5
> ```

Batch query for uncommitted read currently use `MAX_EPOCH`, and will indeed panic here. Maybe we can change to use the frontend current timestamp as the query epoch.

src/storage/src/hummock/store/hummock_storage.rs
```diff
@@ -438,11 +442,11 @@ impl HummockStorageReadSnapshot {
         let ret = if let Some(info) = info
             && epoch <= info.committed_epoch
         {
-            if epoch < info.committed_epoch {
```
hzxa21, line 440: Can you remind me why we disallow reading below MCE before?

src/storage/src/hummock/store/local_hummock_storage.rs
```diff
@@ -384,7 +391,12 @@ impl LocalStateStore for LocalHummockStorage {
     }
 
     fn new_flushed_snapshot_reader(&self) -> Self::FlushedSnapshotReader {
-        self.new_flushed_snapshot_reader_inner()
+        assert_eq!(
+            self.table_option.retention_seconds, None,
+            "lushed snapshot reader should not work with table {} with ttl",
```
hzxa21: `lushed` -> `flushed`

src/storage/src/hummock/store/hummock_storage.rs
wenym1: No specific reason. Previous no error was caused by this limit because we always use u64::MAX as the query epoch for uncommitted read, and for uncommitted read the reader block the barrier of the epoch used to read, and therefore won't hit this error either.

PR #24551, opened 2026-01-21
feat: introduce force resume backfill node/job interface in risectl

### LINKED ISSUE #24424: Support manual resume for a backfill node
Support a barrier command to force backfill to resume. This can be done by:
- job (all fragments start backfilling).
- fragment.

This is meant for troubleshooting purposes, in case some node is stuck.

yezizp2012:

Resolves #24424 .

Usage: 
```
risectl meta resume-backfill --job-id <job_id> / risectl meta resume-backfill --fragment-id <fragment_id>
```

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/meta/src/barrier/command.rs
+++ b/src/meta/src/barrier/command.rs
@@ -407,7 +407,7 @@ pub enum Command {
     },
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Copy)]
 pub enum ResumeBackfillTarget {
     Job(JobId),
     Fragment(FragmentId),
@@ -676,6 +676,9 @@ pub enum PostCollectCommand {
         subscription_id: SubscriptionId,
     },
     ConnectorPropsChange(ConnectorPropsChange),
+    ResumeBackfill {
+        target: ResumeBackfillTarget,
+    },
 }
 
 impl PostCollectCommand {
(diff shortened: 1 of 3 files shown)


src/ctl/src/cmd_impl/meta/resume_backfill.rs
```diff
@@ -0,0 +1,44 @@
...
+) -> Result<()> {
+    let target = match (job_id, fragment_id) {
+        (Some(job_id), None) => Target::JobId(job_id),
+        (None, Some(fragment_id)) => Target::FragmentId(fragment_id),
+        _ => {
+            return Err(anyhow!(
+                "exactly one of job_id or fragment_id must be provided"
+            ));
+        }
+    };
```
kwannoel, line 34: Why don't we use an enum here instead?

src/meta/service/src/ddl_service.rs
```diff
@@ -776,6 +776,62 @@ impl DdlService for DdlServiceImpl {
         Ok(Response::new(RisectlListStateTablesResponse { tables }))
     }
 
+    async fn risectl_resume_backfill(
```
kwannoel, line 779: Shall we also update the backfill orders metadata here?

src/ctl/src/cmd_impl/meta/resume_backfill.rs
yezizp2012: Just refer to and keep consistent with other risectl commands. 
Due to the limitations of the parameters defined by `clap`, adding it would only allow modifying the function's parameters and moving this code to the caller side, which doesn't seem very necessary.

src/meta/service/src/ddl_service.rs
yezizp2012: Added the update logic in post collect.

PR #24555, opened 2026-01-21
feat(storage): refine imm memory limiter management

wenym1:

#### Life cycle of an imm:
1. created in an instance of `LocalHummockStorage`, associated to a `table_id`, added to the `ReadVersion` of `LocalHummockStorage`. When handling read on `LocalHummockStorage`, its reference can be cloned and held by the reader (get, iter and rev_iter)
2. generate `HummockEvent::ImmToUploader` event, and add to event channel
3. event handler handles the event, and add the imm to uploader
4. on uploader flush or sync epoch, do shared buffer compact to transform its data into SST, replace the reference in `ReadVersion` with the newly generated SST.
5. besides above, there can be some `orphan imm`. It can be 1) an imm whose data has already been compacted but not released because it's pinned by a reader, or 2) imm of replicated local state store, which is only cleaned after the epoch is committed.

#### semantic of metrics
* `state_store_per_table_imm_size`: labeled by `table_id`, including all imms in phase 1-5
* `uploading_memory_size`: include imms in phase 1-4, which means all imms that are going to be uploaded as SST. Orphan imm size (in phase 5) is `sum(state_store_per_table_imm_size) over table_id - uploading_memory_size`
* `state_store_uploader_imm_size`: imms in phase 3 and 4, which means the imms managed by the uploader
* `state_store_uploader_uploading_task_size`: imms in phase 4

#### new flush strategy
```
if need_flush() {
      let mut spiller = Spiller::new(&mut data.unsync_data);
      while need_more_flush(curr_batch_flush_size)
          && let Some((epoch, payload, spilled_table_ids)) = spiller.next_spilled_payload()
      {
          // spawn upload task
      }
}
```
`need_flush`: total size of imms in phase 3 is greater than `flush_threshold`, which is by default 80% of shared buffer capacity. Total size of imms in phase 3 is calculated as `state_store_uploader_imm_size - state_store_uploader_uploading_task_size`
`need_more_flush`: `need_flush` or: for the created spiller, the total size of newly spill task has not reached a configured minimum value, which is by default is `800MB`.

Under the new flush strategy, we only control flush based on the imms managed by the uploader, and it's more controllable.

## Checklist

## Documentation

<details>

</details>


wenym1:
* **#24561** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24561?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24555** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24555?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24555?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#24538** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24538?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/storage/benches/bench_imm_compact.rs
```diff
@@ -52,7 +52,7 @@ fn criterion_benchmark(c: &mut Criterion) {
         &batches,
         |b, batches| {
             b.to_async(FuturesExecutor).iter(|| async {
-                let imm = merge_imms_in_memory(TableId::default(), batches.clone(), None).await;
+                let imm = merge_imms_in_memory(TableId::default(), batches.clone()).await;
```
Li0k, line 55: nits: I don't think **merge_imms_in_memory** is being used anymore , can we remove it in a separate PR?

src/storage/src/hummock/compactor/shared_buffer_compact.rs
```diff
@@ -327,7 +326,6 @@ async fn compact_shared_buffer<const IS_NEW_VALUE: bool>(
 pub async fn merge_imms_in_memory(
```
Li0k, line 326: ditto

src/storage/src/hummock/event_handler/hummock_event_handler.rs
```diff
@@ -1248,16 +1222,30 @@ mod tests {
...
+            move |read_version: &HummockReadVersionRef,
+                  instance: &LocalInstanceGuard,
+                  imm: &ImmutableMemtable| {
+                let memory_limiter = memory_limiter.clone();
+                let event_tx = event_tx_clone.clone();
+                let read_version = read_version.clone();
+                let imm = imm.clone();
+                let instance_id = instance.instance_id;
+                async move {
+                    let tracker = memory_limiter.require_memory(imm.size() as _).await;
```
wenym1, line 1237: This is just test, so it should be fine. (modified by AI though)

wenym1:
The logic has been verified to work in prod cluster.

PR #24588, opened 2026-01-26
fix: remove non-deterministic background enabling for ddl tests

### LINKED ISSUE #24586: test failure in madsim e2e recovery test with background ddl
### Buildkite URL

https://buildkite.com/risingwavelabs/main-cron/builds/8563#019bf79a-907f-4423-b371-70f888c8d3ca

### Description and Insights

```
thread '<unnamed>' (248) panicked at /risingwave/src/tests/simulation/src/slt.rs:439:21:
query result mismatch:
[SQL] select background_ddl from rw_catalog.rw_materialized_views where name='m1';
[Diff] (-expected|+actual)
-   f
+   t
at e2e_test/ddl/drop/drop_creating_mv.slt:97
```

cc @kwannoel

yezizp2012:

Closes #24586 . As title, fixes flaky main-cron tests while background ddl is enabled randomly. I think it's not necessary for all the tests under ddl.

Root cause: `--background-ddl-rate` makes m1 background randomly.

https://github.com/risingwavelabs/risingwave/blob/e22462d848d6b04ecab99e6af67016a8bbc38d0f/e2e_test/ddl/drop/drop_creating_mv.slt#L94-L100

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/ci/scripts/deterministic-recovery-test.sh
+++ b/ci/scripts/deterministic-recovery-test.sh
@@ -69,7 +69,6 @@ echo "--- deterministic simulation e2e, ci-3cn-2fe-1meta, recovery, ddl"
 seq "$TEST_NUM" | parallel --tmpdir .risingwave --ungroup 'set -o pipefail && ((MADSIM_TEST_SEED={} $RW_SIM \
 --kill \
 --kill-rate=${KILL_RATE} \
---background-ddl-rate=${BACKGROUND_DDL_RATE} \
 ${EXTRA_ARGS:-} \
 ./e2e_test/ddl/\*\*/\*.slt 2> $LOGDIR/recovery-ddl-{}.log && rm $LOGDIR/recovery-ddl-{}.log) \
 | awk -W interactive "{print \"(seed = {}): \" \$0; fflush()}")'


yezizp2012:
> Interesting. So why haven't we encountered this issue before? Is it purely coincidental?

The issue was exposed after fixing the background ddl status in #24525. Before that the background status in system catalog will always be `false` under some conditions.

wenym1:
```
query I 
 select background_ddl from rw_catalog.rw_materialized_views where name='m1'; 
 ---- 
 f 
```

I think the failed query just wants to check that the job with the specific name exists. Shall we make it `select count(*)` and keep the background ddl rate in the test? I think it can help find out bug under corner cases.

yezizp2012:
> ```
> query I 
>  select background_ddl from rw_catalog.rw_materialized_views where name='m1'; 
>  ---- 
>  f 
> ```
> 
> I think the failed query just wants to check that the job with the specific name exists. Shall we make it `select count(*)` and keep the background ddl rate in the test? I think it can help find out bug under corner cases.

Emmm, In `e2e_test/ddl/drop/drop_creating_mv.slt`, there are many places that actively set background ddl, and the `--background-ddl-rate` will make the state more non-deterministic. `--background-ddl-rate` was introduced when the background ddl function was not stable in the early days. Now there have been many tests on background ddl. I even think `--background-ddl-rate` can be deprecated in the future, At least for now I don't think it's necessary for the ddl tests.

wenym1:
> > ```
> > query I 
> >  select background_ddl from rw_catalog.rw_materialized_views where name='m1'; 
> >  ---- 
> >  f 
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I think the failed query just wants to check that the job with the specific name exists. Shall we make it `select count(*)` and keep the background ddl rate in the test? I think it can help find out bug under corner cases.
> 
> Emmm, In `e2e_test/ddl/drop/drop_creating_mv.slt`, there are many places that actively set background ddl, and the `--background-ddl-rate` will make the state more non-deterministic. `--background-ddl-rate` was introduced when the background ddl function was not stable in the early days. Now there have been many tests on background ddl. I even think `--background-ddl-rate` can be deprecated in the future, At least for now I don't think it's necessary for the ddl tests.

I still prefer to keep the existing tests. These tests helped find out several snapshot backfill related bugs. 

If the current issue is only by showing the expected `background_ddl` field, we can simply mark the query as  `skipif madsim`.

yezizp2012:
> > > ```
> > > query I 
> > >  select background_ddl from rw_catalog.rw_materialized_views where name='m1'; 
> > >  ---- 
> > >  f 
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > I think the failed query just wants to check that the job with the specific name exists. Shall we make it `select count(*)` and keep the background ddl rate in the test? I think it can help find out bug under corner cases.
> > 
> > 
> > Emmm, In `e2e_test/ddl/drop/drop_creating_mv.slt`, there are many places that actively set background ddl, and the `--background-ddl-rate` will make the state more non-deterministic. `--background-ddl-rate` was introduced when the background ddl function was not stable in the early days. Now there have been many tests on background ddl. I even think `--background-ddl-rate` can be deprecated in the future, At least for now I don't think it's necessary for the ddl tests.
> 
> I still prefer to keep the existing tests. These tests helped find out several snapshot backfill related bugs.
> 
> If the current issue is only by showing the expected `background_ddl` field, we can simply mark the query as `skipif madsim`.

How about disabling `--background-ddl-rate` for ddl tests, as what's been done in this PR? Most of these tests relate to the system catalog and schema. We can keep this config for other tests.

wenym1:
> > > > ```
> > > > query I 
> > > >  select background_ddl from rw_catalog.rw_materialized_views where name='m1'; 
> > > >  ---- 
> > > >  f 
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > I think the failed query just wants to check that the job with the specific name exists. Shall we make it `select count(*)` and keep the background ddl rate in the test? I think it can help find out bug under corner cases.
> > > 
> > > 
> > > Emmm, In `e2e_test/ddl/drop/drop_creating_mv.slt`, there are many places that actively set background ddl, and the `--background-ddl-rate` will make the state more non-deterministic. `--background-ddl-rate` was introduced when the background ddl function was not stable in the early days. Now there have been many tests on background ddl. I even think `--background-ddl-rate` can be deprecated in the future, At least for now I don't think it's necessary for the ddl tests.
> > 
> > 
> > I still prefer to keep the existing tests. These tests helped find out several snapshot backfill related bugs.
> > If the current issue is only by showing the expected `background_ddl` field, we can simply mark the query as `skipif madsim`.
> 
> How about disabling `--background-ddl-rate` for ddl tests, as what's been done in this PR? Most of these tests relate to the system catalog and schema. We can keep this config for other tests.

LGTM

PR #24595, opened 2026-01-26
chore: more robust postgres geometry type test

wcy-fdu:

increase retry times...

## Checklist

## Documentation

<details>

</details>


e2e_test/source_inline/cdc/postgres/postgres_postgis_geometry.slt
```diff
@@ -73,13 +73,15 @@ order by ours.id;
 system ok
 psql -c "INSERT INTO test_geo VALUES (3, 'London', ST_SetSRID(ST_MakePoint(-0.1278, 51.5074), 4326));"
 
-query ITI retry 3 backoff 1s
+sleep 5s
```
hzxa21: I think retry + backoff is strictly better than explicitly sleep. Let's avoid doing the sleep to reduce the CI time.

hzxa21 (approved):

Issue #24583, opened 2026-01-25
snapshot backfill fails with ObjectNotFound error


hzxa21:
### Describe the bug

This happened when backfilling from a 10+TB upstream table. I saw full GC triggered 10mins before the snapshot backfill failure. I suspect it is related to the fact that full GC doesn't take version pinned back snapshot backfill into consideration. Note that the default min sst retention time for full gc is 6hr.

Part of the backtrace
```
[...]

PR #24602, opened 2026-01-27
feat(meta): pause time travel vacuum when having snapshot backfilling jobs

wenym1:

Snapshot backfill may keep querying a snapshot epoch long time ago until backfill finishes. We should rely on time travel query instead of pinned version. Therefore, in this PR, every time before we do time travel vacuum, we will check whether there is snapshot backfilling job, and pause the time travel vacuum when there is any.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/meta/src/hummock/manager/time_travel.rs
+++ b/src/meta/src/hummock/manager/time_travel.rs
@@ -658,8 +658,11 @@ impl HummockManager {
             self.env.opts.hummock_time_travel_sst_info_insert_batch_size,
         )
         .await?;
-        // Ignore delta which adds no data.
-        if written > 0 {
+        let has_state_table_info_delta = delta
+            .state_table_info_delta
+            .keys()
+            .any(|table_id| time_travel_table_ids.contains(table_id));
+        if written > 0 || has_state_table_info_delta {
             let m = hummock_time_travel_delta::ActiveModel {
                 version_id: Set(risingwave_meta_model::HummockVersionId::try_from(
                     delta.id.to_u64(),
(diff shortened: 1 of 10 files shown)


wenym1:
* **#24602** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24602?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24602?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/meta/src/hummock/mod.rs
```diff
@@ -109,6 +113,10 @@ pub fn start_vacuum_time_travel_metadata_loop(
                     return;
                 }
             }
+            if should_pause_vacuum_time_travel().await {
```
Li0k, line 116: cc @hzxa21 @zwang28 

If I remember correctly, time-travel causes a surge in data volume, this was also the reason we controlled the retention period for time-travel. If snapshot-backfill runs for a long time, the increase in data volume will also be significant.

hzxa21:
Related: https://github.com/risingwavelabs/risingwave/issues/24583

src/meta/src/hummock/manager/time_travel.rs
```diff
@@ -658,8 +658,11 @@ impl HummockManager {
             self.env.opts.hummock_time_travel_sst_info_insert_batch_size,
         )
         .await?;
-        // Ignore delta which adds no data.
-        if written > 0 {
+        let has_state_table_info_delta = delta
```
zwang28, line 661: time_travel_table_ids never contains internal tables, refer to `list_time_travel_table_ids`.

PR #24611, opened 2026-01-27
feat(optimizer): support mv selection for batch queries

chenzl25:

- Support very basic mv selection at logical plan level.
- New session variable `enable_mv_selection` by default is false.
- If enable the mv selection, we first register mv before the optimizer and then use MvSelectionRule to match against the current plan of mv candidates. Current implementation is naive but work and easy to understand. We can make it more efficient later.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/e2e_test/batch/catalog/pg_settings.slt.part
+++ b/e2e_test/batch/catalog/pg_settings.slt.part
@@ -43,8 +43,8 @@ user enable_datafusion_engine
 user enable_index_selection
 user enable_join_ordering
 user enable_locality_backfill
-user enable_serverless_backfill
 user enable_mv_selection
+user enable_serverless_backfill
 user enable_share_plan
 user enable_two_phase_agg
 user extra_float_digits


src/frontend/src/optimizer/logical_optimization.rs
```diff
@@ -941,4 +957,41 @@ impl LogicalOptimizer {
...
+        let Some(current_user) = user_reader.get_user_by_name(&session.user_name()) else {
+            return;
+        };
+        let db_name = session.database();
+        let Ok(schemas) = catalog_reader.iter_schemas(&db_name) else {
+            return;
+        };
+
+        for schema in schemas {
+            for mv in schema.iter_created_mvs_with_acl(current_user) {
```
BugenZhao, line 973: Potential future improvements: we may narrow the scope of candidates to those who have the same dependency as the currently planning MV.

src/frontend/src/optimizer/logical_optimization.rs
```diff
@@ -941,4 +957,41 @@ impl LogicalOptimizer {
...
+                    continue;
+                };
+                let mut binder = Binder::new_for_batch(session);
+                let Ok(bound_query) = binder.bind_query(&query) else {
+                    continue;
+                };
+                let mut planner = Planner::new_for_batch_dql(context.clone());
+                let Ok(plan_root) = planner.plan_query(bound_query) else {
+                    continue;
+                };
```
BugenZhao, line 992: Since we're replanning using the original definition here, I'm thinking what are the cases when same SQL may lead to different semantics or results.

Basically it could happen when the query involves side effects. For example,
- referencing external sources with retention
- containing impure expressions, like time-related functions or UDF
- encapsulating session configs to be used in execution, like timezone

src/frontend/src/optimizer/logical_optimization.rs
chenzl25: Yes. That's exactly what I am thinking as well. Some pruning optimization could be applied here.

src/frontend/src/optimizer/logical_optimization.rs
chenzl25: > referencing external sources with retention

This one is easy to solve, we can ignore any materialized views that have sources.

> containing impure expressions, like time-related functions or UDF

I think we can use the mv to replace the batch queries directly for impure expressions, because it makes the batch query results more aligned with the streaming query results. 

> encapsulating session configs to be used in execution, like timezone

This one seems hard to detect but it seems we directly use the mv would be ok in most cases, because most users would use only one timezone in their system. We can track it with a future improvement issue.

src/frontend/src/optimizer/plan_node/generic/project.rs
```diff
@@ -52,11 +53,14 @@ fn check_expr_type(expr: &ExprImpl) -> std::result::Result<(), &'static str> {
...
 /// [`Project`] computes a set of expressions from its input relation.
-#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[derive(Debug, Clone, Educe)]
+#[educe(PartialEq, Eq, Hash)]
 #[allow(clippy::manual_non_exhaustive)]
 pub struct Project<PlanRef> {
     pub exprs: Vec<ExprImpl>,
     /// Mapping from expr index to field name. May not contain all exprs.
+    #[educe(PartialEq(ignore))]
+    #[educe(Hash(ignore))]
```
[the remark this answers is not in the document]
chenzl25: Ignore the field name is exactly the thing we want to achieve when we compare the plan, because we use index of `InputRef` to know exactly which column it refers to. The `Eq` trait of the Plan and Expr right now are all used by the same feature i.e. compare the plan to detect it if could be shared or reused.

chenzl25:
After https://github.com/risingwavelabs/risingwave/pull/24706 we can prune mv candidates.

chenzl25:
* **#24782** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24782?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24751** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24751?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24611** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24611?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24611?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #24614, opened 2026-01-27
fix: limit meta refill concurrency

yezizp2012:

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/storage/src/hummock/event_handler/refiller.rs
+++ b/src/storage/src/hummock/event_handler/refiller.rs
@@ -380,6 +380,7 @@ impl CacheRefillTask {
             .map(|info| async {
                 let mut stats = StoreLocalStatistic::default();
                 GLOBAL_CACHE_REFILL_METRICS.meta_refill_attempts_total.inc();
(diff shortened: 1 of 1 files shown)


src/storage/src/hummock/event_handler/refiller.rs
```diff
@@ -380,6 +380,7 @@ impl CacheRefillTask {
             .map(|info| async {
                 let mut stats = StoreLocalStatistic::default();
                 GLOBAL_CACHE_REFILL_METRICS.meta_refill_attempts_total.inc();
+                let _permit = context.concurrency.acquire().await.unwrap();
```
Li0k, line 383: How about splitting the limit params for meta cache and data cache refill?

yezizp2012: @MrCroxx WDUT?

MrCroxx: +1 for spliting. Let me do it.

yezizp2012:
@MrCroxx will open another PR to split the limit params for meta cache and data cache refill.

PR #24620, opened 2026-01-28
feat(observability): add fragment-level immutable memtable memory metrics

EdwinaZhu:

This pull request introduces support for tracking shared buffer memory usage metrics by fragment in addition to by table, and propagates the `fragment_id` through various storage layers and test utilities. Besides, it aggregates shared buffer memory usage metrics by job. Now we can check the sum of shared buffer memory usage and executor cache by job on grafana. The main changes include updating metric collection, modifying options structures to include `fragment_id`, and updating tests and dashboard panels to reflect these changes.

**Metrics and Monitoring Enhancements**
- Change the label of `per_table_imm_size` from `table_id` to `[table_id, fragment_id]`. When the metric level is set into MetricLevel::Info, the label would be relabeled, only keeping `fragment_id`. The metrics are now incremented and decremented alongside the existing per-table metrics.
- Updated Grafana dashboard panels to visualize memory usage by fragment and by relation, including new timeseries for shared buffer memory usage (`grafana/dashboard/streaming_relations.py` and `streaming_fragments.py`).

**API and Test Updates**
- Modified `NewLocalOptions` and related traced/test options to include a `fragment_id` field, ensuring that fragment context is available throughout storage operations and tracing. 
- Added `TracedFragmentId` struct and conversion implementations between `FragmentId` and `TracedFragmentId` for tracing and replay purposes.
- Updated tests and test utilities across storage modules to provide and propagate `fragment_id` where required, ensuring correctness under the new API.

## Checklist

## Documentation

<details>

Now users could monitor memory usage more precisely by fragment/job on grafana, which includes not only the memory usage of executor cache, but the shared buffer memory usage by fragment/job.
</details>

### DIFF
--- a/src/storage/hummock_trace/src/opts.rs
+++ b/src/storage/hummock_trace/src/opts.rs
@@ -189,12 +189,12 @@ pub struct TracedTryWaitEpochOptions {
 
 #[cfg(test)]
 impl TracedNewLocalOptions {
-    pub(crate) fn for_test(table_id: u32, fragment_id: u32) -> Self {
+    pub(crate) fn for_test(table_id: u32) -> Self {
         use risingwave_common::hash::VirtualNode;
 
         Self {
             table_id: TracedTableId { table_id },
-            fragment_id: TracedFragmentId { fragment_id },
+            fragment_id: TracedFragmentId { fragment_id: 0 },
             op_consistency_level: TracedOpConsistencyLevel::Inconsistent,
             table_option: TracedTableOption {
                 retention_seconds: None,
--- a/src/storage/hummock_trace/src/replay/runner.rs
+++ b/src/storage/hummock_trace/src/replay/runner.rs
@@ -92,9 +92,9 @@ mod tests {
         let get_result = TracedBytes::from(vec![54, 32, 198, 236, 24]);
         let sync_id = 4561245432;
 
-        let opts1 = TracedNewLocalOptions::for_test(1, 1);
-        let opts2 = TracedNewLocalOptions::for_test(2, 1);
-        let opts3 = TracedNewLocalOptions::for_test(3, 1);
+        let opts1 = TracedNewLocalOptions::for_test(1);
+        let opts2 = TracedNewLocalOptions::for_test(2);
+        let opts3 = TracedNewLocalOptions::for_test(3);
 
         let storage_type1 = StorageType::Local(0, 1);
         let storage_type2 = StorageType::Local(1, 2);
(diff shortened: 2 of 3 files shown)


EdwinaZhu:
The output screenshot is shown below:
image
image

grafana/dashboard/dev/streaming_fragments.py
```diff
@@ -270,6 +270,17 @@ def _(outer_panels: Panels):
...
+                panels.timeseries_bytes(
+                    "Shared Buffer Memory Usage",
+                    "",
+                    [
+                        panels.target(
+                            f"sum({metric('state_store_per_fragment_imm_size')}) by (fragment_id)",
+                            "fragment {{fragment_id}}",
+                        ),
+                    ],
+                ),
```
kwannoel, line 329: We should also add Total Memory Usage (Executor Cache + Shared Buffer usage) here.

src/storage/src/monitor/hummock_state_store_metrics.rs
```diff
@@ -465,6 +467,34 @@ impl HummockStateStoreMetrics {
             metric_level,
         );
 
+        let per_fragment_imm_size = register_guarded_int_gauge_vec_with_registry!(
```
wenym1: We can reuse the metrics per table_id. Currently the metrics are labeled with `[table_id]`, and the `table_id` will be relabeled to `""` by the RelabeledMetrics when the metrics level is not high enough. We can change to be labeled with `[fragment_id, table_id]`, and when metrics level is not high enough, we only relabel the `table_id`. We can do this by setting the prefix length of not-relabel to be 1

grafana/dashboard/dev/streaming_fragments.py
EdwinaZhu: image

EdwinaZhu:
* **#24688** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24688?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24620** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24620?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24620?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/storage/hummock_trace/src/opts.rs
```diff
@@ -170,11 +189,12 @@ pub struct TracedTryWaitEpochOptions {
 
 #[cfg(test)]
 impl TracedNewLocalOptions {
-    pub(crate) fn for_test(table_id: u32) -> Self {
+    pub(crate) fn for_test(table_id: u32, fragment_id: u32) -> Self {
```
wenym1: Can use 0 as fragment_id in for test

EdwinaZhu: @copilot change all fragment_id in test as 0.

PR #24632, opened 2026-01-28
fix(metrics): fix sink commit metrics time unit

wcy-fdu:

In grafana panel, the unit of the commit duration is second, while we use `as_millis` in our code, lead us to see the commit time multiplied by 1000.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/e2e_test/source_inline/fs/utf8_err_tmp/data_bad.csv
+++ b/e2e_test/source_inline/fs/utf8_err_tmp/data_bad.csv
@@ -1,3 +1,0 @@
-carat,cut,color,depth
-0.22,Premium,I,62
-0.23,Very Good,H,57.5


src/connector/src/sink/coordinate.rs
```diff
@@ -230,7 +230,7 @@ impl<W: SinkWriter<CommitMetadata = Option<SinkMetadata>>> LogSinker for Coordin
                                 .await?;
                             sink_writer_metrics
                                 .sink_commit_duration
-                                .observe(start_time.elapsed().as_millis() as f64);
```
wenym1, line 233: I think we can should instead divide 1000 when displaying in grafana dashboard. Our barrier latency is 1s. Reporting in second granularity can lose many information.

wcy-fdu: Umm I don't think so:
1. bucket of histgrom is default second boundary, we can not first divide 1000 and then put it in bucket. The "bucketing" occurs at the moment `observe()` is called within the Rust process, not within Grafana/PromQL.
2. using `as_secs_f64()` does not limit precision to "second-level granularity".
`as_secs_f64()` is for floating-point seconds; sub-second information such as 0.123s is fully preserved; there is no issue of "losing a lot of information when reporting seconds".

wcy-fdu: If we want to modify Grafana, we need to change this bucket to use milliseconds as the unit, which contradicts the second-level barrier latency. We should still use second-level buckets here.

wenym1: Oh, I see. It's `as_secs_f64` rather than `as_secs() as f64`

PR #24641, opened 2026-01-29
perf: avoid fetching stream nodes in system catalogs

yezizp2012:

As title. Previously `rw_fragments` was used by cloud and some other system views, which fetches the node info for all fragments. This has a high performance pressure on both the frontend and metastore.
1. removed `node` column in `rw_fragments`.
2. introduced new system catalog `rw_fragment_infos` that includes `node` column, same as `rw_fragments`. It's not in use, in case the node info is needed in some scenario in the future.
3. rewrote `rw_sink_decouple` together with a new system catalog `rw_sink_log_store_tables`, makes it not dependent on fragment node information.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/meta/src/controller/fragment.rs
+++ b/src/meta/src/controller/fragment.rs
@@ -1142,6 +1142,7 @@ impl CatalogController {
         let txn = inner.db.begin().await?;
 
         let fragments_query = Self::build_fragment_query(is_creating);
+        #[allow(clippy::type_complexity)]
         let fragments: Vec<(
             FragmentId,
             JobId,


yezizp2012:
🙏  Cc @Gogomoe please help to confirm that the `node` column of `rw_fragments` is not used in any cloud scenario. I have checked in cloud's repo and it seems safe to remove it: https://github.com/search?q=repo%3Arisingwavelabs%2Frisingwave-cloud%20rw_fragments&type=code

kwannoel (approved):
LGTM!

Gogomoe:
> 🙏 Cc @Gogomoe please help to confirm that the node column of rw_fragments is not used in any cloud scenario

Yes, confirmed the `node` column is unused

PR #24647, opened 2026-01-30
feat(google-pubsub): add Application Default Credentials (ADC) support to Google Pub/Sub connector

### LINKED ISSUE #24646: Add Application Default Credentials (ADC) support to Google Pub/Sub connector

Currently, the Google Cloud Pub/Sub connector requires users to explicitly provide the `pubsub.credentials` parameter with a **JSON string containing service account credentials**. This creates several issues:

- **Security concerns**: Credentials must be embedded in connector configurations or SQL statements, potentially exposing sensitive information in logs, version control, or query history.
- **Poor developer experience**: Developers need to manually copy and paste credential JSON strings into configurations, which is error-prone and cumbersome.
-  **Limited Cloud Native support**: When RisingWave is deployed on Google Cloud infrastructure (GKE, GCE, Cloud Run), it cannot automatically leverage the underlying platform's identity and access management.
- **Configuration overhead**: Every source/sink configuration needs explicit credentials even when running in an environment where Google ADC is already configured.

Implement **Google Application Default Credentials (ADC)** support for the Pub/Sub connector. ADC is a Google-standard strategy for authentication that automatically finds credentials based on the environment.

When pubsub.credentials is not explicitly provided, the connector should attempt to authenticate using ADC in the following order:

- `GOOGLE_APPLICATION_CREDENTIALS_JSON` environment variable (JSON content)
- `GOOGLE_APPLICATION_CREDENTIALS` environment variable (file path)
- Well-known file: $HOME/.config/gcloud/application_default_credentials.json
- Google Cloud metadata server (when running on GCE/GKE)

_No response_

### Additional context

_No response_

tuantran0910:

This PR adds support for **Google Application Default Credentials (ADC)** to the Google Cloud Pub/Sub connector.

Previously, users were required to explicitly provide the `pubsub.credentials` parameter with a JSON string containing service account credentials. Now, the connector can automatically use Google's ADC when no explicit credentials are provided, improving the developer experience and simplifying configuration.

Closes #24646 

## Usage

When `pubsub.credentials` is not specified, the connector searches for credentials in the following order:
1. `GOOGLE_APPLICATION_CREDENTIALS_JSON` environment variable (JSON content)
2. `GOOGLE_APPLICATION_CREDENTIALS` environment variable (file path)
3. Well-known file: $HOME/.config/gcloud/application_default_credentials.json
4. Google Cloud metadata server (when running on GCE/GKE)

## How to test

First, create a Google Cloud Pubsub topic:

```
gcloud pubsub topics create risingwave

gcloud pubsub subscriptions create risingwave-sub \
  --topic=risingwave
```

Second, create Risingwave table with connector `google_pubsub`:

```sql
CREATE TABLE pubsub_source_v3 (
  text STRING
) WITH (
  connector = 'google_pubsub',
  pubsub.subscription = 'projects/<PROJECT_NAME>/subscriptions/risingwave-sub'
) FORMAT PLAIN ENCODE JSON;
```

Third, publish message into the topic:

```
gcloud pubsub topics publish risingwave \
  --message='{"text": "Hello from Pub/Sub!"}'

gcloud pubsub topics publish risingwave \
  --message='{"text": "Hello from Pub/Sub! I love RisingWave!"}'
```

Finally, check the result:

```sql
SELECT * FROM pubsub_source_v3;
                   text                    
-------------------------------------------
 Hello from Pub/Sub!
 Hello from Pub/Sub! I love RisingWave!
(2 rows)
```

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/with_options_sink.yaml
+++ b/src/connector/with_options_sink.yaml
@@ -433,15 +433,6 @@ GooglePubSubConfig:
       see the [service-account](https://developers.google.com/workspace/guides/create-credentials#create_credentials_for_a_service_account) credentials guide.
       The provided account credential must have the
       `pubsub.publisher` [role](https://cloud.google.com/pubsub/docs/access-control#roles)
-
-      If not specified, the connector will attempt to use Google Application Default Credentials (ADC).
-      ADC searches for credentials in the following order:
-      1. `GOOGLE_APPLICATION_CREDENTIALS_JSON` environment variable (JSON content)
-      2. `GOOGLE_APPLICATION_CREDENTIALS` environment variable (file path)
-      3. Well-known file: `$HOME/.config/gcloud/application_default_credentials.json`
-      4. Google Cloud metadata server (when running on GCE/GKE)
-
-      See [Google ADC documentation](https://cloud.google.com/docs/authentication/application-default-credentials) for more details.
     required: false
 IcebergConfig:
   fields:
--- a/src/connector/with_options_source.yaml
+++ b/src/connector/with_options_source.yaml
@@ -1222,15 +1222,6 @@ PubsubProperties:
       `credentials` is a JSON string containing the service account credentials.
       See the [service-account credentials guide](https://developers.google.com/workspace/guides/create-credentials#create_credentials_for_a_service_account).
       The service account must have the `pubsub.subscriber` [role](https://cloud.google.com/pubsub/docs/access-control#roles).
-
-      If not specified, the connector will attempt to use Google Application Default Credentials (ADC).
-      ADC searches for credentials in the following order:
-      1. `GOOGLE_APPLICATION_CREDENTIALS_JSON` environment variable (JSON content)
-      2. `GOOGLE_APPLICATION_CREDENTIALS` environment variable (file path)
-      3. Well-known file: `$HOME/.config/gcloud/application_default_credentials.json`
-      4. Google Cloud metadata server (when running on GCE/GKE)
-
-      See [Google ADC documentation](https://cloud.google.com/docs/authentication/application-default-credentials) for more details.
     required: false
   - name: pubsub.start_offset.nanos
     field_type: String


src/connector/src/source/google_pubsub/enumerator/client.rs
```diff
@@ -50,17 +50,19 @@ impl SplitEnumerator for PubsubSplitEnumerator {
...
-        }
-
         let sub = properties.subscription_client().await?;
         if !sub
             .exists(None)
             .await
             .context("error checking subscription validity")?
         {
-            bail!("subscription {} does not exist", &sub.id())
+            bail!(
```
tabVersion, line 59: I'd suggest tighten the error message

  For the source path, the new message on `subscription does not exist` now also includes ADC guidance. That mixes two different failure modes:
  1. authentication/credential initialization failed
  2. authentication succeeded, but the subscription is missing

  It would be clearer to:
  - add ADC-specific guidance only when auth initialization fails
  - keep the resource validation error as just `subscription {id} does not exist`

  Similarly for the sink path, the current ADC fallback error is very long. Maybe shorten it to something like:
  `failed to initialize Google Cloud Pub/Sub ADC; please provide \`pubsub.credentials\`, configure ADC explicitly, or use \`pubsub.emulator_host\``

src/connector/src/sink/google_pubsub.rs
```diff
@@ -230,9 +223,27 @@ impl GooglePubSubSinkWriter {
         } else if let Some(emu_host) = config.emulator_host {
             Environment::Emulator(emu_host)
         } else {
-            return Err(SinkError::GooglePubSub(anyhow!(
-                "Missing emulator_host or credentials in Google Pub/Sub sink"
-            )));
+            let mut auth_config = project::Config::default();
```
tabVersion, line 226: I think we should disable ADC fallback in Cloud Hosted first, and only enable it for BYOC / self-hosted deployments.

you can refer to https://github.com/risingwavelabs/risingwave/blob/b60143545788ec02f2dd534e267653259a194e3c/src/connector/src/connector_common/common.rs#L56-L58

PR #24652, opened 2026-01-30
fix: Java Integration test cannot allocate memory in static TLS block

tabVersion:

Problem
CI observed java.lang.UnsatisfiedLinkError: ... cannot allocate memory in static TLS block failures in risingwave-connector-service tests.

This error occurs when the Rust JNI library (librisingwave_java_binding.so) is dynamically loaded by the JVM. The issue is likely caused by the default allocator (often jemalloc in RisingWave components) using the initial-exec TLS model. When loaded at runtime via System.loadLibrary, libraries using initial-exec require static TLS slots from a limited surplus pool. If this pool is exhausted, the library fails to load.

Solution
Explicitly set the global allocator to std::alloc::System in src/java_binding/src/lib.rs.

Using the system allocator for the JNI binding library avoids the heavy static TLS usage associated with high-performance allocators like jemalloc. This ensures the library can be safely loaded dynamically by the JVM without exhausting the available static TLS resources.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/java_binding/src/lib.rs
+++ b/src/java_binding/src/lib.rs
@@ -15,6 +15,11 @@
 #![feature(type_alias_impl_trait)]
 #![feature(try_blocks)]
 
+use std::alloc::System;
+
+#[global_allocator]
+static GLOBAL: System = System;
+
 mod hummock_iterator;
 use std::ffi::c_void;
 use std::ops::Deref;


wenym1 (approved):
LGTM!

PR #24654, opened 2026-01-30
fix: preallocate compaction task id range per loop and assert bounds

Li0k:

This PR improves compaction task id allocation by reusing prefetched ids across `get_compact_tasks_impl` calls while keeping refill behavior bounded by current remaining selection budget.

### Main changes
- Keep an in-memory prefetched task id range in `HummockManager` (`[next, end)`).
- Extract task id allocation into helpers in `hummock/manager/compaction/mod.rs`:
  - `refill_prefetched_compaction_task_id_range_if_needed`
  - `pop_prefetched_compaction_task_id`
  - `next_compaction_task_id_with_prefetch`
  - `remaining_compaction_task_id_refill_capacity`
- In `get_compact_tasks_impl`, compute refill capacity from remaining budget (`max_select_count - pick_tasks.len()`) to reduce over-prefetch while still avoiding per-group single-id SQL calls.

### Tests
Added unit tests for the new behavior:
- `test_remaining_compaction_task_id_refill_capacity`
- `test_next_compaction_task_id_with_prefetch_reuses_cached_range`
- `test_next_compaction_task_id_with_prefetch_refills_in_batch_when_empty`
- `test_next_compaction_task_id_with_prefetch_respects_refill_capacity`

Verified with:
- `cargo test -p risingwave_meta prefetched_task_id_tests --lib`

## Checklist

## Documentation

<details>

No user-facing behavior changes.

</details>


Li0k:
@zwang28 @hzxa21  Do you remember why each group fetches only one normal-task per pull?

zwang28:
I recommend storing prefetched task IDs within HummockManager to prevent them from being dropped after get_compact_tasks_impl returns.
To be clear, the waste of ids is not introduced by this PR.

src/meta/src/hummock/manager/compaction/mod.rs
```diff
@@ -283,6 +284,73 @@ impl HummockManager {
...
+        }
+
+        let new_range_start = self
+            .env
+            .hummock_seq
+            .next_interval(COMPACTION_TASK_ID, task_id_capacity)
+            .await?;
+        let mut prefetched_task_id_range = self.prefetched_compaction_task_id_range.lock();
+        // Re-check after `await`: another concurrent picker may have already refilled and started
+        // consuming the in-memory range. Do not overwrite that range.
```
zwang28: It means new_range_start is ignored..Can we avoid this waste?

zwang28: i.e. Concurrent pickers can wait for the current refill to complete.

src/meta/src/hummock/manager/compaction/mod.rs
```diff
@@ -368,8 +392,14 @@ impl HummockManager {
                 }
             };
 
-            // StoredIdGenerator already implements ids pre-allocation by ID_PREALLOCATE_INTERVAL.
-            let task_id = next_compaction_task_id(&self.env).await?;
+            // Use prefetched task id if available; when exhausted, refill in chunks first.
+            let task_id_refill_capacity = Self::remaining_compaction_task_id_refill_capacity(
```
zwang28: Why not use a fixed task_id_refill_capacity? Since prefetched IDs are never wasted, a constant capacity wouldn't result in any loss of efficiency.

Li0k: done

Li0k: @zwang28

PR #24659, opened 2026-01-30
fix(stream): fix eliminate no op

chenzl25:

- if [I, U-, U+] and eliminate_adjacent_noop_update optimize it to [invis, invis, U+] and cause dispatcher panic.

## Checklist

## Documentation

<details>

</details>


src/common/src/array/stream_chunk.rs
```diff
@@ -366,6 +366,20 @@ impl StreamChunk {
                 prev_r = Some(curr);
             }
         }
+
+        // Normalize update pairs that became partially invisible.
```
BugenZhao, line 370: Seems that ideally some cases do not need such normalization, i.e., when `U- U+ U- U+` becomes `U- . . U+`.

BugenZhao (approved):
Rest LGTM

PR #24668, opened 2026-01-31
fix(binder): fix schema-qualified column references for tables with the same name

### LINKED ISSUE #24667: Schema-qualified column references don't work correctly with tables that have the same name in different schemas
### Describe the bug

When I have two tables with the same name in different schemas (e.g., `public.users` and `public_staging.users`), using fully-qualified column names like `public_staging.users.id` doesn't correctly distinguish between the two tables in EXISTS subqueries.

### Error message/log

```text

```

### To Reproduce

```sql
-- Step 1: Create schemas
CREATE SCHEMA IF NOT EXISTS public_staging;

-- Step 2: Create first table in public schema
CREATE TABLE public.users (
    id INT PRIMARY KEY,
    name TEXT,
    is_active BOOLEAN
);

INSERT INTO public.users (id, name, is_active) VALUES
    (1, 'Alice', true),
    (2, 'Bob', false),
    (3, 'Charlie', false);

-- Step 3: Create second table with same name in different schema
CREATE TABLE public_staging.users AS
SELECT * FROM public.users WHERE is_active = false;

-- Verify: public_staging.users has 2 rows (Bob and Charlie)
SELECT * FROM public_staging.users;

-- Step 4: Query with EXISTS using fully-qualified column names
-- This SHOULD return 2 rows (Bob and Charlie who exist in both tables)
-- But it may return wrong results due to the bug
SELECT *
FROM public.users
WHERE EXISTS (
    SELECT 1
    FROM public_staging.users
    WHERE public_staging.users.id = public.users.id
    AND public_staging.users.is_active = false
);
```

The result after running the last query is:

```
 id |  name   | is_active
----+---------+-----------
  1 | Alice   | t
  2 | Bob     | f
  3 | Charlie | f
```

### Expected behavior

The expected result should be 2 rows (id=2 Bob, id=3 Charlie) and not (id=1 Alice) as query contains condition `is_active = false`.

```
 id |  name   | is_active
----+---------+-----------
  2 | Bob     | f
  3 | Charlie | f
```

The official Risingwave Docker Compose file.

### The version of RisingWave

```
dev=> select version();
                                    version
--------------------------------------------------------------------------------
 PostgreSQL 13.14.0-RisingWave-2.7.2 (30301dc965a6f30c08de859e2be0e6cb1b66f6b0)
(1 row)
```

```
➜ docker image ls | grep risingwave
risingwavelabs/risingwave                    v2.7.2                  70e134006f14   12 days ago     7.46GB
```

### Additional context

_No response_

tuantran0910:

Closes #24667

Fixed a bug where schema-qualified column references (e.g., schema.table.column) were incorrectly resolved when two tables had the same name but different schemas.

When users have tables with identical names in different schemas (common in multi-tenant apps or staging environments), using fully-qualified column names like `public_staging.users.id` doesn't correctly distinguish between the tables. The schema name was being parsed but then ignored during column binding, causing wrong query results.

```sql
-- Both tables have the same name but different schemas
public.users        -- has 3 rows (Alice, Bob, Charlie)
public_staging.users -- has 2 rows (Bob, Charlie)

-- This query should return 2 rows, but returns all 3
SELECT * FROM public.users
WHERE EXISTS (
SELECT 1 FROM public_staging.users
  WHERE public_staging.users.id = public.users.id
);
```

### To Reproduce

```sql
-- Step 1: Create schemas
CREATE SCHEMA IF NOT EXISTS public_staging;

-- Step 2: Create first table in public schema
CREATE TABLE public.users (
    id INT PRIMARY KEY,
    name TEXT,
    is_active BOOLEAN
);

INSERT INTO public.users (id, name, is_active) VALUES
    (1, 'Alice', true),
    (2, 'Bob', false),
    (3, 'Charlie', false);

-- Step 3: Create second table with same name in different schema
CREATE TABLE public_staging.users AS
SELECT * FROM public.users WHERE is_active = false;

-- Verify: public_staging.users has 2 rows (Bob and Charlie)
SELECT * FROM public_staging.users;

-- Step 4: Query with EXISTS using fully-qualified column names
-- This SHOULD return 2 rows (Bob and Charlie who exist in both tables)
-- But it may return wrong results due to the bug
SELECT *
FROM public.users
WHERE EXISTS (
    SELECT 1
    FROM public_staging.users
    WHERE public_staging.users.id = public.users.id
    AND public_staging.users.is_active = false
);
```

The result after running the last query is:

```
 id |  name   | is_active
----+---------+-----------
  1 | Alice   | t
  2 | Bob     | f
  3 | Charlie | f
```

### Expected behavior

The expected result should be 2 rows (id=2 Bob, id=3 Charlie) and not (id=1 Alice) as query contains condition `is_active = false`.

```
 id |  name   | is_active
----+---------+-----------
  2 | Bob     | f
  3 | Charlie | f
```

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/e2e_test/ddl/not_null.slt
+++ b/e2e_test/ddl/not_null.slt
@@ -118,7 +118,7 @@ id     NO
 name   YES
 
 query TT
-select rw_catalog.rw_columns.name, is_nullable from rw_catalog.rw_columns JOIN rw_catalog.rw_tables ON rw_catalog.rw_columns.relation_id = w_catalog.rw_tables.id WHERE rw_catalog.rw_tables.name = 'test_person' order by rw_catalog.rw_columns.name;
+select rw_catalog.rw_columns.name, is_nullable from rw_catalog.rw_columns JOIN rw_catalog.rw_tables ON rw_catalog.rw_columns.relation_id = rw_catalog.rw_tables.id WHERE rw_catalog.rw_tables.name = 'test_person' order by rw_catalog.rw_columns.name;
 ----
 _rw_timestamp t
 age           f
--- a/e2e_test/batch/duckdb/select/test_multi_column_reference.slt.part
+++ b/e2e_test/batch/duckdb/select/test_multi_column_reference.slt.part
@@ -227,8 +227,10 @@ CREATE TABLE s2.t1(t INT);
 statement ok
 INSERT INTO s2.t1 VALUES (42);
 
-statement error
+query I
 SELECT s1.t1.t FROM s1.t1, s2.t1
+----
+42
 
 # test various failures
 


e2e_test/ddl/not_null.slt
```diff
@@ -118,7 +118,7 @@ id     NO
 name   YES
 
 query TT
-select rw_catalog.rw_columns.name, is_nullable from rw_catalog.rw_columns JOIN rw_catalog.rw_tables ON rw_catalog.rw_columns.relation_id = w_catalog.rw_tables.id WHERE rw_catalog.rw_tables.name = 'test_person' order by rw_catalog.rw_columns.name;
+select rw_catalog.rw_columns.name, is_nullable from rw_catalog.rw_columns JOIN rw_catalog.rw_tables ON rw_catalog.rw_columns.relation_id = rw_catalog.rw_tables.id WHERE rw_catalog.rw_tables.name = 'test_person' order by rw_catalog.rw_columns.name;
```
tuantran0910, line 121: Somehow I find this typo which made the test failed.

tuantran0910:
Hi @xiangjinwu, can I have this PR reviewed :D I think this is actually a bug.

chenzl25 (approved):
Generally LGTM.

src/frontend/src/binder/relation/mod.rs
```diff
@@ -386,7 +388,14 @@ impl Binder {
...
 
-        match self.context.range_of.entry(table_name.clone()) {
+        // When a table has a schema, use schema.table as the key to prevent conflicts
+        // when two tables have the same name but different schemas.
+        let range_key = match &schema_name {
+            Some(s) => format!("{}.{}", s, table_name),
+            None => table_name.clone(),
+        };
+
+        match self.context.range_of.entry(range_key) {
```
chenzl25, line 398: It seems risky to update the key of `range_of` but doesn't change the search part in `select.rs`.

PR #24670, opened 2026-02-01
feat: add RescheduleIntent command and resolve in barrier worker

shanicky:

**Summary**
This PR introduces a new `RescheduleIntent` command variant to represent declarative intents for rescheduling jobs or fragments, deferring actual reschedule command generation to the barrier worker. It enhances the barrier processing pipeline by enforcing a two-phase rescheduling approach: intent declaration followed by dynamic resolution based on current cluster state.

**Details**
- Introduces `RescheduleIntent` enum within the barrier command module to encapsulate rescheduling intents for jobs or fragments without immediate execution.
- Updates barrier worker logic to asynchronously resolve `RescheduleIntent` commands into concrete `RescheduleFragment` commands before barrier injection, leveraging scale controller rendering and reschedule command building functions.
- Enforces strict validation in checkpoint control to reject unresolved `RescheduleIntent` commands at injection, ensuring intents are always resolved beforehand and preventing inconsistent state propagation.
- Extends inflight barrier info and command recognition to include `RescheduleIntent`, enabling seamless integration within the inflight barrier processing pipeline.
- Adds a test case verifying notifier failure notifications when resolution of `RescheduleIntent` fails due to absence of active workers, improving robustness in failure scenarios.

## Checklist

### DIFF
--- a/src/meta/src/barrier/checkpoint/control.rs
+++ b/src/meta/src/barrier/checkpoint/control.rs
@@ -1071,21 +1071,16 @@ impl DatabaseCheckpointControl {
             (None, vec![])
         };
 
-        if matches!(
-            &command,
-            Some(Command::RescheduleIntent {
-                reschedule_plan: None,
-                ..
-            })
-        ) {
-            warn!("reschedule intent should be resolved before injection");
-            for notifier in notifiers {
-                notifier.notify_start_failed(
-                    anyhow!("reschedule intent must be resolved before injection").into(),
-                );
-            }
-            return Ok(());
-        }
+        debug_assert!(
+            !matches!(
+                &command,
+                Some(Command::RescheduleIntent {
+                    reschedule_plan: None,
+                    ..
+                })
+            ),
+            "reschedule intent should be resolved before injection"
+        );
 
         if let Some(Command::DropStreamingJobs {
             streaming_job_ids, ..
--- a/src/meta/src/barrier/command.rs
+++ b/src/meta/src/barrier/command.rs
@@ -755,60 +755,63 @@ impl Command {
                 Some((Some((info.streaming_job.id(), cdc_tracker)), changes))
             }
             Command::RescheduleIntent {
-                reschedule_plan: Some(ReschedulePlan { reschedules, .. }),
-                ..
-            } => Some((
-                None,
-                reschedules
-                    .iter()
-                    .map(|(fragment_id, reschedule)| {
-                        (
-                            *fragment_id,
-                            CommandFragmentChanges::Reschedule {
-                                new_actors: reschedule
-                                    .added_actors
-                                    .iter()
-                                    .flat_map(|(node_id, actors)| {
-                                        actors.iter().map(|actor_id| {
-                                            (
-                                                *actor_id,
-                                                InflightActorInfo {
-                                                    worker_id: *node_id,
-                                                    vnode_bitmap: reschedule
-                                                        .newly_created_actors
-                                                        .get(actor_id)
-                                                        .expect("should exist")
-                                                        .0
-                                                        .0
-                                                        .vnode_bitmap
-                                                        .clone(),
-                                                    splits: reschedule
-                                                        .actor_splits
-                                                        .get(actor_id)
-                                                        .cloned()
-                                                        .unwrap_or_default(),
-                                                },
-                                            )
+                reschedule_plan, ..
+            } => {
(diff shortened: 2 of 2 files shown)


src/meta/src/barrier/command.rs
```diff
@@ -105,6 +107,78 @@ pub struct Reschedule {
...
+            downstream_fragments: HashMap::new(),
+            downstream_relations: HashMap::new(),
+        }
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.loaded.is_empty()
+    }
+
+    pub fn for_database(&self, database_id: DatabaseId) -> Option<Self> {
```
yezizp2012, line 143: It looks a bit strange but it's acceptable, preferably add some comments.

yezizp2012 (approved):
Rest LGTM.

src/meta/src/barrier/command.rs
```diff
@@ -567,6 +745,7 @@ impl Command {
 
                 Some((Some((info.streaming_job.id(), cdc_tracker)), changes))
             }
+            Command::RescheduleIntent { .. } => None,
```
wenym1: Can we directly replace `RescheduleFragment` with `RescheduleIntent`?

wenym1: A dirty but acceptable way is to have a `Option<reschedules>` in RescheduleIntent and set as None when constructing it, and then fill the value to some in `build_reschedule_from_context`, and in later usage in global barrier worker we can expect it to be set already.

shanicky: The main issue is that many parts actually rely on the reschedule fragment command, so changing it might require a lot of modifications.

wenym1: Currently the reschedule fragment command is
```
RescheduleFragment {
    reschedules: HashMap<FragmentId, Reschedule>,
    fragment_actors: HashMap<FragmentId, HashSet<ActorId>>,
},
```

I meant moving these two fields to `RescheduleIntent`
```
RescheduleIntent {
    context: RescheduleContext,
    reschedules: Option<HashMap<FragmentId, Reschedule>>,
    fragment_actors: Option<HashMap<FragmentId, HashSet<ActorId>>>,
},
```

And then just replace all appearance of `Command::RescheduleFragment` with `Command::RescheduleIntent`, and `reschedules` to `reschedules.expect("should set")` and `fragment_actors` to `fragment_actors.expect("should set")`. The work should be trivial.

wenym1: Besides, currently `RescheduleFragment` is only constructed when converting from `RescheduleIntent`, so the replacement won't increase complexity in the constructor side.

src/meta/src/stream/scale.rs
```diff
@@ -403,349 +266,570 @@ impl ScaleController {
...
+        .chain(all_upstream_fragments.values().flatten().map(|(id, _)| *id))
+        .chain(
+            all_downstream_fragments
+                .values()
+                .flatten()
+                .map(|(id, _)| *id),
+        )
+        .collect();
+
+    let read_guard = env.shared_actor_infos().read_guard();
```
wenym1: Since we are calling the method already in global barrier worker, can we instead just use the owned InflightFragmentInfo instead of the shared info?

src/meta/src/barrier/command.rs
shanicky: done

src/meta/src/barrier/checkpoint/control.rs
```diff
@@ -1064,6 +1071,22 @@ impl DatabaseCheckpointControl {
             (None, vec![])
         };
 
+        if matches!(
```
wenym1: I think the check here is unnecessary, because we should have ensured the `reschedule_plan` to be set here in global barrier worker before we reach here.

src/meta/src/barrier/command.rs
```diff
@@ -567,7 +754,10 @@ impl Command {
 
                 Some((Some((info.streaming_job.id(), cdc_tracker)), changes))
             }
-            Command::RescheduleFragment { reschedules, .. } => Some((
```
wenym1, line 570: I think instead of handling the case of `None` reschedules, we can just do `reschedules.expect("should set")`, because here we can expect the value to be already set.

Same for other `None` handling in this PR.

wenym1 (approved):
Rest LGTM!

PR #24678, opened 2026-02-02
Bound default adaptive parallelism to 64

kwannoel:

Updated the system default adaptive parallelism strategy to BOUNDED(64) so streaming jobs are capped unless explicitly configured otherwise. Split the source and table default bounds into separate constants while keeping both defaults at 4 for clarity and future tuning. Documentation and tests from the rebased PR remain intact.

## Checklist

## Documentation

<details>

</details>

```
Updated the defaults for system adaptive parallelism to bounded(64).
Updated the defaults for source and table to bounded(4).
```


kwannoel:
> Can you add a simple simulation test?

Given that this PR changes defaults, I don't think a new simulation test should be added. Testing defaults in simulation test means we have a new brittle test, everytime a default changes our test suite will break.

If we want, we could add a snapshot test, similar to what's done in `example.toml` for all session variables. That's out of scope for this PR though.

PR #24693, opened 2026-02-03
fix(snapshot-backfill): fix deadlock caused by next_epoch

wenym1:

Fix deadlock caused by back-pressure to upstream from `try_wait_epoch` in `next_epoch`.

# Fix Snapshot Backfill Deadlock: Always Consume Until Checkpoint is Buffered

## Problem

During snapshot backfill Phase 2 (consuming log store), a deadlock can occur where:

1. **Creating job stops consuming** from upstream channel because `can_consume_upstream()` returns `false`
   - This happens when `pending_epoch_lag >= max_pending_epoch_lag` and `is_polling_epoch_data = false`
   - `max_pending_epoch_lag` shrinks over time via `saturating_sub(elapsed_epoch / 2)`

2. **No checkpoint is buffered yet** - only non-checkpoint barriers are in the buffer

3. **Channel fills up** with non-checkpoint barriers (E18, E19, ...)

4. **Checkpoint barrier (E20) is blocked** from being sent by main graph actor

5. **Deadlock**: 
   - Creating job waits for `next_epoch()` → needs E20 committed
   - E20 committed → needs E20 collected  
   - E20 collected → needs main graph to complete `channel.send(E20)`
   - `channel.send(E20)` → needs channel space
   - Channel space → needs creating job to drain channel
   - Creating job → is `pending()` forever!

```
+---------------------+          +---------------------+
|   Snapshot Backfill |          |     Main Graph      |
|                     |          |                     |
|  next_epoch(E15)    |          |  Blocked trying to  |
|  waiting for E20    |          |  SEND E20 to        |
|  to be committed    |          |  full channel       |
+----------+----------+          +----------+----------+
           |                                |
           v                                v
+---------------------+          +---------------------+
|  Checkpoint E20     |          |  Channel space      |
|  to be committed    |          |  (need backfill to  |
|                     |          |   drain channel)    |
|  Needs E20 sent     |          |                     |
|  and collected      |          |  But backfill has   |
|                     |          |  pending() upstream |
+---------------------+          +---------------------+
           ^                                |
           +--------------------------------+
                    CIRCULAR WAIT!
```

## Solution

Modify `can_consume_upstream()` to **always return `true`** when no checkpoint is buffered yet:

```rust
fn can_consume_upstream(&self) -> bool {
    // Always consume if no checkpoint is buffered yet (prevents deadlock)
    !self.upstream_pending_barriers.has_checkpoint_epoch()
        || self.is_polling_epoch_data
        || self.pending_epoch_lag() < self.max_pending_epoch_lag
}
```

### 1. Breaks the Deadlock

By ensuring we always consume **until we have a checkpoint buffered**, the main graph can always:
- Send the checkpoint barrier to the channel
- Complete the send operation
- Report the checkpoint as collected
- Allow meta to commit the checkpoint

### 2. Backpressure Still Works

The fix only bypasses backpressure **before** receiving a checkpoint. Once a checkpoint is buffered:
- Original condition applies: `pending_epoch_lag() < max_pending_epoch_lag`
- Backpressure kicks in normally for subsequent epochs
- The checkpoint is already safely buffered, so commit will succeed

### 3. Bounded Buffering

The "unbounded" buffering is actually bounded by `checkpoint_frequency`:

```
Worst case per checkpoint cycle:
  - We buffer at most (checkpoint_frequency) epochs without limit
  - E.g., with checkpoint_frequency=5: buffer E16,E17,E18,E19,E20 = 5 epochs max
  
This is O(checkpoint_frequency), not unbounded!
```

### 4. Downstream Can Still Catch Up

The fix doesn't change steady-state behavior:
- `consumed_epoch()` still reduces `max_pending_epoch_lag` over time
- Once we have a checkpoint, backpressure applies normally
- If downstream processes faster than upstream produces, it catches up
- If downstream is permanently slower, it would fail regardless of this fix

### 5. Invariant Preserved

The original backpressure goal: **"don't buffer too far ahead of what we've consumed"**

With the fix, this invariant is preserved:
- We only force-buffer **up to the next checkpoint** (minimum required)
- After checkpoint is buffered, backpressure applies for additional epochs
- The checkpoint is what we need for `next_epoch()` to return anyway

## Summary

| Aspect | Before Fix | After Fix |
|--------|-----------|-----------|
| Deadlock possible? | YES (channel blocks checkpoint) | NO (always drain to checkpoint) |
| Unbounded buffering? | No (but deadlocks) | Only up to next checkpoint (O(checkpoint_frequency)) |
| Backpressure works? | N/A (deadlock) | YES (after checkpoint received) |
| Catch-up possible? | N/A (deadlock) | YES (same as intended design) |

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/executor/dispatch.rs
+++ b/src/stream/src/executor/dispatch.rs
@@ -539,31 +539,8 @@ impl StreamConsumer for DispatchExecutor {
                     }
                     MessageBatch::BarrierBatch(barrier_batch) => {
                         assert!(!barrier_batch.is_empty());
-                        let stop_barrier = if barrier_batch[0].is_stop(self.inner.actor_id) {
-                            assert_eq!(
-                                barrier_batch.len(),
-                                1,
-                                "stop barrier with mutation must be handled singly"
-                            );
-                            Some(barrier_batch[0].clone())
-                        } else {
-                            None
-                        };
-                        if stop_barrier.is_none() {
-                            // Yield barriers before dispatch. Unless it's a stop barrier.
-                            //
-                            // Assume that there are more than 1 barriers in the batch(let's call their epochs E1, E2, ...),
-                            // and we do the actual dispatch before yielding the barriers, When downstream receive E1, and start
-                            // processing it, it may wait for E1 committed. However, E2 is not received,
-                            // the actual dispatch may not return, which blocks the yielding of E1 and subsequently
-                            // block E1 committed, which causes deadlock.
-                            for barrier in &barrier_batch {
-                                yield barrier.clone();
-                            }
-                        }
-                        let barrier_batch_len = barrier_batch.len();
                         self.inner
-                            .dispatch(MessageBatch::BarrierBatch(barrier_batch))
+                            .dispatch(MessageBatch::BarrierBatch(barrier_batch.clone()))
                             .instrument(tracing::info_span!("dispatch_barrier_batch"))
                             .instrument_await("dispatch_barrier_batch")
                             .await?;
@@ -571,12 +548,9 @@ impl StreamConsumer for DispatchExecutor {
                             .metrics
                             .metrics
                             .barrier_batch_size
-                            .observe(barrier_batch_len as f64);
-                        if let Some(stop_barrier) = stop_barrier {
-                            // For stop barrier, we yield it after dispatch to downstream,
-                            // because after yielding the stop barrier, the executor will have stopped,
-                            // and the subsequent dispatch won't be handled.
-                            yield stop_barrier;
+                            .observe(barrier_batch.len() as f64);
+                        for barrier in barrier_batch {
+                            yield barrier;
                         }
                     }
                     watermark @ MessageBatch::Watermark(_) => {


wenym1:
* **#24693** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24693?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/24693?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

hzxa21 (approved):
Good find!

Wonder how you find out the deadlock

wenym1:
Seems this is not the root cause. The CI in this PR still timeout with the same deadlock. Let me continue investigating.

wenym1:
Potential root cause analyzed by opus-4.5
```
## Snapshot Backfill Deadlock Analysis

### Setup

Configuration:
  - checkpoint_frequency = 5
  - barrier_interval_ms = 250
  - upstream_channel_capacity = 16 (bounded channel)

Epochs Timeline:
  E1 --- E2 --- E3 --- E4 --- E5 --- E6 --- E7 --- E8 --- E9 --- E10
                              ^                                    ^
                          checkpoint                           checkpoint

Components:
  +----------------------------------------------------------------------+
  |                           MAIN GRAPH                                  |
  |  +---------+    +---------+    +---------+                           |
  |  | Source  |--->|  Join   |--->|Dispatcher|--+                       |
  |  | (T162)  |    |         |    |          |  |                       |
  |  +---------+    +---------+    +----------+  |                       |
  |                                              |                       |
  +----------------------------------------------|------------------------+
                                                 |
                               upstream channel  |  (bounded: 16 slots)
                                                 v
  +---------------------------------------------------------------------+
  |                        CREATING JOB (MV)                             |
  |  +-----------------+    +---------------+    +----------------+     |
  |  |  Merge Executor |--->| Snapshot      |--->|  Downstream    |     |
  |  |  (receives      |    | Backfill      |    |  (state store) |     |
  |  |   upstream)     |    | (Actor 665)   |    |                |     |
  |  +-----------------+    +---------------+    +----------------+     |
  |                               |                                      |
  |                               | reads change log                     |
  |                               v                                      |
  |                      +----------------+                              |
  |                      |  Hummock       |                              |
  |                      |  (Table 154)   |                              |
  |                      +----------------+                              |
  +----------------------------------------------------------------------+

CRITICAL INSIGHT:
  For main graph to COLLECT barrier E_n:
    1. Actor processes E_n
    2. Actor sends E_n to dispatcher
    3. Dispatcher sends E_n to channel  <-- BLOCKS if channel full!
    4. Actor reports E_n collected to meta

  If step 3 blocks, step 4 never happens, and E_n cannot be committed!

---

### Phase 1: Consuming Snapshot (max_pending_epoch_lag = u64::MAX)

During Phase 1 (consuming snapshot), max_pending_epoch_lag is initialized to u64::MAX,
meaning there's no limit on buffered epochs:

    // executor.rs line 566
    UpstreamBuffer {
        max_pending_epoch_lag: u64::MAX,  // No limit during snapshot phase
        ...
    }

---

### Transition: Phase 1 -> Phase 2 (max_pending_epoch_lag gets set)

When transitioning from Phase 1 to Phase 2, max_pending_epoch_lag is set to
the CURRENT pending_epoch_lag:

    // executor.rs line 590
    fn start_consuming_log_store(mut self, consumed_epoch: u64) {
        let max_pending_epoch_lag = self.pending_epoch_lag();  // <-- SET HERE!
        ...
    }

Example: At transition time

    State when Phase 1 ends:
      - upstream_pending_barriers = [E3, E4, E5(ckpt)]
      - consumed_epoch = E0 (initial)
      - pending_epoch_lag = E5.prev - E0 = E4 - E0 = 4 epochs

    After start_consuming_log_store(E0):
      - max_pending_epoch_lag = 4  <-- Locked in!

---

### Step 1: Initial State (Phase 2 starts, E5 just committed)

                        MAIN GRAPH                     CREATING JOB
                    +--------------+                +------------------+
                    |              |   upstream     |                  |
                    |   Actors     |---channel----> |  Snapshot        |
                    |              |   [empty]      |  Backfill        |
                    +--------------+                +------------------+

State:
  - Last committed checkpoint: E5
  - Change log for T154: [{non_ckpt: [E1,E2,E3,E4], ckpt: E5}]
  - consumed_epoch = E5
  - upstream_pending_barriers = []
  - pending_epoch_lag = 0
  - max_pending_epoch_lag = 4 (set at Phase 1 -> Phase 2 transition)

---

### Step 2: Processing Barrier E6 - Upstream Buffer Fills

Snapshot Backfill receives barrier E6, calls run_future(next_epoch(E5))

                        MAIN GRAPH                     CREATING JOB
                    +--------------+                +------------------+
                    |              |   upstream     |  run_future:     |
                    |   Sending    |---channel----> |   select! {      |
                    |   E6,E7,E8,  |   [E6,E7,E8,   |     consume()    |
                    |   E9,E10     |    E9,E10]     |     next_epoch() |
                    |              |                |   }              |
                    +--------------+                +------------------+

Action: concurrently_consume_upstream() drains channel until checkpoint E10

                    +--------------+                +------------------+
                    |              |   upstream     |                  |
                    |   Actors     |---channel----> |  Buffered:       |
                    |   (idle)     |   [empty]      |  [E6,E7,E8,E9,   |
                    |              |                |   E10(ckpt)]     |
                    +--------------+                +------------------+

State after draining:
  - upstream_pending_barriers = [E6, E7, E8, E9, E10]
  - is_polling_epoch_data = false (just received checkpoint)
  - pending_epoch_lag = 4 epochs
  - max_pending_epoch_lag = 4
  - can_consume_upstream() = false || (4 < 4) = FALSE!

But next_epoch(E5) returns quickly because E5 was already committed,
so E6 exists in the change log.

---

### Step 3: Over Time - max_pending_epoch_lag Shrinks

As epochs are consumed, max_pending_epoch_lag decreases:

    // executor.rs line 692-695
    self.max_pending_epoch_lag = min(
        self.pending_epoch_lag(),
        self.max_pending_epoch_lag.saturating_sub(elapsed_epoch / 2),
    );

Example progression:
  - After consuming E6:  max_pending_epoch_lag = min(current_lag, 4 - 0) = 4
  - After consuming E8:  max_pending_epoch_lag = min(current_lag, 4 - 1) = 3
  - After consuming E10: max_pending_epoch_lag = min(current_lag, 3 - 1) = 2
  - ...continues shrinking...

---

### Step 4: The Critical Moment - Backpressure Triggers BEFORE Checkpoint

At some point during Phase 2:

State BEFORE receiving the next checkpoint:
  - consumed_epoch = E15
  - upstream_pending_barriers = [E16, E17, E18, E19] <-- NO CHECKPOINT YET!
  - pending_epoch_lag = 4 epochs
  - max_pending_epoch_lag = 2 (shrunk over time)
  - E20 (checkpoint) is STILL IN THE CHANNEL or being produced by main graph

Processing barrier E16, calls run_future(next_epoch(E15)):

+-------------------------------------------------------------------------+
|  can_consume_upstream() evaluation:                                      |
|                                                                          |
|    is_polling_epoch_data = false                                        |
|    pending_epoch_lag = 4           <-- (E19 - E15)                      |
|    max_pending_epoch_lag = 2       <-- decreased over time              |
|                                                                          |
|    can_consume = is_polling_epoch_data || (pending_lag < max_lag)       |
|                = false             || (4 < 2)                           |
|                = false             || false                             |
|                = FALSE  X                                                |
+-------------------------------------------------------------------------+

concurrently_consume_upstream() becomes PENDING IMMEDIATELY!

    async fn concurrently_consume_upstream(&mut self) {
        loop {
            if !self.can_consume_upstream() {
                return pending().await;  // <-- STOPS CONSUMING!
            }
            ...
        }
    }

---

### Step 5: E20 Never Gets Received!

CRITICAL: The creating job stops consuming BEFORE E20 arrives!

                     MAIN GRAPH                         CREATING JOB
                 +------------------+               +------------------+
                 |                  |   upstream    |                  |
                 |  Trying to send  |---channel---> |  NOT READING!    |
                 |  E20 (ckpt)!     |   [E20,E21,   |                  |
                 |                  |    E22,...]   |  pending() in    |
                 |  But channel     |               |  consume_upstream|
                 |  filling up...   |   FILLING!    |                  |
                 +------------------+               +------------------+

Timeline:
  T0: Creating job has [E16,E17,E18,E19] buffered, no checkpoint
  T1: can_consume_upstream() = false, stops consuming
  T2: Main graph produces E20 (checkpoint), sends to channel
  T3: Main graph produces E21, E22, E23... sends to channel
  T4: Channel fills up (16 slots)
  T5: Main graph BLOCKS trying to send E37

---

### Step 6: Main Graph Blocks - E20 Still in Channel, Not Collected!

                     MAIN GRAPH                         CREATING JOB
                 +------------------+               +------------------+
                 |                  |   upstream    |                  |
                 |  Actor trying    |---channel---> |  NOT READING!    |
                 |  to send E37     |   [FULL!]     |                  |
                 |                  |   E20,E21,    |  Buffered only:  |
                 |  BLOCKED on      |   E22,E23,    |  [E16,E17,E18,   |
                 |  channel.send()! |   ...E36      |   E19]           |
                 |                  |               |                  |
                 |  Cannot finish   |               |  Waiting for     |
                 |  barrier loop!   |               |  next_epoch(E15) |
                 +------------------+               +------------------+

KEY INSIGHT: E20 is in the channel, but NOT yet received by creating job!

The creating job stopped consuming at E19. E20 is sitting in the channel.
For E20 to be "collected" by the main graph:
  - The main graph actor must COMPLETE the send operation
  - The send operation blocks when channel is full on E37
  - The actor cannot move on to report E20 collected

Wait, E20 was sent to the channel before the block on E37...
Let me reconsider the collection mechanism.

---

### Step 6 (Corrected): When is E20 "Collected"?

Actually, barrier collection works like this:

    Main Graph Actor Loop:
      for each barrier in stream {
          process(barrier)
          send_to_downstream(barrier)  // Async, queued to channel
          report_collected(barrier)    // Reported after send returns
      }

If send_to_downstream(E37) blocks:
  - E20 through E36 were already sent (channel wasn't full yet)
  - E20's report_collected(E20) was already called
  - E20 SHOULD be collected!

So E20 should be collected... unless the channel was full earlier?

Let's reconsider with a smaller channel:

---

### Step 6 (Re-corrected): Channel Fills BEFORE E20 is Sent

If channel_capacity = 3:

Timeline:
  T0: Creating job buffers [E16, E17, E18, E19], stops consuming
  T1: Creating job calls run_future(next_epoch(E15)), pending()
  
  T2: Main graph sends E20 to channel --> [E20]
  T3: Main graph sends E21 to channel --> [E20, E21]
  T4: Main graph sends E22 to channel --> [E20, E21, E22] FULL!
  T5: Main graph tries to send E23 --> BLOCKED!

E20 was sent before the block, so E20 should be collected.
This still doesn't cause a deadlock on E20...

The REAL deadlock requires E20 itself to be blocked!

---

### Step 6 (Final): The TRUE Deadlock - Channel Fills BEFORE Checkpoint

For a deadlock, the checkpoint barrier must be BLOCKED from sending:

Scenario with very small channel OR fast barrier production:
  - Channel capacity = 2
  - Creating job buffers [E16, E17], stops consuming (pending())
  
Timeline:
  T0: upstream_pending_barriers = [E16, E17], no checkpoint buffered
  T1: can_consume_upstream() = false, stops consuming channel
  
  T2: Main graph sends E18 to channel --> [E18]
  T3: Main graph sends E19 to channel --> [E18, E19] FULL!
  T4: Main graph tries to send E20 (checkpoint) --> BLOCKED!!!
  T5: E20 NEVER SENT!
  T6: E20 NEVER COLLECTED!
  T7: E20 NEVER COMMITTED!

+-------------------------------------------------------------------------+
|  THE TRUE DEADLOCK:                                                      |
|                                                                          |
|  Creating Job:                                                           |
|    - Has [E16, E17] buffered (no checkpoint!)                           |
|    - Waiting for next_epoch(E15) which needs E20 committed              |
|    - Not consuming from channel (pending)                                |
|                                                                          |
|  Upstream Channel:                                                       |
|    - Contains [E18, E19] (no checkpoint!)                               |
|    - FULL!                                                               |
|                                                                          |
|  Main Graph:                                                             |
|    - Trying to send E20 (checkpoint)                                    |
|    - BLOCKED on channel.send()!                                         |
|    - Cannot complete send, cannot report collected                      |
|                                                                          |
|  Meta:                                                                   |
|    - Waiting for all actors to report E20 collected                     |
|    - Will never receive report from blocked actor                       |
|    - E20 NEVER COMMITTED!                                               |
+-------------------------------------------------------------------------+

---

### Step 7: Final Deadlock State

+-------------------------------------------------------------------------+
|                           DEADLOCK CYCLE                                 |
|                                                                          |
|    +---------------------+          +---------------------+             |
|    |   Snapshot Backfill |          |     Main Graph      |             |
|    |                     |          |                     |             |
|    |  next_epoch(E15)    |          |  Blocked trying to  |             |
|    |  waiting for E20    |          |  SEND E20 to        |             |
|    |  to be committed    |          |  full channel       |             |
|    |                     |          |                     |             |
|    +----------+----------+          +----------+----------+             |
|               |                                |                        |
|               |   WAITS FOR                    |   WAITS FOR            |
|               v                                v                        |
|    +---------------------+          +---------------------+             |
|    |  Checkpoint E20     |          |  Channel space      |             |
|    |  to be committed    |          |  (need backfill to  |             |
|    |                     |          |   drain channel)    |             |
|    |  Needs E20 sent     |          |                     |             |
|    |  and collected      |          |  But backfill has   |             |
|    |                     |          |  pending() upstream |             |
|    +---------------------+          +---------------------+             |
|               ^                                |                        |
|               |                                |                        |
|               +--------------------------------+                        |
|                                                                          |
|                        +===================+                            |
|                        |  CIRCULAR WAIT!   |                            |
|                        |    DEADLOCK!      |                            |
|                        +===================+                            |
+-------------------------------------------------------------------------+

---

### Summary: The Key Conditions for Deadlock

+-------------------------------------------------------------------------+
|  CONDITION 1: max_pending_epoch_lag shrinks over time                    |
|  ----------------------------------------------------                    |
|  Each consumed_epoch() call reduces max_pending_epoch_lag.               |
|  Eventually: pending_epoch_lag >= max_pending_epoch_lag                  |
|                                                                          |
|  CONDITION 2: Backpressure triggers BEFORE checkpoint buffered           |
|  --------------------------------------------------------------          |
|  Creating job stops consuming when:                                      |
|    can_consume_upstream() = false                                        |
|  This happens when only non-checkpoint barriers are buffered!            |
|                                                                          |
|  CONDITION 3: Channel fills up with non-checkpoint barriers              |
|  -----------------------------------------------------------             |
|  Channel fills with E18, E19, ... (non-checkpoint epochs)               |
|  E20 (checkpoint) is BLOCKED from being sent!                           |
|                                                                          |
|  CONDITION 4: Checkpoint cannot be sent/collected                        |
|  ------------------------------------------------                        |
|  Main graph actor blocks on channel.send(E20)                           |
|  Actor cannot complete barrier processing                                |
|  Actor cannot report E20 collected to meta                              |
|  Meta cannot commit E20                                                  |
|                                                                          |
|  RESULT: Circular dependency --> DEADLOCK                                |
|                                                                          |
|  Creating job waits for E20 committed                                   |
|       --> needs E20 collected                                           |
|       --> needs E20 sent to channel                                     |
|       --> needs channel space                                           |
|       --> needs creating job to drain channel                           |
|       --> but creating job is pending()!                                |
+-------------------------------------------------------------------------+

---

### The Root Cause in Code

// executor.rs line 609-611 - Back-pressure check
fn can_consume_upstream(&self) -> bool {
    self.is_polling_epoch_data || self.pending_epoch_lag() < self.max_pending_epoch_lag
}

// executor.rs line 617-619 - Becomes pending when back-pressure triggers
async fn concurrently_consume_upstream(&mut self) -> StreamExecutorError {
    loop {
        if !self.can_consume_upstream() {
            return pending().await;  // <-- THE BUG: Stops consuming forever!
        }
        ...
    }
}

The fix should ensure that even when we've buffered many epochs, we still
continue to drain the upstream channel (at least barriers) to prevent blocking
the main graph from sending checkpoint barriers.

```

hzxa21:
Does it mean that this can only happen when checkpoint_frequency is set to >1?

wenym1:
The analysis of AI seems correct, but it's actually wrong. In step 4 of the analysis, it assumes `is_polling_epoch_data = false`. `is_polling_epoch_data` means whether we have polled something else after receiving a checkpoint barrier. In the step, 4 non-checkpoint barrier E16-E19 is in the buffer, which means that we must have polled something after the checkpoint barrier E15, and therefore `is_polling_epoch_data = true` in this scenario, and the deadlock is not true.

The root cause is much more simpler. Calling `next_epoch` on a committed checkpoint epoch won't return immediately, until the next checkpoint epoch is committed. Therefore, for example, when we call `next_epoch(E15)`, it won't return until `E20` has been committed. `E15` is a checkpoint barrier. When we haven't polled anything after `E15`, `is_polling_epoch_data = false`, and if the `max_pending_lag` is too small, the `can_consume_upstream` will return false, and then we will impose backpressure to upstream, which may block the barrier of E20 in the upstream, which causes the deadlock.

To fix it, we can simply disable the `next_epoch` call, because it's only for sanity check.

shanicky (commented):
reschedule part lgtm

wenym1:
The previous fix is fixing a potential deadlock, but it still times out in the e2e source test. The root cause deadlock is like the following
```

                        +-----------------------+
                        | HashJoin (align E63) |
                        +----+--------------+--+
                             |              |
                             |              |
       waits for E65 commit  |              | back-pressure propagates
                             |              v
                  +----------+----+     +---+-------------------+
                  | Snapshot BF   |     | Snapshot BF          |
                  | from log      |     | direct from upstream |
                  | (iter_log E61 |     | (already pulled E63) |
                  |  -> needs E65)|     +-----------+----------+
                  +----------+----+                 |
                             |                      |
                             |                      | stalls upstream main graph on E65
                             +------------------------------------------+
                                                                v
                                              +-----------------+----------------+
                                              | Upstream main graph (E65)        |
                                              +----------------------------------+

```

The fix is that, we won't do iter_log on non-checkpoint epoch. Instead, we save the non-checkpoint epoch and yield the barrier anyway, and when receiving the checkpoint epoch, we do then iter_log on `earliest non-checkpoint epoch` to `checkpoint epoch`. This can actually improve the efficiency, because the key-value are stored together for checkpoint and non-checkpoint together. If we only do iter_log on checkpoint, we don't have to iter_log on a same data block for multiple times.

Now the CI can pass, including all main-cron CI.

xiangjinwu:
Not sure if this timeout (rebased today) is the same issue:
https://buildkite.com/risingwavelabs/pull-request/builds/91587#019c416b-742d-4b4c-a941-3dc2e160a760

yuhao-su:
Looks like we are still seeing some deadlock https://buildkite.com/risingwavelabs/pull-request/builds/92413#019c96fc-b3c9-4aec-84ee-943bdeefe383 This is rebased on latest main

PR #24703, opened 2026-02-04
feat: add a config to control whether state table vnode prunning is enabled

hzxa21:

This pull request introduces a new "dry-run" mode for vnode stats pruning in state tables. The main goal is to allow validation of vnode stats pruning logic without actually applying pruning during reads, making it safer to test and verify correctness before enabling this optimization in production. The changes add a configuration flag, propagate it through the relevant code paths, and implement logic to log warnings if the pruning decision would have been incorrect in dry-run mode.

Key changes include:

**Configuration and Documentation:**

- Added a new config option `enable_state_table_vnode_stats_prunning` (default: false) to enable dry-run mode for vnode stats pruning. This is reflected in `streaming.rs`, the example config (`example.toml`), and the generated documentation (`docs.md`). [[1]](diffhunk://#diff-44452374fc2620c74f12ec3372ab01632d983659155a079d741bde9b8044f1e4R281-R288) [[2]](diffhunk://#diff-0842d7b738eeee546a0239c9a6875dac418ee2533985a6cbe7ed028a80f277a4R423-R426) [[3]](diffhunk://#diff-432b175bfef2be21c05f5aadd4c6a6fbed0eaff05881a070c3692f5f6b769735R241) [[4]](diffhunk://#diff-34e9dbf05e21681c680a9192e753a645e119ce6691a8670e5cf56fb021934939R341)

**State Table Builder and Internal Wiring:**

- Updated the `StateTableBuilder` and related structs to propagate the new dry-run flag, renaming the previous pruning flag to clarify its purpose, and ensuring the new setting is passed to all relevant components. [[1]](diffhunk://#diff-f7503ae24cca193f3b9eee1c478ea36710f7157e9b74af2355a32ceb9c968351L575-R582) [[2]](diffhunk://#diff-f7503ae24cca193f3b9eee1c478ea36710f7157e9b74af2355a32ceb9c968351L592-R600) [[3]](diffhunk://#diff-f7503ae24cca193f3b9eee1c478ea36710f7157e9b74af2355a32ceb9c968351L609-R618) [[4]](diffhunk://#diff-f7503ae24cca193f3b9eee1c478ea36710f7157e9b74af2355a32ceb9c968351L653-R665) [[5]](diffhunk://#diff-f7503ae24cca193f3b9eee1c478ea36710f7157e9b74af2355a32ceb9c968351L677-R694) [[6]](diffhunk://#diff-f7503ae24cca193f3b9eee1c478ea36710f7157e9b74af2355a32ceb9c968351L694-R706) [[7]](diffhunk://#diff-f7503ae24cca193f3b9eee1c478ea36710f7157e9b74af2355a32ceb9c968351L747-R760) [[8]](diffhunk://#diff-f7503ae24cca193f3b9eee1c478ea36710f7157e9b74af2355a32ceb9c968351L963-R977)

**Dry-run Pruning Logic in State Table Operations:**

- Modified `get`, `exists`, `iter`, and `rev_iter` methods in `StateTableRowStore` to support dry-run mode:  
  - When dry-run is enabled, pruning is not actually applied, but if the pruning logic would have incorrectly skipped data, a warning is logged.  
  - Ensures that in dry-run mode, all reads still go to cache/storage even if pruning would indicate no results, and correctness is validated via logs. [[1]](diffhunk://#diff-f7503ae24cca193f3b9eee1c478ea36710f7157e9b74af2355a32ceb9c968351L1138-R1166) [[2]](diffhunk://#diff-f7503ae24cca193f3b9eee1c478ea36710f7157e9b74af2355a32ceb9c968351L1155-R1193) [[3]](diffhunk://#diff-f7503ae24cca193f3b9eee1c478ea36710f7157e9b74af2355a32ceb9c968351L1178-R1224) [[4]](diffhunk://#diff-f7503ae24cca193f3b9eee1c478ea36710f7157e9b74af2355a32ceb9c968351L1198-R1247) [[5]](diffhunk://#diff-f7503ae24cca193f3b9eee1c478ea36710f7157e9b74af2355a32ceb9c968351L1713-R1815) [[6]](diffhunk://#diff-f7503ae24cca193f3b9eee1c478ea36710f7157e9b74af2355a32ceb9c968351L1748-R1836) [[7]](diffhunk://#diff-f7503ae24cca193f3b9eee1c478ea36710f7157e9b74af2355a32ceb9c968351L1771-R1905) [[8]](diffhunk://#diff-f7503ae24cca193f3b9eee1c478ea36710f7157e9b74af2355a32ceb9c968351L1807-R1926)

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/common/table/state_table.rs
+++ b/src/stream/src/common/table/state_table.rs
@@ -1786,9 +1786,9 @@ impl<LS: LocalStateStore, SD: ValueRowSerde> StateTableRowStore<LS, SD> {
 
         let table_id = self.table_id;
         let inspect_fn = move |result: &StreamExecutorResult<(K, OwnedRow)>| {
-            // Only log error when in dry-run mode and we would have pruned but got results
+            // Only log when in dry-run mode and we would have pruned but got results
             if should_prune_entirely && result.is_ok() {
-                tracing::error!(
+                tracing::warn!(
                     table_id = %table_id,
                     "vnode stats pruning dry run fails for iter. This will not affect correctness."
                 );
@@ -1874,9 +1874,9 @@ impl<LS: LocalStateStore, SD: ValueRowSerde> StateTableRowStore<LS, SD> {
 
         let table_id = self.table_id;
         let inspect_fn = move |result: &StreamExecutorResult<(K, OwnedRow)>| {
-            // Only log error when in dry-run mode and we would have pruned but got results
+            // Only log when in dry-run mode and we would have pruned but got results
             if should_prune_entirely && result.is_ok() {
-                tracing::error!(
+                tracing::warn!(
                     table_id = %table_id,
                     "vnode stats pruning dry run fails for rev_iter. This will not affect correctness."
                 );


src/stream/src/common/table/state_table.rs
Li0k: question: The doc indicates we need to invalidate stats, but there is currently no update range operation, so it has no correctness effect. Is this the expected behavior?

src/stream/src/common/table/state_table.rs
```diff
@@ -1135,14 +1149,21 @@ impl<LS: LocalStateStore, SD: ValueRowSerde> StateTableRowStore<LS, SD> {
...
         {
             if let Some(m) = &self.metrics {
                 m.get_vnode_pruned_count.inc();
             }
+            true
+        } else {
+            false
+        };
+
+        // In dry-run mode, we verify pruning correctness but don't apply it
```
Li0k: I'm quite confused why this branch has a dry-run document. It hits the prune logic and returns directly.

src/stream/src/common/table/state_table.rs
```diff
@@ -1175,14 +1207,21 @@ impl<LS: LocalStateStore, SD: ValueRowSerde> StateTableRowStore<LS, SD> {
...
 
         // Try to prune using vnode statistics
-        if let Some(stats) = &self.vnode_stats
+        let should_prune = if let Some(stats) = &self.vnode_stats
             && let (vnode, key) = key_bytes.split_vnode_bytes()
             && let Some(vnode_stat) = stats.get(&vnode)
             && vnode_stat.can_prune(&key)
         {
             if let Some(m) = &self.metrics {
                 m.get_vnode_pruned_count.inc();
```
Li0k, line 1215: Regardless of whether it is a dry-run or not, **get_vnode_pruned_count** will be counted. It is impossible to tell at a glance from the metrics whether pruning is enabled. Should we add an additional actual_pruned_count metric to make a semantic distinction?

src/stream/src/common/table/state_table.rs
hzxa21: The doc is incorrect. We consider delete key as normal key. For example, for the following sequence
```
+k1
+k3
-k3
```

vnode stat can only prune out read with key > k3 instead of key > k1

src/stream/src/common/table/state_table.rs
hzxa21: The documentation is wrong. Fixed.

src/stream/src/common/table/state_table.rs
hzxa21: I think it is fine. The pruning metrics is to help us understand whether the pruning can be efficient. Given that we will eventually deprecate this PR when this feature is fully rolled out and default enable_state_table_vnode_stats_pruning to true, I think it is fine to not introduce extra metric.

PR #24715, opened 2026-02-05
feat: support table functions for datafusion

silver-ymz:

This PR adds support for table functions (e.g. `generate_series`, `unnest`) when executing queries via DataFusion engine.

Because RisingWave’s “project set” semantics (used to implement set-returning functions) differ from DataFusion’s native model, we translate `LogicalProjectSet` into a custom DataFusion extension node `ProjectSetLogicalPlan` and provide a matching physical implementation `ProjectSetExec`. The executor follows the logic of RisingWave’s legacy batch `ProjectSetExecutor` to preserve consistent behavior and results between the old batch engine and the DataFusion engine.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/datafusion/project_set.rs
+++ b/src/frontend/src/datafusion/project_set.rs
@@ -14,12 +14,11 @@
 
 use std::any::Any;
 use std::cmp::Ordering;
-use std::sync::{Arc, LazyLock};
+use std::sync::Arc;
 
 use async_trait::async_trait;
 use datafusion::arrow::datatypes::SchemaRef as ArrowSchemaRef;
 use datafusion::arrow::record_batch::RecordBatch;
-use datafusion::execution::context::QueryPlanner;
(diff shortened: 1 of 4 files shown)


src/frontend/src/datafusion/project_set.rs
```diff
@@ -0,0 +1,424 @@
...
+        Ok(Some(Arc::new(ProjectSetExec::new(
+            physical_inputs[0].clone(),
+            select_list,
+            schema,
+        ))))
+    }
+}
+
+#[derive(Debug)]
+pub struct ProjectSetQueryPlanner;
```
chenzl25: It seems like we can use a more general name for this planner in case in the future we would support more self-defined plan node in this planner.

silver-ymz: refactor to `RwCustomQueryPlanner` in new commit

PR #24718, opened 2026-02-05
refactor(source): skip bad file splits after retry in fs_fetch executor

wcy-fdu:

- Track bad file splits in memory after retries and skip them on subsequent reads.
- Avoid repeated per‑barrier errors without persisting dirty state.
- Keep file attribution accurate for early read errors.

## Checklist

## Documentation

<details>

</details>


wcy-fdu:
Tests are a bit hard to add because the goal is to reduce per‑barrier bad‑file error logs; the source/fetcher logic itself is unchanged.

src/stream/src/executor/source/fs_fetch_executor.rs
```diff
@@ -158,6 +228,9 @@ impl<S: StateStore, Src: OpendalSource> FsFetchExecutor<S, Src> {
                 .await?
                 .map_err(StreamExecutorError::connector_error);
                 let single_file_stream = single_file_stream.map(|reader| reader);
```
tabVersion: useless map op?

src/stream/src/executor/source/fs_fetch_executor.rs
```diff
@@ -253,36 +330,92 @@ impl<S: StateStore, Src: OpendalSource> FsFetchExecutor<S, Src> {
...
+                            retries_done = *retries_done,
+                            max_retries = MAX_RETRIES_PER_SPLIT,
+                            error = %e.as_report(),
+                            "Fetch Error, retrying file split"
+                        );
+                        tokio::time::sleep(backoff).await;
+                    } else {
+                        // Exceeded max retries: mark dirty in memory and skip this split afterwards.
+                        dirty_splits.insert(split_id.clone());
+                        retry_counts.remove(&split_id);
```
tabVersion, line 395: `retry_counts` only clean up when exceeded max retries, shall we remove also when the file failed but succeed?

src/stream/src/executor/source/fs_fetch_executor.rs
```diff
@@ -12,13 +12,17 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::collections::{HashMap, HashSet};
 use std::marker::PhantomData;
 use std::ops::Bound;
+use std::sync::{Arc, Mutex};
```
tabVersion, line 18: maybe parking_lot::Mutex for more fairness mutex

tabVersion (commented):
Overall LGTM, please add metrics for the number of dirty splits, per failed file count.

src/stream/src/executor/source/fs_fetch_executor.rs
wcy-fdu: parking_lot::Mutex isn’t fair by default either (fairness requires unlock_fair), and this is a short critical section, so std::sync::Mutex is fine.

wcy-fdu:
> Overall LGTM, please add metrics for the number of dirty splits, per failed file count.

metrics added.

wcy-fdu:
cc @tabVersion PTAL❤️

src/stream/src/executor/source/fs_fetch_executor.rs
```diff
@@ -46,9 +50,66 @@ use crate::executor::prelude::*;
 use crate::executor::stream_reader::StreamReaderWithPause;
```
tabVersion, line 51: Should these constants be configurable via system parameters or source properties? Hardcoded values make it difficult to tune for different deployment environments. For example, a source reading from a high-latency cloud storage might need more retries, while a local POSIX FS source might want fewer.

PR #24728, opened 2026-02-06
feat(sink): support firestore sink 

xxhZs:

This PR adds Firestore sink support.

It introduces the Firestore connector options needed to create a sink, including `firestore.project_id`, `firestore.database_id`, `firestore.collection`, and `firestore.credentials`. With these options, users can create a sink such as:

```sql
CREATE SINK s1 FROM all_types_test WITH (
  connector = 'firestore',
  primary_key = 'id',
  firestore.project_id = 'xx',
  firestore.database_id = 'xx',
  firestore.collection = 'xx',
  firestore.credentials = 'xx'
);
```

The intent is to allow RisingWave sink data to be written into a configured Firestore collection.

## Checklist

## Documentation

<details>

RisingWave now supports creating Firestore sinks with Firestore project, database, collection, credential, and primary key configuration.

</details>


hzxa21 (commented):
How is this PR tested? If the test cannot be automated in codes, can you provide a document with the following information?
- How to setup the firestore environment
- The test script to perform end-to-end functional test and correctness test.

hzxa21:
There is an official firestore local emulator: https://docs.cloud.google.com/firestore/native/docs/emulator

src/connector/src/sink/firestore.rs
```diff
@@ -0,0 +1,475 @@
...
+            let pk_values: Vec<String> = self
+                .pk_indices
+                .iter()
+                .map(|&idx| {
+                    row.datum_at(idx)
+                        .map(|s| s.to_text())
+                        .unwrap_or_else(|| "null".to_owned())
+                })
+                .collect();
+            Ok(pk_values.join("_"))
```
hzxa21: When `firestore.document_id_column` is not set, this derives the document id by joining PK text values with `_`. That mapping is not injective: for example, composite PKs `(a_b, c)` and `(a, b_c)` both become `a_b_c`, and a string value `null` can collide with a NULL rendered as `null`. The later upsert/delete can overwrite or delete the wrong Firestore document.

Please encode the PK tuple unambiguously, for example with a JSON array plus base64url, length-prefixed components, or a stable hash over typed PK values. Also consider validating/encoding Firestore document-id constraints such as `/` so ordinary PK values do not fail only at write time.

src/connector/src/sink/firestore.rs
```diff
@@ -0,0 +1,475 @@
...
+
+            match op {
+                Op::Insert | Op::UpdateInsert => {
+                    let data = self.format_row(row)?;
+                    operations.insert(doc_id.clone(), FirestoreOperation::Upsert { doc_id, data });
+                }
+                Op::Delete => {
+                    operations.insert(doc_id.clone(), FirestoreOperation::Delete { doc_id });
+                }
+                Op::UpdateDelete => {
```
hzxa21, line 318: Skipping `UpdateDelete` is only safe when the old and new rows map to the same Firestore document id. With a mutable `firestore.document_id_column`, an update from `doc_id = old` to `doc_id = new` will skip deleting `old` and then upsert `new`, leaving a stale document behind.

Please either require `document_id_column` to be one of the downstream PK columns, or treat `UpdateDelete` as a delete. The latter should still coalesce correctly for same-document updates because the later `UpdateInsert` for the same doc id replaces the earlier delete in the `BTreeMap`.

src/connector/src/sink/firestore.rs
```diff
@@ -0,0 +1,475 @@
...
+    }
+
+    async fn new_log_sinker(&self, _writer_param: SinkWriterParam) -> Result<Self::LogSinker> {
+        Ok(FirestoreSinkWriter::new(
+            self.config.clone(),
+            self.schema.clone(),
+            self.pk_indices.clone(),
+        )
+        .await?
+        .into_log_sinker(self.config.max_future_send_nums))
```
hzxa21, line 212: `firestore.max_future_send_nums` is passed directly to `AsyncTruncateLogSinker`, but validation above only rejects `max_batch_size == 0`. If users configure `firestore.max_future_send_nums = 0`, `DeliveryFutureManagerAddFuture::add_future_may_await` waits while `future_count >= max_future_count`; with zero, that is true before any future is added, so the first write can spin/wait forever with no delivery future available.

Please reject `firestore.max_future_send_nums == 0` during validation or config parsing.

hzxa21 (commented):
Posting review findings from local review.

xxhZs:
> There is an official firestore local emulator: https://docs.cloud.google.com/firestore/native/docs/emulator

I tried testing Firestore locally, but local testing failed because the authentication redirect URL differs between local and remote environments. Below are the details of my remote tests:
Create a Firestore instance in Google Cloud.
Create a RisingWave table and Firestore sink as shown below:
sql
CREATE TABLE firestore_t (
    id INT PRIMARY KEY,
    name VARCHAR,
    active BOOLEAN,
    tags VARCHAR[],
    nested STRUCT<a INT, b VARCHAR>
);

CREATE SINK s1 FROM firestore_t WITH (
    connector = 'firestore',
    primary_key = 'id',
    firestore.database_id = 'test',
    firestore.project_id = 'rwc-playground',
    firestore.collection = 'test',
    firestore.credentials = 'xxx'
);
Perform insert, update and delete operations, then verify whether the data in Firestore matches expectations.

chenzl25:
I found one remaining correctness issue around `firestore.document_id_column`.

Today validation only checks that the column exists. If it is not guaranteed to be a unique document key, the sink can conflate different RisingWave rows into the same Firestore document. For example, with RW primary key `id` and `firestore.document_id_column = 'tenant'`, two rows `(1, 't')` and `(2, 't')` will overwrite/coalesce into the same Firestore doc; deleting one row can delete the other's document as well. The default path avoids this by encoding the full downstream PK tuple, but the custom document-id path bypasses that guarantee.

Can we either reject `document_id_column` unless it is exactly the single downstream PK column, or otherwise keep using the full downstream PK for the Firestore document id? If arbitrary columns are intentionally allowed, the docs and validation should make the uniqueness requirement explicit. Also, this raw path should validate or encode Firestore document-id constraints such as `/`, unlike the default base64url PK encoding.

One smaller config validation gap: `firestore.max_batch_size` should reject values greater than Firestore's 500-write batch limit, not only `0`, since it is used directly in `.chunks(self.config.max_batch_size)`.

src/connector/src/sink/firestore.rs
```diff
@@ -0,0 +1,511 @@
...
+                                        SinkError::Firestore(
+                                            anyhow!(e)
+                                                .context("failed to add delete document to batch"),
+                                        )
+                                    })?;
+                            }
+                        }
+                    }
+
+                    batch.write().await.map_err(|e| {
```
hzxa21: `batch.write().await` returning `Ok(_)` only means the BatchWrite RPC itself succeeded; Firestore BatchWrite is not atomic and each write can still fail independently. The `firestore` crate returns a `FirestoreBatchWriteResponse` with a `statuses` entry for each write, but this code drops the response and acknowledges the sink chunk as delivered.

Please inspect the returned statuses and fail the delivery future if any write status is non-OK, ideally including the document id or operation context in the error. Otherwise a partial batch failure can be silently committed by the sink, causing data loss.

xxhZs:
Thanks for the review. I updated this PR with the following changes:

- For `firestore.document_id_column`, arbitrary document-id columns are intentional: this option lets users choose the Firestore document key mapping. I documented that the configured column values must be unique, and that duplicate values intentionally map multiple RisingWave rows to the same Firestore document. I also added validation for explicit Firestore document IDs so invalid values such as empty IDs, `/`, `.` and `..` fail clearly before the batch write.
- Added validation that `firestore.max_batch_size` must be within the Firestore batch write limit (`<= 500`).
- Started checking `BatchWrite` per-write statuses and returning an error if any individual write fails, instead of acknowledging a partially failed batch.

src/connector/src/sink/firestore.rs
```diff
@@ -0,0 +1,574 @@
...
+    };
+
+    let value = match data_type {
+        DataType::Boolean => JsonValue::Bool(scalar.into_bool()),
+        DataType::Int16 => JsonValue::Number(scalar.into_int16().into()),
+        DataType::Int32 => JsonValue::Number(scalar.into_int32().into()),
+        DataType::Int64 => {
+            // JSON numbers have limited precision, use string for large integers
+            let v = scalar.into_int64();
+            if v.abs() > (1i64 << 53) {
```
chenzl25: Firestore supports signed 64-bit integer values directly, and `serde_json::Number::from(i64)` preserves the full `i64`; the Firestore serializer ultimately emits `IntegerValue(i64)`. Converting only values above 2^53 to strings therefore makes one RisingWave `INT64` column data-dependent in Firestore: small values are integers while large values are strings, which breaks numeric filters/order/index semantics. `v.abs()` also panics for `i64::MIN` in overflow-checked builds. Please serialize every `INT64` as `JsonValue::Number(v.into())` (and add boundary tests including `i64::MIN`/`i64::MAX`).

src/connector/Cargo.toml
```diff
@@ -75,9 +75,11 @@ duration-str = "0.18.0"
 easy-ext = "1"
 either = "1"
 elasticsearch = { version = "8.17.0-alpha.1", features = ["rustls-tls"] }
-enum-as-inner = "0.7"
+enum-as-inner = "0.6"
```
chenzl25: This PR also downgrades the connector crate from `enum-as-inner` 0.7 to 0.6. That is unrelated to Firestore and reverses the repository-wide dependency bump from #24680, so it looks like a stale merge/conflict resolution artifact. Please keep 0.7 and regenerate `Cargo.lock`.

chenzl25:
One more correctness issue: NULL values are silently dropped by the current Firestore serialization path. `format_datum()` produces `serde_json::Value::Null`, but the default `firestore-rs` serializer encodes it with an empty `value_type`, which is then filtered out of maps and arrays. In a local repro, `nullable = NULL` removed the field entirely, and `[1, NULL, 2]` became `[1, 2]`; nested NULLs in JSONB/STRUCT/LIST are affected as well.

Please explicitly encode Firestore `NullValue` and add regression tests for both a NULL field and a NULL array element.

chenzl25:
Reviewed head `52903d51b`. Besides the JSONB big-integer wraparound already reported, a few more issues:

1. **`firestore.max_future_send_nums` is a no-op.** [`write_chunk`](https://github.com/risingwavelabs/risingwave/blob/52903d51bc7ce9c795abe98fae3cd3dbdda93cd5/src/connector/src/sink/firestore.rs#L569-L576) awaits `write_chunk_inner` inline and ignores `_add_future`, so the delivery-future pipeline (`into_log_sinker(max_future_send_nums)`, `FirestoreSinkDeliveryFuture`) never engages and this user-facing option (default 256) does nothing. Either register the future via `add_future` or drop the option and type.

2. **NaN/±Infinity floats are silently written as strings.** In [`format_datum`](https://github.com/risingwavelabs/risingwave/blob/52903d51bc7ce9c795abe98fae3cd3dbdda93cd5/src/connector/src/sink/firestore.rs#L476-L487), `Number::from_f64` returns `None` for NaN/Inf and falls back to `JsonValue::String`. Firestore's `doubleValue` natively supports NaN/Infinity, so the same column flips between double and string across rows — another silent shape change.

Minor:
- Deeply nested JSONB (>128 levels) fails `serde_json::from_str` at [L501-L505](https://github.com/risingwavelabs/risingwave/blob/52903d51bc7ce9c795abe98fae3cd3dbdda93cd5/src/connector/src/sink/firestore.rs#L501-L505) and `unwrap_or_else` silently stores the whole JSONB as one string field; the `to_string` + re-parse round trip is also an extra full copy.
- In `FirestoreSinkWriter::new`, `doc_id_col_idx` silently becomes `None` when `document_id_column` is missing from the schema, falling back to PK-encoded doc IDs. `validate()` catches this today, but it should be a hard error here too.

Checked OK: per-chunk dedup gives correct last-op-wins; concurrent sub-batches are safe after dedup; base64(JSON) composite-PK encoding is unambiguous and path-safe; retries are idempotent.

xxhZs:
Regarding local testing, Google provides an official local testing tool, yet its authentication flow differs from ours. Our service authenticates against Google Cloud, whereas the local variant relies on a local port for auth. In my opinion, there is no need for us to implement a full set of logic solely for testing purposes that will never be utilized in production. As such, local testing will not be supported for the time being.

PR #24731, opened 2026-02-06
refactor(metric): remove actor_id from sink metrics

zwang28:

The label `actor_id` in prometheus SINK metrics increases the metric volume significantly. It is never used in grafana.
This PR removes it from SINK metrics.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/executor/sink.rs
+++ b/src/stream/src/executor/sink.rs
@@ -360,11 +360,7 @@ impl<F: LogStoreFactory> SinkExecutor<F> {
             // TODO(hzxa21): support rate limit?
             processed_input.boxed()
         } else {
-            let labels = [
-                &actor_id.to_string(),
-                &sink_id.to_string(),
-                self.sink_param.sink_name.as_str(),
-            ];
+            let labels = [&sink_id.to_string(), self.sink_param.sink_name.as_str()];
             let log_store_first_write_epoch = GLOBAL_SINK_METRICS
                 .log_store_first_write_epoch
                 .with_guarded_label_values(&labels);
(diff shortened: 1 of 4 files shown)


src/stream/src/executor/sink.rs
```diff
@@ -360,11 +360,7 @@ impl<F: LogStoreFactory> SinkExecutor<F> {
             // TODO(hzxa21): support rate limit?
             processed_input.boxed()
         } else {
-            let labels = [
-                &actor_id.to_string(),
-                &sink_id.to_string(),
-                self.sink_param.sink_name.as_str(),
-            ];
+            let labels = [&sink_id.to_string(), self.sink_param.sink_name.as_str()];
```
wenym1, line 363: `actor_id` here is useful in debugging, because when one of the parallelisms is encountering error, only one actor_id can have abnormal metrics.

wenym1:
> The label actor_id in prometheus SINK metrics increases the metric volume significantly.

Is there a concrete number about how it affects the metrics volume?

zwang28:
> > The label actor_id in prometheus SINK metrics increases the metric volume significantly.
> 
> Is there a concrete number about how it affects the metrics volume?

The size of SINK metric is multipled by N, where N is the number of SINK parallelism.

src/stream/src/executor/sink.rs
zwang28: OK. 
If we use relabel instead of removing actor_id, will it work? By default the actor_id is dropped. During debug it's enabled by modifying MetricLevel.

PR #24733, opened 2026-02-06
refactor: improve compaction schedule

Li0k:

## Summary

Optimize compaction scheduling performance and fix a merge compatibility check bug.

### 1. Fix merge config mismatch bug (`63b9286bf1`)

Bug: When validating if two compaction groups can be merged, the original code compared the entire `CompactionConfig`. However, the `split_weight_by_vnode` field is only used for per-table split behavior and will be reset to default (0) after merge. This caused unnecessary merge rejections when only `split_weight_by_vnode` differed between groups.

Fix: Introduced `merge_compatibility_config()` to normalize `split_weight_by_vnode` to 0 before comparison, and `is_merge_compatible_by_semantics()` to perform semantic compatibility check.

### 2. Optimize compaction state lock contention (`7633689818`)

- Added `snapshot_with_shuffled_group_ids()` to reduce lock holding time during periodic compaction scheduling
- The snapshot approach allows quick lock acquisition and release, preventing blocking of `commit_epoch` operations

### 3. Add cooldown for periodic Dynamic trigger (`648a933321`)

- Added `dynamic_cooldown` HashSet to track recently unscheduled groups
- Groups that return no task during periodic Dynamic scheduling enter cooldown until next `commit_epoch`
- Increased `periodic_compaction_interval_sec` from 60s to 300s (5 minutes)

### 4. Introduce CompactionScheduleSnapshot to fix cooldown race condition (`ee9c893548`, `72244ef591`)

- Introduced `CompactionScheduleSnapshot` with `snapshot_time` to solve the race between `commit_epoch` and periodic scheduling
- When `unschedule()` is called, it compares `last_new_data_time` with `snapshot_time` to decide if cooldown should apply
- Added `ScheduleTrigger` enum (`NewData` / `Periodic`) to distinguish scheduling triggers

### 5. Clean up compaction schedule state when group is removed (`5c9afebb30`)

- Added `remove_compaction_group()` to `CompactionState` to clean up `scheduled`, `dynamic_cooldown`, and `last_new_data_time` entries when a group is dropped or merged

### 6. Merge try_sched_compaction functions with ScheduleTrigger enum (`32a9c16880`)

- Unified `try_sched_compaction` and `try_sched_new_data_compaction` into a single function parameterized by `ScheduleTrigger`
- Simplified caller logic by removing duplicated scheduling paths

### 7. Simplify try_sched_compaction return type and improve docs (`c7c779f816`)

- Changed `pick_compaction_groups_and_type` to return `Option<(Vec<CompactionGroupId>, TaskType)>` instead of `(Vec<...>, TaskType)` with empty-vec sentinel
- Added `TASK_TYPE_PRIORITY` const array for priority-based task type selection in `pick_type`

### 8. Add unit tests for CompactionState (`1eb7a76c64`)

- Added 12 unit tests covering: basic scheduling, cooldown behavior, `ScheduleTrigger` semantics, `NewData` cooldown clearing, group removal, snapshot isolation, and mixed-type task scheduling

### 9. Simplify CompactionScheduleSnapshot internals (`624691a64f`)

- `group_ids_shuffled`: replaced `sort_unstable()` + `dedup()` with `itertools::unique()`
- `try_sched_compaction`: replaced `contains()` + `insert()` (5 lines) with `HashSet::insert()` return value (1 line)

### 10. Extract try_dispatch_tasks to structurally guard PullTaskAck (`b889f154e1`)

- Extracted task selection/dispatch logic into `try_dispatch_tasks()` so that `PullTaskAck` is structurally guaranteed to execute
- Prevents accidental PullTaskAck skip (which would cause compactor deadlock) by any future early-return in task selection logic

## Motivation

When there are hundreds of compaction groups, the periodic compaction trigger (every 60s) would schedule all groups regardless of whether they have pending work. This caused:
1. Unnecessary iteration overhead in `auto_pick_compaction_groups_and_type()`
2. Potential lock contention with `commit_epoch` operations

The cooldown mechanism ensures groups without pending compaction work are skipped in subsequent periodic triggers until new data arrives (signaled by `commit_epoch`).

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/meta/src/hummock/manager/compaction/compaction_group_schedule.rs
+++ b/src/meta/src/hummock/manager/compaction/compaction_group_schedule.rs
@@ -1121,7 +1121,7 @@ impl GroupMergeValidator {
             );
 
             return Err(Error::CompactionGroup(format!(
-                "Cannot merge group {} and next_group {} with different compaction config. left_config: {:?}, right_config: {:?}",
+                "Cannot merge group {} and next_group {} with different compaction config (split_weight_by_vnode is excluded from comparison). left_config: {:?}, right_config: {:?}",
                 group.group_id, next_group.group_id, left_config, right_config
             )));
         }
--- a/src/meta/src/hummock/manager/compaction/mod.rs
+++ b/src/meta/src/hummock/manager/compaction/mod.rs
@@ -1100,9 +1100,7 @@ impl HummockManager {
         Ok(())
     }
 
-    /// Sends a compaction request triggered by new data (e.g., from `commit_epoch`).
-    ///
-    /// This clears any cooldown for the group since new data arrived.
+    /// Sends a compaction request for new data (clears cooldown).
     pub fn try_send_compaction_request(
         &self,
         compaction_group: CompactionGroupId,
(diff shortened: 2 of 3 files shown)


src/meta/src/hummock/manager/compaction/compaction_group_schedule.rs
```diff
@@ -927,6 +927,25 @@ impl HummockManager {
...
+    fn merge_compatibility_config(group: &CompactionGroupStatistic) -> CompactionConfig {
+        let mut config = group
+            .compaction_group_config
+            .compaction_config
+            .as_ref()
+            .clone();
+        // Ignore per-table split behavior when validating merge compatibility.
+        // After merge, the hybrid group will reset this field to default(0).
+        config.split_weight_by_vnode = 0;
+        config
```
hzxa21: minor: I think it is better to impl a custom comparator function without considering `split_weight_by_vnode` instead of cloning the configs.

src/meta/src/hummock/mock_hummock_meta_client.rs
```diff
@@ -281,15 +281,14 @@ impl HummockMetaClient for MockHummockMetaClient {
...
+                let (groups, task_type) =
+                    hummock_manager_compact.auto_pick_compaction_groups_and_type();
 
-                if group_and_type.is_none() {
+                if groups.is_empty() {
                     break;
                 }
 
-                let (group, task_type) = group_and_type.unwrap();
+                let group = groups[0];
```
hzxa21: It seems to make the call more heavier because the early stop when getting task from one group is no longer there after replacing `auto_pick_compaction_groups_and_type` with `auto_pick_compaction_group_and_type`. Will this slow down CI tests?

src/meta/src/hummock/manager/compaction/mod.rs
```diff
@@ -1415,17 +1400,40 @@ impl CompactionState {
         self.scheduled.lock().remove(&(compaction_group, task_type));
     }
 
-    pub fn auto_pick_type(&self, group: CompactionGroupId) -> Option<TaskType> {
-        let guard = self.scheduled.lock();
-        if guard.contains(&(group, compact_task::TaskType::Dynamic)) {
+    /// Take a snapshot of the scheduled set to reduce lock contention.
```
hzxa21: What will this not affect correctness?

src/meta/src/hummock/manager/compaction/compaction_group_schedule.rs
Li0k: Certainly, I just thought clone was more concise. And,  this isn't a hot path, so I'm prioritizing readability. Using a comparator means we have to list all configs, okay with you?

src/meta/src/hummock/mock_hummock_meta_client.rs
Li0k: 1. group-count magnitude small in testing
2. HashMap low overhead

Li0k: let's see ci tests result

src/meta/src/hummock/manager/compaction/mod.rs
Li0k:
```
/// A point-in-time snapshot of the compaction schedule state.
///
/// The `snapshot_time` is essential for solving the cooldown race condition:
/// - When `unschedule()` is called, it compares `last_new_data_time` with `snapshot_time`
/// - If `last_new_data_time > snapshot_time`, new data arrived during processing,
///   so the group should NOT enter cooldown
///
/// # Race Condition Example
/// ```text
/// T1: snapshot(t0) ───────────────────── unschedule(A, snapshot_time=t0)
///                                             │
///                                             ├─ last_new_data_time[A] = t1
///                                             │  t1 > t0 ✓ → skip cooldown
///                                             │
/// T2: ───── try_sched_compaction(A) ──────────
///           (records last_new_data_time = t1)
/// ```
```

The snapshot-based design is perfectly valid. Even if the state falls behind, it will not cause errors but only result in a wasted trigger. However, I identified a race condition in the cooldown logic, so I refactored it to keep everything concise

PR #24756, opened 2026-02-09
feat: support atomically adding column for iceberg table

yezizp2012:

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/handler/mod.rs
+++ b/src/frontend/src/handler/mod.rs
@@ -1612,7 +1612,8 @@ fn check_ban_ddl_for_iceberg_engine_table(
                         .get_schema_by_name(&db_name, &schema_name)
                         .ok()
                         .and_then(|schema| schema.get_created_sink_by_name(&sink_name));
-                    sink.and_then(|s| s.auto_refresh_schema_from_table).is_some()
+                    sink.and_then(|s| s.auto_refresh_schema_from_table)
+                        .is_some()
                 } else {
                     false
                 };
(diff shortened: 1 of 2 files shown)


src/meta/src/controller/streaming_job.rs
```diff
@@ -1638,11 +1768,13 @@ impl CatalogController {
                     Table::update(table::ActiveModel {
                         table_id: Set(log_store_table_id),
                         columns: Set(new_log_store_table_columns.clone()),
+                        value_indices: Set(I32Array::from(new_value_indices.clone())),
```
chenzl25: I am wondering why this PR would need to change the log store table? Shouldn't it have been handled by sink auto-schema change?

src/meta/src/rpc/ddl_controller.rs
```diff
@@ -1344,24 +1344,31 @@ impl DdlController {
...
-                        !original_table_column_ids.remove(&ColumnId::new(
+                        !original_table_column_ids.contains(&ColumnId::new(
                             col.column_desc.as_ref().unwrap().column_id as _,
                         ))
                     })
                     .map(|col| ColumnCatalog::from(col.clone()))
                     .collect_vec();
-                if !original_table_column_ids.is_empty() {
-                    return Err(anyhow!("new table columns does not contains all original columns. new: {:?}, original: {:?}, not included: {:?}", table.columns, original_table_columns, original_table_column_ids).into());
-                }
```
chenzl25, line 1454: It seems we loose this check.

src/meta/src/stream/stream_graph/fragment.rs
```diff
@@ -417,9 +418,29 @@ pub fn check_sink_fragments_support_refresh_schema(
     let PbNodeBody::Sink(_) = sink_node.node_body.as_ref().unwrap() else {
         return Err(anyhow!("expect PbNodeBody::Sink but got: {:?}", sink_node.node_body).into());
     };
-    let [stream_scan_node] = sink_node.input.as_slice() else {
+    let [stream_input_node] = sink_node.input.as_slice() else {
         panic!("Sink has more than 1 input: {:?}", sink_node.input);
     };
+    let stream_scan_node = match stream_input_node.node_body.as_ref().unwrap() {
```
chenzl25, line 402: Lots of changes in this files. Involve @Chiro11 to take a look as well.

src/meta/src/rpc/ddl_controller.rs
yezizp2012: You are right. This check should stay for the current scope (add-column only). I have added an explicit guard back: if any original columns are removed, we now return an error (`auto schema change sink only supports add column now`). This avoids silently allowing drop-column in this PR.

yezizp2012:
@chenzl25 I'm still working on it, There may be breaking issues with the previous sink auto schema change implementation. I'm still debugging and will let you know when it's ready for review.

src/meta/src/controller/streaming_job.rs
yezizp2012: reverted.

chenzl25:
@wenym1 Could you please help to review the fragment part together?

src/frontend/src/handler/mod.rs
```diff
@@ -1600,109 +1600,68 @@ fn check_ban_ddl_for_iceberg_engine_table(
...
-            let (table, schema_name) = get_table_catalog_by_table_name(session.as_ref(), name)?;
-            if table.is_iceberg_engine_table() {
-                bail!(
-                    "ALTER TABLE SET SOURCE RATE LIMIT is not supported for iceberg table: {}.{}",
-                    schema_name,
-                    name
-                );
-            }
+        _ => {
+            bail!(
```
chenzl25, line 1661: Actually we support altering iceberg table with connector https://github.com/risingwavelabs/risingwave/pull/22040

chenzl25:
> In `src/meta/src/stream/stream_graph/fragment.rs`, I think the drop-column path in `rewrite_log_store_table` is still wrong. It removes entries from `log_store_table.columns`, but it does not rebuild `log_store_table.value_indices` (`fragment.rs:522-534`). The new test also still expects the old indices (`fragment.rs:2652-2656`).
That looks unsafe because the KV log-store serde uses `value_indices` as direct indexes into the table columns (`src/stream/src/common/log_store_impl/kv_log_store/serde.rs:209-230`). After dropping a column, those old indexes can point to the wrong column or even go out of bounds.
I think `value_indices` needs to be rewritten together with `columns` here.

Maybe drop column need to consider the above case.

yezizp2012:
> > In `src/meta/src/stream/stream_graph/fragment.rs`, I think the drop-column path in `rewrite_log_store_table` is still wrong. It removes entries from `log_store_table.columns`, but it does not rebuild `log_store_table.value_indices` (`fragment.rs:522-534`). The new test also still expects the old indices (`fragment.rs:2652-2656`).
> > That looks unsafe because the KV log-store serde uses `value_indices` as direct indexes into the table columns (`src/stream/src/common/log_store_impl/kv_log_store/serde.rs:209-230`). After dropping a column, those old indexes can point to the wrong column or even go out of bounds.
> > I think `value_indices` needs to be rewritten together with `columns` here.
> 
> Maybe drop column need to consider the above case.

In the previous implementation of sink schema change, `value_indices` would not be changed and used. I initially modified this part, leading to inconsistent behavior. Later I found out that column information changes are maintained separately in the meta store, and IIUC drop column is already supported. Anyway, I will continue implementing drop column after this PR, and then test to see if there are any issues.

PR #24765, opened 2026-02-10
feat(cdc): support postgres 18

wcy-fdu:

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/risedev.yml
+++ b/risedev.yml
@@ -1526,7 +1526,7 @@ template:
     application: "metastore"
 
     # The docker image. Can be overridden to use a different version.
-    image: "postgres:18-alpine"
+    image: "postgres:17-alpine"
 
     # If set to true, data will be persisted at data/{id}.
     persist-data: true


java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/PostgresValidator.java
```diff
@@ -112,10 +112,10 @@ public void validateDbConfig() {
...
             // some time to support it. So even though 18 is not released yet, we put a version
             // guard here.
-            if (pgVersion >= 18) {
-                throw ValidatorUtils.failedPrecondition(
-                        "Postgres major version should be less than or equal to 17.");
-            }
+            // if (pgVersion >= 18) {
+            //     throw ValidatorUtils.failedPrecondition(
+            //             "Postgres major version should be less than or equal to 17.");
+            // }
```
hzxa21: Should we just remove the code directly?

wcy-fdu: I think we should convert it from `>= 18 `to `>18`

hzxa21: Makes sense.

PR #24773, opened 2026-02-10
refactor(meta): split hummock timer task into per-handler loops

### LINKED ISSUE #24675: Critical Stream Starvation in Hummock Timer Task Causes Heartbeat Failures
### Describe the bug

The Hummock manager's timer event loop in `src/meta/src/hummock/manager/timer_task.rs` suffers from a critical **stream starvation bug**. The current architecture utilizes `futures::stream::select_all` to merge multiple timer streams but processes them **sequentially** in a single loop.

This architectural flaw causes time-critical events (specifically `CompactionHeartBeatExpiredCheck`) to be blocked by long-running operations in other event handlers (like `Report`). Under load, this leads to **Head-of-Line Blocking**, causing false positive task cancellations and system instability.

### Root Cause Analysis

The loop waits for the completion of one event handler before polling for the next event.

**Problematic Code Pattern (`src/meta/src/hummock/manager/timer_task.rs`):**

```rust
let event_stream = select_all(triggers); // Merges all timer streams
loop {
    // ⚠️ The loop is blocked here until the previous iteration finishes
    match event_stream.next().await {
        HummockTimerEvent::Report => {
            // 🛑 CRITICAL: This branch holds locks and performs IO/Calculation
            // If this takes 5 seconds, the loop halts for 5 seconds.
            let versioning_guard = hummock_manager.versioning.read().await;
            let configs = hummock_manager.get_compaction_group_map().await;
            // ...
        }
        HummockTimerEvent::CompactionHeartBeatExpiredCheck => {
            // ❌ STARVED: This event cannot fire until Report completes.
            // Even if scheduled for every 1s, it gets delayed by the Report duration.
        }
    }
}

```

### Visualizing the Blocking Behavior

The following Gantt chart illustrates how the current sequential architecture causes the liveness gaps observed in our tests:

```mermaid
gantt
    title Current Architecture: Sequential Processing (Starvation)
    dateFormat S
    axisFormat %S

    section Event Loop
    Select Next Event       :a1, 0, 0.1
    Handle Report (BLOCKING):crit, a2, 0.1, 5.1
    Select Next Event       :a3, 5.1, 5.2
    Handle Heartbeat        :a4, 5.2, 5.3
    
    section Real Time
    Heartbeat Due (T=1s)    :milestone, m1, 1, 0
    Heartbeat Due (T=2s)    :milestone, m2, 2, 0
    Heartbeat Due (T=3s)    :milestone, m3, 3, 0
    
    %% Result: Heartbeats at T=1,2,3,4,5 are MISSED because 'Handle Report' is occupying the thread.

```

### Error message/log

```text
No error infos
```

### 1. Minimal Reproducible Example (Rust Test)

I have implemented a standalone reproduction test. It sets up a high-frequency "Heartbeat" stream (10ms) and a low-frequency "Report" stream (100ms) that simulates a brief 50ms block.

<details>
<summary>Click to expand the Rust reproduction test code</summary>

```rust
#[cfg(test)]
mod starvation_test {
    use super::*;
    use std::sync::atomic::{AtomicU64, Ordering};
    use std::sync::Arc;
    use tokio::time::{Duration, Instant};
    use tokio::sync::RwLock;
    use futures::stream::{select_all, BoxStream, StreamExt};

    #[tokio::test]
    async fn test_timer_starvation() {
        let report_count = Arc::new(AtomicU64::new(0));
        let heartbeat_count = Arc::new(AtomicU64::new(0));
        let heartbeat_timestamps = Arc::new(tokio::sync::Mutex::new(Vec::new()));
        
        let report_count_clone = report_count.clone();
        let heartbeat_count_clone = heartbeat_count.clone();
        let heartbeat_timestamps_clone = heartbeat_timestamps.clone();
        let shared_lock = Arc::new(RwLock::new(0u64));
        let shared_lock_clone = shared_lock.clone();

        // High frequency heartbeat (10ms) vs Low frequency Report (100ms)
        let report_trigger = tokio_stream::wrappers::IntervalStream::new(
            tokio::time::interval(Duration::from_millis(100))
        ).map(|_| "Report");
        
        let heartbeat_trigger = tokio_stream::wrappers::IntervalStream::new(
            tokio::time::interval(Duration::from_millis(10))
        ).map(|_| "Heartbeat");

        let mut event_stream = select_all(vec![
            Box::pin(report_trigger) as BoxStream<'static, &str>,
            Box::pin(heartbeat_trigger) as BoxStream<'static, &str>,
        ]);

        let start = Instant::now();
        let test_duration = Duration::from_secs(1);

        while start.elapsed() < test_duration {
            if let Some(event) = event_stream.next().await {
                match event {
                    "Report" => {
                        report_count_clone.fetch_add(1, Ordering::SeqCst);
                        // Simulate blocking: acquire lock + computation for just 50ms
                        let _guard = shared_lock_clone.read().await;
                        std::thread::sleep(Duration::from_millis(50)); 
                    }
                    "Heartbeat" => {
                        heartbeat_count_clone.fetch_add(1, Ordering::SeqCst);
                        heartbeat_timestamps_clone.lock().await.push(start.elapsed());
                    }
                    _ => {}
                }
            }
        }

        let timestamps = heartbeat_timestamps.lock().await;
        let max_gap = timestamps.windows(2)
            .map(|w| w[1] - w[0])
            .max()
            .unwrap_or(Duration::from_millis(0));

        println!("Report events: {}", report_count.load(Ordering::SeqCst));
        println!("Heartbeat events: {}", heartbeat_count.load(Ordering::SeqCst));
        println!("Expected heartbeats: ~100 (every 10ms)");
        println!("Maximum gap between heartbeats: {:?}", max_gap);
        println!("Expected gap: ~10ms, Actual max gap shows starvation");
    }
}

```

</details>

### 2. Test Results & Analysis

Running the above test yields the following output:

```text
running 1 test
test is runner,hummock::manager::timer_task::starvation_test::test_timer_starvation
Report events: 11
Heartbeat events: 100
Expected heartbeats: ~100 (every 10ms)
Maximum gap between heartbeats: 33.807789ms
Expected gap: ~10ms, Actual max gap shows starvation
test is drop,hummock::manager::timer_task::starvation_test::test_timer_starvation
test hummock::manager::timer_task::starvation_test::test_timer_starvation ... ok

```

**Conclusion from Data:**

1. **Latency Spike (338%):** Even with a minimal simulated block (50ms), the heartbeat interval spiked from the expected **10ms** to **33.8ms**.
2. **Linear Scaling Risk:** In production, the `Report` task involves heavy iteration and lock contention that can take **seconds**. This 33ms gap scales linearly—a 5-second block in `Report` results in a 5-second gap in heartbeats, guaranteeing timeout failures (threshold is usually 1s).
3. **Jitter Confirmation:** The variability confirms that the `Heartbeat` stream is **not independent** but is strictly coupled to the completion of the `Report` stream's items.

## Concrete Impact

In a production environment where `Report` takes 5 seconds:

1. **Heartbeat Starvation:** `CompactionHeartBeatExpiredCheck` (interval: 1s) misses ~5 ticks entirely.
2. **False Positives:** Valid compaction tasks are incorrectly marked as expired and cancelled.
3. **Log Evidence:** Users observe warnings like:
```
[WARN] Heartbeat expired compaction tasks detected. Attempting to cancel tasks.
[ERROR] Attempt to remove compaction task due to elapsed heartbeat failed.

```

## Proposed Solution: Independent Task Architecture

To resolve the Head-of-Line blocking, we should refactor `hummock_timer_task` to spawn **independent Tokio tasks** for each timer.

```mermaid
gantt
    title Proposed Fix: Independent Tasks (No Starvation)
    dateFormat S
    axisFormat %S

    section Report Task
    Handle Report (BLOCKING):crit, a1, 0.1, 5.1
    
    section Heartbeat Task
    Handle Heartbeat (T=1s) :active, b1, 1, 1.1
    Handle Heartbeat (T=2s) :active, b2, 2, 2.1
    Handle Heartbeat (T=3s) :active, b3, 3, 3.1
    
    %% Result: Heartbeats fire on schedule regardless of Report task duration.

```

**Draft Implementation:**

```rust
pub fn hummock_timer_task(
    hummock_manager: Arc<Self>,
    backup_manager: Option<BackupManagerRef>,
) -> Vec<(JoinHandle<()>, Sender<()>)> {
    vec![
        // Spawn each as a separate task
        spawn_report_task(hummock_manager.clone()),
        spawn_heartbeat_task(hummock_manager.clone()),
        spawn_dead_task_check(hummock_manager.clone()),
        // ...
    ]
}

fn spawn_heartbeat_task(hummock_manager: Arc<HummockManager>) -> (JoinHandle<()>, Sender<()>) {
    let (shutdown_tx, mut shutdown_rx) = oneshot::channel();
    let handle = tokio::spawn(async move {
        let mut interval = tokio::time::interval(Duration::from_secs(1));
        interval.set_missed_tick_behavior(MissedTickBehavior::Skip);
        
        loop {
            tokio::select! {
                _ = interval.tick() => {
                    // Running in its own task, immune to 'Report' blocks
                    check_heartbeat_expired(&hummock_manager).await;
                }
                _ = &mut shutdown_rx => break,
            }
        }
    });
    (handle, shutdown_tx)
}

```

I run RisingWave on my x86_64 Linux Server 

### The version of RisingWave

Detected on `main` branch at 2026/02/02.

## 🔴 __Category A: Write Limits Circular Lock Dependency (Critical)__

__Core Function:__ `try_update_write_limits` in `src/meta/src/hummock/manager/versioning.rs`

__Affected Locations:__

- `versioning.rs:198:49` - Lock acquisition point
- `versioning.rs:199:68` - Secondary lock acquisition point

__Call Sites Creating Cycles:__

- `commit_epoch.rs:341:13` - Called during epoch commit
- `compaction/compaction_group_manager.rs:394:13` - Called during compaction group management
- `compaction/mod.rs:1045:13` - Called during compaction operations

__Problem:__ The same function `try_update_write_limits` is called from multiple locations (commit_epoch, compaction_group_manager, compaction module), and these different call chains acquire locks in inconsistent orders, creating circular dependencies.

__Impact:__ Can cause system hang during write limit updates, affecting write throughput control.

---

## 🟠 __Category B: Version Stats Access Contention (High Priority)__

__Core Function:__ `get_version_stats` in `src/meta/src/hummock/manager/versioning.rs`

__Affected Location:__

- `versioning.rs:188:32` - Lock acquisition point

__Call Sites Creating Cycles:__

- `barrier/context/context_impl.rs:52:12` - Called from barrier context
- `barrier/context/recovery.rs:777:48` - Called during recovery
- `dashboard/mod.rs:177:29` - Called from dashboard

__Problem:__ Concurrent access from barrier subsystem and dashboard creates circular lock dependencies when both try to read version statistics.

__Impact:__ Can block barrier operations and dashboard queries simultaneously.

---

## 🟠 __Category C: Version Operation Multi-Point Contention (High Priority)__

__Core Function:__ `on_current_version` in `src/meta/src/hummock/manager/versioning.rs`

__Affected Location:__

- `versioning.rs:156:35` - Lock acquisition point

__Call Sites Creating Cycles:__

- `barrier/context/recovery.rs:232:13` - Recovery initialization
- `barrier/context/recovery.rs:705:82` - Recovery checkpoint
- `barrier/context/recovery.rs:887:70` - Recovery finalization
- `backup_restore/backup_manager.rs:351:17` - Backup operations
- `versioning.rs:152:9` - Self-recursive call
- `versioning.rs:160:9` - Self-recursive call
- `compaction/mod.rs:1057:9` - Compaction operations
- `manager/diagnose.rs:551:36` - Diagnostic operations

__Problem:__ The function `on_current_version` is called from multiple critical subsystems (recovery, backup, compaction, diagnostics), and these different execution paths create complex circular dependencies.

__Impact:__ Can deadlock during recovery, backup, or compaction operations, affecting system availability.

---

## 🟡 __Category D: Initialization State Loading (Medium Priority)__

__Core Functions:__ `load_meta_store_state_impl` and `load_meta_store_state` in `src/meta/src/hummock/manager/mod.rs`

__Affected Locations:__

- `mod.rs:433:18` - First lock acquisition point
- `mod.rs:491:51` - Second lock acquisition point

__Call Sites:__

- `mod.rs:369:9` - `load_meta_store_state_impl` call
- `mod.rs:350:9` - `load_meta_store_state` call

__Problem:__ Circular dependency between `load_meta_store_state` and `load_meta_store_state_impl` during initialization, with locks held across yield points.

__Impact:__ Can cause deadlock during system startup or state reload.

---

## 🟡 __Category E: Compaction Group Scheduling (Medium Priority)__

__Core Functions:__

- `move_state_tables_to_dedicated_compaction_group` in `compaction/compaction_group_schedule.rs`
- `compaction_group_ids` in `compaction/compaction_group_manager.rs`

__Affected Locations:__

- `compaction_group_schedule.rs:626:59` - Lock in move_state_tables function
- `compaction_group_manager.rs:96:58` - Lock in compaction_group_ids function

__Call Sites Creating Cycles:__

- `compaction_group_schedule.rs:797:19` - First call to move_state_tables
- `compaction_group_schedule.rs:844:31` - Second call to move_state_tables
- `compaction/mod.rs:290:40` - First call to compaction_group_ids
- `compaction/mod.rs:307:40` - Second call to compaction_group_ids
- `timer_task.rs:590:22` - Timer task call to compaction_group_ids

__Problem:__ Dynamic compaction group scheduling operations create circular dependencies when moving tables and querying group IDs concurrently.

__Impact:__ Can block compaction group reorganization operations.

---

## 🟡 __Category F: Context Release Contention (Medium Priority)__

__Core Function:__ `release_contexts` in `src/meta/src/hummock/manager/context.rs`

__Affected Location:__

- `context.rs:97:58` - Lock acquisition point

__Call Sites Creating Cycles:__

- `context.rs:288:9` - Internal call
- `worker.rs:98:17` - Worker cleanup call

__Problem:__ Context release called from multiple locations creates circular lock dependency during resource cleanup.

__Impact:__ Can block context cleanup operations, potentially causing resource leaks.

---

## 🟢 __Category G: Test Code Issues (Low Priority)__

__Core Function:__ `report_compact_task_for_test` in `src/meta/src/hummock/manager/compaction/mod.rs`

__Affected Location:__

- `compaction/mod.rs:1359:53` - Lock acquisition point

__Call Sites:__

- `mock_hummock_meta_client.rs:350:37` - Mock client call
- `test_utils.rs:113:5` - Test utility call

__Problem:__ Test function has circular lock dependency.

__Impact:__ Only affects test environment, no production impact.

---

Li0k:

Fixes https://github.com/risingwavelabs/risingwave/issues/24675.

This PR refactors the Hummock timer task to avoid head-of-line blocking in the previous single-loop implementation. On `main`, all periodic triggers were driven by one `select_all(triggers)` loop and each selected handler was awaited inline. A slow handler such as statistics reporting could therefore delay the 1s compaction heartbeat expiration check and cause false worker liveness issues.

The new structure keeps the behavior aligned with the original timer logic, but splits the work into semantic lanes:

- `heartbeat`: runs `handle_timer_compaction_heartbeat_expired_check` independently, so liveness detection is not blocked by slower background work.
- `report`: runs statistics reporting independently. The reporting path now snapshots in-memory state while holding locks and emits metrics outside those locks.
- `scheduling`: a single serialized lane for all compaction scheduling related timers:
  - dynamic compaction,
  - space reclaim compaction plus vnode watermark compaction,
  - TTL reclaim compaction,
  - tombstone reclaim compaction,
  - compaction group split scheduling,
  - compaction group merge scheduling.
- `maintenance`: a serialized lane for low-frequency maintenance such as dead task checking and full GC.

The `scheduling` lane intentionally serializes compaction triggers and group split/merge scheduling together. This preserves the old main-loop mutual exclusion semantics for scheduling-related handlers and avoids introducing additional timer-vs-timer lock contention on `versioning` / `compaction_group_manager`. At the same time, the heartbeat and report lanes are isolated from that scheduling lane, which addresses the starvation issue from #24675.

Tests were also updated to cover the periodic-loop behavior and to assert that a serial timer lane does not run handlers concurrently.

Local validation:

```text
rustfmt --edition 2021 src/meta/src/hummock/manager/timer_task.rs src/meta/src/hummock/manager/compaction/compaction_group_manager.rs
git diff --check
cargo test -p risingwave_meta hummock::manager::timer_task --lib
cargo check -p risingwave_meta --lib
```

## Checklist

- [x] I have written necessary rustdoc comments, or no new rustdoc is needed for this internal refactor.
- [x] I have added/updated necessary unit tests.

## Documentation

No user-facing documentation update is needed. This is an internal meta-node timer-loop bug fix/refactor.

<details>

Fix a Hummock compaction timer-loop issue where slow periodic background handlers could delay compaction heartbeat expiration checks and affect compactor liveness detection.

</details>


src/meta/src/hummock/manager/timer_task.rs
```diff
@@ -202,296 +115,370 @@ impl HummockManager {
...
             }
+
+            tracing::info!("Hummock timer loop is stopped");
         });
         (join_handle, shutdown_tx)
     }
+
+    async fn handle_timer_report(&self) {
+        // Avoid holding the `versioning` lock across await points to prevent potential deadlocks.
+        let (current_version, version_stats) = {
```
zwang28, line 425: These clones seem unnecessary. They are also inconsistent with subsequent lines, where the versioning lock is re-acquired within the same method.

zwang28: To maintain consistency throughout the method—read either the cloned version object, or the original version object reference.

faketygg (commented):
Thanks for pushing this fix and the timer-loop tests. I went through it with our #24675 risk checklist and have one non-blocking design concern:

- This PR splits all timer handlers into independent loops. The original issue is specifically heartbeat starvation by heavy handlers (/etc.). A full split fixes it, but it also broadens concurrency surface (more interleavings and lock contention paths) beyond what is needed for the bug.

Could we consider a smaller-scope variant (extract only  into an independent loop, keep other handlers serialized), or alternatively add one more stress test that validates lock/contention behavior under mixed  +  + heartbeat workload?

I’m OK either way; I just want us to make the risk/perf tradeoff explicit before merge.

faketygg:
Correction to my previous comment (CLI escaped some identifiers): the specific handlers I meant are `Report`, `CompactionHeartBeatExpiredCheck`, and `Split/Merge`.

faketygg (commented):
Good refactor! Reviewed with focus on concurrency safety. Key observations:

1. **`split_merge_mutex`**: Smart use of `Arc<Mutex<()>>` to preserve the implicit mutual exclusion between `GroupScheduleSplit` and `GroupScheduleMerge` that existed in the original `select_all` loop. This prevents potential deadlocks from concurrent split/merge operations that both acquire `compaction.write() + versioning.write() + compaction_group_manager.write()`.

2. **Lock scope narrowing in `handle_timer_report`**: Releasing `versioning.read()` before calling `get_compaction_group_map()` reduces lock contention. Note that the second `versioning.read()` at the end is held until function return — this is fine since it's only a read lock.

3. **Lock scope narrowing in `check_dead_task`**: Same pattern + added `configs.get()` fallback with warning instead of panicking on missing group config — good defensive improvement.

4. **`spawn_periodic_loop` with shutdown preemption**: Clean abstraction. The inner `tokio::select!` allowing shutdown to cancel a running handler is important for graceful shutdown.

5. **Tests**: All three test cases (`isolated_progress`, `shared_mutex_guards`, `shutdown_preempts_handler`) cover the critical behavioral properties.

One minor note: `CompactionHeartBeatExpiredCheck` (1s period) now runs concurrently with all other handlers. Its `cancel_compact_tasks` path acquires `compaction.write() + versioning.write()`, which follows the global lock ordering (compaction → versioning), so no deadlock risk. The lock contention with split/merge is acceptable given split/merge's low frequency (600s+).

LGTM from a concurrency safety perspective.

faketygg (approved):
Reviewed from #24675 perspective. I left one non-blocking scope/risk comment; no blocker from me for merge.

faketygg (commented):
Nice refactoring! The `spawn_periodic_loop` abstraction with graceful shutdown and the `split_merge_mutex` for preserving Split/Merge mutual exclusion are well-designed.

One observation: with all handlers now running as independent tasks, `CompactionHeartBeatExpiredCheck`'s `cancel_compact_tasks()` path acquires `compaction.write()` + `versioning.write()`, which can now run concurrently with `Report`'s `versioning.read()`. The lock ordering is consistent (compaction → versioning), so no deadlock risk, but the write lock may need to wait for Report's read lock to release. Since Report's lock hold time is reduced by the refactoring (early drop of versioning guard), this should be acceptable in practice.

Overall LGTM from a concurrency safety perspective.

YZL0v3ZZ:
This PR seems to have been blocked for a long time.

Li0k:
> This PR seems to have been blocked for a long time.

Hi, Are you currently blocked? I'll continue to push this PR forward.

src/meta/src/hummock/manager/compaction/compaction_group_manager.rs
```diff
@@ -428,39 +429,56 @@ impl HummockManager {
...
-                infos.push(group_info);
-            }
-        };
+        for group_id in current_version.levels.keys() {
+            let Some(compaction_group_config) = id_to_config.get(group_id) else {
+                tracing::warn!(
+                    compaction_group_id = %group_id,
+                    "Missing compaction group config when calculating group statistics. Skip."
+                );
+                continue;
```
zwang28: Notes: Regardless of whether we use unwrap or continue when getting a group config, we must remain cautious: the calculate_compaction_group_statistic may return a subset of compaction groups to its callsite, which may lead to incorrectness in certain scenarios.
I recommend we add comments to clarify the correctness for all callsites of calculate_compaction_group_statistic.

zwang28 (approved):
Note: This refactor is expected to increase lock contention. For example, check_dead_task and handle_timer_report may now compete for the versioning lock concurrently, whereas previously only one could run at a time.

src/meta/src/hummock/manager/timer_task.rs
```diff
@@ -203,287 +198,339 @@ impl HummockManager {
...
+            for (name, handle) in child_handles {
+                if let Err(e) = handle.await {
+                    warn!(
+                        handler = name,
+                        error = %e.as_report(),
+                        "Hummock timer handler loop failed"
+                    );
                 }
             }
+
```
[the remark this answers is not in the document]
hzxa21: This is a valid comment.

hzxa21 (approved):
LGTM

PR #24781, opened 2026-02-11
fix: update heartbeat to include CPU and memory resource information for consistency

### LINKED ISSUE #24780: Resource (CPU/Memory) fields in rw_worker_nodes are 0 after meta pod restart until compute/compactor are restarted
### Problem

There is a reproducible issue with the order of pod startup in RisingWave clusters. If the meta pod is started after the compute or compactor pod (or if the meta pod is restarted), the `SELECT * FROM rw_worker_nodes;` query can show `0` for CPU/memory fields in compute/compactor pods. Only after restarting the compute/compactor pods do the correct resource values appear.

#### Where the resource values are collected:

- The logic for displaying resource data in `rw_worker_nodes` is implemented in [src/frontend/src/catalog/system_catalog/rw_catalog/rw_worker_nodes.rs](https://github.com/risingwavelabs/risingwave/blob/a2301a8edcf4c91c39366bc1894ca0e403e2b071/src/frontend/src/catalog/system_catalog/rw_catalog/rw_worker_nodes.rs) ([lines 24–100](https://github.com/risingwavelabs/risingwave/blob/a2301a8edcf4c91c39366bc1894ca0e403e2b071/src/frontend/src/catalog/system_catalog/rw_catalog/rw_worker_nodes.rs#L24-L100)):

```rust
// Table struct for rw_worker_nodes
struct RwWorkerNode {
    ...
    system_total_memory_bytes: Option<i64>,
    system_total_cpu_cores: Option<i64>,
    ...
}

// Data collected via meta_client.list_all_nodes()
async fn read_rw_worker_nodes_info(reader: &SysCatalogReaderImpl) -> Result<Vec<RwWorkerNode>> { ... }
```

- The `system_total_memory_bytes` and `system_total_cpu_cores` values are coming from the `worker.resource` field in the catalog, which in turn depends on what the compute/compactor pods report to meta.

- Compute node startup parameters and reporting are handled in [src/compute/src/server.rs](https://github.com/risingwavelabs/risingwave/blob/a2301a8edcf4c91c39366bc1894ca0e403e2b071/src/compute/src/server.rs), and compactor reporting in [src/storage/compactor/src/lib.rs](https://github.com/risingwavelabs/risingwave/blob/a2301a8edcf4c91c39366bc1894ca0e403e2b071/src/storage/compactor/src/lib.rs).

### Steps to Reproduce
1. Start/restart meta pod after compute or compactor pods are already running.
2. Run `SELECT * FROM rw_worker_nodes;`.
3. Observe that `system_total_memory_bytes` and `system_total_cpu_cores` are `0` for running compute/compactor pods.
4. Restart the affected compute/compactor pods, and the correct values appear.

### Expected Behavior
- Resource values should populate in `rw_worker_nodes` correctly even if meta pod is started/restarted after other pods.

### Suggested Area to Investigate
- Check how the worker nodes publish or refresh their resource info to meta, especially on meta restart.

---
#### References:
- [Resource struct definition and population](https://github.com/risingwavelabs/risingwave/blob/a2301a8edcf4c91c39366bc1894ca0e403e2b071/src/frontend/src/catalog/system_catalog/rw_catalog/rw_worker_nodes.rs#L24-L100)
- [Compute node server startup](https://github.com/risingwavelabs/risingwave/blob/a2301a8edcf4c91c39366bc1894ca0e403e2b071/src/compute/src/server.rs)
- [Compactor node server and resource options](https://github.com/risingwavelabs/risingwave/blob/a2301a8edcf4c91c39366bc1894ca0e403e2b071/src/storage/compactor/src/lib.rs)

### Environment
- RisingWave Version: 2.7.1
---

yezizp2012:

Fixes #24780 .

## Checklist

## Documentation

<details>

</details>


yezizp2012:
> Wait... Should we just refresh the resource info on worker heartbeat given that the resource info is only use for o11y purpose?

Yes, that also works. However, if without persistent, this information may still be empty during the meta restart to the node heartbeat. Register and re-register are both low-frequency operations, and changing the heartbeat RPC might have a slight impact on performance.

Emmm, it's acceptable for me. Let me try it with modifications to the heartbeat RPC interface.

hzxa21 (approved):
LGTM

PR #24784, opened 2026-02-11
chore: recommend user to use locality backfill

hzxa21:

Recommend user to use locality backfill if there are more than 1 locality provider in the plan.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/optimizer/mod.rs
+++ b/src/frontend/src/optimizer/mod.rs
@@ -623,21 +623,25 @@ impl LogicalPlanRoot {
             ).into());
         }
 
-        let locality_provider_count = LocalityProviderCounter::count(plan.clone());
-        if ctx.session_ctx().config().enable_locality_backfill() {
-            if locality_provider_count > 5 {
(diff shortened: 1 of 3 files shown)


src/frontend/src/optimizer/mod.rs
```diff
@@ -625,10 +623,21 @@ impl LogicalPlanRoot {
...
+            if locality_provider_count > 5 {
+                risingwave_common::license::Feature::LocalityBackfill.check_available()?
+            }
+        } else if locality_provider_count > 0
+            && risingwave_common::license::Feature::LocalityBackfill
+                .check_available()
+                .is_ok()
         {
-            risingwave_common::license::Feature::LocalityBackfill.check_available()?
+            ctx.warn_to_user(
```
chenzl25: Actually, locality provider would be created only if `enable_locality_backfill = true`, so when you see `locality_provider_count > 0`, it seems it is enabled.

hzxa21: Change to hook into try_enforce_locality_requirement for the check. PTAL

Issue #8084, opened 2023-02-21
Should we ban full outer join for streaming query?


chenzl25:
### Describe the bug

A null row from either left or right side produces the same row (null, null) to the downstream.
```
create table t (a int primary key);
insert into t values(null);
create materialized view v as select t1.* from t as t1 full join t as t2 on t1.a = t2.a; -- panic
```

chenzl25:
cc @yuhao-su

fuyufjh:
What's the cause of `panic`?

chenzl25:
> What's the cause of `panic`?

left side:  +[null]  --> Full Join -> +[null, null]
right side:  +[null]  --> Full Join -> +[null, null]

From the downstream's view, the full join operator inserts the same row twice.

yuhao-su:
The correct behavior should be:
left side: +[null] --> Full Join -> +[null, null]
left side: +[null] --> Full Join -> -[null, null] +[null, null]
It might be a bug

chenzl25:
> The correct behavior should be: left side: +[null] --> Full Join -> +[null, null] left side: +[null] --> Full Join -> -[null, null] +[null, null] It might be a bug

I don't think it is a bug. The batch query will output 2 rows instead of one row.

yuhao-su:
> I don't think it is a bug. The batch query will output 2 rows instead of one row.

You are right! I can't think of any easy way to fix this. Maybe we should ban it for now. cc. @st1page

st1page:
:innocent: in fact, I prefer to ban all null primary keys. The only origin of that is Materialized view with `GROUP BY K` where k could be null. But if there any users need that behavior?

BugenZhao:
Link to https://github.com/risingwavelabs/risingwave/issues/8059.

fuyufjh:
> 😇 in fact, I prefer to ban all null primary keys. The only origin of that is Materialized view with `GROUP BY K` where k could be null. But if there any users need that behavior?

Group by `CUBE` will generate a NULL group by default. :lark-cry:

fuyufjh:
I prefer to keep the full outer join. In batch query it's almost useless but I guess it might be more useful for stream joining stream. Just guess.

But I don't have any idea to fix it now 🤔

chenzl25:
[...]

PR #24792, opened 2026-02-12
fix(optimizer): filter_out_all_null_keys precidate with no stream keys

### LINKED ISSUE #24790: Bug: FULL OUTER JOIN on global-aggregate MVs (no GROUP BY) returns empty result
### Describe the bug

A downstream materialized view using `FULL OUTER JOIN` can return an empty result when both join inputs are materialized views built from global aggregation (no `GROUP BY`).

This is unexpected because both sides produce one row and the join keys match.

### Error message/log

```text

```

### To Reproduce

```sql
drop table if exists t;
create table t (
  id int primary key,
  key string,
  amount int
);

insert into t values
  (1, 'k', 100),
  (2, 'k', 200);

-- two global-agg MVs (no GROUP BY)
create materialized view t1_v as
select sum(amount) as amount, max(key) as key
from t
where id = 1;

create materialized view t2_v as
select sum(amount) as amount, max(key) as key
from t
where id = 2;

-- downstream FULL OUTER JOIN MV
create materialized view t_stat as
select
  coalesce(a.key, b.key) as key,
  coalesce(a.amount, 0) as amount_1,
  coalesce(b.amount, 0) as amount_2
from t1_v a
full outer join t2_v b
  on a.key = b.key;

select * from t_stat;
```

### Expected behavior

Expected behavior

```
 key | amount_1 | amount_2
-----+----------+----------
 k   |      100 |      200
```

Actual behavior
```
 key | amount_1 | amount_2
-----+----------+----------
(0 rows)
```

_No response_

### The version of RisingWave

<=2.7.2

### Additional context

_No response_

qccash:

closes https://github.com/risingwavelabs/risingwave/issues/24790

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/optimizer/plan_node/logical_filter.rs
+++ b/src/frontend/src/optimizer/plan_node/logical_filter.rs
@@ -66,6 +66,11 @@ impl LogicalFilter {
 
     /// Create a `LogicalFilter` to filter out rows where all keys are null.
     pub fn filter_out_all_null_keys(input: PlanRef, key: &[usize]) -> PlanRef {
+        // No key means no "all-null key" row to filter. `ExprImpl::or([])` is `false`,
+        // which would incorrectly drop all rows.
+        if key.is_empty() {
+            return input;
+        }
         let schema = input.schema();
         let cond = ExprImpl::or(key.iter().unique().map(|&i| {
             FunctionCall::new_unchecked(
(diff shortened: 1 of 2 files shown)


src/frontend/src/optimizer/plan_node/logical_filter.rs
```diff
@@ -66,6 +66,11 @@ impl LogicalFilter {
 
     /// Create a `LogicalFilter` to filter out rows where all keys are null.
     pub fn filter_out_all_null_keys(input: PlanRef, key: &[usize]) -> PlanRef {
+        // No key means no "all-null key" row to filter. `ExprImpl::or([])` is `false`,
+        // which would incorrectly drop all rows.
+        if key.is_empty() {
+            return input;
+        }
```
BugenZhao, line 73: If we simply skip filtering for this case, it seems that we will encounter the duplicated stream key issue in #8084 again, if the join key is `NULL`.

You can reproduce with:
```sql
create table t (a int, b int);
create materialized view mv as select max(a) ma, max(b) mb from t;
insert into t values (1, NULL);
flush;
select * from mv;
create materialized view mvj as select mv1.ma as ma1 from mv as mv1 full outer join mv as mv2 on mv1.mb = mv2.mb;
```

I'm thinking that whether we should also consider the `join key` here for filtering, i.e., change the filtering keys from `concatenated stream key of two sides` into `stream key of the join output`.

cc @chenzl25 for discussion

BugenZhao (commented):
Would you also add planner tests and e2e tests for the reproduction query?

src/frontend/src/optimizer/plan_node/logical_filter.rs
chenzl25: I think we just need to add join key together to the filter_out_all_null_keys argument.

chenzl25: I open a PR here https://github.com/risingwavelabs/risingwave/pull/24899

PR #24798, opened 2026-02-12
fix(meta): narrow reschedule blocking for creating backfill jobs

shanicky:

This PR narrows reschedule blocking during creating backfill jobs to only truly related jobs/fragments instead of blocking all creating jobs.
It builds initial blocked fragments from creating jobs by checking unreschedulable online scan fragments and their upstream fragments, then expands through no-shuffle connectivity, maps to blocked job ids, and rejects only reschedules that actually touch those blocked jobs.
It also fixes an over-strict regression where every creating job could be treated as blocked, which made `ALTER ... SET PARALLELISM` fail even when `streaming_use_snapshot_backfill=false`.
Simulation coverage is included for concurrent background MV backfill parallelism switching, and this fix is validated with targeted simulation runs.

## Checklist

### DIFF
--- a/src/meta/src/barrier/checkpoint/creating_job/mod.rs
+++ b/src/meta/src/barrier/checkpoint/creating_job/mod.rs
@@ -33,7 +33,7 @@ use risingwave_pb::id::{ActorId, FragmentId};
 use risingwave_pb::stream_plan::barrier::PbBarrierKind;
 use risingwave_pb::stream_plan::barrier_mutation::Mutation;
 use risingwave_pb::stream_plan::stream_node::NodeBody;
-use risingwave_pb::stream_plan::{AddMutation, StopMutation, StreamScanType};
+use risingwave_pb::stream_plan::{AddMutation, StopMutation};
 use risingwave_pb::stream_service::BarrierCompleteResponse;
 use risingwave_pb::stream_service::streaming_control_stream_response::ResetPartialGraphResponse;
 use status::CreatingStreamingJobStatus;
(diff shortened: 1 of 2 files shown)


src/meta/src/stream/scale.rs
```diff
@@ -1107,11 +1114,29 @@ impl GlobalStreamManager {
...
             .catalog_controller
             .get_job_parallelisms(job_id)
-            .await?;
+            .await
+        {
+            Ok(parallelisms) => parallelisms,
+            Err(e)
+                if matches!(
+                    e.inner(),
+                    crate::error::MetaErrorInner::CatalogIdNotFound("streaming job", _)
```
shanicky: any better idea?

src/meta/src/stream/scale.rs
```diff
@@ -903,20 +904,23 @@ impl GlobalStreamManager {
...
+            // Retry periodically while some jobs are temporarily blocked by creating
+            // unreschedulable backfill jobs. This allows us to scale them
+            // automatically once the creating jobs finish.
+            return Ok(has_blocked_jobs);
         }
 
         let active_workers =
             ActiveStreamingWorkerNodes::new_snapshot(self.metadata_manager.clone()).await?;
 
         if job_ids.is_empty() {
```
wenym1: The two checks on `job_ids.is_empty()` seems duplicated.

src/meta/src/barrier/checkpoint/creating_job/mod.rs
```diff
@@ -82,6 +84,33 @@ pub(crate) struct CreatingStreamingJobControl {
     upstream_lag: LabelGuardedIntGauge,
 }
 
+fn fragment_has_online_unreschedulable_scan(fragment: &InflightFragmentInfo) -> bool {
+    let mut has_unreschedulable_scan = false;
+    visit_stream_node_cont(&fragment.nodes, |node| {
+        if let Some(NodeBody::StreamScan(stream_scan)) = node.node_body.as_ref() {
+            let scan_type = stream_scan.stream_scan_type();
+            if scan_type != StreamScanType::Unspecified && !scan_type.is_reschedulable(true) {
```
wenym1: The check on `!= StreamScanType::Unspecified` seems unnecessary.

Issue #24794, opened 2026-02-12
Expose cross-db changelog lag related metrics for cross-db stream scan


qccash:
### Is your feature request related to a problem? Please describe.

When creating a materialized view(sink) from a cross-database table, the source relies on a subscription with retention.
If stream consumption falls behind beyond the retention window, consumption can fail.
We need first-class observability to track lag after backfill is finished, during normal stream scan.

### Describe the solution you'd like

Expose two metrics:

1. **Meta-side metric (per table)**
- `rw_crossdb_changelog_min_epoch{table_id}`
- Meaning: the minimum available epoch in table change logs (within retention window / truncate boundary approximation).

2. **Compute-side metric (per consumer execution unit)**
- `rw_crossdb_last_consumed_min_epoch{table_id,actor_id,fragment_id}`
- Meaning: the minimum consumed epoch for the cross-db stream scan progress (reflecting the slowest frontier).

External monitoring can join by `table_id` and compute headroom/risk.

hzxa21:
@wenym1

PR #24802, opened 2026-02-13
feat: add metrics for cross database consuming epoch

qccash:

closes https://github.com/risingwavelabs/risingwave/issues/24794

## Checklist

## Documentation

<details>

</details>


src/meta/src/rpc/metrics.rs
```diff
@@ -557,6 +559,14 @@ impl MetaMetrics {
         )
         .unwrap();
 
+        let crossdb_changelog_min_epoch = register_int_gauge_vec_with_registry!(
+            "crossdb_changelog_min_epoch",
```
wenym1: Table change log is not dedicated for cross db query. May change  the name to `table_change_log_min_epoch`

src/meta/src/hummock/metrics_utils.rs
```diff
@@ -507,6 +507,11 @@ pub fn trigger_gc_stat(
             .table_change_log_object_size
             .with_label_values(&[&format!("{table_id}")])
             .set(object_size as _);
+        let min_epoch = logs.epochs().min().unwrap_or_default();
```
wenym1, line 512: May generate `table_id_string` before the code to set `table_change_log_object_count`, and then always reuse the table_id_string.

qccash: Updated, added table_id_label and labels, reuse labels

src/meta/src/rpc/metrics.rs
qccash: Rename to storage_table_change_log_min_epoch.

qccash:
> Thanks for the contribution!
> 
> Can you help to add the two metrics to our grafana dashboard?
> 
> You may add table_change_log_min_epoch to the following section in `hummock_manager.py`.
> 
> ```
> panels.timeseries_epoch(
>                     "Epoch",
>                     "",
>                     [
>                         panels.target(
>                             f"{metric('storage_max_committed_epoch')}",
>                             "max committed epoch",
>                         ),
>                         panels.target(
>                             f"{metric('storage_min_committed_epoch')}",
>                             "min committed epoch",
>                         ),
>                     ],
>                 ),
> ```
> 
> and the other in `streaming_backfill.py`.

I think we don't need to add storage_table_change_log_min_epoch to Epoch panel, but will add a new panel 'Table Change Log Min Epoch'.

wenym1:
> I think we don't need to add storage_table_change_log_min_epoch to Epoch panel, but will add a new panel 'Table Change Log Min Epoch'.

Both LGTM.

Issue #24825, opened 2026-02-18
flaky e2e alter_connector_props_non_shared_source


yuhao-su:
https://buildkite.com/risingwavelabs/pull-request/builds/92134#019c7267-94f7-479f-b055-6123e2c2a92e

```

e2e_test/source_inline/kafka/alter/alter_connector_props_non_shared_source.slt .. [FAILED] after 10539 ms
--
 
failed to run `e2e_test/source_inline/kafka/alter/alter_connector_props_non_shared_source.slt`
 
Caused by:
statement failed: db error: ERROR: Failed to run the query
 
Caused by these errors (recent errors listed first):
1: gRPC request to meta service (call `/ddl_service.DdlService/CreateSource`) failed: Internal error
2: The cluster is recovering
3: Foreign Key Constraint Violated: FOREIGN KEY constraint failed
 
[SQL] create source non_shared_conn_source (x int) with (
connector = 'kafka',
connection = kafka_conn_non_shared,
topic = 'test_alter_non_shared_source_conn'
) format plain encode json;
at e2e_test/source_inline/kafka/alter/alter_connector_props_non_shared_source.slt:160

```

yuhao-su:
https://buildkite.com/risingwavelabs/pull-request/builds/92147#019c74f0-3aac-4de2-9e19-ef8b8b604f42
Another one

```

e2e_test/source_inline/kafka/alter/rate_limit_source_kafka_shared.slt.serial .. [FAILED] after 10389 ms
--
 
failed to run `e2e_test/source_inline/kafka/alter/rate_limit_source_kafka_shared.slt.serial`
 
Caused by:
query result mismatch:
[SQL] select * from rl_mv1;
[Diff] (-expected\|+actual)
-   1000
+   1250
at e2e_test/source_inline/kafka/alter/rate_limit_source_kafka_shared.slt.serial:52


```

tabVersion:
> [buildkite.com/risingwavelabs/pull-request/builds/92147#019c74f0-3aac-4de2-9e19-ef8b8b604f42](https://buildkite.com/risingwavelabs/pull-request/builds/92147#019c74f0-3aac-4de2-9e19-ef8b8b604f42) Another one
> 
> ```
> 
> e2e_test/source_inline/kafka/alter/rate_limit_source_kafka_shared.slt.serial .. [FAILED] after 10389 ms
> --
>  
[...]

PR #24834, opened 2026-02-20
test: fix kafka rate limit flaky test

tabVersion:

## Whats changed and whats your intention?

This PR is now scoped to the flaky SQLLogicTest only:

- `e2e_test/source_inline/kafka/alter/rate_limit_source_kafka_shared.slt.serial`

The flaky behavior came from timing-sensitive checks (`sleep` + immediate assert) and replay-sensitive counting (`count(*)`) on a shared Kafka source.

Changes:

- Replace fixed waits with `retry ... backoff ...` assertions for eventual consistency.
- Use `count(distinct v1)` instead of `count(*)` so replayed records do not cause false negatives.
- Insert a non-overlapping second batch (`1001..2000`) to ensure post-alter assertions reflect new data only.
- Add an explicit check for resetting `source_rate_limit` to `default`.

Stable invariants checked by this test:

- With `source_rate_limit = 0`, `rl_mv1` stays at `1000` after inserting the second batch.
- After setting `source_rate_limit = 1000`, `rl_mv1` reaches `2000`.
- `rw_rate_limit` reflects the expected SOURCE/SOURCE_BACKFILL limits during each alter step.
- After `set source_rate_limit to default`, the SOURCE rate-limit entry disappears while `rl_mv2` backfill limit remains.

## Checklist

## Documentation

<details>

N/A

</details>

### DIFF
--- a/e2e_test/source_inline/kafka/alter/rate_limit_source_kafka_shared.slt.serial
+++ b/e2e_test/source_inline/kafka/alter/rate_limit_source_kafka_shared.slt.serial
@@ -68,9 +68,9 @@ SET BACKGROUND_DDL=true;
 statement ok
 SET BACKFILL_RATE_LIMIT=0;
 
-statement ok
 # Shared Kafka source can replay records during recovery/reconfiguration.
 # Use DISTINCT here so this test validates rate-limit behavior instead of replay multiplicity.
+statement ok
 create materialized view rl_mv2 as select count(distinct v1) from kafka_source;
 
 sleep 1s


e2e_test/source_inline/kafka/alter/rate_limit_source_kafka_shared.slt.serial
```diff
@@ -72,17 +68,19 @@ SET BACKGROUND_DDL=true;
 statement ok
 SET BACKFILL_RATE_LIMIT=0;
 
+# Shared Kafka source can replay records during recovery/reconfiguration.
```
xiangjinwu, line 71: 1. Is this expected behavior?
2. Why was there recovery/reconfiguration during the occasional ci failure?

It is true these changes would alleviate the symptoms (less ci failure), but I am not sure whether we are just hiding the real problem deeper.

tabVersion: No, it is not expected. I also observed the same backtrace in other tests, ie. irrelevant to the kafka test itself. You can check details here https://github.com/risingwavelabs/risingwave/issues/24825#issuecomment-3943228474 . I think it is more like a meta bug. 

For the flaky test itself, because we run tests on hotspot EC2, the alter rate limit may not take effect on the time exactly what we want. So it can produce more data than expected, leading to test failure.

xiangjinwu: I am asking about the duplicate rows due to replay, while the linked comment is about another issue (meta foreign key constraint).

> the alter rate limit may not take effect on the time exactly what we want

This seems a third topic and not observed in tests, no matter before or after this PR.

tabVersion: agree, I am also investigating the replay issue. Seems there is some flaky scheduling on CI environment when transferring from backfill stage to forward stage.

tabVersion: The meta part is resolved by https://github.com/risingwavelabs/risingwave/pull/24887

Issue #24829, opened 2026-02-19
source backfill cannot complete when consuming from an idle Kafka partition ending in a control record


jgraettinger:
Author's note: :wave: I'm the CTO at [estuary.dev](https://estuary.dev/). We have a mutual customer who encountered a hang in RisingWave when attempting to create a materialized view from an Estuary collection using [dekaf](https://docs.estuary.dev/guides/dekaf_reading_collections_from_kafka/). We've traced it back to what appears to be a legitimate bug in the handling of Kafka control records, where RisingWave never considers a partition backfill "complete" during creation DDL because rdkafka doesn't surface control records to the application, despite their being accounted for in the high watermark. I used AI to generate this report, but have reviewed it in detail and confirmed the reproductions work as described:

## Describe the bug

Source backfill can never complete when consuming from a transactional Kafka topic whose last record in a partition is a transaction control record (commit/abort marker).

RisingWave's Kafka consumer is configured with `isolation.level=read_committed` ([`src/connector/src/source/kafka/mod.rs:224`](https://github.com/risingwavelabs/risingwave/blob/main/src/connector/src/source/kafka/mod.rs#L224)). Under this isolation level, the consumer never delivers [Kafka control records](https://kafka.apache.org/documentation/#controlrecord) (transaction commit/abort markers) to the application — they are silently skipped by librdkafka. However, these control records still occupy offsets in the partition log and are included in the broker-reported `high_watermark`.

The source backfill executor uses `high_watermark - 1` as the `target_offset` — the offset it must reach before considering backfill complete ([`src/connector/src/source/kafka/source/reader.rs:135-141`](https://github.com/risingwavelabs/risingwave/blob/main/src/connector/src/source/kafka/source/reader.rs#L135-L141)):

```rust
backfill_info.insert(
    split.id(),
    BackfillInfo::HasDataToBackfill {
        latest_offset: (high - 1).to_string(),
    },
);
```

When the last record in a partition is a control record at offset `N`:
- `high_watermark` = `N + 1`
- `target_offset` = `N`
- The last *consumable* message has an offset `< N`
- Backfill compares the last consumed offset against `target_offset` and will never find it `>=` ([`src/stream/src/executor/source/source_backfill_executor.rs:234-235`](https://github.com/risingwavelabs/risingwave/blob/main/src/stream/src/executor/source/source_backfill_executor.rs#L234-L235))
- Backfill hangs indefinitely

[...]

PR #24835, opened 2026-02-20
fix(kafka): unblock backfill on control-record tails

tabVersion:

This PR fixes Kafka source backfill hanging when a transactional partition ends with an unseen control record (issue #24829). It introduces a backfill-only reader option that enables `enable.partition.eof`, keeps normal readers unchanged, and computes backfill target offsets from the last readable message instead of `high_watermark - 1`. The Kafka reader now treats `PartitionEOF` as a control signal while probing/streaming in backfill mode, so backfill can finish without consuming uncommitted records. It also adds a source_inline Kafka regression test (`e2e_test/source_inline/kafka/issue_24829.slt`) with a transactional producer script to verify the MV creation no longer hangs.

## Checklist

## Documentation

<details>

Kafka source backfill no longer hangs when a transactional Kafka partition ends with commit/abort control records that are invisible under `read_committed`.

</details>

### DIFF
--- a/e2e_test/source_inline/kafka/issue_24829.slt
+++ b/e2e_test/source_inline/kafka/issue_24829.slt
@@ -51,6 +51,7 @@ drop materialized view txn_mv_24829;
 # consumer. Without the fix, the batch scan also hangs indefinitely because the
 # consumer waits for a message at stop_offset - 1 (a control record).
 
+skipif madsim
 query I
 select count(*) from txn_src_24829;
 ----


e2e_test/streaming/tpch/views/q12.slt.part
```diff
@@ -1,4 +1,4 @@
-statement ok
+statement ok retry 3 backoff 5s
```
xiangjinwu: Avoid unrelated change in the PR

e2e_test/streaming/tpch/views/q11.slt.part
```diff
@@ -1,4 +1,4 @@
-statement ok
+statement ok retry 3 backoff 5s
```
xiangjinwu: Avoid unrelated change in the PR

e2e_test/source_inline/kafka/tpch.slt
```diff
@@ -3,6 +3,46 @@ include ../../tpch/create_sources_kafka.slt.part
 
 control substitution off
 
+query I retry 20 backoff 3s
```
xiangjinwu: Avoid unrelated change in the PR

e2e_test/source_inline/kafka/alter/rate_limit_source_kafka_shared.slt.serial
```diff
@@ -104,7 +104,7 @@ rl_mv2	SOURCE_BACKFILL	{SOURCE_SCAN}	0
 
 sleep 3s
 
-query I
+query I retry 10 backoff 2s
```
xiangjinwu: Avoid unrelated change in the PR

e2e_test/source_inline/connection/ddl.slt
```diff
@@ -87,6 +81,12 @@ create sink sink_kafka from data_table with (
   force_append_only='true'
 );
 
+statement ok
```
xiangjinwu: Avoid unrelated change in the PR.

e2e_test/source_inline/kafka/issue_24829.slt
```diff
@@ -0,0 +1,10 @@
+control substitution on
+
+system ok
+rpk topic create test_topic_issue_24829;
+
+system ok
+python e2e_test/source_inline/kafka/kafka_txn_backfill_issue_24829.py --db-name $__DATABASE__ --topic test_topic_issue_24829
+
+system ok
+rpk topic delete test_topic_issue_24829;
```
xiangjinwu, line 64: This is against the design of slt. `CREATE SOURCE` and `CREATE MATERIALIZED VIEW` shall be individual statements in slt, and only resort to external script for preparing the testing environment (populating the kafka topic with transactional record in this case).

src/batch/executors/src/executor/source.rs
```diff
@@ -158,7 +158,13 @@ impl SourceExecutor {
         ));
         let (stream, _) = self
             .source
-            .build_stream(Some(self.split_list), self.column_ids, source_ctx, false)
+            .build_stream(
+                Some(self.split_list),
+                self.column_ids,
+                source_ctx,
+                false,
+                false,
```
xiangjinwu: Is batch source also supposed to stop-on-eof like backfill?

src/connector/src/source/kafka/mod.rs
```diff
@@ -180,6 +180,9 @@ pub struct KafkaProperties {
 
     #[serde(flatten)]
     pub unknown_fields: HashMap<String, String>,
+
+    #[serde(skip)]
```
xiangjinwu: How is this `serde` used, and why should it skip this property?

src/connector/src/source/base.rs
```diff
@@ -728,15 +732,11 @@ impl ConnectorProperties {
...
-            splits,
-            parser_config,
-            source_ctx,
-            columns,
-            opt
-        )
-        .await)
+        dispatch_source_prop!(self, |prop| {
+            let mut prop = prop;
+            <PropType as SourceProperties>::init_from_create_split_reader_opt(&mut *prop, &opt);
```
xiangjinwu: Why not `prop.init_from_create_split_reader_opt`?

src/connector/src/source/base.rs
```diff
@@ -123,6 +123,9 @@ pub trait SourceProperties:
 
     /// Load additional info from `ExternalTableDesc`. Currently only used by CDC.
     fn init_from_pb_cdc_table_desc(&mut self, _table_desc: &ExternalTableDesc) {}
+
+    /// Load extra options before creating split readers.
+    fn init_from_create_split_reader_opt(&mut self, _opt: &CreateSplitReaderOpt) {}
```
xiangjinwu: Extending this way can be error-prone. The caller may forget to call it when they have both `prop` and `opt`, and callee may forget to implement it. But that may be a larger refactor so I am not suggesting anything in this PR.

src/connector/src/with_options_test.rs
```diff
@@ -157,7 +157,11 @@ fn generate_with_options_yaml_inner(path: &Path) -> BTreeMap<String, StructInfo>
                     default_func,
                     rename,
                     alias,
+                    skip_deserializing,
                 } = extract_serde_properties(&field);
+                if skip_deserializing {
+                    continue;
```
xiangjinwu: Again, why omit the option in yaml when it is omitted by serde?

src/batch/executors/src/executor/source.rs
tabVersion: I think yes, batch scan is scanning from some timestamp to the latest, then stop. So it should stop when reaching the end.

src/connector/src/source/kafka/mod.rs
tabVersion: enable_partition_eof is supposed to control internally, and do not want users to config it.
We only allow `enable_partition_eof` in source backfill exec (and batch exec), otherwise will be disabled.

xiangjinwu: Oh makes sense. This property is not user configurable.

Is `serde` on properties solely for the yaml generation? And `java_catalog_props` also hidden from user?

src/batch/executors/src/executor/source.rs
xiangjinwu: Maybe we shall also add it to the test.

tabVersion: Yes. Batch source has the same hang problem: `stop_offset` is set to `high_watermark` by the enumerator, so when the tail offsets are control records, `cur_offset == stop_offset - 1` can never be satisfied and the batch query hangs.

Fixed in 4d30819955: set `for_backfill: true` for batch executor and use the `PartitionEOF` signal to remove completed partitions from `stop_offsets` in `into_data_stream()`. The streaming path is unaffected because streaming splits have `stop_offset = None`, so the `stop_offsets` map is empty and `remove()` returns `None`.

tabVersion:
replace with https://github.com/risingwavelabs/risingwave/pull/25121
close as found another approach

Issue #23763, opened 2025-11-13
Feature request for make_interval()


karenlo-tt:
### Is your feature request related to a problem? Please describe.

Currently, there is no easy way to create an interval from integer values. The only option that I have found is to first cast those integers into strings.

For example, if I wanted to create an interval of 10 days, I need to run something like:
```
select format('%s day', some_col)::interval
``` 

### Describe the solution you'd like

Ideally, we would support the solution provided by Postgres:
```
make_interval(days => some_col)
```

As specified here: https://www.postgresql.org/docs/current/functions-datetime.html

james-johnston-thumbtack:
This still seems relevant.

hezijiangjiang:
Hi! I'd like to work on implementing `make_interval()` (starting with `days => int`) in this issue. Could you assign it to me?

PR #24837, opened 2026-02-23
feat(expr): support make_interval(days => int)

hezijiangjiang:
## Summary
Implement a minimal day-based `make_interval`:
- add scalar function `make_interval(int4) -> interval`
- maps input to `Interval::from_month_day_usec(0, days, 0)`
- add unit tests for positive and negative day values

## Important scope note
RisingWave parser supports named function-arg syntax (`arg => value`), but binder currently has `FunctionArg::Named => todo!` in `src/frontend/src/binder/expr/function/mod.rs`.
So this PR only introduces positional form for now (`make_interval(10)`).

## Related issue
Part of #23763

## Tests
Added unit test in `make_time.rs` (`test_make_interval_days`).

> I could not run `cargo test` in this environment because Rust toolchain (`cargo`) is not installed on the host.

### DIFF
--- a/src/expr/impl/src/scalar/make_time.rs
+++ b/src/expr/impl/src/scalar/make_time.rs
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 use chrono::{NaiveDate, NaiveDateTime, NaiveTime};
-use risingwave_common::types::{Date, F64, FloatExt, Time, Timestamp, Timestamptz};
+use risingwave_common::types::{Date, F64, FloatExt, Interval, Time, Timestamp, Timestamptz};
 use risingwave_expr::expr_context::TIME_ZONE;
 use risingwave_expr::{ExprError, Result, capture_context, function};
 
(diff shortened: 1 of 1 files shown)


hezijiangjiang:
I did a second pass review on this PR and found an important compatibility gap: named function arguments are currently not bound in frontend binder ().\n\nSo this PR is intentionally scoped to positional usage () as an incremental step for #23763. If maintainers prefer strict PostgreSQL-style named syntax (), we likely need binder support first (or in parallel).

hezijiangjiang:
Follow-up clarification: this PR currently supports positional call style only (example: make_interval(10)). Named-argument style (example: make_interval(days => 10)) is not wired in binder yet because FunctionArg::Named is still TODO in frontend binder code.

hezijiangjiang:
Closing this initial attempt because it had a PostgreSQL compatibility gap (positional semantics mismatch). Re-opening with a corrected implementation and tests in a new PR.

PR #24852, opened 2026-02-24
feat: meta generalize compaction-group normalize with batched split deltas

Li0k:
## Summary

Generalize compaction-group normalize into an incremental plan/apply loop that revalidates one split per step against the latest version, runs before both split and merge scheduling, and supports a configurable per-round split budget.

## What's changed

- **Normalize execution model** (`compaction_group_schedule.rs`): build one normalize plan from a read snapshot, revalidate it under the write lock, and apply one split in a short transaction. Repeat until there is no more overlap in the current schedulable segment or the per-round limit is reached. Expired parent-group compaction tasks are cancelled after each successful normalize split.
- **Scheduler integration** (`timer_task.rs`): invoke normalize before both split scheduling and merge scheduling so overlap repair still works when the regular split timer is disabled.
- **Config and docs**: add `max_normalize_splits_per_round` (default `4`), wire it through meta config, and clarify that `periodic_scheduling_compaction_group_split_interval_sec` only controls the regular split timer. When `enable_compaction_group_normalize` is enabled, normalize may still emit split deltas before split/merge scheduling.
- **Tests** (`tests.rs`, `timer_task.rs`): add coverage for cascading normalize, per-round limiting, disabled boundaries, expired task cancellation, merge-path normalize when split scheduling is disabled, and stronger membership/non-overlap assertions.

## Motivation

Overlapping compaction-group table-id ranges still need automatic repair, but doing all normalize splits in one long write transaction is brittle. This change makes normalize incremental, revalidates each step against the latest state, and bounds the amount of normalize work done in one scheduler round while preserving the current split-only repair model within schedulable segments.

### DIFF
--- a/src/meta/src/hummock/manager/compaction/compaction_group_schedule.rs
+++ b/src/meta/src/hummock/manager/compaction/compaction_group_schedule.rs
@@ -999,9 +999,11 @@ impl HummockManager {
             let mut version = HummockVersionTransaction::new(
                 &mut versioning.current_version,
                 &mut versioning.hummock_version_deltas,
+                &mut versioning.table_change_log,
                 self.env.notification_manager(),
                 None,
                 &self.metrics,
+                &self.env.opts,
             );
             let mut new_version_delta = version.new_delta();
             let split_key = plan.split_key();
--- a/src/meta/src/hummock/manager/tests.rs
+++ b/src/meta/src/hummock/manager/tests.rs
@@ -187,7 +187,6 @@ async fn setup_compute_env_with_meta_opts(
             Property {
                 is_streaming: true,
                 is_serving: true,
-                is_unschedulable: false,
                 parallelism: 4,
                 ..Default::default()
             },
@@ -3098,7 +3097,7 @@ async fn test_normalize_overlapping_compaction_groups_cancels_expired_compact_ta
         .unwrap();
 
     let task = hummock_manager
-        .get_compact_task(cg_64, &mut default_compaction_selector())
+        .get_compact_task(cg_64, &mut *default_compaction_selector())
         .await
         .unwrap()
         .expect("should produce a compaction task");
--- a/src/meta/src/hummock/manager/timer_task.rs
+++ b/src/meta/src/hummock/manager/timer_task.rs
@@ -764,7 +764,6 @@ mod tests {
                 Property {
                     is_streaming: true,
                     is_serving: true,
-                    is_unschedulable: false,
                     parallelism: 4,
                     ..Default::default()
                 },


src/meta/src/hummock/manager/compaction/compaction_group_schedule.rs
```diff
@@ -740,6 +740,268 @@ impl HummockManager {
...
+                }
+
+                groups.sort_by_key(|group| group.table_ids[0]);
+
+                let mut split_candidate = None;
+                for pair in groups.windows(2) {
+                    let left = &pair[0];
+                    let right = &pair[1];
+
+                    if left.table_ids.len() <= 1 || right.table_ids.is_empty() {
```
zwang28: right.table_ids.is_empty() is unreachable?

src/meta/src/hummock/manager/compaction/compaction_group_schedule.rs
```diff
@@ -740,6 +740,268 @@ impl HummockManager {
...
+                self.env.notification_manager(),
+                None,
+                &self.metrics,
+            );
+
+            loop {
+                let mut groups = vec![];
+                let latest_version = version.latest_version();
+                for group_id in latest_version.levels.keys() {
+                    let table_ids = latest_version
```
zwang28: What about retaining the original BTreeSet type for table_ids? Otherwise, the ordering of table_ids is not explicitly enforced.

src/meta/src/hummock/manager/compaction/compaction_group_schedule.rs
```diff
@@ -740,6 +740,268 @@ impl HummockManager {
...
+                let split_full_key =
+                    group_split::build_split_full_key(boundary_table_id, VirtualNode::ZERO);
+                let (table_ids_left, table_ids_right) =
+                    group_split::split_table_ids_with_table_id_and_vnode(
+                        &parent_table_ids,
+                        split_full_key.user_key.table_id,
+                        split_full_key.user_key.get_vnode_id(),
+                    );
+                if table_ids_left.is_empty() || table_ids_right.is_empty() {
+                    tracing::debug!(
```
zwang28: What triggers this branch? Once triggered, the method will always fail at this point and never proceed.

src/meta/src/hummock/manager/compaction/compaction_group_schedule.rs
```diff
@@ -740,6 +852,219 @@ impl HummockManager {
 }
 
 impl HummockManager {
+    async fn build_normalize_plan(&self) -> Result<Option<NormalizePlan>> {
```
zwang28: unreachable Error

src/meta/src/hummock/manager/compaction/compaction_group_schedule.rs
```diff
@@ -46,6 +47,117 @@ use crate::hummock::table_write_throughput_statistic::{
...
+            .then_some((table_ids_left, table_ids_right))
+    }
+}
+
+fn build_normalize_plan_from_group_statistics(
+    groups: &[CompactionGroupStatistic],
+) -> Option<NormalizePlan> {
+    let mut groups = groups
+        .iter()
+        .filter(|group| !group.table_statistic.is_empty())
```
zwang28, line 114: Under what conditions does group.table_statistic.is_empty() return true?

zwang28, line 114: Under what conditions does group.table_statistic.is_empty() return true?

src/meta/src/hummock/manager/compaction/compaction_group_schedule.rs
```diff
@@ -740,6 +852,219 @@ impl HummockManager {
 }
 
 impl HummockManager {
+    async fn build_normalize_plan(&self) -> Result<Option<NormalizePlan>> {
+        let groups = self.calculate_compaction_group_statistic().await;
+        let Some(plan) = build_normalize_plan_from_group_statistics(&groups) else {
+            return Ok(None);
+        };
+
+        if plan.split_table_ids().is_none() {
```
zwang28: Under what conditions does plan.split_table_ids().is_none()?
Will it fail all subsequent attempts infinitely?

src/meta/src/hummock/manager/compaction/compaction_group_schedule.rs
Li0k: This is due to a historical legacy issue. The two groups cg2 and cg3 may exist without any tables and yet not be cleaned up.

Li0k: ditto

src/meta/src/hummock/manager/compaction/compaction_group_schedule.rs
Li0k: Added a stricter assertion

src/meta/src/hummock/manager/compaction/compaction_group_schedule.rs
```diff
@@ -46,6 +47,126 @@ use crate::hummock::table_write_throughput_statistic::{
...
+        assert!(!table_ids_left.is_empty() && !table_ids_right.is_empty());
+        (table_ids_left, table_ids_right)
+    }
+}
+
+fn gen_normalize_plan(
+    left: &CompactionGroupStatistic,
+    right: &CompactionGroupStatistic,
+) -> Option<NormalizePlan> {
+    let left_table_ids = left.table_statistic.keys().copied().collect_vec();
```
zwang28, line 82: Keep the BTreeSet type to enfore the order of table ids?

Li0k: table_statistic is a BTreeMap<StateTableId, u64>, so keys() already yields table ids in sorted order. Here collect_vec() only materializes that ordered keys for last() / partition_point(); it does not weaken the ordering guarantee.

PR #24855, opened 2026-02-24
fix(mysql-cdc): fix MySQL CDC auto schema change for UNSIGNED integer types

wcy-fdu:

- Fix MySQL CDC auto schema change failing on Debezium typeName like BIGINT UNSIGNED / INT UNSIGNED by normalizing qualified type names and mapping UNSIGNED integer family to safe RisingWave types.
- Map BIGINT UNSIGNED to DECIMAL to avoid overflow beyond i64, and promote other UNSIGNED ints to wider signed types when needed.

Related:
* #23278 (cherry-pick [#23616](https://github.com/risingwavelabs/risingwave/pull/23616))
* #23244

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/src/source/cdc/external/mysql.rs
+++ b/src/connector/src/source/cdc/external/mysql.rs
@@ -259,57 +259,81 @@ pub fn timestamp_val_to_timestamptz(value_text: &str) -> ConnectorResult<String>
 }
 
 pub fn type_name_to_mysql_type(ty_name: &str) -> Option<ColumnType> {
-    macro_rules! column_type {
-        ($($name:literal => $variant:ident),* $(,)?) => {
-            match ty_name.to_lowercase().as_str() {
-                $(
-                    $name => Some(ColumnType::$variant(Default::default())),
-                )*
-                "json" => Some(ColumnType::Json),
-                "date" => Some(ColumnType::Date),
-                "bool" => Some(ColumnType::Bool),
-                "tinyblob" => Some(ColumnType::TinyBlob),
-                "mediumblob" => Some(ColumnType::MediumBlob),
-                "longblob" => Some(ColumnType::LongBlob),
-                _ => None,
-            }
-        };
+    // Debezium schema change message may include extra qualifiers, e.g. `BIGINT UNSIGNED`,
+    // `BIGINT(20) UNSIGNED`, `INT UNSIGNED ZEROFILL`, etc.
+    let ty = ty_name.trim().to_lowercase();
+    let base = ty
+        .split_whitespace()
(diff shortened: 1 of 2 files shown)


src/connector/src/source/cdc/external/mysql.rs
```diff
@@ -259,57 +259,81 @@ pub fn timestamp_val_to_timestamptz(value_text: &str) -> ConnectorResult<String>
...
+        .unwrap_or_default()
+        .split('(')
+        .next()
+        .unwrap_or_default();
+    let is_unsigned = ty.contains("unsigned");
+
+    // Handle MySQL `... UNSIGNED` integer family in Debezium schema change.
+    //
+    // Notes about mapping:
+    // - `TINYINT UNSIGNED`  : 0..255            -> fits in RW Int16
```
xiangjinwu: nit: use `0..=255` notation

src/connector/src/source/cdc/external/mysql.rs
```diff
@@ -259,57 +259,81 @@ pub fn timestamp_val_to_timestamptz(value_text: &str) -> ConnectorResult<String>
...
+    // - `MEDIUMINT UNSIGNED`: 0..16777215       -> fits in RW Int32
+    // - `INT UNSIGNED`      : 0..4294967295     -> needs RW Int64
+    // - `BIGINT UNSIGNED`   : 0..18446744073709551615 -> needs RW Decimal
+    if is_unsigned {
+        match base {
+            // Keep as-is: RW mapping is already wide enough.
+            "tinyint" => return Some(ColumnType::TinyInt(Default::default())),
+            "mediumint" => return Some(ColumnType::MediumInt(Default::default())),
+            // Promote to a wider signed type.
+            "smallint" => return Some(ColumnType::Int(Default::default())),
```
xiangjinwu: This does not look like the correct abstraction. Given the function signature, it is returning a `sea_schema::mysql::def::ColumnType` rather than a `risingwave_common::types::DataType`. This function shall return MySQL types faithfully and populate `NumericAttr::unsigned` (instead of `Default::default()`) rather than assuming it has only one caller and the caller want RisingWave data types eventually.

xiangjinwu (commented):
Could you link to related PRs on `unsigned` support? So that we learn how to avoid missing similar updates in the future

wcy-fdu:
Related pr https://github.com/risingwavelabs/risingwave/pull/23244 https://github.com/risingwavelabs/risingwave/pull/23616

xiangjinwu (approved):

PR #24846, opened 2026-02-24
feat(sink): support alter jdbc sink url , username and password

xxhZs:
This pull request updates the handling of sink connector fields that can be altered on the fly, with a focus on improving support for JDBC sinks and ensuring string values are parsed correctly in SQL options. The changes enhance configurability and robustness, particularly for JDBC connectors.

Add "jdbc.url" ,"username","password" to the whitelist. After this PR, we will support users to execute ALTER sink  with them

**JDBC Sink Support and Configuration:**

* Added `JdbcSink` to the list of sinks that allow certain fields (`jdbc.url`, `user`, `password`) to be altered on the fly in `SINK_ALLOW_ALTER_ON_FLY_FIELDS`. [[1]](diffhunk://#diff-a365fc0d179182dc053c563a1b8d16f283b155ca2152d47859c9e5a1c40f9652R23-R24) [[2]](diffhunk://#diff-a365fc0d179182dc053c563a1b8d16f283b155ca2152d47859c9e5a1c40f9652R167-R175)
* Updated the logic in `check_sink_allow_alter_on_fly_fields` to correctly handle JDBC sinks by name, ensuring proper field validation for this connector type.

**SQL Option Parsing Improvements:**

* Modified the `TryFrom<(&String, &String)> for SqlOption` implementation to always wrap values in single quotes and escape embedded single quotes, preventing parsing errors for values with special characters (such as JDBC URLs).I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## Checklist

- [ ] I have written necessary rustdoc comments.
- [ ]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  I have checked the [Release Timeline](https://github.com/risingwavelabs/rw-commits-history/blob/main/release_timeline.md) and [Currently Supported Versions](https://docs.risingwave.com/changelog/release-support-policy#support-end-dates-for-recent-releases) to determine which release branches I need to cherry-pick this PR into. 

## Documentation

- [ ]  My PR needs documentation updates. 

<details>
<summary><b>Release note</b></summary>

</details>

PR #24861, opened 2026-02-25
fix: support jdbc sink alter-on-fly fields

tabVersion:

following https://github.com/risingwavelabs/risingwave/pull/24846

## Checklist

## Documentation

<details>

</details>


src/connector/src/with_options_test.rs
xxhZs, line 714: May need to revert.
https://github.com/risingwavelabs/risingwave/pull/24846 ’s bug

tabVersion: Let it be. Let's review the impl afterwards.

PR #24871, opened 2026-02-26
feat: add secret support in UDF

tabVersion:

This PR adds **user-facing secret support for UDF options** and enforces secret dependency management for UDF lifecycle.

### User-facing impact

- Users can now bind secret-backed custom options when creating UDFs via:
  - `WITH (api_token = SECRET my_schema.my_secret)`
- For embedded **Python** and **JavaScript** UDFs, these options are exposed at runtime through an immutable global:
  - `RW_UDF_OPTIONS`
- UDFs now track dependencies on referenced secrets. If a secret is used by a UDF, `DROP SECRET ...` is rejected until the dependent UDF is dropped.
- Secret existence and privilege are validated during `CREATE FUNCTION`.
- Plaintext custom UDF options are rejected for security (custom options must use `SECRET ...` references).

### How this works

- SQL parser supports secret refs in `CREATE FUNCTION ... WITH (...)` and stores them in `secret_refs`.
- Frontend resolves secret names to secret IDs and checks privileges.
- Meta persists function->secret references and creates object dependency edges.
- Expr runtime resolves secret refs via local secret manager and injects resolved options into Python/JS runtimes.

### Current limitations

- Custom UDF options currently support **secret references only** (no arbitrary plaintext key-value custom options).
- Runtime option injection is currently implemented for embedded Python/JavaScript UDF runtimes.

### Tests

- Added E2E SQLLogicTest for function+secret behavior:
  - `e2e_test/udf/tests/management/secret_dependency.slt`
- Added parser tests for secret option syntax.
- Added meta catalog test for function-secret dependency creation and drop restriction.

## Checklist

## Documentation

<details>

Users can now pass secrets to UDFs securely by referencing catalog secrets in `CREATE FUNCTION ... WITH (...)` (for example, `api_token = SECRET my_schema.my_secret`).

For embedded Python and JavaScript UDFs, resolved secret values are available at runtime through `RW_UDF_OPTIONS`. RisingWave now tracks function->secret dependencies, so dropping a secret that is still referenced by a UDF is prevented until the dependent function is removed.

This improves secure credential handling for user-defined functions and prevents accidental secret breakage.

</details>

### DIFF
--- a/src/frontend/src/binder/expr/function/mod.rs
+++ b/src/frontend/src/binder/expr/function/mod.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::collections::{HashMap, HashSet};
+use std::collections::{BTreeMap, HashMap, HashSet};
 use std::str::FromStr;
 use std::sync::Arc;
 
@@ -21,9 +21,11 @@ use itertools::Itertools;
 use risingwave_common::acl::AclMode;
 use risingwave_common::bail_not_implemented;
 use risingwave_common::catalog::INFORMATION_SCHEMA_SCHEMA_NAME;
+use risingwave_common::secret::LocalSecretManager;
 use risingwave_common::types::{DataType, MapType};
 use risingwave_expr::aggregate::AggType;
 use risingwave_expr::window_function::WindowFuncKind;
+use risingwave_pb::secret::PbSecretRef;
 use risingwave_sqlparser::ast::{
     self, Expr as AstExpr, Function, FunctionArg, FunctionArgExpr, FunctionArgList, Ident,
     OrderByExpr, Statement, Window,
@@ -35,8 +37,8 @@ use crate::binder::bind_context::Clause;
 use crate::catalog::function_catalog::FunctionCatalog;
 use crate::error::{ErrorCode, Result, RwError};
 use crate::expr::{
-    Expr, ExprImpl, ExprType, FunctionCallWithLambda, InputRef, TableFunction, TableFunctionType,
-    UserDefinedFunction,
+    Expr, ExprImpl, ExprType, FunctionCallWithLambda, InputRef, Literal, TableFunction,
+    TableFunctionType, UserDefinedFunction,
 };
 use crate::handler::privilege::ObjectCheckItem;
 
@@ -698,6 +700,7 @@ impl Binder {
         body: &str,
         arg_names: &[String],
         args: Vec<ExprImpl>,
+        secret_refs: &BTreeMap<String, PbSecretRef>,
     ) -> Result<ExprImpl> {
         // This represents the current user defined function is `language sql`
         let ast = Parser::parse_sql(body)?;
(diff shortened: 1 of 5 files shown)


src/frontend/src/handler/create_sql_function.rs
```diff
@@ -177,6 +177,7 @@ pub async fn handle_create_sql_function(
         is_batched: None,
         created_at_epoch: None,
         created_at_cluster_version: None,
+        secret_refs: Default::default(),
```
hzxa21: I suggest we support secret in sql udf as well because:
- only SQL UDF can call other UDF inside its body
- we can put common logic in non-SQL udf
- we can have multiple SQL udf to call the same non-SQL udf with different secrets
- moreover, we can have finer-grain access control
    - User A creates non-SQL UDF `foo` and SQL UDFs `foo_B`, `foo_C` with different secrets but both call `foo` inside.
    - User A grant usage privilege for `foo_B` to user B and `foo_C` to user C.
    - User B and C cannot see the non-SQL UDF definitions and they cannot hack and see the secret value either.

Exmaples:
```
CREATE FUNCTION fetch_slack_msg(channel_id VARCHAR, api_key VARCHAR) RETURNS VARCHAR
  LANGUAGE javascript AS $$ ... $$;

-- we can discuss on the sql syntax
CREATE FUNCTION fetch_slack_msg_for_B(channel_id VARCHAR) RETURNS VARCHAR
  LANGUAGE sql AS $$
      SELECT fetch_slack_msg(channel_id, SECRET secret_for_B)
  $$;

CREATE FUNCTION fetch_slack_msg_for_C(channel_id VARCHAR) RETURNS VARCHAR
  LANGUAGE sql AS $$
      SELECT fetch_slack_msg(channel_id, SECRET secret_for_C)
  $$;
```

src/frontend/src/handler/create_aggregate.rs
```diff
@@ -147,6 +147,7 @@ pub async fn handle_create_aggregate(
         is_batched: None,
         created_at_epoch: None,
         created_at_cluster_version: None,
+        secret_refs: Default::default(),
```
hzxa21, line 151: I saw we have the fill_secrets logic in UDAF implementaion but we ignore the secret refs here. Is this intentional?

tabVersion: It is a bug, already fixed

hzxa21: It seems that it is not fixed?

src/sqlparser/src/ast/mod.rs
```diff
@@ -3694,27 +3696,52 @@ impl TryFrom<Vec<SqlOption>> for CreateFunctionWithOptions {
...
+                    } else {
+                        return Err(StrError(format!(
+                            "option {} only accepts boolean values",
+                            option_name
+                        )));
+                    }
                 }
                 _ => {
-                    return Err(StrError(format!("unknown option: {}", option.name)));
+                    let SqlOptionValue::SecretRef(secret_ref) = option.value else {
```
hzxa21, line 3732: file secret won't work in the current implementation. Let's check and ban it.

src/frontend/src/binder/expr/function/mod.rs
```diff
@@ -718,6 +721,21 @@ impl Binder {
                 arguments.insert(arg_names[i].clone(), arg);
             }
         }
+        for (option_key, secret_value) in
```
hzxa21, line 724: @xiangjinwu may know better than me. But I suspect resolving the secret in binding phase can cause secret to leak more easily. For example, it can leak in `EPLAIN` run by the user of the UDF, even he/she is not the owner of it. Though it can be hard to completely prevent revealing secret in UDF use case, I think it is better to do secret resolution in CN instead of in binder.


One way is to resolve it in CN is to do it in expr building. I attempt to do this once before: https://github.com/risingwavelabs/risingwave/pull/24292/changes#diff-6aaa44e6e5af7bd24bfcb359cbd2708076ca0c529066573d17d5db1f3c46977b

src/sqlparser/src/ast/mod.rs
tabVersion: for embedded UDF, I think we can still let it access file? I am not sure about it.

src/frontend/src/binder/expr/function/mod.rs
tabVersion: might need have a clearer design first internally. I'd close this one first.

PR #24883, opened 2026-02-26
feat: reserve memory headroom for unspillable DataFusion operators

silver-ymz:

### Problem

The DataFusion execution memory pool (`RwMemoryPool`) was a single shared pool. Under memory pressure:

- **Spillable** operators (e.g. sort, hash join) can spill to disk when allocation fails, so they can keep allocating until the pool is full.
- **Unspillable** operators have no spill path; they must get memory or the query fails.

When both run in the same query, spillable operators could use the entire pool. Unspillable operators would then get allocation failures even though the query could have succeeded if some memory had been kept available for them. This led to unnecessary "insufficient memory" errors and poor behavior when mixing operator types.

### Solution

1. **Split by consumer type**:
   Each `RwMemoryPool` has two `MemoryContext` — `spillable_ctx` for operators that can spill, `unspillable_ctx` for those that cannot. `grow` / `shrink` / `try_grow` route based on `consumer.can_spill()`.

3. **Reserve headroom via context hierarchy**:
   A shared intermediate context (`df_spillable_budget_ctx`, limit = 90% of pool) sits between the root and all per-query spillable contexts:
   ```
   batch_mem_context (limit = L)
   ├── df_spillable_budget_ctx (limit = L × 0.9)   ← shared, DataFusion-only
   │   ├── query_1.spillable_ctx
   │   └── query_2.spillable_ctx
   ├── query_1.unspillable_ctx
   └── query_2.unspillable_ctx
   ```
   This structurally caps total spillable usage at 90%, guaranteeing 10% headroom for unspillable operators — without transient overcharges or spurious failures under concurrency.

### Related

https://github.com/apache/datafusion/issues/17334

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6570,7 +6570,7 @@ dependencies = [
 [[package]]
 name = "iceberg"
 version = "0.7.0"
-source = "git+https://github.com/risingwavelabs/iceberg-rust.git?rev=4d5f635572a8e90c77a3c56c12aa40d5362f0d68#4d5f635572a8e90c77a3c56c12aa40d5362f0d68"
+source = "git+https://github.com/risingwavelabs/iceberg-rust.git?rev=10410fa883791d3b38d3751d050c79223ad4317f#10410fa883791d3b38d3751d050c79223ad4317f"
 dependencies = [
  "anyhow",
  "apache-avro 0.20.0",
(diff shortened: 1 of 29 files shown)


chenzl25 (approved):
LGTM. Does it resolve the TPCH OOM issues?

silver-ymz:
> Does it resolve the TPCH OOM issues?

Seems not 😢

PR #24887, opened 2026-02-26
fix(meta): fence drop vs split updates and skip stale fragment splits

tabVersion:

fix https://github.com/risingwavelabs/risingwave/issues/24825

## Background

  We observed a flaky failure where `fragment_splits` updates raced with fragment deletion:

  - `fragment_splits.fragment_id` has FK to `fragment.fragment_id`.
  - split persistence can happen in barrier post-collect paths (`SourceChangeSplit` / `Reschedule`).
  - if a fragment is dropped concurrently, direct upsert into `fragment_splits` may hit FK violation.
  - the barrier command then fails in post-collect, the cluster enters recovering state, and subsequent DDLs can fail with "The cluster is recovering".

  ## Root Cause

  The drop path and split persistence path were not fully fenced at the right boundary:

  - drop cleanup could proceed while split persistence was still possible.
  - `update_fragment_splits` previously upserted blindly without checking fragment existence.

  ## Changes

  ### 1) Fence drop before catalog deletion

  In `DdlController::drop_object`, acquire fences **before** `catalog_controller.drop_object(...)`:

  - hold `reschedule_lock_read_guard`
  - hold `source_manager.pause_tick`

  This narrows the race window so drop does not proceed while new reschedule/tick-driven split updates are being introduced.

  ### 2) Make split persistence tolerant to stale fragments

  In `CatalogController::update_fragment_splits`:

  - query existing fragment IDs first
  - filter out stale/missing fragment IDs
  - upsert only valid rows
  - if all fragment IDs are stale, return success as no-op

  This prevents FK violation from escalating to barrier failure/recovery.

**Summary**  
This PR enhances the handling of fragment splits by filtering out stale fragment IDs before processing updates, preventing foreign key violations and improving data integrity. It also introduces synchronization measures in DDL operations, such as acquiring a read lock and pausing source ticks, to mitigate race conditions between fragment drops, split updates, and source processing. These changes ensure operational consistency during catalog modifications while adding logging for better observability of skipped updates.

**Details**  
- **Stale Fragment ID Filtering**: To maintain referential integrity and avoid database constraint violations during split updates, the code now queries existing fragment IDs and skips updates for non-existent (stale) ones, ensuring only valid fragments are processed and enabling an early exit if no updates are needed.  
- **Synchronization in DDL Operations**: To prevent race conditions where split updates or source ticks could interfere with fragment drops, read locks are acquired and source ticks are paused early in drop functions, establishing a fencing boundary that protects catalog consistency during deletions.  
- **Logging for Skipped Updates**: To provide observability into potential issues like races or failures, warning logs are emitted when stale fragment updates are skipped, helping operators monitor and debug inconsistent states without affecting core functionality.  
- **Removal of Redundant Locks**: To simplify the locking strategy and reduce unnecessary overhead, previously redundant guards in DDL functions are eliminated, clarifying the synchronization flow while maintaining the same protection levels.

## Checklist

### DIFF
--- a/src/meta/src/controller/fragment.rs
+++ b/src/meta/src/controller/fragment.rs
@@ -18,7 +18,7 @@ use std::sync::Arc;
 
 use anyhow::{Context, anyhow};
 use futures::TryStreamExt;
-use itertools::Itertools;
+use itertools::{Either, Itertools};
 use risingwave_common::bail;
 use risingwave_common::bitmap::Bitmap;
 use risingwave_common::catalog::{FragmentTypeFlag, FragmentTypeMask};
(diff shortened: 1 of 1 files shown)


yezizp2012 (approved):
LGTM

PR #24896, opened 2026-02-27
refactor(hummock): unify level normalization for table-id mutations

Li0k:

This PR hardens Hummock against SSTs with unexpected `table_ids` state after table-id truncation and makes related maintenance logic more consistent.

Main changes:
- Refactor compaction-group level mutation paths in `hummock_version_ext`:
  - Introduce/centralize `Level::{recompute_size, normalize, delete_ssts, truncate_tables}`.
  - Introduce `OverlappingLevel::normalize` and `Levels::truncate_tables`.
  - Apply truncation via one unified entry in `GroupDeltaCommon::TruncateTables`.
  - Remove duplicated ad-hoc size bookkeeping by normalizing levels after mutation.
- Add/expand unit tests in `hummock_version_ext` to verify:
  - empty-`table_ids` SST cleanup,
  - size recomputation correctness,
  - sub-level normalization,
  - truncate flow and delta application behavior.
- Guard SST iterators (`forward` and `backward`) with explicit assertions that `table_ids` is non-empty, so violations fail fast with clear SST/object context.
- Fix `get` path candidate selection in `version.rs`:
  - after locating an SST by key range in non-overlapping levels, additionally require `table_id` membership in `sst.table_ids`;
  - skip non-matching SSTs instead of reading them.
- Add a focused storage test (`test_get_skips_sst_by_table_id_filter`) for the `get` path table-id filter.

Why:
- We observed a panic from `table_ids.first().unwrap()` in SST iterators.
- Empty `table_ids` should be pruned during version-delta application; this PR both strengthens that pruning path and adds defensive checks at read sites.
- The `get` path fix avoids touching SSTs that overlap by key-range but do not belong to the queried table id.

## Checklist

## Documentation

<details>

No user-facing SQL behavior change. This PR improves internal Hummock robustness for table-id truncation and SST selection.

</details>

### DIFF
--- a/src/storage/src/hummock/store/version.rs
+++ b/src/storage/src/hummock/store/version.rs
@@ -772,6 +772,15 @@ impl HummockVersionReader {
                         continue;
                     }
                     table_info_idx = table_info_idx.saturating_sub(1);
+
+                    if level.table_infos[table_info_idx]
+                        .table_ids
+                        .binary_search(&table_id)
+                        .is_err()
+                    {
+                        continue;
+                    }
+
                     let ord = level.table_infos[table_info_idx]
                         .key_range
                         .compare_right_with_user_key(full_key.user_key.as_ref());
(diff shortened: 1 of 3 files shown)


src/storage/src/hummock/store/version.rs
```diff
@@ -772,6 +772,15 @@ impl HummockVersionReader {
                         continue;
                     }
                     table_info_idx = table_info_idx.saturating_sub(1);
+
+                    if level.table_infos[table_info_idx]
+                        .table_ids
+                        .binary_search(&table_id)
+                        .is_err()
+                    {
+                        continue;
```
zwang28, line 781: In what scenario would this line be reached?

Li0k: The get operation only filters by key-range. As mentioned in the PR, once we remove the table-id via truncate-table, this path may be triggered.

PR #24907, opened 2026-02-28
refactor(cdc): serialize MySQL CDC schema history writes to prevent lost updates

wcy-fdu:

This PR makes MySQL CDC schema history persistence more robust by serializing `OpendalSchemaHistory` read-modify-write updates and by making CDC engine shutdown idempotent to reduce noisy failures during error paths.
### Why is this needed?
We observed MySQL CDC sources getting stuck with Debezium errors like “table schema isn’t known to this connector”, which is consistent with incomplete/corrupted schema history.
`OpendalSchemaHistory` persists history via read → append → write-back to object storage. Without serialization, concurrent store operations (e.g., snapshot/recovery/DDL handling or fast restart paths) can cause lost updates where one write overwrites another, leaving history incomplete.
### What does this PR defend against?
Lost-update corruption of schema history due to concurrent read-modify-write on the same history object.
Repeated close/stop paths that throw IllegalStateException (“Engine has been already shut down”), which can obscure the original failure.
### Performance impact
Low. Schema history writes happen mainly during startup/snapshot/DDL/schema changes, not per-row ingestion. The lock only serializes writes within a single source history instance, and object-store IO remains the dominant cost.

## Checklist

## Documentation

<details>

</details>


zwang28:
If OpendalSchemaHistory::doStart can be invoked concurrently with OpendalSchemaHistory::doStoreRecord, the new lock should also be applied to both methods.

PR #24937, opened 2026-03-03
feat(meta): add compression and part-read for hummock version checkpoint

Li0k:

- Add `CheckpointCompressionAlgorithm` enum (`NONE`, `ZSTD`, `LZ4`) and `HummockVersionCheckpointEnvelope` to `hummock.proto`
- Read checkpoint objects with metadata + bounded ranged reads to avoid single whole-file read timeout on large checkpoints (for example, ~1GB checkpoints)
- Compress checkpoint data with zstd by default before writing
- Add xxhash64 checksum for envelope payload integrity verification
- Support backward-compatible reading: try envelope format first, and fall back to legacy raw `PbHummockVersionCheckpoint`
- Update backup restore to write the compressed envelope format
- Keep empty checkpoint objects as decode errors instead of treating them as missing checkpoints, matching the previous behavior
- Clarify format detection docs: readers use `checksum` as the discriminator because legacy bytes may still decode as an envelope through protobuf unknown-field skipping

## Checklist

## Documentation

<details>

Improve reliability of Hummock version checkpoint loading for large checkpoint objects and add compressed checkpoint persistence with integrity verification.

</details>


src/meta/src/controller/catalog/drop_op.rs
```diff
@@ -469,6 +469,7 @@ impl CatalogController {
         }
 
         subscription.delete(&txn).await?;
+        txn.commit().await?;
```
zwang28, line 472: Was this a bug in main?

src/meta/src/hummock/manager/checkpoint.rs
```diff
@@ -69,26 +80,235 @@ impl HummockVersionCheckpoint {
...
 impl HummockManager {
     /// Returns Ok(None) if not found.
+    ///
+    /// Reads large checkpoints using bounded parallel chunked reads (128MB per chunk) to avoid
+    /// single-request timeout issues.
+    /// Supports both compressed (envelope) and uncompressed (legacy) checkpoint formats.
     pub async fn try_read_checkpoint(&self) -> Result<Option<HummockVersionCheckpoint>> {
-        use prost::Message;
-        let data = match self
+        const CHUNK_SIZE: usize = 128 * 1024 * 1024; // 128MB
```
zwang28: Should we consider making these configurable?

zwang28 (approved):
rest LGTM

src/meta/src/controller/catalog/drop_op.rs
Li0k: Yes, this is a bug that exists on the main branch. I have confirmed it with @yezizp2012  . Since it severely interferes with CI, I have made the change.

src/meta/src/hummock/manager/checkpoint.rs
Li0k: I don't think I'll change it, but why not? I (Codex) will modify it.

PR #24951, opened 2026-03-04
fix(stream): split batched barriers for coalescer FIFO alignment

lmatz:
## Summary

Implement **multiplexed barrier coalescing** for remote exchanges: when N upstream actors on the same node send to the same downstream actor, barrier messages are coalesced from N to 1 per epoch, while data continues to flow through independent per-actor channels at full throughput.

This PR contains the core feature plus a series of correctness fixes that were discovered during integration testing (see [Correctness fixes](#correctness-fixes-discovered-during-integration) below).

## Motivation

In the current architecture (main branch), each `(upstream_actor, downstream_actor)` pair has its own gRPC stream carrying both data and barriers. When a fragment has N upstream actors on the same node, that means N barrier messages per epoch per downstream actor—all carrying identical barrier metadata. This is redundant and becomes a scalability bottleneck for high-parallelism deployments.

A previous approach (branch `lz/barrier`) tried to solve this by multiplexing ALL traffic (data + barriers) through a single shared channel and gRPC stream. However, this caused 10–46% throughput regression on Nexmark benchmarks because the shared data path created bottlenecks: single mpsc contention, single gRPC stream serialization, extra async hops in the demuxer, and permit return amplification.

## Architecture

The key insight is: **keep data on per-actor channels (identical to non-multiplexed mode), only coalesce barriers on a separate channel**. Use message counting to synchronize the two paths.

```
SENDER NODE                                    RECEIVER NODE
============                                   =============
Actor A1 ──data──> [per-actor channel] ──gRPC──> RemoteInput ──> CountingMergeInput ──> MergeExecutor
         ──barrier──> Coalescer                                       ↑ barrier_rx
Actor A2 ──data──> [per-actor channel] ──gRPC──> RemoteInput ──> CountingMergeInput ──> MergeExecutor
         ──barrier──> Coalescer                                       ↑ barrier_rx
                        │                                              │
                        └── coalesced barrier ──> [barrier ch] ──gRPC──> run_barrier_receiver
                            (with per-actor data counts)                (distributes to CountingMergeInputs)
```

### Sender side

- **`Output::CoalescedBarrier`** (new enum variant): Each upstream actor's output. Data (Chunk/Watermark) goes directly through `data_ch` (a normal per-actor `permit::Sender`), incrementing `epoch_msg_count`. Barriers are sent individually to the coalescer as `(actor_id, barrier, epoch_msg_count)`, then count resets to 0.

- **`BarrierCoalescer`**: Queue-based coalescer. Each actor has a `VecDeque<(DispatcherBarrier, u64)>`. Fast actors can submit barriers for future epochs before slow actors catch up. When all actors have submitted for the front epoch, produces `(barrier, actor_ids, HashMap<ActorId, u64>)`.

- **`MultiplexedOutputCoordinator`**: Background tokio task. Receives actor submissions from an mpsc channel, feeds the coalescer, and sends coalesced barriers through a barrier-only `permit::Sender` (which feeds into a barrier-only gRPC stream).

### Receiver side

- **`run_barrier_receiver`**: Background tokio task. Reads coalesced barriers from the barrier-only gRPC stream, extracts `actor_data_counts` from protobuf, and distributes `(barrier, expected_count)` to each actor's `CountingMergeInput` via per-actor mpsc channels.

- **`CountingMergeInput`**: Wraps a `RemoteInput` (data-only) and a barrier channel. Counts data messages from the inner stream. When a barrier arrives with `expected_count`, drains remaining data from inner until `msg_count == expected_count`, then yields the barrier. This preserves the ordering invariant: `[epoch E data] → barrier E → [epoch E+1 data]`.

### Protocol changes

- **`BarrierBatch` protobuf**: Added `coalesced_actor_ids` and `actor_data_counts` fields for coalesced barrier metadata.
- **`GetStreamRequest.Get`**: Added `up_actor_ids` field. When non-empty, triggers multiplexed mode—creates a barrier-only stream on the sender side.

### Connection setup flow

1. Receiver calls `get_stream_barrier(up_actor_ids, ...)` → sender creates barrier group (per-actor data channels stored in `pre_created_data_receivers`, coalescer + coordinator spawned, `CoalescedBarrierRemote` output requests sent to upstream actors' DispatchExecutors).
2. Receiver calls `get_stream(up_actor_id, down_actor_id, ...)` for each actor → sender finds pre-created data channel receiver in `pre_created_data_receivers`.
3. Receiver wraps each `RemoteInput` with `CountingMergeInput`, spawns `run_barrier_receiver`.

### Why message counting instead of sequence numbers?

Message counting is simpler and equally correct. The sender increments a counter for each Chunk/Watermark, resets on barrier, and embeds the count in the coalesced barrier. The receiver counts messages from `RemoteInput`. Since TCP/gRPC preserves ordering, the count is an exact measure of "how many data messages belong to this epoch." No need for per-message sequence numbers or out-of-order handling.

### Why split barrier batches into individual coalescer submissions?

`DispatchExecutor::try_batch_barriers()` may batch consecutive barriers differently across actors based on timing (Actor A1 batches `[b1, b2]` while Actor A2 sends `b1` then `b2` separately). If entire batches were submitted as single coalescer entries, the FIFO queue pairing would become misaligned:

```
A1 queue: [ ([b1, b2], count=5) ]          ← 1 entry covering 2 epochs  
A2 queue: [ ([b1], count=3), ([b2], count=7) ]  ← 2 entries, one per epoch
```

First drain pairs A1's 2-epoch entry with A2's epoch-1-only entry → wrong counts → deadlock. The fix splits batches at the sender: first barrier carries the epoch's data count, subsequent barriers carry count=0 (since batching only groups consecutive barriers with no data between them).

### Why not block the sender on barrier coalescing?

The sender never blocks. `Output::CoalescedBarrier::send()` submits barriers to an unbounded mpsc channel and returns immediately. This avoids backpressure from slow actors propagating to fast actors through the coalescer. The queue-based coalescer handles fast-actor-ahead-of-slow-actor naturally.

### Why are data channels created by the barrier group setup, not by individual get_stream calls?

The barrier-only gRPC stream (`get_stream_barrier`) must be established first because it triggers `CreateBarrierGroup` on the sender, which creates per-actor data channels and stores their receivers. Subsequent per-actor `get_stream` calls then find pre-created receivers via `pre_created_data_receivers`. This ordering ensures data channels exist before `RemoteInput` tries to connect.

### Why unbounded channels for barrier distribution?

Both the coalescer submission channel (`Output → Coordinator`) and the barrier distribution channels (`run_barrier_receiver → CountingMergeInput`) are unbounded. Barriers are small and infrequent relative to data. The data path remains bounded by the permit system. Using bounded channels here would risk deadlocks when multiple epochs' barriers queue up (which is normal during epoch pipelining).

### Why retry DML reader lookup with exponential backoff?

During recovery and immediately after a streaming job is created, there is a brief window where the DML executor may not have registered its reader yet. The new connection-setup flow (`CreateBarrierGroup` → per-actor `get_stream` calls → actor build / register) widens this race, so batch DML operations occasionally hit `DmlError::NoReader`. The new `DmlManager::table_dml_handle_with_retry` resolves this with bounded exponential backoff (up to 10 retries, capped at 2s per attempt) instead of failing the DML statement.

## Correctness fixes discovered during integration

In addition to the core feature, this PR contains several correctness fixes uncovered by integration testing and exploratory chaos. Each is its own commit:

- **`fix(stream): split batched barriers into individual coalescer submissions`** — FIFO alignment fix described in [Why split barrier batches](#why-split-barrier-batches-into-individual-coalescer-submissions).
- **`fix(stream): fix actor ID mismatch in coalescer and coordinator drain bug`** — coalescer was keyed on the wrong actor identifier in one path, mis-pairing entries during drain.
- **`fix(stream): fix barrier receiver lifecycle and CountingMergeInput error handling`** — `CountingMergeInput` now surfaces `ExchangeChannelClosed` instead of silently ending the stream when the barrier-only channel disconnects; matches the error semantics of the legacy non-multiplexed path.
- **`fix(stream): use downstream actor ID for dispatch matching in CoalescedBarrier`** — `Output::actor_id()` keeps the downstream actor ID for dispatch matching while the coalescer submission uses upstream IDs; required to preserve existing dispatch-by-downstream semantics.
- **`fix(stream): queue CreateBarrierGroup when partial graph not yet running`** — when a `CreateBarrierGroup` arrives before the partial graph has reached `Running`, the request is now queued and replayed instead of rejected; addresses a real race observed during streaming-job startup and partial-graph recovery.
- **`fix(stream): wait for first barrier in CountingMergeInput before yielding data`** — without this, a `CountingMergeInput` could yield data from inner `RemoteInput` before the first epoch's coalesced barrier arrived, breaking the `[epoch E data] → barrier E` invariant on cold start.
- **`fix(stream): carry over excess msg_count in CountingMergeInput to prevent deadlock`** — if data for epoch E+1 arrived before the coalesced barrier for E, the inner counter would overshoot `expected_count`; the carry-over preserves the surplus into the next epoch instead of waiting indefinitely.
- **`fix(batch): retry on NoReader when DML executor not yet registered`** — implemented `table_dml_handle_with_retry` in `DmlManager`, called from batch insert/delete/update executors; covers the brief window described in [Why retry DML reader lookup](#why-retry-dml-reader-lookup-with-exponential-backoff).
- **`fix(stream): suppress repetitive barrier receiver warnings`** — log-noise reduction during expected lifecycle events (channel closed on shutdown).
- **`fix(stream): suppress clippy type_complexity warning on test helpers`** — non-functional.

## Files changed

| File | Change |
|------|--------|
| `proto/stream_plan.proto` | Add `coalesced_actor_ids` and `actor_data_counts` to `BarrierBatch` |
| `proto/task_service.proto` | Add `up_actor_ids` to `GetStreamRequest.Get` |
| `src/stream/src/executor/exchange/multiplexed.rs` | **New file**: `BarrierCoalescer`, `MultiplexedOutputCoordinator`, `CountingMergeInput`, `run_barrier_receiver`, `create_multiplexed_output` |
| `src/stream/src/executor/exchange/output.rs` | `Output` enum with `Direct` and `CoalescedBarrier` variants; downstream actor ID for dispatch matching |
| `src/stream/src/executor/exchange/permit.rs` | Add `coalesced_actor_ids` / `actor_data_counts` to `MessageWithPermits`, add `send_coalesced()` |
| `src/stream/src/executor/exchange/input.rs` | Multiplexed input creation in `new_inputs()` |
| `src/stream/src/executor/exchange/mod.rs` | Module registration |
| `src/stream/src/executor/dispatch.rs` | Handle `NewOutputRequest::CoalescedBarrierRemote` |
| `src/stream/src/executor/merge.rs` | Re-export / minor integration |
| `src/stream/src/executor/upstream_sink_union.rs` | Match new `NewOutputRequest::CoalescedBarrierRemote` variant in tests |
| `src/stream/src/executor/mod.rs` | Re-exports |
| `src/stream/src/from_proto/merge.rs` | Wire multiplexed input path through `from_proto` |
| `src/stream/src/task/barrier_worker/mod.rs` | `NewOutputRequest::CoalescedBarrierRemote`, `CreateBarrierGroup` operation; queue when partial graph not yet running |
| `src/stream/src/task/barrier_worker/managed_state.rs` | `create_barrier_group()`, `pre_created_data_receivers` |
| `src/stream/src/task/stream_manager.rs` | `create_barrier_group()` RPC handler |
| `src/compute/src/rpc/service/stream_exchange_service.rs` | `get_barrier_stream_impl()` for barrier-only gRPC stream |
| `src/rpc_client/src/compute_client.rs` | `get_stream_barrier()` client method |
| `src/dml/src/dml_manager.rs` | New `table_dml_handle_with_retry` with exponential backoff for the DML-reader-not-yet-registered window |
| `src/batch/executors/src/executor/insert.rs` | Switch to `table_dml_handle_with_retry` |
| `src/batch/executors/src/executor/delete.rs` | Switch to `table_dml_handle_with_retry` |
| `src/batch/executors/src/executor/update.rs` | Switch to `table_dml_handle_with_retry` |

## Test plan

- [x] 17 unit tests in `multiplexed.rs` covering:
  - `BarrierCoalescer`: basic, queued, add/remove actor, split-batches alignment
  - `CountingMergeInput`: basic, idle actor, barrier before data, data before barrier, multiple epochs, first-barrier wait, carry-over across epochs, error on barrier disconnect (with and without buffered data)
  - `Output`: sends upstream actor ID to coalescer; wrong actor ID panics in coalescer
  - `distribute_barrier`: removes closed actors, all-actors-closed case
- [x] Clippy clean, `cargo fmt` clean, `cargo doc` clean
- [x] Docker image build: [Buildkite docker build 18354](https://buildkite.com/risingwavelabs/docker/builds/18354), image tag `ghcr.io/risingwavelabs/risingwave:v2.9.0-44e1de4b26` (later iterations: `v2.9.0-2711dba95e`)
- [x] Nexmark backlog + RW Cloud aggregation queries on `pr24951-barrier-counting-imgtest` and `pr24951-barrier-counting-32c` (manual, see [thread context](https://github.com/risingwavelabs/risingwave/pull/24951#issuecomment-4637583312))
- [x] Performance validation: same-throughput, zero-data, and saturated-capacity comparisons against `nightly-20260530`; see [comment with full setup + results + Grafana links](https://github.com/risingwavelabs/risingwave/pull/24951#issuecomment-4637583312). Headline: at the same ~20k source rows/s the PR uses **78.2 compute cores vs baseline's 112.8 (~31% reduction)**, gRPC traffic **193.9 → 38.4 MB/s (~80% reduction)**, and meta barrier p95 **5.48s → 1.19s (~4.6× faster)**.
- [ ] Chaos-mesh runs on PR image — pending: builds [1888](https://buildkite.com/risingwave-test/chaos-mesh/builds/1888) (longevity / Nexmark chaos), [1889](https://buildkite.com/risingwave-test/chaos-mesh/builds/1889) (decoupled-sink + postgres), [1890](https://buildkite.com/risingwave-test/chaos-mesh/builds/1890) (TPC / shared-source backfill + compute pod-kill), [1891](https://buildkite.com/risingwave-test/chaos-mesh/builds/1891) (ch-benchmark-pg-cdc).
- [ ] Local `risedev` recovery test with tracing on coalescer / barrier receiver / `CountingMergeInput` / `CreateBarrierGroup` paths — in progress to verify recovery-barrier path, asymmetric stream-death handling, and partial-graph reset interaction.

### Known caveats / open items

- Performance results are single 3-minute windows without repeats; tail-latency outliers (zero-data `sync-storage` and saturated `meta barrier p95`) need PR-only repeat runs to discriminate noise from regression.
- Workload coverage is synthetic datagen + symmetric base+MV chains; high-fanout joins, real customer DAG shapes, and low-parallelism cases are not characterized here.
- Mixed-version / rolling upgrade compatibility: the new `up_actor_ids` field in `GetStreamRequest.Get` and the new `get_stream_barrier` RPC mean old↔new compute pairs need to be validated for fallback behavior or strict simultaneous-upgrade documented as a deployment constraint.


lmatz:
## Performance validation for multiplexed barrier exchange

I ran an end-to-end performance validation for this PR against a baseline image used in this validation to isolate whether the multiplexed barrier exchange path improves high-parallelism streaming workloads.

### Images compared

- PR image: `ghcr.io/risingwavelabs/risingwave:v2.9.0-2711dba95e`
- Baseline image used in this validation: `ghcr.io/risingwavelabs/risingwave:nightly-20260530` (`6fa726ce...` from `select version()`)

### Cluster configuration

The final validation used the same cluster shape for both images:

- Compute: 8 pods x 16 CPU limit, total 128 compute cores
- Compactor: 4 pods x 16 CPU limit, total 64 compactor cores
- Serverless compaction: disabled
- All tables and MVs were set to `parallelism = 256`

Each test case was run with a fresh rebuild of the workload: drop all test tables/MVs, confirm no remaining `pr24951_*` relations, create the workload, alter all relations to parallelism 256, warm up, collect a 3-minute metrics window, then drop the workload again.

### Workload shape

The final workload was two independent pipelines. Each pipeline was:

```text
base -> mv1 -> mv2 -> mv3
```

Each base table used the built-in `datagen` source with:

```text
datagen.split.num = 256
```

Important: the configured source rate below is **per base table**. Since there are two independent pipelines, the total configured source rate is 2x the per-base value.

### SQL used for the final 2x3 workload

The final validation workload used two independent copies of the same 3-MV pipeline. The only difference between the two pipelines is the relation prefix (`p1` vs `p2`).

For one pipeline, the SQL shape is:

```sql
CREATE TABLE pr24951_zero_p1_base (
  id BIGINT,
  k BIGINT,
  v BIGINT
) WITH (
  connector = 'datagen',
  datagen.rows.per.second = '<RATE_PER_BASE>',
  datagen.split.num = '256',
  fields.id.kind = 'sequence', fields.id.start = '1', fields.id.end = '2000000000',
  fields.k.kind = 'sequence', fields.k.start = '1', fields.k.end = '2000000000',
  fields.v.kind = 'sequence', fields.v.start = '1', fields.v.end = '2000000000'
) FORMAT NATIVE ENCODE NATIVE;

CREATE MATERIALIZED VIEW pr24951_zero_p1_mv1 AS
SELECT a.k, COUNT(*) AS cnt, SUM(a.v + b.v) AS s
FROM pr24951_zero_p1_base AS a
JOIN pr24951_zero_p1_base AS b ON a.k = b.k
GROUP BY a.k;

CREATE MATERIALIZED VIEW pr24951_zero_p1_mv2 AS
SELECT a.k, COUNT(*) AS cnt, SUM(a.s + b.s) AS s
FROM pr24951_zero_p1_mv1 AS a
JOIN pr24951_zero_p1_mv1 AS b ON a.k = b.k
GROUP BY a.k;

CREATE MATERIALIZED VIEW pr24951_zero_p1_mv3 AS
SELECT a.k, COUNT(*) AS cnt, SUM(a.s + b.s) AS s
FROM pr24951_zero_p1_mv2 AS a
JOIN pr24951_zero_p1_mv2 AS b ON a.k = b.k
GROUP BY a.k;
```

The second pipeline used the same SQL with names changed to:

```text
pr24951_zero_p2_base
pr24951_zero_p2_mv1
pr24951_zero_p2_mv2
pr24951_zero_p2_mv3
```

After creation, every table and MV was explicitly set to parallelism 256:

```sql
ALTER TABLE pr24951_zero_p1_base SET PARALLELISM = 256;
ALTER MATERIALIZED VIEW pr24951_zero_p1_mv1 SET PARALLELISM = 256;
ALTER MATERIALIZED VIEW pr24951_zero_p1_mv2 SET PARALLELISM = 256;
ALTER MATERIALIZED VIEW pr24951_zero_p1_mv3 SET PARALLELISM = 256;

ALTER TABLE pr24951_zero_p2_base SET PARALLELISM = 256;
ALTER MATERIALIZED VIEW pr24951_zero_p2_mv1 SET PARALLELISM = 256;
ALTER MATERIALIZED VIEW pr24951_zero_p2_mv2 SET PARALLELISM = 256;
ALTER MATERIALIZED VIEW pr24951_zero_p2_mv3 SET PARALLELISM = 256;
```

For the zero-data runs, `<RATE_PER_BASE>` was `0`. For the data-rate runs, `<RATE_PER_BASE>` was `10000` or `50000`. Since the workload has two base tables, the total configured source rate was `2 x <RATE_PER_BASE>`.

`MV input rows/s` in the result tables is summed across MV layers and both pipelines. It is not the final output throughput. The key completed-throughput comparison is `actual source rows/s`.

### Metrics collected

For each run I collected:

- actual source rows/s
- MV input rows/s
- compute CPU cores
- max compute pod CPU
- compactor CPU cores
- gRPC MB/s
- exchange send/recv MB/s
- meta barrier p95/p99/p999/max
- stream barrier inflight p95/p99/p999/max
- stream barrier sync-storage p95/p99/p999/max

The `max` value is the maximum non-empty Prometheus histogram bucket upper bound, not the raw maximum event duration.

---

## Result 1: same-throughput comparison at per-base rate = 10k

This is the cleanest CPU-saving comparison. The total configured source rate was 20k rows/s, and both images completed approximately the same throughput.

Grafana windows:

- PR: [2026-06-05 19:18-19:24 UTC](https://grafana.test.risingwave-cloud.xyz/d/EpkBw5W4k/risingwave-dev-dashboard?orgId=1&timezone=browser&var-datasource=f2ad7358-1f42-73f7-8dc9-a4517433b212&var-namespace=rwc-g1jq8m2k0geiracl1budtebbum-pr24951-exchange-sql&var-cluster=test-useast1-eks-a&var-instance=risingwave&var-pod=$__all&var-component=$__all&var-table=$__all&from=1780687080000&to=1780687440000)
- Baseline: [2026-06-05 19:43-19:50 UTC](https://grafana.test.risingwave-cloud.xyz/d/EpkBw5W4k/risingwave-dev-dashboard?orgId=1&timezone=browser&var-datasource=f2ad7358-1f42-73f7-8dc9-a4517433b212&var-namespace=rwc-g1jq8m2k0geiracl1budtebbum-pr24951-exchange-sql&var-cluster=test-useast1-eks-a&var-instance=risingwave&var-pod=$__all&var-component=$__all&var-table=$__all&from=1780688580000&to=1780689000000)

| image | actual source rows/s | MV input rows/s | compute CPU | max pod CPU | compactor CPU | gRPC MB/s | exchange send/recv MB/s |
|---|---:|---:|---:|---:|---:|---:|---:|
| PR `2711dba` | 19,999 | 79,935 | 78.2 cores | 10.3 / 16 | 1.68 cores | 38.4 | 18.1 / 17.5 |
| baseline `nightly-20260530` | 20,024 | 79,699 | 112.8 cores | 14.9 / 16 | 1.86 cores | 193.9 | 53.6 / 44.4 |

Barrier latency:

| image | meta barrier p95 / p99 / p999 / max | inflight p95 / p99 / p999 / max | sync-storage p95 / p99 / p999 / max |
|---|---|---|---|
| PR | 1.19s / 2.18s / 2.52s / 2.56s | 1.10s / 1.68s / 2.47s / 2.56s | 0.221s / 0.307s / 0.404s / 0.506s |
| baseline | 5.48s / 7.72s / 8.56s / 8.65s | 5.04s / 6.98s / 8.48s / 8.65s | 0.284s / 0.334s / 0.608s / 0.759s |

Conclusion from this run:

- At the same completed throughput (~20k source rows/s), the PR reduces compute CPU from 112.8 cores to 78.2 cores.
- This is a reduction of ~34.6 compute cores, or ~31% versus baseline.
- gRPC traffic drops from 193.9 MB/s to 38.4 MB/s, roughly an 80% reduction.
- Meta barrier p95 improves from 5.48s to 1.19s.
- The compactor is not the bottleneck in this comparison; compactor CPU is low and similar in both runs.

This is the strongest evidence that the PR improves the targeted high-parallelism barrier/control-plane bottleneck.

---

## Result 2: zero-data barrier/control-path fixed cost

I also ran zero-data tests with `datagen.rows.per.second = 0`. In these runs:

- source rows/s = 0
- MV rows/s = 0
- actor in/out rows/s = 0

So the remaining CPU and traffic are attributable to maintaining the active high-parallelism streaming graph and sending barriers/control-plane messages, not row processing.

### 5-layer single pipeline: `base -> mv1 -> mv2 -> mv3 -> mv4 -> mv5`

Grafana windows:

- PR: [2026-06-05 18:12-18:21 UTC](https://grafana.test.risingwave-cloud.xyz/d/EpkBw5W4k/risingwave-dev-dashboard?orgId=1&timezone=browser&var-datasource=f2ad7358-1f42-73f7-8dc9-a4517433b212&var-namespace=rwc-g1jq8m2k0geiracl1budtebbum-pr24951-exchange-sql&var-cluster=test-useast1-eks-a&var-instance=risingwave&var-pod=$__all&var-component=$__all&var-table=$__all&from=1780683120000&to=1780683660000)
- Baseline: [2026-06-05 18:26-18:35 UTC](https://grafana.test.risingwave-cloud.xyz/d/EpkBw5W4k/risingwave-dev-dashboard?orgId=1&timezone=browser&var-datasource=f2ad7358-1f42-73f7-8dc9-a4517433b212&var-namespace=rwc-g1jq8m2k0geiracl1budtebbum-pr24951-exchange-sql&var-cluster=test-useast1-eks-a&var-instance=risingwave&var-pod=$__all&var-component=$__all&var-table=$__all&from=1780683960000&to=1780684500000)

| image | compute CPU | meta barrier p95 | inflight p95 | gRPC MB/s | exchange send/recv MB/s |
|---|---:|---:|---:|---:|---:|
| PR | 27.8 cores | 0.50s | 0.33s | 25.6 | 12.2 / 11.9 |
| baseline | 57.7 cores | 0.75s | 0.73s | 147.6 | 38.7 / 31.4 |

### Two independent 3-MV pipelines

Grafana windows:

- PR: [2026-06-05 18:45-18:52 UTC](https://grafana.test.risingwave-cloud.xyz/d/EpkBw5W4k/risingwave-dev-dashboard?orgId=1&timezone=browser&var-datasource=f2ad7358-1f42-73f7-8dc9-a4517433b212&var-namespace=rwc-g1jq8m2k0geiracl1budtebbum-pr24951-exchange-sql&var-cluster=test-useast1-eks-a&var-instance=risingwave&var-pod=$__all&var-component=$__all&var-table=$__all&from=1780685100000&to=1780685520000)
- Baseline: [2026-06-05 18:56-19:03 UTC](https://grafana.test.risingwave-cloud.xyz/d/EpkBw5W4k/risingwave-dev-dashboard?orgId=1&timezone=browser&var-datasource=f2ad7358-1f42-73f7-8dc9-a4517433b212&var-namespace=rwc-g1jq8m2k0geiracl1budtebbum-pr24951-exchange-sql&var-cluster=test-useast1-eks-a&var-instance=risingwave&var-pod=$__all&var-component=$__all&var-table=$__all&from=1780685760000&to=1780686180000)

| image | compute CPU | max pod CPU | source/MV rows/s | gRPC MB/s | exchange send/recv MB/s |
|---|---:|---:|---:|---:|---:|
| PR | 36.65 cores | 4.64 / 16 | 0 / 0 | 26.0 | 12.1 / 11.8 |
| baseline | 67.63 cores | 8.56 / 16 | 0 / 0 | 181.9 | 47.7 / 38.7 |

Conclusion from zero-data tests:

- Even with no row processing, the baseline burns about 58-68 compute cores just maintaining the high-parallelism graph and barrier/control path.
- The PR reduces this fixed cost to about 28-37 compute cores.
- This proves that a substantial part of the improvement is from reducing barrier/control-plane overhead, not from changes in row processing, hash join, aggregation, or materialization.

---

## Result 3: saturated-capacity comparison at per-base rate = 50k

This run is not a same-throughput CPU-saving comparison. The configured total source rate was 100k rows/s, but neither image reached that rate. Both were close to the 128-core compute limit.

Grafana windows:

- PR: [2026-06-05 19:28-19:35 UTC](https://grafana.test.risingwave-cloud.xyz/d/EpkBw5W4k/risingwave-dev-dashboard?orgId=1&timezone=browser&var-datasource=f2ad7358-1f42-73f7-8dc9-a4517433b212&var-namespace=rwc-g1jq8m2k0geiracl1budtebbum-pr24951-exchange-sql&var-cluster=test-useast1-eks-a&var-instance=risingwave&var-pod=$__all&var-component=$__all&var-table=$__all&from=1780687680000&to=1780688100000)
- Baseline: [2026-06-05 19:53-20:00 UTC](https://grafana.test.risingwave-cloud.xyz/d/EpkBw5W4k/risingwave-dev-dashboard?orgId=1&timezone=browser&var-datasource=f2ad7358-1f42-73f7-8dc9-a4517433b212&var-namespace=rwc-g1jq8m2k0geiracl1budtebbum-pr24951-exchange-sql&var-cluster=test-useast1-eks-a&var-instance=risingwave&var-pod=$__all&var-component=$__all&var-table=$__all&from=1780689180000&to=1780689600000)

| image | actual source rows/s | MV input rows/s | compute CPU | max pod CPU | compactor CPU | gRPC MB/s | exchange send/recv MB/s |
|---|---:|---:|---:|---:|---:|---:|---:|
| PR | 40,120 | 164,537 | 121.9 cores | 15.9 / 16 | 6.79 cores | 49.4 | 23.5 / 22.8 |
| baseline | 30,900 | 124,829 | 118.9 cores | 15.5 / 16 | 3.48 cores | 174.1 | 53.4 / 46.1 |

Barrier latency:

| image | meta barrier p95 / p99 / p999 / max | inflight p95 / p99 / p999 / max | sync-storage p95 / p99 / p999 / max |
|---|---|---|---|
| PR | 42.33s / 43.50s / 43.76s / 43.79s | 41.75s / 43.38s / 43.75s / 43.79s | 0.458s / 0.808s / 1.20s / 1.71s |
| baseline | 32.16s / 41.46s / 43.56s / 43.79s | 28.96s / 38.31s / 43.24s / 43.79s | 0.393s / 0.498s / 0.721s / 0.759s |

Conclusion from this run:

- This run is saturated/backpressured for both images, so it should not be used to claim CPU savings at the same throughput.
- However, at similar compute CPU near the cluster limit, the PR processes about 40.1k source rows/s while the baseline processes about 30.9k source rows/s.
- That is about 30% higher actual throughput under saturation.
- gRPC traffic remains much lower in the PR: 49.4 MB/s versus 174.1 MB/s.
- Both images show very high barrier tail latency in this saturated regime.
- At saturation, the PR has higher p95 barrier latency while processing more rows, so this is a throughput-vs-tail-latency trade-off signal, not a clean latency win.

---

## Overall conclusion

The PR is strongly validated for its target bottleneck: high-parallelism barrier/control-plane overhead.

The strongest same-throughput result is:

- actual source throughput: ~20k rows/s on both images
- compute CPU: 112.8 cores -> 78.2 cores
- CPU reduction: ~34.6 cores, ~31%
- gRPC traffic: 193.9 MB/s -> 38.4 MB/s
- meta barrier p95: 5.48s -> 1.19s

The zero-data runs further show that the improvement is not an artifact of row processing: even with 0 source rows/s and 0 MV rows/s, the PR saves roughly 30 compute cores over baseline.

Under saturation, the PR also pushes more completed throughput at similar compute CPU: about 40.1k source rows/s versus 30.9k for baseline.

### Caveats

This does not prove the PR is “perfect” or regression-free in all cases. The validation has the following limits:

- The workload is synthetic datagen + symmetric MV chains, not a real customer workload.
- The final results are single 3-minute windows, without N>=3 repeats or error bars.
- The workload uses high parallelism (`parallelism=256`); it does not characterize low-parallelism cases.
- Failure and upgrade paths were not covered: rolling upgrade, mixed-version operation, actor restart, node restart, network hiccups, and recovery were not tested here.
- In the saturated rate=50k run, both images have very high barrier tail latency; the PR has higher p95 barrier latency there while also processing more rows, so that run should be treated as a throughput/capacity trade-off, not a clean latency win.
- One zero-data run showed PR-side meta/sync-storage tail outliers; this needs repeated runs or log correlation before concluding whether it is noise or a real tail-latency regression.

A precise summary would be:

> This PR is clearly effective for the targeted high-parallelism barrier/control-plane bottleneck. In the clean same-throughput test it reduced compute CPU by ~31%, reduced gRPC traffic by ~80%, and improved barrier p95 by ~4.6x. Zero-data tests confirm that the improvement comes from reducing barrier/control fixed overhead. Saturated tests show higher throughput at similar CPU. Remaining caveats are workload coverage, lack of repeated windows/error bars, and untested failure/upgrade paths.

hzxa21:
@chenzl25 @zwang28 you may be interested at the results above: https://github.com/risingwavelabs/risingwave/pull/24951#issuecomment-4637583312

chenzl25:
**Recommend holding this PR — one correctness blocker + one compat risk.**

**1. (Blocker) `CountingMergeInput` can yield E+1 data ahead of barrier E.** With no pending barrier, `run_counting_merge` races `barrier_rx` vs `inner.next()` and yields data immediately if data wins (`multiplexed.rs:420`). The count gate only ensures barrier E is not delivered too early; it does not stop next-epoch data from overtaking barrier E. This breaks the required `E data -> barrier E -> E+1 data` ordering.

**2. (Risk) `up_actor_ids` is a new protocol field with no mixed-version handling.** During rolling upgrade, a new downstream talking to an old upstream can have this field ignored and the request handled as a legacy single-actor stream, routing to the wrong/absent receiver. PR body already lists mixed-version as a caveat; this needs version negotiation, fallback, or an explicit upgrade block.

**Minor:** On recovery/scaling, is the first barrier's `expected_count == 0` assertion guaranteed when an actor joins the coalescer mid-stream? Otherwise it panics instead of failing gracefully.

lmatz:
> CountingMergeInput can yield E+1 data ahead of barrier E. With no pending barrier, run_counting_merge races barrier_rx vs inner.next() and yields data immediately if data wins (multiplexed.rs:420). The count gate only ensures barrier E is not delivered too early; it does not stop next-epoch data from overtaking barrier E. This breaks the required E data -> barrier E -> E+1 data ordering.

True, I mistook which part of the rows needs to be counted in this case.


Currently, it is counting the number of rows between the current barrier and the barrier alignment, while it should be counting the number of rows between the current barrier alignment in the upstream and the next barrier arrival, so that the downstream knows how much data to be processed before stopping, as the true barrier hasn't been sent by the upstream yet.
So counting does not work.


Two possible solutions I can think of are:
1. either strictly align both barriers and data in the upstream and thus disable the data pipelining completely, i.e., not sending data from upstream to downstream before sending the barrier.
2. or send a barrier-like thing only when there is a new row coming after the barrier alignment and before the next barrier arrival along the data path. But this essentially brings back M x M barriers if there is always data coming in to each actor, which cannot be good in reality. So rule out (2).

(1) has the downside of no data pipelining and the benefit of lower CPU usage. If the cluster has no CPU saturation, the throughput can only be lower. 
Will benchmark the saturation case to see if it can be a special optimization.

------
maybe a special scheduling strategy to reduce M while maintaining the amount of CPU that each fragment can use the same.
Since the more exchanges/fragments, the more overhead there is, if let's say there are 8 fragments, instead of scheduling each fragment on all the machines, it may make sense to schedule each fragment only on 1/8 of the machines in the cluster.

PR #24968, opened 2026-03-05
fix(stream): sync ops with visibility after noop elimination in HashDataDispatcher

jw-itq:

Fix a bug where [HashDataDispatcher](cci:2://file:///Users/shiwanming/job/open-source-project/risingwave/src/stream/src/executor/dispatch.rs:878:0-886:1) uses pre-elimination ops with post-elimination visibility, causing downstream panic: `expect a U- before U+`.

**Root Cause:**

In [HashDataDispatcher::dispatch_data](cci:1://file:///Users/shiwanming/job/open-source-project/risingwave/src/stream/src/executor/dispatch.rs:1219:4-1228:5), `new_ops` is computed **before** [output_mapping.apply()](cci:1://file:///Users/shiwanming/job/open-source-project/risingwave/src/stream/src/executor/dispatch/output_mapping.rs:58:4-99:5), which internally calls [eliminate_adjacent_noop_update()](cci:1://file:///Users/shiwanming/job/open-source-project/risingwave/src/common/src/array/stream_chunk.rs:329:4-383:5). This function may:
1. Hide rows by setting visibility to false (noop pairs)
2. Normalize partially-visible U-/U+ pairs (U- → Delete, U+ → Insert) via #24659

However, `combined_vis` is computed **after** the elimination, incorporating the new visibility. The resulting chunk uses stale `new_ops` (without normalize) but updated `combined_vis` (with normalize), creating a mismatch — orphan `UpdateInsert` ops become visible without a preceding `UpdateDelete`, causing downstream iterators to panic.

**Fix:**

After [output_mapping.apply()](cci:1://file:///Users/shiwanming/job/open-source-project/risingwave/src/stream/src/executor/dispatch/output_mapping.rs:58:4-99:5), merge `new_ops` (which contains dist-key-changed rewrites) with [chunk.ops()](cci:1://file:///Users/shiwanming/job/open-source-project/risingwave/src/common/src/array/stream_chunk.rs:261:4-263:5) (which contains post-elimination normalized ops). This ensures ops and visibility are always consistent.

**How to Reproduce:**

When the same stream key has multiple consecutive updates within a single chunk **and** the dispatcher's output columns are fewer than input columns (triggering noop elimination), the middle U-/U+ pair gets eliminated, leaving the outer pair's ops un-normalized.

Example:
U- row_A (visible) U+ row_B (eliminated by noop) U- row_B (eliminated by noop) U+ row_C (visible, but ops still says UpdateInsert → panic)

**Related:**
- #24580 — introduced `combined_vis` but did not sync ops
- #24645 — removed adjacent U-/U+ assertion
- #24659 — added normalize in [eliminate_adjacent_noop_update](cci:1://file:///Users/shiwanming/job/open-source-project/risingwave/src/common/src/array/stream_chunk.rs:329:4-383:5), but [HashDataDispatcher](cci:2://file:///Users/shiwanming/job/open-source-project/risingwave/src/stream/src/executor/dispatch.rs:878:0-886:1) ignores it

## Checklist

## Documentation

<details>

Fix a critical bug introduced in v2.8.0 where compute nodes crash with `expect a U- before U+` panic when processing streams with consecutive updates on the same key through [HashDataDispatcher](cci:2://file:///Users/shiwanming/job/open-source-project/risingwave/src/stream/src/executor/dispatch.rs:878:0-886:1) with column pruning. This is caused by ops/visibility mismatch after noop update elimination in the dispatcher.

</details>


hzxa21:
Good catch. Can you write a e2e slt test? You can use `noop_update_elimination.slt` as a reference.

chenzl25:
Let's add this test, thanks.

```rust
    #[tokio::test]
    async fn test_hash_dispatcher_internal_projection_normalize_single_u_plus() {
        let (output_tx, mut output_rx) = channel_for_test();
        let outputs = vec![Output::new(ActorId::new(1), output_tx)];
        let hash_mapping = vec![ActorId::new(1); VirtualNode::COUNT_FOR_TEST];
        let mut hash_dispatcher = HashDataDispatcher::new(
            outputs,
            vec![0],
            DispatchOutputMapping::Simple(vec![0]),
            hash_mapping,
            0.into(),
        );

        let input = StreamChunk::from_pretty(
            "  I I
            + 1 10
            U- 1 10
            U+ 1 30",
        );

        hash_dispatcher.dispatch_data(input).await.unwrap();

        let output = output_rx.recv().await.unwrap();
        assert_eq!(
            *output.as_chunk().unwrap(),
            StreamChunk::from_pretty(
                "  I
                + 1 D
                U- 1 D
                + 1",
            )
        );
    }
```

chenzl25:
The e2e test actually failed to reproduce the bug, so I think we'd better use the unit test above instead.

e2e_test/streaming/noop_update_elimination_ops_sync.slt
```diff
@@ -0,0 +1,145 @@
+# Test that ops are correctly synced with visibility after noop update elimination
```
chenzl25: Please remove this e2e test since it actually can't reproduce the bug.

PR #24979, opened 2026-03-05
feat(config): remove session variables for rate limit

### LINKED ISSUE #17954: remove session variables for rate limit
We can just set it locally to the ddl.

faketygg:

Remove the four session-level rate limit variables (`backfill_rate_limit`, `source_rate_limit`, `dml_rate_limit`, `sink_rate_limit`) from `SessionConfig`. Rate limits should now be set locally on each DDL statement via:

1. **WITH clause** at creation time: `CREATE MATERIALIZED VIEW mv WITH (backfill_rate_limit = 1) AS ...`
2. **ALTER DDL** for existing objects: `ALTER MATERIALIZED VIEW mv SET BACKFILL_RATE_LIMIT TO 1`

This removes the global session-wide defaults in favor of per-object rate limit configuration, making the behavior more explicit and less error-prone.

### Changes:
- **Core**: Removed 4 session variables and their constants from `SessionConfig`
- **Core**: Simplified `OverwriteOptions` to only read from WITH clause (no session fallback)
- **Tests**: Updated ~40 test files (e2e + simulation) to use WITH clause instead of SET session variable
- **CI**: Updated `run-backfill-tests.sh` to use WITH clause

Fixes #17954.

## Checklist

## Test plan

All existing rate limit tests have been updated to use WITH clause instead of session variables. The test coverage remains the same — only the mechanism for setting rate limits has changed.

— Kodomo

### DIFF
--- a/src/tests/simulation/tests/integration_tests/recovery/locality_backfill.rs
+++ b/src/tests/simulation/tests/integration_tests/recovery/locality_backfill.rs
@@ -22,8 +22,7 @@ const SET_LOCALITY_BACKFILL: &str = "SET enable_locality_backfill = true;";
 const SET_BACKGROUND_DDL: &str = "SET background_ddl = true;";
 const CREATE_TABLE: &str = "CREATE TABLE t(a int);";
 const SEED_TABLE: &str = "INSERT INTO t SELECT * FROM generate_series(1, 10000, 1);";
-const CREATE_MV: &str =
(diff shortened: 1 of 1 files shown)


hzxa21:
This is a breaking change. I think session level rate limit won't cause much maintenance burden and confusion in usability. Let's close this PR.

Issue #15596, opened 2024-03-11
notice user about cascading drops when dropping a relation


BugenZhao:
In Postgres, when there are dependent relations being dropped while executing `DROP .. CASCADE`, we'll get their names through the notices:

```
postgres=# drop table t33 cascade;
NOTICE:  drop cascades to view v33
DROP TABLE
```

That's more user friendly. It'll be good for us to provide such feature. Better to be asynchronous and reflect the progress of the dropped relations.

PR #24980, opened 2026-03-05
feat: notice user about cascading drops when dropping a table

Standing-Man:

- Closes part of #15596.

This PR adds a message during cascading deletes to inform users which related objects are removed as part of the cascade. Currently, this implementation does not support asynchronous execution or displaying the progress of deleting relations. 

⚠️  : The cascading delete notice currently only supports tables, and can be extended to other object types in the future.

```sql
dev=> CREATE TABLE orders (
    order_id BIGINT,
    user_id BIGINT,
    amount DECIMAL
);
CREATE_TABLE
dev=> CREATE VIEW order_view AS SELECT * FROM orders;
CREATE_VIEW
dev=>CREATE MATERIALIZED VIEW order_mv AS 
SELECT user_id, SUM(amount) AS total_amount FROM orders 
GROUP BY user_id;
CREATE_MATERIALIZED_VIEW
dev=> DROP TABLE orders CASCADE;
NOTICE:  drop cascades to 2 other objects
DETAIL:  drop cascades to materialized view public.order_mv
drop cascades to view public.order_view
DROP_TABLE
```

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/proto/ddl_service.proto
+++ b/proto/ddl_service.proto
@@ -16,6 +16,16 @@ message WaitVersion {
   uint64 hummock_version_id = 2;
 }
 
+message CascadeObject {
+  common.ObjectType object_type = 1;
+  optional string schema_name = 2;
+  string object_name = 3;
+}
+
+message DropResult {
+  repeated CascadeObject cascade_objects = 1;
+}
+
 message CreateDatabaseRequest {
   catalog.Database db = 1;
 }
@@ -441,6 +451,7 @@ message DropTableRequest {
 message DropTableResponse {
   common.Status status = 1;
   WaitVersion version = 2;
+  DropResult drop_result = 3;
 }
 
 // Used by risectl (and in the future, dashboard)
--- a/src/frontend/src/catalog/mod.rs
+++ b/src/frontend/src/catalog/mod.rs
@@ -107,8 +107,26 @@ pub fn check_schema_writable(schema: &str) -> Result<()> {
 }
 
 pub(crate) fn notice_drop_cascade_objects(objects: &[CascadeObject]) {
-    for object in objects {
-        notice_to_user(format_drop_cascade_object(object));
+    if let Some(notice) = format_drop_cascade_notice(objects) {
+        notice_to_user(notice);
+    }
+}
+
+fn format_drop_cascade_notice(objects: &[CascadeObject]) -> Option<String> {
+    match objects {
+        [] => None,
+        [object] => Some(format_drop_cascade_object(object)),
+        objects => {
+            let detail = objects
+                .iter()
+                .map(format_drop_cascade_object)
+                .collect::<Vec<_>>()
+                .join("\n");
+            Some(format!(
+                "drop cascades to {} other objects\nDETAIL:  {detail}",
+                objects.len()
+            ))
+        }
     }
 }
 
@@ -273,6 +291,30 @@ mod tests {
                 object_name: "select".to_owned(),
             }),
             "drop cascades to view \"My Schema\".\"select\""
+        );
+    }
+
+    #[test]
+    fn test_format_multiple_drop_cascade_objects() {
+        let objects = [
+            CascadeObject {
+                object_type: PbObjectType::View as _,
+                schema_name: Some("public".to_owned()),
+                object_name: "order_view".to_owned(),
+            },
+            CascadeObject {
+                object_type: PbObjectType::Mview as _,
+                schema_name: Some("public".to_owned()),
+                object_name: "order_mv".to_owned(),
+            },
+        ];
+
+        assert_eq!(
+            format_drop_cascade_notice(&objects).as_deref(),
+            Some(
+                "drop cascades to 2 other objects\nDETAIL:  drop cascades to view public.order_view\n\
+                 drop cascades to materialized view public.order_mv"
+            )
         );
     }
 }
--- a/src/frontend/src/catalog/mod.rs
+++ b/src/frontend/src/catalog/mod.rs
@@ -23,10 +23,14 @@ use risingwave_common::catalog::{
 };
 use risingwave_common::error::code::PostgresErrorCode;
 use risingwave_connector::sink::catalog::SinkCatalog;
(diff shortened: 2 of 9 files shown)


proto/ddl_service.proto
```diff
@@ -425,6 +438,7 @@ message DropTableRequest {
 message DropTableResponse {
   common.Status status = 1;
   WaitVersion version = 2;
+  repeated CascadeObject cascade_objects = 3;
```
hzxa21: > > Cascade deletion notifications are implemented only for tables as an experimental feature.

I suggest you only keep this pb change and leave the other changes to further PR when supporting other db objects.

Standing-Man: Ok, I have a question: can I use SLT testing to verify the SQL output of cascading deletes? If not, how should I test it? Regression test?

hzxa21: Need to modify [sqllogictest-rs](https://github.com/risinglightdb/sqllogictest-rs) in order to test NOTICE output in SLT test.

I think this feat is straight-forward so testing it with only UT should be fine.

Standing-Man: From what I can see, `MockCatalogWriter` does not support [cascading deletes for tables](https://github.com/risingwavelabs/risingwave/blob/main/src/frontend/src/test_utils.rs#L463-L469). 
If the cascading delete behavior for tables is correct, then the generated notice should also be correct.
Additionally, I have included an example in the PR description to show the expected output and verify that it is consistent with PostgreSQL.

The cascading delete support in `MockCatalogWriter` can be added in a future update.

could you help trigger the workflows? Thank you!

hzxa21: SGTM

hzxa21 (approved):
LGTM

src/meta/service/src/ddl_service.rs
```diff
@@ -710,6 +710,15 @@ impl DdlService for DdlServiceImpl {
         let source_id = request.source_id;
         let table_id = request.table_id;
 
+        let cascade_objects = if request.cascade {
```
yezizp2012: I think it's better to get the cascade objects in https://github.com/risingwavelabs/risingwave/blob/53f4265a0f0eb11c9b7f4f339f7ac8106d5f67e3/src/meta/src/controller/catalog/drop_op.rs#L25. Not only for improved performance and consistency, but this logic can also be reused for other drop DDLs.

yezizp2012 (commented):
Rest LGTM

src/meta/service/src/ddl_service.rs
Standing-Man: I’ve moved the cascade object collection logic into `drop_op` so that the cascade objects are computed only once, avoiding redundant work and improving performance.

src/frontend/src/catalog/mod.rs
```diff
@@ -102,6 +106,57 @@ pub fn check_schema_writable(schema: &str) -> Result<()> {
...
+        [] => None,
+        [object] => Some(format_drop_cascade_object(object)),
+        objects => {
+            let detail = objects
+                .iter()
+                .map(format_drop_cascade_object)
+                .collect::<Vec<_>>()
+                .join("\n");
+            Some(format!(
+                "drop cascades to {} other objects\nDETAIL:  {detail}",
```
Standing-Man, line 126: Currently, the `NOTICE` and `DETAIL` messages are simply concatenated and displayed together. There is room for improvement here: we could extend `pgwire‘s ErrorOrNoticeMessage` to include a dedicated `DETAIL` field. I’ll open a separate PR to address this enhancement.

Standing-Man:
Hi @wenym1 and @hzxa21, This PR has been open for a little while. If you happen to have some time, I’d greatly appreciate a review. Please let me know if there are any concerns or additional changes I should make. Thanks!

hzxa21:
@Standing-Man Thanks for the heads up. I thought this PR has already been merged since my last approval. We will get back to you for the review.

src/frontend/src/catalog/mod.rs
```diff
@@ -102,6 +106,57 @@ pub fn check_schema_writable(schema: &str) -> Result<()> {
...
+        PbObjectType::Mview => "materialized view",
+        PbObjectType::Source => "source",
+        PbObjectType::Sink => "sink",
+        PbObjectType::View => "view",
+        PbObjectType::Index => "index",
+        PbObjectType::Function => "function",
+        PbObjectType::Connection => "connection",
+        PbObjectType::Subscription => "subscription",
+        PbObjectType::Secret => "secret",
+        PbObjectType::Unspecified => "object",
```
Standing-Man, line 147: I’m wondering whether it’s appropriate to display `PbObjectType::Unspecified` as object in the error message.

src/meta/src/controller/catalog/drop_op.rs
```diff
@@ -95,6 +119,21 @@ impl CatalogController {
             },
         };
 
+        let cascade_objects =
+            if drop_mode == DropMode::Cascade && object_type != ObjectType::Database {
+                Some(
+                    build_cascade_objects_for_notice(
+                        object_id,
+                        &implicit_object_ids,
+                        &removed_objects,
```
hzxa21: This notice list is built before the sink-into-table special case below extends `removed_objects` from `Sink::TargetTable` (around lines 171-201). A sink that is only discovered through that path will still be dropped, but it will not be included in `DropResult.cascade_objects`, so `DROP TABLE ... CASCADE` can remove a user-visible sink without mentioning it in the notice.

Can we build `cascade_objects` after those late additions, or explicitly append the late-discovered sink objects before returning the drop result?

hzxa21 (changes requested):
Requesting changes for the cascade notice correctness issue.

src/meta/src/controller/catalog/drop_op.rs
Standing-Man: Good catch! I’ve updated the implementation so that cascade object notices are built after discovering sinks through `Sink::TargetTable`.

PR #24983, opened 2026-03-05
fix(meta): ignore stale refresh finish events after source drop

tabVersion:
## Summary
- handle stale refresh completion events (`ListFinish`/`LoadFinish`) when the associated source is already dropped
- treat `CatalogIdNotFound(object, ..)` from `get_object_database_id(associated_source_id)` as idempotent skip instead of bubbling to barrier completion failure
- keep other error types unchanged

## Why
In `REFRESH TABLE` + immediate `DROP TABLE` races, refresh finish events can arrive after catalog cleanup. Previously this propagated as:
`Failed to get database id for table -> object id not found`, which could block command completion and trigger recovery noise.

## Tests
- `cargo test -p risingwave_meta --lib skip_refresh_finish_when_associated_source_missing`
- `cargo test -p risingwave_meta --lib do_not_skip_refresh_finish_for_other_not_found_types`

### DIFF
--- a/e2e_test/source_inline/kafka/alter/rate_limit_source_kafka_shared.slt.serial
+++ b/e2e_test/source_inline/kafka/alter/rate_limit_source_kafka_shared.slt.serial
@@ -108,9 +108,8 @@ order by name;
 kafka_source	SOURCE	{SOURCE}	1000
 rl_mv2	SOURCE_BACKFILL	{SOURCE_SCAN}	0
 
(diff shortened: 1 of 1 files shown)


src/stream/src/executor/source/source_backfill_executor.rs
shanicky, line 329: can we use some struct to replace the `false,true`?

PR #25041, opened 2026-03-10
fix(meta): commit transaction after aborting subscription creation

Li0k:
## Summary

Fix a bug where `try_abort_creating_subscription` was not committing the transaction after deleting a subscription, causing the deletion to not be persisted.

## Changes

- Add `txn.commit().await?` after `subscription.delete(&txn).await?` in `try_abort_creating_subscription`
- Add integration test `test_abort_creating_subscription_commits_delete` to verify the fix

## Checklist

- [x] Integration test added


yezizp2012 (approved):
LGTM!

PR #25043, opened 2026-03-10
feat(frontend): add iceberg engine storage selection

silver-ymz:

Iceberg engine tables in RisingWave currently have two storage paths for batch reads:

- Hummock row store, which is better for point reads and highly selective access
- Iceberg columnar storage, which is better for wide scans and analytical reads

Before this change, users could not explicitly control which path a batch query should use, and the planning flow for Iceberg engine tables was also harder to extend for cost-based or heuristic storage selection.

This PR introduces **configurable storage selection** for batch queries on Iceberg engine tables, so users can either force a storage mode or let the optimizer choose one automatically based on the query shape.

### User-facing behavior

This PR adds a session config:

- `iceberg_engine_storage_mode`

`iceberg_engine_storage_mode` supports:

- `iceberg`: always use Iceberg columnar scan for batch `SELECT`
- `hummock`: always use Hummock table scan for batch `SELECT`
- `auto`: let the optimizer choose the storage path

In `auto` mode, the current heuristic prefers Hummock when:

- the query is effectively a point lookup on the table primary key

### Detailed Update

Included in this PR:

- new session config for Iceberg engine storage selection
- heuristic-based storage selection rule for batch queries in `auto` mode
- planner/optimizer refactor so LogicalIcebergIntermediateScan can store more infomation about hummock table
- support for preserving predicate pushdown and schema/type alignment when switching between Iceberg and Hummock plans
- e2e coverage for config parsing, explain output, and query correctness across `iceberg`, `hummock`, and `auto` modes

Not in scope for this PR (Maybe we can do it in the future):

- a full cost-based storage selection model
- DataFusion engine for hummock table scan

## Checklist

## Documentation

<details>

RisingWave now supports configurable storage selection for batch queries on Iceberg engine tables.

Users can choose whether batch `SELECT` queries read from Iceberg columnar storage, Hummock row storage, or let the optimizer decide automatically through the new session setting `iceberg_engine_storage_mode`. In `auto` mode, RisingWave will currently prefer Hummock for point lookups on primary keys.

This is useful for mixed workloads on Iceberg engine tables, especially when users want to tune or debug query performance for selective lookups versus wider scans.

</details>

### DIFF
--- a/src/frontend/src/optimizer/rule/iceberg_engine_storage_selection_rule.rs
+++ b/src/frontend/src/optimizer/rule/iceberg_engine_storage_selection_rule.rs
@@ -24,6 +24,7 @@ use risingwave_common::session_config::IcebergQueryStorageMode;
 use super::prelude::{PlanRef, *};
 use crate::TableCatalog;
 use crate::catalog::source_catalog::SourceCatalog;
+use crate::optimizer::plan_node::generic::GenericPlanRef;
 use crate::optimizer::plan_node::{Logical, LogicalIcebergIntermediateScan, LogicalScan, generic};
 use crate::optimizer::rule::InfallibleRule;
 use crate::session::SessionImpl;
@@ -43,24 +44,12 @@ impl InfallibleRule<Logical> for IcebergEngineStorageSelectionRule {
         let source_catalog = scan.source_catalog()?;
         let table = get_table_from_iceberg_source(session, source_catalog)?;
 
-        macro_rules! try_strategy {
-            ($($strategy:ident),*) => {
-                'select_strategy: {
-                    $(
-                        if $strategy(session, scan, &table) {
-                            break 'select_strategy true;
-                        }
-                    )*
-                    false
-                }
-            };
-        }
-        let prefer_rowstore = try_strategy!(check_point_lookup);
+        let prefer_rowstore = check_point_lookup(scan, &table);
         if !prefer_rowstore {
             return None;
         }
 
-        rewrite_to_table_scan(&plan, scan, &table)
+        rewrite_to_table_scan(scan, &table)
     }
 }
 
@@ -72,7 +61,6 @@ impl IcebergEngineStorageSelectionRule {
 
 /// Rewrite the intermediate Iceberg scan to a Hummock `LogicalScan`.
 fn rewrite_to_table_scan(
-    plan: &PlanRef,
     scan: &LogicalIcebergIntermediateScan,
     table: &Arc<TableCatalog>,
 ) -> Option<PlanRef> {
@@ -91,7 +79,7 @@ fn rewrite_to_table_scan(
         table.clone(),
         vec![],
         vec![],
-        plan.ctx(),
+        scan.ctx(),
         scan.hummock_rewrite.origin_condition.clone(),
         scan.core.as_of.clone(),
     );
@@ -114,11 +102,7 @@ fn get_table_from_iceberg_source(
 /// Returns `true` when the predicate has equality-to-constant conditions on
 /// *all* PK columns of the table, making this a point lookup that benefits
 /// from the row store's key-value access pattern.
-fn check_point_lookup(
-    _session: &SessionImpl,
-    scan: &LogicalIcebergIntermediateScan,
-    table: &TableCatalog,
-) -> bool {
+fn check_point_lookup(scan: &LogicalIcebergIntermediateScan, table: &TableCatalog) -> bool {
     let pk_column_names: HashSet<&str> = table.pk_column_names().into_iter().collect();
     if pk_column_names.is_empty() {
         return false;


silver-ymz:
* **#25043** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25043?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25043?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/frontend/src/optimizer/plan_node/logical_iceberg_intermediate_scan.rs
```diff
@@ -50,79 +56,115 @@ use crate::utils::{ColIndexMapping, Condition, to_iceberg_predicate};
...
+    // They will be added together during predicate pushdown. The input index in origin_condition
+    // is based on table columns rather than output columns, and the column mapping is stored in output_column_mapping.
+    #[educe(Hash(ignore))]
+    pub iceberg_predicate: Predicate,
     #[educe(Hash(ignore))]
-    pub predicate: Predicate,
-    pub output_columns: Vec<String>,
+    pub origin_condition: Condition,
+    #[educe(Hash(ignore))]
+    pub output_column_mapping: ColIndexMapping,
```
chenzl25: Any reason we add `#[educe(Hash(ignore))]` to all these fields? Is it because `Predicate` doesn't impl `Hash` trait?

src/common/src/session_config/mod.rs
```diff
@@ -109,6 +111,11 @@ pub struct SessionConfig {
     #[parameter(default = QueryMode::default())]
     query_mode: QueryMode,
 
+    /// For Iceberg engine tables, which storage to use for batch SELECT: Iceberg (columnar) or
+    /// Hummock (row). Only affects batch SELECT on tables with ENGINE = ICEBERG.
+    #[parameter(default = IcebergEngineStorageMode::default())]
+    iceberg_engine_storage_mode: IcebergEngineStorageMode,
```
chenzl25: How about just using `query_storage_mode`? `iceberg_engine_storage_mode` sounds like we can change the storage of iceberg engine.

src/frontend/src/optimizer/rule/iceberg_engine_storage_selection_rule.rs
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    let table_scan = generic::TableScan::new(
+        output_col_idx,
+        table.clone(),
+        vec![],
+        vec![],
+        plan.ctx(),
+        scan.risingwave_condition.clone(),
+        scan.core.as_of.clone(),
```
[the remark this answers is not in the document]
chenzl25: How to fix it?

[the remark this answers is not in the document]
silver-ymz: Maybe we can add another `LogicalFilter` outside of `LogicalScan`

src/frontend/src/optimizer/plan_node/logical_iceberg_intermediate_scan.rs
silver-ymz: `Predicate` (from the iceberg crate) don't implement Hash. We could work around this with wrapper types, but `base` and `core` already provide sufficient discrimination for the optimizer's dedup — two nodes with different predicates or column mappings will almost certainly also differ in base/core.

So the extra complexity isn't worth it.

src/frontend/src/optimizer/rule/iceberg_engine_storage_selection_rule.rs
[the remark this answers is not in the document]
silver-ymz: add an initial column mapping at `relation.rs`

src/common/src/session_config/mod.rs
silver-ymz: update to `iceberg_query_storage_mode`

src/frontend/src/optimizer/rule/iceberg_engine_storage_selection_rule.rs
```diff
@@ -0,0 +1,141 @@
...
+        .0
+        .iter()
+        .copied()
+        .try_collect()?;
+    let table_scan = generic::TableScan::new(
+        output_col_idx,
+        table.clone(),
+        vec![],
+        vec![],
+        plan.ctx(),
```
chenzl25: It seems we don't need `plan` as a parameter, because we can use the `scan.ctx()` directly.

silver-ymz: done

PR #25047, opened 2026-03-10
refactor(meta): remove vnode-aligned level size threshold

Li0k:
## Summary
- remove the `vnode_aligned_level_size_threshold` capability from meta compaction scheduling
- stop exposing the threshold in config and `risectl`
- keep proto field `31` as deprecated compatibility placeholder and ignore it on the server side

## Validation
- `./risedev c`
- `cargo test -p risingwave_common config::tests::test_example_up_to_date -- --exact`


Li0k:
after https://github.com/risingwavelabs/risingwave/pull/24176

PR #25050, opened 2026-03-10
fix: unify and simplify streaming parallelism configuration

shanicky:

- This PR unifies user-facing streaming parallelism configuration around `streaming_parallelism` and per-type overrides such as `streaming_parallelism_for_table` / `streaming_parallelism_for_materialized_view`, and removes the legacy `streaming_parallelism_strategy{_for_*}` and cluster-wide `adaptive_parallelism_strategy` parameter surface.
- The session-config and job-level alter paths now accept and resolve `adaptive`, `0`, fixed integers, `bounded(n)`, and `ratio(r)` through the new `ConfigParallelism` path. Fragment-level alter parallelism remains fixed/adaptive only.
- Meta performs one-time migrations to:
  - derive the new unified session parameter values from legacy split parallelism + strategy settings,
  - backfill missing legacy per-job adaptive strategies,
  - add persisted backfill adaptive strategy metadata for streaming jobs,
  - delete deprecated legacy parameter rows instead of keeping runtime compatibility fallback logic.
- System catalog and diagnose-style output now show the effective user-facing policy more precisely, instead of flattening most adaptive jobs to coarse labels.
- Documentation and tests are updated to reflect the simplified configuration model and the new adaptive variants.

### User-facing changes

Before this PR:
- `ALTER ... SET PARALLELISM` effectively accepted only `default`, `adaptive`, `0`, or a fixed integer such as `4`.
- System catalogs and diagnose output mostly displayed coarse values such as `FIXED(4)`, `ADAPTIVE`, or `CUSTOM`.
- `pg_catalog.pg_settings` still exposed the cluster-wide `adaptive_parallelism_strategy` parameter.
- Default adaptive behavior was mostly opaque to users: catalogs usually showed `ADAPTIVE` instead of the concrete default policy.

After this PR:
- `ALTER TABLE|SOURCE|SINK|INDEX|MATERIALIZED VIEW ... SET PARALLELISM` can also use `bounded(n)` and `ratio(r)`.
- `rw_streaming_parallelism`, `rw_streaming_jobs`, `rw_table_fragments`, and diagnose-style output now show the actual policy:
  - fixed parallelism displays as `4` instead of `FIXED(4)`
  - adaptive default displays as `adaptive`
  - bounded adaptive displays as `bounded(16)`
  - ratio adaptive displays as `ratio(0.5)`
- Built-in defaults are now visible in catalog output. For example:
  - table/source default shows as `bounded(4)`
  - materialized view/sink/index default shows as `bounded(64)`
- `pg_catalog.pg_settings` no longer exposes `adaptive_parallelism_strategy`.
- `EXPLAIN DISTSQL` for streaming object creation now uses the actual job type when resolving default streaming parallelism, so plans better reflect table/source/sink/index/materialized-view-specific defaults.

- No tracking issue.

## Checklist

## Documentation

<details>

Streaming parallelism configuration is now unified under `streaming_parallelism` and per-type overrides. Users can now configure job-level streaming parallelism with `bounded(n)` and `ratio(r)`, and catalogs/diagnostics show the concrete policy such as `4`, `bounded(4)`, `bounded(64)`, or `ratio(0.5)` instead of mostly coarse `FIXED(...)` / `ADAPTIVE` labels. Legacy strategy-based parameters are migrated once on upgrade and then removed.

</details>


src/frontend/src/stream_fragmenter/parallelism.rs
```diff
@@ -12,84 +12,151 @@
...
+    global_streaming_parallelism: ConfigParallelism,
+) -> ConfigParallelism {
+    let resolved_global_parallelism = resolve_global_parallelism(global_streaming_parallelism);
+
+    match job_type {
+        Some(GraphJobType::Table | GraphJobType::Source) => {
+            if matches!(resolved_global_parallelism, ConfigParallelism::Fixed(_)) {
+                resolved_global_parallelism
+            } else {
+                ConfigParallelism::Bounded(NonZeroU64::new(4).unwrap())
```
hzxa21: If the `resolved_global_parallelism` is not `ConfigParallelism::Fixed`, it can be `ConfigParallelism::Ratio`, `ConfigParallelism::Adaptive` as well. Will it be a problem if we use `bounded(4)` in all the cases here?

hzxa21: Consider the following example:
```
set streaming_parallelism to 'RATIO(0.5)'
CREATE TABLE x ....
```

Should we use `RATIO(0.5)` or the default `BOUNDED(4)` for table x?

src/meta/model/migration/src/m20260311_000000_legacy_streaming_parallelism_session_params.rs
```diff
@@ -0,0 +1,1046 @@
...
+                    StreamingJob::AdaptiveParallelismStrategy,
+                    legacy_system_strategy.to_string(),
+                )
+                .and_where(Expr::col(StreamingJob::AdaptiveParallelismStrategy).is_null())
+                .to_owned(),
+        )
+        .await?;
+
+    Ok(())
+}
```
hzxa21, line 179: Will this make the source/table created in version <= 2.8 slightly change its parallelism to `legacy_system_strategy` (`bounded(64)` by default)?

hzxa21 (commented):
A potential bug existed prior to this PR: we use None job type here. Will it affect the explain result?
https://github.com/risingwavelabs/risingwave/blob/9bb17e5857a59baaeac8a26787ac2ec5802969f6/src/frontend/src/handler/explain.rs#L289

hzxa21:
To ensure all corner cases are considered,

I think we need slt tests to reveal the parallelism for the combination of different options in new version:
1. `set streaming_parallelism to default/fixed/ratio/bounded/adaptive` - 5 possibilities
2. `set streaming_parallelism_for_<type> to default/fixed/ratio/bounded/adaptive` - 5 possibilities

For each `CREATE <type>` statement, we need to test the 5x5 = 25 cases and check whether the expected parallelism for this type is correct.


Moreover, we should also add backward compatible slt tests to make sure the job parallelism doesn't change for the combination of different options:
1. `set streaming_parallelism to default/fixed` in old version - 2 possibilities
2. `set streaming_parallelism_for_<type> to default/fixed` in old version - 2 possibilities
3. `set streaming_parallelism_strategy to default/bouned/ratio/auto` in old version - 4 possibilities
4. `set streaming_parallelism_strategy_for_<type> to default/bouded/ratio/auto` in old version - 4 possibilities
5. `alter system set adaptive_parallelism_strategy to default/bounded/ratio/auto` in old version - 4 possibilities

For each `CREATE <type>` statement, we need to test 2x2x4x4x4 cases and make sure the parallelism doesn't change after upgrading the cluster to new version.

src/frontend/src/stream_fragmenter/parallelism.rs
shanicky: Historically, `BOUNDED(4)` was the default adaptive strategy for table/source, not the default parallelism value itself.

Before this refactor, the logic was split into two layers: parallelism first, then adaptive strategy. After removing the separate strategy layer, we are now overloading `default` to mean both "fall back to global" and "use the type-specific legacy default". That is the source of the ambiguity here.

shanicky: Before this change, the behavior was effectively two-stage: `streaming_parallelism_for_<type>` decided fixed vs adaptive, and `streaming_parallelism_strategy_for_<type>` decided the adaptive strategy. For table/source, the default strategy was `BOUNDED(4)`.

Codex proposes the following design:

1. If `streaming_parallelism_for_<type>` is explicitly set, use it directly.
2. Otherwise, if `streaming_parallelism` is explicitly set, inherit the global `streaming_parallelism`.
3. Only when both remain at `default` should we fall back to built-in effective defaults:
   - `table/source` => `BOUNDED(4)`
   - `mv/sink/index` => follow the global default, which ultimately resolves to `BOUNDED(64)`

So the precedence becomes:

explicit type-specific setting > explicit global setting > built-in defaults

Under this design, `BOUNDED(4)` is only the effective default for the untouched path. It should not behave like a hard override over an explicit global setting such as `RATIO(0.5)`.

src/meta/model/migration/src/m20260311_000000_legacy_streaming_parallelism_session_params.rs
shanicky: No. This only updates jobs whose `adaptive_parallelism_strategy` is still `NULL`.

Jobs that already have a job-level strategy are untouched, including source/table jobs that were already created with an explicit per-job adaptive strategy.

For the older jobs with `NULL` here, the previous behavior was to fall back to the system-level `adaptive_parallelism_strategy` at scheduling/scaling time. Since this migration removes that legacy system parameter, we materialize the current `legacy_system_strategy` into those jobs to preserve their existing effective behavior.

So this is not retroactively applying the new table/source default (`BOUNDED(4)`) to old jobs. It is only making the old system-level fallback explicit before we drop the legacy system parameter.

src/frontend/src/stream_fragmenter/parallelism.rs
hzxa21: > Codex proposes the following design:
> 
> 1. If `streaming_parallelism_for_<type>` is explicitly set, use it directly.
> 2. Otherwise, if `streaming_parallelism` is explicitly set, inherit the global `streaming_parallelism`.
> 3. Only when both remain at `default` should we fall back to built-in effective defaults:
>    
>    * `table/source` => `BOUNDED(4)`
>    * `mv/sink/index` => follow the global default, which ultimately resolves to `BOUNDED(64)`


This is my understanding too. It looks more intuitive to me. Examples:
```sql
-- parallelism(x) = bounded(4)
set streaming_parallelism to default
CREATE TABLE x ....

-- parallelism(x) = fixed(8)
set streaming_parallelism to 8
CREATE TABLE x ....

-- parallelism(x) = bounded(32)
set streaming_parallelism_for_table to 'bounded(32)'
CREATE TABLE x ....

-- parallelism(x) = bounded(32)
set streaming_parallelism to 8
set streaming_parallelism_for_table to 'bounded(32)'
CREATE TABLE x ....
```

shanicky:
> To ensure all corner cases are considered,
> 
> I think we need slt tests to reveal the parallelism for the combination of different options in new version:
> 
> 1. `set streaming_parallelism to default/fixed/ratio/bounded/adaptive` - 5 possibilities
> 2. `set streaming_parallelism_for_<type> to default/fixed/ratio/bounded/adaptive` - 5 possibilities
> 
> For each `CREATE <type>` statement, we need to test the 5x5 = 25 cases and check whether the expected parallelism for this type is correct.
> 
> Moreover, we should also add backward compatible slt tests to make sure the job parallelism doesn't change for the combination of different options:
> 
> 1. `set streaming_parallelism to default/fixed` in old version - 2 possibilities
> 2. `set streaming_parallelism_for_<type> to default/fixed` in old version - 2 possibilities
> 3. `set streaming_parallelism_strategy to default/bouned/ratio/auto` in old version - 4 possibilities
> 4. `set streaming_parallelism_strategy_for_<type> to default/bouded/ratio/auto` in old version - 4 possibilities
> 5. `alter system set adaptive_parallelism_strategy to default/bounded/ratio/auto` in old version - 4 possibilities
> 
> For each `CREATE <type>` statement, we need to test 2x2x4x4x4 cases and make sure the parallelism doesn't change after upgrading the cluster to new version.

The first part has been added via the standard ddl slt tests.
[#diff-1b9eaf5a68](https://github.com/risingwavelabs/risingwave/compare/shanicky%2Fstream-parallelism...shanicky%2Fstream-parallelism-v1?body=&expand=1#diff-1b9eaf5a689c20f86e1f06346b4c9f0ebf61bf522951dc17545d4ed056e47b50)

For the second part, it looks like the implementation is the same for every type, so in theory testing the 64 table cases should be sufficient. Or would you prefer to write out all 256 cases?
[#diff-cc271395f0](https://github.com/risingwavelabs/risingwave/compare/shanicky%2Fstream-parallelism...shanicky%2Fstream-parallelism-v1?body=&expand=1#diff-cc271395f03b1cd351699809d8b6f3a8e0255a9ace10cc37265011538234bf93)

shanicky:
This PR has been tested and is stable. To make review easier, I've split it into a stack of 6 smaller PRs with a linear dependency chain:

**PR Stack: Streaming Parallelism Unification**

| # | PR | Description |
|---|-----|-------------|
| 1 | #25125 | **Config types & parser** — Extend `ConfigParallelism` with `Bounded`/`Ratio` variants; add `SET key = bounded(n)` syntax support |
| 2 | #25126 | **Session config & frontend resolution** — Remove legacy `streaming_parallelism_strategy*` fields; rewrite `derive_parallelism()` to return `ResolvedParallelism` |
| 3 | #25127 | **ALTER PARALLELISM & RPC** — Propagate `adaptive_parallelism_strategy` through the alter parallelism handler and RPC layer |
| 4 | #25128 | **Meta model & controller** — DB migration for legacy strategy params; persist strategy in streaming job model; update scale/fragment controllers |
| 5 | #25129 | **Cleanup** — Remove deprecated `adaptive_parallelism_strategy` system parameter; update diagnostics and display |
| 6 | #25130 | **Tests** — E2E SLT tests for all job types, backwards-compatibility coverage, simulation test updates |

**Merge order:** strictly #25125 → #25126 → #25127 → #25128 → #25129 → #25130. Each PR is based on the previous one.

Please review starting from PR 1. Thanks!

src/frontend/src/catalog/system_catalog/mod.rs
```diff
@@ -222,15 +224,33 @@ fn extract_parallelism_from_table_state(state: &StreamingJobState) -> String {
...
+        Some(PbParallelism::Auto(_)) | Some(PbParallelism::Adaptive(_)) => state
+            .adaptive_parallelism_strategy
+            .as_deref()
+            .and_then(format_adaptive_parallelism_strategy)
+            .unwrap_or_else(|| "adaptive".to_owned()),
+        Some(PbParallelism::Custom(_)) => state
+            .adaptive_parallelism_strategy
+            .as_deref()
+            .and_then(format_adaptive_parallelism_strategy)
+            .unwrap_or_else(|| "custom".to_owned()),
```
hzxa21, line 239: Just curious. In which case, will custom parallelism be set?

src/common/src/session_config/parallelism.rs
```diff
@@ -27,9 +27,9 @@ const KEYWORD_DEFAULT_STRATEGY: &str = "default";
 
 #[derive(Copy, Debug, Clone, PartialEq, Default)]
 pub enum ConfigParallelism {
-    #[default]
     Default,
     Fixed(NonZeroU64),
+    #[default]
```
hzxa21: This will break `test_parallelism_default_variant`, which I think it is okay. The test needs to be fixed.

src/frontend/src/catalog/system_catalog/mod.rs
```diff
@@ -222,15 +224,33 @@ fn extract_parallelism_from_table_state(state: &StreamingJobState) -> String {
         .as_ref()
         .and_then(|parallelism| parallelism.parallelism.as_ref())
     {
-        Some(PbParallelism::Auto(_)) | Some(PbParallelism::Adaptive(_)) => "adaptive".to_owned(),
-        Some(PbParallelism::Fixed(PbFixedParallelism { parallelism })) => {
-            format!("fixed({parallelism})")
-        }
-        Some(PbParallelism::Custom(_)) => "custom".to_owned(),
+        Some(PbParallelism::Fixed(PbFixedParallelism { parallelism })) => parallelism.to_string(),
```
hzxa21, line 229: This is a breaking change because it changes the output value of the rw system table. I think it is fine. Just a note.

src/meta/src/model/stream.rs
```diff
@@ -298,6 +298,9 @@ pub struct StreamContext {
 
     /// The adaptive parallelism strategy for this job if it overrides the system default.
     pub adaptive_parallelism_strategy: Option<AdaptiveParallelismStrategy>,
```
hzxa21: We still have "Auto" and "Full" defined in `AdaptiveParallelismStrategy`. Do you think we should only keep one (e.g. Full or Auto) given that we already deprecated to global `adaptive_parallelism_strategy` system param.

src/meta/service/src/ddl_service.rs
```diff
@@ -1114,18 +1116,41 @@ impl DdlService for DdlServiceImpl {
         let parallelism = *req.get_parallelism()?;
         let deferred = req.get_deferred();
 
-        let parallelism = match parallelism.get_parallelism()? {
+        let adaptive_parallelism_strategy = req.adaptive_parallelism_strategy;
```
hzxa21, line 1119: I missed this in the sub-PR. Shall we make `adaptive_parallelism_strategy` in the PB a enum instead of a string? I saw that we are putting hardcoded string in https://github.com/risingwavelabs/risingwave/blob/f9dbb9734731973a0698f82670a3638ee271e6e4/src/frontend/src/handler/alter_parallelism.rs#L167

src/frontend/src/stream_fragmenter/parallelism.rs
```diff
@@ -31,7 +31,10 @@ pub(crate) struct ResolvedParallelism {
 fn resolve_global_parallelism(
     global_streaming_parallelism: ConfigParallelism,
 ) -> ConfigParallelism {
-    global_streaming_parallelism
+    match global_streaming_parallelism {
+        ConfigParallelism::Default => ConfigParallelism::Bounded(NonZeroU64::new(64).unwrap()),
```
hzxa21: I suggest we store the default value for table/source parallelism `ConfigParallelism::Bounded(NonZeroU64::new(4).unwrap())` and for others `ConfigParallelism::Bounded(NonZeroU64::new(64).unwrap())` as const at the same place as the `ConfigParallelism` struct in [session_config/parallelism.rs](https://github.com/risingwavelabs/risingwave/blob/f9dbb9734731973a0698f82670a3638ee271e6e4/src/common/src/session_config/parallelism.rs#L29) instead of filling in hardcoded value here. This can help us track the default value more explicitly.

hzxa21:
```rust
// Current
pub enum ConfigParallelism {
    #[default]
    Default,
    Fixed(NonZeroU64),
    Adaptive,
    Bounded(NonZeroU64),
    Ratio(f32),
}
```
Actually we can have something like this to make the logic clearer:
```
// After
pub enum Parallelism {
    Fixed(NonZeroU64),
    Adaptive,
    Bounded(NonZeroU64),
    Ratio(f32),
}

pub enum ConfigParallelism {
    #[default]
    Default: Parallelism,
    Custom: Parallelism
}

pub struct SesstionConfig {
    ...

    #[parameter(default = ConfigParallelism::Default(Parallelism::Bounded(64)))]
    streaming_parallelism: ConfigParallelism,

    #[parameter(default = ConfigParallelism::Default(Parallelism::Bounded(4)))]
    streaming_parallelism_for_source: ConfigParallelism,

    #[parameter(default = ConfigParallelism::Default(Parallelism::Bounded(4)))]
    streaming_parallelism_for_table: ConfigParallelism,


}
```

hzxa21 (commented):
Rest LGTM

hzxa21:
@zwang28 @wenym1 PTAL

ci/scripts/e2e-test-parallel.sh
```diff
@@ -51,7 +51,7 @@ kill_cluster() {
 
 host_args=(-h localhost -p 4565 -h localhost -p 4566 -h localhost -p 4567)
 
-export RUST_LOG="info,risingwave_stream=info,risingwave_batch=info,risingwave_storage=info,risingwave_storage::hummock::compactor=error,risingwave_hummock_sdk::compaction_group::hummock_version_ext=error,risingwave_storage::hummock::event_handler::hummock_event_handler=error"
+export RUST_LOG="info,risingwave_stream=info,risingwave_batch=info,risingwave_storage=info,risingwave_storage::hummock::compactor=error,risingwave_hummock_sdk::compaction_group::hummock_version_ext=error,risingwave_storage::hummock::event_handler::hummock_event_handler=error,await_tree=error,risingwave_connector::sink::log_store=warn,risingwave_stream::executor::sink=warn,risingwave_meta::barrier::info=error"
```
hzxa21, line 54: why do we want to change the log level?

ci/scripts/e2e-test-serial.sh
```diff
@@ -87,7 +87,7 @@ chmod +x ./target/debug/risingwave_e2e_extended_mode_test
 echo "--- Install Python Dependencies"
 python3 -m pip install --break-system-packages -r ./e2e_test/requirements.txt
 
-export RUST_LOG="info,risingwave_stream=info,risingwave_batch=info,risingwave_storage=info,risingwave_stream::common::table::state_table=warn,risingwave_storage::hummock::compactor=error,risingwave_hummock_sdk::compaction_group::hummock_version_ext=error"
+export RUST_LOG="info,risingwave_stream=info,risingwave_batch=info,risingwave_storage=info,risingwave_stream::common::table::state_table=warn,risingwave_storage::hummock::compactor=error,risingwave_hummock_sdk::compaction_group::hummock_version_ext=error,await_tree=error,risingwave_connector::sink::log_store=warn,risingwave_stream::executor::sink=warn,risingwave_meta::barrier::info=error"
```
hzxa21, line 90: ditto

src/meta/src/controller/fragment.rs
```diff
@@ -1560,15 +1608,10 @@ impl CatalogController {
...
-        let adaptive_parallelism_strategy = {
-            let system_params_reader = self.env.system_params_reader().await;
-            system_params_reader.adaptive_parallelism_strategy()
-        };
-
         let RenderedGraph { fragments, .. } = render_actor_assignments(
             self.env.actor_id_generator(),
             worker_nodes.current(),
-            adaptive_parallelism_strategy,
+            AdaptiveParallelismStrategy::default(),
```
zwang28: Is this adaptive parallelism strategy now set to default still being used?

src/meta/src/rpc/ddl_controller.rs
```diff
@@ -1902,30 +1902,42 @@ impl DdlController {
...
+                    .unwrap_or_default()
+                    .compute_target_parallelism(resolved_parallelism.get())
             };
-            adaptive_strategy.compute_target_parallelism(parallelism.get())
+
+            Ok(NonZeroUsize::new(resolved_parallelism).expect("parallelism must be positive"))
         };
 
-        let parallelism = NonZeroUsize::new(parallelism).expect("parallelism must be positive");
+        let has_backfill_override = specified_backfill_parallelism.is_some()
```
zwang28: What will happen if session var backfill_parallelism is gloablly set and applied to stream job that doesn't have the backfill phase? i.e. is it possible the job will remain in backfill_parallelism rathe than expected stream_parallelism?

src/meta/src/stream/scale.rs
```diff
@@ -1423,14 +1440,7 @@ impl GlobalStreamManager {
...
                     match notification {
-                        LocalNotification::SystemParamsChange(reader) => {
-                            let new_strategy = reader.adaptive_parallelism_strategy();
-                            if new_strategy != previous_adaptive_parallelism_strategy {
-                                tracing::info!("adaptive parallelism strategy changed from {:?} to {:?}", previous_adaptive_parallelism_strategy, new_strategy);
-                                should_trigger = true;
-                                previous_adaptive_parallelism_strategy = new_strategy;
-                            }
-                        }
+                        LocalNotification::SystemParamsChange(_) => {}
```
zwang28, line 1446: Missing logic to trigger parallelism adjustments upon session variable updates. (Legacy issue).

zwang28 (commented):
Comments left.

src/meta/src/rpc/ddl_controller.rs
hzxa21: Confirmed this path is safe. The frontend only fills `fragment_graph.backfill_parallelism` and `backfill_adaptive_parallelism_strategy` when `state.has_any_backfill` is true in `stream_fragmenter/mod.rs`. For jobs without any backfill phase, both fields are left empty, so `stream_ctx.backfill_adaptive_parallelism_strategy` stays `None` and `has_backfill_override` remains false here. In that case we still resolve the initial scheduling parallelism from the normal stream parallelism path rather than the backfill one.

src/meta/src/stream/scale.rs
hzxa21: Tracked this as #25478 and left a TODO in `src/meta/src/stream/scale.rs` next to the no-op notification branch. The gap is that relevant session-parameter updates do not currently emit a meta-local notification that reaches `GlobalStreamManager`, so there is no trigger for a new parallelism-control round yet.

src/meta/src/controller/fragment.rs
hzxa21: After walking through the codebase together with codex, I think it is no longer needed because the adaptive strategy will be part of the job metadata and its value will be backfilled for legacy jobs. I remove the adaptive_parallelism_strategy for the relevant function call and added a warn log in [render_new](https://github.com/risingwavelabs/risingwave/blob/00ab39a0a33c7a05a8470ca24c5abdb8a3a5065f/src/meta/src/controller/scale.rs#L1157) if we do need the default fallback.

hzxa21:
Note for reviewers: 

Don't panic on the LoCs. most LoCs are from SLT test because we need to test all the possible combinations mentioned here: https://github.com/risingwavelabs/risingwave/pull/25050#issuecomment-4067553915.

ci/scripts/e2e-test-parallel.sh
shanicky: We need to disable some logs here; otherwise, it will generate a large amount of logs and cause issues with the CI.

zwang28:
The PR removes the system parameter `adaptive_parallelism_strategy`, which may have been widely used.

As a system parameter, `adaptive_parallelism_strategy` can be set via toml during cluster initialization. Other runtime parameters cannot be set this way.

src/meta/src/stream/scale.rs
zwang28: > so there is no trigger for a new parallelism-control round yet.

I prefer current no-trigger to avoid unnecessary parallelism-control. i.e. only triggered by recover.

zwang28 (approved):
LGTM

wenym1:
Pushed two follow-up commits on `shanicky/stream-parallelism`:
- `e805488aa4`
- `24eecb9984`

Main follow-up changes in these commits:
- removed the dead `PbTableFragments.parallelism` RPC field and reserved the proto tag/name instead of keeping a deprecated unused field
- removed `StreamJobFragments.assigned_parallelism` and the related carry-through logic, since runtime/recovery already reads job-level parallelism from `streaming_job` metadata
- simplified `ActorGraphBuilder` / `Scheduler` by removing the resolved job parallelism argument that was only used for validation
- removed the now-dead `resolve_stream_parallelism_inner` helper and its unit tests after the surrounding flow no longer needed it
- tightened the frontend alter-parallelism path to use `AdaptiveParallelismStrategy` as a strong type through the handler, `CatalogWriter`, and `MetaClient`, with string conversion only at the protobuf boundary
- corrected the legacy migration fallback to match the old default system behavior (`Bounded(64)`), and clarified the migration comments for both legacy session-param and backfill-strategy migrations
- restored the `src/meta/src/model/stream.rs` tests and kept only the assertions that still match the current `StreamContext` shape

Local verification after these follow-ups:

chenzl25 (approved):

wenym1:
* **#25050** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25050?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25050?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #25057, opened 2026-03-12
fix(cdc): fix compatibility issues with SQL Server CDC table name

wcy-fdu:

## Summary
This PR fixes a SQL Server CDC backward-compatibility issue for tables created before table-name normalization.

For legacy CDC tables, the persisted `CdcFilter` literal may be `db.schema.table` (for example, `prod.dbo.Answer`), while newer runtime metadata emits `_rw_table_name` as `schema.table` (for example, `dbo.Answer`).  
As a result, the filter condition never matches and all rows are dropped (`in_record_cnt > 0`, `out_record_cnt = 0`).

## Root Cause
`CdcFilter` uses an equality predicate on `_rw_table_name`:
- Left side: runtime table name from CDC metadata
- Right side: persisted table literal from historical table definition

After SQL Server table-name normalization changes, legacy persisted literals and runtime values can differ only by the database prefix:
- Legacy literal: `db.schema.table`
- Runtime value: `schema.table`

This causes a silent mismatch and zero output for affected legacy tables.

## Fix
Add compatibility rewrite for `CdcFilter` search condition during executor build:
- Detect legacy CDC equality pattern (`_rw_table_name = <literal>`)
- If literal is in `db.schema.table` form, derive normalized `schema.table`
- Rewrite predicate as:
  - `original_condition OR (_rw_table_name = normalized_literal)`

So legacy and normalized forms both match, without requiring table recreation.

## Scope
- Targeted to CDC filter condition handling
- No catalog migration required
- No behavior change for already-normalized table names

## Validation
- Added unit tests for:
  - legacy name normalization (`db.schema.table -> schema.table`)
  - predicate rewrite for legacy literals
  - no rewrite for already-normalized literals
- Local test command:
  - `cargo test -p risingwave_stream cdc_filter -- --nocapture`

## User Impact
- Existing legacy SQL Server CDC tables can continue working after upgrade
- No need to drop/recreate CDC tables solely for this table-name format mismatch

- Closes #[ISSUE_NUMBER]

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/from_proto/cdc_filter.rs
+++ b/src/stream/src/from_proto/cdc_filter.rs
@@ -52,6 +52,7 @@ impl ExecutorBuilder for CdcFilterExecutorBuilder {
 fn build_search_condition_with_compat(node: &CdcFilterNode) -> StreamResult<ExprNode> {
     Ok(with_legacy_sqlserver_table_name_compat(
         node.get_search_condition()?,
+        node.upstream_source_id.as_raw_id(),
     ))
 }
 
@@ -63,7 +64,10 @@ fn build_search_condition_with_compat(node: &CdcFilterNode) -> StreamResult<Expr
 /// NOTE: This only rewrites a top-level equality predicate.
 /// `CdcFilter` search condition is expected to be a single equality on `_rw_table_name`.
 /// If that assumption changes in planner/proto, this function should be updated accordingly.
-fn with_legacy_sqlserver_table_name_compat(search_condition: &ExprNode) -> ExprNode {
+fn with_legacy_sqlserver_table_name_compat(
+    search_condition: &ExprNode,
+    upstream_source_id: u32,
+) -> ExprNode {
     let Some((table_name_ref_expr, table_name_literal_expr, table_name_literal)) =
         extract_cdc_filter_eq_condition(search_condition)
     else {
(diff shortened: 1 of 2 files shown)


src/stream/src/from_proto/cdc_filter.rs
```diff
@@ -30,10 +38,212 @@ impl ExecutorBuilder for CdcFilterExecutorBuilder {
...
+}
+
+/// For legacy SQL Server CDC tables created before table-name normalization, preserve
+/// compatibility between:
+/// - historical filter literal: `db.schema.table`
+/// - runtime `_rw_table_name`: `schema.table`
+///
+/// NOTE: This only rewrites a top-level equality predicate.
+/// `CdcFilter` search condition is expected to be a single equality on `_rw_table_name`.
+/// If that assumption changes in planner/proto, this function should be updated accordingly.
```
tabVersion, line 66: shall we also note it when create cdc table, people often got wrong about compatible logic.

src/stream/src/from_proto/cdc_filter.rs
```diff
@@ -13,13 +13,21 @@
...
 use super::*;
 use crate::executor::FilterExecutor;
 
 pub struct CdcFilterExecutorBuilder;
 
+/// Column index of `_rw_table_name` in CDC source chunks.
+///
+/// Current CDC source schema is `(payload, _rw_offset, _rw_table_name)`.
+/// Keep this value in sync with CDC source schema definition.
+const RW_TABLE_NAME_COLUMN_IDX: u32 = 2;
```
tabVersion, line 31: although it is a convention, I'd suggest getting the col by name instead of fixed id.

src/stream/src/from_proto/cdc_filter.rs
```diff
@@ -30,10 +38,212 @@ impl ExecutorBuilder for CdcFilterExecutorBuilder {
...
+
+    let Some(normalized_table_name) = normalize_legacy_sqlserver_table_name(table_name_literal)
+    else {
+        return search_condition.clone();
+    };
+
+    // The filter is already in normalized format.
+    if normalized_table_name == table_name_literal {
+        return search_condition.clone();
+    }
```
tabVersion, line 85: chore: let's print a log in the else branch, ie. we are applying the compatible logic

```
if normalized_table_name == table_name_literal { ... } else {
      tracing::info!(
          original = table_name_literal,
          normalized = %normalized_table_name,
          "rewriting legacy SQL Server CdcFilter expression for compatibility"
      );
  }
```

PR #25074, opened 2026-03-14
feat(source): adaptive splits + streaming pull for PubSub source

tabVersion:
## Summary

- **Enable adaptive split mode for Google PubSub source**, unblocking cluster scaling (scale up/down).
  - The `pubsub.parallelism` parameter is deprecated (kept for backward compatibility with existing DDL).
- **Switch PubSub reader from synchronous pull to streaming pull** (`subscribe()` API), providing a persistent bidirectional gRPC connection with automatic heartbeat, reconnection, and flow control.
  - **Fix message loss during parallelism reschedule** by properly disposing streaming pull subscribers instead of relying on the racy default `Drop` impl.
- **Expose `pubsub.ack_deadline_seconds`** as an alter-on-fly SQL property (range 10-600s, default 60s).

## Changes

| File | Change |
|------|--------|
| `source/base.rs` | Add `GooglePubsub` to `enable_drop_split()` and `enable_adaptive_splits()` |
| `source/google_pubsub/enumerator/client.rs` | Remove `split_count`; `list_splits()` returns 1 template split |
| `source/util.rs` | Add `GooglePubsub` branch in `fill_adaptive_split()` + unit tests |
| `source/google_pubsub/mod.rs` | Deprecate `pubsub.parallelism`; add `pubsub.ack_deadline_seconds` with `allow_alter_on_fly` |
| `source/google_pubsub/source/reader.rs` | Replace sync `pull()` loop with streaming `subscribe()` + batch drain via `now_or_never()` + auto-reconnect + `DisposableMessageStream` wrapper for safe cleanup |
| `allow_alter_on_fly_fields.rs` | Register `pubsub.ack_deadline_seconds` for PubsubProperties |
| `with_options_source.yaml` | Auto-generated: new field entry |

## Streaming Pull Details

- Uses `subscription.subscribe(SubscribeConfig)` which internally opens a `StreamingPull` bidirectional gRPC stream.
- Messages are batched (up to 1024) using `now_or_never()` to drain ready messages without blocking, matching the old pull batch size.
- On stream termination the reader automatically reconnects (outer `loop`), with a warn log.
- `stream_ack_deadline_seconds` is set on the `StreamingPullRequest`; the server extends deadlines via bidirectional ping as long as the stream is alive.
- Ack mechanism (`WaitCheckpointTask::AckPubsubMessage`) is unchanged -- it uses a separate `Subscription` client with batch `ack(ack_ids)` after checkpoint.

## DisposableMessageStream: Fixing Message Loss on Reschedule

When parallelism scales down, old split readers are dropped. The `gcloud-pubsub` crate default `MessageStream::drop()` has a race condition with `Subscriber::drop()`:

1. `MessageStream::drop()` spawns a background task to drain buffered messages and nack them
2. `Subscriber::drop()` immediately calls `task.abort()` on the receive task
3. The abort wins the race -- in-flight messages are lost permanently without being nacked back to the subscription

**Impact observed in testing**: 10 out of 50 messages permanently lost after `ALTER MV SET PARALLELISM = 1`, with the remaining 40 delayed ~5 minutes waiting for ack deadline expiry.

**Fix**: `DisposableMessageStream` wraps `MessageStream` and calls `dispose()` on drop (via `tokio::spawn`). `dispose()` properly sequences cleanup: takes subscribers out first, drains channel, nacks all unprocessed messages, then disposes each subscriber. This avoids the race because subscribers are consumed by `dispose()` before their `Drop` can fire.

## Backward Compatibility

- **Split ID format unchanged**: `{subscription}-{index}`.
- **Existing DDL accepted**: `pubsub.parallelism` is still parsed (just ignored with a warning log).
- **At-least-once semantics preserved**: Ack mechanism operates on per-message ack_ids at the subscription level.
- **Upgrade transition**: On first meta tick after upgrade, `fill_adaptive_split(template, actor_count)` generates splits with the same IDs as before, seamless.

## Test plan

- [x] Unit tests for `fill_adaptive_split` with PubSub template (3 actors, 1 actor, unsupported type)
- [x] `cargo check -p risingwave_connector` passes
- [x] `./risedev generate-with-options` passes (yaml + allow_alter_on_fly in sync)
- [x] E2E: PubSub emulator + RisingWave -- publish messages, query MV, data flows correctly
- [x] E2E: Cluster recovery (kill + restart) -- at-least-once holds, 0 duplicates, downtime messages recovered
- [x] E2E: Scale down (SET PARALLELISM = 1) -- confirmed message loss without DisposableMessageStream, fixed with it
- [ ] Existing PubSub e2e tests (e2e_test/source_inline/pubsub/) should pass unchanged
- [ ] Manual: ALTER SOURCE SET pubsub.ack_deadline_seconds -- verify takes effect on next reader creation

### DIFF
--- a/src/connector/src/source/google_pubsub/source/reader.rs
+++ b/src/connector/src/source/google_pubsub/source/reader.rs
@@ -44,21 +44,17 @@ struct DisposableMessageStream(Option<MessageStream>);
 
 impl Drop for DisposableMessageStream {
     fn drop(&mut self) {
+        #[cfg(not(madsim))]
         if let Some(stream) = self.0.take() {
-            // Use block_in_place + block_on to guarantee dispose() completes
-            // before Drop returns. A tokio::spawn here would race with task
-            // cancellation during RECOVER: if the spawned task never runs, the
-            // gRPC stream stays open and the library's internal ack-deadline
-            // extension keeps the message locked, preventing redelivery.
-            let count = tokio::task::block_in_place(|| {
-                tokio::runtime::Handle::current().block_on(stream.dispose())
+            tokio::spawn(async move {
+                let count = stream.dispose().await;
+                if count > 0 {
+                    tracing::info!(
+                        "disposed pubsub streaming pull, nacked {} in-flight messages",
+                        count
+                    );
+                }
             });
-            if count > 0 {
-                tracing::info!(
-                    "disposed pubsub streaming pull, nacked {} in-flight messages",
-                    count
-                );
-            }
         }
     }
 }
--- a/e2e_test/source_inline/pubsub/pubsub.slt.serial
+++ b/e2e_test/source_inline/pubsub/pubsub.slt.serial
@@ -103,7 +103,7 @@ set visibility_mode = 'all';
 
 # Wait for redelivery. Poll instead of a fixed sleep, because the emulator's
 # streaming-pull redelivery timing is non-deterministic.
-query I retry 60 backoff 1s
+query I retry 120 backoff 1s
 select count(distinct v1) from s1;
 ----
 10


tabVersion:
## Why the e2e test (`pubsub.slt.serial`) was updated

The original test expected `count(*) = 2` for each `(v1, v2)` group before checkpoint, because the old `pull()`-based implementation did **not** ack messages until checkpoint. With the default 10s ack deadline, Pub/Sub would redeliver unacked messages, resulting in duplicates (count=2).

After switching to **streaming pull** (`subscribe()`), the streaming pull protocol **automatically extends the ack deadline** for in-flight messages as long as the stream is active. This means messages are no longer redelivered while the reader is connected — so `count(*)` is now correctly `1`.

The post-`RECOVER` part of the test still validates redelivery: when recovery disconnects the stream, the ack deadline expires and Pub/Sub redelivers the uncheckpointed messages as expected.

tabVersion:
## PR Relationship Map

| Role | PR | Branch | Base |
|------|-----|--------|------|
| **pr** (main feature) | #25074 | `feat/pubsub-adaptive-splits` | `main` |
| **pr-cherry-pick-2.7** | #25089 | `feat/pubsub-adaptive-splits-release-2.7` | `release-2.7` |
| **pr-fix** (checkpoint fix + metrics) | #25144 | `fix/pubsub-checkpoint-ack-metrics` | `feat/pubsub-adaptive-splits` |
| **pr-fix-cherry-pick-2.7** | #25145 | `fix/pubsub-checkpoint-ack-metrics-release-2.7` | `feat/pubsub-adaptive-splits-release-2.7` |

**Merge order:** #25144 → #25074 → `main`, #25145 → #25089 → `release-2.7`

src/connector/src/source/google_pubsub/source/reader.rs
```diff
@@ -95,10 +145,19 @@ impl SplitReader for PubsubSplitReader {
         );
         let split = splits.into_iter().next().unwrap();
 
+        let ack_deadline_seconds = properties
+            .ack_deadline_seconds
+            .unwrap_or(DEFAULT_ACK_DEADLINE_SECONDS);
+
+        if !(10..=600).contains(&ack_deadline_seconds) {
+            bail!("pubsub.ack_deadline_seconds must be between 10 and 600");
```
xiangjinwu, line 154: Is this check too late or not?

xiangjinwu (approved):
Rest lgtm

src/connector/src/source/google_pubsub/source/reader.rs
tabVersion: resolve by 1972c450380f9078ba854a7399e1d5df8c3bafa8 , adding the check also when creating enumerator.

PR #25076, opened 2026-03-15
feat(source): support consuming multiple Kafka topics via regex pattern

tabVersion:
## Summary

- Add `topic.regex` property for Kafka sources, enabling subscription to multiple topics matching a regex pattern
- New topics matching the pattern are discovered dynamically (~30s interval) without source recreation
- Maintain exactly-once semantics with unique `{topic}:{partition}` split IDs for independent offset tracking per topic-partition

### New feature: `topic.regex` for Kafka sources

Use a regex pattern to consume from multiple Kafka topics at once:

```sql
CREATE TABLE events (id int, data varchar)
INCLUDE partition AS kafka_partition
INCLUDE topic AS kafka_topic
WITH (
    connector = 'kafka',
    "topic.regex" = '^events\..*',
    properties.bootstrap.server = 'localhost:9092',
    scan.startup.mode = 'earliest'
) FORMAT PLAIN ENCODE JSON;
```

- **`topic.regex`**: A regex pattern matching Kafka topic names. Mutually exclusive with `topic`.
- **Dynamic discovery**: New topics matching the regex are automatically picked up (~30s interval).
- **`INCLUDE topic`**: New additional column for Kafka sources. Returns the topic name the message was consumed from. Useful with `topic.regex` to distinguish messages from different topics.
- **`INCLUDE partition`**: Behavior unchanged - always shows the pure Kafka partition number, even in multi-topic mode.
- **Validation**: `topic` and `topic.regex` cannot be specified at the same time. At least one must be set. Errors are reported at `CREATE SOURCE` / `CREATE TABLE` time (not deferred to runtime).

### Breaking changes

None. Existing single-topic sources (`topic = '...'`) are unaffected.

## Motivation

Customers want to consume events from many Kafka topics sharing a naming convention (e.g., `events.orders`, `events.users`, `events.payments`) through a single source, with automatic discovery when new topics are added.

## Changes

| File | Change |
|------|--------|
| `connector_common/common.rs` | Add `topic.regex: Option<String>` to `KafkaCommon`; make `topic` serde-default |
| `kafka/split.rs` | Add `#[serde(default)] multi_topic: bool`; split ID becomes `{topic}:{partition}` when true |
| `kafka/enumerator.rs` | Add regex-based topic discovery; fix metric key from weak i32 hash to String |
| `kafka/source/message.rs` | Add `topic` and `partition` fields to `KafkaMeta` |
| `kafka/source/reader.rs` | Detect `multi_topic` from splits; propagate to message creation and metrics |
| `kafka/mod.rs` | Add `validate_topic_config()`, `is_multi_topic()`; 6 unit tests |
| `sink/kafka.rs` | Validate `topic` non-empty for sinks |
| `proto/plan_common.proto` | Add `AdditionalColumnTopicName` message (tag 15) |
| `parser/additional_columns.rs` | Add `"topic"` to Kafka compatible columns |
| `parser/chunk_builder.rs` | Handle `TopicName` column; decouple `Partition` from raw `split_id` |
| `parser/utils.rs` | Add `extract_partition_from_meta()` and `extract_topic_from_meta()` |
| `frontend/validate.rs` | Add DDL-time validation for `topic` vs `topic.regex` |
| `topic_regex.slt` | E2E test covering all scenarios |

## Why there are no backward compatibility issues

1. **Serialized split format**: `multi_topic` uses `#[serde(default)]` - existing splits deserialize with `multi_topic: false`.
2. **Single-topic code path unchanged**: When `topic` is set, execution follows the exact same path as before.
3. **INCLUDE partition unchanged**: Always shows the pure partition number, extracted from `KafkaMeta.partition`.
4. **No split ID collision**: `{topic}:{partition}` vs `{partition}` formats never overlap.
5. **Sink unaffected**: Validates `topic` non-empty in `from_btreemap()`.

## Test plan

- [x] Unit tests: split ID format, backward compat deserialization, config validation
- [x] E2E SLT test:
  - Validation errors (both set, neither set, invalid regex)
  - Single-topic backward compatibility
  - Multi-topic regex consumption with topic filtering
  - Dynamic topic addition
  - `INCLUDE partition` shows pure partition number
  - `INCLUDE topic` shows topic name
  - Exactly-once verification

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/frontend/src/handler/create_source.rs
+++ b/src/frontend/src/handler/create_source.rs
@@ -1024,9 +1024,9 @@ HINT: use `CREATE TABLE <name> WITH (...)` instead of `CREATE TABLE <name> (<col
             }
             (_, Some(pattern)) => {
                 if let Err(e) = regex::Regex::new(pattern) {
+                    let err_msg = e.to_string();
                     return Err(RwError::from(ProtocolError(format!(
-                        "invalid topic.regex pattern '{}': {}",
-                        pattern, e
+                        "invalid topic.regex pattern '{pattern}': {err_msg}",
                     ))));
                 }
 
--- a/src/frontend/src/handler/create_source/validate.rs
+++ b/src/frontend/src/handler/create_source/validate.rs
@@ -269,9 +269,9 @@ pub fn validate_compatibility(
         if has_regex {
             let pattern = &props["topic.regex"];
             if let Err(e) = regex::Regex::new(pattern) {
+                let err_msg = e.to_string();
                 return Err(RwError::from(ProtocolError(format!(
-                    "invalid topic.regex pattern '{}': {}",
-                    pattern, e
+                    "invalid topic.regex pattern '{pattern}': {err_msg}",
                 ))));
             }
         }


chenzl25:
@xiangjinwu Please help to review this PR.

src/connector/with_options_sink.yaml
```diff
@@ -750,9 +750,18 @@ KafkaConfig:
   fields:
   - name: topic
     field_type: String
-    required: true
+    required: false
+    default: Default::default
     alias:
     - kafka.topic
+  - name: topic.regex
```
hzxa21: This is not applicable to sink. I think we should avoid adding it to the sink option.

src/connector/src/parser/chunk_builder.rs
```diff
@@ -383,13 +383,18 @@ impl SourceStreamChunkRowWriter<'_> {
...
+                        Some(row_meta) => {
+                            extract_partition_from_meta(row_meta.source_meta, row_meta.split_id)
+                        }
+                        None => DatumCow::Borrowed(None),
+                    }))
                 }
+                (_, &Some(AdditionalColumnType::TopicName(_))) => Ok(A::output_for(
+                    self.row_meta
+                        .as_ref()
+                        .and_then(|ele| extract_topic_from_meta(ele.source_meta)),
```
hzxa21, line 396: This change makes the additional partition column changing from the split id to the bare Kafka partition number in multi-topic mode.

However, in source executor, we rely on the value of additional partition column to retrieve the split id: https://github.com/risingwavelabs/risingwave/blob/a94aad01ac086409c67c0bac3bbd834a9613f734/src/stream/src/executor/source/mod.rs#L90

This will mess up with the source state IMO

e2e_test/source_inline/kafka/topic_regex.slt
```diff
@@ -0,0 +1,200 @@
...
+INCLUDE topic AS kafka_topic
+WITH (
+    ${RISEDEV_KAFKA_WITH_OPTIONS_COMMON},
+    "topic.regex" = '^tr_events_.*',
+    scan.startup.mode = 'earliest'
+) FORMAT PLAIN ENCODE JSON;
+
+# Should consume from tr_events_orders and tr_events_users, but NOT tr_logs_system
+query IT retry 10 backoff 2s
+SELECT id, data FROM regex_multi ORDER BY id;
```
hzxa21, line 117: We should check source state table and see whether the offset and split id matches.

tabVersion:
replaced by https://github.com/risingwavelabs/risingwave/pull/25323

PR #25078, opened 2026-03-15
fix(meta): prevent unnecessary actor rebuilds from spurious parallelism strategy changes

### LINKED ISSUE #25077: bug(simulation): deterministic e2e DDL test fails during adhoc recovery with random vnode count
## Summary

The `end-to-end test (madsim, random vnode count)` step in the `main-cron` pipeline has been **consistently failing** on the `main` branch (Reliability: 0%). The failure occurs in `e2e_test/ddl/alter_swap_rename.slt` at line 15, where a DML statement hits an adhoc recovery window.

## Failing builds

All recent `main-cron` builds on `main` are affected:

| Build | Commit | Status |
|-------|--------|--------|
| [#9053](https://buildkite.com/risingwavelabs/main-cron/builds/9053) | `95a5b11ab2` | Failed |
| [#9052](https://buildkite.com/risingwavelabs/main-cron/builds/9052) | `95a5b11ab2` | Failed |
| [#9048](https://buildkite.com/risingwavelabs/main-cron/builds/9048) | `5a29ff4cf1` | Failed |
| [#9042](https://buildkite.com/risingwavelabs/main-cron/builds/9042) | `f6b71b466f` | Failed |
| [#9032](https://buildkite.com/risingwavelabs/main-cron/builds/9032) | `42aacf121c` | Failed |

The `pull-request` pipeline sim tests pass because they use fewer seeds and no random vnode count.

## Reproduction

```bash
MADSIM_TEST_SEED=32 RW_SIM_RANDOM_VNODE_COUNT=true ./target/debug/risingwave_simulation ./e2e_test/ddl/**/*.slt
# also reproduces with MADSIM_TEST_SEED=4
```

## Root cause analysis

The failure is **NOT a simple race condition**. It is a deterministic chain of events triggered by `RW_SIM_RANDOM_VNODE_COUNT=true`:

### The causal chain (from `ddl-32.log`)

1. **Random vnode count injects `SET STREAMING_MAX_PARALLELISM = <random>`** before each CREATE TABLE/MV. This creates tables with varying vnode counts (2–1056).

2. **The parallelism control monitor detects a system parameter change**:
   ```
   adaptive parallelism strategy changed from Auto to Bounded(64)
   ```
   When a DDL test file sets session-level `STREAMING_MAX_PARALLELISM`, this triggers a `SystemParamsChange` notification. The scale controller at `src/meta/src/stream/scale.rs:1084-1089` detects the change and sets `should_trigger = true`.

3. **`trigger_parallelism_control()` is called**, which examines all `CREATED` jobs and tries to reschedule them to match the new parallelism target. This calls into the barrier manager to execute the reschedule.

4. **The reschedule triggers an adhoc recovery**:
   ```
   trigger parallelism control
   trigger parallelism control for jobs: [...]
   adhoc_recovery{error=adhoc recovery triggered}: cluster marked as blocked
   adhoc_recovery{error=adhoc recovery triggered}: recovery start!
   ```

5. **Recovery completes** at `T+0.703s`:
   ```
   recovery success
   cluster marked as ready blocked_databases={}
   ```

6. **DML arrives before table readers are re-registered**: The SLT runner continues executing `e2e_test/ddl/alter_swap_rename.slt:15` — `INSERT INTO t2 VALUES(2,(2,(2,4)))`. The cluster is marked "ready" but the DML executor on the compute node hasn't received its first barrier yet, so `TableDmlHandle::write_handle()` finds `changes_txs` empty:
   ```
   DML is not permitted during cluster recovery (no available table reader in streaming executors)
   ```

7. **SLT runner panics** because `run_no_kill` (line 514) does not retry on error — it immediately panics.

### Why this only happens with random vnode count

Without `RW_SIM_RANDOM_VNODE_COUNT`, all tables use the default `STREAMING_MAX_PARALLELISM = 256` and the `adaptive_parallelism_strategy` stays at `Auto`. The parallelism control monitor never detects a change, so no reschedule → no adhoc recovery → no DML failure.

With random vnode count enabled, the injected `SET STREAMING_MAX_PARALLELISM = <N>` changes the system-level strategy from `Auto` to `Bounded(64)`, which triggers the parallelism control monitor to reschedule existing jobs. The reschedule causes an adhoc recovery, and the immediately following DML in the SLT test hits the recovery window.

### The gap: `run_no_kill` doesn't retry

In the kill-enabled code path (line 558), the SLT runner retries with exponential backoff. But `run_no_kill` (line 512-514) assumes no failures are possible and panics immediately:

```rust
if !env.kill() {
    match run_no_kill(&mut tester, random_vnode_count, &cmd, record).await {
        Ok(_) => continue,
        Err(e) => panic!("{}", e),  // <-- no retry
    }
}
```

This is incorrect when random vnode count is enabled, because the parallelism changes can trigger adhoc recovery even without any node kills.

## Possible fixes

1. **Quick fix**: Add retry logic in `run_no_kill` for transient errors like "DML is not permitted during cluster recovery" when `random_vnode_count` is enabled.
2. **Better fix**: Prevent `SET STREAMING_MAX_PARALLELISM` (session-level) from triggering system-level `adaptive_parallelism_strategy` changes that cause reschedules of existing jobs.
3. **Proper fix**: Ensure the cluster doesn't accept DML until table readers are fully re-registered after recovery — the "ready" state should mean truly ready.

## Affected test

- File: `e2e_test/ddl/alter_swap_rename.slt:15`
- Statement: `INSERT INTO t2 VALUES(2,(2,(2,4)));`
- Artifact logs: `risedev-logs/ddl-32.log`, `risedev-logs/ddl-4.log`

tabVersion:
## Summary

Two-layer fix for the Scale Controller triggering unnecessary adhoc recovery when the adaptive parallelism strategy appears to change but the actual topology remains the same.

Closes #25077.

### 1. Initialize `previous_adaptive_parallelism_strategy` from system params (`scale.rs`)

In `ScaleController::run()`, the cached strategy was initialized with `AdaptiveParallelismStrategy::default()` (`Auto`), but the actual system parameter default is `Bounded(64)`. On the first `SystemParamsChange` notification, the monitor falsely detected `Auto → Bounded(64)` and triggered `trigger_parallelism_control()`.

Now reads the real value via `self.env.system_params_reader().await.adaptive_parallelism_strategy()`, consistent with how the Barrier Worker initializes the same field.

### 2. Add semantic no-op check in `build_reschedule_from_context()` (`worker.rs`)

The reschedule path always allocates fresh actor IDs — `diff_fragment` is a full rebuild by design (`debug_assert!(kept_ids.is_empty())`). This means even when parallelism and worker distribution are unchanged, it produces a "remove all + add all" reschedule plan, which is disruptive (can trigger adhoc recovery).

Added `is_reschedule_noop()` that compares the rendered result against the current inflight state:
- **Actor count** per fragment
- **Per-worker actor distribution** (which worker has how many actors)

When both match for every fragment, the reschedule is skipped (`Ok(None)`), and the caller is notified as success without injecting a barrier.

This makes the system **idempotent** — any spurious trigger from the Scale Controller (initialization mismatch, `Auto ↔ Full` switch that computes to the same parallelism, redundant `alter system set`, etc.) becomes a harmless no-op at the Barrier Worker level.

## Why initialization fix alone is not enough

Fixing initialization only eliminates one specific false detection (`Auto → Bounded(64)`). But it introduces the reverse: tests and users that `alter system set adaptive_parallelism_strategy to AUTO` from the `Bounded(64)` default now trigger a real `Bounded(64) → Auto` change. Since both strategies compute to `total_parallelism` for the same cluster, the reschedule produces identical topology — yet without the no-op check, it still does a full actor rebuild.

## Test plan

- [x] `cargo check -p risingwave_meta` passes
- [ ] Madsim deterministic simulation tests (no false reschedule triggers)
- [ ] Integration tests (`adaptive_strategy`, `auto_parallelism`) — strategy changes that don't alter actual parallelism are now correctly skipped

### DIFF
--- a/src/meta/src/barrier/worker.rs
+++ b/src/meta/src/barrier/worker.rs
@@ -347,9 +347,7 @@ fn is_reschedule_noop(
                     return false;
                 }
 
-                if worker_distribution(&new_info.actors)
-                    != worker_distribution(&prev_info.actors)
-                {
+                if worker_distribution(&new_info.actors) != worker_distribution(&prev_info.actors) {
                     return false;
                 }
             }


shanicky (approved):

tabVersion:
@shanicky @hzxa21 @wenym1 Can you guys take a look at the no-op added when the cluster is not necessary to trigger a scale

src/meta/src/barrier/worker.rs
```diff
@@ -294,14 +296,66 @@ fn build_reschedule_from_context(
...
+/// per-worker actor distribution as the current inflight state. In this case
+/// the reschedule would be a pure rebuild with no semantic change, since
+/// `diff_fragment` always allocates fresh actor IDs (full rebuild by design).
+fn is_reschedule_noop(
+    rendered: &FragmentRenderMap,
+    prev_fragments: &HashMap<FragmentId, &InflightFragmentInfo>,
+) -> bool {
+    fn worker_distribution(
+        actors: &HashMap<ActorId, InflightActorInfo>,
+    ) -> BTreeMap<WorkerId, usize> {
```
shanicky, line 330: These checks are not sufficient to prove semantic equivalence. There are cases where they all pass, but the `vnode_bitmap` still differs, for example when worker weights change under certain conditions.

PR #25081 provides a more complete check.

PR #25081, opened 2026-03-16
fix(meta): skip no-op reschedules before actor allocation

shanicky:

This PR teaches reschedule planning to compare a normalized layout before consuming real actor ids, so no-op parallelism-control triggers no longer allocate actor ids or rebuild actors unnecessarily.
It first adds cross-fragment layout normalization to detect no-op reschedules, then renders preview assignments with temporary actor ids and only materializes real ids after the layout is confirmed to change.
This keeps the existing full-rebuild path for real reschedules while making no-op triggers idempotent and avoiding actor id growth from skipped plans.

- Related to #25077

## Checklist

### DIFF
--- a/src/meta/src/controller/scale.rs
+++ b/src/meta/src/controller/scale.rs
@@ -1330,6 +1330,14 @@ impl<'a> ComponentFragmentAligner<'a> {
         }
     }
 
+    pub(crate) fn new_persistent(
+        actor_template: &'a EnsembleActorTemplate,
+        actor_id_counter: &AtomicU32,
+    ) -> Self {
+        let mut actor_id_allocator = RenderActorIdAllocator::Persistent(actor_id_counter);
+        Self::new(actor_template, &mut actor_id_allocator)
+    }
+
     pub(crate) fn align_component_actor(
         &self,
         distribution_type: DistributionType,
--- a/src/meta/src/barrier/checkpoint/state.rs
+++ b/src/meta/src/barrier/checkpoint/state.rs
@@ -346,7 +346,8 @@ fn render_actors(
             }
             let fragment = &fragments.inner.fragments[&fragment_id];
             let distribution_type: DistributionType = fragment.distribution_type.into();
-            let aligner = ComponentFragmentAligner::new(&actor_template, actor_id_counter);
+            let aligner =
+                ComponentFragmentAligner::new_persistent(&actor_template, actor_id_counter);
             let assignments = aligner.align_component_actor(distribution_type);
             actor_assignments.insert(fragment_id, assignments);
         }
@@ -964,8 +965,10 @@ impl DatabaseCheckpointControl {
                         let actor_template = EnsembleActorTemplate::from_existing_inflight_fragment(
                             original_frag_info,
                         );
-                        let new_aligner =
-                            ComponentFragmentAligner::new(&actor_template, actor_id_counter);
+                        let new_aligner = ComponentFragmentAligner::new_persistent(
+                            &actor_template,
+                            actor_id_counter,
+                        );
                         let distribution_type: DistributionType =
                             sink_ctx.new_fragment.distribution_type.into();
                         let actor_assignments =


src/meta/src/controller/scale.rs
```diff
@@ -592,6 +592,100 @@ pub(crate) fn render_actor_assignments(
...
+    // Empty fragments should not appear, but fragment_id keeps the ordering
+    // deterministic if they do.
+    ordered_fragments.sort_by_key(|(_, _, fragment_id, fragment)| {
+        (
+            fragment
+                .actors
+                .keys()
+                .copied()
+                .min()
+                .map(|actor_id| actor_id.as_raw_id())
```
tabVersion, line 664: **Correctness concern: implicit ordering invariant for no-shuffle ensembles**

The sort by `min(preview_actor_id)` is the **sole guarantee** that no-shuffle ensemble members get their actor IDs allocated in the same relative order as the original single-phase renderer. This is correct today because the preview allocator is a single monotonically-increasing counter, but it is an implicit invariant — if `render_actors_with_allocator` ever changes traversal order or the preview allocator becomes non-monotonic, no-shuffle pairings will silently break with no compile-time or runtime error.

Suggestion: add a `debug_assert!` after materialization in `build_reschedule_from_context` to verify the invariant explicitly:

```rust
#[cfg(debug_assertions)]
{
    let materialized_normalized = build_normalized_layout(&rendered.fragments);
    let preview_normalized = build_normalized_layout(&previewed.fragments);
    assert_eq!(materialized_normalized, preview_normalized,
        "materialize changed the normalized layout; preview allocation order is not stable");
}
```

This turns a silent corruption into a loud failure during development/testing.

src/meta/src/stream/scale.rs
```diff
@@ -68,6 +69,97 @@ use crate::controller::utils::{
...
+                fragment_id
+            )));
+        };
+
+        let rendered_layout = build_normalized_fragment_layout(rendered_fragment);
+        let current_layout = build_normalized_fragment_layout(prev_fragment);
+
+        if rendered_layout != current_layout {
+            return Ok(false);
+        }
```
tabVersion, line 161: **Correctness concern: `rendered_layout_matches_current` only checks fragments present in `render_result`**

This is fine for the reschedule path (as the doc comment correctly states), but worth noting: if a caller ever reuses this function for a path where fragment **deletion** is semantically meaningful (e.g., the current layout has fragment X but the rendered layout does not), the current logic would return `true` (match) even though fragments were dropped.

The function name `rendered_layout_matches_current` doesn't convey this asymmetry. Consider renaming to something like `rendered_fragments_unchanged` or adding a brief doc note that this is a **subset check** (rendered ⊆ current), not a full bidirectional equality.

src/meta/src/barrier/worker.rs
```diff
@@ -282,22 +283,29 @@ fn build_reschedule_from_context(
...
-        actor_id_counter,
+    let all_prev_fragments = database_info
+        .fragment_infos()
+        .map(|fragment| (fragment.fragment_id, fragment))
+        .collect();
+
+    let previewed = preview_actor_assignments(
         &worker_nodes,
         adaptive_parallelism_strategy,
         &context.loaded,
```
tabVersion, line 300: **Correctness concern: TOCTOU between preview comparison and materialization**

The `all_prev_fragments` snapshot is taken from `database_info.fragment_infos()` before the preview comparison, and the same snapshot is reused for `build_reschedule_commands` after materialization. This is correct **as long as** `database_info` is immutable for the duration of this function.

If `database_info` could be mutated concurrently (e.g., by a concurrent DDL or another reschedule), then the preview could see layout-A == current-A (no-op), but by the time materialization happens, current has moved to layout-B, making the skip incorrect.

From my reading, this runs inside the barrier worker's serialized path, so this should be safe. But it would be worth adding a brief comment noting this assumption, since the correctness of the no-op skip depends on it.

src/meta/src/stream/scale.rs
```diff
@@ -68,6 +69,97 @@ use crate::controller::utils::{
...
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+struct NormalizedFragmentLayout {
+    distribution_type: DistributionType,
+    actors: Vec<NormalizedActor>,
+}
+
+fn normalize_actor_info(info: &InflightActorInfo) -> NormalizedActor {
+    let vnode_bitmap = info.vnode_bitmap.as_ref().map(|bitmap| {
+        bitmap
```
tabVersion, line 90: **Nit / robustness: `NormalizedActor` derives `PartialOrd, Ord` — field ordering matters**

The derived `Ord` compares fields in declaration order: `worker_id`, then `vnode_bitmap`, then `splits`. This means two actors on the same worker with `vnode_bitmap: None` (stateless/Single distribution) and empty splits will be considered identical by the sort + equality check.

This is actually correct for no-op detection (if we have N such identical actors in both current and rendered, `Vec` comparison preserves count). Just worth being aware that any future field additions to `NormalizedActor` must be added in the right position to maintain meaningful ordering, or `Ord` should be manually implemented.

PR #25090, opened 2026-03-17
feat(sink): support serverless task for snowflake sink

xxhZs:

This PR adds support for serverless Snowflake tasks in Snowflake v2 sinks.

It introduces two new task options:

- `task.serverless`: creates the Snowflake merge task as a serverless task instead of using a user-managed warehouse.
- `task.target_completion_interval`: optionally sets `TARGET_COMPLETION_INTERVAL` for Snowflake serverless tasks.

The task creation logic now supports both warehouse-based and serverless tasks. The merge task also checks whether the intermediate table has data before running, so it can skip unnecessary work when the table is empty.

## Checklist

## Documentation

<details>

Snowflake v2 sinks can now use Snowflake serverless tasks via `task.serverless` and optionally configure `task.target_completion_interval`.

</details>


hzxa21:
For some reason, the PR description of your recent PRs seem to be filling in the wrong sections. Please polish them for this PR and make sure description of your future PRs are well-formatted and polished.

This can help reviewer and for future tracking.

src/connector/src/sink/snowflake_redshift/snowflake.rs
```diff
@@ -1099,37 +1115,49 @@ fn build_create_merge_into_task_sql(snowflake_task_context: &SnowflakeTaskContex
...
-        WHERE dedupe_id = 1
-    ) AS source
-    ON {pk_names_eq_str}
-    WHEN MATCHED AND source."{snowflake_sink_op}" IN (2, 4) THEN DELETE
-    WHEN MATCHED AND source."{snowflake_sink_op}" IN (1, 3) THEN UPDATE SET {all_column_names_set_str}
-    WHEN NOT MATCHED AND source."{snowflake_sink_op}" IN (1, 3) THEN INSERT ({all_column_names_str}) VALUES ({all_column_names_insert_str});
-
-    DELETE FROM {cdc_table_name}
-    WHERE "{snowflake_sink_row_id}" <= :max_row_id;
+    IF (EXISTS (SELECT 1 FROM {cdc_table_name} LIMIT 1)) THEN
```
hzxa21: Why do we need the `EXISTS` here?

src/connector/src/sink/snowflake_redshift/snowflake.rs
```diff
@@ -1244,28 +1274,30 @@ WAREHOUSE = test_warehouse
...
-        WHERE dedupe_id = 1
-    ) AS source
-    ON target."v1" = source."v1"
-    WHEN MATCHED AND source."__op" IN (2, 4) THEN DELETE
-    WHEN MATCHED AND source."__op" IN (1, 3) THEN UPDATE SET target."v1" = source."v1", target."v2" = source."v2"
-    WHEN NOT MATCHED AND source."__op" IN (1, 3) THEN INSERT ("v1", "v2") VALUES (source."v1", source."v2");
-
-    DELETE FROM "test_db"."test_schema"."test_cdc_table"
-    WHERE "__row_id" <= :max_row_id;
+    IF (EXISTS (SELECT 1 FROM "test_db"."test_schema"."test_cdc_table" LIMIT 1)) THEN
```
hzxa21: ditto

src/connector/src/sink/snowflake_redshift/snowflake.rs
```diff
@@ -1292,28 +1326,30 @@ WAREHOUSE = multi_pk_warehouse
...
-        WHERE dedupe_id = 1
-    ) AS source
-    ON target."id1" = source."id1" AND target."id2" = source."id2"
-    WHEN MATCHED AND source."__op" IN (2, 4) THEN DELETE
-    WHEN MATCHED AND source."__op" IN (1, 3) THEN UPDATE SET target."id1" = source."id1", target."id2" = source."id2", target."val" = source."val"
-    WHEN NOT MATCHED AND source."__op" IN (1, 3) THEN INSERT ("id1", "id2", "val") VALUES (source."id1", source."id2", source."val");
-
-    DELETE FROM "test_db"."test_schema"."cdc_multi_pk"
-    WHERE "__row_id" <= :max_row_id;
+    IF (EXISTS (SELECT 1 FROM "test_db"."test_schema"."cdc_multi_pk" LIMIT 1)) THEN
```
hzxa21: ditto

src/connector/src/sink/snowflake_redshift/snowflake.rs
```diff
@@ -1099,37 +1115,49 @@ fn build_create_merge_into_task_sql(snowflake_task_context: &SnowflakeTaskContex
...
+            .map(|interval| format!("TARGET_COMPLETION_INTERVAL = '{}'", interval))
+    } else {
+        Some(format!("WAREHOUSE = {}", warehouse.as_ref().unwrap()))
+    };
+
     format!(
         r#"CREATE OR REPLACE TASK {task_name}
-WAREHOUSE = {warehouse}
+{compute_clause}
 SCHEDULE = '{writer_target_interval_seconds} SECONDS'
```
hzxa21, line 1161: The snowflake document explicitly mentions 
> Define the target stream using the WHEN clause. (Do not include the SCHEDULE parameter.)
for serverless task. https://docs.snowflake.com/en/user-guide/tasks-triggered#create-a-triggered-task

We don't seem to follow their doc at all. Have you tested this?

xxhZs: This parameter is valid for serverless tasks. The triggered-task doc is specifically about tasks defined with a `WHEN SYSTEM$STREAM_HAS_DATA(...)` clause, where `SCHEDULE` should not be included:
https://docs.snowflake.com/en/user-guide/tasks-triggered

Our sink still creates a scheduled task, not a triggered task. For scheduled serverless tasks, Snowflake's `CREATE TASK` docs allow `TARGET_COMPLETION_INTERVAL`
https://docs.snowflake.com/en/sql-reference/sql/create-task

xxhZs:
@hzxa21  Can proceed with the review.

hzxa21:
I think we should add validation for combinations of the new task options before accepting the sink config.

Right now `SnowflakeV2Config::from_btreemap` accepts `task.target_completion_interval` even when `task.serverless` is not true, but `build_create_merge_into_task_sql` only emits `TARGET_COMPLETION_INTERVAL` in the serverless branch, so the option is silently ignored for warehouse-backed tasks. Similarly, `task.serverless = true` can be combined with `warehouse`, but `build_snowflake_task_ctx_jdbc_client` skips the warehouse when serverless is true, so the warehouse value is silently ignored.

Can we reject these combinations during config parsing? Suggested checks:

- `task.target_completion_interval` requires `task.serverless = true`.
- `task.serverless = true` must not be combined with `warehouse`.
- Optionally validate interval format up front, e.g. `<num> HOURS|MINUTES|SECONDS`.

This would make invalid `WITH` options fail in RisingWave DDL instead of producing a Snowflake task that does not match the user config.

src/connector/src/sink/snowflake_redshift/snowflake.rs
```diff
@@ -228,6 +235,16 @@ impl SnowflakeV2Config {
                 SINK_TYPE_UPSERT
             )));
         }
+        if config.task_target_completion_interval.is_some() && !config.task_serverless {
```
hzxa21, line 250: Could we also gate these task options on the Snowflake sink type being `upsert`? The new checks reject invalid serverless combinations, but `type = append-only` with both `task.serverless = true` and `task.target_completion_interval` still passes this block. Later, these fields are only copied into `SnowflakeTaskContext` inside the `if !is_append_only` branch, so append-only configs accept the options and then ignore them.

I think we should reject `task.serverless` / `task.target_completion_interval` unless `config.r#type == SINK_TYPE_UPSERT` (or add explicit append-only semantics/tests if they are meant to apply there).

hzxa21 (approved):
LGTM. The previous task option validation issues are fixed, and the focused Snowflake/config checks passed.

PR #25101, opened 2026-03-18
fix(mysql-cdc): tolerate 2-column SHOW BINARY LOGS results

### LINKED ISSUE #25100: bug(mysql-cdc): MySQL CDC panics when SHOW BINARY LOGS returns 2 columns instead of 3
### Describe the bug

MySQL CDC can panic in the meta source worker when monitoring upstream binlog progress.

The current implementation assumes `SHOW BINARY LOGS` always returns 3 columns:

- `Log_name`
- `File_size`
- `Encrypted`

But this is not portable across MySQL-compatible variants and versions.

## Version

Observed on `v2.8.0`.

This appears to be a regression introduced by:
- [PR #23692](https://github.com/risingwavelabs/risingwave/pull/23692)

That PR added MySQL binlog progress monitoring, including a `SHOW BINARY LOGS` query in the CDC enumerator.

## Error

```text
thread 'rw-main' panicked at /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/mysql_common-0.35.5/src/row/convert/mod.rs:84:39:
Couldn't convert Row { Log_name: Bytes("mysql-bi.."), File_size: Bytes("53688032..") } to type (alloc::string::String, u64, alloc::string::String). (see FromRow documentation)
```

The panic originates from tuple decoding in:

- `risingwave_connector::source::cdc::enumerator::DebeziumSplitEnumerator<Mysql>::query_binlog_files`

## Root Cause

The code currently does:

```rust
conn.query_map(
    "SHOW BINARY LOGS",
    |(log_name, file_size, _encrypted): (String, u64, String)| (log_name, file_size),
)
```

This hardcodes the result shape to 3 columns.

However, `SHOW BINARY LOGS` is not consistently 3-column across versions / distributions:

- Oracle MySQL 5.7: 2 columns
  - `Log_name`, `File_size`
- Oracle MySQL 8.0.14+ / 8.4 / 9.x: 3 columns
  - `Log_name`, `File_size`, `Encrypted`
- MariaDB: 2 columns
  - `Log_name`, `File_size`
- Percona Server: behavior depends on series / implementation, so assuming 3 columns is unsafe

In other words, only the first two columns are portable. The third column is optional.

## Suggested Fix

Avoid tuple decoding with a fixed `(String, u64, String)` shape.

Instead:

- fetch rows as `mysql_async::Row`
- decode only column 0 as `String`
- decode only column 1 as `u64`
- ignore any extra columns

This makes the implementation compatible with both 2-column and 3-column servers.

### Error message/log

```text
thread 'rw-main' panicked at /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/mysql_common-0.35.5/src/row/convert/mod.rs:84:39:
Couldn't convert Row { Log_name: Bytes("mysql-bi.."), File_size: Bytes("53688032..") } to type (alloc::string::String, u64, alloc::string::String). (see FromRow documentation)


The panic originates from tuple decoding in:

- `risingwave_connector::source::cdc::enumerator::DebeziumSplitEnumerator<Mysql>::query_binlog_files`
```

### To Reproduce

_No response_

### Expected behavior

RisingWave should tolerate both result shapes:

1. 2 columns: `(Log_name, File_size)`
2. 3 columns: `(Log_name, File_size, Encrypted)`

Only `Log_name` and `File_size` are needed for current logic.

## Actual Behavior

If the upstream server returns only 2 columns, RisingWave panics in the CDC monitor path instead of handling the result gracefully.

_No response_

### The version of RisingWave

v2.8.0

### Additional context

This issue was not present before the binlog progress monitoring logic added in `#23692`.

Relevant compatibility references:

- Oracle MySQL 5.7 documentation shows 2 columns for `SHOW BINARY LOGS`
- Oracle MySQL 8.0.14 release notes introduced showing encryption status in `SHOW BINARY LOGS`
- MariaDB documentation shows 2 columns
- Percona behavior may vary by series, so fixed 3-column decoding is not safe

qccash:

- Closes #25100

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/src/source/cdc/enumerator/mod.rs
+++ b/src/connector/src/source/cdc/enumerator/mod.rs
@@ -21,6 +21,7 @@ use std::sync::Arc;
 use anyhow::{Context, anyhow};
 use async_trait::async_trait;
 use itertools::Itertools;
+use mysql_async::Row;
 use mysql_async::prelude::*;
 use prost::Message;
 use risingwave_common::global_jvm::Jvm;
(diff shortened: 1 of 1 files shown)


wcy-fdu (approved):
Thanks for the fix!

PR #25111, opened 2026-03-18
feat(cdc): support ALTER SOURCE RESET TO

zwang28:

This PR adds support for resetting the CDC source's start_offset to an arbitrary value.
It also adds a check requiring the source to be explicitly paused prior to a reset; this prevents active source readers from overwriting the offset.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/proto/ddl_service.proto
+++ b/proto/ddl_service.proto
@@ -88,6 +88,8 @@ message AlterSourceResponse {
 // Reset CDC source offset to latest (for handling expired binlog/oplog)
 message ResetSourceRequest {
   uint32 source_id = 1;
+  // Optional start offset to set. When unset, clear offsets to None (latest).
+  optional string start_offset = 2;
 }
 
 message ResetSourceResponse {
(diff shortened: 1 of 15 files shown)


wcy-fdu:
Two correctness issue reviewed by agent:
- Silent success when reset is ignored (major correctness issue)
The PR can return success for ALTER SOURCE ... RESET TO '...' even when the reset is not actually applied.
In source_executor, if the source is not explicitly paused, the ResetSource mutation is ignored with only an info log (src/stream/src/executor/source/source_executor.rs:993+). But frontend/meta still report success (src/frontend/src/handler/reset_source.rs:55).
This creates a correctness risk: users see a successful SQL response while offsets remain unchanged (silent no-op).

- Misleading logs from non-target source actors (minor correctness/observability issue)
The branch order in source_executor checks !is_source_paused_explicitly before checking whether source_id matches the current actor’s source.
As a result, actors for other sources may log “reset ignored because source is not paused,” even though the mutation was never for them (src/stream/src/executor/source/source_executor.rs:993+).
This does not directly break functionality, but it can mislead debugging and operational diagnosis.

Feel free to ignore if it's wrong.

wcy-fdu:
I had some discussion with @hzxa21 a few days ago, and we reach an agreement that "reset source to <offset>" is better implemented in risectl🤔, @tabVersion may have more context.

PR #25114, opened 2026-03-18
fix(connector): reject postgres sink tables with foreign keys

### LINKED ISSUE #25018: bug(cdc sink): pg native cdc sink panicked with FK in downstream
### Describe the bug

Adding a foreign key to the downstream pg could cause panic in cdc pg sink.

### Error message/log

```text
2026-03-02 19:50:20.792	
thread 'rw-streaming' (108) panicked at /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/sea-schema-0.16.0/src/postgres/parser/table_constraints.rs:52:63:
2026-03-02 19:50:20.727
```

### To Reproduce

_No response_

### Expected behavior

_No response_

_No response_

### The version of RisingWave

_No response_

### Additional context

_No response_

xxhZs:

check whether the target Postgres table has foreign key constraints before sink validation continues
perform the same check again when creating the sink writer to avoid reaching the sea-schema foreign key parser panic path
reuse a shared helper for tcp keepalive configuration in the sink
- Closes  https://github.com/risingwavelabs/risingwave/issues/25018

## Checklist

## Documentation

<details>

</details>


src/connector/src/sink/postgres.rs
```diff
@@ -304,13 +357,9 @@ impl PostgresSinkWriter {
...
-                .tcp_keepalive
-                .or_else(|| Some(TcpKeepaliveConfig::default()))
-        } else {
-            None
-        };
+        ensure_no_foreign_key(&config).await?;
+
+        let tcp_keepalive = tcp_keepalive_from_config(&config);
 
         let client = create_pg_client(
```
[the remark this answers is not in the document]
hzxa21: I think this makes sense @xxhZs

xxhZs:
@tabVersion @xiangjinwu  Can proceed with the review.

src/connector/src/sink/postgres.rs
```diff
@@ -182,6 +243,8 @@ impl Sink for PostgresSink {
             )));
         }
 
+        ensure_no_foreign_key(&self.config).await?;
```
hzxa21, line 246: Could we add a targeted regression test for this path? `e2e_test/sink/postgres_sink.slt` already provisions a real Postgres database, so a small negative case there should cover the important behavior: create a Postgres target table that owns an FK constraint, then assert `CREATE SINK` fails with this new error.

For example:

```sql
system ok
PGDATABASE=sink_test psql -c "CREATE TABLE pg_fk_parent (id BIGINT PRIMARY KEY); CREATE TABLE pg_fk_child (id BIGINT PRIMARY KEY, parent_id BIGINT REFERENCES pg_fk_parent(id));"

statement ok
CREATE TABLE rw_fk_child (id BIGINT, parent_id BIGINT);

statement error Postgres sink does not support target table "public"."pg_fk_child" with foreign key constraints
CREATE SINK postgres_fk_sink FROM rw_fk_child WITH (
    connector='postgres',
    host='$PGHOST',
    port='$PGPORT',
    user='$PGUSER',
    password='$PGPASSWORD',
    database='sink_test',
    table='pg_fk_child',
    type='upsert',
    primary_key='id',
);
```

This would exercise the real Postgres `pg_constraint` query, including schema/table matching, and verify that `CREATE SINK` rejects a target table that owns an FK constraint. A small positive/control case for sinking to `pg_fk_parent` would also be useful if we want to lock in the `conrelid` vs `confrelid` direction.

xxhZs: Added a targeted regression test in 18f093ff1a. It creates a Postgres child table that owns an FK and asserts CREATE SINK is rejected, plus a parent-table control case to lock in the conrelid/confrelid direction.

hzxa21 (approved):

PR #7264, opened 2023-01-09
feat(kafka): option of setting max number of messages

lmatz:
I hereby agree to the terms of the [Singularity Data, Inc. Contributor License Agreement](https://gist.github.com/skyzh/0663682a70b0edde7ae991492f2314cb#file-s9y_cla).

## What's changed and what's your intention?
As per the title, it will be used for Kafka performance testing setting as discussed with @huangjw806, who will use this to control how much time a query runs while all the query reads from the same kafka topic (one by one)

This is only used internally, so it will not go into the Doc.

## Checklist

- [x] I have written necessary rustdoc comments
- [x] I have added necessary unit tests and integration tests
- [x] All checks passed in `./risedev check` (or alias, `./risedev c`)

PR #25121, opened 2026-03-19
fix(source): resolve Kafka backfill/batch hang on transaction control records

### LINKED ISSUE #24717: Bug: Kafka Backfill Hangs
### Describe the bug

RisingWave's backfill hangs on Kafka topics where the offset of the last regular message is not topic offset - 1.
This happens when Kafka transactions are used.

(The topic where I first observed the behavior is produced by Kafka Streams, so all of these topics could be affected, but I didn't test this.)

### Error message/log

```text
N/A
```

### To Reproduce

Start Kafka:
```sh
docker run --name kafka --rm -p 9092:9092 apache/kafka:latest
```

Create a topic:
```sh
./kafka-topics.sh --bootstrap-server localhost:9092 --create --topic demo --partitions 1 --replication-factor 1
```

Now run the attached Rust program [1]. It will send a message in a transaction.

Note that the topic's offset (LEO) has increased by **two**!

What's going on? In addition to our message, Kafka has produced a `COMMIT` control message.
I don't think it's possible to display it in a consumer, but inside the Kafka container, we can use `kafka-dump-log`:

```sh
docker exec -it kafka \
	/opt/kafka/bin/kafka-dump-log.sh --files /tmp/kraft-combined-logs/demo-0/00000000000000000000.log --print-data-log
```

```
Dumping /tmp/kraft-combined-logs/demo-0/00000000000000000000.log
Log starting offset: 0
baseOffset: 0 lastOffset: 0 count: 1 baseSequence: 0 lastSequence: 0 producerId: 0 producerEpoch: 0 partitionLeaderEpoch: 0 isTransactional: true isControl: false deleteHorizonMs: OptionalLong.empty position: 0 CreateTime: 1770284165146 size: 88 magic: 2 compresscodec: none crc: 966004153 isvalid: true
| offset: 0 CreateTime: 1770284165146 keySize: 8 valueSize: 12 sequence: 0 headerKeys: [] key: test-key payload: test-payload
baseOffset: 1 lastOffset: 1 count: 1 baseSequence: -1 lastSequence: -1 producerId: 0 producerEpoch: 0 partitionLeaderEpoch: 0 isTransactional: true isControl: true deleteHorizonMs: OptionalLong.empty position: 88 CreateTime: 1770284165190 size: 78 magic: 2 compresscodec: none crc: 2017695611 isvalid: true
| offset: 1 CreateTime: 1770284165190 keySize: 4 valueSize: 6 sequence: -1 headerKeys: [] endTxnMarker: COMMIT coordinatorEpoch: 0
```

Now to the problem.
Let's create a source in RisingWave:
```sql
create source test_kafka_source (value bytea)
with (
  connector = 'kafka',
  topic = 'demo',
  properties.bootstrap.server = 'localhost:9092'
)
format plain encode bytes;
```

And an MV. **This is going to hang!**
```sql
set BACKGROUND_DDL = false;
create materialized view test_kafka_mv as (select * from test_kafka_source);
```

If we look at the matching `__internal…` table, we'll see a `target_offset`. It is `topic offset - 1`.
That's the offset of the hidden COMMIT message, not of the last _visible_ message!

⇒ We're waiting for a message that will never come. (Until new messages are sent, so the problem probably won't be very visible on high-throughput topics — while it makes it impossible to use RW with low-volume topics using transactions.)

```sql
dev=> show jobs;
 Id |                                  Statement                                  | Create Type |    Progress     
----+-----------------------------------------------------------------------------+-------------+-----------------
  8 | CREATE MATERIALIZED VIEW test_kafka_mv AS (SELECT * FROM test_kafka_source) | FOREGROUND  | 1 rows consumed
(1 row)

dev=> show internal tables;
                    Name                     
---------------------------------------------
 __internal_test_kafka_source_1_source_0
 __internal_test_kafka_mv_2_sourcebackfill_1
(2 rows)

dev=> select * from  __internal_test_kafka_mv_2_sourcebackfill_1;
 partition_id |                               backfill_progress                               
--------------+-------------------------------------------------------------------------------
 0            | {"num_consumed_rows": 1, "state": {"Backfilling": "0"}, "target_offset": "1"}
(1 row)
```

In the attached program I'm using `enable.partition.eof` on my consumer, maybe this would work in RW, too?

[1]
```rust
use rdkafka::consumer::{BaseConsumer, Consumer};
use rdkafka::error::KafkaError;
use rdkafka::message::Message;
use rdkafka::producer::{FutureProducer, FutureRecord, Producer};
use rdkafka::ClientConfig;
use rdkafka::TopicPartitionList;
use std::time::Duration;

const BROKERS: &str = "localhost:9092";
const TOPIC: &str = "demo";
const PARTITION: i32 = 0;

#[tokio::main]
async fn main() -> anyhow::Result<()> {
    print_offsets()?;
    send_message_in_transaction().await?;
    print_offsets()?;
    dump_messages().await?;

    println!("Done.");
    Ok(())
}

async fn send_message_in_transaction() -> anyhow::Result<()> {
    let producer: FutureProducer = ClientConfig::new()
        .set("bootstrap.servers", BROKERS)
        .set("transactional.id", "demo-tx")
        .set("enable.idempotence", "true")
        .set("acks", "all")
        .create()?;

    producer.init_transactions(Duration::from_secs(10))?;

    println!("Sending a message in a transaction...");
    producer.begin_transaction()?;
    producer
        .send(
            FutureRecord::to(TOPIC)
                .partition(0)
                .key("test-key")
                .payload("test-payload"),
            Duration::from_secs(5),
        )
        .await
        .map_err(|(e, _)| e)?;
    producer.commit_transaction(Duration::from_secs(10))?;
    println!("Message sent and transaction committed.");

    Ok(())
}

fn print_offsets() -> anyhow::Result<()> {
    let consumer: BaseConsumer = ClientConfig::new()
        .set("bootstrap.servers", BROKERS)
        .set("group.id", "demo-group")
        .create()?;

    let offsets = consumer.fetch_watermarks(TOPIC, PARTITION, Duration::from_secs(5))?;
    println!("Offsets for topic '{}', partition {}: low={}, high={}", TOPIC, PARTITION, offsets.0, offsets.1);

    Ok(())
}

async fn dump_messages() -> anyhow::Result<()> {
    let consumer: BaseConsumer = ClientConfig::new()
        .set("bootstrap.servers", BROKERS)
        .set("group.id", "demo-group")
        .set("auto.offset.reset", "earliest")
        .set("enable.partition.eof", "true")
        .set("enable.auto.commit", "false")
        .set("enable.auto.offset.store", "false")
        .create()?;

    let mut tpl = TopicPartitionList::new();
    tpl.add_partition(TOPIC, PARTITION);
    consumer.assign(&tpl)?;

    println!("Messages from {TOPIC}[{PARTITION}]:");
    loop {
        match consumer.poll(Duration::from_millis(250)) {
            None => {}

            Some(Ok(msg)) => {
                let offset = msg.offset();
                let key = msg.key().map(|k| String::from_utf8_lossy(k).into_owned());
                let payload = msg
                    .payload()
                    .map(|p| String::from_utf8_lossy(p).into_owned());

                println!(
                    "[offset {}] key={:?} payload={:?}",
                    offset,
                    key.as_deref(),
                    payload.as_deref()
                );
            }

            Some(Err(KafkaError::PartitionEOF(partition))) => {
                if partition == 0 {
                    println!("Reached EOF for {TOPIC}[{PARTITION}].");
                    break;
                }
            }

            Some(Err(e)) => {
                eprintln!("Kafka error: {e}");
            }
        }
    }

    Ok(())
}
```

### Expected behavior

_No response_

I tested this on RisingWave cloud and on two local deployments.

### The version of RisingWave

PostgreSQL 13.14.0-RisingWave-2.7.2 (30301dc965a6f30c08de859e2be0e6cb1b66f6b0)

### Additional context

_No response_

### LINKED ISSUE #24829: source backfill cannot complete when consuming from an idle Kafka partition ending in a control record

tabVersion:
## Summary

- Fix source backfill (`CREATE MATERIALIZED VIEW`) and batch queries (`SELECT * FROM source`) hanging indefinitely on Kafka topics where the last record in a partition is a transaction control record (commit/abort marker)
- Affects any topic written by a transactional producer (Kafka Streams, Flink, Debezium, Kafka Connect with EOS)
- Root cause: `isolation.level=read_committed` makes librdkafka skip control records, but `fetch_watermarks()` high_watermark includes their offsets — the stop condition points to an invisible offset never delivered

## Changes

**`SourceCtrlOpts`** — Added `enable_partition_eof: bool` field. Only batch source and source backfill set it to `true`; all other executors (streaming, FS fetch, Iceberg, CDC) remain `false`.

**`KafkaSplitReader`** — When `enable_partition_eof=true`:
- Sets rdkafka `enable.partition.eof` to `"true"`
- Auto-injects `stop_offset = high_watermark` for bounded readers
- Handles `KafkaError::PartitionEOF` gracefully (per-partition tracking)
- Pre-push guard (`>= stop_offset`) + post-push fast-path (`== stop_offset - 1`)
- Signals completion via `Arc<AtomicBool>` only when ALL partitions done

**`SourceBackfillExecutor`** — Checks `bounded_eof_flag` on each barrier. When set, marks remaining `Backfilling` splits as `Finished`. State transition persisted in same barrier epoch for crash safety.

**Batch source** — Stream terminates naturally when reader completes; no explicit flag check needed.

## Key Design Decisions

| Decision | Rationale |
|---|---|
| `enable.partition.eof` signal, not `consumer.position()` | `position()` is undefined before first `poll()`. EOF covers all invisible-tail cases. |
| Batch + backfill only, not streaming | Streaming runs forever, doesn't need completion signals. |
| `Arc<AtomicBool>` side-channel | `select_with_strategy` doesn't expose Right-side exhaustion. Flag set only after all data yielded. |
| Pre-push + post-push guards | Pre-push prevents out-of-range rows (compaction gaps). Post-push is fast-path for normal case. Complementary, not redundant. |

## Test plan

- [ ] `cargo check` — all crates compile clean
- [ ] `cargo clippy` — zero warnings
- [ ] `./risedev c` — all checks passed (fmt, clippy, typos, machete)
- [ ] Verified streaming `SourceExecutor` unaffected (zero `bounded_eof_flag` references, `enable_partition_eof: false`)
- [ ] E2E test with transactional Kafka producer: `CREATE MATERIALIZED VIEW` completes in seconds (was hanging indefinitely)
- [ ] E2E test: `SELECT * FROM source` returns rows without hanging
- [ ] Empty topic edge case: PartitionEOF handled gracefully when no stop_offsets exist

Closes #24829, closes #24717

### DIFF
--- a/src/stream/src/executor/source/source_backfill_executor.rs
+++ b/src/stream/src/executor/source/source_backfill_executor.rs
@@ -360,6 +360,15 @@ impl<S: StateStore> SourceBackfillExecutorInner<S> {
             split_handoff_stream,
             prefer_split_handoff,
         );
+        // Backfill timing for one split:
+        // 1. the bounded reader emits historical rows as `Chunk`;
+        // 2. when it catches up to the split's current Kafka head, it emits `SplitHandoff`;
+        // 3. `SourceBackfillExecutor` switches that split from `Backfilling` to
+        //    `SourceCachingUp` / `Finished` and only then acks the handoff;
+        // 4. after the ack, the reader is allowed to suppress that split locally.
+        //
+        // This ordering closes the gap where the reader could stop producing a split
+        // before the upstream side had taken over responsibility for new rows of that split.
         // Wrap the plain chunk stream with an in-band EOF sentinel.
         // When the bounded reader finishes (stream terminates), the chained
         // `BoundedEof` event is delivered in causal order after all data chunks,
@@ -534,7 +543,17 @@ impl<S: StateStore> SourceBackfillExecutorInner<S> {
                 * WAIT_BARRIER_MULTIPLE_TIMES;
             let mut last_barrier_time = Instant::now();
 
-            // The main logic of the loop is in handle_upstream_row and handle_backfill_row.
+            // Backfill stage timing:
+            // - Left stream (`input`) is the live upstream source executor. Its rows are filtered by
+            //   the per-split backfill state in `handle_upstream_row`.
+            // - Right stream is the bounded backfill reader. It emits either historical chunks or
+            //   `SplitHandoff` events for individual splits.
+            // - A split is handed off as soon as the bounded reader catches up to the split's current
+            //   head; we do not wait for all splits to finish before switching that split.
+            // - `BoundedEof` only means the bounded reader as a whole has drained all unfinished splits.
+            //
+            // The main state transitions are implemented in `handle_upstream_row`,
+            // `handle_backfill_row`, and the `SplitHandoff` branch below.
             'backfill_loop: while let Some(either) = backfill_stream.next().await {
                 match either {
                     // Upstream
@@ -820,6 +839,13 @@ impl<S: StateStore> SourceBackfillExecutorInner<S> {
                                 backfill_offset,
                                 ack_tx,
                             } => {
+                                // This is the split-level ownership handoff point.
+                                //
+                                // Kafka `PartitionEOF` only means "caught up for now"; new rows for the
+                                // same split may still arrive later. Therefore the bounded reader must not
+                                // stop covering the split until the executor has switched that split's
+                                // state and the live upstream path can safely take over. The ack makes this
+                                // ordering explicit.
                                 tracing::info!(
                                     source_id = %self.source_id,
                                     split_id = %split_id,


src/connector/src/source/kafka/source/reader.rs
```diff
@@ -282,42 +334,64 @@ impl KafkaSplitReader {
...
+                        if !res.is_empty() {
+                            yield std::mem::take(&mut res);
+                        }
+                        if let Some(ref flag) = self.bounded_eof_flag {
+                            flag.store(true, Ordering::Release);
+                        }
+                        break 'for_outer_loop;
+                    }
+                    continue;
+                }
```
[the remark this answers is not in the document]
tabVersion: Good catch on the theoretical gap. In practice this is safe to leave as-is:

- `stop_offset` is set to `high_watermark` at reader creation time. For messages beyond that offset to appear in the same `ready_chunks` batch, new data would have to arrive between the watermark snapshot and consumption — an extremely narrow window.
- Even if a few extra rows leak through, they are harmless: for **backfill**, the executor's `handle_upstream_row` deduplicates by offset; for **batch**, it is equivalent to a slightly later snapshot.
- The original stop-offset logic (before this PR) had the same behavior — `o.remove()` followed by `continue` without guarding the rest of the batch.

Adding a `finished_partitions` set would increase per-message overhead for a race condition that is both unlikely and consequence-free.

src/connector/src/source/kafka/source/reader.rs
```diff
@@ -282,42 +334,64 @@ impl KafkaSplitReader {
...
+                    if stop_offsets.is_empty() {
+                        yield std::mem::take(&mut res);
+                        if let Some(ref flag) = self.bounded_eof_flag {
+                            flag.store(true, Ordering::Release);
                         }
-
-                        continue;
+                        break 'for_outer_loop;
                     }
                 }
```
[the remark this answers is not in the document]
tabVersion: Same reasoning as above — this mirrors the pre-existing behavior of the original `stop_offsets.entry(split_id)` check, which also `o.remove()`s and continues without guarding subsequent messages in the batch. The bounded reader is a best-effort snapshot up to `high_watermark`; a handful of extra rows from a concurrent producer are deduplicated downstream.

src/connector/src/source/kafka/source/reader.rs
```diff
@@ -252,11 +280,35 @@ impl KafkaSplitReader {
...
-            let msgs: Vec<_> = msgs
-                .into_iter()
-                .collect::<std::result::Result<_, KafkaError>>()?;
+        'for_outer_loop: for chunk in self.consumer.stream().ready_chunks(max_chunk_size) {
+            let mut msgs = Vec::with_capacity(chunk.len());
+            for item in chunk {
+                match item {
+                    Ok(msg) => msgs.push(msg),
+                    Err(KafkaError::PartitionEOF(partition)) if enable_eof => {
+                        let split_id: SplitId = partition.to_string().into();
```
xiangjinwu, line 465: This `to_string` will be affected by #25076 right?

src/connector/src/source/kafka/source/reader.rs
```diff
@@ -168,6 +184,13 @@ impl SplitReader for KafkaSplitReader {
                 .expect("max.num.messages expect usize"),
         };
 
+        let has_stop_offsets = offsets.values().any(|(_, stop)| stop.is_some());
+        let bounded_eof_flag = if enable_partition_eof && has_stop_offsets {
```
xiangjinwu: When is `has_stop_offsets` false and `enable_partition_eof` is true?

src/connector/src/source/kafka/source/reader.rs
```diff
@@ -330,6 +404,20 @@ impl KafkaSplitReader {
...
+
+            // After processing chunk: check if EOF events emptied stop_offsets.
+            // Only check when in bounded mode (enable_eof), otherwise streaming
+            // sources with no stop_offsets would break after the first chunk.
+            if enable_eof && stop_offsets.is_empty() {
+                if !res.is_empty() {
+                    yield std::mem::take(&mut res);
+                }
+                if let Some(ref flag) = self.bounded_eof_flag {
+                    flag.store(true, Ordering::Release);
```
xiangjinwu: Is it possible that `enable_eof == true` but `bounded_eof_flag.is_none()`?

src/connector/src/source/kafka/source/reader.rs
```diff
@@ -68,8 +71,15 @@ impl SplitReader for KafkaSplitReader {
         let bootstrap_servers = &properties.connection.brokers;
         let broker_rewrite_map = properties.privatelink_common.broker_rewrite_map.clone();
 
-        // disable partition eof
-        config.set("enable.partition.eof", "false");
+        let enable_partition_eof = source_ctx.source_ctrl_opts.enable_partition_eof;
+        config.set(
+            "enable.partition.eof",
```
hzxa21, line 82: We can include https://docs.confluent.io/platform/current/clients/librdkafka/html/md_CONFIGURATION.html in the comment to let people know an error will be thrown by rdkafka when the partition eof is met.

src/connector/src/source/kafka/source/reader.rs
hzxa21: I have the same question

hzxa21: We should list out the possible combinations of `enable_partition_eof` and `has_stop_offsets` with their use case.

src/connector/src/source/kafka/source/reader.rs
```diff
@@ -282,42 +334,64 @@ impl KafkaSplitReader {
...
-                            "stop offset reached for split {}, stop reading, offset: {}, stop offset: {}",
-                            o.key(),
-                            cur_offset,
-                            stop_offset
-                        );
+                let split_id: SplitId = msg.partition().to_string().into();
+
+                // PRE-PUSH GUARD: stop_offset is exclusive — skip rows at or beyond it
+                if let Entry::Occupied(o) = stop_offsets.entry(split_id.clone())
+                    && cur_offset >= *o.get()
```
hzxa21: Notes:
I think the changes added here fix another bug when consuming from a log compacted kafka topic. The original recorded stop offset can disapear after the async log compaction. Without the pre-push guard added here, the outer loop may never stop.

src/connector/src/source/kafka/source/reader.rs
hzxa21: Same question here.

hzxa21: I feel like we should have a struct wrapping enable_eof, stop_offset and bounded_eof_flag to make the code clearer.

src/connector/src/source/kafka/source/reader.rs
```diff
@@ -252,11 +280,35 @@ impl KafkaSplitReader {
...
+                    if !res.is_empty() {
+                        yield std::mem::take(&mut res);
+                    }
+                    if let Some(ref flag) = self.bounded_eof_flag {
+                        flag.store(true, Ordering::Release);
+                    }
+                    break 'for_outer_loop;
+                }
+                continue;
+            }
```
hzxa21: I felt that the codes here are unneccessary given that we have the check in L411

src/stream/src/executor/source/source_backfill_executor.rs
```diff
@@ -644,14 +654,31 @@ impl<S: StateStore> SourceBackfillExecutorInner<S> {
...
+                                    && flag.load(std::sync::atomic::Ordering::Acquire)
+                                {
+                                    for state in backfill_stage.states.values_mut() {
+                                        if matches!(state.state, BackfillState::Backfilling(_)) {
+                                            tracing::info!(
+                                                source_id = %self.source_id,
+                                                "marking backfill split finished due to bounded reader EOF"
+                                            );
+                                            state.state = BackfillState::Finished;
+                                        }
```
hzxa21: `bounded_eof_flag` introduces an out-of-band runtime signal alongside the reader stream. This makes EOF/completion detection depend on polling shared state at barrier handling time, instead of consuming a first-class event from the reader, which seems a bit hacky to me. 

I feel that this can create race condition and cause data loss. One example I can think of:
kafka partition
```
T1:
1 | 2 | 3 (txn begin control) | 4 | 5 | 6 (txn end control)
T2:
1 | 2 | 3 (txn begin control) | 4 | 5 | 6 (txn end control) | 7 | 8
```
- Before T1: source backfill started
- Between T1 and T2: source backfill stream completed with EOF, set `bounded_eof_flag` to true
- After T2: before the next barrier comes, source upstream stream polls data 7 but ignore, because `BackfillState` is not set to finish.
- Next barrier comes, source backfill is done and source upstream now polls data 8 and yield to downstream. data 7 is lost.

src/connector/src/source/kafka/source/reader.rs
tabVersion: I think yes, and it does not matter here because all data in memory.

src/stream/src/executor/source/source_backfill_executor.rs
tabVersion: After making the EOF signal part of SourceChunkEvent, we take everything in band. The data stream can help preserve the order and should not have similar race condition issue.

src/connector/src/source/kafka/source/reader.rs
tabVersion: resolve because in the current impl, I removed all atomic bool signals, no more confusing cases

src/connector/src/source/kafka/source/reader.rs
tabVersion: Yes, `POST-PUSH` relies on `cur_offset == stop_offset-1`. But if `stop_offset-1` is invisible because of compaction or other reasons, it can hang forever. 
`PRE-PUSH` checks `cur_offset >= stop_offset` as fallback. Upon getting the message after the end_offset, it can return and will not stuck.

src/connector/src/source/kafka/source/reader.rs
tabVersion: One is checking: the chunk contains only EOF without data; and the other one is for chunk with data and EOF signal. We must check EOF signal after handling data chunks, otherwise may lose data.

tabVersion:
rely on https://github.com/risingwavelabs/risingwave/pull/25215 for fixing iceberg issues in CI

src/connector/src/source/kafka/source/reader.rs
hzxa21: I don't get it. The is how the current codes are organized:
```rust
    'for_outer_loop: for chunk in self.consumer.stream().ready_chunks(max_chunk_size) {
            let mut msgs = Vec::with_capacity(chunk.len());
            for item in chunk {
                ...
            }
            
            // section 1
            if msgs.is_empty() {
                // Chunk contained only PartitionEOF events, no data messages.
                // If all partitions are done, break; otherwise wait for more.
                if enable_eof && stop_offsets.is_empty() {
                    if !res.is_empty() {
                        yield std::mem::take(&mut res);
                    }
                    break 'for_outer_loop;
                }
                continue;
            }

            // section 2
            ...

            for msg in &msgs {
                ...
            }

            ...

            // section 3
            if enable_eof && stop_offsets.is_empty() {
                if !res.is_empty() {
                    yield std::mem::take(&mut res);
                }
                break 'for_outer_loop;
            }

            // section 4
            let mut cur = Vec::with_capacity(res.capacity());
            swap(&mut cur, &mut res);
            yield cur;
```

IIUC, section 2 is purely no-op if `msgs.is_empty()` is true. That means we can do something like this
```rust
   'for_outer_loop: for chunk in self.consumer.stream().ready_chunks(max_chunk_size) {
            let mut msgs = Vec::with_capacity(chunk.len());
            for item in chunk {
                ...
            }

            // section 2 (remove section 1)
            ...

            // section 3
            if enable_eof && stop_offsets.is_empty() {
                if !res.is_empty() {
                    yield std::mem::take(&mut res);
                }
                break 'for_outer_loop;
            }

            // modifed section 4 (avoid empty chunk)
            if !res.is_empty() {
                let mut cur = Vec::with_capacity(res.capacity());
                swap(&mut cur, &mut res);
                yield cur;
            }

            
```

Just nits, not a correctness issue.

src/connector/src/source/kafka/source/reader.rs
```diff
@@ -330,6 +393,18 @@ impl KafkaSplitReader {
                     break 'for_outer_loop;
                 }
             }
```
hzxa21: By codex: BoundedEof is now emitted for any normal termination of the bounded reader, but the Kafka reader still has a non-EOF early-exit path when max.num.messages is hit. In that case the stream ends because of the throttle/debug knob, chain(...BoundedEof...) still fires, and backfill state is advanced as if the snapshot were fully consumed. That can make MV creation finish with unread historical rows still missing.

mext.num.messages is introduced [here](https://github.com/risingwavelabs/risingwave/blob/1b615e8798110d8caecfbd86dfcefb865b7166e9/src/connector/src/source/kafka/mod.rs#L125) in #7264 for perf test. That means this PR may affect perf test. cc @lmatz

src/connector/src/source/kafka/source/reader.rs
```diff
@@ -140,6 +138,23 @@ impl SplitReader for KafkaSplitReader {
                     },
                 );
             }
+
+            // Auto-set stop_offset for bounded readers
+            if enable_partition_eof && split.stop_offset.is_none() && low < high {
```
hzxa21: When will `split.stop_offset` be `none`?

If the `stop_offset` is `none` for kafka source backfill, auto set stop offset to the high offset at the time split reader is created can cause data loss.

T1: backfill stream is created with stop offset `5`, upstream stream is at offset `4`
T2: upstream stream is at offset `8` before backfill stream catches up
T3: backfill stream is caught up to offset `5`. emitting BoundedEof, marking backfilling as done. However, upstream is already past `5`. Message (5, 8] will be lost.

tabVersion:
close for fix in https://github.com/risingwavelabs/risingwave/pull/25258

PR #25122, opened 2026-03-19
feat(iceberg): add Iceberg source ingestion metrics

chenzl25:
## Summary
- Add 11 new Prometheus metrics for Iceberg source ingestion observability
- Instrument all 3 Iceberg executors: `IcebergListExecutor`, `IcebergFetchExecutor`, `BatchIcebergFetchExecutor`
- Update Grafana dashboard with panels for all new metrics

### New Metrics

| Metric | Type | Labels | Description |
|--------|------|--------|-------------|
| `iceberg_source_snapshot_lag_seconds` | Gauge | source_id, source_name, table_name | Ingestion lag (0 when caught up) |
| `iceberg_source_snapshots_discovered_total` | Counter | source_id, source_name, table_name | Snapshots discovered via incremental scan |
| `iceberg_source_list_duration_seconds` | Histogram | source_id, source_name, table_name | Time spent planning files from a snapshot |
| `iceberg_source_files_discovered_total` | Counter | source_id, source_name, table_name, file_type | Files discovered per scan by type |
| `iceberg_source_file_read_duration_seconds` | Histogram | table_name | Per-file read duration |
| `iceberg_source_rows_read_total` | Counter | table_name | Rows read from Iceberg source |
| `iceberg_source_files_read_total` | Counter | table_name, file_type | Files read by type |
| `iceberg_source_delete_rows_applied_total` | Counter | table_name, delete_type | Approximate rows removed by delete processing |
| `iceberg_source_delete_files_per_data_file` | Histogram | source_id, source_name, table_name | Delete files attached per data file |
| `iceberg_source_checkpoint_file_count` | Gauge | source_id, source_name, table_name | Files tracked in checkpoint state |
| `iceberg_source_scan_errors_total` | Counter | source_id, source_name, table_name, error_type | Categorized scan errors |

image

### Design Decisions
- Executor-level metrics use `[source_id, source_name, table_name]` labels for full context
- Scan-function metrics use `[table_name]` labels only (matching existing `iceberg_read_bytes`) since the shared `scan_task_to_chunk_with_deletes` function lacks source-level context
- Delete rows metric is explicitly approximate (`sdk_applied_approx` label) and only fires when `handle_delete_files` is enabled
- Snapshot lag reports 0 when ingestion is caught up, not growing snapshot age

## Test plan
- [ ] `cargo check -p risingwave_stream` passes (verified locally)
- [ ] Verify Grafana dashboard renders correctly with `./risedev generate-grafana-dashboard`
- [ ] Deploy with an Iceberg source and verify metrics appear in Prometheus/Grafana
- [ ] Verify snapshot lag goes to 0 when source is caught up
- [ ] Verify scan error counter increments on transient failures


tabVersion:
Can you share a screenshot of new panel added?

chenzl25:
> Can you share a screenshot of new panel added?

In the PR description.

src/stream/src/executor/source/batch_source/batch_iceberg_list.rs
```diff
@@ -244,20 +282,73 @@ impl<S: StateStore> BatchIcebergListExecutor<S> {
...
+            let mut pos_delete_count: u64 = 0;
+
             #[for_await]
             for scan_task in snapshot_scan
                 .plan_files()
                 .await
                 .context("failed to plan iceberg files")?
             {
                 let scan_task = scan_task.context("failed to get scan task")?;
+                data_file_count += 1;
```
Li0k, line 295: correct me: If I understand correctly, if a scan-task includes the same delete-task, will the delete-file be counted repeatedly?

src/stream/src/executor/source/iceberg_fetch_executor.rs
```diff
@@ -600,8 +649,14 @@ impl<S: StateStore> IcebergFetchExecutor<S> {
                         }) => {
                             // TODO: support persist progress after supporting reading part of a file.
                             if true {
```
Li0k, line 643: why need `if true`

src/stream/src/executor/source/iceberg_list_executor.rs
```diff
@@ -308,14 +350,89 @@ async fn incremental_scan_stream(
...
             .context("failed to plan iceberg files")?
         {
-            yield scan_task.context("failed to get scan task")?;
+            let scan_task = scan_task.context("failed to get scan task")?;
+
+            // Count file types: the main task is a data file, deletes are attached.
+            data_file_count += 1;
+            for delete_task in &scan_task.deletes {
+                match delete_task.data_file_content {
+                    DataContentType::EqualityDeletes => eq_delete_count += 1,
```
Li0k, line 386: ditto

src/stream/src/executor/source/batch_source/batch_iceberg_list.rs
chenzl25: Good question. It will be. But actually the delete file metrics are collected by `eq_delete_count` and `pos_delete_count` these number  equal exactly the number of actual file io it would perform.

src/stream/src/executor/source/iceberg_fetch_executor.rs
chenzl25: It is a todo item. We don't support reading part of a file.

chenzl25:
@Li0k Please take a look again

dashboard/next-env.d.ts
```diff
@@ -1,6 +1,6 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
-import "./.next/dev/types/routes.d.ts";
```
tabVersion, line 3: unrelated change?

src/stream/src/executor/source/iceberg_list_executor.rs
```diff
@@ -288,13 +318,25 @@ async fn incremental_scan_stream(
         };
 
         if Some(current_snapshot.snapshot_id()) == last_snapshot {
+            // Ingestion is caught up — lag is 0.
+            metrics
+                .iceberg_source_snapshot_lag_seconds
+                .with_guarded_label_values(&label_values)
+                .set(0);
```
tabVersion, line 325: I do not see other place to set it to non-zero value in the diff.

src/stream/src/executor/source/iceberg_list_executor.rs
```diff
@@ -308,14 +350,83 @@ async fn incremental_scan_stream(
...
+            // Record delete files per data file.
+            metrics
+                .iceberg_source_delete_files_per_data_file
+                .with_guarded_label_values(&label_values)
+                .observe(scan_task.deletes.len() as f64);
+
+            yield scan_task;
+        }
+
+        let list_duration = list_start.elapsed();
```
tabVersion: the list time includes the yield time, we can over-estimate the time when in back-pressure.

src/stream/src/executor/source/batch_source/batch_iceberg_list.rs
```diff
@@ -129,6 +141,18 @@ impl<S: StateStore> BatchIcebergListExecutor<S> {
                         self.associated_table_id.to_string(),
                     ]);
 
+                    if !iceberg_table_name.is_empty() {
```
tabVersion: in which case will the table name be empty?

tabVersion:
btw, shall we add some context info for scan_task_to_chunk_with_deletes, to make source_id and source_name available when sending the metrics.

chenzl25:
> btw, shall we add some context info for scan_task_to_chunk_with_deletes, to make source_id and source_name available when sending the metrics.

I think not every caller would have source id and source name, so let's keep it simple is better.

PR #25093, opened 2026-03-17
feat(frontend): support consuming snapshot of batch refreshable job (frontend part)

wenym1:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## What's changed and what's your intention?

This PR adds frontend support for consuming snapshot backfill in batch refresh materialized views.

- Parse and validate `refresh.interval.sec` from materialized view `WITH` options and pass it through the frontend catalog writer and meta RPC request.
- Force batch refresh materialized views to use snapshot backfill and reject plans that cannot use snapshot backfill.
- Reject unsupported SQL operations around batch refresh materialized views, including `ALTER MATERIALIZED VIEW` and creating downstream streaming jobs on batch refresh MVs.
- Add parser, e2e SLT, and deterministic recovery simulation coverage for batch refresh snapshot creation, validation, periodic refresh behavior, and recovery.

This is the frontend part for batch refresh MV snapshot consumption, built on top of the meta-side support merged in #25251.

## Checklist

- [ ] I have written necessary rustdoc comments.
- [x]  I have added necessary unit tests and integration tests.
- [x]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  I have checked the [Release Timeline](https://github.com/risingwavelabs/rw-commits-history/blob/main/release_timeline.md) and [Currently Supported Versions](https://docs.risingwave.com/changelog/release-support-policy#support-end-dates-for-recent-releases) to determine which release branches I need to cherry-pick this PR into. 

## Documentation

- [x]  My PR needs documentation updates. 

<details>
<summary><b>Release note</b></summary>

Batch refresh materialized views can be created with `WITH (refresh.interval.sec = <seconds>)` to consume a snapshot-backfill upstream and refresh periodically.

Batch refresh materialized views currently require snapshot backfill, do not support `ALTER MATERIALIZED VIEW`, cannot depend directly on sources, and cannot be used as upstreams for streaming jobs.

</details>

PR #25123, opened 2026-03-19
refactor(meta): unify epoch node management of partial graphs

wenym1:

This PR refactors meta barrier handling to unify epoch-state management around `PartialGraphManager`, so barrier lifecycle data stays with the partial graph that owns the barrier instead of being split across checkpoint control, creating-job control, and completion code paths.

Concretely, this change introduces `PartialGraphBarrierInfo` as the per-barrier state container for:
- `BarrierInfo`
- `PostCollectCommand`
- notifiers
- table IDs to commit
- enqueue time / latency tracking

With that state moved into partial-graph-owned barrier queues, the completion path no longer reconstructs barrier context from separate structures. `CheckpointControl` and creating-job barrier control now pull collected barriers and their post-collect metadata directly from `PartialGraphManager`, and build `CompleteBarrierTask` from those partial-graph results.

The refactor also makes checkpoint vs non-checkpoint behavior more explicit:
- `PostCollectCommand::should_checkpoint()` is added to centralize whether a collected barrier needs commit-time handling.
- Non-checkpoint barriers can be consumed earlier in the partial-graph completion path.
- Checkpoint barriers continue through the normal commit/post-collect flow.

Metrics collection is also moved closer to the real barrier owner:
- database-level barrier latency / inflight-count metrics are implemented through a new `PartialGraphStat` interface
- creating-streaming-job barrier metrics use the same mechanism
- old duplicated barrier-number updates in checkpoint control are removed
- the unused snapshot backfill wait-commit metric is removed

Error cleanup is simplified as well. Instead of draining notifier state from checkpoint control queues, `PartialGraphManager` now owns and notifies pending barrier waiters on failure, which better matches the new ownership model.

Overall, the intention is to reduce duplicated epoch-state bookkeeping, make barrier completion ordering easier to reason about, and prepare the barrier subsystem for follow-up changes with a cleaner separation of responsibilities.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/meta/src/rpc/metrics.rs
+++ b/src/meta/src/rpc/metrics.rs
@@ -102,8 +102,6 @@ pub struct MetaMetrics {
     // ********************************** Snapshot Backfill ***************************
     /// The barrier latency in second of `table_id` and snapshto backfill `barrier_type`
     pub snapshot_backfill_barrier_latency: LabelGuardedHistogramVec, // (table_id, barrier_type)
-    /// The latency of commit epoch of `table_id`
-    pub snapshot_backfill_wait_commit_latency: LabelGuardedHistogramVec, // (table_id, )
     /// The lags between the upstream epoch and the downstream epoch.
     pub snapshot_backfill_lag: LabelGuardedIntGaugeVec, // (table_id, )
     /// The number of inflight barriers of `table_id`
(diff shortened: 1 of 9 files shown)


wenym1:
* **#25157** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25157?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#25123** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25123?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25123?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/meta/src/rpc/metrics.rs
```diff
@@ -334,13 +332,6 @@ impl MetaMetrics {
             registry
         )
         .unwrap();
-        let opts = histogram_opts!(
```
wenym1, line 337: This metrics is not needed any more, because it duplicates with the global wait commit timer

PR #25125, opened 2026-03-19
feat(common): add Bounded/Ratio parallelism config types

shanicky:
## Summary
- Extend `ConfigParallelism` with `Bounded(NonZeroU64)` and `Ratio(f32)` variants
- Add `ConfigBackfillParallelism` enum for backfill-specific parallelism config
- Add legacy migration helpers (`migrate_parallelism` / `migrate_adaptive_strategy`)
- Extend sqlparser to support `SET key = bounded(4)` function-style syntax via `SetVariableValueSingle::Raw`
- Exhaustive match adaptation in frontend `derive_parallelism` (Bounded/Ratio → adaptive fallback)

**PR 1/6** of the streaming parallelism stack. Pure additive — no behavior change.

## Test plan
- [x] `cargo test -p risingwave_common session_config::parallelism --lib`
- [x] `cargo test -p risingwave_sqlparser --test sqlparser_postgres parse_set -- --exact`
- [x] `cargo check -p risingwave_frontend --lib`

### DIFF
--- a/src/common/src/session_config/parallelism.rs
+++ b/src/common/src/session_config/parallelism.rs
@@ -346,6 +346,33 @@ mod tests {
     }
 
     #[test]
+    fn test_backfill_parallelism_parse_bounded() {
+        let parallelism: ConfigBackfillParallelism = "Bounded(4)".parse().unwrap();
+        assert_eq!(
+            parallelism,
+            ConfigBackfillParallelism::Bounded(NonZeroU64::new(4).unwrap())
+        );
+        assert_eq!(parallelism.to_string(), "bounded(4)");
+        assert_eq!(
+            parallelism.adaptive_strategy(),
+            Some(AdaptiveParallelismStrategy::Bounded(
+                NonZeroUsize::new(4).unwrap()
+            ))
+        );
+    }
+
+    #[test]
+    fn test_backfill_parallelism_parse_ratio() {
+        let parallelism: ConfigBackfillParallelism = "ratio(0.5)".parse().unwrap();
+        assert_eq!(parallelism, ConfigBackfillParallelism::Ratio(0.5));
+        assert_eq!(parallelism.to_string(), "ratio(0.5)");
+        assert_eq!(
+            parallelism.adaptive_strategy(),
+            Some(AdaptiveParallelismStrategy::Ratio(0.5))
+        );
+    }
+
+    #[test]
     fn test_strategy_parse_default() {
         assert_eq!(
             "default"
--- a/src/sqlparser/tests/sqlparser_postgres.rs
+++ b/src/sqlparser/tests/sqlparser_postgres.rs
@@ -429,6 +429,16 @@ fn parse_set() {
         }
     );
 
+    let stmt = verified_stmt("SET a = ratio(0.5)");
+    assert_eq!(
+        stmt,
+        Statement::SetVariable {
+            local: false,
+            variable: "a".into(),
+            value: SetVariableValueSingle::Raw("ratio(0.5)".into()).into(),
+        }
+    );
+
     let stmt = verified_stmt("SET a = DEFAULT");
     assert_eq!(
         stmt,


tabVersion:
Hi, thanks for the PR! I took a look at the changes and have a few observations — mostly around the gap between what's already parse-reachable and what the downstream code handles.

---

### 1. `derive_parallelism` silently drops Bounded/Ratio strategy info

In `src/frontend/src/stream_fragmenter/parallelism.rs`, `Bounded(_)` and `Ratio(_)` are mapped to `None` (same as `Adaptive`):

```rust
ConfigParallelism::Default
| ConfigParallelism::Adaptive
| ConfigParallelism::Bounded(_)
| ConfigParallelism::Ratio(_) => None,
```

However, the actual strategy is resolved separately via `derive_parallelism_strategy()`, which reads from `ConfigAdaptiveParallelismStrategy` — a completely independent session config field. It has no visibility into the new `ConfigParallelism::Bounded/Ratio` variants.

Since the sqlparser change + `FromStr` impl already allow users to run `SET streaming_parallelism = bounded(4)`, the full path is reachable today: the user thinks they've set a bounded strategy, but the strategy that actually gets applied is whatever the *separate* strategy config happens to be.

If this is intentionally deferred to a later PR in the 1/6 stack, it might be safer to either:
- Not yet wire up `Bounded()/Ratio()` parsing in `ConfigParallelism::FromStr` until the downstream is ready, or
- Gate the new variants behind a feature flag / return an explicit "unsupported" error for now.

### 2. Backfill parallelism path has the same gap

In `src/frontend/src/stream_fragmenter/mod.rs` (around line 230):

```rust
match config.streaming_parallelism_for_backfill() {
    ConfigParallelism::Default => None,
    override_parallelism => {
        derive_parallelism(Some(override_parallelism), streaming_parallelism)
            .or(normal_parallelism)
    }
}
```

If `streaming_parallelism_for_backfill()` returns `Bounded(n)` or `Ratio(r)`, `derive_parallelism` returns `None`, and `.or(normal_parallelism)` silently falls back to the normal parallelism. The backfill-specific bounded/ratio config is effectively ignored.

### 3. `ConfigBackfillParallelism` duplicates `ConfigParallelism` entirely

`ConfigBackfillParallelism` has identical variants, `FromStr`, `Display`, and `adaptive_strategy()` implementations as `ConfigParallelism`. This creates a maintenance risk where a future fix to one enum could be missed in the other.

Would it make sense to either use a type alias / newtype wrapper, or consolidate into a single enum with a marker generic (e.g., `ConfigParallelism<Backfill>` vs `ConfigParallelism<Streaming>`) if you need distinct types for session config registration?

---

These are non-blocking observations — happy to discuss if any of this is intentionally deferred to later PRs in the stack. Thanks!

shanicky:
> Hi, thanks for the PR! I took a look at the changes and have a few observations — mostly around the gap between what's already parse-reachable and what the downstream code handles.
> 
> ### 1. `derive_parallelism` silently drops Bounded/Ratio strategy info
> In `src/frontend/src/stream_fragmenter/parallelism.rs`, `Bounded(_)` and `Ratio(_)` are mapped to `None` (same as `Adaptive`):
> 
> ```rust
> ConfigParallelism::Default
> | ConfigParallelism::Adaptive
> | ConfigParallelism::Bounded(_)
> | ConfigParallelism::Ratio(_) => None,
> ```
> 
> However, the actual strategy is resolved separately via `derive_parallelism_strategy()`, which reads from `ConfigAdaptiveParallelismStrategy` — a completely independent session config field. It has no visibility into the new `ConfigParallelism::Bounded/Ratio` variants.
> 
> Since the sqlparser change + `FromStr` impl already allow users to run `SET streaming_parallelism = bounded(4)`, the full path is reachable today: the user thinks they've set a bounded strategy, but the strategy that actually gets applied is whatever the _separate_ strategy config happens to be.
> 
> If this is intentionally deferred to a later PR in the 1/6 stack, it might be safer to either:
> 
> * Not yet wire up `Bounded()/Ratio()` parsing in `ConfigParallelism::FromStr` until the downstream is ready, or
> * Gate the new variants behind a feature flag / return an explicit "unsupported" error for now.
> 
> ### 2. Backfill parallelism path has the same gap
> In `src/frontend/src/stream_fragmenter/mod.rs` (around line 230):
> 
> ```rust
> match config.streaming_parallelism_for_backfill() {
>     ConfigParallelism::Default => None,
>     override_parallelism => {
>         derive_parallelism(Some(override_parallelism), streaming_parallelism)
>             .or(normal_parallelism)
>     }
> }
> ```
> 
> If `streaming_parallelism_for_backfill()` returns `Bounded(n)` or `Ratio(r)`, `derive_parallelism` returns `None`, and `.or(normal_parallelism)` silently falls back to the normal parallelism. The backfill-specific bounded/ratio config is effectively ignored.
> 
> ### 3. `ConfigBackfillParallelism` duplicates `ConfigParallelism` entirely
> `ConfigBackfillParallelism` has identical variants, `FromStr`, `Display`, and `adaptive_strategy()` implementations as `ConfigParallelism`. This creates a maintenance risk where a future fix to one enum could be missed in the other.
> 
> Would it make sense to either use a type alias / newtype wrapper, or consolidate into a single enum with a marker generic (e.g., `ConfigParallelism<Backfill>` vs `ConfigParallelism<Streaming>`) if you need distinct types for session config registration?
> 
> These are non-blocking observations — happy to discuss if any of this is intentionally deferred to later PRs in the stack. Thanks!

Thanks for the careful review. These are good observations, and you are right that there is a gap between what is parse-reachable in this PR and what the downstream path fully interprets today.

For (1) and (2): this is intentionally deferred in the split PR stack. This PR is only the first step, focused on introducing the config/parser surface (`Bounded` / `Ratio` and `SET ... = bounded(...)` syntax) while keeping the current execution path behavior unchanged as much as possible. In the follow-up PRs, the frontend path is reworked to carry the resolved adaptive strategy alongside the fixed parallelism, and the backfill path is also split out explicitly instead of falling back through the normal path. So your observation is correct for this PR in isolation, but the gap is expected to be closed by the later PRs in the stack.

For (3): also a fair point. At the moment the duplication is intentional to keep backfill config separate from the main streaming config as we migrate the session config / frontend / meta layers incrementally. The follow-up PRs keep the two types distinct as well, so this is not something that is automatically addressed later in the stack. I agree there is some maintenance cost here, but for this split we preferred explicit separation over introducing a more abstract generic/newtype layer at the same time.

I’ll make sure this is called out more clearly so the temporary gap in this first PR is explicit, rather than looking like an accidental partial wiring.

PR #25131, opened 2026-03-20
fix(sqlserver): fix sqlserver data type mismatch issue

wcy-fdu:

The fix applies safe upcasting (for example, Int32 -> Int64) during SQL Server snapshot row parsing so frontend-accepted type compatibility is honored at runtime, preventing backfill panics from type mismatches.

- Closes #[ISSUE_NUMBER]

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/src/parser/sql_server.rs
+++ b/src/connector/src/parser/sql_server.rs
@@ -53,7 +53,9 @@ pub fn sql_server_row_to_owned_row(row: &mut Row, schema: &Schema) -> OwnedRow {
                 }
             },
             false => match row.try_get::<ScalarImplTiberiusWrapper, usize>(i) {
-                Ok(datum) => datum.map(|d| d.0),
+                Ok(datum) => datum
+                    .map(|d| d.0)
+                    .map(|scalar| coerce_scalar_to_target_type(scalar, &rw_field.data_type)),
                 Err(err) => {
                     log_error!(name, err, "parse column failed");
                     None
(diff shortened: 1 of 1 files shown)


wenym1:
@codex[agent] review the PR. Summarize the previous problem and the current fix.

src/connector/src/parser/sql_server.rs
```diff
@@ -66,6 +68,19 @@ pub fn sql_server_row_to_owned_row(row: &mut Row, schema: &Schema) -> OwnedRow {
...
+fn coerce_scalar_to_target_type(scalar: ScalarImpl, target_type: &DataType) -> ScalarImpl {
+    match (scalar, target_type) {
+        // SQL Server validator allows integer upcast (e.g. `int` -> `BIGINT`).
+        // Coerce snapshot values to the target RW type to keep validation and execution consistent.
+        (ScalarImpl::Int16(v), DataType::Int32) => ScalarImpl::Int32(v as i32),
+        (ScalarImpl::Int16(v), DataType::Int64) => ScalarImpl::Int64(v as i64),
+        (ScalarImpl::Int32(v), DataType::Int64) => ScalarImpl::Int64(v as i64),
+        // SQL Server `real` may map to `FLOAT` in RW validator.
+        (ScalarImpl::Float32(v), DataType::Float64) => ScalarImpl::Float64((v.0 as f64).into()),
+        (scalar, _) => scalar,
```
wenym1, line 80: shall we check and return error if the two types are different to avoid similar panic in the future?

wenym1 (approved):
Rest LGTM

src/connector/src/parser/sql_server.rs
wcy-fdu: Leave in future pr.

PR #25133, opened 2026-03-20
feat(meta): support wait jobs

chenzl25:

- This PR enhances the `WAIT` SQL statement by allowing users to wait for the completion of specific background jobs, such as tables, materialized views, sinks, or indexes.
- `wait materialized view <object>`,`wait table <object>`, `wait sink <object>`, `wait index <object>`
-  change the pull interval from 1ms to 100ms to reduce overhead.
- change timeout from 30min to 2h.
- This feature could be used in dbt-risingwave background ddl job in the future.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/rpc_client/src/meta_client.rs
+++ b/src/rpc_client/src/meta_client.rs
@@ -1147,9 +1147,7 @@ impl MetaClient {
     }
 
     pub async fn wait(&self, job_id: Option<JobId>) -> Result<WaitVersion> {
-        let request = WaitRequest {
-            job_id: job_id.map(|job_id| job_id.as_raw_id()),
-        };
+        let request = WaitRequest { job_id };
         let resp = self.inner.wait(request).await?;
         Ok(resp
             .version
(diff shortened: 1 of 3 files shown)


wenym1:
For a specific job, we can actually call `wait_streaming_job_finished`, which is more light weight, and will be returned immediately when the job is created. This is the same method for foreground ddl to wait for job finish.

src/rpc_client/src/meta_client.rs
```diff
@@ -1146,8 +1146,10 @@ impl MetaClient {
         Ok(resp.hummock_version_id)
     }
 
-    pub async fn wait(&self) -> Result<WaitVersion> {
-        let request = WaitRequest {};
+    pub async fn wait(&self, job_id: Option<JobId>) -> Result<WaitVersion> {
+        let request = WaitRequest {
+            job_id: job_id.map(|job_id| job_id.as_raw_id()),
```
wenym1: can specify the `job_id` as strong typed `JobId` in the pb generated struct, in `build.rs`

PR #25005, opened 2026-03-07
test: add slt test for tracking alter sink config feature

Standing-Man:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## What's changed and what's your intention?

- Closes #22479.

This PR adds tests for the sink-related properties mentioned in #20691.
Specifically, it verifies the ability to alter the following properties:
- For ClickHouse, StarRocks, DeltaLake, and Iceberg:
  - `commit_checkpoint_interval`

- For Kafka:
  - `properties.allow.auto.create.topics`
  - `properties.batch.num.messages`
  - `properties.batch.size`
  - `properties.enable.idempotence`
  - `properties.max.in.flight.requests.per.connection`
  - `properties.message.max.bytes`
  - `properties.message.send.max.retries`
  - `properties.message.timeout.ms`
  - `properties.queue.buffering.max.kbytes`
  - `properties.queue.buffering.max.messages`
  - `properties.queue.buffering.max.ms`
  - `properties.request.required.acks`
  - `properties.retry.backoff.ms`
  - `properties.receive.message.max.bytes`

These tests ensure that the above properties can be altered correctly and behave as expected.

## Checklist

- [ ] I have written necessary rustdoc comments.
- [x]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  I have checked the [Release Timeline](https://github.com/risingwavelabs/rw-commits-history/blob/main/release_timeline.md) and [Currently Supported Versions](https://docs.risingwave.com/changelog/release-support-policy#support-end-dates-for-recent-releases) to determine which release branches I need to cherry-pick this PR into. 

## Documentation

- [ ]  My PR needs documentation updates. 

<details>
<summary><b>Release note</b></summary>

add e2e tests to cover ALTER SINK functionality

</details>

PR #25148, opened 2026-03-22
feat(sink): support altering secrets for sinks

### LINKED ISSUE #22181: refactor: handle alter sink connector props with secret

Standing-Man:

- Closes #22181

This PR adds complete secret-aware support for altering sink connector properties, including direct `ALTER SINK`, Iceberg table connector updates, and `ALTER CONNECTION` propagation to dependent sinks.
Previously, the sink alter path mainly handled plaintext properties. Secret-backed properties could become inconsistent across the catalog, fragment metadata, runtime connector configuration, SQL definitions, and object dependency records. As a result, secret rotation could leave stale secret references, prevent old secrets from being dropped, or fail to deliver the updated secret to running sinks.
The PR makes the following changes:
- Uses `WithOptionsSecResolved::handle_update` to merge plaintext updates and Secret references while correctly handling transitions between plaintext and Secret-backed values.
- Persists plaintext properties and `secret_ref` separately, ensuring secret values are never stored in catalog definitions or connector properties as plaintext.
- Updates `sink_desc.properties` and `sink_desc.secret_refs` in stream fragments so runtime metadata remains consistent with the catalog.
- Resolves Secret references through `LocalSecretManager::fill_secrets` when validating connector configurations and broadcasting runtime property changes.
- Maintains `object_dependency` records when rotating secrets, allowing an unused old Secret to be dropped while continuing to protect the currently referenced Secret.
- Aligns the dependent-sink path in `ALTER CONNECTION` with direct `ALTER SINK` and dependent-source behavior.
- Separates alter-on-the-fly validation from full connector configuration validation: only changed fields are checked against the alterable-field allowlist, while the complete resolved configuration is used for semantic validation.
- Prevents the connector type from being changed through an ALTER operation.
- Rewrites stored SQL definitions using typed `SqlOptionValue::SecretRef` nodes instead of formatting `"SECRET ..."` as a normal string. Secret metadata is loaded by ID together with its schema and reference mode, preserving schema-qualified names and AS FILE semantics without exposing the secret value. 
- Extends Iceberg table alter handling so its implicit sink, source, table catalog entries, fragment metadata, Secret references, and dependencies remain synchronized.
- Adds end-to-end coverage for direct sink secret rotation, connection-level propagation to dependent sinks, catalog state, allowed and disallowed on-the-fly fields, and `DROP SECRET` dependency behavior.
The intention is to make Secret-backed connector properties a fully supported part of ALTER operations rather than a special case handled only during object creation. After this change, rotating a connector credential from `SECRET s1` to `SECRET s2` consistently updates catalog metadata, SQL definitions, stream fragments, runtime connector state, and dependency tracking.

## Checklist

## Documentation

<details>

Now we supports updating secret-backed connector properties through `ALTER SINK ... CONNECTOR WITH` and `ALTER TABLE ... CONNECTOR WITH` for Iceberg tables.

Secret rotation is also propagated correctly when `ALTER CONNECTION ... CONNECTOR WITH` updates a connection used by dependent sinks. We updates the catalog, running fragments, and secret dependencies consistently, allowing obsolete secrets to be dropped while preventing active secrets from being removed.

Connector configurations are validated using resolved secret values, while stored SQL definitions retain typed, schema-qualified `SECRET` references, including `AS FILE`, without exposing secret contents in plaintext.

</details>

## Summary by CodeRabbit

- **New Features**
  - Connector credentials and authentication settings can now be updated at runtime across supported sinks, sources, and Iceberg connections.
  - Added support for rotating secret-backed configuration without recreating connectors or sinks.
  - Iceberg table properties now support secret references during updates.

- **Bug Fixes**
  - Secret changes are correctly validated, resolved, persisted, and propagated to running jobs.
  - Unsupported sink properties continue to be rejected, while supported password and credential changes can switch between secrets and plaintext.

### DIFF
--- a/src/meta/src/controller/streaming_job.rs
+++ b/src/meta/src/controller/streaming_job.rs
@@ -3146,9 +3146,16 @@ impl CatalogController {
                         .unwrap_or_default(),
                 );
 
-                let (_sink_to_add_secret_dep, _sink_to_remove_secret_dep) =
-                    sink_options_with_secret
-                        .handle_update(alter_props.clone(), alter_secret_refs.clone())?;
+                let (sink_to_add_secret_dep, sink_to_remove_secret_dep) = sink_options_with_secret
+                    .handle_update(alter_props.clone(), alter_secret_refs.clone())?;
+
+                update_secret_dependencies(
+                    &txn,
+                    sink_to_add_secret_dep,
+                    sink_to_remove_secret_dep,
+                    sink_id.as_object_id(),
+                )
+                .await?;
 
                 // Prepare sink update
                 let active_sink = sink::ActiveModel {
--- a/src/meta/src/controller/streaming_job.rs
+++ b/src/meta/src/controller/streaming_job.rs
@@ -2402,11 +2402,8 @@ impl CatalogController {
 
         // Only check alter-on-fly restrictions for SQL ALTER SOURCE, not for admin risectl operations
         if !skip_alter_on_fly_check {
-            let prop_keys: Vec<String> = alter_props
-                .keys()
-                .chain(alter_secret_refs.keys())
-                .cloned()
-                .collect();
+            let prop_keys: Vec<String> =
+                build_altered_field_names(&alter_props, &alter_secret_refs);
             risingwave_connector::allow_alter_on_fly_fields::check_source_allow_alter_on_fly_fields(
                 &connector, &prop_keys,
             )?;
@@ -2576,8 +2573,19 @@ impl CatalogController {
             .await?
             .ok_or_else(|| MetaError::catalog_id_not_found(ObjectType::Sink.as_str(), sink_id))?;
 
-        // check if the alter-ed props are valid for all connector
-        validate_sink_props(&sink, &alter_props, &alter_secret_refs)?;
+        let connector_type = sink
+            .properties
+            .inner_ref()
+            .get(CONNECTOR_TYPE_KEY)
+            .ok_or_else(|| SinkError::Config(anyhow!("connector not specified when alter sink")))?
+            .to_lowercase();
+
+        validate_connector_type(&connector_type, &alter_props)?;
+
+        let prop_keys = build_altered_field_names(&alter_props, &alter_secret_refs);
+
+        check_sink_allow_alter_on_fly_fields(&connector_type, &prop_keys)
+            .map_err(|e| SinkError::Config(anyhow!(e)))?;
 
         let mut options_with_secret = WithOptionsSecResolved::new(
             sink.properties.0.clone(),
@@ -2585,8 +2593,12 @@ impl CatalogController {
                 .map(|secret_ref| secret_ref.to_protobuf())
                 .unwrap_or_default(),
         );
+
         let (to_add_secret_dep, to_remove_secret_dep) =
             options_with_secret.handle_update(alter_props.clone(), alter_secret_refs)?;
+
+        // check if the alter-ed props are valid for all connector
+        validate_sink_props(&connector_type, options_with_secret.clone())?;
 
         tracing::info!(
             "applying new properties to sink: sink_id={}, options_with_secret={:?}",
@@ -2690,7 +2702,20 @@ impl CatalogController {
             .one(&txn)
             .await?
             .ok_or_else(|| MetaError::catalog_id_not_found(ObjectType::Sink.as_str(), sink_id))?;
-        validate_sink_props(&sink, &alter_props, &alter_secret_refs)?;
+
+        let connector_type = sink
+            .properties
+            .inner_ref()
+            .get(CONNECTOR_TYPE_KEY)
+            .ok_or_else(|| SinkError::Config(anyhow!("connector not specified when alter sink")))?
+            .to_lowercase();
+
+        validate_connector_type(&connector_type, &alter_props)?;
+
+        let prop_keys = build_altered_field_names(&alter_props, &alter_secret_refs);
+
+        check_sink_allow_alter_on_fly_fields(&connector_type, &prop_keys)
+            .map_err(|e| SinkError::Config(anyhow!(e)))?;
 
         let mut options_with_secret = WithOptionsSecResolved::new(
             sink.properties.0.clone(),
@@ -2698,8 +2723,12 @@ impl CatalogController {
                 .map(|secret_ref| secret_ref.to_protobuf())
                 .unwrap_or_default(),
         );
+
         let (to_add_secret_dep, to_remove_secret_dep) =
             options_with_secret.handle_update(alter_props.clone(), alter_secret_refs)?;
+
+        // check if the alter-ed props are valid for all connector
+        validate_sink_props(&connector_type, options_with_secret.clone())?;
 
         // rewrite the sql for definition field
         let rewrite_sql = {
@@ -3135,8 +3164,14 @@ impl CatalogController {
             for (sink, _obj) in sinks_with_objs {
                 let sink_id = sink.sink_id;
 
-                // Validate that sink props can be altered
-                validate_sink_props(&sink, &alter_props.clone(), &alter_secret_refs.clone())?;
+                let connector_type = sink
+                    .properties
+                    .inner_ref()
+                    .get(CONNECTOR_TYPE_KEY)
+                    .ok_or_else(|| {
+                        SinkError::Config(anyhow!("connector not specified when alter sink"))
+                    })?
+                    .to_lowercase();
 
                 let mut sink_options_with_secret = WithOptionsSecResolved::new(
                     sink.properties.0.clone(),
@@ -3148,6 +3183,9 @@ impl CatalogController {
 
                 let (sink_to_add_secret_dep, sink_to_remove_secret_dep) = sink_options_with_secret
                     .handle_update(alter_props.clone(), alter_secret_refs.clone())?;
+
+                // check if the alter-ed props are valid for all connector
+                validate_sink_props(&connector_type, sink_options_with_secret.clone())?;
 
                 update_secret_dependencies(
                     &txn,
@@ -3417,18 +3455,39 @@ impl CatalogController {
 }
 
 fn validate_sink_props(
-    sink: &sink::Model,
+    connector_type: &str,
+    with_properties: WithOptionsSecResolved,
+) -> MetaResult<()> {
+    let (options, secret_refs) = with_properties.into_parts();
+    let resolved_props = LocalSecretManager::global()
+        .fill_secrets(options, secret_refs)
+        .map_err(MetaError::from)?;
+
+    match_sink_name_str!(
+        connector_type,
+        SinkType,
+        SinkType::validate_alter_config(&resolved_props),
+        |sink: &str| Err(SinkError::Config(anyhow!("unsupported sink type {}", sink)))
+    )?;
+
+    Ok(())
+}
+
+fn build_altered_field_names(
     alter_props: &BTreeMap<String, String>,
     alter_secret_refs: &BTreeMap<String, PbSecretRef>,
+) -> Vec<String> {
+    alter_props
+        .keys()
+        .chain(alter_secret_refs.keys())
+        .cloned()
+        .collect()
+}
+
+fn validate_connector_type(
+    connector: &str,
+    alter_props: &BTreeMap<String, String>,
 ) -> MetaResult<()> {
-    let connector = sink
-        .properties
-        .inner_ref()
-        .get(CONNECTOR_TYPE_KEY)
-        .ok_or_else(|| SinkError::Config(anyhow!("connector not specified when alter sink")))?;
-
-    let connector_type = connector.to_lowercase();
-
     if let Some(new_connector) = alter_props.get(CONNECTOR_TYPE_KEY)
         && new_connector != connector
     {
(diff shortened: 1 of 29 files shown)


Standing-Man:
Hi @tabVersion and @xxhZs, This PR is ready for review, but it currently lacks SLT tests for validation.
I'm not sure whether the SLT tests for altering secrets should be handled in a separate PR (#25005) after it is merged, or if we should include them here using a unified SLT file to cover ALTER SINK secret functionality. Any suggestions would be appreciated.

tabVersion (commented):
## Code Review for PR #25148

Overall this PR is a solid improvement — it correctly introduces `WithOptionsSecResolved` to track secret refs alongside plaintext props, properly manages secret dependency tracking, and fixes a pre-existing double-quoting bug. However, there are a few issues worth addressing:

---

### 🔴 Bug: Iceberg source `properties` and `secret_ref` not updated

**File:** `src/meta/src/controller/streaming_job.rs` — `update_iceberg_table_props_by_table_id`

The iceberg table ALTER path updates three entities: **sink**, **source**, and **table**. The sink gets `properties`, `secret_ref`, and `definition` updated. But the source only gets `definition` updated:

```rust
let active_source = source::ActiveModel {
    source_id: Set(source_id),
    definition: Set(rewrite_sql.clone()),
    ..Default::default()  // ← missing with_properties and secret_ref
};
```

The `source::Model` has both `with_properties: Property` and `secret_ref: Option<SecretRef>` fields (see `src/meta/model/src/source.rs:34,42`). After ALTER, the source's runtime properties and secret refs will be stale — they'll still reference the old config while only the SQL definition text is updated. This could cause the source to use old credentials after an ALTER, or worse, cause inconsistency between what the catalog shows vs what the source actually uses.

**Suggested fix:** Update the source's `with_properties` and `secret_ref` similarly to how the sink is updated:
```rust
let active_source = source::ActiveModel {
    source_id: Set(source_id),
    with_properties: Set(risingwave_meta_model::Property(
        options_with_secret.as_plaintext().clone(),
    )),
    secret_ref: Set((!options_with_secret.as_secret().is_empty())
        .then(|| SecretRef::from(options_with_secret.as_secret().clone()))),
    definition: Set(rewrite_sql.clone()),
    ..Default::default()
};
```

---

### 🟡 Code duplication: Secret dependency management logic

The secret dependency add/remove block is copy-pasted verbatim between `update_sink_props_by_sink_id` and `update_iceberg_table_props_by_table_id`. Consider extracting it into a helper like:

```rust
async fn update_secret_dependencies(
    txn: &DatabaseTransaction,
    to_add: Vec<SecretId>,
    to_remove: Vec<SecretId>,
    used_by: ObjectId,
) -> MetaResult<()> { ... }
```

This would reduce the ~30 lines of duplication and make future maintenance easier.

---

### 🟡 `validate_sink_props` doesn't validate secret refs

`validate_sink_props` only validates the plaintext `alter_props` but not the `alter_secret_refs`. This means:
- A user could pass a secret ref for a key that isn't allowed to be altered on-the-fly (bypassing `check_sink_allow_alter_on_fly_fields`)
- The merged config validation (`SinkType::validate_alter_config`) only sees plaintext props, not the full picture including secrets

Consider merging `alter_secret_refs` keys into the field validation, or at minimum checking that secret ref keys are also in the allowed alter fields.

---

### 🟢 Good fix: Double-quoting bug in `format_with_option_secret_resolved`

The old code used `format!("'{}'", v)` before passing to `SqlOption::try_from`, but `try_from` already wraps the value in single quotes internally (see `src/sqlparser/src/ast/mod.rs:3296-3297`). This caused double-quoting in the stored SQL definition.

The new code correctly uses `v.to_owned()`. Note this also silently fixes the **source** alter path that shares this function — worth mentioning in the PR description.

---

### 🟡 Minor: `panic!` on unexpected statement type

Both methods use `panic!("definition is not a create sink/iceberg table statement")` when the parsed SQL doesn't match expectations. In a production meta service, this would crash the node. Consider returning a `MetaError` instead:

```rust
return Err(MetaError::invalid_parameter("definition is not a create sink statement"));
```

(Note: the old code also panicked, so this is pre-existing, but worth fixing while touching this code.)

---

### 🔵 Nit: Unnecessary block scope

The secret dependency management code in both methods is wrapped in a bare `{ ... }` block that serves no purpose (no variables need scoping). It can be removed for clarity.

---

### 🔵 Nit: `MetadataManager::update_sink_props_by_sink_id` removed but inconsistently

The PR removes the `MetadataManager::update_sink_props_by_sink_id` wrapper (which was just a passthrough to `catalog_controller`) and calls `catalog_controller` directly in `stream_service.rs`. Good cleanup! But `update_iceberg_table_props_by_table_id` still goes through the `MetadataManager` wrapper. Consider being consistent — either keep both wrappers or remove both.

---

**Summary:** The main blocker is the missing source `properties`/`secret_ref` update in the iceberg table path. The rest are improvements that would be nice to have.

Standing-Man:
Hi @tabVersion and @xxhZs, please take a look when you have time.

tabVersion:
Thanks for adding sink secret alter support! The direct `ALTER SINK ... WITH (... = SECRET ...)` path looks nicely covered by the new JDBC SLT, including dependency cleanup when switching back to plaintext.

One thing I think is worth checking before merge: the connection-level propagation path still seems asymmetric for dependent sinks.

When `ALTER CONNECTION ... WITH (password = SECRET new_secret)` is applied, dependent sources are updated through `WithOptionsSecResolved`, their `secret_ref`s are persisted, fragments get `source_inner.secret_refs`, and the runtime broadcast is built with `LocalSecretManager::fill_secrets(...)`. For dependent sinks, though, the current path still appears to only extend `sink.properties` with plaintext `alter_props` and then broadcast that plaintext map. It does not appear to apply `alter_secret_refs` to the sink catalog, update `sink.secret_ref`, update `sink_desc.secret_refs` in fragments, or build `updated_sinks_with_props` via `fill_secrets(...)`.

Could we mirror the direct sink alter logic for dependent sinks as well? Concretely, I think the dependent-sink branch in `update_connection_and_dependent_objects_props` should probably:

- construct a `WithOptionsSecResolved` from `sink.properties` and `sink.secret_ref`;
- call `handle_update(alter_props.clone(), alter_secret_refs.clone())`;
- persist both `properties` and `secret_ref`;
- update both `sink_desc.properties` and `sink_desc.secret_refs` in fragments;
- build the `ConnectorPropsChange` payload using `LocalSecretManager::fill_secrets(plaintext_props, secret_refs)`.

Otherwise, rotating a secret through `ALTER CONNECTION` may update the connection and dependent sources, but dependent sinks can keep stale/missing secret refs in catalog metadata and may not receive the rotated plaintext secret at runtime.

A focused regression test could be: create a connection with `password = SECRET s1`, create a sink using that connection, run `ALTER CONNECTION ... password = SECRET s2`, then verify `rw_sinks.connector_props` points to `s2`, `DROP SECRET s1` succeeds, and `DROP SECRET s2` is still blocked while the sink exists.

Standing-Man:
> One thing I think is worth checking before merge: the connection-level propagation path still seems asymmetric for dependent sinks.

Thanks for the detailed review. I have resolved the issue where changes to connection-level properties were not propagated to dependent sinks, ensuring consistency with the direct ALTER SINK path. Additionally, I have added corresponding test cases.

The CI is still encountering flaky test failures, so [that issue](https://github.com/risingwavelabs/risingwave/issues/25243) may not be fully resolved yet.

Standing-Man:
Hi @tabVersion, just a friendly ping — would you mind taking a look at this PR when convenient?

src/meta/src/controller/streaming_job.rs
Standing-Man: Hi @tabVersion, Sink needs to update its dependencies accordingly, and the same applies to Source. I will open a separate PR to address this.

tabVersion:
Thanks for the follow-up fixes here — I checked the current head and the previous `ALTER CONNECTION` dependent-sink propagation gap looks much better now: the sink catalog `properties`/`secret_ref`, sink fragment `secret_refs`, secret dependencies, and the runtime `fill_secrets(...)` broadcast path are all handled, and the Kafka SLT covers the old/new secret dependency behavior.

I found one remaining edge case that I think is worth fixing before merge: `validate_sink_props` still validates the merged sink config using only `sink.properties` plus plaintext `alter_props`, but not the existing/new `secret_ref` keys as actual config values.

That matters once a required connector option is stored as a secret. For example, this PR now marks required auth fields like `clickhouse.password`, `doris.password`, `starrocks.password`, and `sqlserver.password` as alterable. After a sink is created with one of those fields as `SECRET ...` (or after the first alter moves it from plaintext to secret), that key is removed from `sink.properties` and lives only in `sink.secret_ref`. A later valid alter — e.g. rotating the same password secret again, or changing `commit_checkpoint_interval` on a ClickHouse/StarRocks sink — will call `SinkType::validate_alter_config(&new_props)` where `new_props` is missing the required password field, so deserialization/validation can reject a valid sink state before `WithOptionsSecResolved::handle_update(...)` gets a chance to materialize the full option set.

Could we validate against the same merged representation that will be persisted/broadcast? Concretely, one option is to build/apply `WithOptionsSecResolved` first, then validate with a secret-filled map (or at least insert placeholder values for secret-backed keys before `SinkType::validate_alter_config`). A focused regression test with a connector whose password field is required, such as ClickHouse or StarRocks, would catch this:

1. create the sink with `...password = SECRET s1`;
2. `ALTER SINK ... WITH (...password = SECRET s2)`;
3. optionally alter an unrelated allowed property after that, such as `commit_checkpoint_interval`;
4. verify the catalog still points to the active secret and old secrets can be dropped.

The existing JDBC/Kafka coverage is helpful, but I don't think it exercises this required-secret-field path.

tabVersion:
@xxhZs plz take a look

src/meta/src/controller/streaming_job.rs
Standing-Man: Resolved in #25582.

src/meta/src/controller/streaming_job.rs
```diff
@@ -3317,61 +3383,56 @@ impl CatalogController {
...
+    alter_secret_refs: &BTreeMap<String, PbSecretRef>,
+) -> Vec<String> {
+    alter_props
+        .keys()
+        .chain(alter_secret_refs.keys())
+        .cloned()
+        .collect()
+}
+
+fn validate_connector_type(
```
xxhZs: It seems we don’t need to validate this, since it’s not in the allow_alter_on_fly  to begin with.

src/meta/src/controller/streaming_job.rs
```diff
@@ -2558,30 +2501,89 @@ impl CatalogController {
...
+                            "Sink definition is not a CREATE SINK statement"
+                        )),
+                    )));
+                }
+            }
+
+            stmt.to_string()
+        };
+
+        update_secret_dependencies(
```
xxhZs, line 3136: Maybe we need to use this method with the source ID?

xxhZs (commented):
If Iceberg tables are supported, the code at this link also needs to be modified:
https://github.com/risingwavelabs/risingwave/blob/05cb6806e4380f46657aefdbb2d46aed9ddd8cb1/src/frontend/src/handler/alter_table_props.rs#L97

src/meta/src/controller/streaming_job.rs
Standing-Man: `ALTER SINK`, `ALTER SOURCE`, and dependent source/sink updates in `ALTER CONNECTION` now consistently use `update_secret_dependencies` to handle modified secret parameters.

src/meta/src/controller/streaming_job.rs
Standing-Man: Got it. 👍🏻

Standing-Man:
Hi @tabVersion, @xxhZs and @hzxa21, I’ve pushed a follow-up fix in bd5f86b9 to address the remaining issue with secret references in rewritten connector options.

The new commit:
- constructs a typed `SqlOptionValue::SecretRef` instead of serializing `SECRET ...` as a quoted string;
- preserves schema-qualified secret names and `TEXT`/`AS FILE` semantics;
- adds sqlparser unit tests for typed secret references;
- adds focused Kafka SLT coverage verifying that `rw_sinks.definition` retains the correct `SECRET public.<name>` syntax after `ALTER SINK`.

Together with the earlier fixes for dependent-sink propagation, secret dependency updates, resolved-config validation, and Iceberg table handling, the known review concerns should now be addressed.

Could you please take another look at the latest revision when convenient? Thanks!

src/meta/src/controller/streaming_job.rs
```diff
@@ -3200,54 +3201,117 @@ impl CatalogController {
...
-            definition: Set(definition),
+            definition: Set(rewrite_sql),
             ..Default::default()
         };
         Sink::update(active_sink).exec(&txn).await?;
         Source::update(active_source).exec(&txn).await?;
         Table::update(active_table).exec(&txn).await?;
 
-        update_sink_fragment_props(&txn, sink_id, new_config).await?;
+        update_sink_fragment_props(&txn, sink_id, &options_with_secret).await?;
```
hzxa21, line 3314: @chenzl25 could you confirm whether `ALTER ICEBERG TABLE ... CONNECTOR WITH` also needs to update the implicit source fragments and broadcast a `ConnectorPropsChange` entry keyed by `source_id`?

This path updates the implicit source catalog row and source/sink secret dependencies above, but the only fragment update here is for the sink. In `stream_service.rs`, the Iceberg-table branch returns `sink_id.as_object_id()`, so the runtime mutation is also keyed only by the sink id. Source executors rebuild only when the mutation map contains their own `source_id`, so the running/recovered Iceberg source reader can keep the old connector properties/secret refs after secret rotation.

Standing-Man:
The work has been continued and completed in PR #27222. PR #25148 will be closed.

PR #25150, opened 2026-03-23
fix(meta/frontend): ensure streaming mapping visibility before DDL completion

shanicky:

DDL operations could return success before the frontend had a complete view of the corresponding streaming fragment mappings, causing \"streaming vnode mapping not found\" errors on immediate queries. In the worst case, waiting longer did not help because the mapping was not merely delayed — it could be dropped during initialization or recovery.

**Root causes fixed:**

1. **Non-atomic snapshot**: `get_worker_slot_mapping_snapshot` previously read mappings and the notification version in separate steps. A mapping added between those reads could be covered by the returned version but missing from the snapshot payload, causing the frontend to permanently drop the corresponding notification during initialization.

2. **`wait_version` didn't gate on mapping visibility**: DDL completion only waited for catalog version and hummock version. The frontend could report DDL success before streaming mappings became visible locally.

3. **Multi-fragment mapping batches advanced the version too early**: a whole batch of fragment mappings shared one version, but frontend notifications were still sent one fragment at a time. The frontend could observe the target version on the first notification in the batch and unblock `wait_version` before the rest of the batch had been applied.

4. **Observer recovery could be marked complete before mappings were authoritative**: during meta recovery, a snapshot can legitimately contain streaming jobs while still lacking complete fragment mappings. The old `CompletedObserverRecovery` logic treated replay completion as sufficient and could unblock waits before mapping state was actually complete.

**Changes:**

- Added an independent `streaming_worker_slot_mapping_version` counter in `SharedActorInfosInner`, with atomic snapshot reads for mapping state and version.
- Extended `WaitVersion` and frontend `wait_version` to also wait for streaming mapping visibility before returning DDL success.
- Kept the batch version on only the final notification in a fragment-mapping batch; intermediate notifications now carry the ignored version so `wait_version` does not unblock early.
- Preserved ignored-version intermediate mapping notifications during initialization replay so newer buffered mappings are not dropped.
- Added `streaming_worker_slot_mapping_ready` to `MetaSnapshot::SnapshotVersion` and deferred `CompletedObserverRecovery` until mapping state is authoritative.
- Sent an explicit recovery-complete version-only mapping notification from meta once recovery finishes, so a frontend that subscribed mid-recovery can safely unblock deferred waits.
- Kept the frontend observer's reset/advance split and stale-notification filtering so meta restart still behaves correctly.

**Tests:**

- Added unit tests for initialization replay keeping ignored-version intermediate notifications.
- Added tests that only the final notification in a mapping batch advances the visible version.
- Added frontend tests covering deferred recovery completion, zero-mapping recovery, and `wait_version` behavior across recovery and mapping waits.
- Retained the previous tests for initialization ordering, stale-notification filtering, and `wait_version` gating.

## Checklist

### DIFF
--- a/src/meta/src/controller/fragment.rs
+++ b/src/meta/src/controller/fragment.rs
@@ -78,7 +78,8 @@ use crate::controller::utils::{
 };
 use crate::error::MetaError;
 use crate::manager::{
-    ActiveStreamingWorkerNodes, LocalNotification, NotificationManager, NotificationVersion,
+    ActiveStreamingWorkerNodes, IGNORED_NOTIFICATION_VERSION, LocalNotification,
+    NotificationManager, NotificationVersion,
 };
 use crate::model::{
     DownstreamFragmentRelation, Fragment, FragmentActorDispatchers, FragmentDownstreamRelation,
@@ -189,11 +190,17 @@ impl NotificationManager {
             return;
         }
         // notify all fragment mappings to frontend.
-        for fragment_mapping in fragment_mappings {
+        let final_notification_index = fragment_mappings.len() - 1;
+        for (index, fragment_mapping) in fragment_mappings.into_iter().enumerate() {
+            let notification_version = if index == final_notification_index {
+                version
+            } else {
+                IGNORED_NOTIFICATION_VERSION
+            };
             self.notify_frontend_with_version(
                 operation,
                 NotificationInfo::StreamingWorkerSlotMapping(fragment_mapping),
-                version,
+                notification_version,
             );
         }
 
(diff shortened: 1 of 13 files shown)


src/meta/src/controller/fragment.rs
```diff
@@ -178,6 +180,7 @@ impl NotificationManager {
         &self,
         operation: NotificationOperation,
         fragment_mappings: Vec<PbFragmentWorkerSlotMapping>,
+        version: NotificationVersion,
```
zwang28, line 183: Since all calls to notify_fragment_mapping currently pass IGNORED_NOTIFICATION_VERSION, the changes to notify_fragment_mapping have no functional impact.

PR #25152, opened 2026-03-23
fix(source): drop stale wait checkpoint task

### LINKED ISSUE #25071: bug(cdc/postgres): Debezium periodic offset flush advances Postgres replication slot `restart_lsn` without a RisingWave checkpoint, causing unrecoverable WAL gap on pod restart
## Describe the bug

We observed this in a production deployment on 2026-03-11 after a rolling update of RisingWave compute pods.

**What was observed:**

1. A rolling update replaced the compute pods. A Postgres CDC source (`_erp_source`) was mid-way through its initial snapshot/backfill scan at the time — no WAL streaming had yet been durably checkpointed for this source.
2. Immediately before and during the rolling update, the Debezium connector was periodically logging `Flush offsets successfully` for this source, indicating that Postgres's replication slot `restart_lsn` was being advanced to the current WAL position.
3. After the pod restarted, Debezium attempted to resume from the last offset stored in RisingWave's checkpoint state (`lsn=LSN{3/E33CCAD8}`, timestamp `2026-03-11T14:02:33Z`).
4. The replication slot's `restart_lsn` had advanced to `LSN{3/E3481010}` — ahead of the stored offset. The WAL between the two positions had been reclaimed by Postgres.
5. Debezium could not resume and immediately failed with: _"The connector is trying to read change stream starting at [...] but this is no longer available on the server."_
6. The source entered a permanent 1s retry loop with no backoff. Each retry reinitializes the JVM, reloading thousands of Postgres custom types, driving sustained memory pressure. All compute pods OOMKilled repeatedly, keeping the cluster in a recovery loop until the source was manually dropped and recreated.

**Root cause:** Two decoupled code paths can advance the Postgres replication slot's `restart_lsn`. Only one is gated by a durable RisingWave checkpoint.
The other — the Kafka Connect embedded engine's periodic `offset.flush.interval.ms=60000` timer — fires independently and advances `restart_lsn` without any checkpoint participation.
The combination of this timer with the CDC backfill phase (during which the checkpoint-gated path is a no-op) creates an unbounded race window that makes data loss guaranteed rather than probabilistic on any pod restart during snapshot.

---

## Error message / log

From `kubectl logs --previous` on the affected compute pod (repeated every ~1s):

```text
INFO  Flush offsets successfully, partition: {server=RW_CDC_74518},
      offsets: {lsn_commit=16697314008, lsn=16697317456, txId=560940372,
      ts_usec=1773237753274507, ...}
      thread="rw-dbz-engine-runner-74518"
      class="com.risingwave.connector.cdc.debezium.internal.ConfigurableOffsetBackingStore"

INFO  Obtained valid replication slot ReplicationSlot
      [active=false, latestFlushedLsn=LSN{3/E3481010}, catalogXmin=560940806]
      thread="pool-3953-thread-1"
      class="io.debezium.connector.postgresql.connection.PostgresConnection"

ERROR Engine has failed with error="...io.debezium.DebeziumException:
      The connector is trying to read change stream starting at
      PostgresOffsetContext [...lsn=LSN{3/E33CCAD8},
      timestamp=2026-03-11T14:02:33.274507Z, snapshot=FALSE...],
      but this is no longer available on the server.
      Reconfigure the connector to use a snapshot mode when needed."
      thread="rw-dbz-engine-runner-74518"
      class="io.debezium.embedded.async.AsyncEmbeddedEngine"

ERROR build stream source reader error, retry in 1s
      source_name="_erp_source" source_id=74518 actor_id=1226890
```

Note the contradiction in the first two lines: `Flush offsets successfully` writes the stale durable offset (`lsn_commit=16697314008` = `3/E33CCAD8`) back into `ConfigurableOffsetBackingStore` on every retry, yet the slot's `latestFlushedLsn` is already at `3/E3481010`.
The engine then immediately tries to stream from `3/E33CCAD8` and fails. This repeats indefinitely.

---

## Root cause analysis

There are two independent code paths that call `replicationStream.flushLsn()` and thereby advance the Postgres replication slot's `restart_lsn`:

**Path 1 — safe (checkpoint-gated):**

```
Barrier checkpoint committed
  → SourceExecutor.persist_state_and_clear_cache()        [src/stream/src/executor/source/source_executor.rs]
    → WaitCheckpointWorker.try_wait_epoch(Committed)      # blocks until Hummock flushes to S3
      → commitOffset(encoded_offset)                      [JniDbzSourceHandler.java:103]
        → DbzChangeEventConsumer.commitOffset()           [DbzChangeEventConsumer.java:334]
          → markProcessed() + markBatchFinished()
            → PostgresStreamingChangeEventSource.commitOffset()
              → replicationStream.flushLsn(lsn)           # advances restart_lsn ✓
```

**Path 2 — dangerous (independent 60s timer, no checkpoint gate):**

```
offset.flush.interval.ms=60000 timer fires               [debezium.properties:28]
  → Kafka Connect WorkerSourceTask internal periodic flush
    → ConfigurableOffsetBackingStore.set()               [ConfigurableOffsetBackingStore.java:190]
      (in-JVM heap; populated from SourceRecord offsets as they pass through the engine)
      → ChangeEventSourceCoordinator.commitOffset()
        → replicationStream.flushLsn(lsn)                # advances restart_lsn WITHOUT checkpoint ✗
```

**Why `noNeedCommitOffset()` does not close the gap:**

`DbzChangeEventConsumer.noNeedCommitOffset()` (line 107) returns `false` for Postgres, causing `handleBatch()` to skip `markProcessed()` and `markBatchFinished()` (lines 306–319).
The intent is to gate all LSN advancement through Path 1. However, the Kafka Connect embedded engine tracks `SourceRecord` offsets internally as they pass through the pipeline, and its periodic flush operates on those internally-tracked offsets regardless of whether `markBatchFinished()` was called.
The suppression is incomplete — it covers only the explicit commit call from `handleBatch()`, not the timer-driven flush.

**Why CDC backfill makes the risk window unbounded:**

During the snapshot scan, `CdcBackfillExecutor` reads rows directly from the external table. No WAL chunks are consumed by `SourceExecutor` during this phase, so `updated_splits_in_epoch` is empty at every barrier.
This causes `update_task_on_checkpoint` to emit `CommitCdcOffset(None)`, which is a no-op in `JniDbzSourceHandler.commitOffset()` (line 105). Path 1 never fires during the snapshot scan.

The risk window is therefore not 60 seconds — it is the entire snapshot duration. For large tables this is minutes to hours. Any pod restart during this window produces an unrecoverable WAL gap.

**Why other connected databases were unaffected:**

Post-backfill sources (`snapshotDone=true`) receive WAL events continuously. Path 1 fires at every barrier interval (~1s globally). The 60s timer can never advance `restart_lsn` further than ~1s of WAL ahead of the durable checkpoint. In practice the rolling update did not land in that narrow window for the already-streaming sources.

---

## Expected behavior

`replicationStream.flushLsn()` for Postgres should only be called after a durable RisingWave checkpoint has confirmed the corresponding offset is persisted. The Kafka Connect periodic flush timer should have no effect on `restart_lsn` advancement for Postgres sources.

---

## To reproduce

1. Create a Postgres CDC source over a large table such that the initial snapshot takes more than 60 seconds.
2. While the snapshot scan is in progress (`snapshotDone=false`), restart or evict the compute node running the CDC source.
3. On restart, Debezium will attempt to resume from the last durably checkpointed LSN. Because no checkpoint-gated LSN commit ever fired during the snapshot (Path 1 is a no-op), the durable offset is from before the snapshot started. The replication slot's `restart_lsn` has been advanced by the 60s timer to a position for which WAL has already been reclaimed.

---

## How deployed

Helm on EKS (Kubernetes), distributed mode.

## Version

`ghcr.io/risingwavelabs/risingwave:v2.8.0-rc.1`

---

## Proposed fix

Remove or effectively disable `offset.flush.interval.ms` in `debezium.properties`:

```diff
-offset.flush.interval.ms=60000
+# offset.flush.interval.ms is intentionally not set for Postgres.
+# LSN advancement is handled exclusively by the checkpoint-gated path in
+# WaitCheckpointWorker → commitOffset(). The periodic Kafka Connect flush timer
+# advances restart_lsn without a durable RisingWave checkpoint, creating an
+# unrecoverable WAL gap on any pod restart during CDC backfill.
```

Path 1 via `WaitCheckpointWorker` already handles all necessary LSN advancement for Postgres. Removing the timer has no effect on streaming correctness; it only removes the dangerous uncheckpointed advancement.

If removing the property entirely is undesirable, setting it to a value larger than any conceivable snapshot duration (e.g. `Long.MAX_VALUE` ms) achieves the same effect.

zwang28:

This PR fixes a bug that a stale WaitCheckpointTask may be incorrectly committed after `try_wait_epoch`.

- Closes #25071

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/executor/source/source_executor.rs
+++ b/src/stream/src/executor/source/source_executor.rs
@@ -1257,7 +1257,13 @@ impl<S: StateStore> WaitCheckpointWorker<S> {
                             },
                         )
                         .await;
-
+                    if self.wait_checkpoint_rx.is_closed() {
+                        // The task must not be committed since the associated epoch may not succeed.
+                        // The wait_checkpoint_rx lifetime is tied to the lifecycle of the source executor actor.
+                        // The old actor must be dropped before any subsequent recovery can succeed; see PartialGraphState::abort_and_wait_actors
+                        tracing::debug!(epoch = epoch.0, "Drop stale wait checkpoint task.");
+                        break;
+                    }
                     match ret {
                         Ok(()) => {
                             tracing::debug!(epoch = epoch.0, "wait epoch success");


src/stream/src/executor/source/source_executor.rs
```diff
@@ -1257,7 +1257,13 @@ impl<S: StateStore> WaitCheckpointWorker<S> {
                             },
                         )
                         .await;
-
+                    if self.wait_checkpoint_rx.is_closed() {
```
wenym1, line 1260: Shall we change to, instead of spawning this `WaitCheckpointWorker`, we store the future in the source executor, and concurrently poll it in a select? In this way, the task can be clean together with abort actors

wcy-fdu:
Looks good overall. Small hardening suggestion: please also check `wait_checkpoint_rx.is_closed()` one more time immediately before` task.run_with_on_commit_success(...).` That would further reduce the tiny race window between the current check and the actual JNI offset commit.

zwang28:
> Looks good overall. Small hardening suggestion: please also check `wait_checkpoint_rx.is_closed()` one more time immediately before` task.run_with_on_commit_success(...).` That would further reduce the tiny race window between the current check and the actual JNI offset commit.

If try_wait_epoch succeeds and self.wait_checkpoint_rx.is_closed() is false, it means the epoch already succeeds.

src/stream/src/executor/source/source_executor.rs
wcy-fdu: +1 for this.

zwang28: > concurrently poll it in a select

The downside is the barrier latency is increased due to the additional tasks in `run_with_on_commit_success`. These tasks typically  communication with external systems, e.g. 
- postgres when flush confirmed LSN for `WaitCheckpointTask::CommitCdcOffset`
- pulsa/pubsub/natsjet ACK message for other `WaitCheckpointTask::`. @tabVersion

zwang28: > These tasks typically communication with external systems

Actually the `run_with_on_commit_success` after `try_wait_epoch` can be spawned to avoid increasing the barrier latency.
But the `try_wait_epoch` must not be spawned. So it still increases the barrier latency. A `try_wait_epoch` may take a long time sometimes.

So I prefer to keep current implementation.

WDYT @wenym1

wcy-fdu (approved):
LGTM

src/stream/src/executor/source/source_executor.rs
wenym1: When we do it in a select concurrently, the two select only competes for CPU time, and won't block each other. The `try_wait_epoch ` is waiting on some external event, rather than consuming CPU time, so I don't think it will block executor and increase barrier latency. But let's merge this PR ASAP and can improve it later.

zwang28: True.

zwang28: > Actually the run_with_on_commit_success after try_wait_epoch can be spawned to avoid increasing the barrier latency.

Implementation Note: Ensure that multiple run_with_on_commit_success calls maintain their original execution order.

PR #25153, opened 2026-03-23
feat(pg-cdc): support convert upstream composite type into `text`

wcy-fdu:

### Summary
Support ingesting Postgres user-defined **composite types** via PG CDC by mapping them to `VARCHAR` and carrying the value as Postgres' native textual form `(a,b,c)`. Covers both snapshot (backfill) and streaming paths, and plays well with auto schema change.

### Background
Before this PR, a Postgres table with a user-defined composite column (e.g. `CREATE TYPE money_type AS (currency varchar, amount decimal)`) couldn't be consumed by PG CDC:

- **Streaming path**: Debezium emits composite columns as `STRUCT` with no usable Kafka Connect schema, and they were silently dropped unless `debezium.include.unknown.datatypes=true` was set — and even then the raw bytes weren't something our parser could turn into a RW datum.
- **Snapshot path**: `tokio-postgres` has no native decoder for user-defined composite types, so the backfill SELECT would fail on those columns.
- **Auto schema change**: `ALTER TABLE ... ADD COLUMN <composite>` events carry `typeName = <user-chosen>` which isn't in our `pg_type` whitelist, so the new column couldn't be added to the MV.

### What Changed

**Java (Debezium side)** — new `PgCompositeToStringConverter.java`, wired up in `postgres.properties` as the `st2str` converter:
- Registered as a Debezium `CustomConverter`. For any column with `jdbcType == Types.STRUCT` (which is how Debezium reports PG composite types), rewrite the schema to `SchemaBuilder.string()` and convert the raw value (`byte[]` / `ByteBuffer` / `toString()`) into a UTF-8 string. The result is the same `(currency,amount)` textual form Postgres itself prints.

**Rust — snapshot reader** (`src/connector/src/source/cdc/external/postgres.rs`):
- At reader init, query `pg_attribute`/`pg_type` for columns where `typtype = 'c'` (composite), and for each such column emit `"col"::text AS "col"` in the snapshot SELECT list. Other `VARCHAR` columns are untouched so RW's native text rendering (`NaN`, `POSITIVE_INFINITY`, etc.) is preserved. On pg_catalog lookup failure we log and fall back to no cast.

**Rust — auto schema change** (`src/connector/src/parser/unified/debezium.rs`):
- In `parse_schema_change`, treat a newly added column as `VARCHAR` if **either** it has `enumValues` (enum, existing behavior) **or** its `jdbcType == 2002` (STRUCT / composite, new). Previously only enum was handled; composite columns would fail the `type_name_to_pg_type` lookup.

**Tests**
- `cdc.share_stream.slt` + `postgres_cdc.sql`: new composite types (`money_type`, `inner_type`, nested `outer_type`) and two MV styles — schema-specified and `(*)` auto-derived — exercising snapshot + streaming INSERT/UPDATE.
- `auto_schema_change_pg.slt`: `ALTER TABLE ... ADD COLUMN price money_type` is picked up as `VARCHAR` and streams correctly.

### User-facing note
To use this, the source must be created with `debezium.include.unknown.datatypes = 'true'` (Debezium native option — otherwise Debezium drops composite columns before our converter ever sees them). Composite columns surface as `VARCHAR` with Postgres text-form values; cast/parse on the RW side if you need structured access.

## Checklist

## Documentation

<details>

</details>


src/connector/src/parser/unified/json.rs
```diff
@@ -519,7 +533,15 @@ impl JsonParseOptions {
                 .into(),
             // ---- Varchar -----
             (DataType::Varchar, ValueType::String) => {
-                return Ok(DatumCow::Borrowed(Some(value.as_str().unwrap().into())));
+                let raw = value.as_str().unwrap();
+                // Fallback for Debezium unknown Postgres composite values that may still arrive
+                // as base64-encoded textual records like "(USD,66.66)".
+                if matches!(self.bytea_handling, ByteaHandling::Base64)
+                    && let Some(decoded) = try_decode_base64_composite_text(raw)
```
xiangjinwu: When the input is a real varchar `KEEp`, it may be mistakenly decoded as `(A)`

xiangjinwu:
There is no e2e tests in this PR. Can we cover backfill + incremental and schema specified + auto derived + auto change?

xiangjinwu:
Check https://github.com/risingwavelabs/risingwave/compare/feat-cdc-struct-converter?expand=1 for an alternative way to handle the unnecessary base64 for structs. (The maven compiler version is being separated here: #25294)

wcy-fdu:
> Check https://github.com/risingwavelabs/risingwave/compare/feat-cdc-struct-converter?expand=1 for an alternative way to handle the unnecessary base64 for structs. (The maven compiler version is being separated here: #25294)

Thanks for investigating this new solution, it's much better. The code has been changed to the new way and tests are applied.

xiangjinwu (approved):
The PR description is a little bit outdated. We no longer do the base64 decoding on strings when a series of conditions are met. Instead, we patched debezium with custom converter so that structs are sent as strings without base64

PR #25165, opened 2026-03-23
feat(iceberg): add variant type for iceberg source

### LINKED ISSUE #26207: file_scan: unwrap on unknown Arrow extension type panics the batch executor on cross-file schema divergence
Found while reviewing #25165; pre-existing (the line predates that PR, `git blame` → 72e5ad3e067, 2024-11-19).

## The bug

`read_parquet_file`'s schema-inference branch unwraps the Arrow→RW type conversion for every field of the file (src/connector/src/source/iceberg/parquet_file_handler.rs:374):

```rust
let data_type = IcebergArrowConvert.type_from_field(field_ref).unwrap();
```

`type_from_field` returns `Err("unsupported extension type: ...")` for any Arrow extension it doesn't know (only `arrow.parquet.variant`, `arrowudf.decimal`, `arrowudf.json` are known). Canonical extensions found in the wild — e.g. `arrow.uuid` or `arrow.json` written by pyarrow and restored from the embedded `ARROW:schema` by `parquet_to_arrow_schema` — make the `.unwrap()` panic the batch worker.

## Reachability

Only the batch `file_scan(...)` executors reach this branch (s3/gcs/azblob pass `rw_columns=None, parser_columns=None`); the streaming file source uses the `Some` branch. The frontend bind path runs the same conversion on `files[0]` with `?` (src/frontend/src/expr/table_function.rs), so a single file or a uniform directory fails cleanly at bind. The panic window is **cross-file schema divergence in a directory scan**: bind samples the first file (clean), runtime hits a sibling file carrying an unsupported extension.

Compounding: `from_field` also has no arm for `Time32(_)`/`Time64(Nanosecond)` (see the companion TIME-unit issue), so a sibling file with a `TIME(millis)` column trips the same unwrap.

## Fix

Propagate the error with context instead of unwrapping — the same treatment #25165 applied to the identical pattern in `extract_iceberg_columns` (frontend iceberg schema derivation). Per the repo panic policy, externally-supplied file schemas must not be able to panic an executor.

yuhao-su:

Support reading Iceberg **VARIANT** columns (format v3) from the Iceberg source and refreshable Iceberg tables. Variant values are decoded to RisingWave's native **VARIANT** type (#26080, which this PR is stacked on; the stack sits on #26354, which makes the parquet decode follow the declared schema at every nesting depth), including variants nested in struct / list / map columns. Variant type identity is preserved (`variant_typeof`, `variant_get` work on the decoded values), and values can be converted with the `variant → jsonb` / `variant → varchar` casts.

### How it works

- Schema derivation maps Iceberg `variant` to RisingWave `variant`: the Arrow struct field carrying the `arrow.parquet.variant` extension is typed as `DataType::Variant` (`src/common/src/array/arrow/arrow_iceberg.rs`).
- Arrays are decoded per row via `VariantVal::from_parts(metadata, value)`, which fully validates the untrusted bytes and re-encodes them into RW's canonical variant form (sorted dictionaries, canonical float bit patterns) that the byte-wise `Eq`/`Ord` of the type relies on.
- `IcebergArrowConvert::to_arrow_field` maps `DataType::Variant` to the extension-tagged struct field (schema-only; converting variant *arrays* to Arrow remains unsupported), so parquet schema matching for declared variant columns works.
- The heavy lifting lives in the iceberg-rust fork (risingwavelabs/iceberg-rust#145, merged; pinned rev `edf72a59` on `dev_rebase_main_20260303`): variant extension metadata on both schema-conversion directions, projection-mask support for variant's id-less Parquet sub-fields, unique Avro record names, rejection of shredded variant files, and enabling parquet's `variant_experimental` feature so Arrow schemas derived from Spark/iceberg-java-written files (Parquet `VARIANT` logical type, no embedded Arrow schema) carry the extension metadata. The workspace also declares `variant_experimental` on its own `parquet` dependency, so the generic parquet path does not depend on feature unification from the fork.
- `iceberg-compaction-core` is pinned to the nimtable/iceberg-compaction main rev `2f1d3714` (nimtable/iceberg-compaction#161, merged), which uses the same iceberg-rust rev — keeping a single copy of the `iceberg` crates in the graph.
- **Shredded variant columns read as NULL instead of failing the file.** The fork's reader rejects a file with shredded variant columns (`typed_value` sub-field) before yielding any batch; previously that error failed batch scans and made streaming ingestion retry the same file every barrier, forever, while the job looked healthy. The single per-task scan funnel now catches that rejection, retries without the variant columns, and splices all-NULL variant columns back at their positions with a per-file warning — the rest of the file stays readable. The decode path (generic parquet sources) likewise reads a shredded column as whole-column NULL rather than erroring: decoding only metadata/value would yield silently *partial* objects, which is worse than NULL.

### Bug fixes included

- **Predicate pushdown on non-primitive columns.** iceberg-rust only builds field accessors for primitive columns, so a pushed-down `IS [NOT] NULL` on a struct/list/map (pre-existing on main) or variant column failed the whole scan with `Accessor for Field ... not found`. Such conjunctions now stay in the RisingWave filter. The check is enforced at a single point (`input_ref_to_reference`) shared by all predicate arms, the two near-identical pushdown implementations (batch rule vs. `utils::iceberg_predicate`) are deduplicated, and pushability is judged by the **iceberg-side** column type — so Iceberg engine tables, whose scan schema is remapped to Hummock types (e.g. jsonb backed by an iceberg string), keep their `IS [NOT] NULL` pushdown.
- **Corrupt variant values no longer panic.** Decoding goes through the fully-validating `VariantVal::from_parts`; malformed bytes from an external writer decode to NULL with a warning instead of panicking inside the upstream decoder (which would abort the compute node on the streaming scan path). Shredded variant arrays (`typed_value` present) read as whole-column NULL at the converter boundary — the upstream decoder panics or silently nulls on them, and decoding only metadata/value would yield partial objects.
- **Generic parquet file sources.** With parquet's `variant_experimental` enabled, `file_scan()` and `ENCODE PARQUET` sources also infer variant columns, but the parquet-vs-source schema match only compared bare `DataType`s, silently NULL-filling the column. A field-aware match (`is_parquet_field_match_source_schema`) now recognizes variant columns (including nested in struct/list/map) and converts them via the extension-aware path; the `arrow.parquet.variant` extension binds **exclusively** to `variant`, so declaring such a column as its raw physical struct is an illegal type mismatch and NULL-fills. See the breaking-change note below — enabling `variant_experimental` is what first puts the extension on these fields, so this changes behaviour for pre-existing sources. The struct projection path (declared subset/reordered struct) always decodes each child by the declared field — extensions on file-side children no longer override the declared type (a child tagged with an unrelated extension, e.g. the canonical `arrow.uuid` on a bytea-declared child, previously failed or mistyped the column) — and takes child types from the conversion result, so a variant child decodes correctly there too (the declared `variant` field carries the extension itself). The CREATE SOURCE variant gate exempts `ENCODE PARQUET`, which can actually produce variant values; the same exemption applies to `ALTER SOURCE ... ADD COLUMN` (the gate itself now also covers the ALTER and CDC-table bind paths, fixed in #26080).
- **Flipped comparison pushdown** — the semantic fix (swapped strictness for literal-on-the-left comparisons) landed separately in #26195; this PR only collapses the six duplicated comparison arms into one helper, keeping the mirroring pinned by the unit test.
- **Declared-schema enforcement at all nesting depths** lives in the base PR #26354: decode follows the declared field everywhere (extras dropped, reorders resolved by name, file-side extensions ignored), so no strict matcher rules are needed. This PR adds the one deliberate matcher exception on top: the `arrow.parquet.variant` extension **binds exclusively to `variant`** at every depth — declaring such a column as any other type (e.g. its raw physical struct) is an illegal type mismatch and NULL-fills, and an undeclared variant extra inside a container is simply dropped by the declared-side decode (relevant since the `variant_experimental` parquet feature enabled by this PR's dependency bump now tags nested VARIANT fields with the extension).
- **A failed `CREATE TABLE ... ENGINE = ICEBERG` no longer drops a pre-existing external iceberg table.** `create_table_if_not_exists` now reports whether it actually created the table, and the failure cleanup only drops it in that case. Previously any late failure — e.g. the meta-side VARIANT sink rejection, made reachable by this type — deleted an external (e.g. Spark-created) table the statement never created.
- **`file_scan()` no longer panics on files with un-decodable extension metadata** (fixes #26207): inferring column types went through a raw `.unwrap()`; a file whose embedded metadata carries an unknown or misplaced extension type now reports a contextual per-file error instead of aborting the batch worker.
- **VARIANT is excluded from the DataFusion batch path.** `to_arrow_field(Variant)` succeeding (schema-only) would have admitted variant-bearing plans into the default-on DataFusion engine, where a top-level variant expression output hits the unsupported RW→Arrow array conversion at runtime with no fallback. (Keying on a variant — grouping, ordering, joining — is already rejected outright by the batch `reject_variant_keys` gate in the base PR, so it never reaches either engine.) The DataFusion execute checker now rejects any plan whose node schemas contain VARIANT (including nested in struct/list/map), falling back to the RisingWave engine at plan time. A variant value created and consumed *inside* one expression (e.g. a `varchar → variant` cast under `IS NOT NULL`) never surfaces in a node schema, so the RwScalarFunction fallback wrapper additionally refuses any subtree whose return type contains VARIANT — plan conversion then errors and the query falls back at plan time instead of hitting the unsupported RW→Arrow conversion at runtime.

### Limitations

- **Read-only.** Writing VARIANT through the Iceberg sink and creating Iceberg tables with VARIANT columns (`create_table_if_not_exists`, Iceberg engine tables) are not supported; sinking VARIANT is rejected at sink creation (#26080).
- **Shredded variant files** (with `typed_value` sub-fields) read the variant column as NULL with a per-file warning; sibling columns stay readable. Files registered via `add_files` from writers that embed no iceberg field ids take a different reader path and currently fail with a generic error instead (#26528).
- A variant value with corrupt encoding decodes to NULL with a warning. Non-finite floats are preserved (NaN normalizes to the canonical bit pattern) and render as the strings `"NaN"` / `"Infinity"` / `"-Infinity"` through the jsonb/text casts, following `to_jsonb(float8)`. Scalar predicates on variant columns are not pushed down (per spec, variant does not support them).

### Follow-ups (TODO)

- **Cache variant metadata validation.** `variant_array_to_variant` re-validates and re-encodes the variant metadata dictionary for every row, but rows of a batch virtually always share one metadata blob; caching the validated metadata would remove O(rows × dict size) redundant work on the decode hot path. (TODO comment in `variant_array_to_variant`.)
- **Included: the iceberg-rust predicate fix.** A variant column's `metadata`/`value` Parquet leaves carry no field ID, so `build_field_id_map` reported the whole file as ID-less and the predicate path fell back to position-based IDs — shifting every column after the variant. On `(payload variant, a int, b int)`, `where b = 200` silently returned 0 of 3 rows and `where a = 2` failed with "isn't a root column in Parquet schema". Projection was unaffected, so it only showed up once a predicate was pushed down, and it happened to be invisible when the variant column was last (which is why the e2e table here, filtering only on its first column `id`, does not catch it). Reproduced end to end against a Spark-written v3 table. Fixed in risingwavelabs/iceberg-rust#190 and picked up here (`edf72a59`); nimtable/iceberg-compaction#161 moves the compaction side to the same rev (`2f1d3714`), so a single copy of the `iceberg` crates stays in the graph. Verified end to end after the bump: every previously-wrong query returns the right rows, and control tables are unaffected.
- **#26368 will remove most of the matcher scaffolding added here.** Once declared-side decode is driven by `DataType` instead of an arrow `Field`, `to_arrow_field(Variant)` can go back to erroring and `is_parquet_field_match_source_schema` folds into the decoder. The one thing that must survive that refactor is the deliberate semantic it encodes — the `arrow.parquet.variant` extension binds exclusively to `variant` — otherwise decode silently reverts to reading the raw physical struct. The e2e case `parquet_variant_as_struct` pins it independently of the matcher.
- New issues from the pre-merge review, tracked separately: #26525 (the streaming iceberg fetch executor retries a deterministically failing file forever with no user-visible error — pre-existing on main, independent of variant), #26526 (variant decode/render recursion has no depth limit; a hostile deeply-nested value can stack-overflow the process — this PR is the first channel feeding untrusted file bytes into it), #26527 (corrupt variant values map to silent NULLs with no metric or row accounting), #26528 (id-less shredded files miss the NULL fallback and fail with an opaque Arrow error).
- Pre-existing issues found during review, tracked separately: #26353 (pushed-down `NOT IN` returns wrong rows for schema-evolved files missing the column), #26206 (matcher/decoder extension-registry gap — still open), #26213 (iceberg scan does no schema validation of upstream parquet; bare `assert_eq!` panics the batch executor), #26375 (legal `int → long` promotion panics batch scan when the source catalog and the snapshot schema disagree), #26461 (`extract_iceberg_columns` does not reject user columns colliding with RisingWave's reserved names). #26208, #26209 and #26216 are now closed.

## Checklist

### Tests

- Unit: `risedev test variant` — variant decode for all variant primitive types (compared via canonical text), variants nested in struct/list/map, map-key rejection, malformed-value handling, shredded rejection, projected-struct decoding (type and value), field-aware schema matching (variant recognized at any depth; variant-extension-declared-as-non-variant rejection incl. scalars and the raw physical struct), nested-struct reorder/superset/depth-2-extension decoding by the declared schema, and comparison-pushdown operator mirroring.
- E2E: `e2e_test/iceberg/test_case/iceberg_source_variant.toml` (+ `.slt`) — Spark writes a v3 table with 21 variant columns plus struct/list/map-nested variants across 3 rows / 2 data files; asserts types (`pg_typeof` = `variant`, variant type identity via `variant_typeof`), values (via `variant → jsonb` casts and `variant_get`) through both the batch source and a `refresh_mode = 'FULL_RELOAD'` table, covering SQL NULL vs variant null, empty object/array, deep nesting, NaN preserved as a canonical variant double, projections with and without variant columns, and the DataFusion-fallback shape of a bare top-level variant output. The table also carries an `evolved array<struct<a int, v variant>>` column asserting native decode of a variant inside `list<struct>`, plus a DataFusion-fallback query with a variant-typed subexpression over a variant-free plan schema. The same data files are also read through the generic parquet path: `file_scan()` (incl. variants nested in struct/list/map), an `ENCODE PARQUET` source with a declared subset + reordered struct exercising the declared-side decode, and a source declaring `evolved` as `struct<a int>[]` asserting the undeclared variant extra is dropped and the declared subset decodes. `e2e_test/ddl/variant.slt` additionally pins the `ALTER SOURCE ... ADD COLUMN` gate (rejected for JSON encode, allowed for PARQUET) and a plain-table `ALTER TABLE` variant round trip. The case also covers: schema evolution (a variant column added after files exist NULL-fills in old files, decodes in new ones), predicates on a column positioned *after* a variant column (pinning the field-id-shift fix from the fork), `IS [NOT] NULL` on variant/struct columns staying RW-side, streaming MVs over the iceberg source (regular decode and the shredded NULL fallback), and an `ALTER SOURCE ... ADD COLUMN variant` that actually decodes file data through the added column. Verified locally.
- E2E (shredded): `e2e_test/iceberg/test_case/iceberg_source_variant_shredded.toml` (+ `.slt`) — Spark's **native** parquet writer produces genuinely shredded files (`spark.sql.variant.writeShredding.enabled` + a forced shredding schema; the iceberg writer never shreds), a helper stamps the iceberg field ids the projection path requires and registers them via `add_files`; asserts the variant column reads as NULL while sibling columns keep their values, and that row counts survive when only the shredded column is projected (empty-projection path). Verified locally including the per-file warning in the compute logs.
- E2E (sink gate): `e2e_test/ddl/variant.slt` now also asserts external sinks with a variant column (top-level and nested in a struct) are rejected before any connection attempt, while blackhole and sink-into-table stay allowed.
- Fork: 1271 lib tests incl. projection-mask/schema-conversion variant tests and a regression test for Arrow schemas derived from Parquet `VARIANT`-annotated files.

## Documentation

<details>

The Iceberg source and refreshable Iceberg tables can now read Iceberg v3 **VARIANT** columns. Variant values are exposed as RisingWave's native `variant` type (including variants nested inside struct, list, and map columns), preserving variant type identity; use `variant_get` / `variant_typeof` / casts to `jsonb` and `varchar` to work with them. Previously a v3 table containing a VARIANT column could not be read at all — `CREATE SOURCE` failed while parsing the table metadata, so even the non-variant columns were unreachable.

Writing VARIANT via the Iceberg sink is not supported yet. Shredded variant columns cannot be reconstructed yet: they read as NULL (with a warning in the logs) while the file's other columns stay readable.

**Breaking changes for generic Parquet sources.** This release enables Parquet's `variant_experimental` feature, so a Parquet `VARIANT` column now arrives tagged with the `arrow.parquet.variant` Arrow extension instead of as an untagged `struct<metadata, value>`. Two consequences for sources reading such files:

- An `ENCODE PARQUET` source or table that **declared** the column as `struct<metadata bytea, value bytea>` previously read the raw bytes; it now NULL-fills, because the variant extension binds exclusively to `variant`. Redeclare the column as `variant`.
- `file_scan()` **infers** its schema, so such a column now comes back as `variant` instead of `struct<metadata bytea, value bytea>`. Queries that destructured `.metadata` / `.value` need updating to `variant_get` / casts.

</details>


yuhao-su:
* **#25165** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25165?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25165?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#26080** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/26080?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#26354** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/26354?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

chenzl25 (approved):
@yuhao-su The read part of iceberg variant type LGTM. Could you please make it ready for review and merge this PR?

e2e_test/iceberg/README.md
chenzl25, line 1: Is this README still validated?

yuhao-su: Yes

chenzl25:
Shall we consider rebasing this PR onto https://github.com/risingwavelabs/risingwave/pull/26080 to support the true variant type?

src/connector/src/source/iceberg/mod.rs
```diff
@@ -585,6 +622,44 @@ pub async fn scan_task_to_chunk_with_deletes(
...
+
+/// The projected VARIANT columns of a task: their field ids, their positions in the
+/// projected column order, and their names.
+fn projected_variant_columns(task: &FileScanTask) -> (Vec<i32>, Vec<usize>, Vec<String>) {
+    let mut field_ids = Vec::new();
+    let mut positions = Vec::new();
+    let mut names = Vec::new();
+    for (position, field_id) in task.project_field_ids.iter().enumerate() {
+        if let Some(field) = task.schema.field_by_id(*field_id)
+            && matches!(field.field_type.as_ref(), iceberg::spec::Type::Variant(_))
```
chenzl25, line 640: Non-blocking: this fallback only covers top-level VARIANT columns. A shredded VARIANT nested in a projected struct/list/map remains unsupported and will still fail the file read.

Issue #24763, opened 2026-02-09
[Snowflake Sink V2] PIPE object uses incorrect path pattern causing cross-table data loading


sriram1337:
### Problem
The automatically created PIPE object uses the path pattern @STAGE_NAME/table_name instead of @STAGE_NAME/table_name/ (missing trailing slash).
### Impact
The pipe loads data from ALL directories matching the prefix, not just the exact table directory. For example:

- PIPE for dim_project loads from both:
   - s3://bucket/dim_project/
   - s3://bucket/dim_project_contract/

This causes:

- Data corruption: Wrong data loaded into tables
- Duplicate processing: Same files processed by multiple pipes
- Very difficult to debug: No obvious errors, just incorrect query results

### Expected Behavior
PIPE should use exact directory matching with trailing slash: @STAGE_NAME/table_name/

### Error message/log

```text

```

hzxa21:
@xxhZs PTAL

PR #25169, opened 2026-03-24
fix: fix snowflake decimal and stage path name

### LINKED ISSUE #24762: [Snowflake Sink V2] DECIMAL type defaults to (38, 0) causing precision loss
### Problem
The Snowflake sink creates tables with DECIMAL(38, 0) for all decimal columns, which truncates decimal values from currency and other precision-dependent fields.
### Root Cause
RisingWave doesn't support specifying precision and scale for DECIMAL types, so the sink uses Snowflake's default DECIMAL type.
### Impact
- Data quality issues: Currency values lose cents/fractional components
- Silent data corruption: No errors or warnings, just truncated values
- Difficult to debug: Requires comparing source and sink data to discover
### Expected Behavior
Either:
 - Allow users to specify column type mappings/overrides in sink configuration
 - Preserve the source precision/scale when creating target tables
 - Use a more sensible default like DECIMAL(38, 9) for general use

### Error message/log

```text

```

### To Reproduce

_No response_

### Expected behavior

_No response_

_No response_

### The version of RisingWave

_No response_

### Additional context

_No response_

xxhZs:

This pull request improves the handling of data type conversions for the Redshift and Snowflake sinks, with a particular focus on correct and consistent mapping of the `Decimal` type. Now the default is fixed as decimal (38, 10).

Also fixed the name error of the stage path in the pipe.

close: https://github.com/risingwavelabs/risingwave/issues/24762
https://github.com/risingwavelabs/risingwave/issues/24763

## Checklist

## Documentation

<details>

</details>


src/connector/src/sink/snowflake_redshift/redshift.rs
```diff
@@ -844,7 +839,10 @@ fn convert_redshift_data_type(data_type: &DataType) -> Result<String> {
         DataType::Timestamp => "TIMESTAMP".to_owned(),
         DataType::Timestamptz => "TIMESTAMPTZ".to_owned(),
         DataType::Jsonb => "VARCHAR(MAX)".to_owned(),
-        DataType::Decimal => "DECIMAL".to_owned(),
+        // Redshift's DECIMAL without explicit precision defaults to (38,0), which drops all
+        // fractional digits. We use DECIMAL(38, 10) consistent with the Iceberg/Snowflake sink
+        // convention, though values with more than 10 fractional digits may still lose precision.
+        DataType::Decimal => "DECIMAL(38, 10)".to_owned(),
```
tabVersion, line 845: Risingwave decimal has 28 digits, much smaller range than `decimal (38, 10)`. How do you think about the data range here?

src/connector/src/sink/snowflake_redshift/snowflake.rs
```diff
@@ -975,8 +981,11 @@ fn build_create_pipe_sql(
     target_table_name: &str,
 ) -> String {
     let pipe_name = format!(r#""{}"."{}"."{}""#, database, schema, pipe_name);
+    // Trailing `/` is required to enforce exact directory matching.
+    // Without it, a PIPE for table "dim_project" would also match files under
+    // "dim_project_contract/" since Snowflake uses prefix matching.
     let stage = format!(
-        r#""{}"."{}"."{}"/{}"#,
+        r#""{}"."{}"."{}"/{}/"#,
```
tabVersion, line 988: good catch

src/connector/src/sink/snowflake_redshift/redshift.rs
xxhZs: This is aligned with Iceberg.

chenzl25 (approved):

PR #25180, opened 2026-03-25
feat: disable bloom filter for kv log store tables

hzxa21:
## Summary
- add a generic internal `disable_bloom_filter` table catalog flag
- mark kv log store state tables with that flag
- skip bloom filter key extraction for tables with `disable_bloom_filter = true`

## Changes
- add `disable_bloom_filter` to `catalog.Table`, frontend `TableCatalog`, meta model/controller, plus a meta migration
- set `disable_bloom_filter = true` in kv log store table builders and kv log store test table helper
- make `FilterKeyExtractorImpl::from_table` return `DummyFilterKeyExtractor` for flagged tables
- add a storage unit test to verify flagged tables produce no bloom filter key contribution

## Verification
- `cargo check -p risingwave_meta_model`
- `cargo test -p risingwave_frontend table_catalog --no-run`
- `cargo test -p risingwave_storage compaction_catalog_manager --no-run`
- `cargo test -p risingwave_storage disable_bloom_filter_uses_dummy_extractor`


Li0k (approved):
LGTM.

hzxa21:
@wenym1 can you help review this PR as well? I haven't conduct any test other than UT.

Li0k (approved):
Is using **prefix_hint** to implicitly bypass BF a bit too implicit? How can we make sure the upper layer won’t get it wrong?

In other words, how do we guarantee the read behavior stays consistent regardless of whether the BF is empty or not? 

cc @hzxa21 @wenym1

wenym1:
Update pushed in b8508def4b.

Summary:
- Propagated `disable_bloom_filter` into `StorageTableDesc`/`TableDesc` so storage readers can see the table-level flag.
- Suppressed point-get and prefix-scan prefix hints in `BatchTable` and `StateTable` when bloom filters are disabled.
- Added a simulation regression test for sink log-store internal table point gets after flush.

Validation:
- `cargo fmt`

wenym1:
Update pushed in `5114eee0e4`.

Motivation:
- `read_prefix_len_hint = 0` already means the normal `StateTable`/`BatchTable` read paths do not set `ReadOptions.prefix_hint`, and Hummock only checks SST bloom filters when a prefix hint is present.
- Therefore KV log-store tables do not need a separate persisted catalog flag to disable bloom participation. Using the existing zero read-prefix signal avoids schema/protobuf/meta migration churn and keeps the reader/writer semantics aligned.

Summary:
- Removed the dedicated `disable_bloom_filter` catalog/proto/meta plumbing and its migration.
- Set KV log-store table catalogs to use `read_prefix_len_hint = 0`.
- Changed `FilterKeyExtractorImpl::from_table` so zero read-prefix tables use `DummyFilterKeyExtractor`, avoiding bloom entries for tables whose normal readers will not use bloom.
- Kept a shared `should_calculate_prefix_hint` helper for `StateTable`/`BatchTable` prefix-hint decisions.
- Adjusted low-level hummock test table-id registration to preserve legacy full-key bloom behavior where tests manually pass full-key prefix hints.
- Added regression coverage for the zero-prefix dummy extractor and sink log-store point gets after flush.

Validation:
- `cargo check -p risingwave_pb`
- `cargo fmt`
- `cargo test -p risingwave_storage -p risingwave_hummock_test --all-features -- --test-threads=1`

Note: the default parallel storage test run exposed existing order sensitivity in `hummock::event_handler::hummock_event_handler::tests::test_clear_tables`; it passed when run alone and in the serial full storage+hummock run above.

Li0k (approved):
LGTM

wenym1:
Buildkite unit-test failure was the planner golden for `risingwave_planner_test::planner_test_runner sink`: after changing KV log-store tables to use `read_prefix_len_hint = 0`, the distsql sink output still expected `read pk prefix len hint: 3` for the sink log-store table.

Fix pushed in `524bef3a83` updates `src/frontend/planner_test/tests/testdata/output/sink.yaml` so the two affected sink cases expect `read pk prefix len hint: 0`.

Local validation:
- `cargo test -p risingwave_planner_test --test planner_test_runner sink -- --nocapture`

hzxa21 (commented):
LGTM.

PR #25184, opened 2026-03-26
feat(cdc): support ingest upstream pg vector type

wcy-fdu:

### Summary
Add end-to-end support for PostgreSQL pgvector in CDC (vector -> vector), covering both snapshot and Debezium incremental paths.

### What changed
Postgres snapshot parser

- Added vector decoding in src/connector/src/parser/postgres.rs.

  - Supports both pgvector binary format and ~~text fallback~~ (AI hallucination).
  - Parses into RW DataType::Vector, with dimension checks.
  - Debezium JSON parser

- Added DataType::Vector parsing in src/connector/src/parser/unified/json.rs.

  - Supports vector values encoded as:
  - ~~string: "[...]"~~ (AI hallucination),
  - array: [ ... ].
  - Added finite/range checks for float conversion.

- Postgres source validator

  - Updated PostgresValidator to allow USER-DEFINED vector -> VECTOR.
  - Explicitly keeps halfvec / sparsevec unsupported for now.

###  Tests

- Added unit tests for pgvector ~~text~~/binary decoding in postgres.rs.
- Added e2e CDC test:

  - e2e_test/source_inline/cdc/postgres/postgres_pgvector_vector.slt
  - covers backfill + update/insert/delete + upstream/RW value consistency checks.
  - We perform validations during table creation, auto schema map, and auto schema change, and will directly report errors for cases where vector dimensions are not defined or do not match.

### Outcome
pgvector columns are no longer parsed as NULL in CDC when mapped to RW vector(n), and both snapshot and incremental ingestion paths are validated.

## Checklist

## Documentation

<details>

</details>


src/connector/src/parser/postgres.rs
```diff
@@ -14,17 +14,80 @@
...
+
+    fn from_sql(
+        _ty: &Type,
+        raw: &'a [u8],
+    ) -> Result<Self, Box<dyn std::error::Error + Sync + Send>> {
+        // PostgreSQL usually sends values in binary format through tokio-postgres.
+        // Keep this text fallback for compatibility with format negotiation fallback paths.
+        if raw.starts_with(b"[") {
+            return Self::parse_text(raw);
+        }
```
xiangjinwu: When does this fallback happen? The `ToSql` for all other types do not handle text parsing, and I did not see any doc about this previously

src/connector/src/parser/unified/json.rs
```diff
@@ -679,6 +679,55 @@ impl JsonParseOptions {
                 builder.finish()
             })
             .into(),
+            // ---- Vector -----
+            (DataType::Vector(size), ValueType::String) => {
+                let text = value.as_str().unwrap();
```
xiangjinwu: When is this branch reachable? Isn't it always array?
https://debezium.io/documentation/reference/stable/connectors/postgresql.html#postgresql-pgvector-types

src/connector/src/parser/unified/json.rs
```diff
@@ -679,6 +679,55 @@ impl JsonParseOptions {
...
+            (DataType::Vector(size), ValueType::Array) => {
+                let array = value.as_array().unwrap();
+                if array.len() != *size {
+                    Err(create_error())?
+                }
+                let mut elems = Vec::with_capacity(array.len());
+                for v in array {
+                    let value = match v.value_type() {
+                        ValueType::I64 | ValueType::I128 | ValueType::U64 | ValueType::U128 => {
+                            v.try_as_i64().map_err(|_| create_error())? as f32
```
xiangjinwu: For I128/U64/U128, `try_as_i64` would definitely be a problem.

java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/PostgresValidator.java
```diff
@@ -811,6 +811,13 @@ private boolean isDataTypeCompatible(ColumnInfo colInfo, Data.DataType.TypeName
                         case "geometry":
                             // PostGIS GEOMETRY -> BYTEA (stored as EWKB bytes)
                             return val == Data.DataType.TypeName.BYTEA_VALUE;
+                        case "vector":
+                            // pgvector VECTOR -> VECTOR
+                            return val == Data.DataType.TypeName.VECTOR_VALUE;
```
xiangjinwu, line 859: What about auto schema derivation? How do we get the vector length from upstream, and return an error when length is flexible in upstream?

xiangjinwu:
To resolve the CI issue, I [prepared](https://gist.github.com/xiangjinwu/c586f01fb944467f2f0ffc0df7890fb2) a PostgreSQL image with both PostGIS and pgvector:
https://github.com/risingwavelabs/risingwave/commit/1ff824c9a59625d14bb427504e808345b0a20580 (Thanks @cyliu0 for help)
https://buildkite.com/risingwavelabs/pull-request/builds/94370

src/connector/src/parser/unified/json.rs
```diff
@@ -679,6 +679,46 @@ impl JsonParseOptions {
...
+                let array = value.as_array().unwrap();
+                if array.len() != *size {
+                    Err(create_error())?
+                }
+                let mut elems = Vec::with_capacity(array.len());
+                for v in array {
+                    let value = match v.value_type() {
+                        ValueType::I64 | ValueType::I128 => {
+                            let i128_value = v.try_as_i128().map_err(|_| create_error())?;
+                            if i128_value < f32::MIN as i128 || i128_value > f32::MAX as i128 {
```
xiangjinwu: `as` does not report error properly; just like you can `-1i8 as u8` which gives you `255` silently. Into/TryInto is the proper conversion tools.

wcy-fdu:
ready for review cc @xiangjinwu

e2e_test/source_inline/cdc/postgres/postgres_pgvector_vector.slt
```diff
@@ -0,0 +1,159 @@
...
+  database.name = '${PGDATABASE:postgres}',
+  schema.name = 'public',
+  slot.name = 'pgvector_slot'
+);
+
+statement ok
+create table test_pgvector_cdc (
+    id int primary key,
+    name varchar,
+    embedding vector(3)
```
xiangjinwu, line 41: What if it does not equal to upstream?

xiangjinwu (commented):
Can we also test that new columns added by auto schema change can detect the length correctly as well?

wcy-fdu:
> Can we also test that new columns added by auto schema change can detect the length correctly as well?

In auto schema change events, pgvector columns sometimes come through as just vector without the dimension n, so we can’t reliably map them to RisingWave’s required vector(n) type. I’m working on a fix.

java/connector-node/risingwave-connector-service/src/main/java/com/risingwave/connector/source/common/PostgresValidator.java
wcy-fdu: Should return error in frontend

wcy-fdu:
Note: 
We perform validations during table creation, auto schema map, and auto schema change, and will directly report errors for cases where vector dimensions are not defined or do not match.

src/connector/src/parser/unified/debezium.rs
```diff
@@ -239,6 +366,43 @@ pub fn parse_schema_change(
...
+                                let Some((schema_name, table_name_only)) =
+                                    parse_schema_table_from_debezium_id(id.as_str())
+                                else {
+                                    return Err(AccessError::CdcAutoSchemaChangeError {
+                                        ty: type_name,
+                                        table_name: format!("{}.{}", source_name, table_name),
+                                    });
+                                };
+
+                                let cache_key = format!("{}.{}", schema_name, table_name_only);
```
xiangjinwu: Why split the `id` into `(schema_name, table_name_only)` above and then combine them again here? Can't we just use `id`?

src/connector/src/parser/unified/debezium.rs
```diff
@@ -61,6 +62,128 @@ fn is_bigserial_default(default_value_expression: &str) -> bool {
     expr.starts_with("nextval(")
 }
 
+fn parse_schema_table_from_debezium_id(id: &str) -> Option<(String, String)> {
```
xiangjinwu, line 79: Need doc comment. What does the input id look like, and why are there two different formats to detect and parse?

src/connector/src/parser/unified/debezium.rs
```diff
@@ -61,6 +62,128 @@ fn is_bigserial_default(default_value_expression: &str) -> bool {
...
+    let host = props
+        .get("hostname")
+        .ok_or_else(|| AccessError::Uncategorized {
+            message: "missing `hostname` in postgres-cdc properties".to_owned(),
+        })?;
+    let port = props
+        .get("port")
+        .ok_or_else(|| AccessError::Uncategorized {
+            message: "missing `port` in postgres-cdc properties".to_owned(),
+        })?;
```
xiangjinwu: This logic must already exist somewhere. Consider reuse a shared utility to ease future maintenance.

src/connector/src/parser/unified/debezium.rs
```diff
@@ -61,6 +62,128 @@ fn is_bigserial_default(default_value_expression: &str) -> bool {
...
+        .query(
+            r#"
+            SELECT a.attname AS column_name,
+                   a.atttypmod AS atttypmod
+            FROM pg_catalog.pg_attribute a
+            JOIN pg_catalog.pg_class c ON c.oid = a.attrelid
+            JOIN pg_catalog.pg_namespace n ON n.oid = c.relnamespace
+            JOIN pg_catalog.pg_type t ON t.oid = a.atttypid
+            WHERE n.nspname = $1
+              AND c.relname = $2
```
xiangjinwu: This looks familiar to me, actually in this same PR as `DISCOVER_FORMATTED_TYPE_QUERY`. Any chance to avoid maintaining two copies?

xiangjinwu (commented):
Seems auto schema change is not included in e2e tests. Can we `add column` in upstream PostgreSQL?

src/connector/src/parser/unified/debezium.rs
```diff
@@ -61,6 +62,76 @@ fn is_bigserial_default(default_value_expression: &str) -> bool {
...
+        if parts.len() >= 2 {
+            let schema = parts[parts.len() - 2].to_owned();
+            let table = parts[parts.len() - 1].to_owned();
+            if !schema.is_empty() && !table.is_empty() {
+                return Some((schema, table));
+            }
+        }
+    }
+
+    let cleaned = trimmed.trim_matches('"');
```
xiangjinwu, line 95: This line does not make sense to me but let's wait until it breaks.

src/connector/src/connector_common/postgres.rs
```diff
@@ -47,6 +47,116 @@ const DISCOVER_PRIMARY_KEY_QUERY: &str = r#"
...
+
+pub struct PgConnectionConfig {
+    pub host: String,
+    pub port: String,
+    pub user: String,
+    pub password: String,
+    pub database: String,
+    pub ssl_mode: SslMode,
+    pub ssl_root_cert: Option<String>,
+}
```
xiangjinwu, line 78: https://github.com/risingwavelabs/risingwave/blob/a3597677be46f1c3636b7402d42e80ea6387deaa/src/connector/src/sink/postgres.rs#L44-L70
Also:
https://github.com/risingwavelabs/risingwave/blob/a3597677be46f1c3636b7402d42e80ea6387deaa/src/batch/executors/src/executor/postgres_query.rs#L39-L47
And (this seems to cover MySQL as well)
https://github.com/risingwavelabs/risingwave/blob/a3597677be46f1c3636b7402d42e80ea6387deaa/src/connector/src/source/cdc/external/mod.rs#L282-L306

src/connector/src/connector_common/postgres.rs
```diff
@@ -47,6 +47,116 @@ const DISCOVER_PRIMARY_KEY_QUERY: &str = r#"
...
+        password: props.get("password").cloned().unwrap_or_default(),
+        database: props
+            .get("database.name")
+            .context("missing `database.name` in postgres-cdc properties")?
+            .clone(),
+        ssl_mode: props
+            .get("ssl.mode")
+            .and_then(|v| v.parse::<SslMode>().ok())
+            .unwrap_or_default(),
+        ssl_root_cert: props.get("ssl.root.cert").cloned(),
```
xiangjinwu, line 105: https://github.com/risingwavelabs/risingwave/blob/a3597677be46f1c3636b7402d42e80ea6387deaa/src/connector/src/source/cdc/enumerator/mod.rs#L214-L242

src/connector/src/connector_common/postgres.rs
```diff
@@ -47,6 +47,116 @@ const DISCOVER_PRIMARY_KEY_QUERY: &str = r#"
...
+    JOIN pg_class c ON c.oid = a.attrelid
+    JOIN pg_namespace n ON n.oid = c.relnamespace
+    JOIN pg_type t ON t.oid = a.atttypid
+    WHERE n.nspname = $1
+      AND c.relname = $2
+      AND t.typname = 'vector'
+      AND a.attnum > 0
+      AND NOT a.attisdropped
+    ORDER BY a.attnum
+"#;
```
xiangjinwu, line 68: This two could just be a single query filtering on vector and expose 3 columns:
```
    SELECT
      a.atttypmod as atttypmod,
      format_type(a.atttypid, a.atttypmod) as formatted_type,
      a.attname as column_name
    FROM pg_attribute a
    JOIN pg_class c ON c.oid = a.attrelid
    JOIN pg_namespace n ON n.oid = c.relnamespace
    JOIN pg_type t ON t.oid = a.atttypid
    WHERE n.nspname = $1
      AND c.relname = $2
      AND t.typname = 'vector'
      AND a.attnum > 0
      AND NOT a.attisdropped
    ORDER BY a.attnum
```

xiangjinwu (approved):
Let's unify the two SQLs in this PR, and cleanup the duplicated connection definition and parsing in a dedicated follow-up PR.

src/connector/src/connector_common/postgres.rs
wcy-fdu: Will leave these configs for the next pull request.

PR #25187, opened 2026-03-27
fix(binder): handle generated column validation for INSERT with/without column list

### LINKED ISSUE #17628: bug: not compatible with PG when binding insert into table with generated column
In RisingWave, when inserting into a table with a generated column, RisingWave automatically removes the generated column during the insert column mapping. However, PostgreSQL does not do this.
https://github.com/risingwavelabs/risingwave/blob/9dabb6e5c158ab87603dfa31ec513f5e56811283/src/frontend/src/catalog/table_catalog.rs#L432-L436

RW behavior:
```SQL
dev=> create table t(v1 int, v2 int AS (v1+1), v3 int default 10);
CREATE_TABLE
dev=> insert into t values (0, 1);
INSERT 0 1
dev=> select * from t;
 v1 | v2 | v3 
----+----+----
  0 |  1 |  1
```

PG behavior: (PostgreSQL 16.2)
```SQL
postgres=# create table t(v1 int, v2 int GENERATED ALWAYS AS (v1+1) STORED, v3 int default 10);
CREATE TABLE
postgres=# insert into t values (0, 1);
ERROR:  cannot insert a non-DEFAULT value into column "v2"
DETAIL:  Column "v2" is a generated column.
postgres=# insert into t(v1, v3) values (0, 1);
INSERT 0 1
postgres=# insert into t values (0, 1, 2);
ERROR:  cannot insert a non-DEFAULT value into column "v2"
DETAIL:  Column "v2" is a generated column.
```

Standing-Man:

- Closes #17628

Handle validation of inserts into generated columns in two cases:
1. When column names are explicitly specified:
   - Directly check whether any generated column is included.
2. When column names are not specified:
   - Validate based on column order by ensuring the number of inserted values
     does not exceed the position of the first generated column.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/src/binder/insert.rs
+++ b/src/frontend/src/binder/insert.rs
@@ -298,7 +298,6 @@ impl Binder {
             }
         };
 
-        let has_user_specified_columns = !cols_to_insert_by_user.is_empty();
         let num_target_cols = if has_user_specified_columns {
             cols_to_insert_by_user.len()
         } else {


Standing-Man:
Hi @chenzl25, Could you help me take a look when you have time?

chenzl25:
I found one **blocking issue** in this PR.

**Problem**
The new `check_generated_insert_violation(Some(bound_query.schema().len()))` / `check_generated_insert_violation(Some(values_len))` logic in `src/frontend/src/binder/insert.rs` uses:
- `number of bound input columns`
- compared against
- `the visible index of the first generated column`

to decide whether the user is trying to insert into a generated column.

That only works when the user does **not** specify an explicit column list and is inserting by the table’s visible column order. Once the user provides an explicit column list and skips the generated column, this becomes a false positive.

**Counterexample**
```sql
create table t(v1 int, v2 int as (v1 + 1), v3 int default 10);
insert into t (v1, v3) values (1, 2);
```
This should be valid, because the user is not assigning any value to `v2` at all.

But with the current logic:
- `values_len == 2`
- `first_generated_column()` returns `(1, "v2")`
- `2 > 1`

so the statement is incorrectly rejected as if the user were inserting a non-`DEFAULT` value into the generated column.

**Root cause**
The PR is treating “the source has N columns” as equivalent to “the user is filling the first N visible table columns”. That is not true when an explicit insert column list is present.

**Suggested fix**
- If the user specified `INSERT INTO t ( ... )`, the earlier **name-based** generated-column check is already the right one.
- The new count-based check should only be applied to the path where the user did **not** specify a column list.

**Test gap**
The new e2e coverage adds several error cases, but it is missing the corresponding valid case where an explicit column list skips the generated column. I suggest adding at least:
```sql
create table t(v1 int, v2 int as (v1+1), v3 int default 10);
insert into t (v1, v3) values (1, 2);
select * from t;
```

Aside from this, I agree with the overall direction of the fix. The main issue is that the new check is currently too broad.

Standing-Man:
> **Root cause** The PR is treating “the source has N columns” as equivalent to “the user is filling the first N visible table columns”. That is not true when an explicit insert column list is present.

Thanks for your review and suggestions! I’ve now separated the checks for explicit and implicit column names. Please let me know if there’s anything I should update.

PR #25188, opened 2026-03-27
feat(stream): support predicate pushdown for snapshot backfill

chenzl25:

- Support predicate pushdown for snapshot backfill, so the pk predicate could be used in snapshot backfill to speedup the backfill process.
- In the future we can extend this optimization to index selection for streaming queries and arrangement backfill and no shuffle backfill.
-  For cases where users want to use different predicates to build multiple downstream mvs, like `create materialized view mv as select * from t where pk_prefix = 'xxx'` and `create materialized view mv as select * from t where pk_range > 1 and pk_range < 10`  It can improve the performance quite a lot.
- https://github.com/risingwavelabs/risingwave/issues/25189

## Checklist

## Documentation

<details>

</details>


src/frontend/src/optimizer/plan_node/logical_scan.rs
```diff
@@ -633,6 +634,32 @@ impl ToStream for LogicalScan {
...
             .into())
         } else {
+            if ctx.backfill_type() == BackfillType::SnapshotBackfill {
+                let (scan_ranges, _residual) = self.predicate().clone().split_to_scan_ranges(
+                    self.table(),
+                    self.base.ctx().session_ctx().config().max_split_range_gap() as u64,
+                )?;
+
+                if scan_ranges.len() == 1
+                    && scan_ranges[0].has_eq_conds()
```
hzxa21: Do we plan to support range filter on pk prefix as well, like we did in row seq scan?

src/frontend/planner_test/tests/testdata/output/backfill.yaml
```diff
@@ -222,6 +222,30 @@
       "public.fact4" -> "public.fact1";
       "public.fact4" -> "public.fact2";
     }
+- name: snapshot backfill prefix equality pushdown
```
hzxa21, line 225: I think we can add e2e test in e2e_test/snapshot_backfill to verify the result as well.

src/frontend/src/optimizer/plan_node/logical_scan.rs
chenzl25: Yes. I will support it in this PR directly to avoid unnecessary breaking change in the future.

src/frontend/planner_test/tests/testdata/output/backfill.yaml
chenzl25: Done

chenzl25:
* **#25221** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25221?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#25207** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25207?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#25188** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25188?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25188?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

chenzl25:
@wenym1 I just implement the predicate pushdown for snapshot backfill. Please take a look if you have time, thanks.

src/stream/src/task/actor_manager.rs
```diff
@@ -140,6 +232,7 @@ impl StreamActorManager {
         .await?;
 
         let table_desc: &StorageTableDesc = node.get_table_desc()?;
+        let pk_scan_range = Self::decode_pk_scan_range(node, table_desc)?;
```
wenym1: can move this logic to inside `SnapshotBackfillExecutor::new`

chenzl25: Done

chenzl25:
@wenym1 ready for the next round review

chenzl25:
@wenym1 ready for review, PTAL

src/stream/src/executor/backfill/snapshot_backfill/executor.rs
```diff
@@ -77,14 +84,82 @@ pub struct SnapshotBackfillExecutor<S: StateStore> {
     metrics: Arc<StreamingMetrics>,
 
     snapshot_epoch: Option<u64>,
+    /// (`eq_prefix`, `range_bounds`) for pk scan range pushdown.
+    pk_scan_range: Option<PkScanRange>,
```
wenym1: I think we can always have some `pk_scan_range` instead of being optional. For the case of no `pk_scan_range` specified, we can initialize it as the default `PkScanRange::full()`

src/stream/src/executor/backfill/snapshot_backfill/executor.rs
```diff
@@ -77,14 +84,82 @@ pub struct SnapshotBackfillExecutor<S: StateStore> {
...
 }
 
 impl<S: StateStore> SnapshotBackfillExecutor<S> {
+    /// Decode a scan range (eq prefix + range bounds) from the scan range proto.
```
wenym1: can we possibly reuse the existing `build_scan_ranges_from_pb` like the row seq scan executor?

src/storage/src/table/batch_table/mod.rs
```diff
@@ -975,6 +975,166 @@ impl<S: StateStore, SD: ValueRowSerde> BatchTableInner<S, SD> {
         Ok(iter.map_ok(|(_, row)| row))
     }
 
+    pub async fn batch_iter_vnode_with_pk_prefix(
```
wenym1: Is this method used in this PR? I didn't see any usage. If not used, can exclude from this PR.

src/storage/src/table/batch_table/mod.rs
```diff
@@ -975,6 +975,166 @@ impl<S: StateStore, SD: ValueRowSerde> BatchTableInner<S, SD> {
...
+            rebuild_interval,
+        )
+        .await?;
+        let iter = self.iter_stream_from_state_store_iter::<(), _>(iter, pk_serializer);
+        Ok(iter.map_ok(|(_, row)| row))
+    }
+
+    /// Like `batch_iter_vnode_with_pk_prefix` but additionally supports range bounds
+    /// on the column after the equality prefix.
+    pub async fn batch_iter_vnode_with_pk_range(
```
wenym1, line 1157: It seems that `batch_iter_vnode` is only a special case of `batch_iter_vnode_with_pk_range` that has full scan range. If so, instead of re-impl a new similar method for `BatchTable`, can we possibly only modify the body of current `batch_iter_vnode` to let it support accepting the scan range, and we can rename it to `batch_iter_vnode_with_pk_range`? The `batch_iter_vnode` can have a new method body that simply call `batch_iter_vnode_with_pk_range` with a full scan range.

wenym1 (commented):
It looks like many optimization can be shared with batch table scan. For example, besides the current predicate push down on all vnodes, we can further limit to scan from a single vnode if the predicate falls onto a single vnode.

Can we possibly refactor to extract some shared logic from batch table scan to some common utils methods? After the refactor, we can use AI to tentatively verify whether the logic of the existing batch table scan is equivalent before and after the refactor, and then in snapshot backfill we can reuse these extracted common utils methods whose correctness has been verified by the existing batch table scan logic. In this way, more existing optimization implemented for batch table scan can be ported to snapshot backfill with less efforts.

chenzl25:
> It looks like many optimization can be shared with batch table scan. For example, besides the current predicate push down on all vnodes, we can further limit to scan from a single vnode if the predicate falls onto a single vnode.
> 
> Can we possibly refactor to extract some shared logic from batch table scan to some common utils methods? After the refactor, we can use AI to tentatively verify whether the logic of the existing batch table scan is equivalent before and after the refactor, and then in snapshot backfill we can reuse these extracted common utils methods whose correctness has been verified by the existing batch table scan logic. In this way, more existing optimization implemented for batch table scan can be ported to snapshot backfill with less efforts.

Thanks for the valuable insight. The refactoring is done, PTAL if you have time.

src/stream/src/executor/backfill/snapshot_backfill/executor.rs
```diff
@@ -920,7 +941,12 @@ async fn make_snapshot_stream(
     rate_limit: RateLimit,
     chunk_size: usize,
     snapshot_rebuild_interval: Duration,
-) -> StreamExecutorResult<VnodeStream<impl super::vnode_stream::ChangeLogRowStream>> {
+    pk_scan_range: &PkScanRange,
```
wenym1, line 943: Most changes on this `make_snapshot_stream` can be avoided. The return type can still be `VnodeStream<impl super::vnode_stream::ChangeLogRowStream>`, because after we unify to call `batch_iter_vnode_with_pk_range`, the return stream is static and fixed, and in the method body we don't need to have async block with `.boxed()`, but instead only change the previous call to `batch_iter_vnode` to `batch_iter_vnode_with_pk_range`

src/storage/src/table/batch_table/mod.rs
```diff
@@ -60,6 +61,167 @@ use crate::store::{
...
+impl PkScanRange {
+    /// Decode a scan range from the protobuf representation and the primary-key column types.
+    pub fn new(scan_range: PbScanRange, pk_types: Vec<DataType>) -> StorageResult<Self> {
+        let mut index = 0;
+        let pk_prefix = OwnedRow::new(
+            scan_range
+                .eq_conds
+                .iter()
+                .map(|v| {
+                    let ty = pk_types.get(index).unwrap();
```
wenym1: Semantic difference found by codex: we used to return error on this case, but not we panic with `unwrap`.

src/storage/src/table/batch_table/mod.rs
```diff
@@ -60,6 +61,167 @@ use crate::store::{
...
+        }
+
+        let build_bound =
+            |bound: &scan_range::Bound, mut index| -> StorageResult<Bound<OwnedRow>> {
+                let range_bound = OwnedRow::new(
+                    bound
+                        .value
+                        .iter()
+                        .map(|v| {
+                            let ty = pk_types.get(index).unwrap();
```
wenym1: ditto

src/storage/src/table/batch_table/mod.rs
```diff
@@ -940,12 +1102,89 @@ impl<S: StateStore, SD: ValueRowSerde> BatchTableInner<S, SD> {
...
+        // For DESC PK columns, the planner's lower/upper bounds are in logical (ascending)
+        // order, but storage key ordering is reversed. Swap bounds so that the serialized
+        // start_key < end_key in storage key space, matching what batch's
+        // `convert_to_range_bounds` does.
+        //
+        // If the equality prefix already covers the full PK, there is no "next column" whose
+        // order type can be consulted. In that case, keep the bounds as-is and let
+        // `serialize_pk_bound` interpret them against the full prefix directly.
+        let next_col_idx = pk_prefix.len();
+        let (logical_start, logical_end) =
```
wenym1: This reimplements the ASC/DESC swap locally, but it does not carry over the NULL-handling from `PkScanRange::convert_to_range_bounds` (the logic that excludes `NULL` on one-sided ranges when nulls sort on the bounded side). For nullable index columns, snapshot backfill can therefore diverge from batch semantics for predicates like `col > 5` or `col < 5`. Can we reuse `convert_to_range_bounds` here so the two paths stay identical?

wenym1: From codex

src/frontend/src/optimizer/plan_node/logical_scan.rs
```diff
@@ -633,6 +633,29 @@ impl ToStream for LogicalScan {
             )
             .into())
         } else {
+            if ctx.backfill_type() == BackfillType::SnapshotBackfill {
+                let (scan_ranges, _residual) = self.predicate().clone().split_to_scan_ranges(
+                    self.table(),
+                    self.base.ctx().session_ctx().config().max_split_range_gap() as u64,
+                )?;
+
+                if scan_ranges.len() == 1 && !scan_ranges[0].is_full_table_scan() {
```
wenym1, line 642: This fast path bypasses `LogicalFilter::to_stream`, which is where we currently reject unsupported predicates containing `now()`. As a result, a snapshot-backfill scan with a pushdown-able PK predicate plus an invalid `now()` predicate can now build a direct `StreamFilter` here instead of raising the frontend error. Can we preserve that validation before taking this branch, or lower through `LogicalFilter` so the semantics stay consistent?

wenym1: From codex

src/stream/src/executor/backfill/snapshot_backfill/executor.rs
chenzl25: Resolved.

wenym1 (approved):
LGTM!

PR #25196, opened 2026-03-29
fix(binder): adjust the position of default columns

### LINKED ISSUE #17626: bug: type mismatch when insert.
```SQL
dev=> create table t(v1 int, v2 int as v1 + v1, v3 int default 10, v4 varchar);
CREATE_TABLE
dev=> insert into t(v3) values (1);

  1: gRPC request to batch service failed: Internal error
  2: broken fifo_channel
```

CN log 
```
thread 'rw-batch' panicked at src/dml/src/table.rs:127:10:
table source write txn_msg schema check failed: "column type mismatched at position 2: expected Some(Varchar), found Some(\"Int32\")"
stack backtrace:
   0: rust_begin_unwind
             at /rustc/4a0cc881dcc4d800f10672747f61a94377ff6662/library/std/src/panicking.rs:645:5
   1: core::panicking::panic_fmt
             at /rustc/4a0cc881dcc4d800f10672747f61a94377ff6662/library/core/src/panicking.rs:72:14
   2: core::result::unwrap_failed
             at /rustc/4a0cc881dcc4d800f10672747f61a94377ff6662/library/core/src/result.rs:1654:5
   3: core::result::Result<T,E>::expect
             at /rustc/4a0cc881dcc4d800f10672747f61a94377ff6662/library/core/src/result.rs:1034:23
   4: risingwave_dml::table::TableDmlHandle::check_chunk_schema
             at ./src/dml/src/table.rs:121:9
   5: risingwave_batch::executor::insert::InsertExecutor::do_execute::{{closure}}::{{closure}}::{{closure}}
             at ./src/batch/src/executor/insert.rs:167:13
   6: risingwave_batch::executor::insert::InsertExecutor::do_execute::{{closure}}
             at ./src/batch/src/executor/insert.rs:189:47
.......
```

additional info: this case is ok
```
dev=> create table t(v1 int, v2 int as v1 + v1, v3 int default 10, v4 int);
CREATE_TABLE
dev=> insert into t(v3) values (1);
INSERT 0 1
```

Standing-Man:

- Closes #17626

In `BoundInsert`, the `default_columns` field is indexed according to the table catalog. However, the current implementation adjusts default column indices based on `cols_to_insert_in_table`, which only contains visible and non-generated columns.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/e2e_test/ddl/table/default_column.slt.part
+++ b/e2e_test/ddl/table/default_column.slt.part
@@ -13,7 +13,7 @@ SELECT * FROM t;
 ----
 1 2 10 4
 
-# Only provide v3; v1/v2 should be NULL in this row, and let v4 use its default. 
+# Only provide v3; v1/v2 should be NULL in this row, and let v4 use its default.
 statement ok
 INSERT INTO t(v3) VALUES (1);
 


Standing-Man:
Hi @chenzl25, could you take a look when you have time? 😂 

Also, I’ve fixed the issue, but I’m still trying to understand why a `FLUSH` statement is required in the SLT test. Without it, the SELECT query returns empty results. Please let me know if I’m missing something.

chenzl25:
> Hi @chenzl25, could you take a look when you have time? 😂
> 
> Also, I’ve fixed the issue, but I’m still trying to understand why a `FLUSH` statement is required in the SLT test. Without it, the SELECT query returns empty results. Please let me know if I’m missing something.

It is because by default data only committed into storage after checkpoint. RisingWave would checkpoint data every second. Flush command would force a checkpoint immediately so that the data could be selected in SLT.

Standing-Man:
> It is because by default data only committed into storage after checkpoint. RisingWave would checkpoint data every second. Flush command would force a checkpoint immediately so that the data could be selected in SLT.

Thanks for the clarification! Please let me know if there’s anything else I should update.

Standing-Man:
Hi @chenzl25, just a gentle bump on this PR when you have time.

PR #25197, opened 2026-03-30
feat(meta): add periodic stale compaction group purge

Li0k:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwave-labs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

- Add a periodic, version-only cleanup for compaction groups (no catalog/table-id reconciliation).
- Periodically detect **dynamic** compaction groups that have **zero member tables** in the latest Hummock version (`state_table_info`) and destroy those groups (emit `GroupDestroy` deltas).
  - "Empty" is defined as having no member tables in `state_table_info` (does not require having no SSTs).
  - Safety relies on the invariant that `state_table_info` in the latest version fully reflects tables that can read from the version.
- Never destroy a group referenced as `parent_group_id` by another group, to avoid dangling parent references / violating version invariants used by group split scheduling.
- Purge stale compaction-group configs that are not referenced by the latest version, and avoid meta-store writes on no-op rounds (when there is nothing to purge).
- Keep the existing meta config `periodic_purge_stale_compaction_group_interval_sec` for this behavior. The production default is `3600` seconds (1 hour); tests keep it disabled by default unless explicitly enabled.
- Disable this periodic cleanup path in `compaction_deterministic_test`, consistent with other periodic background metadata-changing tasks.
- Add focused unit tests for the version-only cleanup behavior (empty group with/without SSTs, mixed empty/non-empty/static, stale config purge, parent lifecycle, and no-op).

- No related issue.

## Checklist

## Documentation

<details>

No user-facing change. This adds an internal meta timer that periodically destroys empty dynamic compaction groups from the latest Hummock version and purges stale compaction-group configs.

</details>


src/meta/src/hummock/manager/timer_task.rs
```diff
@@ -484,6 +512,76 @@ impl HummockManager {
...
+            Err(err) => {
+                warn!(
+                    error = %err.as_report(),
+                    "failed to list fragment state tables for stale compaction group purge"
+                );
+                return;
+            }
+        };
+
+        if let Err(err) = self.purge(&valid_table_ids).await {
```
zwang28: This purge is unsafe.
pls find the comment of purge method: The caller should ensure the set of valid table ids remains stable during `purge`. i.e. no new valid table id should be added during purge, otherwise it will be incorrectly purged from hummock.

Li0k: Good catch.

The initial periodic path indeed had a race: if a new valid table id is added while we build `valid_table_ids`, it could be treated as stale and unregistered.

To make this runtime cleanup best-effort and conservative, this PR now:
- Skips the whole round when there are any creating streaming jobs, or any pending dropped tables (create/drop transition windows).
- Re-checks the two conditions again right before calling `purge(...)` to narrow the remaining race window.
- Disables the periodic purge in `compaction_deterministic_test`.
- Adds unit tests for the two safety gates and documents the assumptions.

Latest commit addressing the extra re-check: b5413915e1.

src/meta/src/hummock/manager/compaction/compaction_group_manager.rs
```diff
@@ -371,6 +370,128 @@ impl HummockManager {
...
+            &self.metrics,
+            &self.env.opts,
+        );
+        let mut new_version_delta = version.new_delta();
+
+        for &(group_id, _) in &groups_to_destroy {
+            new_version_delta
+                .group_deltas
+                .entry(group_id)
+                .or_default()
```
zwang28, line 464: minor: the entry must be absent.

src/meta/src/hummock/manager/compaction/compaction_group_manager.rs
```diff
@@ -371,6 +370,128 @@ impl HummockManager {
...
+        }
+
+        new_version_delta.pre_apply();
+
+        let mut compaction_group_manager = self.compaction_group_manager.write().await;
+        let mut compaction_groups_txn = compaction_group_manager.start_compaction_groups_txn();
+        compaction_groups_txn.purge(HashSet::from_iter(get_compaction_group_ids(
+            version.latest_version(),
+        )));
+        commit_multi_var!(self.meta_store_ref(), version, compaction_groups_txn)?;
```
zwang28, line 476: The 2 lock can be explictly dropped after commit.

src/meta/src/hummock/manager/compaction/compaction_group_manager.rs
```diff
@@ -371,6 +370,128 @@ impl HummockManager {
...
+            })
+            .collect();
+
+        if groups_to_destroy.is_empty() {
+            // No groups to destroy — just purge stale configs if any.
+            let mut compaction_group_manager = self.compaction_group_manager.write().await;
+            let mut compaction_groups_txn = compaction_group_manager.start_compaction_groups_txn();
+            let existing_groups =
+                HashSet::from_iter(get_compaction_group_ids(&versioning.current_version));
+            if compaction_groups_txn
```
zwang28, line 436: The version lock can be dropped here.

zwang28 (approved):
rest LGTM

PR #25202, opened 2026-03-30
feat(iceberg): support order key

xxhZs:
## Summary

This PR adds `order_key` support for Iceberg tables and ensures the configured sort order is carried through both file writing and Iceberg compaction metadata.

Main changes:
- parse and validate `order_key` for Iceberg table creation
- reject invalid expressions and unsupported/system columns early in frontend/connector validation
- build Iceberg `SortOrder` from the configured key when creating the table
- write `sort_order_id` into generated Iceberg data files so the sort order is visible in metadata
- preserve the same sort-order metadata on compaction outputs
- add e2e coverage for Iceberg engine, append-only Iceberg engine, and Iceberg sink

In practice, this means a table like:

```sql
WITH (order_key = 'v1 desc nulls last, id asc') ENGINE = ICEBERG
```

will materialize an Iceberg sort order in table metadata, and the produced data files can be observed from `rw_iceberg_files.sort_order_id`.

## Validation

Validated locally with Iceberg engine + storage catalog.

Test case:
- `order_key = 'v1 desc nulls last, v2 asc nulls first, id desc'`
- inserted rows in 3 batches with `FLUSH` after each batch:
  - `(1, 100, 2, 'a'), (2, 100, 1, 'b')`
  - `(3, 100, 1, 'c'), (4, 90, 2, 'd')`
  - `(5, 90, 2, 'e'), (6, 90, 1, 'f')`

Before compaction:
- `rw_iceberg_files` showed `6` data files
- all data files had non-null `sort_order_id = 1`

After starting a dedicated iceberg compactor and running:

```sql
VACUUM FULL t_order_multi;
```

After compaction:
- `rw_iceberg_files` showed `1` data file
- the compacted file still had non-null `sort_order_id = 1`

The compacted parquet file was read directly and its physical row order was:

```text
(3, 100, 1, 'c')
(2, 100, 1, 'b')
(1, 100, 2, 'a')
(6, 90, 1, 'f')
(5, 90, 2, 'e')
(4, 90, 2, 'd')
```

This matches the configured multi-column order key:
- `v1 desc`
- `v2 asc`
- `id desc`

## Dependency

The compaction-side ordering behavior depends on the upstream change in:
- [nimtable/iceberg-compaction#124](https://github.com/nimtable/iceberg-compaction/pull/124)

This PR should wait for that upstream PR to be merged.

### DIFF
--- a/src/frontend/src/handler/create_table.rs
+++ b/src/frontend/src/handler/create_table.rs
@@ -2293,7 +2293,7 @@ pub async fn create_iceberg_engine_table(
         .map(|v| v.to_owned());
     if let Some(order_key) = &order_key {
         validate_order_key_columns(order_key, table.columns().iter().map(|col| col.name()))
-            .map_err(|err| ErrorCode::InvalidInputSyntax(err.to_string()))?;
+            .map_err(|err| ErrorCode::InvalidInputSyntax(err.to_report_string()))?;
 
         sink_with.insert(ORDER_KEY.to_owned(), order_key.to_owned());
 


src/connector/src/sink/iceberg/mod.rs
```diff
@@ -1920,9 +1961,12 @@ impl SinkWriter for IcebergSinkWriter {
                 inner.uncommitted_write_bytes = 0;
                 let format_version = inner.table.metadata().format_version();
                 let partition_type = inner.table.metadata().default_partition_type();
+                let table_sort_order_id = table_sort_order_id(inner.table.metadata());
                 let data_files = result
                     .into_iter()
                     .map(|f| {
+                        let f = apply_sort_order_id_to_data_file(f, table_sort_order_id)
```
chenzl25: I think we can't apply sort order key to the parquet file written by iceberg sink because there is no order guarantee in the streaming sink.

chenzl25:
@xxhZs any updates?

xxhZs:
can review againt @chenzl25

PR #25203, opened 2026-03-30
feat(sink): support sink skip error and record error in table

wenym1:

This PR adds the sink-side feature work for persisting skipped sink rows into an internal error table when `skip_error = 'true'`.

Stack context:

- depends on #25885, which adds the log-store plumbing to carry reported sink error rows through truncate/flush
- depends on #25883, which fixes the async truncate ordering issue for non-decoupled sinks

On top of those prerequisites, this PR:

- extends sink planning/protobuf so a sink can own an internal error table
- defines the sink error table schema and catalog inference, with fixed metadata columns followed by visible sink columns
- makes the sink executor buffer reported rows by epoch and persist them into the error table before checkpoint barriers are yielded
- updates auto schema change rewriting so the sink error table schema stays aligned with the sink schema
- adds deterministic simulation coverage for sink error table behavior in both decoupled and non-decoupled modes, including schema change
- includes regression coverage around async truncate behavior and internal-table discovery

<details>
<summary><b>Design supplement</b></summary>

# Sink Error Table Design Supplement

## Trait Changes on `LogReader` and `LogWriter`

The sink runtime needs to carry skipped row-level errors from the sink delivery path back to the sink executor in a way that stays consistent with truncate and barrier progress.

The trait changes are:

- `LogReader::truncate(offset, error_rows)`
- `LogWriter::flush_current_epoch(...) -> FlushCurrentEpochResult { post_flush, reported_error_rows }`

`ReportedSinkErrorRow` is the row-level unit transported through this protocol. Each item contains:

- `epoch`: the read epoch the row belongs to
- `op`: the original row operation
- `row`: the original row payload
- `extra_info`: structured sink-specific error context

With this design, reported error rows are carried through the same truncate / flush path that already defines sink progress, instead of introducing a separate callback channel.

## How `SinkExecutor` Interacts with `LogWriter`

At each barrier, the sink executor calls `log_writer.flush_current_epoch(...)`.

The flush result returns:

- `post_flush`: the existing post-yield callback
- `reported_error_rows`: skipped rows that are ready to be surfaced at this point in progress

The sink executor then:

1. adds `reported_error_rows` into an internal pending queue ordered by epoch
2. drains rows with `reported_row.epoch <= barrier.epoch.curr`
3. converts each drained row into an internal error-table row
4. commits the error table for the barrier epoch
5. yields the barrier
6. runs `post_flush.post_yield_barrier()`

The error table commit must happen before yielding the barrier so that the error table follows the same state-table barrier contract as other executor-managed state.

## Error Table Schema

Each sink owns an internal error table.

The schema is:

- `sink_error_epoch`
- `sink_error_row_id`
- `sink_error_vnode`
- `sink_error_row_op`
- `sink_error_read_epoch`
- `sink_error_extra_info`
- visible sink columns

Column semantics:

- `sink_error_epoch`
  - the barrier epoch used when this error row is persisted into the internal table
  - this is the commit epoch of the error-table write, not necessarily the epoch when the original row was read
- `sink_error_row_id`
  - a per-`sink_error_epoch` sequence number assigned by the sink executor when draining ready rows for that barrier
  - together with vnode and write epoch, it makes each persisted error row unique
- `sink_error_vnode`
  - the vnode chosen for distributing the error-table row
  - this is an internal distribution column rather than source-row business data
- `sink_error_row_op`
  - the original stream operation of the skipped row
  - stored so queries can distinguish insert / delete / update-derived records
- `sink_error_read_epoch`
  - the read epoch attached to the reported sink error row when it comes back from the log-store layer
  - this identifies when the original row was read by the sink runtime
- `sink_error_extra_info`
  - structured sink-specific diagnostic information
  - intended for fields such as connector error details, raw payload metadata, or column-level decode context
- visible sink columns
  - the original visible sink payload columns copied from the skipped row
  - hidden/internal sink columns are not persisted into this part of the schema

The visible sink columns are appended after the metadata prefix so the original sink payload can be queried directly.

Primary key:

- `(sink_error_epoch, sink_error_row_id, sink_error_vnode)`

Distribution key:

- `sink_error_vnode`

## Schema Change Handling

The error table schema is derived from the sink schema, so it must be rewritten when sink auto schema refresh changes the visible sink columns.

The rewrite rule is:

- keep the fixed metadata prefix unchanged
- remove dropped visible sink columns
- append newly added visible sink columns
- rebuild `value_indices` to match the rewritten column layout

This rewrite is applied both to:

- the sink fragment definition
- the persisted catalog table definition

That keeps the internal error table aligned with the sink schema after auto schema change.

### Querying the error table

The current design stores skipped rows in an internal table, but the user-facing query surface is still minimal.

One future direction is to expose a dedicated query interface instead of making users discover internal table names manually. Two possible shapes are:

- a table-valued function, for example:
  - `select * from rw_sink_errors('my_sink')`
  - `select * from rw_sink_errors(sink_id => 1001, from_epoch => 123456)`
- a system catalog relation that resolves sink identity to the internal error table automatically

I checked current ClickHouse documentation as a reference point. Their Kafka engine supports:

- `kafka_handle_error_mode = 'stream'`, which exposes `_error` and `_raw_message` virtual columns on the stream source
- `kafka_handle_error_mode = 'dead_letter_queue'`, which persists error-related data into `system.dead_letter_queue`

That suggests two reasonable directions for RisingWave:

- expose a dedicated read surface on top of the persisted error rows
- keep the internal storage table hidden and avoid forcing users to depend on generated internal table names

The table-valued-function shape above is a proposal for RisingWave, not existing ClickHouse syntax.

### Truncating the error table

The current design appends skipped rows but does not yet define a first-class retention or truncation interface.

One future direction is to support controlled deletion through SQL, but only in a monotonic form that matches storage behavior. For example:

- `delete from rw_sink_errors('my_sink') where sink_error_epoch < 123456`
- or equivalent deletion against the internal error table with strict predicate validation

The intended restriction is:

- only allow predicates that advance a lower bound on retained epochs
- reject arbitrary row-level deletes or non-monotonic predicates

This can then be implemented as a storage-level watermark on the error table id in Hummock, instead of a general-purpose row rewrite path. In other words:

- SQL delete is only the user-facing control plane
- the execution path becomes "advance retained epoch watermark for this error table"

That keeps the retention model simple and aligned with append-only error-table semantics.

</details>

## Validation

- `cargo test -p risingwave_connector test_truncate_barrier_reader -- --nocapture`
- `cargo test -p risingwave_stream test_in_memory_log_store -- --nocapture`
- `./risedev sit-test test_sink_error_table_nondecouple`
- `./risedev sit-test test_sink_error_table_decouple`
- `./risedev sit-test test_async_truncate_nondecouple_waits_for_checkpoint_truncate`

## Checklist

## Documentation

<details>

Sinks with `skip_error = 'true'` can now persist skipped rows into an internal error table, and that table stays aligned when the sink schema is refreshed automatically.

</details>

### DIFF
--- a/proto/stream_plan.proto
+++ b/proto/stream_plan.proto
@@ -377,6 +377,7 @@ message SinkNode {
   catalog.Table table = 2;
   SinkLogStoreType log_store_type = 3;
   optional uint32 rate_limit = 4;
+  optional catalog.Table error_table = 5;
 }
 
 message IcebergWithPkIndexWriterNode {
--- a/src/common/src/constants.rs
+++ b/src/common/src/constants.rs
@@ -94,3 +94,48 @@ pub mod log_store {
         });
     }
 }
+
+pub mod sink_error {
+    use std::sync::LazyLock;
+
+    use crate::types::DataType;
+    use crate::util::sort_util::OrderType;
+
+    pub const EPOCH_COLUMN_NAME: &str = "sink_error_epoch";
+    pub const ROW_ID_COLUMN_NAME: &str = "sink_error_row_id";
+    pub const VNODE_COLUMN_NAME: &str = "sink_error_vnode";
+    pub const ROW_OP_COLUMN_NAME: &str = "sink_error_row_op";
+    pub const READ_EPOCH_COLUMN_NAME: &str = "sink_error_read_epoch";
+    pub const EXTRA_INFO_COLUMN_NAME: &str = "sink_error_extra_info";
+
+    pub const EPOCH_COLUMN_TYPE: DataType = DataType::Int64;
+    pub const ROW_ID_COLUMN_TYPE: DataType = DataType::Int32;
+    pub const VNODE_COLUMN_TYPE: DataType = DataType::Int16;
+    pub const ROW_OP_COLUMN_TYPE: DataType = DataType::Int16;
+    pub const READ_EPOCH_COLUMN_TYPE: DataType = DataType::Int64;
+    pub const EXTRA_INFO_COLUMN_TYPE: DataType = DataType::Jsonb;
+
+    pub const PREDEFINED_COLUMNS: [(&str, DataType); 6] = [
+        (EPOCH_COLUMN_NAME, EPOCH_COLUMN_TYPE),
+        (ROW_ID_COLUMN_NAME, ROW_ID_COLUMN_TYPE),
+        (VNODE_COLUMN_NAME, VNODE_COLUMN_TYPE),
+        (ROW_OP_COLUMN_NAME, ROW_OP_COLUMN_TYPE),
+        (READ_EPOCH_COLUMN_NAME, READ_EPOCH_COLUMN_TYPE),
+        (EXTRA_INFO_COLUMN_NAME, EXTRA_INFO_COLUMN_TYPE),
+    ];
+
+    pub const EPOCH_COLUMN_INDEX: usize = 0;
+    pub const ROW_ID_COLUMN_INDEX: usize = 1;
+    pub const VNODE_COLUMN_INDEX: usize = 2;
+    pub const ROW_OP_COLUMN_INDEX: usize = 3;
+    pub const READ_EPOCH_COLUMN_INDEX: usize = 4;
+    pub const EXTRA_INFO_COLUMN_INDEX: usize = 5;
+
+    pub static PK_ORDERING: LazyLock<[OrderType; 3]> = LazyLock::new(|| {
+        [
+            OrderType::ascending(),
+            OrderType::ascending(),
+            OrderType::ascending(),
+        ]
+    });
+}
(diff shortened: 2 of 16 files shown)


wenym1:
* **#25203** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25203?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25203?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#25442** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25442?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#25440** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25440?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

Issue #25189, opened 2026-03-27
Tracking: Selective Backfill


chenzl25:
### Is your feature request related to a problem? Please describe.

Currently, backfill processes all historical data from the upstream. However, if the materialized view supports predicate pushdown, we can use a primary key prefix to skip unnecessary data.

[ ] snapshot backfill. https://github.com/risingwavelabs/risingwave/pull/25188
[ ] arrangement backfill.
[ ] no-shuffle backfill.
[ ] index selection for backfill predicate pushdown. https://github.com/risingwavelabs/risingwave/pull/25207

PR #25207, opened 2026-03-31
feat(optimizer): support index selection for seletive backfill

chenzl25:

## What's changed and what's intention?

- https://github.com/risingwavelabs/risingwave/issues/25189

This PR implements streaming index selection optimizations for snapshot backfill operations. The changes introduce two key optimizations:

**1. Covering Index Selection for Backfill**
- Adds a new `StreamingIndexSelectionRule` that selects the lowest-cost covering index during snapshot backfill
- Unlike batch index selection, this only considers covering indexes since streaming backfill cannot perform lookup joins
- The rule estimates costs and picks the most efficient index that covers all required columns

**2. IN Predicate Expansion**
- Expands `IN` predicates (e.g., `WHERE a IN (1, 2, 3)`) into a `LogicalUnion` of separate `LogicalScan` nodes
- Each branch gets a single equality predicate (e.g., `a = 1`, `a = 2`, `a = 3`) for better scan range optimization
- This enables more efficient primary key range scans and proper post-backfill data routing

**Implementation Details:**
- The optimization is triggered during `logical_rewrite_for_stream` for snapshot backfill operations
- Added `RewriteStreamContext` to pass backfill type information through the rewrite process
- Extended test coverage with new end-to-end tests for IN predicates and covering index scenarios
- Added planner tests to verify the generated stream plans produce the expected `StreamUnion` structures

The changes ensure that materialized view backfill operations can leverage indexes more effectively and handle IN predicates with optimal scan patterns.

## Checklist

## Documentation

<details>

Improved performance of materialized view backfill operations through automatic index selection and IN predicate optimization. When creating materialized views with filtered queries, RisingWave now automatically selects the most efficient covering index and optimizes IN predicates by splitting them into parallel scan operations. This results in faster backfill completion times, especially for queries with selective predicates on indexed columns.

</details>


chenzl25:
* **#25221** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25221?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#25207** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25207?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25207?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#25188** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25188?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

chenzl25:
@wenym1 This is a follow-up PR of backfill predicate pushdown.

chenzl25:
@wenym1 Could you please take a look at this PR?

src/frontend/src/optimizer/plan_node/logical_scan.rs
```diff
@@ -672,6 +673,24 @@ impl ToStream for LogicalScan {
...
+            && !self.predicate().always_true()
+            && self
+                .base
+                .ctx()
+                .session_ctx()
+                .config()
+                .enable_index_selection()
+            && let Some(rewritten) = StreamingIndexSelectionRule::rewrite(self)
+        {
+            return rewritten.logical_rewrite_for_stream(ctx);
```
wenym1, line 691: Can we add an assertion or throw error that if the rewritten is also a table scan, `rewritten.predicate().always_true()` must be true, and otherwise I am afraid we will keep calling `logical_rewrite_for_stream` on the same plan node

src/frontend/src/optimizer/rule/streaming_index_selection_rule.rs
```diff
@@ -0,0 +1,160 @@
...
+
+    /// Picks the lowest-cost covering index for a streaming scan.
+    fn select_covering_index(logical_scan: &LogicalScan) -> Option<LogicalScan> {
+        let indexes = logical_scan.table_indexes();
+        if indexes.is_empty() {
+            return None;
+        }
+
+        let rule = IndexSelectionRule {};
+        let primary_table_row_size = TableScanIoEstimator::estimate_row_size(logical_scan);
```
wenym1, line 70: I see we have a similar logic of calculating the `primary_cost` in batch query index selection. Can we possibly extract it as a shared method of logical scan?

src/frontend/src/optimizer/rule/streaming_index_selection_rule.rs
```diff
@@ -0,0 +1,160 @@
...
+        }
+
+        best_scan
+    }
+
+    /// Expands an IN predicate into a `LogicalUnion` of `LogicalScan`s.
+    ///
+    /// Each branch gets a single-value equality predicate (e.g., `a = 1`) plus residual
+    /// conditions. Only applies when all scan ranges are non-trivial.
+    fn try_expand_in(logical_scan: &LogicalScan) -> Option<PlanRef> {
```
wenym1, line 104: Does it mean that, for example if we have `IN (1, 3, 5)`, after backfill finish, the upstream MV data will be dispatched to the current MV for 3 exactly same copies for each 3 branches of the union, and then we do the filter in each branch? This extra cost seems unnecessary. Can we possibly support multiple scan ranges for stream scan?

src/frontend/src/optimizer/plan_node/logical_scan.rs
chenzl25: We avoid repeatedly calling index selection by checking the table's index catalog. Therefore, even if a table scan occurs, the next call returns immediately because there is no index to select.

src/frontend/src/optimizer/rule/streaming_index_selection_rule.rs
chenzl25: I think the current shared method is proper. Once we extract it as a shared method of logical scan, the logical scan actually could be an index which is not a primary table.

src/frontend/src/optimizer/rule/streaming_index_selection_rule.rs
chenzl25: Indeed, there is some overhead after backfilling, as we will end up with three branches. Not sure it would introduce some breaking changes to the backfill executor if we want to support multiple scan ranges, especially we have persisted the progress state for singe range scan already.

wenym1: Theoretically, if the scan ranges don't overlap, they can be sorted in pk order, and then after we resume the progress, we can know which scan range we are currently in, and then continue. But having multiple vnodes may be tricky, because different vnode can be in different scan range after scale. Maybe we can keep the current impl first, and optimize it later.

wenym1 (approved):
LGTM

PR #25212, opened 2026-03-31
feat(frontend): support secret references as function call arguments

tabVersion:
## Summary

Support `SELECT udf(data, secret my_key) FROM stream` syntax by introducing secret references as first-class function arguments for **user-defined functions only**.

This PR adds `ExprImpl::SecretRef` so frontend can bind a secret reference, serialize only `secret_id` into the plan, and let compute resolve the value at runtime via `LocalSecretManager`.

Current scope:
- Works with **external scalar UDFs** (for example Python / JavaScript)
- Works with **SQL UDFs** after inline expansion
- **Does not** allow secret refs in built-in functions
- **Does not** support named function arguments yet; they now return a user-facing error instead of panicking

## Design

`ExprImpl::SecretRef` follows the same general pattern as other special expression nodes:
- Frontend parser/binder recognizes `SECRET name` in function argument position
- Frontend expression tree uses `ExprImpl::SecretRef`
- Proto adds `SecretRefNode { secret_id, ref_as }` under `ExprNode.rex_node`
- Backend resolves the referenced secret to a literal at expression build time via `LocalSecretManager::fill_secret()`
- Display keeps a readable secret name for EXPLAIN/debugging without serializing the value itself

Key properties:
- Secret values never appear in protobuf plans or EXPLAIN output
- `PartialEq` / `Hash` exclude `secret_name` because it is display-only and not preserved across proto round-trip
- Secret dependencies are tracked through binder so streaming jobs can block `DROP SECRET` when referenced
- Secret refs are marked impure to avoid unsafe constant-folding assumptions

### Supported
- `SELECT echo_with_secret('data', secret test_secret);`
- `SELECT sql_echo('sql:', secret test_secret);`
- `CREATE MATERIALIZED VIEW mv AS SELECT echo_with_secret(data, secret test_secret) FROM t;`

### Rejected
- Built-in calls such as `concat('x', secret test_secret)`
- Secret refs outside function-argument context
- Named function arguments such as `func(arg => secret test_secret)`

### Error handling improvements in this PR
- Built-in/non-UDF usage now reports `secret reference is only allowed in user-defined function arguments` before secret lookup
- Named function arguments now return `named function arguments are not supported yet` instead of hitting `todo!()`

## Changes

| Layer | Files | Change |
|-------|-------|--------|
| Proto | `proto/expr.proto` | Add `SecretRefNode` and `secret_ref` rex node |
| Parser | `src/sqlparser/src/ast/mod.rs`, `src/sqlparser/src/parser.rs` | Parse `SECRET name` in function args |
| Frontend expr | `src/frontend/src/expr/secret_ref.rs`, `src/frontend/src/expr/mod.rs` | Add `ExprImpl::SecretRef` |
| Binder | `src/frontend/src/binder/expr/function/mod.rs`, `src/frontend/src/binder/mod.rs` | Bind secret refs, track `included_secrets`, reject invalid contexts clearly |
| Runtime | `src/expr/core/src/expr/build.rs` | Resolve `SecretRefNode` via `LocalSecretManager` |
| Optimizer / expr infra | `src/frontend/src/expr/expr_visitor.rs`, `src/frontend/src/expr/expr_rewriter.rs`, `src/frontend/src/expr/expr_mutator.rs`, `src/frontend/src/expr/pure.rs` | Handle `SecretRef` exhaustively |
| Streaming dependency tracking | `src/frontend/src/handler/create_mv.rs`, `src/frontend/src/handler/alter_mv.rs`, `src/frontend/src/handler/query.rs` | Thread secret dependencies into MV / prepared-statement flows |
| Tests | planner + e2e tests | Cover UDF success path, invalid built-in usage, dependency tracking, and named-arg failure |

## Test plan

- [x] Parser unit test for `SECRET ...` function args
- [x] Planner tests for invalid built-in usage / binder errors
- [x] E2E test coverage in `e2e_test/udf/tests/scalar/secret_ref.slt`
  - external UDF success path
  - SQL UDF success path
  - invalid built-in usage
  - missing secret in UDF path
  - MV dependency tracking
  - named-arg graceful error

## Notes

This PR intentionally keeps the product boundary narrow: secret references are a controlled capability for UDF arguments, not a general-purpose expression feature for all functions.

### DIFF
--- a/proto/expr.proto
+++ b/proto/expr.proto
@@ -684,7 +684,6 @@ message UserDefinedFunction {
   // additional options for javascript functions
   optional bool is_async = 13;
   optional bool is_batched = 14;
-  reserved 15;
 
   UdfExprVersion version = 1000;
 }
@@ -706,7 +705,6 @@ message UserDefinedFunctionMetadata {
   // additional options for javascript functions
   optional bool is_async = 14;
   optional bool is_batched = 15;
-  reserved 16;
 
   UdfExprVersion version = 1000;
 }
--- a/src/frontend/src/expr/mod.rs
+++ b/src/frontend/src/expr/mod.rs
@@ -1180,7 +1180,7 @@ mod tests {
                 "{:?}",
                 ExprDisplay {
                     expr: &expr,
-                    input_schema: &Schema::empty(),
+                    input_schema: Schema::empty(),
                 }
             ),
             "Secret(test_secret)"


proto/expr.proto
```diff
@@ -675,6 +684,7 @@ message UserDefinedFunction {
   // additional options for javascript functions
   optional bool is_async = 13;
   optional bool is_batched = 14;
+  reserved 15;
```
chenzl25: Why do we need this reserved 15?

proto/expr.proto
```diff
@@ -696,6 +706,7 @@ message UserDefinedFunctionMetadata {
   // additional options for javascript functions
   optional bool is_async = 14;
   optional bool is_batched = 15;
+  reserved 16;
```
chenzl25: Why do we need this reserved 16?

chenzl25 (approved):
Generally LGTM

proto/expr.proto
tabVersion: will remove, introduced due to an failed attempt.

tabVersion:
Issue found by codex:

```
- High: CREATE SINK AS ... still drops secret dependencies on the floor, so sinks created from queries like SELECT my_udf(col, secret api_key) won't block DROP SECRET and can be left with a dangling runtime dependency. In src/frontend/src/handler/create_sink.rs:278 only included_relations and included_udfs are captured, and the final dependency set at src/frontend/src/handler/create_sink.rs:449 never adds included_secrets, unlike the new MV path in src/frontend/src/handler/create_mv.rs:166 and src/frontend/src/handler/create_mv.rs:405. Please thread dependent_secrets through sink creation and include them in the catalog dependency set.

- High: ordinary CREATE VIEW has the same regression: views can now contain secret refs in UDF arguments, but create_view persists only relation dependencies. src/frontend/src/handler/create_view.rs:53 builds a batch plan and keeps only dependent_relations, then src/frontend/src/handler/create_view.rs:114 stores just those dependencies. That means DROP SECRET can succeed even while a view still references it, leaving the view invalid at query time. Please plumb secret dependencies through the batch-plan / view-creation path the same way this patch does for materialized views.
```

tabVersion:
> Issue found by codex:
> 
> ```
> - High: CREATE SINK AS ... still drops secret dependencies on the floor, so sinks created from queries like SELECT my_udf(col, secret api_key) won't block DROP SECRET and can be left with a dangling runtime dependency. In src/frontend/src/handler/create_sink.rs:278 only included_relations and included_udfs are captured, and the final dependency set at src/frontend/src/handler/create_sink.rs:449 never adds included_secrets, unlike the new MV path in src/frontend/src/handler/create_mv.rs:166 and src/frontend/src/handler/create_mv.rs:405. Please thread dependent_secrets through sink creation and include them in the catalog dependency set.
> 
> - High: ordinary CREATE VIEW has the same regression: views can now contain secret refs in UDF arguments, but create_view persists only relation dependencies. src/frontend/src/handler/create_view.rs:53 builds a batch plan and keeps only dependent_relations, then src/frontend/src/handler/create_view.rs:114 stores just those dependencies. That means DROP SECRET can succeed even while a view still references it, leaving the view invalid at query time. Please plumb secret dependencies through the batch-plan / view-creation path the same way this patch does for materialized views.
> ```

fixed in https://github.com/risingwavelabs/risingwave/pull/25291

PR #25157, opened 2026-03-23
refactor(meta): unify barrier completion flow to partial-graph manager

wenym1:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## What's changed and what's your intention?

This PR refactors meta barrier completion handling so that partial-graph barrier lifecycle management is owned by `PartialGraphManager` instead of being split across multiple ad hoc queues in checkpoint control and creating-job control.

Concretely:

- Introduce a reusable `BarrierItemCollector` utility in `src/meta/src/barrier/utils.rs` to manage ordered inflight/collected barrier items.
- Migrate partial-graph barrier tracking to the shared collector, including inflight counting, collected ordering, notifier cleanup, and completion state.
- Move checkpoint barrier completion sequencing into `PartialGraphManager` with explicit `start_completing` / `ack_completed` flow.
- Remove the dedicated creating-job barrier control queue and reuse the same partial-graph completion flow for creating streaming jobs.
- Update checkpoint control and recovery paths to acknowledge completed barriers through `PartialGraphManager`.
- Adjust barrier progress application code to consume worker responses from maps directly, matching the new collector-based flow.

The intention is to unify the barrier completion model for normal databases and creating jobs, reduce duplicated state machines, and make the ordering/ownership of barrier completion easier to reason about.

Behavior is intended to stay unchanged. This is a refactor of internal meta barrier management logic to centralize bookkeeping and reduce the risk of state divergence between different barrier control paths.

- Closes #[ISSUE_NUMBER]

## Checklist

- [ ] I have written necessary rustdoc comments.
- [ ]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  I have checked the [Release Timeline](https://github.com/risingwavelabs/rw-commits-history/blob/main/release_timeline.md) and [Currently Supported Versions](https://docs.risingwave.com/changelog/release-support-policy#support-end-dates-for-recent-releases) to determine which release branches I need to cherry-pick this PR into. 

## Documentation

- [ ]  My PR needs documentation updates. 

<details>
[...]

PR #25219, opened 2026-04-01
refactor(meta): unify barrier item collection for partial graphs and finishing jobs

wenym1:

- Closes #[ISSUE_NUMBER]

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/meta/src/barrier/checkpoint/control.rs
+++ b/src/meta/src/barrier/checkpoint/control.rs
@@ -49,7 +49,7 @@ use crate::barrier::partial_graph::{CollectedBarrier, PartialGraphManager, Parti
 use crate::barrier::progress::TrackingJob;
 use crate::barrier::rpc::{from_partial_graph_id, to_partial_graph_id};
 use crate::barrier::schedule::{NewBarrier, PeriodicBarriers};
-use crate::barrier::utils::collect_creating_job_commit_epoch_info;
+use crate::barrier::utils::{BarrierItemCollector, collect_creating_job_commit_epoch_info};
 use crate::barrier::{
     BackfillProgress, Command, CreateStreamingJobType, FragmentBackfillProgress, Reschedule,
 };
@@ -616,6 +616,9 @@ pub(in crate::barrier) struct DatabaseCheckpointControl {
     partial_graph_id: PartialGraphId,
     pub(super) state: BarrierWorkerState,
 
+    finishing_jobs_collector:
+        BarrierItemCollector<JobId, (Vec<BarrierCompleteResponse>, TrackingJob), ()>,
+
     /// Save the state and message of barrier in order.
     /// Key is the `prev_epoch`.
     command_ctx_queue: BTreeMap<u64, BarrierEpochState>,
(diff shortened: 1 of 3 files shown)


wenym1:
* **#25157** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25157?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#25219** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25219?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25219?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/meta/src/barrier/checkpoint/control.rs
```diff
@@ -1025,15 +1033,11 @@ impl DatabaseCheckpointControl {
...
-
-    creating_jobs_to_wait: HashSet<JobId>,
-
-    finished_jobs: HashMap<JobId, (Vec<BarrierCompleteResponse>, TrackingJob)>,
 }
 
 impl BarrierEpochState {
     fn is_inflight(&self) -> bool {
-        !self.is_collected || !self.creating_jobs_to_wait.is_empty()
+        !self.is_collected
```
zwang28, line 1040: Should we also check creating_jobs_to_wait (via finishing_jobs_collector) at the is_inflight callsite to preserve the original semantics?

src/meta/src/barrier/partial_graph.rs
```diff
@@ -470,7 +417,7 @@ impl PartialGraphManager {
             let PartialGraphStatus::Running(state) = graph else {
                 panic!("graph to be explicitly removed should be running");
             };
-            assert!(state.inflight_barriers.is_empty());
+            assert!(state.is_empty());
```
zwang28, line 420: The sematics of is_empty has changed. i.e. previously it's inflight_barriers.is_empty(), now it's inflight_barriers.is_empty() && collected_barriers.is_empty().
Will it cause any issue?

zwang28 (commented):
Left some comments.

src/meta/src/barrier/partial_graph.rs
wenym1: The previous naming `inflight_barriers` is ambiguous. It actually stores both the inflight and the collected ones.

src/meta/src/barrier/checkpoint/control.rs
wenym1: One purpose of this PR is to separate the management of `creating_jobs_to_wait` from `BarrierEpochState`. The only important caller of the `is_inflight` is when trying to figure out what epoch to commit, we check whether the oldest pending epoch is able to be committed. This guard logic is refactored as
```
let first_unfinished_job_epoch = self.finishing_jobs_collector.first_inflight_epoch();
while let Some((epoch, state)) = self.command_ctx_queue.first_key_value()
    && !state.is_inflight()
    && first_unfinished_job_epoch
        .is_none_or(|first_unfinished_epoch| *epoch < first_unfinished_epoch.prev)
```

which previously was
```
while let Some((epoch, state)) = self.command_ctx_queue.first_key_value()
    && !state.is_inflight()
```

PR #25220, opened 2026-04-01
fix(meta): load fragment parallelism from meta store table directly

wenym1:

This PR fixes a meta recovery bug where serving vnode mappings could not be rebuilt until barrier recovery had already completed.

Previously, serving info refresh read fragment parallelism from the in-memory shared_actor_infos. That state is only populated after recovery finishes, which creates a bad dependency during restart:
if a serving compute node comes back before streaming recovery completes, frontend queries may fail because the serving mapping is still empty.

This PR changes the source of fragment parallelism for serving mapping rebuilds to the persistent fragment catalog in meta store. Concretely:

- add fragment_parallelisms() in the catalog controller to load fragment distribution type and vnode count directly from the fragment table
- update serving info initialization to use this async DB-backed path instead of the in-memory runtime state
- remove the now-unused metadata wrapper and actor_count field
- add a simulation recovery test that restarts meta and only the serving compute node first, then verifies frontend queries still succeed before streaming recovery completes
- extend simulation cluster config to support explicit compute node roles so the recovery scenario can be reproduced reliably

This makes serving vnode mapping reconstruction independent from shared_actor_infos and allows batch/serving queries to work earlier in partial recovery scenarios.

- Closes #[ISSUE_NUMBER]

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/meta/src/controller/fragment.rs
+++ b/src/meta/src/controller/fragment.rs
@@ -417,19 +417,15 @@ impl CatalogController {
     /// from the in-memory `shared_actor_infos`.  This is critical because the
     /// serving vnode mapping must be available even before the barrier manager's
     /// recovery has completed and populated `shared_actor_infos`.
-    pub async fn running_fragment_parallelisms(
+    pub async fn fragment_parallelisms(
         &self,
-        id_filter: Option<HashSet<FragmentId>>,
     ) -> MetaResult<HashMap<FragmentId, FragmentParallelismInfo>> {
         let inner = self.inner.read().await;
-        let mut query = FragmentModel::find().select_only().columns([
+        let query = FragmentModel::find().select_only().columns([
             fragment::Column::FragmentId,
             fragment::Column::DistributionType,
             fragment::Column::VnodeCount,
         ]);
-        if let Some(id_filter) = id_filter {
-            query = query.filter(fragment::Column::FragmentId.is_in(id_filter));
-        }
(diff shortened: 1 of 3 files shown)


wenym1:
* **#25220** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25220?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25220?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/meta/src/controller/fragment.rs
```diff
@@ -412,31 +411,40 @@ impl CatalogController {
...
-                },
-            );
+        let inner = self.inner.read().await;
+        let mut query = FragmentModel::find().select_only().columns([
+            fragment::Column::FragmentId,
+            fragment::Column::DistributionType,
+            fragment::Column::VnodeCount,
+        ]);
+        if let Some(id_filter) = id_filter {
+            query = query.filter(fragment::Column::FragmentId.is_in(id_filter));
```
zwang28: The id_filter can be included in the SQL.

zwang28: It seems the `id_filter` is always None when `running_fragment_parallelisms` is called from `fetch_serving_infos`, which is the only callsite.
So `id_filter` can be removed as well.

PR #25224, opened 2026-04-01
feat(sink): support limiting the max concurrent initializing kv log store reader

wenym1:

This PR adds a compute-node level limit for concurrent historical reads during kv log store reader initialization.

Today, when many sink actors recover at the same time, each kv log store reader may start scanning historical data from the state store immediately. That can create a large burst of state store I/O during recovery or restart. This PR introduces a new developer config, `streaming.developer.max_concurrent_kv_log_store_historical_read`, so we can throttle how many kv log store readers are allowed to perform that initialization read concurrently on the same compute node.

The change works as follows:

- A new optional semaphore is created in `StreamEnvironment` from the developer config.
- The semaphore is passed into `KvLogStoreFactory` and then into each `KvLogStoreReader`.
- When a reader starts from historical state, it acquires a permit before opening the historical stream.
- The permit is held only while the reader is consuming the historical stream and is released automatically once the stream is exhausted or the reader state is reset.
- Setting the config to `0` keeps the existing behavior and disables the limit.

This PR also adds config docs and example config entries for the new knob.

For validation, a new unit test covers the key safety property here: when one reader is blocked waiting for the historical-read permit, the writer/log-store buffer path can still make progress and barrier flow is not blocked.

- Closes #[ISSUE_NUMBER]

## Checklist

## Documentation

<details>

</details>


wenym1:
* **#25224** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25224?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25224?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #25241, opened 2026-04-02
fix(sink): fix file sink name conflict issue

wcy-fdu:

The file sink could generate duplicate object names when the same writer rolled over multiple files within one second (`writer_uuid + created_time.as_secs()`), so we fixed it by adding a per-writer monotonically increasing `file_seq` suffix to each filename to guarantee uniqueness.

- Closes #[ISSUE_NUMBER]

## Checklist

## Documentation

<details>

</details>


wcy-fdu:
Local test passed.

src/connector/src/sink/file_sink/opendal_sink.rs
```diff
@@ -251,17 +252,18 @@ impl OpenDalSinkWriter {
...
             match sink_writer {
                 FileWriterEnum::ParquetFileWriter(w) => {
-                    if w.bytes_written() > 0 {
-                        let metadata = w.close().await?;
+                    let bytes_written = w.bytes_written();
+                    if bytes_written > 0 {
+                        w.close().await?;
                         tracing::info!(
-                            "writer {} (executor_id: {}, created_time: {}) finish write file, metadata: {:?}",
+                            "writer {} (executor_id: {}, created_time: {}) finish write file, bytes_written: {}",
```
hzxa21, line 259: How about making it a debug/trace log?

hzxa21 (approved):
Rest LGTM

src/connector/src/sink/file_sink/opendal_sink.rs
wcy-fdu: I've considered it, but I still think deleting the metadata field is better than downgrading the log level. Typically, this file writing won't be very frequent (usually not too often with a normal configuration (default `10240`)).

hzxa21: It will be very frequent. It is generally not a good idea to emit a log every couple seconds, especially when the information can be retrieved via metrics.

Even worse, this log entry is amplified by number of sink and number of sink parallelism in the cluster. Please do change it to debug/trace log. We do have a customer cluster running with 1000 s3 sinks. This log is too excessive.

PR #25245, opened 2026-04-03
fix(cdc): fix CDC false-ready handshake by setting Debezium Connected only after real streaming connection

wcy-fdu:

### Summary
This PR fixes a silent failure path in CDC backfill startup: Debezium could report `Connected=true` before the real streaming DB connection was established, causing RisingWave handshake to pass too early and then spin with no data when streaming actually failed.

### Problem
In backfill mode, RisingWave waits for Debezium streaming readiness via JMX `Connected` (`waitForStreamingRunning()` path).

Previously, `Connected` was marked too early in coordinator/init flow, while real streaming connection happened later inside connector-specific streaming code:
- PostgreSQL: `startStreaming(...)`
- MySQL: binlog client `onConnect`
- SQL Server: streaming iteration readiness path

This led to:
1. `Connected` becomes true too early
2. Handshake returns success
3. Real streaming connection fails afterwards
4. Debezium retries internally while polling loop keeps running
5. Source appears healthy but produces no data

### Fix
We changed the signal timing to “connected only after real streaming connect”.

- `ChangeEventSourceCoordinator` now registers connector-specific `onConnected` callbacks instead of eagerly setting `streamingConnected(true)`.
- PostgreSQL keeps callback trigger after successful `startStreaming(...)`.
- Added forked MySQL `MySqlStreamingChangeEventSource` with callback trigger on binlog `onConnect` (one-time guarded).
- Added forked SQL Server `SqlServerStreamingChangeEventSource` with callback trigger on streaming iteration readiness path (one-time guarded).

### Result
`Connected=true` now reflects actual streaming connectivity, so startup failures no longer pass handshake as false-ready, avoiding silent no-data spinning.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/java/connector-node/risingwave-source-cdc/src/main/java/io/debezium/connector/postgresql/PostgresStreamingChangeEventSource.java
+++ b/java/connector-node/risingwave-source-cdc/src/main/java/io/debezium/connector/postgresql/PostgresStreamingChangeEventSource.java
@@ -215,11 +215,6 @@ public class PostgresStreamingChangeEventSource
                         null, replicationConnection.startStreaming(walPosition));
             }
 
-            // Replication connection is now truly established. Signal connected.
-            if (onConnectedCallback != null) {
-                onConnectedCallback.run();
-            }
-
             // Start keep alive thread to prevent connection timeout during time-consuming
             // operations the DB side.
             ReplicationStream stream = this.replicationStream.get();
@@ -260,6 +255,11 @@ public class PostgresStreamingChangeEventSource
                                 PostgresConnector.class,
                                 connectorConfig.getLogicalName(),
                                 KEEP_ALIVE_THREAD_NAME));
+            }
+
+            // Signal connected only when the final streaming session is ready.
+            if (onConnectedCallback != null) {
+                onConnectedCallback.run();
             }
             processMessages(context, partition, this.effectiveOffset, stream);
         } catch (Throwable e) {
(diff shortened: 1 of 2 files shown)


wcy-fdu:
Pls review https://github.com/risingwavelabs/debezium/pull/9 for debezium's part's changes.

java/connector-node/risingwave-source-cdc/src/main/java/io/debezium/connector/postgresql/PostgresStreamingChangeEventSource.java
zwang28, line 248: The stream may reconnect and fail.
Please also review other connectors for similar corner cases.

wcy-fdu: Nice catch!

wcy-fdu: Fix this in both postgres and sqlerver.
MySQL does not need this extra change because its connected signal is emitted on the binlog client’s real onConnect event and it has no PostgreSQL-style startup “connect, search, reconnect” phase that can create a false-ready window.
Also apply the fix in https://github.com/risingwavelabs/debezium/pull/9

PR #25249, opened 2026-04-03
refactor(meta): extract independent ckpt job enum

wenym1:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## What's changed and what's your intention?

This PR refactors meta checkpoint control by introducing a dedicated IndependentCheckpointJobControl abstraction for jobs that run on their own partial graphs, and migrating the existing creating-
streaming-job path onto that abstraction.

Concretely:

- Add checkpoint/independent_job as the new home for independent partial-graph checkpoint logic.
- Introduce IndependentCheckpointJobControl as the wrapper type for independently-checkpointed jobs.
- Move the existing creating-streaming-job control code under checkpoint/independent_job/creating_job.

## Checklist

- [ ] I have written necessary rustdoc comments.
- [ ]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  I have checked the [Release Timeline](https://github.com/risingwavelabs/rw-commits-history/blob/main/release_timeline.md) and [Currently Supported Versions](https://docs.risingwave.com/changelog/release-support-policy#support-end-dates-for-recent-releases) to determine which release branches I need to cherry-pick this PR into. 

## Documentation

- [ ]  My PR needs documentation updates. 

<details>
<summary><b>Release note</b></summary>

</details>

PR #25250, opened 2026-04-03
feat(stream): allow snapshot backfill executor without upstream merge

wenym1:

This PR adds support for batch refresh jobs in `SnapshotBackfillExecutor`.

In a batch refresh job, snapshot backfill rebuilds the target from the table snapshot only. It should not receive any upstream changes through the merge input, so the snapshot backfill actor may legitimately have no upstream merge actors at runtime.

Previously, snapshot backfill always assumed there was a merge input and unconditionally waited for the first upstream barrier. That assumption does not hold for batch refresh, and could cause executor construction or startup to fail even though the job should be able to finish by consuming only the snapshot.

This PR fixes that by making the upstream input optional for snapshot backfill and teaching the executor to handle the no-upstream case explicitly:

- `MergeExecutorBuilder::new_input` now returns `Option<MergeExecutorInput>`. It returns `None` when `initial_upstream_actors` does not contain the upstream fragment, instead of trying to build an input from an empty actor list.
- Regular merge executor creation still rejects missing upstream actors with a clear error, so the relaxed behavior only applies to callers that intentionally support an empty upstream.
- `SnapshotBackfillExecutor` now accepts `Option<MergeExecutorInput>` and only waits for the first upstream barrier when an upstream is present.
- For batch refresh jobs without upstream input, the executor always enters the snapshot-backfill path when `snapshot_backfill_epoch` is set, instead of expecting barrier alignment with an upstream merge.
- The snapshot/log-store transition is refactored with a `SnapshotBackfillUpstream` wrapper so the existing buffering logic is reused when upstream exists, while log-store and upstream consumption are skipped cleanly when there is no upstream.

As a result, batch refresh jobs can execute snapshot backfill correctly without a merge input, while the existing behavior for normal streaming cases with upstream barriers and log-store replay remains unchanged.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/executor/backfill/snapshot_backfill/executor.rs
+++ b/src/stream/src/executor/backfill/snapshot_backfill/executor.rs
@@ -1116,3 +1116,485 @@ async fn make_consume_snapshot_stream<'a, S: StateStore>(
     }
     trace!(?barrier_epoch, "finish consuming snapshot");
 }
+
+#[cfg(test)]
+mod tests {
+    use std::collections::HashSet;
+    use std::sync::Arc;
+
+    use risingwave_common::array::StreamChunk;
+    use risingwave_common::catalog::{ColumnDesc, ColumnId, Field, Schema, TableId};
+    use risingwave_common::row::OwnedRow;
+    use risingwave_common::test_prelude::StreamChunkTestExt;
+    use risingwave_common::types::DataType;
+    use risingwave_common::util::epoch::{EpochPair, test_epoch};
+    use risingwave_common::util::sort_util::OrderType;
+    use risingwave_hummock_test::test_utils::{HummockTestEnv, prepare_hummock_test_env};
+    use risingwave_rpc_client::HummockMetaClient;
+    use risingwave_storage::hummock::HummockStorage;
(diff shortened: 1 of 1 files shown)


wenym1:
* **#25271** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25271?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#25251** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25251?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#25250** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25250?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25250?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#25249** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25249?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #25253, opened 2026-04-03
fix(sink): handle postgres upsert when all columns are primary keys

### LINKED ISSUE #25252: bug(sink): handle Postgres upsert when all sink columns are primary keys
## Summary

The native PostgreSQL sink generates invalid upsert SQL when all sink columns are part of the primary key and there are no non-primary-key columns to update.

## Impact

Sink execution fails because the generated SQL ends with an empty `DO UPDATE SET` clause.

Generated SQL shape:

```sql
INSERT INTO "schema"."table" ("col1", "col2")
VALUES ($1, $2)
ON CONFLICT ("col1", "col2") DO UPDATE SET
```

This is invalid PostgreSQL syntax.

## Expected Behavior

If there are no non-primary-key columns, the sink should not emit an empty update clause.

The sink should instead generate conflict handling equivalent to:

```sql
ON CONFLICT ("col1", "col2") DO NOTHING
```

## Actual Behavior

The native `connector = 'postgres'` sink builds an upsert statement with an empty `DO UPDATE SET` clause.

`connector = 'jdbc'` does not have this issue and works as a temporary workaround.

## Reproduction

Use a sink target table where every column belongs to the primary key and there are no additional columns.

Example shape:

```sql
CREATE TABLE target_table (
  user_id bigint,
  client_id bigint,
  PRIMARY KEY (user_id, client_id)
);
```

Then create a native PostgreSQL sink with `connector = 'postgres'` targeting that table.

## Suspected Root Cause

The PostgreSQL sink SQL builder filters update assignments to non-PK columns, but does not handle the empty-result case before formatting the final `ON CONFLICT ... DO UPDATE SET ...` statement.

## Proposed Fix

Add a guard in the PostgreSQL sink SQL generation path:

- If non-PK update columns exist, keep `DO UPDATE SET ...`
- If no non-PK update columns exist, generate `ON CONFLICT (...) DO NOTHING`

## Test Coverage

Add a regression test for a target table where:

- all sink columns are primary-key columns
- non-PK update column list is empty
- generated SQL uses `DO NOTHING` instead of `DO UPDATE SET`

## Source

Internal support thread reference: https://web.clearfeed.app/collections/6499?account=c3dc5f08-6bab-486f-962b-ef70c904ea81&request=3605

hzxa21:
## Summary
- generate `ON CONFLICT ... DO NOTHING` for native Postgres upsert sinks when every sink column is part of the primary key
- keep the existing `DO UPDATE SET ...` path when there are non-primary-key columns to update
- add a regression unit test for the all-primary-key case

## Testing
- cargo test -p risingwave_connector test_create_upsert_sql --lib

Fixes #25252

### DIFF
--- a/e2e_test/sink/postgres_sink.slt
+++ b/e2e_test/sink/postgres_sink.slt
@@ -590,6 +590,64 @@ DROP SINK rw_types_to_pg_types;
 statement ok
 DROP TABLE rw_types_subset_table;
 
+################### Test upsert sink when all sink columns are primary keys
+
+system ok
+PGDATABASE=sink_test psql -c "DROP TABLE IF EXISTS pg_pk_only_table"
+
+system ok
+PGDATABASE=sink_test psql -c "CREATE TABLE pg_pk_only_table (
+    user_id BIGINT,
+    client_id BIGINT,
+    PRIMARY KEY (user_id, client_id)
+)"
+
+system ok
+PGDATABASE=sink_test psql -c "INSERT INTO pg_pk_only_table VALUES (1, 10)"
+
+statement ok
+CREATE TABLE rw_pk_only_table (
+    user_id BIGINT,
+    client_id BIGINT,
+    PRIMARY KEY (user_id, client_id)
+);
+
+statement ok
+CREATE SINK rw_pk_only_sink FROM rw_pk_only_table WITH (
+    connector='postgres',
+    host='$PGHOST',
+    port='$PGPORT',
+    user='$PGUSER',
+    password='$PGPASSWORD',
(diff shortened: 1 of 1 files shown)


hzxa21:
Added a CI-path regression test in `e2e_test/sink/postgres_sink.slt` for the PK-only native Postgres sink case.

It creates a sink whose target table has only primary-key columns and verifies the sink can flush rows successfully, which would fail with the old empty `DO UPDATE SET` SQL.

Also re-ran `cargo test -p risingwave_connector test_create_upsert_sql --lib` locally.

PR #25258, opened 2026-04-06
feat(source): add source reader progress events

tabVersion:
## Problem (What)
Kafka transactional producers emit control records (commit/abort). Under `read_committed`, librdkafka does not deliver these control records, but `high_watermark` still includes their offsets. When the last offset in a partition is a control record, source backfill / bounded scans wait for `high_watermark - 1`, which is unreachable, so they hang (issue #24829).

## Fix (How)
1) Add `SourceReaderEvent::SplitProgress` so readers can report split progress even when no data chunks arrive.  
2) For Kafka, enable `enable.partition.eof`, and on `PartitionEOF` use `consumer.position()` to compute the reachable inclusive offset and emit `SplitProgress`.  
3) Source executor/backfill consumes these progress-only events to advance split state and drain stop offsets, allowing bounded/backfill to finish.

## Why it works
- `PartitionEOF` means the consumer has reached the end of the partition, even if the tail is an invisible control record.  
- `consumer.position()` returns the next readable offset, so `offset - 1` is the maximum reachable offset.  
- Propagating that as `SplitProgress` lets backfill advance to completion without relying on unseen control records.

### DIFF
--- a/src/connector/src/source/kafka/split.rs
+++ b/src/connector/src/source/kafka/split.rs
@@ -68,12 +68,4 @@ impl KafkaSplit {
             stop_offset,
         }
     }
-
-    pub fn get_topic_and_partition(&self) -> (String, i32) {
-        (self.topic.clone(), self.partition)
-    }
-
-    pub fn start_offset(&self) -> Option<i64> {
-        self.start_offset
-    }
 }


Cargo.toml
```diff
@@ -203,7 +203,7 @@ prost-derive = { version = "=0.14.3" }
 prost-reflect = { version = "0.16.3", features = ["serde"] }
 prost-types = { version = "=0.14.3" }
 rand = { version = "0.9", features = ["small_rng"] }
-rdkafka = { package = "madsim-rdkafka", version = "0.4.4", features = [
+rdkafka = { package = "madsim-rdkafka", git = "https://github.com/tabVersion/madsim.git", rev = "e572bf3", features = [
```
tabVersion: https://github.com/tabVersion/madsim/pull/1
change madsim-kafka to add position call support

Cargo.toml
```diff
@@ -179,7 +179,8 @@ jsonbb = { version = "0.2.3", features = ["float_roundtrip"] }
 jsonwebtoken = { version = "10", features = ["aws_lc_rs"] }
 linkme = { version = "0.3.32", features = ["used_linker"] }
 lru = "0.16"
-madsim = "0.2.34"
+# madsim = "0.2.34"
+madsim = { git = "https://github.com/tabVersion/madsim.git", rev = "498afcee77a0856b86af66cad61598ac968afcf1" }
```
tabVersion: ditto

src/connector/src/source/kafka/source/reader.rs
```diff
@@ -251,12 +339,39 @@ impl KafkaSplitReader {
...
+
+            let split_progress = if saw_partition_eof {
+                self.snapshot_split_progress(&mut latest_progress_offsets)?
+            } else {
+                None
+            };
+
+            if msgs.is_empty() {
+                if is_bounded && stop_offsets.is_empty() {
+                    if !res.is_empty() {
```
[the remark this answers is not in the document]
chenzl25: @tabVersion That's a valid point, PTAL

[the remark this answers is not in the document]
tabVersion: This should be fine now.

In the current implementation, `split_progress` is propagated from `PartitionEOF` and immediately passed through `apply_split_progress`, so bounded reads can drain stop_offsets before any early exit. Then, in the `msgs.is_empty()` path, the reader explicitly emits `SourceMessageEvent::SplitProgress(progress)` when
  progress exists instead of dropping it.

That means the control-record EOF case now works as intended even when no deliverable messages are returned:

  - progress-only updates are still propagated,
  - stop_offsets can still be drained for bounded scans,
  - and downstream backfill / source progress tracking can advance without waiting for an invisible control record.

src/connector/src/source/kafka/split.rs
```diff
@@ -72,4 +72,8 @@ impl KafkaSplit {
     pub fn get_topic_and_partition(&self) -> (String, i32) {
         (self.topic.clone(), self.partition)
     }
+
+    pub fn start_offset(&self) -> Option<i64> {
+        self.start_offset
+    }
```
hzxa21: unused?

tabVersion:
blocked by https://github.com/risingwavelabs/risingwave/pull/25351

src/connector/src/source/kafka/split.rs
tabVersion: removed

PR #25269, opened 2026-04-07
fix(cdc): use shared state to prevent concurrent schema history corruption

wcy-fdu:

Fix concurrent `OpendalSchemaHistory` instances corrupting each other's writes for the same `sourceId`.

### Root cause

When a Debezium engine is rebuilt (split change, error recovery, connector props change), a new `OpendalSchemaHistory` instance is created while the old engine may still be running. Each instance held **its own** `cachedLatestFile`, `cachedLatestFileRecords`, and `sequenceNumber`. This caused:

- **Stale cache reads**: Instance B doesn't see Instance A's writes, derives the same sequence number, and overwrites A's history file
- **Lost schema history records**: The overwritten file loses A's appended records

### Fix

Two mechanisms working together:

**1. Shared state per sourceId**

Replace per-instance mutable state with a per-sourceId `SharedState` object stored in a `static ConcurrentHashMap`. All instances for the same `sourceId` share one cache (`cachedLatestFile`, `cachedLatestFileRecords`, `sequenceNumber`), synchronized on the `SharedState` object. This prevents stale cache reads — every instance sees the latest sequence number and file info.

**2. Version-based stale instance rejection**

Each `SharedState` has a monotonically increasing `activeVersion`. On `doStart()`, the new instance increments `activeVersion` and records it as `myVersion`. In `doStoreRecord()`, stale instances (`myVersion != activeVersion`) are rejected immediately — first without the lock (fast path), then re-checked under the lock (double-check).

### Why this is better than #24935's lock-only approach

#24935 adds a per-sourceId `ReentrantLock` to serialize `doStart()` and `doStoreRecord()`. This prevents concurrent writes but **does not prevent stale cache reads**:

```
Lock approach:
  A: lock → write history_6.dat (seq=6) → unlock
  B: lock → own cache still says seq=5 → write history_6.dat → OVERWRITES A's file!

Shared state approach:
  A: synchronized → write history_6.dat, sharedState.seq=6 → release
  B: synchronized → sees sharedState.seq=6 → write history_7.dat → correct!
```

### Three guarantees

| Guarantee | Mechanism |
|-----------|-----------|
| Stale instances cannot block new ones | Version check rejects stale instances before acquiring the lock |
| Stale instances cannot write after a newer instance starts | `myVersion != activeVersion` → skip |
| No concurrent writes can corrupt files | `synchronized(sharedState)` serializes all access |

### Behavior change

- `sequenceNumber` changed from `AtomicLong` to plain `long` (all access is within `synchronized` block)
- No change to object store file format, directory layout, or external behavior
- Fully compatible with existing schema history files

### Note on the ideal fix

The ideal fix would be to prevent concurrent instances entirely — stop the old Debezium engine before starting the new one (stop-before-replace at the registry level). However, this is hard because:
- The old engine's shutdown is passive (detects channel closure), not actively triggered
- Debezium graceful shutdown latency depends on DB connection teardown, can hang
- Blocking new engine startup risks Rust-side handshake timeout and slower failover recovery
- A proper solution requires a cross-boundary (Rust ↔ Java) shutdown signal mechanism

This PR is a targeted fix at the `OpendalSchemaHistory` level that ensures correctness without changing engine lifecycle management.

## Checklist

## Documentation


wcy-fdu:
Note: 
### Ideal fix:
stop-before-replace at the registry level — ensure the old Debezium engine is fully stopped before starting a new one. One engine per sourceId at a time, no concurrency at all.

### Why it's hard:

- The old engine's shutdown is passive (detects channel closure), not actively triggered
- Debezium graceful shutdown has unpredictable latency (depends on DB connection teardown), can hang
- Blocking in register() waiting for the old engine would delay new engine startup, risking Rust-side handshake timeout and slower failover recovery
- A proper solution requires a cross-boundary (Rust ↔ Java) shutdown signal mechanism — large refactor


Thus this pr promotes per-instance mutable state (`sequenceNumber, cachedLatestFile`, etc.) to a per-sourceId SharedState object. Even when multiple instances coexist during the overlap window, they share the same cache and never write based on stale state.

java/connector-node/risingwave-source-cdc/src/main/java/com/risingwave/connector/cdc/debezium/internal/OpendalSchemaHistory.java
```diff
@@ -154,38 +186,63 @@ protected void doPreStoreRecord(HistoryRecord record) {}
 
     @Override
     protected void doStoreRecord(HistoryRecord record) {
+        // Fast path: reject stale instances without acquiring the lock.
```
zwang28: I recommend removing this "fast path". Without a "happens-before" relationship, it introduces additional risk like memory visibility.
synchronized (sharedState) is lightweight considering concurrent access is very rare.

java/connector-node/risingwave-source-cdc/src/main/java/com/risingwave/connector/cdc/debezium/internal/OpendalSchemaHistory.java
```diff
@@ -275,18 +332,18 @@ private long extractSequenceFromFileName(String fileName) {
 
     /**
      * Initialize sequence number from existing files. This ensures we don't reuse sequence numbers
-     * that have already been used.
+     * that have already been used. Caller must hold {@code synchronized(sharedState)}.
      */
     private void initializeSequenceNumber(List<String> historyFiles) {
```
zwang28, line 325: sharedState.sequenceNumber is reset upon every new instance.

wcy-fdu: Fixed.

java/connector-node/risingwave-source-cdc/src/main/java/com/risingwave/connector/cdc/debezium/internal/OpendalSchemaHistory.java
wcy-fdu: Make sense, removed.

PR #25251, opened 2026-04-03
feat(meta): support consuming snapshot of batch refreshable job (meta part)

wenym1:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## What's changed and what's your intention?

This PR introduces the **meta-side support** for *batch-refreshable* materialized views — MVs that consume a one-shot snapshot of their upstream at creation time, then sit idle until the next refresh interval, instead of continuously consuming changes like a regular streaming MV.

```sql
CREATE MATERIALIZED VIEW mv_refresh
WITH (refresh.interval.sec = 600) AS
SELECT * FROM mv_upstream;
```

This PR contains only the meta/control-plane part. Stream-executor changes and the periodic re-trigger of refresh cycles are out of scope and will land separately.

### High-level design

A batch refresh job runs in its **own partial graph**, decoupled from the database's main barrier stream — similar in spirit to `CreatingStreamingJobControl`, but with a different lifecycle:

| Aspect              | Batch Refresh                   | Regular Creating Job          |
|---------------------|---------------------------------|-------------------------------|
| Allowed upstream    | Must be an MV (no source/source-backfill) | Any source                    |
| Lifecycle           | Snapshot → Idle → (rerun)       | Snapshot → Steady-state       |
| External barriers   | None — fake barriers internally | Yes (global barrier)          |
| Partial graph       | Removed when snapshot completes | Persists                      |
| Log-epoch pinning   | Only while consuming snapshot   | Throughout backfill           |

To unify handling on the checkpoint-control side, the existing `CreatingStreamingJobControl` and the new `BatchRefreshJobCheckpointControl` are wrapped in a new `IndependentCheckpointJobControl` enum (`independent_job/mod.rs`). Common helpers like `new_fake_barrier` are factored out at the module root.

### State machine

`BatchRefreshJobCheckpointControl` (in `independent_job/batch_refresh_job/mod.rs`, ~1.1k lines) drives four states:

1. **ConsumingSnapshot** — actors run paused-at-creation, then unpause as fragments report ready. Synthetic ("fake") barriers are injected since there is no real upstream barrier source. Per-fragment progress is tracked.
2. **FinishingSnapshot** — final checkpoint + stop barrier injected, awaiting commit.
[...]

PR #25271, opened 2026-04-07
feat(meta): support periodic refresh run for batch refresh job

wenym1:

This PR adds periodic refresh execution for batch-refresh materialized views after the initial snapshot run completes.

Before this change, a batch-refresh MV could complete its snapshot backfill and become idle, but it did not start later refresh runs from upstream changes. This branch keeps the batch-refresh job control alive after snapshot completion, records the last committed epoch, and triggers a refresh run when the upstream committed epoch has advanced by at least the configured `refresh.interval.sec`.

Main changes:

- Extend `BatchRefreshJobCheckpointControl` with an `Idle { last_committed_epoch }` state and a new `ConsumingLogStore` state for periodic refresh runs.
- Add batch-refresh trigger polling in checkpoint control so idle jobs can emit a `BatchRefreshTrigger` event once their interval is due.
- Load trigger-time metadata and upstream changelog epochs from the barrier worker context before starting each refresh.
- Re-render actors for each refresh run, inject log-store barriers, and stop the temporary partial graph once the target upstream epoch commits.
- Update pinned upstream log epoch handling so snapshot runs and log-store runs pin the correct upstream range, while idle jobs do not pin logs.
- Refresh shared actor info when a periodic run starts, and keep drop / reset handling working across snapshot, idle, and log-store states.

Tests added:

- `e2e_test/streaming/batch_refresh_periodic.slt` verifies periodic refresh picks up inserts and deletes from the upstream MV.
- `test_batch_refresh_periodic_update` verifies refresh cycles continue after CN recovery and meta+CN recovery.
- Existing barrier worker mocks were updated for the new batch-refresh trigger context API.

No related issue.

## Checklist

## Documentation

<details>

Batch-refresh materialized views configured with `refresh.interval.sec` now periodically consume upstream MV changes after the initial snapshot completes. This allows the batch-refresh MV to reflect later inserts, deletes, and other upstream updates on the configured refresh interval instead of remaining at the initial snapshot result.

</details>

### DIFF
--- a/e2e_test/streaming/batch_refresh_periodic.slt
+++ b/e2e_test/streaming/batch_refresh_periodic.slt
@@ -1,0 +1,95 @@
+# Tests for the periodic-refresh trigger on batch-refresh materialized views.
+#
+# After the `refresh.interval.sec` elapses, a periodic refresh run should
+# consume upstream changes and surface them in the batch-refresh MV.
+
+statement ok
+DROP MATERIALIZED VIEW IF EXISTS mv_batch;
+
+statement ok
+DROP MATERIALIZED VIEW IF EXISTS mv_up;
+
+statement ok
+DROP TABLE IF EXISTS t_batch;
+
+statement ok
+CREATE TABLE t_batch (v1 int);
+
+statement ok
+INSERT INTO t_batch VALUES (1), (2);
+
+statement ok
+CREATE MATERIALIZED VIEW mv_up AS SELECT * FROM t_batch;
+
+statement ok
+CREATE MATERIALIZED VIEW mv_batch WITH (refresh.interval.sec = 3) AS
+SELECT * FROM mv_up;
+
+# Initial snapshot.
+query I rowsort
+SELECT * FROM mv_batch;
+----
+1
+2
+
+# ─── first periodic refresh picks up new upstream rows ───────────────────────
+
+statement ok
+INSERT INTO t_batch VALUES (3), (4);
+
+statement ok
+FLUSH;
+
+query I rowsort retry 5 backoff 1s
+SELECT * FROM mv_up;
+----
+1
+2
+3
+4
+
+# Wait for at least one refresh cycle to kick in and complete. The retry
+# window covers variance in the periodic trigger's firing time.
+query I rowsort retry 15 backoff 1s
+SELECT * FROM mv_batch;
+----
+1
+2
+3
+4
+
+# ─── second periodic refresh picks up further changes (incl. deletes) ────────
+
+statement ok
+DELETE FROM t_batch WHERE v1 = 1;
+
+statement ok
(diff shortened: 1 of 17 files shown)


wenym1:
* **#25271** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25271?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25271?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#25251** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25251?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#25250** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25250?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#25249** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25249?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

chenzl25:
I think there is still a correctness blocker for mixed same-db + cross-db batch refresh MVs.

During creation, `fill_snapshot_backfill_epoch` receives both `snapshot_backfill_info` and `cross_db_snapshot_backfill_info`, so a query that contains both `SnapshotBackfill` and `CrossDbSnapshotBackfill` can be initialized. But the periodic refresh path later only tracks the same-db side: `BatchRefreshJobCheckpointControl` stores `snapshot_backfill_upstream_tables` from `batch_refresh_info.snapshot_backfill_info`, and `load_batch_refresh_trigger_context_impl` calls `collect_snapshot_backfill_info_impl(...).0`, discarding the cross-db info.

As a result, `target_upstream_epoch`, changelog epoch loading, and changelog pinning can miss the cross-db upstream. If only that upstream advances, or if it advances together with same-db upstreams, the refresh run can finish without replaying the cross-db changelog, leaving the batch refresh MV stale.

Could we either reject batch refresh plans containing `CrossDbSnapshotBackfill`, or include the cross-db upstream tables in the trigger/log-pin/replay path as well? A mixed same-db + cross-db regression test would cover this.

wenym1:
> As a result, target_upstream_epoch, changelog epoch loading, and changelog pinning can miss the cross-db upstream.

We don't have a mechanism for changelog pinning for cross-db downstream yet, because the control logic is very complicated for it at the moment. Currently its changelog pinning depends on creating subscription on the upstream tables/MVs with large enough retention time, and if the lag exceeds the retention time, there is no way to resume. I added alert and some error isolation in https://github.com/risingwavelabs/risingwave/pull/25884 to improve the related user experience.

chenzl25 (approved):
LGTM!

PR #25272, opened 2026-04-07
fix: ignore stale refresh finish events after source drop

tabVersion:
Co-authored-by: tab <tabversion@bupt.icu>I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## What changed

* handle stale refresh completion events (ListFinish/LoadFinish) when the associated source is already dropped
* treat CatalogIdNotFound(object, ..) from get_object_database_id(associated_source_id) as idempotent skip instead of bubbling to barrier completion failure

## Why

- In REFRESH TABLE + immediate DROP TABLE races, refresh finish events can arrive after catalog cleanup. Previously this propagated as:
`Failed to get database id for table -> object id not found`, which could block command completion and trigger recovery noise.

---

cherry pick https://github.com/risingwavelabs/risingwave/pull/24983 to main

- Why the prev pr is not in main
  - #24983 was part of https://github.com/risingwavelabs/risingwave/pull/24835 , but later I switched to another approach and the FK issue is not observed on main branch
  - In https://github.com/risingwavelabs/risingwave/pull/25258 's [CI](https://buildkite.com/risingwavelabs/pull-request/builds/94748#019d6730-232d-4cb8-80ba-7fb4dda04c60) , I saw the same FK issue. 

```
2026-04-07T09:25:05.929195Z  INFO sqllogictest::system_command: system command executed command="echo \"8,henry\" > ./e2e_test/source_inline/refresh/refresh_table_tmp/new2.csv" status=ExitStatus(unix_wait_status(0)) stdout="" stderr=""
--
e2e_test/source_inline/kafka/upsert_source.slt               .. [FAILED] after 107053 ms
 
failed to run `e2e_test/source_inline/kafka/upsert_source.slt`
 
Caused by:
statement failed: db error: ERROR: Failed to run the query
 
1: gRPC request to meta service (call `/ddl_service.DdlService/CreateSink`) failed: Internal error
2: failed to wait streaming job finish: failed over: Foreign Key Constraint Violated: FOREIGN KEY constraint failed
 
[SQL] CREATE SINK sk INTO students AS
SELECT
rw_key,
"ID",
"firstName" \|\| ' ' \|\| "lastName" AS name,
age
FROM upsert_students;
at e2e_test/source_inline/kafka/upsert_source.slt:53
 

2026-04-07T09:25:13.107398495Z  INFO bootstrap_recovery:recovery_attempt:failure_recovery: risingwave_meta::barrier::schedule: cluster marked as blocked error=Foreign Key Constraint Violated: FOREIGN KEY constraint failed
2026-04-07T09:25:13.108488311Z  INFO bootstrap_recovery:recovery_attempt:failure_recovery: risingwave_meta::barrier::worker: recovery start! error=Foreign Key Constraint Violated: FOREIGN KEY constraint failed
2026-04-07T09:25:13.108878669Z ERROR bootstrap_recovery:recovery_attempt: risingwave_meta::rpc::ddl_controller: failed to create streaming job id=306 error=failed to wait streaming job finish: failed over: Foreign Key Constraint Violated: FOREIGN KEY constraint failed
2026-04-07T09:25:13.108906514Z ERROR bootstrap_recovery:recovery_attempt: risingwave_meta::rpc::ddl_controller: failed to create streaming job id=187 error=failed to wait streaming job finish: failed over: Foreign Key Constraint Violated: FOREIGN KEY constraint failed
2026-04-07T09:25:13.108966989Z ERROR bootstrap_recovery:recovery_attempt: risingwave_meta::rpc::ddl_controller: failed to create streaming job id=407 error=failed to wait streaming job finish: failed over: Foreign Key Constraint Violated: FOREIGN KEY constraint failed
2026-04-07T09:25:13.125935695Z  WARN bootstrap_recovery:recovery_attempt: risingwave_meta::rpc::ddl_controller: aborted streaming job id=306
```

So the fix should be cherry-picked

- Closes #[ISSUE_NUMBER]

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/meta/src/barrier/context/context_impl.rs
+++ b/src/meta/src/barrier/context/context_impl.rs
@@ -558,12 +558,12 @@ mod tests {
     #[test]
     fn test_skip_refresh_finish_when_associated_source_missing() {
         let err = MetaError::catalog_id_not_found("object", 42);
-        assert!(should_skip_refresh_finish_for_missing_object(&err));
+        assert!(err.is_catalog_id_not_found("object"));
     }
 
     #[test]
     fn test_do_not_skip_refresh_finish_for_other_not_found_types() {
         let err = MetaError::catalog_id_not_found("table", 42);
-        assert!(!should_skip_refresh_finish_for_missing_object(&err));
+        assert!(!err.is_catalog_id_not_found("object"));
     }
 }


src/meta/src/controller/catalog/alter_op.rs
wenym1, line 1183: Any reason to have this `ensure_refresh_job` in the first place? IIUC, should it be inserted with the job's other metadata in a single transaction? And then we shouldn't see this `foreign key constrain` error and don't have to handle it here.

src/meta/src/barrier/context/context_impl.rs
```diff
@@ -246,11 +245,41 @@ impl GlobalBarrierWorkerContext for GlobalBarrierWorkerContextImpl {
...
+            Err(err) => Err(err),
+        }
+    }
+
     fn set_status(&self, new_status: BarrierManagerStatus) {
         self.status.store(Arc::new(new_status));
     }
 }
 
+fn should_skip_refresh_finish_for_missing_object(err: &MetaError) -> bool {
```
wenym1: Can instead introduce a helper method `is_catalog_id_not_found` in `MetaError`

src/meta/src/controller/catalog/alter_op.rs
```diff
@@ -1261,4 +1286,28 @@ impl CatalogController {
...
+        &self,
+        table_id: TableId,
+        err: &sea_orm::DbErr,
+    ) -> MetaResult<bool> {
+        if !is_refresh_job_stale_drop_candidate(err) {
+            return Ok(false);
+        }
+
+        let inner = self.inner.read().await;
+        let table_exists = Table::find_by_id(table_id).one(&inner.db).await?.is_some();
```
wenym1: Any reason to check the db again? If we don't need to handle the foreign key error as mentioned above, and meanwhile we can get `sea_orm::DbErr::RecordNotUpdated`, we can be sure that the table has been deleted.

src/meta/src/controller/catalog/alter_op.rs
tabVersion: `ensure_refresh_job` is not in the main path when creating refreshable jobs. It serves as an additional check to heal from somehow missing entries. We can remove it in future pr.

src/meta/src/controller/catalog/alter_op.rs
```diff
@@ -1262,3 +1287,26 @@ impl CatalogController {
...
+    }
+
+    if !matches!(
+        err.sql_err(),
+        Some(SqlErr::ForeignKeyConstraintViolation(_))
+    ) {
+        return Ok(false);
+    }
+
+    let table_exists = Table::find_by_id(table_id).one(db).await?.is_some();
```
wenym1, line 1310: any comment on https://github.com/risingwavelabs/risingwave/pull/25272#discussion_r3045973650?

src/meta/src/controller/catalog/alter_op.rs
tabVersion: Good catch. Given that `ensure_refresh_job` is called right before the update, RecordNotUpdated here effectively means the row disappeared between those two steps, which indicates the table was deleted concurrently and this is a stale refresh event.

The extra DB check is only necessary for the foreign key violation case, since that signal is not specific enough by itself.

wenym1 (approved):
LGTM

PR #25275, opened 2026-04-08
feat(stream): add change buffer executor to solve join amplification

EdwinaZhu:

- Summary: This PR introduces a `ChangeBufferExecutor` that sorts StreamChunk within the same epoch before forwarding them downstream. Once the in-memory buffer is full, the executor spills data into the log store. When data needs to be emitted, it drains the log store first and then drains the in-memory buffer.
- How it works: The executor reconstructs two local old/new-value streams from the memtable, staging imm, and staging SSTs (both old and new sides), and then reuses `ChangeLogIteratorInner` to produce the uncommitted changelog output.
- Links: this is the demo pr https://github.com/risingwavelabs/risingwave/pull/23755

## Checklist

## Documentation

<details>

</details>


src/storage/src/store.rs
```diff
@@ -366,10 +366,23 @@ pub trait StateStore: StateStoreReadLog + StaticSendSync + Clone {
     ) -> impl Future<Output = Self::VectorWriter> + Send + '_;
 }
 
+/// Reads the net changelog of the current unsealed epoch from a local state store instance.
+///
+/// Unlike [`StateStoreReadLog`], this trait is intended for executor-private local state and does
+/// not expose committed-epoch or key-range semantics or any implementation detail such as imm/sst
+/// layout.
+pub trait LocalStateStoreReadLog: StaticSendSync {
```
EdwinaZhu: This new trait exposes a single capability: read back the **net changelog** that the current actor has written in the (still uncommitted) current epoch. It is intentionally narrower than `StateStoreReadLog` — no committed-epoch / key-range semantics, and no leakage of imm/sst layout — so it can be consumed by executor-private local state (used by `ChangeBufferExecutor` here) without coupling streaming code to hummock internals.

Note that it is added as a supertrait of `LocalStateStore` below, which forces every local state store implementation (hummock, memory, panic, traced, monitored wrappers) to provide an implementation or trivial forwarding.

EdwinaZhu:
截屏2026-05-22 18 52 54
截屏2026-05-22 18 53 04

I ran the sort_buffer.slt, and the output is shown above. The blue part is the output when `enable_locality_sort_buffer = false`, the red part is the output when `enable_locality_sort_buffer = true`. As we can see, when using locality sort buffer, it shortened the runtime. Meanwhile, the memory usage didn't increase too much nearly.

PR #25283, opened 2026-04-08
feat(meta): track iceberg compaction with commit backlog

### LINKED ISSUE #25211: iceberg compactor accesses iceberg catalog too frequently
Compactor PullTask loop (every 1s × all sinks) sends a PullTask → evaluates should_trigger() for every sink. Because of the early snapshot-count trigger, t must call load_table() on every registered sink in parallel to get the live snapshot count.

Li0k:

This PR is a V1 fix for Iceberg sink compaction scheduling pressure in meta.

- Meta now tracks per-sink Iceberg compaction backlog in memory with `pending_commit_count` and an `Idle/PendingDispatch/InFlight` state machine instead of probing Iceberg catalog metadata on every `PullTask`.
- Iceberg sink commits now emit update events, and the Iceberg compactor reports task-level `SUCCESS/FAILED`, so meta consumes the dispatched commit baseline only after the task completes.
- Timed-out tasks are treated as lease expiry instead of immediate worker loss. The report timeout is configurable via the global meta config `iceberg_compaction_report_timeout_sec`.
- Iceberg compactor stream handling now fences stale same-`context_id` streams with per-context stream generations, so an old stream cannot remove a newly reconnected session.
- Manual compaction now waits for task reports instead of polling Iceberg snapshots for completion.
- The manager/report implementation is also cleaned up into smaller modules to keep the scheduling and report flow reviewable.

Behavioral notes for this V1:

- Meta does not bootstrap or recover historical Iceberg compaction backlog across restart.
- Timeout is a lease expiry signal, not proof that the original task has stopped running.
- `enable_compaction=false` is still not an immediate lifecycle clear in this PR; schedule config continues to reconcile on sink update events.
- `compaction.trigger_snapshot_count` must be greater than `0`.

Validation:

- `cargo test -p risingwave_meta iceberg_compaction --lib --quiet`
- `cargo clippy -p risingwave_meta --lib --tests -- -D warnings`
- `cargo test -p risingwave_storage iceberg_compaction --lib --quiet`
- `cargo clippy -p risingwave_storage --lib --tests -- -D warnings`

- Closes #25211

## Checklist

## Documentation

<details>

Reduce Iceberg compaction scheduling pressure in meta by tracking pending compaction work with in-memory commit backlog state instead of hot-path catalog polling. Also add a global meta config to tune the Iceberg compaction report timeout lease.

</details>


Li0k:
@chenzl25 PTAL

chenzl25 (approved):
Generally LGTM

PR #25288, opened 2026-04-09
feat(iceberg): sync comments to iceberg metadata

### LINKED ISSUE #25147: [Feature] Support comments propagation to Iceberg table metadata

Currently, when commenting to Risingwave table/columns with engine Iceberg, these comments are only created in internal Risingwave catalog and are not propagated to Iceberg table metadata.

I would like to request a feature to propagate comments from a Risingwave table with engine Iceberg or Iceberg sinked from a normal Risingwave table.

_No response_

### Additional context

_No response_

xxhZs:

This PR syncs RisingWave table and column comments to Iceberg metadata for Iceberg tables and Iceberg sinks.

It adds a shared Iceberg metadata update path that maps:
- RisingWave table comments to the Iceberg table property `comment`
- RisingWave column comments to Iceberg field `doc`

The sync now happens in two places:
- after creating an Iceberg engine table or Iceberg sink, so existing comments on the source table are written to Iceberg metadata immediately
- when running `COMMENT ON TABLE` / `COMMENT ON COLUMN`, so later comment updates are propagated to the corresponding Iceberg table metadata

This PR also fixes two issues found during local validation:
- avoid rebinding the same source relation in `create_sink`, which caused `Duplicated table name while binding table to context` for Iceberg engine table creation
- use the correct internal Iceberg sink naming rule (`__iceberg_sink_*`) when looking up the internal sink for Iceberg engine tables during comment sync

- Closes #25147

## Checklist

## Documentation

<details>

RisingWave now syncs table and column comments to Iceberg metadata for Iceberg engine tables and Iceberg sinks. Table comments are written to the Iceberg table `comment` property, and column comments are written to Iceberg field `doc`.

</details>


tuantran0910:
Hi @xxhZs, may I ask that are there any updates on this PR :D I am looking forward to see this feature in the upcoming releases of Risingwave :v Moreover, I have a concern, we are using dbt risingwave and this will generate tons of comment updates on the table, therefore, will it create storm of update schema requests to Iceberg catalog? Thanks

xxhZs:
> Hi @xxhZs, may I ask that are there any updates on this PR :D I am looking forward to see this feature in the upcoming releases of Risingwave :v Moreover, I have a concern, we are using dbt risingwave and this will generate tons of comment updates on the table, therefore, will it create storm of update schema requests to Iceberg catalog? Thanks

Hi, thanks for checking in! I’ve just updated the PR and resolved the latest conflicts.

Regarding the dbt concern: the implementation only sends an Iceberg metadata update when the desired table/column comments are different from the current Iceberg metadata, so repeated syncs with the same comments should be no-op. That said, if dbt generates many actual COMMENT ON changes, each change can still result in an Iceberg catalog metadata update. I think that matches the current semantics, but if this becomes too noisy in practice, we can consider batching/debouncing comment syncs in a follow-up.

chenzl25:
I found two correctness blockers in the current head `e52b7f64f1bf7d3716d76e0365d9b376fe752e67`:

1. `rest_rust` / `glue_rust` column-comment sync can fail directly. `build_comment_updates` converts column doc changes into an Iceberg schema update with `CurrentSchemaIdMatch`, but the `rest_rust | glue_rust` path bails when it sees that requirement. That means `COMMENT ON COLUMN` on REST vended credentials / Glue Rust catalogs can return an error instead of just updating RW's local comment, and `CREATE TABLE` / `CREATE SINK` with column comments can hit the same failure path.

2. Create-time Iceberg comment sync runs after the RW catalog commit. In both `create_sink` and `create_table`, the RW catalog object has already been created/replaced before we try to sync comments to the external Iceberg catalog. Any external catalog/network failure, including the `rest_rust` / `glue_rust` column-comment path above, makes SQL report `CREATE ...` failure even though the RW object already exists. That leaves inconsistent DDL semantics.

I think the comment sync should either be best-effort/non-fatal after the RW catalog commit, or it needs a clear transactional/compensation story. At minimum, please cover table comments, column comments, Rust REST/Glue catalog behavior, and the external-sync-failure case in tests. The PR currently changes `CREATE TABLE` / `CREATE SINK` / `COMMENT ON` side effects but does not appear to add or update tests for these paths.

xxhZs:
> I found two correctness blockers in the current head `e52b7f64f1bf7d3716d76e0365d9b376fe752e67`:
> 
> 1. `rest_rust` / `glue_rust` column-comment sync can fail directly. `build_comment_updates` converts column doc changes into an Iceberg schema update with `CurrentSchemaIdMatch`, but the `rest_rust | glue_rust` path bails when it sees that requirement. That means `COMMENT ON COLUMN` on REST vended credentials / Glue Rust catalogs can return an error instead of just updating RW's local comment, and `CREATE TABLE` / `CREATE SINK` with column comments can hit the same failure path.
> 2. Create-time Iceberg comment sync runs after the RW catalog commit. In both `create_sink` and `create_table`, the RW catalog object has already been created/replaced before we try to sync comments to the external Iceberg catalog. Any external catalog/network failure, including the `rest_rust` / `glue_rust` column-comment path above, makes SQL report `CREATE ...` failure even though the RW object already exists. That leaves inconsistent DDL semantics.
> 
> I think the comment sync should either be best-effort/non-fatal after the RW catalog commit, or it needs a clear transactional/compensation story. At minimum, please cover table comments, column comments, Rust REST/Glue catalog behavior, and the external-sync-failure case in tests. The PR currently changes `CREATE TABLE` / `CREATE SINK` / `COMMENT ON` side effects but does not appear to add or update tests for these paths.

fixed

chenzl25:
Thanks for working on this. I think there is one correctness issue that should be fixed before merge:

`COMMENT ON` currently finds all sinks that depend on the source table, then applies the source table/column comments to each Iceberg sink schema. That seems unsafe for `CREATE SINK ... AS SELECT ... FROM t`, joins, renames, or expressions: commenting `t.a` could update the doc of an unrelated output column named `a` on a derived Iceberg table. The create-sink path only syncs comments for direct sinks; the comment path should have the same restriction, or use an explicit column mapping.

Two smaller follow-ups:
- The PR text says column comments are synced to Iceberg field `doc`, but the native Rust REST/Glue path only warns and updates table properties. Please either implement this path or document/test the limitation.
- Iceberg catalog/storage I/O happens while holding the catalog write lock and before committing the meta txn. It would be safer to commit RW comments first and do this external sync asynchronously/best-effort, otherwise a slow external catalog can block catalog writes and an Iceberg-success/RW-commit-failure can leave state diverged.

Could you also add tests covering direct sink sync, AS SELECT sink should not sync, and native Rust catalog column-comment behavior?

src/meta/src/controller/catalog/mod.rs
```diff
@@ -719,9 +727,129 @@ impl CatalogController {
...
+                sink,
+                sink_obj.ok_or_else(|| MetaError::catalog_id_not_found("object", sink_id))?,
+                None,
+            )));
+        }
+        Ok(sinks)
+    }
+
+    fn is_direct_iceberg_comment_sink(sink: &sink::Model, table: &table::Model) -> bool {
+        sink.name == format!("{}{}", ICEBERG_SINK_PREFIX, table.name)
```
chenzl25: This directness check is name-based, but renaming the source relation does not update `sink.sink_from_name` or the internal sink name: `rename_relation_refer` only rewrites the stored `definition`. After `ALTER TABLE ... RENAME TO ...`, a direct Iceberg sink (and especially `__iceberg_sink_<old_name>`) still depends on the same table ID, but both comparisons here fail, so later `COMMENT ON TABLE/COLUMN` silently stops syncing. Please base this on stable catalog IDs/metadata, or update the identifying fields transactionally on rename, and add a rename regression test.

xxhZs: fixed，please review again

chenzl25:
I don't think comment sync is a valid issue given the complexity it involves. I don't think we need to support this feature right now. Let's close it until we really need ti.

PR #25331, opened 2026-04-13
refactor(streaming): replace BatchTable with ReplicatedStateTable in lookup executor

wenym1:

This PR continues the `BatchTable` → `ReplicatedStateTable` migration in the streaming executor layer, extending the work from temporal join (commit `7a91a34602`) to the lookup executor used in delta join operations.

**Summary of changes:**
- Replace `BatchTable<S>` with `ReplicatedStateTable<S, SD>` in the lookup executor
- Add `row.project(output_indices)` after `iter_with_prefix()` calls to properly project rows to output columns (similar to temporal join implementation)
- Update the lookup executor's table initialization to use `ReplicatedStateTable` with proper column ID mapping
- Update related deserialization and message handling to work with the new table type

**Why this change:**
- Standardizes state table access patterns across streaming executors
- Enables proper support for replicated state tables in delta join operations
- Maintains consistency with the temporal join refactor completed in the parent commit

**Testing:**
- All 5 e2e delta join SLT tests pass
- Unit tests for lookup executor pass
- Clippy checks pass
- Type safety verified: row projection fix ensures column types match between deserial and output

## Checklist

## Documentation

<details>
<summary>Release note</summary>

</details>

### DIFF
--- a/src/stream/src/common/table/state_table.rs
+++ b/src/stream/src/common/table/state_table.rs
@@ -670,11 +670,6 @@ impl<S: StateStore, SD: ValueRowSerde, const IS_REPLICATED: bool, PreloadAllRow>
         self
     }
 
-    pub fn with_output_column_ids(mut self, output_column_ids: Vec<ColumnId>) -> Self {
-        self.output_column_ids = Some(output_column_ids);
-        self
-    }
-
     pub fn enable_vnode_key_stats(mut self, enable: bool, config: &StreamingConfig) -> Self {
         self.enable_vnode_key_stats = Some(enable);
         self.enable_state_table_vnode_stats_pruning =
@@ -684,6 +679,15 @@ impl<S: StateStore, SD: ValueRowSerde, const IS_REPLICATED: bool, PreloadAllRow>
 
     pub fn with_metrics(mut self, metrics: StateTableMetrics) -> Self {
         self.metrics = Some(metrics);
+        self
+    }
+}
+
+impl<S: StateStore, SD: ValueRowSerde, PreloadAllRow>
+    StateTableBuilder<S, SD, true, PreloadAllRow>
+{
+    pub fn with_output_column_ids(mut self, output_column_ids: Vec<ColumnId>) -> Self {
+        self.output_column_ids = Some(output_column_ids);
         self
     }
 }
@@ -1065,7 +1069,7 @@ impl<S: StateStore, SD: ValueRowSerde, const IS_REPLICATED: bool>
         table_desc: &StorageTableDesc,
         store: S,
         vnodes: Option<Arc<Bitmap>>,
-        fragment_id: u32,
+        fragment_id: FragmentId,
     ) -> Self {
         let table_id = table_desc.table_id;
         let table_columns: Vec<ColumnDesc> =
@@ -1107,7 +1111,7 @@ impl<S: StateStore, SD: ValueRowSerde, const IS_REPLICATED: bool>
             prefix_hint_len: table_desc.read_prefix_len_hint as usize,
             retention_seconds: table_desc.retention_seconds,
             versioned: table_desc.versioned,
-            fragment_id: fragment_id.into(),
+            fragment_id,
             clean_watermark_index: None,
             store,
             vnodes,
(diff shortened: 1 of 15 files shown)


wenym1:
* **#25331** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25331?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25331?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#25319** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25319?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

chenzl25:
I think `commit_assert_no_update_vnode_bitmap` may not be the right behavior here. Lookup can still receive update-vnode barriers; asserting would make rescale fail instead of applying the vnode bitmap update.

Should this follow the temporal join / old `BatchTable` behavior instead: commit first, yield the barrier, then handle `post_yield_barrier(update_vnode_bitmap)` and clear `LookupCache` if the keyed cache becomes stale?

src/stream/src/common/table/state_table.rs
```diff
@@ -1841,7 +1845,13 @@ where
...
             .iter_kv_with_pk_range::<()>(pk_range, vnode, prefetch_options)
             .await?
-            .map_ok(|(_, row)| row))
+            .map_ok(|(_, row)| {
+                if IS_REPLICATED {
+                    row.project(&self.output_indices).into_owned_row()
+                } else {
+                    row
+                }
+            }))
```
chenzl25, line 1854: Is it a bug before this fix?

src/stream/src/common/table/state_table.rs
```diff
@@ -2069,7 +2066,13 @@ where
...
         let stream = self.iter_with_prefix_inner::</* REVERSE */ false, ()>(pk_prefix, sub_range, prefetch_options)
             .await?;
-        Ok(stream.map_ok(|(_, row)| row))
+        Ok(stream.map_ok(|(_, row)| {
+            if IS_REPLICATED {
+                row.project(&self.output_indices).into_owned_row()
+            } else {
+                row
+            }
+        }))
```
chenzl25, line 2075: ditto

chenzl25 (approved):
Rest LGTM

src/stream/src/common/table/state_table.rs
wenym1: no. The previous caller will manually do the map project, and now we unify to put it inside the state table returned stream, just like the current get_row.

wenym1:
> I think `commit_assert_no_update_vnode_bitmap` may not be the right behavior here. Lookup can still receive update-vnode barriers; asserting would make rescale fail instead of applying the vnode bitmap update.
> 
> Should this follow the temporal join / old `BatchTable` behavior instead: commit first, yield the barrier, then handle `post_yield_barrier(update_vnode_bitmap)` and clear `LookupCache` if the keyed cache becomes stale?

In our current scale logic, there is actually no online update vnode bitmap any more. Scaling a fragment will stop all actors of this fragment and create new actors with the new vnode distribution, so there is no need to handle in place update vnode bitmap any more.

chenzl25:
> In our current scale logic, there is actually no online update vnode bitmap any more. Scaling a fragment will stop all actors of this fragment and create new actors with the new vnode distribution, so there is no need to handle in place update vnode bitmap any more.

I see.

chenzl25:
> > In our current scale logic, there is actually no online update vnode bitmap any more. Scaling a fragment will stop all actors of this fragment and create new actors with the new vnode distribution, so there is no need to handle in place update vnode bitmap any more.
> 


 I see, so we don't need to handle the update vnode bitmap for all executors anymore, right?

wenym1:
> > > In our current scale logic, there is actually no online update vnode bitmap any more. Scaling a fragment will stop all actors of this fragment and create new actors with the new vnode distribution, so there is no need to handle in place update vnode bitmap any more.
> 
> I see, so we don't need to handle the update vnode bitmap for all executors anymore, right?

Yes. For newly added executors in the future, we don't have to handle update vnode bitmap so that we can reduce complexity. For logic of handling update vnode bitmap in existing executors, we can keep them in case we want to bring back in-place update vnode bitmap to avoid in memory cache invalidation, unless the existing logic increase the complexity when add new features to them.

Issue #25336, opened 2026-04-13
Breaking change >=2.7: JSONB distribution-key hash change can remap the same logical key to a different vnode


maybe-vibe:
## Summary

When upgrading from a version before `v2.7` to `v2.7+`, tables whose primary key / distribution key contains `JSONB` can compute a different vnode for the same logical JSONB value.

This is because vnode assignment hashes the distribution-key datum, and JSONB hashing relies on Rust's `Hash` trait behavior for the underlying JSONB representation. That hash behavior changed across the upgrade even when the JSONB value is logically equal and has the same key-serialized representation.

This can leave old rows persisted under the old vnode while new operations compute a new vnode for the same JSONB key.

#23497 -> https://github.com/risingwavelabs/jsonbb/pull/7

## Simple Repro

On a pre-`v2.7` cluster:

```sql
CREATE TABLE t (a JSONB PRIMARY KEY);
INSERT INTO t VALUES ('"foo"'::jsonb);
```

Upgrade the cluster to `v2.7+`.

Then insert or otherwise operate on the same logical key:

```sql
INSERT INTO t VALUES ('"foo"'::jsonb);
```

The same logical JSONB key may now be assigned to a different vnode.

A symptom can be:

```sql
SELECT DISTINCT a FROM t;
```

returning duplicate-looking rows, while equality still says they are equal:

```sql
SELECT t1.a, t2.a, t1.a = t2.a
FROM t AS t1, t AS t2;
```

## What Leads To This

The persisted Hummock table key is effectively:

```text
vnode bytes + memcomparable primary-key bytes
```

The vnode prefix is computed separately from the serialized key payload.

For a table like:

```sql
CREATE TABLE t (a JSONB PRIMARY KEY);
```

the `JSONB` column `a` is both the primary key and the distribution key.

The write path computes the vnode from the primary/distribution key:

```text
StateTable::serialize_pk
  -> compute_vnode_by_pk
  -> compute_vnode
  -> VirtualNode::compute_row
  -> hash distribution-key row with Crc32FastBuilder
  -> hash % vnode_count
```

Relevant code paths:

```text
src/stream/src/common/table/state_table.rs
src/common/src/hash/table_distribution.rs
src/common/src/hash/consistent_hash/vnode.rs
src/common/src/row/mod.rs
src/common/src/types/jsonb.rs
```

[...]

PR #25345, opened 2026-04-14
fix(common): [BREAK] `Hash for jsonbb::StringRef` shall be same as `&str`

xiangjinwu:

See #25336. In short, it is unfortunate that hash of `Jsonb` (rather than its memcomparable encoding) is persisted and can NEVER change for backward compatibility. This PR patches the `jsonbb` crate to revert its hashing behavior to the same as `&str` before introducing `StringRef` - which was supposed to be an internal change.

**NOTE**:
This "fix" means we are restoring compatibility with <2.7, BUT BREAKING compatibility with 2.7.x and 2.8.x series.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/common/src/types/jsonb.rs
+++ b/src/common/src/types/jsonb.rs
@@ -636,3 +636,24 @@ impl ToSql for JsonbRef<'_> {
         Ok(IsNull::No)
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_hash_backward_compatible() {
+        use std::hash::Hasher as _;
+
+        // Hash of sample input `"foo"` required to be the the magic number below.
+        // See #25336 for how the backward compatibility is doomed.
+        let s = r#""foo""#;
+        let j: JsonbVal = s.parse().unwrap();
+        let expected = expect_test::expect!["10172337927241793445"];
+
+        let mut state = std::hash::DefaultHasher::new();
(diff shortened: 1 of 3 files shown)


chenzl25 (approved):
LGTM, thanks

src/common/src/types/jsonb.rs
```diff
@@ -636,3 +636,24 @@ impl ToSql for JsonbRef<'_> {
...
+
+    #[test]
+    fn test_hash_backward_compatible() {
+        use std::hash::Hasher as _;
+
+        // Hash of sample input `"foo"` required to be the the magic number below.
+        // See #25336 for how the backward compatibility is doomed.
+        let s = r#""foo""#;
+        let j: JsonbVal = s.parse().unwrap();
+        let expected = expect_test::expect!["10172337927241793445"];
```
xiangjinwu, line 652: I may replace this test case with one later that:
* comprehensively cover all data types and special values
* placed closer to vnode calculation
* ensures the hashing algorithm is always `Crc32FastBuilder`

It is here just to demonstrate the effectiveness of the fix contained in this PR, in lieu of a more involved e2e backward compatibility test suite.

chenzl25: Totally agree.

chenzl25:
@xiangjinwu I think we need to cherry-pick this fix to release 2.8 and 2.7 as well

PR #25351, opened 2026-04-14
refactor: switching to rw-fork madsim

tabVersion:
…cross workspaces

- Changed madsim dependency to point to a specific GitHub commit for consistency.
- Updated rdkafka, tonic, and tonic-build to use the same GitHub source for madsim.
- Aligned tokio dependency across multiple Cargo.toml files to use workspace features instead of specifying package versions.
- Ensured all instances of tokio in the project utilize the workspace configuration for better dependency management.

- Closes #[ISSUE_NUMBER]

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -287,7 +287,14 @@ tikv-jemallocator = { git = "https://github.com/risingwavelabs/jemallocator.git"
     "profiling",
     "stats",
 ], rev = "64a2d988d687a94cd859855e19241cd8b0705466" }
-tokio = { package = "madsim-tokio", git = "https://github.com/risingwavelabs/madsim.git", rev = "595455fd05058b3b48bf281bea3ada1bca3d6646" }
+tokio = { package = "madsim-tokio", git = "https://github.com/risingwavelabs/madsim.git", rev = "595455fd05058b3b48bf281bea3ada1bca3d6646", features = [
+    "rt",
+    "rt-multi-thread",
+    "sync",
+    "macros",
+    "time",
+    "signal",
+] }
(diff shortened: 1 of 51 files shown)


src/batch/executors/Cargo.toml
```diff
@@ -37,7 +37,7 @@ risingwave_rpc_client = { workspace = true }
 risingwave_storage = { workspace = true }
 rw_futures_util = { workspace = true }
 
-tokio = { version = "0.2", package = "madsim-tokio", features = [
+tokio = { workspace = true, features = [
```
wenym1: I think we can even put the feature list to the workspace dependency declaration, so that we don't have to always copy the whole list.

Cargo.toml
```diff
@@ -179,7 +179,7 @@ jsonbb = { version = "0.2.3", features = ["float_roundtrip"] }
 jsonwebtoken = { version = "10", features = ["aws_lc_rs"] }
 linkme = { version = "0.3.32", features = ["used_linker"] }
 lru = "0.16"
-madsim = "0.2.34"
+madsim = { git = "https://github.com/risingwavelabs/madsim.git", rev = "595455fd05058b3b48bf281bea3ada1bca3d6646" }
```
wenym1, line 182: seems unnecessary to change the version number in this dependency, because the target will be overwritten by `patch.crate-io`

wenym1 (approved):
Rest LGTM

Cargo.toml
tabVersion: Let's change all occurrence, making it less error prune.

PR #25381, opened 2026-04-16
fix(storage): filter compact table ids in compactor

Li0k:
## Summary
- align normal compactor input filtering with compact table ids instead of existing table ids
- simplify compaction input SST selection with a CompactTaskInputSsts helper
- reuse the precomputed compaction input when deciding task parallelism and copy-block optimization

## Testing
- cargo test -p risingwave_storage test_filter_block_metas --lib
- cargo clippy -p risingwave_storage --lib -- -D warnings


hzxa21 (commented):
Please add some tests.

src/storage/src/hummock/compactor/iterator.rs
```diff
@@ -479,9 +478,18 @@ impl ConcatSstableIterator {
                 None => self.key_range.clone(),
             };
 
+            let compact_sstable_table_ids = {
```
zwang28, line 481: minor, optional: compact_sstable_table_ids.is_empty can replace the variable `found` above.

zwang28 (approved):

hzxa21 (approved):

PR #25391, opened 2026-04-16
fix(optimizer): remove spurious BatchSort above BatchTopN on index with eq-prefix

### LINKED ISSUE #22570: improving optimizer TopNOnIndexRule

chagelo:

Fixes #22570. Supersedes #22581.

When `TopNOnIndexRule` rewrites `LogicalTopN + LogicalScan` into an index scan with equality predicates (e.g. `WHERE a = 1 ORDER BY b LIMIT 10`), the old `clone_with_input_and_prefix` helper prepended the eq-prefix columns to `core.order` (turning `[b ASC]` into `[a ASC, b ASC]`). `BatchTopN::new` then advertised that inflated order as its output order. The parent `BatchSort` insertion logic saw this as "not yet satisfying" the user-requested `[b ASC]` and inserted a redundant `BatchSort` above `BatchTopN`.

The root fix is to stop prepending the eq-prefix in `top_n_on_index_rule.rs` and instead rely on the fact that `BatchSeqScan` already exposes prefix-trimmed order candidates via `orders()`. Concretely:

- **`top_n_on_index_rule.rs`**: `finish_top_n` now calls `clone_with_input` (keeps the original user-requested TopN order) instead of `clone_with_input_and_prefix`. `try_on_pk` simplified to just skip the fixed prefix without collecting it.
- **`logical_topn.rs`**: deleted the now-dead `clone_with_input_and_prefix` helper that was the source of the bug.
- **`batch_topn.rs`**: `one_phase_topn` and `two_phase_topn` now use `input.orders().iter().any(|o| o.satisfies(...))` instead of `input.order().satisfies(...)`, so they recognise the prefix-trimmed order candidates that `BatchSeqScan` already exposes in `orders()`. `two_phase_topn` also passes the TopN order (not `Order::any()`) to the exchange so the global phase can further collapse to `BatchLimit`.
- **`batch_limit.rs`**: `BatchLimit::new` propagates `input.orders()` (all candidates) via `new_batch_with_core_and_orders` so order information flows through limit nodes.

Before this fix, `SELECT * FROM t WHERE a = 1 ORDER BY b LIMIT 10` on an index `(a, b)` produced:
```
BatchSort { order: [b ASC] }          ← spurious
└─BatchTopN { order: [a ASC, b ASC] }
  └─BatchExchange
    └─BatchScan { scan_ranges: [a=1] }
```

After:
```
BatchLimit { limit: 10, offset: 0 }
└─BatchExchange { order: [b ASC] }
  └─BatchLimit { limit: 10, offset: 0 }
    └─BatchScan { scan_ranges: [a=1], limit: 10 }
```

- Closes #22570

## Checklist

## Documentation

<details>

Queries of the form `SELECT ... FROM t WHERE pk_prefix = const ORDER BY remaining_key LIMIT n` on a table with a matching index no longer generate a redundant `BatchSort` above `BatchTopN`. The optimizer now correctly collapses the plan to `BatchLimit → BatchExchange → BatchLimit → BatchScan`, improving query performance.

</details>

### DIFF
--- a/src/frontend/planner_test/tests/testdata/input/index_selection.yaml
+++ b/src/frontend/planner_test/tests/testdata/input/index_selection.yaml
@@ -517,6 +517,20 @@
     select * from t1 where a = 1 order by b desc limit 10
   expected_outputs:
     - batch_plan
+- name: topn index selection + limit pushdown with point get + offset
+  sql: |
+    create table t1 (a int, b int);
+    create index idx1 on t1(a, b desc);
+    select * from t1 where a = 1 order by b desc limit 10 offset 5
+  expected_outputs:
+    - batch_plan
+- name: topn index selection + limit pushdown with point get + with ties
+  sql: |
+    create table t1 (a int, b int);
+    create index idx1 on t1(a, b desc);
+    select * from t1 where a = 1 order by b desc fetch next 10 rows with ties
+  expected_outputs:
+    - batch_plan
 - name: topn index selection + limit pushdown with multi point get
   sql: |
     create table t1 (a int, b int, c int);
@@ -598,6 +612,13 @@
     create table t (c int, b int, a int);
     create index idx on t(a, b desc, c desc);
     select b from t where a = 1 order by b desc, c desc limit 1;
+  expected_outputs:
+    - batch_plan
+- name: topn index selection for scan with output column idx + offset
+  sql: |
+    create table t (c int, b int, a int);
+    create index idx on t(a, b desc, c desc);
+    select b from t where a = 1 order by b desc, c desc limit 2 offset 1;
   expected_outputs:
     - batch_plan
 - name: topn index selection for scan with output column idx
--- a/src/frontend/planner_test/tests/testdata/output/basic_query.yaml
+++ b/src/frontend/planner_test/tests/testdata/output/basic_query.yaml
@@ -193,8 +193,8 @@
     create table t (v1 bigint, v2 double precision);
     select * from t order by v1 limit 1
   batch_plan: |-
-    BatchTopN { order: [t.v1 ASC], limit: 1, offset: 0 }
-    └─BatchExchange { order: [], dist: Single }
+    BatchLimit { limit: 1, offset: 0 }
+    └─BatchExchange { order: [t.v1 ASC], dist: Single }
       └─BatchTopN { order: [t.v1 ASC], limit: 1, offset: 0 }
         └─BatchScan { table: t, columns: [t.v1, t.v2], distribution: SomeShard }
 - sql: |
--- a/src/frontend/planner_test/tests/testdata/output/functional_dependency.yaml
+++ b/src/frontend/planner_test/tests/testdata/output/functional_dependency.yaml
@@ -8,8 +8,8 @@
     └─LogicalProject { exprs: [t1.id, t1.i] }
       └─LogicalScan { table: t1, columns: [t1.id, t1.i, t1._rw_timestamp] }
   batch_plan: |-
-    BatchTopN { order: [t1.id ASC], limit: 2, offset: 0 }
-    └─BatchExchange { order: [], dist: Single }
+    BatchLimit { limit: 2, offset: 0 }
+    └─BatchExchange { order: [t1.id ASC], dist: Single }
       └─BatchLimit { limit: 2, offset: 0 }
         └─BatchScan { table: t1, columns: [t1.id, t1.i], limit: 2, distribution: UpstreamHashShard(t1.id) }
   stream_plan: |-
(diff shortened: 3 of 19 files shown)


chagelo:
@wenym1 @chenzl25 PTAL

PR #25396, opened 2026-04-17
feat(sink): add HTTP sink

wenym1:

This PR adds a new HTTP sink connector that sends append-only rows to an HTTP endpoint with `POST` requests.

### Main changes
- Adds a new `http` sink implementation in `src/connector/src/sink/http.rs`.
- Registers the new sink in `src/connector/src/sink/mod.rs`.
- Adds e2e coverage for the HTTP sink in `e2e_test/sink/http_sink.slt`.
- Adds a lightweight mock HTTP server for e2e verification in `e2e_test/sink/http_sink_mock_server.py`.
- Extends `ci/scripts/e2e-sink-test.sh` to run the HTTP sink e2e test.

### Behavior
- Supports append-only HTTP sink writes.
- Supports single-column payloads with:
  - `varchar` -> request body sent as plain text, default `Content-Type: text/plain`
  - `jsonb` -> request body sent as JSON string, default `Content-Type: application/json`
- Supports custom request headers through `header.*` sink properties.
- Validates unsupported configurations early, including:
  - non-append-only usage
  - multi-column input
  - unsupported column types
  - invalid URL
  - `auto.schema.change = 'true'`

### Why
This provides a simple sink option for pushing RisingWave output to generic HTTP services and webhooks, while keeping the initial scope small and well-validated.

### Tests
- Added e2e test cases for:
  - invalid schema and URL validation
  - auto schema change rejection
  - successful `varchar` and `jsonb` delivery
  - inferred default `Content-Type` behavior
  - custom header propagation

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/src/sink/http.rs
+++ b/src/connector/src/sink/http.rs
@@ -1,4 +1,4 @@
-// Copyright 2025 RisingWave Labs
+// Copyright 2026 RisingWave Labs
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -14,7 +14,7 @@
 
 use std::collections::BTreeMap;
 
-use anyhow::anyhow;
+use anyhow::{Context, anyhow};
 use risingwave_common::array::{Op, StreamChunk};
 use risingwave_common::catalog::Schema;
 use risingwave_common::row::Row;
@@ -108,7 +108,8 @@ fn validate_http_sink(
 
     let parsed_url = url
         .parse()
-        .map_err(|e| SinkError::Config(anyhow!("invalid URL: {}", e)))?;
+        .context("invalid URL")
+        .map_err(SinkError::Config)?;
 
     Ok((col_type, parsed_url))
 }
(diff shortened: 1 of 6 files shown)


wenym1:
* **#25396** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25396?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25396?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/connector/src/sink/http.rs
```diff
@@ -0,0 +1,269 @@
...
+            let name: reqwest::header::HeaderName = k
+                .parse()
+                .with_context(|| format!("invalid header name '{k}'"))
+                .map_err(SinkError::Http)?;
+            let value: reqwest::header::HeaderValue = v
+                .parse()
+                .with_context(|| format!("invalid header value for '{k}'"))
+                .map_err(SinkError::Http)?;
+            header_map.insert(name, value);
+        }
```
hzxa21: Let's do the header the content type validation in `validate` so that we can detect malformed with option earlier.

src/connector/src/sink/http.rs
```diff
@@ -0,0 +1,269 @@
...
+impl TryFrom<SinkParam> for HttpSink {
+    type Error = SinkError;
+
+    fn try_from(param: SinkParam) -> std::result::Result<Self, Self::Error> {
+        let schema = param.schema();
+        let url = param
+            .properties
+            .get("url")
+            .ok_or_else(|| SinkError::Config(anyhow!("missing 'url' in WITH")))?;
+        let (col_type, _) = validate_http_sink(param.sink_type.is_append_only(), &schema, url)?;
```
hzxa21: We should also allow sink with `ignore_delete = true` to be eligible here. See https://github.com/risingwavelabs/risingwave/blob/35a2d166adff672b0937a1773600fcee4af260b6/src/stream/src/executor/sink.rs#L347

src/connector/src/sink/http.rs
```diff
@@ -0,0 +1,269 @@
...
+                        "unexpected column type, expected varchar or jsonb"
+                    )));
+                }
+                None => continue, // skip NULL rows
+            };
+
+            let resp = self
+                .client
+                .post(&self.endpoint)
+                .body(payload)
```
hzxa21, line 239: Just a note for performance. We are sending one row per http request, which can potentially be a performance issue. If there is performance concern, we may need to support batch http sink (sending ndjson or json array for each stream chunk). IIUC, there is no way user can express that via SQL.

We can address this when the demand comes.

hzxa21 (approved):
Rest LGTM

PR #25399, opened 2026-04-17
perf(optimizer): skip order enforcement for singleton batch plans

chagelo:

This PR skips batch required-order enforcement when the optimized batch plan is guaranteed to return at most one row.

Before this change, `BatchOptimizedLogicalPlanRoot::gen_batch_plan` always kept `required_order` intact, so later order enforcement could still insert a redundant `BatchSort` even when `CardinalityVisitor` had already proven the batch plan is a singleton. This shows up in cases such as point lookups with `ORDER BY`, or predicates that are statically false but still carry `ORDER BY`.

This PR clears `required_order` to `Order::any()` after constant evaluation if the batch plan cardinality is at most one. That lets the existing order-enforcement path short-circuit naturally without changing non-singleton behavior.

Planner regressions are added to cover:
- singleton `ORDER BY` planning in `functional_dependency`
- always-false predicate + `ORDER BY` in `order_by`

Validated with:
- `./risedev run-planner-test functional_dependency`
- `./risedev run-planner-test order_by`

- No issue

## Checklist

## Documentation

<details>

Avoid redundant `BatchSort` operators for batch queries that are guaranteed to produce at most one row.

</details>

### DIFF
--- a/src/frontend/planner_test/tests/testdata/input/functional_dependency.yaml
+++ b/src/frontend/planner_test/tests/testdata/input/functional_dependency.yaml
@@ -26,6 +26,8 @@
     - logical_plan
     - stream_plan
     - batch_plan
+    - batch_local_plan
+    - batch_distributed_plan
 # Order key should not be pruned for index,
 # since it uses it as distribution key as well.
 - name: test functional dependency for order key pruning (index)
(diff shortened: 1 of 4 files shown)


chagelo:
@wenym1 PTAL

PR #25401, opened 2026-04-17
fix(sink): support parameterized BigQuery decimal types in schema check

### LINKED ISSUE #25404: bug: BigQuery sink rejects widened numeric destination types for RisingWave NUMERIC columns
## Summary

BigQuery sink schema validation rejects a target BigQuery numeric column when the RisingWave source column is `NUMERIC` and the BigQuery side must use higher precision / scale handling.

## Problem

Users can hit this failure pattern when syncing a RisingWave `NUMERIC` column to BigQuery:

- BigQuery `NUMERIC` supports up to 9 decimal places, so values requiring higher scale cannot fit cleanly.
- Switching the BigQuery destination column to a wider numeric type still fails validation with a type mismatch.

Observed error shape:

```text
Data type mismatch for column "...". BigQuery side: "NUMERIC(31, 2)", RisingWave side: "NUMERIC".
```

This makes some BigQuery sink migrations fail even when the destination side is adjusted to a wider compatible numeric representation.

## Expected Behavior

One of the following should work consistently:

1. RisingWave `NUMERIC` should map cleanly to an appropriate BigQuery numeric type for sink validation.
2. BigQuery sink schema validation should accept compatible destination numeric widening where precision/scale differ but data remains representable.
3. Documentation should clearly state current numeric-type limitations and required workarounds if this behavior is intentional.

## Impact

- Blocks BigQuery sink migrations for tables with higher-precision numeric columns.
- Forces manual schema compromises or prevents using the sink for affected tables.

## Reproduction

1. Create a RisingWave source relation with a `NUMERIC` column containing values requiring more than BigQuery `NUMERIC` scale limits.
2. Create a BigQuery sink targeting a BigQuery table.
3. Use a destination schema that avoids precision loss by widening the numeric type on the BigQuery side.
4. Observe sink creation or validation fail with a numeric type mismatch error.

## Notes

- Related BigQuery sink work recently landed in [#23837](https://github.com/risingwavelabs/risingwave/pull/23837), but this appears to be a separate numeric compatibility issue.

## Source

Filed from an internal support thread. Customer identifiers and environment-specific details were intentionally omitted.

xxhZs:

This PR fixes BigQuery sink schema validation for parameterized decimal types.

BigQuery reports decimal columns as `NUMERIC(p, s)` or `BIGNUMERIC(p, s)` through `INFORMATION_SCHEMA.COLUMNS`. RisingWave sink columns are typed as `NUMERIC`, so the previous exact string comparison could reject compatible existing BigQuery tables.

The schema check now treats BigQuery `NUMERIC(p, s)` and `BIGNUMERIC(p, s)` as compatible decimal type-family matches for RisingWave `NUMERIC`. This PR also aligns the auto-create option with the `create_table_if_not_exists` property name and adds targeted coverage for inline credentials and decimal compatibility.

Closes #25404

## Checklist

## Documentation

<details>

BigQuery sinks now accept parameterized `NUMERIC` and `BIGNUMERIC` columns when validating existing decimal columns, avoiding false schema mismatch errors during sink creation.

</details>

## Testing

- `cargo fmt --check`
- `cargo test -p risingwave_connector --lib --features sink-bigquery test_decimal_type_family_compatibility`
- `cargo test -p risingwave_connector --lib --features sink-bigquery test_decimal_schema_check_accepts_parameterized_numeric_types`
- `cargo test -p risingwave_connector --lib --features sink-bigquery test_bigquery_config_accepts_inline_credentials_json`
- `./risedev d`
- Created BigQuery tables with `NUMERIC(31, 2)` and `BIGNUMERIC(35, 12)` columns and verified `CREATE SINK` succeeds.
- `./risedev k`


src/connector/src/sink/big_query.rs
```diff
@@ -92,7 +92,7 @@ pub struct BigQueryCommon {
     pub dataset: String,
     #[serde(rename = "bigquery.table")]
     pub table: String,
-    #[serde(default)] // default false
+    #[serde(default, rename = "create_table_if_not_exists")] // default false
```
hzxa21, line 95: Should we use alias instead? Otherwise it is a breaking change?

src/connector/src/sink/big_query.rs
```diff
@@ -92,7 +92,7 @@ pub struct BigQueryCommon {
     pub dataset: String,
     #[serde(rename = "bigquery.table")]
     pub table: String,
-    #[serde(default)] // default false
+    #[serde(default, alias = "create_table_if_not_exists")] // default false
```
tabVersion, line 95: unintentional change?

src/connector/src/sink/big_query.rs
```diff
@@ -323,6 +323,22 @@ impl BigQuerySink {
 }
 
 impl BigQuerySink {
+    fn is_decimal_type_compatible(bigquery_type: &str) -> bool {
```
tabVersion, line 326: then how to define `parameterized BigQuery decimal types`?

tabVersion (commented):
can you help share examples about the fix?

src/connector/src/sink/big_query.rs
hzxa21: I think it is intentional based on the PR description.

xxhZs:
@hzxa21  @tabVersion  Can proceed with the review.

hzxa21 (approved):

PR #25405, opened 2026-04-17
fix(connector): drain kafka enumerator events during meta split discovery

tabVersion:
## Summary
Poll the Kafka split enumerator's `BaseConsumer` once at the start of `list_splits()` so meta's periodic source-manager tick can serve queued librdkafka callbacks, including statistics events.

## Why
When `statistics.interval.ms` is enabled, the enumerator path can accumulate queued events because `BaseConsumer` has no background polling thread. Meta keeps a long-lived Kafka enumerator in source manager, so this path is a better fit for a minimal mitigation than producer or reader changes.

## What changed
- added one non-blocking `poll(Duration::ZERO)` in `KafkaSplitEnumerator::list_splits()`
- only log `tracing::warn!` when that poll returns `Some(Err(_))`
- kept split discovery behavior unchanged otherwise

## Validation
- `rustfmt --edition 2024 src/connector/src/source/kafka/enumerator.rs`
- `cargo check -p risingwave_connector --lib --quiet`

## Risk
Narrow. This only touches the Kafka enumerator path used by meta for split discovery and does not add a background task or alter producer/reader behavior.

### DIFF
--- a/src/connector/src/source/kafka/enumerator.rs
+++ b/src/connector/src/source/kafka/enumerator.rs
@@ -182,6 +182,26 @@ impl SplitEnumerator for KafkaSplitEnumerator {
     }
 
     async fn list_splits(&mut self) -> ConnectorResult<Vec<KafkaSplit>> {
+        // `KafkaSplitEnumerator` uses `BaseConsumer`, which does not have a background polling
+        // thread. Poll once per `list_splits` invocation so meta's periodic source-manager tick
+        // can serve queued callbacks like librdkafka statistics events.
+        if let Some(Err(poll_err)) = {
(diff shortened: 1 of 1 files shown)


hzxa21 (approved):
Is `statistics.interval.ms` useful for the kafka consumer in meta? I wonder whether it makes sense to turn off `statistics.interval.ms` for the meta consumer in addition to this PR.

tabVersion:
> Is `statistics.interval.ms` useful for the kafka consumer in meta? I wonder whether it makes sense to turn off `statistics.interval.ms` for the meta consumer in addition to this PR.

make sense, will add in the next commit.

PR #25431, opened 2026-04-21
fix(cdc): avoid aborting auto schema change on non-constant default expressions

wcy-fdu:

wcy/pg_now_schema_change_fix
### Problem

Auto schema change aborts the entire CDC stream when an upstream column has a non-constant default expression (`now()`, `gen_random_uuid()`, `CURRENT_TIMESTAMP`, etc.). Root cause: `ScalarImpl::from_text` only parses simple literals; any function call returns an error which is propagated via `?`, killing the schema change. Users have no clean recovery — they must drop and recreate the source.

### Why a complete fix is hard

Properly handling defaults requires distinguishing "literal constants" from "real expressions" and evaluating each correctly. But PG/MySQL literal grammar is complex (bit / hex / escape strings, arrays, literals with casts, ...) and expression evaluation needs an evaluator. That is feature-level work (extend `ColumnDesc` to carry an `ExprNode`, wire in evaluation), not a hot fix.

### Known limitation

A default that PG treats as a constant (fast-path, no row rewrite) but that `from_text` still cannot parse (e.g. `B'1010'::bit(4)`, `'active'::my_enum`, some array literals) will leave existing rows with `NULL` in the new column. The warn log contains the full expression for diagnosis. This is strictly better than the previous "entire stream dead" behavior.

### Future work

- Process only the delta columns in the schema change event so we can precisely flag which columns actually have NULL existing rows.

## Checklist

## Documentation

<details>

</details>


xiangjinwu (approved):
* Is it possible to verify the new behavior with an e2e test?
* The CI failure is supposed to be fixed by [#25437](https://github.com/risingwavelabs/risingwave/pull/25437) already

xiangjinwu:
~~I was expecting this to fail: https://buildkite.com/risingwavelabs/pull-request/builds/95334~~ (wrong step passed)
failed as expected: https://buildkite.com/risingwavelabs/pull-request/builds/95335

e2e_test/source_legacy/cdc_inline/auto_schema_change_pg.slt
```diff
@@ -61,12 +61,15 @@ distribution key id NULL NULL
...
+# through CDC, not to pin down v7's value.
+query TTTTTTTTTTTT retry 4 backoff 1s
+SELECT id, name, age, v1, v2, v3, v4, v5, v6, v8, v9, v10 from rw_test_schema_change order by id;
 ----
-1 name1 20 1.1 2.2 3.3 f 2020-01-01 12:34:56 2020-01-01 12:34:56 2020-01-01 12:34:56+00:00 1 day {}
-2 name2 21 1.1 2.2 3.3 f 2020-01-01 12:34:56 2020-01-01 12:34:56 2020-01-01 12:34:56+00:00 1 day {}
-3 name3 22 1.1 2.2 3.3 f 2020-01-01 12:34:56 2020-01-01 12:34:56 2020-01-01 12:34:56+00:00 1 day {}
+1 name1 20 1.1 2.2 3.3 f 2020-01-01 12:34:56 2020-01-01 12:34:56+00:00 1 day {}
+2 name2 21 1.1 2.2 3.3 f 2020-01-01 12:34:56 2020-01-01 12:34:56+00:00 1 day {}
+3 name3 22 1.1 2.2 3.3 f 2020-01-01 12:34:56 2020-01-01 12:34:56+00:00 1 day {}
```
xiangjinwu: FYI you can keep using `select *` and just do wildcard match on `v7` with `<slt:ignore>`

PR #25438, opened 2026-04-21
refactor(storage): normalize compact task table ids

Li0k:

- Closes #N/A
- Move stale table-id cleanup to Hummock version-delta apply, so later compact tasks are picked from already-pruned SST metadata instead of rewriting `CompactTask.input_ssts` during task construction.
- Rename the Rust-side delta to `PruneTableIdsFromSsts` while keeping the existing protobuf `truncate_tables` wire format for compatibility. Truncate table and drop/unregister table both use this pruning path; full dynamic-group removal still uses `GroupDestroy`.
- Keep `CompactTask::existing_table_ids` as version live/member table ids. Task-local readable table ids now come from `input_ssts[*].table_ids`, and `CompactTask::build_compact_table_ids()` is removed.
- Simplify compactor-side handling around this invariant: do not register `StateCleanUpCompactionFilter` on the main path, read only SSTs with non-empty table ids, and keep block-level table-id pruning as the physical-SST safety guard.
- Add/update tests for version SST metadata pruning, truncate/drop apply paths, unregister table pruning, compact-task read semantics, and drop-table compaction behavior.

## Checklist

## Documentation

<details>

N/A

</details>

### DIFF
--- a/src/frontend/src/catalog/system_catalog/rw_catalog/rw_hummock_compact_task_assignment.rs
+++ b/src/frontend/src/catalog/system_catalog/rw_catalog/rw_hummock_compact_task_assignment.rs
@@ -14,6 +14,7 @@
 
 use risingwave_common::types::{Fields, JsonbVal};
 use risingwave_frontend_macro::system_catalog;
+use risingwave_hummock_sdk::compact_task::CompactTask;
 use risingwave_pb::id::CompactionGroupId;
 use serde_json::json;
 
@@ -35,6 +36,7 @@ struct RwHummockCompactTaskAssignment {
     target_sub_level_id: i64,
     compression_algorithm: i32,
     table_ids: JsonbVal,
+    existing_table_ids: JsonbVal,
 }
 
 #[system_catalog(table, "rw_catalog.rw_hummock_compact_task_assignment")]
@@ -44,9 +46,18 @@ async fn read(reader: &SysCatalogReaderImpl) -> Result<Vec<RwHummockCompactTaskA
 
     let mut rows = vec![];
     for compact_task_assignment in compact_task_assignments {
-        let compact_task = compact_task_assignment.compact_task.unwrap();
+        let compact_task = CompactTask::from(compact_task_assignment.compact_task.unwrap());
 
         let select_level = compact_task.input_ssts[0].level_idx;
+        let table_ids = compact_task
+            .get_table_ids_from_input_ssts()
+            .map(|table_id| table_id.as_raw_id())
+            .collect::<Vec<_>>();
+        let existing_table_ids = compact_task
+            .existing_table_ids
+            .iter()
+            .map(|table_id| table_id.as_raw_id())
+            .collect::<Vec<_>>();
 
         rows.push(RwHummockCompactTaskAssignment {
             compaction_group_id: compact_task.compaction_group_id,
@@ -60,7 +71,8 @@ async fn read(reader: &SysCatalogReaderImpl) -> Result<Vec<RwHummockCompactTaskA
             target_file_size: compact_task.target_file_size as _,
             target_sub_level_id: compact_task.target_sub_level_id as _,
             compression_algorithm: compact_task.compression_algorithm as _,
-            table_ids: json!(compact_task.existing_table_ids).into(),
+            table_ids: json!(table_ids).into(),
+            existing_table_ids: json!(existing_table_ids).into(),
         });
     }
 
--- a/src/meta/src/hummock/manager/compaction/compact_task_builder.rs
+++ b/src/meta/src/hummock/manager/compaction/compact_task_builder.rs
@@ -1,0 +1,240 @@
+// Copyright 2026 RisingWave Labs
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::collections::{BTreeMap, HashMap, HashSet};
+use std::sync::Arc;
+
+use itertools::Itertools;
+use risingwave_common::catalog::TableId;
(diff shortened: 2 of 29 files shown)


zwang28 (approved):
LGTM

zwang28 (approved):

src/frontend/src/catalog/system_catalog/rw_catalog/rw_hummock_compact_task_assignment.rs
```diff
@@ -60,7 +65,7 @@ async fn read(reader: &SysCatalogReaderImpl) -> Result<Vec<RwHummockCompactTaskA
             target_file_size: compact_task.target_file_size as _,
             target_sub_level_id: compact_task.target_sub_level_id as _,
             compression_algorithm: compact_task.compression_algorithm as _,
-            table_ids: json!(compact_task.existing_table_ids).into(),
+            table_ids: json!(table_ids).into(),
```
hzxa21, line 74: Given that we don't deprecate `existing_table_ids`, I think it is better we display them as well for better o11y.

src/meta/src/hummock/manager/compaction/compact_task_builder.rs
```diff
@@ -0,0 +1,240 @@
...
+    for input_level in &mut *input_ssts {
+        for sst in &mut input_level.table_infos {
+            let mut inner = sst.get_inner();
+            inner
+                .table_ids
+                .retain(|table_id| table_id_set.contains(table_id));
+            sst.set_inner(inner);
+        }
+    }
+}
```
hzxa21: This works, but I wonder if we’re fixing this at the wrong layer.

The cleanup in here is essentially the same operation as `truncate_tables` in [hummock_version_ext.rs](https://github.com/risingwavelabs/risingwave/blob/e3fac55a60a6e319148c0c21e07af038a8ddae0a/src/storage/hummock_sdk/src/compaction_group/hummock_version_ext.rs#L1344): remove unregistered `table_ids` from `SstableInfo`, and then treat empty SSTs specially. That makes me think `drop table` and `truncate table` should probably be unified from the Hummock version’s perspective.

Instead of repairing stale `SstableInfo.table_ids` in `get_compact_task`, could we do the metadata cleanup when a table is unregistered? In other words, when `unregister_table_ids` happens, apply a version-level delta similar to `TruncateTables` to remove those table ids from SST metadata directly.

I think that would have a few advantages:

- the Hummock version stays self-consistent, instead of relying on task construction to patch it up later
- we avoid having compaction-task semantics diverge from version semantics
- the compactor can treat normalized `input_ssts[*].table_ids` as a normal invariant, instead of a defensive cleanup result

So my preference would be to push this cleanup into version update / unregister flow if possible, and keep `get_compact_task` focused on building tasks from already-normalized version state.

hzxa21: Open an issue to track this idea: https://github.com/risingwavelabs/risingwave/issues/25507

zwang28 (approved):

src/meta/src/hummock/manager/transaction.rs
```diff
@@ -197,7 +197,9 @@ impl<'a> HummockVersionTransaction<'a> {
                 .or_default()
                 .group_deltas;
 
-            group_deltas.push(GroupDelta::TruncateTables(table_ids.into_iter().collect()));
+            group_deltas.push(GroupDelta::PruneTableIdsFromSsts(
+                table_ids.into_iter().collect(),
+            ));
```
hzxa21, line 202: Not introduced by this PR:

Correct me if I am wrong, it seems that it is possible to have a both NewL0SubLevel and PruneTableIdsFromSsts group delta in the same version delta.

In build_sst_delta_infos used by cache refill, we currently continue unless all group deltas are IntraLevel or NewL0SubLevel. In that case we skip building SstDeltaInfo for the whole group, so the inserted SSTs in the same commit won’t participate in refill.

Will this be a problem? This is relevant to the meta cache sharding thing we discussed previously as well.

src/storage/src/hummock/compactor/compaction_utils.rs
```diff
@@ -173,13 +166,8 @@ pub fn build_multi_compaction_filter(compact_task: &CompactTask) -> MultiCompact
...
         CompactionFilterFlag::from_bits(compact_task.compaction_filter_mask).unwrap_or_default();
-    if compaction_filter_flag.contains(CompactionFilterFlag::STATE_CLEAN) {
-        let state_clean_up_filter = Box::new(StateCleanUpCompactionFilter::new(
-            HashSet::from_iter(compact_task.existing_table_ids.clone()),
-        ));
-
-        multi_filter.register(state_clean_up_filter);
-    }
+    // STATE_CLEAN is handled by compact-task table-id normalization and iterator-level block
+    // pruning. Keep the flag as a no-op here for compatibility with existing task configs.
```
hzxa21, line 170: For the SST created prior to this PR, they can contain table ids that are dropped. How do we guarantee compatibility? I suspect we need to do a onetime clean up on the table ids in sstable info on meta startup.

src/storage/hummock_sdk/src/compaction_group/hummock_version_ext.rs
```diff
@@ -91,6 +91,49 @@ impl<L> HummockVersionCommon<SstableInfo, L> {
...
+        if live_table_ids.is_empty()
+            && self.levels.values().any(|levels| {
+                #[expect(deprecated)]
+                {
+                    !levels.member_table_ids.is_empty()
+                }
+            })
+        {
+            return 0;
+        }
```
hzxa21, line 111: I saw that [may_fill_backward_state_table_info](https://github.com/risingwavelabs/risingwave/blob/fee1709abddc743bebd48d5fe7509d2559747ebd/src/meta/node/src/server.rs#L537) has existed for a long time. cc @wenym1 to confirm. Do we still need to handle the case when a cluster is upgraded from a version with `stateb_table_info` not present but `levels.member_table_ids` present?

- If we do need to handle such case, I suggest we move `prune_stale_table_ids_from_ssts` before `may_fill_backward_state_table_info`. This is to make sure the "no stale table ids in ssts" assumption always hold in new version. It will not hold if this check is hit.
- If we don't need to handle such case, we can change the current check into an assertion.

wenym1: The migration is introduced in v1.10.0, which is already out of support. I think we can ignore this case, and even remove the migration logic. I will raise a PR to deprecate the migration logic and remove the deprecated field completely.

PR #25440, opened 2026-04-21
fix(sink): ensure barrier progress must be truncated on log store

wenym1:

This PR fixes sink log reader truncation progress when barriers are yielded but not explicitly truncated before a later truncate offset arrives.

Previously, a log reader could yield multiple barriers and then receive a truncate call on a later chunk or barrier offset. In that case, earlier yielded barriers could be skipped instead of being truncated in order. That breaks the expectation that all yielded barrier progress is reflected in log store truncation.

This PR introduces a generic `TruncateBarrierLogReader` wrapper in `src/connector/src/sink/log_store.rs` that:

- records yielded barrier offsets,
- buffers pending untruncated barriers,
- truncates all earlier pending barriers before applying the requested truncate offset,
- clears its pending state on `init`, `start_from`, and `rewind`.

The wrapper is then applied to both current log store factories:

- `BoundedInMemLogStoreFactory`
- `KvLogStoreFactory`

This keeps the behavior generic and avoids duplicating barrier-truncation bookkeeping inside individual reader implementations.

The PR also adds unit tests for:
- truncating earlier barriers before a later barrier truncate,
- preserving later barriers when truncating a chunk offset first.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/connector/src/sink/log_store.rs
+++ b/src/connector/src/sink/log_store.rs
@@ -217,6 +217,31 @@ pub struct TransformChunkLogReader<F: Fn(StreamChunk) -> StreamChunk, R: LogRead
     inner: R,
 }
 
+pub struct TruncateBarrierLogReader<R: LogReader> {
+    inner: R,
+    pending_barriers: VecDeque<TruncateOffset>,
+}
+
+impl<R: LogReader> TruncateBarrierLogReader<R> {
+    pub fn new(inner: R) -> Self {
+        Self {
+            inner,
+            pending_barriers: VecDeque::new(),
+        }
+    }
+
+    fn truncate_pending_barriers_before(&mut self, offset: TruncateOffset) -> LogStoreResult<()> {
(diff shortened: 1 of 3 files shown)


wenym1:
* **#25203** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25203?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#25442** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25442?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#25440** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25440?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25440?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

hzxa21 (approved):
LGTM

PR #25442, opened 2026-04-21
fix(sink): update kv log store value indices after schema change

wenym1:

This PR fixes auto schema change for decoupled sinks backed by the kv log store.

Previously, when `ALTER TABLE ... ADD COLUMN` triggered sink schema change, the internal log store table schema was updated with the new `columns`, but its `value_indices` could remain stale in the rewritten table metadata and in the catalog update path. That led to incorrect serialization/deserialization for rows written after the schema change:
- newly added column values could be dropped from retained log rows
- scans on the internal log store table could fail with type mismatches because columns were read with misaligned indices

This PR fixes that by:
- rebuilding the log store table `value_indices` after extending the schema during sink schema change
- carrying the full rewritten `PbTable` through the auto-refresh schema flow instead of only forwarding the new columns
- persisting both `columns` and `value_indices` when finishing the streaming job replacement in meta
- adding regression coverage in both the kv log store test and the madsim sink schema change test

The new tests cover:
- batch scan behavior after schema change, ensuring old-schema rows are skipped and new-schema rows remain readable
- an end-to-end decoupled sink schema change flow that verifies retained log rows expose the newly added column correctly after `ALTER TABLE ADD COLUMN`

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/common/log_store_impl/kv_log_store/writer.rs
+++ b/src/stream/src/common/log_store_impl/kv_log_store/writer.rs
@@ -162,17 +162,6 @@ impl<LS: LocalStateStore> LogWriter for KvLogStoreWriter<LS> {
         }
         flush_info.report(&self.metrics);
 
-        if let Some(truncate_offset) = self.tx.pop_truncation(epoch) {
-            self.last_truncate_offset = Some(truncate_offset);
-        }
-        let post_seal_epoch = self.state.seal_current_epoch(
-            next_epoch,
-            self.last_truncate_offset
-                .map(|(epoch, seq_id)| {
-                    LogStoreVnodeProgress::Aligned(self.state.vnodes().clone(), epoch, seq_id)
-                })
-                .unwrap_or(LogStoreVnodeProgress::None),
-        );
         let has_schema_change = options.schema_change.is_some();
         // Barrier's new_vnode_bitmap field does not need to be passed to log-reader, because when sink is decoupled, we
         // always rebuild sink when update vnode bitmap.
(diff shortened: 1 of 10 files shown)


wenym1:
* **#25203** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25203?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#25442** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25442?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25442?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#25440** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25440?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/stream/src/common/log_store_impl/kv_log_store/writer.rs
```diff
@@ -188,6 +177,17 @@ impl<LS: LocalStateStore> LogWriter for KvLogStoreWriter<LS> {
             assert_eq!(truncate_offset, (epoch, None));
             self.last_truncate_offset = Some(truncate_offset);
         }
+        if let Some(truncate_offset) = self.tx.pop_truncation(epoch) {
+            self.last_truncate_offset = Some(truncate_offset);
+        }
+        let post_seal_epoch = self.state.seal_current_epoch(
```
wenym1, line 183: call `seal_current_epoch` to be after `wait_for_barrier_truncation` during schema change, to ensure that the truncate_offset at the epoch of schema change can be applied to `seal_current_epoch` to ensure that all pending data in the log store have been truncated.

wenym1:
Although the LoC in this PR is large, most changes are about test.

hzxa21 (approved):
LGTM

xxhZs (approved):
lGTM

PR #25441, opened 2026-04-21
fix(dml): wait for persistence before completing fast insert in dml executor

wenym1:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## What's changed and what's your intention?

- This PR extracts the DML executor and webhook persistence-waiting part from `yiming/test-tapdata`.
- Fast insert no longer waits for persistence in the batch RPC layer by returning an epoch and then calling `try_wait_epoch` outside the DML path.
- Instead, `WriteHandle` now supports waiting for durable persistence directly, and `DmlExecutor` holds persistence notifiers until the closing barrier's previous epoch has passed `try_wait_epoch`.
- This keeps the persistence waiting mechanism inside the normal DML execution flow and makes webhook-triggered fast inserts complete only after their data is durably persisted.
- The executor builder and related tests are updated to pass the state store into `DmlExecutor` and verify the new persistence behavior.
- This PR intentionally excludes the new websocket ingest implementation and the new RPC method introduced in the original branch.

## Checklist

- [ ] I have written necessary rustdoc comments.
- [ ]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  I have checked the [Release Timeline](https://github.com/risingwavelabs/rw-commits-history/blob/main/release_timeline.md) and [Currently Supported Versions](https://docs.risingwave.com/changelog/release-support-policy#support-end-dates-for-recent-releases) to determine which release branches I need to cherry-pick this PR into. 

## Documentation

- [ ]  My PR needs documentation updates. 

<details>
<summary><b>Release note</b></summary>

</details>

PR #25443, opened 2026-04-21
feat(frontend): support TLS for webhook endpoint and multi-column webhook tables

wenym1:

This PR extends webhook ingestion in two directions:

1. Webhook sources can now ingest typed multi-column rows directly from JSON payloads instead of requiring a single `jsonb` column.
2. The webhook service can serve HTTPS when `RW_SSL_CERT` and `RW_SSL_KEY` are configured, using the same TLS configuration as pgwire.

Breaking change:

- On a cluster where `RW_SSL_CERT` and `RW_SSL_KEY` are already set, pre-existing webhook sources will start serving over HTTPS instead of HTTP after this change. Existing HTTP webhook senders must switch to HTTPS for webhook delivery.

For multi-column webhook tables, RisingWave now reads the request body as JSON and decodes each top-level field into the target table schema. This allows users to define normal typed columns such as `int`, `varchar`, `timestamp`, `timestamptz`, `time`, `decimal`, and primary keys, while still keeping the existing webhook signature validation flow.

The request handling flow is now:

- Resolve the webhook target table from `/webhook/<database>/<schema>/<table>`.
- Authenticate the request by evaluating `VALIDATE AS secure_compare(...)` against the raw request body and request headers.
- Build the row payload:
  - For legacy single-`jsonb` webhook tables, keep the existing raw-payload behavior.
  - For typed multi-column webhook tables, decode JSON into a full row using the table schema.
- Send the fully materialized chunk to fast insert without the old `column_indices` path.

How to use the feature:

```sql
CREATE TABLE orders (
  id INT,
  customer_name VARCHAR,
  amount DOUBLE PRECISION,
  PRIMARY KEY (id)
) WITH (
  connector = 'webhook'
) VALIDATE AS secure_compare(
  headers->>'x-hub-signature',
  'sha1=' || encode(hmac('TEST_WEBHOOK', payload, 'sha1'), 'hex')
);
```

Then send a webhook request to:

```text
POST /webhook/<database>/<schema>/<table>
```

Example:

```bash
curl -X POST "http://127.0.0.1:4560/webhook/dev/public/orders" \
  -H "content-type: application/json" \
  -H "x-hub-signature: sha1=<signature>" \
  -d '{"id":1234,"customer_name":"Alice","amount":99.99}'
```

Protocol and requirements:

- Transport:
  - HTTP is still supported by default.
  - HTTPS is enabled when both `RW_SSL_CERT` and `RW_SSL_KEY` are set.
  - `RW_SSL_CERT` and `RW_SSL_KEY` must be set together.
  - `RW_SSL_ENFORCE` now also requires both TLS env vars to be present.
- Request path:
  - `/webhook/<database>/<schema>/<table>`
- Request body for typed multi-column webhook tables:
  - Single-row mode: one JSON object per request.
  - Batched mode: set `WITH (..., is_batched = true)` and send newline-delimited JSON objects.
- Decoder protocol for typed multi-column JSON payloads:
  - `x-rw-webhook-format: plain` (default)
  - `x-rw-webhook-encode: json` (default)
  - `x-rw-webhook-json-timestamp-handling-mode: milli | guess_number_unit`
  - `x-rw-webhook-json-timestamptz-handling-mode: utc_string | utc_without_suffix | micro | milli | guess_number_unit`
  - `x-rw-webhook-json-time-handling-mode: milli | micro`
  - `x-rw-webhook-json-bigint-unsigned-handling-mode: long | precise`
  - `x-rw-webhook-json-handle-toast-columns: true | false`
- Validation behavior:
  - Signature validation still runs on the raw request body bytes, not the decoded row.
  - For multi-column webhook tables, `secure_compare(...)` should reference `payload` as the raw body input.
  - Existing single-`jsonb` webhook tables keep the previous validation behavior and can still reference the actual jsonb column name.
- Table requirements and limitations:
  - Missing non-primary-key fields are padded as `NULL`.
  - Missing primary-key fields are rejected during decoding.
  - Generated columns are not supported for webhook tables.
  - Default values are not supported for webhook tables.
  - Inline column options other than `NULL` are rejected for webhook tables.
  - Single-column `jsonb` webhook tables remain supported and intentionally bypass the typed JSON decoder path.

This PR also updates the JSON parser configuration path so webhook requests can pass request-scoped JSON decoding options through headers, and adds E2E coverage for multi-column inserts, batching, timestamp/time/timestamptz options, precise bigint handling, and invalid decoder headers.

- Closes #[ISSUE_NUMBER]

## Checklist

## Documentation

<details>

Webhook sources can now ingest typed multi-column JSON payloads directly, including batched JSON-lines requests and request-scoped JSON decoding options for timestamp, timestamptz, time, bigint unsigned, and TOAST handling.

Webhook endpoints also serve HTTPS when `RW_SSL_CERT` and `RW_SSL_KEY` are configured. This is a breaking change for existing webhook sources on clusters that already set those environment variables, because webhook traffic will move from HTTP to HTTPS and senders must be updated accordingly.

</details>

### DIFF
--- a/src/frontend/src/lib.rs
+++ b/src/frontend/src/lib.rs
@@ -238,7 +238,7 @@ pub fn start(
         let webhook_listen_addr = opts.webhook_listen_addr.parse().unwrap();
         let tcp_keepalive =
             TcpKeepalive::new().with_time(Duration::from_secs(opts.tcp_keepalive_idle_secs as _));
-        let tls_config = TlsConfig::new_default();
+        let tls_config = TlsConfig::new_default().unwrap();
 
         let session_mgr = Arc::new(SessionManagerImpl::new(opts).await.unwrap());
         SESSION_MANAGER.get_or_init(|| session_mgr.clone());
--- a/src/frontend/src/lib.rs
+++ b/src/frontend/src/lib.rs
@@ -238,6 +238,7 @@ pub fn start(
         let webhook_listen_addr = opts.webhook_listen_addr.parse().unwrap();
         let tcp_keepalive =
             TcpKeepalive::new().with_time(Duration::from_secs(opts.tcp_keepalive_idle_secs as _));
+        let tls_config = TlsConfig::new_default();
 
         let session_mgr = Arc::new(SessionManagerImpl::new(opts).await.unwrap());
         SESSION_MANAGER.get_or_init(|| session_mgr.clone());
@@ -257,14 +258,15 @@ pub fn start(
             frontend_config.max_single_query_size_bytes,
         ));
 
-        let webhook_service = crate::webhook::WebhookService::new(webhook_listen_addr);
+        let webhook_service =
+            crate::webhook::WebhookService::new(webhook_listen_addr, tls_config.clone());
         let _task = tokio::spawn(webhook_service.serve());
         pg_serve(
             &listen_addr,
             tcp_keepalive,
             session_mgr.clone(),
             ConnectionContext {
-                tls_config: TlsConfig::new_default(),
+                tls_config,
                 redact_sql_option_keywords: Some(redact_sql_option_keywords),
                 message_memory_manager,
             },
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -24,7 +24,7 @@ dependencies = [
  "arrow-schema 57.3.0",
  "libloading 0.8.9",
  "toml 0.9.12+spec-1.1.0",
- "windows-registry 0.6.1",
+ "windows-registry",
  "windows-sys 0.61.2",
 ]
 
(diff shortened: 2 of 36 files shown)


wenym1:
* **#25444** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25444?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#25443** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25443?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25443?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#25441** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25441?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

chenzl25 (approved):
LGTM!

src/frontend/src/lib.rs
```diff
@@ -238,6 +238,7 @@ pub fn start(
         let webhook_listen_addr = opts.webhook_listen_addr.parse().unwrap();
         let tcp_keepalive =
             TcpKeepalive::new().with_time(Duration::from_secs(opts.tcp_keepalive_idle_secs as _));
+        let tls_config = TlsConfig::new_default().unwrap();
```
hzxa21, line 241: Note: this is a breaking change for pre-existing webhook source on a cluster with `RW_SSL_CERT` and `RW_SSL_KEY` set.

hzxa21 (approved):
LGTM

src/frontend/src/lib.rs
hzxa21: Please update the release note section for this breaking change.

PR #25444, opened 2026-04-21
feat(frontend): add WebSocket ingest with async acks

wenym1:

This PR adds a WebSocket-based ingest endpoint for `webhook` tables, and makes it reuse the same webhook-side JSON decoding logic that is already used by the HTTP webhook path for multi-column payloads.

The main intention is to support low-latency streaming DML ingestion for webhook tables, while keeping the row decoding semantics consistent with the existing webhook endpoint and source connector JSON decoder. In particular:

- Add a WebSocket endpoint for webhook-table ingestion at `/ingest/{database}/{schema}/{table}`.
- Authenticate the connection with the same webhook signature validation logic already used by the HTTP webhook endpoint.
- Validate the init timestamp against the configured clock skew window.
- Accept batched DML requests over the WebSocket connection.
- Reuse the webhook/source JSON accessor path to decode JSON payloads into typed rows for multi-column tables, instead of maintaining a separate handwritten JSON-to-datum conversion path.
- Keep ingest-specific semantics outside the decoder:
  - monotonic `dml_batch_id`
  - `upsert` / `delete` op handling
  - batch-level ack / fatal reporting
  - connection termination on any error

The implementation keeps the webhook HTTP logic close to its original layout, and moves the WebSocket endpoint under `src/frontend/src/webhook/websocket.rs` so the webhook-related entrypoints stay together.

### How to use the feature

1. Create a webhook table with signature validation, for example:

```sql
create table ws_orders (
  id int,
  customer_name varchar,
  amount double precision,
  primary key (id)
) with (
  connector = 'webhook'
) validate as secure_compare(
  headers->>'x-rw-signature',
  'sha256=' || encode(hmac('TEST_WEBSOCKET', payload, 'sha256'), 'hex')
);
```

2. Open a WebSocket connection to:

```text
ws://<frontend-host>:4560/ingest/<database>/<schema>/<table>
```

Example:

```text
ws://127.0.0.1:4560/ingest/dev/public/ws_orders
```

3. Put the webhook signature in the handshake headers. The signature is computed from the raw init message body, using the same secret/validation expression as the HTTP webhook endpoint.

Example header:

```text
x-rw-signature: sha256=<hmac_of_init_message>
```

4. Send the first WebSocket text frame as the authenticated init message:

```json
{"type":"init","timestamp":1760000000000}
```

5. After init succeeds, send DML batches as JSON objects:

```json
{
  "dml_batch_id": 1,
  "items": [
    {"op":"upsert","data":{"id":1,"customer_name":"Alice","amount":99.99}},
    {"op":"delete","data":{"id":1}}
  ]
}
```

#### Client -> Server: init frame

The first text frame must be:

```json
{"type":"init","timestamp":1760000000000}
```

Rules:
- `type` must be `init`
- `timestamp` must be a non-negative epoch millisecond
- the timestamp must be within the configured webhook auth clock-skew window
- the init frame is authenticated by the webhook signature header

If init authentication or validation fails, the server returns a fatal message and closes the connection.

#### Client -> Server: DML batch

After init, each text frame must be a JSON object with a monotonic batch id and an `items` array:

```json
{
  "dml_batch_id": 1,
  "items": [
    {"op":"upsert","data":{"id":1,"name":"foo"}},
    {"op":"delete","data":{"id":1}}
  ]
}
```

Fields:
- `dml_batch_id: u64`
  - must increase monotonically across the connection
- `items: [DML request]`
  - may be empty; an empty batch is valid and is acked immediately
- `op: string`
  - `upsert`, `insert`, and `update` are treated as upsert/insert semantics
  - `delete` is treated as delete semantics
- `data: json`
  - JSON object for normal multi-column webhook tables
  - scalar JSON is accepted for single-`jsonb` payload schema

#### Server -> Client: ack

Each accepted batch is acked once:

```json
{"ack":1}
```

For non-empty batches, the ack is sent after the batch is accepted downstream. Empty batches are acked immediately by the frontend.

#### Server -> Client: fatal error

Any error returns a fatal message and closes the connection:

```json
{"fatal":"failed to prepare DML batch: dml_batch_id 2 item 1: failed to decode webhook JSON payload"}
```

This covers cases such as:
- missing required PK fields
- type conversion failures
- malformed row payload for the table schema
- malformed batch payloads

Examples:
- invalid DML payload
- invalid init signature
- invalid/stale init timestamp
- malformed init message
- invalid decoder option headers
- ingest stream/channel failures

### JSON decoding behavior

For multi-column webhook tables, this PR makes the WebSocket ingest path reuse the same webhook/source JSON decoder path as the HTTP webhook endpoint.

That means:
- typed row decoding follows the existing source-connector JSON accessor semantics
- decoder option headers are shared with webhook JSON decoding
- PK fields are enforced through the same accessor-based row extraction logic
- non-PK columns may be omitted and become `NULL`
- delete payloads can contain only PK columns

Supported decoder option headers on the WebSocket handshake include the existing webhook JSON options, such as:
- `x-rw-webhook-json-timestamp-handling-mode`
- `x-rw-webhook-json-timestamptz-handling-mode`
- `x-rw-webhook-json-time-handling-mode`
- `x-rw-webhook-json-bigint-unsigned-handling-mode`
- `x-rw-webhook-json-handle-toast-columns`

### Rust unit tests

Added/updated WebSocket ingest unit tests to cover:
- valid init parsing
- stale init timestamp rejection
- missing `op` rejection
- empty batch is accepted as valid protocol input
- building one `StreamChunk` from a batched DML payload
- returning the first failing batch/item location in the error string
- allowing `delete` with only PK columns
- rejecting incomplete composite PK on `upsert`
- rejecting incomplete composite PK on `delete`
- applying supported decoder option headers
- rejecting invalid decoder option headers
- returning the first type-conversion error with the correct batch/item location
- accepting scalar JSON for single-`jsonb` payload schema

### WebSocket e2e tests

Added `e2e_test/webhook/websocket_ingest.slt` and its Python driver to cover:
- happy path with batched `upsert` and `delete`
- empty batch ack
- `delete` with only PK columns
- missing PK fatal closes the stream without writing a row
- incomplete composite PK on `upsert` returns fatal and closes the stream without writing a row
- incomplete composite PK on `delete` returns fatal, closes the stream, and leaves the seeded row unchanged
- type-conversion fatal closes the stream without writing a row
- timestamp decoding with `x-rw-webhook-json-timestamp-handling-mode: milli`
- time decoding with `x-rw-webhook-json-time-handling-mode: milli`
- bigint precise decoding with `x-rw-webhook-json-bigint-unsigned-handling-mode: precise`
- invalid init signature -> fatal
- invalid init timestamp -> fatal
- invalid decoder option header -> fatal

### CI coverage

The new WebSocket e2e test is added to `ci/scripts/e2e-source-test.sh`, alongside the existing webhook source e2e test, so it runs in CI together with the webhook endpoint coverage.

- Closes #[ISSUE_NUMBER]

## Checklist

## Documentation

<details>

Added a WebSocket ingest endpoint for `webhook` tables. Clients can now stream authenticated batched `upsert` / `delete` JSON DMLs to `/ingest/{database}/{schema}/{table}` using batch-level `dml_batch_id` acknowledgements, with the same signature validation and JSON decoding behavior as the existing webhook endpoint.

</details>


wenym1:
* **#25444** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25444?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25444?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#25443** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25443?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#25441** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25441?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

wenym1:
Most LoC are about tests.

src/frontend/src/webhook/websocket.rs
```diff
@@ -0,0 +1,1030 @@
...
+    let session_mgr = SESSION_MANAGER.get().expect("session manager initialized");
+    let max_clock_skew_ms = session_mgr
+        .env()
+        .frontend_config()
+        .webhook_auth_max_clock_skew_ms;
+    let webhook_source_info = table_info.webhook_source_info;
+    let table_id = table_info.table_id;
+    let table_version_id = table_info.table_version_id;
+    let compute_client = table_info.compute_client;
+    let payload_schema = table_info.payload_schema;
```
hzxa21, line 240: `acquire_table_info()` also carries `row_id_index` for webhook tables whose DML schema includes the hidden row-id column. The HTTP webhook path forwards that into `FastInsertRequest`, and `FastInsertExecutor` reinserts a `NULL` serial column before writing. This WebSocket path drops that information here, and `IngestDml` later writes the chunk unchanged, so websocket ingest will not match the existing HTTP behavior for webhook tables without a user-defined PK (for example the existing single-`jsonb` webhook-table shape). Can we thread `row_id_index` through the ingest request and apply the same hidden-column fixup on the compute side?

hzxa21 (commented):
One inline comment about the missing `row_id_index` wiring in the WebSocket ingest path.

src/frontend/src/webhook/websocket.rs
wenym1: fixed

chenzl25 (approved):
Generally LGTM

src/frontend/src/webhook/websocket.rs
```diff
@@ -0,0 +1,1047 @@
...
+            Some(ingest_dml_response::Response::Init(_)) => {}
+            _ => {
+                return Err("unexpected init response from ingest stream".to_owned());
+            }
+        },
+        Ok(None) => return Err("ingest stream closed during init".to_owned()),
+        Err(e) => {
+            return Err(format!("ingest stream init error: {}", e.as_report()));
+        }
+    }
```
hzxa21, line 303: I think there is a small race here around ack delivery.

`tokio::select!` is racing `ws_rx.next()` against `ingest_resp_stream.message()`. If batch N has already been persisted and its ack is ready, but the next WebSocket frame is malformed (or fails local validation), the `ws_msg` branch can win first and close the connection before we forward the pending ack for batch N.

This does not introduce a new correctness issue by itself, since the protocol is not exactly-once anyway, but it does make duplicate replays more likely: the client may reconnect without having seen the ack for a batch that was already accepted.

Would it make sense to prefer draining ready acks before failing on a later input error, so we reduce unnecessary duplicate retries?

hzxa21 (approved):
Rest LGTM!

PR #25453, opened 2026-04-22
fix(pgwire): make JWK alg optional and pin verification to an allow-list

xuhui-lu:
## Summary

- **RFC 7517 compliance** — `alg` is OPTIONAL per RFC 7517 §4.4, but our `Jwk` deserializer required it, so any JWKS that omits `alg` fails to parse (a legitimate JWKS today won't authenticate at all).
- **Alg-confusion hardening** — when the JWK doesn't pin an `alg`, we now fall back to the header's `alg` but only after checking it against a server-side allow-list (RS/PS 256/384/512). The `DecodingKey` is RSA-only, and the allow-list blocks the classic `alg: none` / `alg: HS256`-forged-with-public-key confusion attacks even if someone later constructs a `DecodingKey` from a different key type.
- Behaviour when the JWK *does* pin an `alg` is unchanged: header must match it.

## Test plan

New unit tests in `src/utils/pgwire/src/pg_server.rs`:

- [x] `test_jwt_with_jwk_missing_alg_succeeds` — JWK without `alg` + RS256 header + valid signature → accepted
- [x] `test_jwt_with_jwk_missing_alg_rejects_disallowed_header_alg` — JWK without `alg` + HS256 header → rejected by allow-list before signature check
- [x] `test_jwt_alg_mismatch_between_header_and_jwk` — JWK pins RS384 but header declares RS256 → rejected

All 11 tests in `pg_server::tests::jwt_validation_tests` pass; `cargo clippy -p pgwire --tests -- -D warnings` is clean.


hzxa21 (approved):
LGTM

PR #25470, opened 2026-04-23
chore(meta): promote create streaming job lifecycle logs to info

yuhao-su:

Today, when a user runs `CREATE MATERIALIZED VIEW` (or any other streaming-job DDL), the default meta log at `INFO` is essentially silent on success — every lifecycle message inside the DDL controller / stream manager is at `DEBUG`. Operators have no way to tell from the log whether meta received the job, whether actors were built, or whether the first barrier was collected, unless they flip the whole meta to `DEBUG`.

This PR promotes three per-DDL log sites from `debug!` to `info!`, chosen to bracket the three externally observable phases of job creation:

- `ddl_controller.rs` — `starting streaming job` (meta accepted the DDL and began processing)
- `stream_manager.rs` — `first barrier collected for stream job` (job is actually running downstream)
- `stream_manager.rs` — `stream job finish` (creation completed)

Intermediate implementation-detail logs (`building streaming job`, `built actors finished`, `sending Command::CreateStreamingJob`) are intentionally left at `debug!` — they fire on the same code path as the three above and would only add noise at `INFO`.

Each of these logs fires at most once per streaming job, so the extra volume at `INFO` is negligible, while the debuggability win for "job stuck in creation" investigations is significant.

## Checklist

- [ ]  I have checked the Release Timeline and Currently Supported Versions to determine which release branches I need to cherry-pick this PR into.

## Documentation

<details>

Meta now logs the start, first-barrier, and finish of streaming-job creation at `INFO` level instead of `DEBUG`, so operators can see per-DDL progress in the default meta log.

</details>

### DIFF
--- a/src/meta/src/rpc/ddl_controller.rs
+++ b/src/meta/src/rpc/ddl_controller.rs
@@ -1064,7 +1064,7 @@ impl DdlController {
             }
         };
         let job_id = streaming_job.id();
-        tracing::debug!(
+        tracing::info!(
             id = %job_id,
             definition = streaming_job.definition(),
             create_type = streaming_job.create_type().as_str_name(),
--- a/src/meta/src/rpc/ddl_controller.rs
+++ b/src/meta/src/rpc/ddl_controller.rs
@@ -1064,7 +1064,8 @@ impl DdlController {
             }
         };
         let job_id = streaming_job.id();
-        tracing::info!(
+        tracing::info!(id = %job_id, "starting streaming job");
+        tracing::debug!(
             id = %job_id,
             definition = streaming_job.definition(),
             create_type = streaming_job.create_type().as_str_name(),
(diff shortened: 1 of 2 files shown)


src/meta/src/rpc/ddl_controller.rs
```diff
@@ -1064,7 +1064,7 @@ impl DdlController {
             }
         };
         let job_id = streaming_job.id();
-        tracing::debug!(
+        tracing::info!(
             id = %job_id,
             definition = streaming_job.definition(),
```
hzxa21, line 1070: Will printing out job definition cause log being too large?

yuhao-su: My initial purpose was to connect the frontend ddl event and meta ddl event. But it actually fall into the scope of tracing.
let's hold the pr for now

yuhao-su: I updated the pr to only add a job id as info.
tracing is a separate issue here

hzxa21 (approved):
LGTM

PR #25476, opened 2026-04-24
refactor(connector): unify Postgres connection config across sink/source/batch/frontend

wcy-fdu:
## Summary

Dedup the 4 Postgres connection config structs and the duplicated `create_pg_client` / `PgConnectOptions` builder / TLS wiring.

- `PgConnectionConfig` is now the canonical struct. Sink / CDC external configs keep their user-facing serde shape and project via a `pg_connection_config()` method. `PostgresConnectionParams` in the batch executor is removed.
- `TcpKeepaliveConfig` moved from `sink::postgres` to `connector_common::postgres` (fixes the reverse dep).
- `create_pg_client` and `PostgresExternalTable::{connect, type_mapping}` now take `&PgConnectionConfig` instead of 8-10 loose args.
- `cdc/enumerator::query_postgres_lsns` replaces 30 lines of manual property extraction with `pg_connection_config_from_properties`.

## Behavior note

`cdc/enumerator::query_postgres_lsns` previously looked up `database.ssl.root.cert`, which doesn't match the `ssl.root.cert` key used by every other path (sink, source, validator). Now unified to `ssl.root.cert`. Only observable when the PG CDC source uses `verify-ca` / `verify-full` with a user-provided root cert.

## Test plan
- [x] CI green


chenzl25 (approved):
LGTM

PR #25480, opened 2026-04-24
fix(meta): clean dropped streaming jobs outside post collect

### LINKED ISSUE #25466: bug: cancel job doesn't notify hummock correctly
### Describe the bug

`HummockManager::unregister_table_ids` is not called for `CANCEL JOB` when streaming_use_snapshot_backfill=true, regardless of whether background_ddl is enabled or disabled.

It triggers a compactor crash loop that requires a cluster recovery to resolve.

### Error message/log

```text

```

### To Reproduce

```
alter system set streaming_parallelism to 1;
alter system set backfill_rate_limit to 1;
create table t (k int primary key, v int);
insert into t  select i,i from generate_series(1,10000) as i;
create materialized view mv as select * from t;

--- Wait for seconds to allow the MV commits some data.

-- see multiple tables, expected.
select member_tables from rw_hummock_compaction_group_configs;

show jobs;
cancel job [id];

-- expect only 1 table, but still see multiple tables. Because cancel job doesn't notify hummock to remove dropped state tables.
select member_tables from rw_hummock_compaction_group_configs;

-- forcefully purge stale tables from hummock.
recover;

-- see 1 tables, expected.
select member_tables from rw_hummock_compaction_group_configs;
```

### Expected behavior

_No response_

_No response_

### The version of RisingWave

v2.8

### Additional context

_No response_

wenym1:

This PR fixes the dropped-state-table cleanup path for streaming job drop/cancel flows, especially the snapshot backfill cancel and recovery cases from #25466.

Previously, `unregister_table_ids` and `complete_dropped_tables` were tied to barrier-manager post-collect handling, while some cancel paths could bypass the expected cleanup sequence. That could leave state table IDs in Hummock version metadata after the catalog state had already been removed, which later caused compactor failures when resolving table info.

This PR changes the flow as follows:

- Moves dropped streaming job cleanup into the issuing side / recovery side, instead of relying on `PostCollectCommand::DropStreamingJobs`.
- Introduces a shared cleanup helper to consistently perform:
  - refresh progress tracker removal
  - `unregister_table_ids`
  - `complete_dropped_tables`
- Applies the shared cleanup logic to:
  - normal drop streaming job flow
  - background snapshot backfill cancel flow
  - recovery pre-apply drop/cancel flow
  - replace-stream cleanup path for dropped state tables
- Keeps `PostCollectCommand::DropStreamingJobs` only as a checkpoint marker and removes the redundant cleanup from post-collect.
- Adds / updates simulation integration tests to cover:
  - snapshot backfill cancel unregistering table IDs
  - snapshot backfill cancel + database recovery + manual compaction repro path

This makes the cleanup behavior consistent even when snapshot backfill cancel goes through a separate path, and prevents stale table IDs from remaining in Hummock version metadata after the corresponding tables are already dropped from catalog state.

- Closes #25466

## Checklist

## Documentation

<details>

Fixed a meta cleanup issue that could leave stale state table IDs after snapshot backfill job cancel / recovery, which could later cause compactor failures during table info resolution.

</details>

### DIFF
--- a/src/meta/node/src/server.rs
+++ b/src/meta/node/src/server.rs
@@ -527,7 +527,9 @@ pub async fn start_service_as_election_leader(
             env.clone(),
             metadata_manager.clone(),
             barrier_scheduler.clone(),
+            hummock_manager.clone(),
             source_manager.clone(),
+            refresh_manager.clone(),
             scale_controller.clone(),
         )
         .unwrap(),
--- a/src/meta/src/barrier/checkpoint/state.rs
+++ b/src/meta/src/barrier/checkpoint/state.rs
@@ -799,9 +799,9 @@ impl DatabaseCheckpointControl {
 
             Some(Command::DropStreamingJobs {
                 streaming_job_ids,
-                unregistered_state_table_ids,
                 unregistered_fragment_ids,
                 dropped_sink_fragment_by_targets,
+                ..
             }) => {
                 let actors = self
                     .database_info
@@ -835,10 +835,7 @@ impl DatabaseCheckpointControl {
                     table_ids,
                     None,
                     node_actors,
-                    PostCollectCommand::DropStreamingJobs {
-                        streaming_job_ids,
-                        unregistered_state_table_ids,
-                    },
+                    PostCollectCommand::DropStreamingJobs,
                 )
             }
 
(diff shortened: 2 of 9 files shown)


wenym1:
* **#25480** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25480?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25480?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

zwang28 (approved):

PR #25481, opened 2026-04-24
fix(meta): avoid panic when dropping initial job

### LINKED ISSUE #25452: bug: meta panic on dropping initial job
### Describe the bug

Root cause:
[The assertion](https://github.com/risingwavelabs/risingwave/blob/f49ef55bb80c1628c48d879100f3c93c7e9a998f/src/meta/src/rpc/ddl_controller.rs#L1596) is actually broken because now catalog is sent to frontend before the job is updated from JobStatus::Initial to JobStatus::Creating, regardless of whether background_ddl is enabled or disabled.

Related: 
- https://github.com/risingwavelabs/risingwave/pull/22928
- https://github.com/risingwavelabs/risingwave/pull/20460
- https://github.com/risingwavelabs/risingwave/pull/22435

### Error message/log

```text
Job with Initial status should not notify frontend and therefore should not arrive here
```

### To Reproduce

Use [this branch](https://github.com/risingwavelabs/risingwave/tree/wangzheng/reproduce_25452), which intentionally pauses the collection of the MV creation barrier.

```
-- The creation should succeed.
create table t (k int primary key, v int);

-- The creation is blocked intentionally in this branch, i.e. the sleep in hash join executor.
create materialized view panic_mv as select t1.k,t2.v from t t1 join t t2 on t1.k=t2.v;

-- Verify the job creation is not finished.
show jobs;

-- Optionally verify in meta store the job_status is INITIAL
-- select * from streaming_job;

drop materialized view panic_mv;
-- Confirm meta panics: "internal error: entered unreachable code: Job with Initial status should not notify frontend and therefore should not arrive here"

```

### Expected behavior

_No response_

_No response_

### The version of RisingWave

v2.8.1

### Additional context

#24609
#22928

wenym1:

- Fix the meta `DROP` path for streaming jobs that are still in `Initial` status by aborting them through `try_abort_creating_streaming_job(..., true)` instead of panicking on an outdated assumption.
- Add a regression test that creates an initial materialized view job, aborts it with cancellation semantics, verifies the create waiter receives a cancellation, and checks the catalog rows are removed.
- Closes #25452

## Checklist

## Documentation

<details>

Meta no longer panics when a user drops a streaming job that is still in `Initial` status during creation.

</details>

### DIFF
--- a/src/meta/src/controller/catalog/test.rs
+++ b/src/meta/src/controller/catalog/test.rs
@@ -14,8 +14,10 @@
 
 #[cfg(test)]
 mod tests {
+    use risingwave_meta_model::table::HandleConflictBehavior;
     use risingwave_pb::catalog::StreamSourceInfo;
     use risingwave_pb::catalog::subscription::SubscriptionState;
+    use tokio::sync::oneshot;
 
     use crate::controller::catalog::*;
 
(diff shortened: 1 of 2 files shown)


wenym1:
* **#25481** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25481?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25481?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

zwang28 (approved):

Li0k (approved):

PR #25519, opened 2026-04-28
refactor: replace #[allow] with #[expect] across the codebase

wenym1:

- Replace eligible `#[allow(lint)]` attributes with `#[expect(lint)]` so clippy reports stale suppressions automatically.
- Keep `#[allow(...)]` in macro expansions, generated code, and non-code contexts where `#[expect(...)]` is not appropriate or produces unhelpful failures.
- For `clippy::unused_async`, remove `async` and corresponding `.await` call sites where feasible instead of suppressing the lint.
- Preserve `async` plus `#[expect(clippy::unused_async)]` only where trait or macro interfaces require an async signature, such as parser helpers reused by `only_parse_payload!`.
- Address follow-up review and CI issues by reverting macro-only lint conversions, restoring selected crate-level and generated-code annotations, fixing feature-gated UDF module compilation for `rw-static-link`, and keeping the branch aligned with `cargo clippy` across both feature configurations.

- No linked issue.

## Checklist

## Documentation

<details>

This PR is an internal lint-suppression cleanup and follow-up CI/review fix. It does not change user-facing behavior.

</details>

## Test plan

- `cargo clippy --all-targets --features rw-static-link --no-default-features`
- `cargo test -p risingwave_connector allow_alter_on_fly_fields_rust_up_to_date -- --nocapture`


yuhao-su (approved):

src/frontend/src/optimizer/optimizer_context.rs
```diff
@@ -134,8 +134,7 @@ impl OptimizerContext {
 
     // TODO(TaoWu): Remove the async.
```
wenym1: remove the comment

src/object_store/src/object/mod.rs
```diff
@@ -12,10 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#![expect(
-    unexpected_cfgs,
-    reason = "feature(hdfs-backend) is banned https://github.com/risingwavelabs/risingwave/pull/7875"
-)]
+#![allow(unexpected_cfgs)]
```
wenym1: revert it

yuhao-su:
Hopefully there can be a check to disallow the use of `allow`

src/object_store/src/object/mod.rs
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#![expect(
+#![allow(
```
wenym1: change it back to expect

PR #25522, opened 2026-04-28
feat: 2/10 add RBAC role membership storage contract

tabVersion:
Part of the re-cut RBAC stack. This stack was split so each PR is an independently reviewable functional slice around ~1k changed lines, with e2e coverage added where the SQL runtime becomes available.

Review-only fixes included in this recut:
- keep #25521 as parser/fallback-only instead of carrying the full runtime;
- move meta revoke/drop tests out of the production-code PR;
- replace stale `role options are not supported yet` frontend tests with positive CREATEROLE/INHERIT coverage once frontend support lands;
- restore role membership ddl e2e coverage in the frontend runtime test slice.

Local verification run for the recut:
- `git diff --check origin/main ralph/rbac-v2-10-pg-catalog-compat`

No compile/build/e2e run in this recut pass.

### DIFF
--- a/src/meta/model/src/user.rs
+++ b/src/meta/model/src/user.rs
@@ -32,6 +32,7 @@ pub struct Model {
     pub can_create_user: bool,
     pub can_login: bool,
     pub is_admin: bool,
+    pub can_inherit: bool,
     pub auth_info: Option<AuthInfo>,
 }
 
@@ -60,6 +61,7 @@ impl From<PbUserInfo> for ActiveModel {
             can_create_user: Set(user.can_create_user),
             can_login: Set(user.can_login),
             is_admin: Set(user.is_admin),
+            can_inherit: Set(user.can_inherit.unwrap_or(true)),
             auth_info: Set(user.auth_info.as_ref().map(AuthInfo::from)),
         }
     }
@@ -75,6 +77,7 @@ impl From<Model> for PbUserInfo {
             can_create_user: val.can_create_user,
             can_login: val.can_login,
             is_admin: val.is_admin,
+            can_inherit: Some(val.can_inherit),
             auth_info: val.auth_info.map(|x| x.to_protobuf()),
             grant_privileges: vec![], // fill in later
         }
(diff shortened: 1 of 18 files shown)


tabVersion:
Heads up for the stacked branch: #25521 now includes parser fix `6a8bbd3ca8` for privilege-keyword role names (`connect`, `create`, `select`, `usage`) in role-membership `GRANT` / `REVOKE`.

Because this PR is based on the previous #25521 head and also touches `src/sqlparser/src/parser.rs`, please rebase/carry that parser fix before merging this step; otherwise the stack can reintroduce the old first-token dispatch behavior. While rebasing, also keep the `parse_role_option_for_clause` lookahead that lets `REVOKE admin FROM user_a` / `REVOKE inherit FROM user_a` / `REVOKE set FROM user_a` parse as role names when they are not followed by `OPTION FOR`.

src/meta/model/src/user.rs
```diff
@@ -60,6 +61,7 @@ impl From<PbUserInfo> for ActiveModel {
             can_create_user: Set(user.can_create_user),
             can_login: Set(user.can_login),
             is_admin: Set(user.is_admin),
+            can_inherit: Set(user.can_inherit),
```
tabVersion: Because `can_inherit` is a proto3 `bool`, any caller that builds `PbUserInfo { name, ..Default::default() }` now persists `NOINHERIT` through this conversion. The frontend create path sets `can_inherit: true`, but the meta/controller test fixture still uses that default shape, and downstream grant semantics use the member's `can_inherit` as the default for `GRANT role TO member` when `WITH INHERIT` is omitted. That means direct meta-created test users silently get `inherit_option = false`, which diverges from the PostgreSQL/default contract this PR is introducing. Please update the server-side role/user fixtures or constructors to set the default `can_inherit: true` and add/keep a regression around an omitted-`WITH INHERIT` role grant.

tabVersion (commented):
Stack-aware RBAC review after rebasing this PR onto #25521 head `69f15214e2`: the new role-membership read contract mostly matches the intended 2/7 boundary, but I found one default-inheritance edge that should be fixed before this slice becomes the base for the grant-authority PR.

PR #25536, opened 2026-04-29
feat(compaction): support compaction for all watermark type 

### LINKED ISSUE #24076: feat: support compaction for all vnode watermark type

Currently compaction only work for [WatermarkSerdeType::PkPrefix](https://github.com/risingwavelabs/risingwave/blob/wangzheng/watermark-iter/src/meta/src/hummock/compaction/selector/vnode_watermark_selector.rs#L90).

_No response_

_No response_

### Additional context

_No response_

zwang28:

This PR introduces a new field `max_watermark_column_value` in SstableInfo.
- It will be Some when table_ids.len() ==1 when builidng the SST and the table has watermark.
- It will be None when table_ids.len() !=1.

The vnode watermark compaction will pick and drop SSTs whose `max_watermark_column_value`  < current watermark of the table.

And the resulted compaction task will all be trivial reclaim compaction.

- Closes #24076

## Checklist

## Documentation

<details>

</details>


src/storage/src/compaction_catalog_manager.rs
```diff
@@ -543,13 +547,72 @@ impl CompactionCatalogAgent {
...
+                } else {
+                    Ok(None)
+                }
+            }
+            WatermarkSerdeType::Value => {
+                let Some(value_serde) = self.value_watermark_serde(table_id) else {
+                    tracing::warn!(?table_id, ?watermark_type, "Failed to get watermark serde.");
+                    return Ok(None);
+                };
+                if let HummockValue::Put(val) = value {
```
hzxa21, line 608: I think this introduces a correctness gap for `WatermarkSerdeType::Value`.

In the SST-level path we derive `max_watermark_column_value` only from `Put` records and return `None` for tombstones, but the existing row-wise value-watermark compaction logic explicitly keeps tombstones (`ValueSkipWatermarkState::should_delete` returns `false` for non-`Put`). That safeguard matters because a delete tombstone may still be needed to hide an older lower-level version of the same PK.

A concrete example:
- lower level SST: `Put(id=1, ts='01:40')`
- newer SST: `Delete(id=1)` plus `Put(id=2, ts='01:41')`
- current watermark: `01:45`

This code will compute the SST max from the `Put(id=2, ts='01:41')` only, so meta can trivially reclaim the whole SST. But that also drops `Delete(id=1)`, and then the older lower-level `Put(id=1, ts='01:40')` becomes visible again.

Can we either exclude `Value` watermark SSTs from the trivial reclaim path, or otherwise prove that dropping tombstones at SST granularity is safe here?

src/storage/src/hummock/sstable/builder.rs
```diff
@@ -238,6 +251,10 @@ impl<W: SstableWriter, F: FilterBuilder> SstableBuilder<W, F> {
         largest_key: Vec<u8>,
         mut meta: BlockMeta,
     ) -> HummockResult<bool> {
+        // Values cannot be accessed from raw blocks.
```
hzxa21, line 254: This looks like it can permanently lose the new watermark metadata on the fast-compaction raw-block path.

`add_raw_block` clears `max_watermark_column_value` immediately, and `fast_compactor_runner` uses this path whenever the current watermark / filters cannot delete the block yet. So an SST can:
1. already have a correct `max_watermark_column_value`
2. go through a fast compaction that raw-copies its blocks
3. come out with `max_watermark_column_value = None`

After that, a later vnode-watermark reclaim has no SST-level metadata to match on anymore, even if the watermark advances enough to reclaim the SST.

Should we preserve / merge the copied block's source SST watermark metadata here, or disable raw-block copy for watermark-tracked SSTs if we cannot preserve it?

hzxa21 (commented):
Thanks for the change. I left two inline comments for correctness / feature-retention concerns that looked worth addressing before relying on the new SST-level watermark reclaim path.

src/storage/src/compaction_catalog_manager.rs
zwang28: Good catch.
I'll refactor it to only perform trivial reclaim in the lowest level, restoring the behavior from prior to this PR

zwang28: https://github.com/risingwavelabs/risingwave/pull/25536/changes/c11a58c939e2d39f39a2b21eb456c70bd1811c5a

src/storage/src/hummock/sstable/builder.rs
zwang28: > disable raw-block copy for watermark-tracked SSTs

Refactored.

https://github.com/risingwavelabs/risingwave/pull/25536/changes/b5c1afb206b7d4764afd2fe0a40d872d05026bdb

zwang28: > Should we preserve / merge the copied block's source SST watermark metadata here

This method is achievable. I applied "disable raw-block copy for watermark-tracked SSTs" instead because
- It requires simpler implementation.
- It triggers early garbage collection when only a subset of Key-Value pairs falls below the watermark.

hzxa21:
Do we still need this PR?

PR #25537, opened 2026-04-29
fix(iceberg): skip empty chunks in fetch executor to avoid unwrap panic

ericnoam:
## Summary
Fix a panic in the Iceberg fetch executor that occurs when processing empty or fully-deleted files, by skipping chunks when none were produced for a task.

## Context
The `IcebergFetchExecutor` iterates over Iceberg data files and collects `StreamChunk`s for each. After collecting, it calls `chunks.last().unwrap()` to retrieve the last chunk's metadata (file path and position) before yielding. If a file produces zero rows (e.g., an empty file or one where all rows were deleted), `chunks` will be empty and `unwrap()` panics at runtime.

## Changes
- Added an early `continue` in the per-file loop inside `IcebergFetchExecutor::fetch_iceberg_data` when `chunks` is empty, skipping files that produced no data.

### Key Implementation Details
The guard is inserted immediately after the chunk-collection loop and before the `chunks.last().unwrap()` call (`src/stream/src/executor/source/iceberg_fetch_executor.rs`, lines 384–387). The comment explains the intent: empty or fully-deleted files are valid Iceberg states and should be silently skipped rather than causing a panic.

## Testing
```bash
# Build the stream crate
cargo build -p risingwave_stream

# Run unit tests for the executor module
cargo test -p risingwave_stream -- iceberg_fetch
```

To reproduce the original panic, point an Iceberg source at a table containing an empty data file (or a v2 table with all rows deleted via equality-delete files) and run a streaming job that scans it.

Sample error log:
```
thread 'rw-streaming' (2696642) panicked at src/stream/src/executor/source/iceberg_fetch_executor.rs:384:44:
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


*** await tree context of current task ***

Actor 1694: `some_name` [3.243s]
  Epoch 10496055749705728 [161.237ms]
    Project 69E00000016 [162.969ms]
      Filter 69E00000015 [162.964ms]
        RowIdGen 69E00000014 [162.964ms]
          StreamFsFetch 69E00000013 [162.964ms]  <== current
            Merge 69E00000012 [162.974ms]
              LocalInput (actor 1748) [162.968ms]
```


hzxa21:
Thanks for the contribution!

I saw that the PR branch is based on 2.8 instead of main. Can you develop the fix based on main?

ericnoam:
Will close and make a cleaner PR

PR #25538, opened 2026-04-29
fix(iceberg): prevent stuck state on empty tasks in fetch executor

ericnoam:
## Summary
Fix a panic — and a follow-on state-management bug — in the Iceberg fetch executor when a `FileScanTask` produces no rows (empty data file or fully equality-deleted file).

## Context
`IcebergFetchExecutor::build_batched_stream_reader` iterates over Iceberg data files, collects `StreamChunk`s for each, then calls `chunks.last().unwrap()` to extract the file path and read position from the last row before yielding a `ChunksWithState`.  If a file produces zero rows the `Vec` is empty and `unwrap()` panics.

The initial fix used `if chunks.is_empty() { continue; }` to skip the yield.  That silenced the panic but introduced a second bug: skipping the yield means `into_stream` never receives a `ChunksWithState` for that task, so it never:
- decrements `splits_on_fetch`, and
- calls `state_store_handler.delete(&data_file_path)`.

The file assignment is left stuck in the state table, preventing subsequent batches from being built and processed.

## Changes

**`src/stream/src/executor/source/iceberg_fetch_executor.rs`**

- Removed the `continue` guard.
- `task.data_file_path` is now captured before the scan (the task is moved into `scan_task_to_chunk_with_deletes`).
- After the scan a `ChunksWithState` is **always** yielded:
  - Non-empty task: `data_file_path` and `last_read_pos` are extracted from the last chunk's metadata columns (unchanged behaviour).
  - Empty task: `data_file_path` falls back to the path captured from the task; `last_read_pos` is `None`.
- The `into_stream` handler's `for chunk in &chunks` loop is a no-op for an empty vec, so no spurious rows are emitted, but state cleanup always runs.
- Replaced the unit test that re-implemented the fix logic in a local closure with two tests that directly construct `ChunksWithState` values and assert on the properties `into_stream` depends on.

## Testing
```bash
# Build the affected crate
cargo build -p risingwave_stream

# Run executor unit tests
cargo test -p risingwave_stream -- iceberg_fetch
```

To reproduce the original panic: point an Iceberg source at a table that contains an empty data file (or an Iceberg v2 table where all rows in a file are deleted via equality-delete files) and run a streaming job that scans it.

Sample error log from the original panic:
```
thread 'rw-streaming' (2696642) panicked at src/stream/src/executor/source/iceberg_fetch_executor.rs:384:44:

*** await tree context of current task ***

Actor 1694: `some_name` [3.243s]
  Epoch 10496055749705728 [161.237ms]
    Project 69E00000016 [162.969ms]
      Filter 69E00000015 [162.964ms]
        RowIdGen 69E00000014 [162.964ms]
          StreamFsFetch 69E00000013 [162.964ms]  <== current
            Merge 69E00000012 [162.974ms]
              LocalInput (actor 1748) [162.968ms]
```


ericnoam:
Probably better if I add a `test`

tabVersion (commented):
Thanks for working on this! I agree with the main direction of the fix: in the Iceberg fetch executor, `ChunksWithState` is not only a data carrier, but also the per-`FileScanTask` completion signal. So even when a task produces no rows, yielding an empty `ChunksWithState` is the right way to let `into_stream` decrement `splits_on_fetch` and delete the persisted file assignment. Falling back to `task.data_file_path` also matches the key used by the list/fetch state-table path.

One thing I think we should still harden before merging is the interaction with the lower-level source/parser/chunk contracts. The current fallback only covers the case where the scan produces no chunks at all:

```rust
if let Some(last_chunk) = chunks.last() {
    let last_row = last_chunk.row_at(last_chunk.cardinality() - 1).1;
    ...
} else {
    (task_data_file_path, None)
}
```

However, `scan_task_to_chunk_with_deletes` converts every Iceberg/Arrow `RecordBatch` into a `DataChunk` and yields it without filtering `row_count == 0`. The common byte-stream parser path often avoids emitting empty chunks, but `StreamChunk`/`SourceChunkStream` itself does not seem to provide a global “non-empty chunk” guarantee, and this Iceberg path bypasses the normal parser chunk builder. If the Iceberg reader emits a zero-row `RecordBatch` after predicate/delete filtering, we can get a non-empty `chunks` vec containing a zero-row `StreamChunk`; then `chunks.last()` is `Some`, but `last_chunk.cardinality()` is `0`, so `cardinality() - 1` can still panic before the state cleanup runs.

Could we make the fallback condition be “no readable row” rather than only “no chunk”? A minimal shape could be to skip zero-capacity chunks before pushing them, and use `capacity()` for the `StreamChunk::from_parts` ops length to match the physical-row contract, similar to the batch Iceberg fetch path:

```rust
let chunk = chunk?;
if chunk.capacity() == 0 {
    continue;
}
chunks.push(StreamChunk::from_parts(
    itertools::repeat_n(Op::Insert, chunk.capacity()).collect_vec(),
    chunk,
));
```

Then the existing `task_data_file_path` fallback would cover both “no batches” and “only zero-row batches”. Alternatively, we could derive the state from the last chunk with `capacity() > 0` and fall back to the task path if none exists.

The current tests document the intended empty `ChunksWithState` shape, but they do not exercise this production path or the zero-row chunk case. It may be worth adding a small helper-level test around deriving the completion state from `(task path, chunks)` so regressions like reintroducing a skip/continue or missing zero-row chunks are caught.

So overall: the state-machine fix looks architecturally right to me, but I think we should cover this zero-row chunk/source-contract edge before calling it fully safe.

src/stream/src/executor/source/iceberg_fetch_executor.rs
```diff
@@ -701,3 +722,114 @@ impl<S: StateStore> Debug for IcebergFetchExecutor<S> {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use risingwave_common::array::{DataChunk, Op, StreamChunk};
+
+    use super::*;
```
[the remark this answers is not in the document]
ericnoam: Fixed in 835b0a6. A couple of notes on the diagnosis:

- The **E0252 duplicate-import claim is a false positive** — `use super::*` only surfaces items that are `pub`/`pub(crate)` *defined* in the parent module. Private `use` statements in the parent (like `use risingwave_common::array::{DataChunk, Op, ...}`) are not re-exported through a glob import, so there was no actual name collision.
- That said, the underlying suggestion to narrow the glob is correct and was applied: `use super::*` → `use super::ChunksWithState`.
- Also caught and fixed a latent issue you did not flag: `test_zero_cardinality_chunks_are_excluded` calls `.collect_vec()` (an `Itertools` trait method). Since `Itertools` is only privately imported in the parent module it would not have been brought in by either `super::*` or the new narrowed import, causing a compile error when the tests are actually run. Added `use itertools::Itertools;` explicitly to cover this.

src/stream/src/executor/source/iceberg_fetch_executor.rs
```diff
@@ -389,20 +392,38 @@ impl<S: StateStore> IcebergFetchExecutor<S> {
                 Some(metrics.clone()),
             ) {
                 let chunk = chunk?;
+                // Skip zero-cardinality chunks: a RecordBatch with 0 visible rows after
+                // predicate/delete filtering would cause `cardinality() - 1` to underflow
+                // below when we extract the last-row metadata. Skipping here is safe
+                // because the existing `task_data_file_path` fallback already covers
+                // the case where no readable rows are produced.
+                if chunk.cardinality() == 0 {
```
tabVersion, line 400: maybe `capacity`?

tabVersion (approved):
LGTM, thanks for contribution

src/stream/src/executor/source/iceberg_fetch_executor.rs
ericnoam: Maybe the zero-visible-row guard should stay based on cardinality(), but the ops vector should use capacity() so that StreamChunk ops matches the physical chunk length? Wdyt?

src/stream/src/executor/source/iceberg_fetch_executor.rs
```diff
@@ -389,20 +392,38 @@ impl<S: StateStore> IcebergFetchExecutor<S> {
...
+                    continue;
+                }
                 chunks.push(StreamChunk::from_parts(
                     itertools::repeat_n(Op::Insert, chunk.cardinality()).collect_vec(),
```
ericnoam, line 404: @tabVersion I mean leaving L400 as it is but change here to:
`itertools::repeat_n(Op::Insert, chunk.capacity()).collect_vec(),`
So that th vector matches physical length. wdyt?

chenzl25 (approved):
LGTM

PR #25547, opened 2026-05-01
fix(stream): restore committed watermark by natural value

yuhao-su:

When initializing a `StateTable`, the committed watermark is restored by picking the most-progressed value across vnodes. Previously this used `.max()` on the raw serialized bytes:

```rust
let max_watermark_of_vnodes = distribution
    .vnodes()
    .iter_vnodes()
    .filter_map(|vnode| local_state_store.get_table_watermark(vnode))
    .max();
```

This is only correct for `WatermarkDirection::Ascending`. For descending columns (i.e. `PkPrefix` / `NonPkPrefix` watermark whose column is sorted DESC), memcomparable encoding inverts byte order, so byte-level `.max()` actually returns the **least**-progressed watermark in natural value space. After recovery the state table would resume from a watermark behind where cleanup actually happened.

This PR deserializes each vnode's watermark and picks the largest by natural value via `cmp_datum(_, _, OrderType::ascending())`. The comparison is direction-agnostic: a watermark always progresses upward in natural value space; `WatermarkDirection` only flips the byte encoding.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/stream/src/common/table/state_table.rs
+++ b/src/stream/src/common/table/state_table.rs
@@ -40,7 +40,7 @@ use risingwave_common::types::{DataType, ScalarImpl};
 use risingwave_common::util::column_index_mapping::ColIndexMapping;
 use risingwave_common::util::epoch::EpochPair;
 use risingwave_common::util::row_serde::OrderedRowSerde;
-use risingwave_common::util::sort_util::OrderType;
+use risingwave_common::util::sort_util::{OrderType, cmp_datum};
 use risingwave_common::util::value_encoding::BasicSerde;
 use risingwave_hummock_sdk::HummockReadEpoch;
 use risingwave_hummock_sdk::key::{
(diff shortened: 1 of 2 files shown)


yuhao-su:
* **#25547** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25547?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25547?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

wenym1 (approved):
LGTM assuming that `a watermark always progresses upward in natural value space` is correct.

In practice, in what SQL will we have descending watermark direction? I ask codex, but it says in our current code base we won't generate such case.

yuhao-su:
> LGTM assuming that `a watermark always progresses upward in natural value space` is correct.
> 
> In practice, in what SQL will we have descending watermark direction? I ask codex, but it says in our current code base we won't generate such case.

That's a good point. Currently there is no way to write query can trigger this bug. I'll remove the block 3.0 tag.

PR #25479, opened 2026-04-24
chore(deps): upgrade deltalake and align arrow

Li0k:
I hereby agree to the terms of the [RisingWave Labs, Inc. Contributor License Agreement](https://raw.githubusercontent.com/risingwavelabs/risingwave/17af8a747593ebdbfa826691daf75bdab7d14fa0/.github/contributor-license-agreement.txt).

## What's changed and what's your intention?

Upgrade Delta Lake sink dependencies and align RisingWave-facing Arrow adapters with Arrow 58.

- Upgrade `deltalake` to `0.32.0`.
- Upgrade ADBC Snowflake crates to `0.23` and switch the local Snowflake Arrow adapter from Arrow 57 aliases to Arrow 58 aliases.
- Remove the old Arrow 57 adapter module from `risingwave_common`.
- Upgrade `gcloud-bigquery` to `1.6.0`.
- Bump the pinned real `tokio` runtime crate to `1.49.0`, which is required by the newer AWS SDK pulled by `deltalake-aws 0.15.0`.
- Adapt small API changes in Delta Lake and GCloud connection options.

Note: `gcloud-bigquery 1.6.0` itself still depends on Arrow 56, but it does not expose Arrow types through the RisingWave BigQuery sink path we use. Therefore this PR does not vendor or patch `gcloud-bigquery`.

No issue.

## Checklist

- [ ] I have written necessary rustdoc comments.
- [ ]  I have added necessary unit tests and integration tests.
- [ ]  I have added test labels as necessary. 
- [ ]  I have added fuzzing tests or opened an issue to track them. 
- [ ]  My PR contains breaking changes. 
- [ ]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results. 
- [ ]  I have checked the [Release Timeline](https://github.com/risingwavelabs/rw-commits-history/blob/main/release_timeline.md) and [Currently Supported Versions](https://docs.risingwave.com/changelog/release-support-policy#support-end-dates-for-recent-releases) to determine which release branches I need to cherry-pick this PR into. 

## Documentation

- [ ]  My PR needs documentation updates. 

<details>
<summary><b>Release note</b></summary>

Not user-facing.

</details>

## Verification

- `cargo fmt`
- `cargo check -p gcloud-bigquery --locked` (before removing the local patch experiment)
- `cargo check -p risingwave_connector --features sink-bigquery`
- `cargo check -p risingwave_connector --features sink-deltalake,sink-bigquery,source-adbc_snowflake --locked`
[...]

PR #25557, opened 2026-05-04
feat(session): allow unitless `statement_timeout` to mean `ms`

### LINKED ISSUE #25556: PG-compat: `SET statement_timeout = <bare integer>` rejected, breaking standard Postgres clients (sqlx, JDBC, libpq, ...)
### Summary

RisingWave's PG-protocol layer rejects `SET statement_timeout = <integer>` (without explicit time unit), which is a 20-year-old canonical Postgres behavior. This breaks **session establishment** for virtually every standard Postgres client driver (`libpq`, JDBC, `pg-node`, `psycopg2/3`, `sqlx`, etc.), because most of them issue this exact statement immediately after connecting.

The result: any ETL/CDC tool that uses a stock Postgres client driver as a sink (Goldsky, Airbyte, Fivetran, dbt-core's PG adapter, Debezium PG sink, Sequin, Estuary, …) cannot even open a session against RisingWave Cloud — let alone run any INSERT.

### Reproduction

Tested against RisingWave Cloud `2.8.1`:

```bash
psql "host=<your-rw-host> port=4566 user=... dbname=dev sslmode=require" \
  -c "SET statement_timeout = 60000;"
```

Result:

```

  1: Failed to get/set session config
  2: Invalid value `60000` for `statement_timeout`
  3: time unit is required: ms s min h d
```

The same statement *with* an explicit unit works:

```sql
SET statement_timeout = '60s';   -- OK
SET statement_timeout = '60000ms'; -- OK
```

### Expected behavior

Match Postgres's documented behavior: a bare integer for `statement_timeout` should be interpreted as **milliseconds**.

From the [Postgres docs](https://www.postgresql.org/docs/current/runtime-config-client.html#GUC-STATEMENT-TIMEOUT):

> A value of zero (the default) disables the timeout. … If this value is specified without units, it is taken as milliseconds.

### Why this matters (real-world impact)

We hit this trying to set up a **direct Goldsky → RisingWave Cloud** integration via Goldsky's standard Postgres sink (no Kafka, no intermediate Postgres — relying on RisingWave's PG-wire compatibility). Goldsky's sink uses Rust's `sqlx` driver, which calls `SET statement_timeout = 60000` on every new connection. The session is killed before any query can execute.

This is not a Goldsky-specific problem. The same `SET` statement is sent by:

- **`libpq` / `psql`** when `PGOPTIONS='-c statement_timeout=60000'` is used (very common in ops/CI scripts)
- **JDBC** (`org.postgresql:postgresql`) — sets `statement_timeout` from the `socketTimeout`/`loginTimeout` connection params
- **`node-postgres` / `pg-node`** — when `statement_timeout` is passed in the pool config
- **`psycopg2` / `psycopg3`** — when set as a connection option
- **`sqlx` (Rust)** — used by Goldsky and many others; sends as part of `after_connect` hook
- **Go `pgx`** in some configurations

For RisingWave's positioning as a Postgres-protocol-compatible streaming database, this is a **single-line parser change with very high leverage**: it instantly unblocks dozens of ETL tools that "just want to write to a Postgres."

### Suggested fix

In the session-config validation path, when the setting is `statement_timeout` (and similar PG `*_timeout` settings: `lock_timeout`, `idle_in_transaction_session_timeout`), if the value is a bare integer literal, default the unit to milliseconds before parsing.

This is non-breaking — values with explicit units continue to work exactly as before.

### Environment

- RisingWave Cloud, version `2.8.1` (commit `f49ef55bb80c1628c48d879100f3c93c7e9a998f`)
- Client: `psql 18.3` (Homebrew); also reproduced via Rust `sqlx` (Goldsky's sink)

### Related observation (separate issue worth filing later, not blocking this one)

The downstream blocker once `statement_timeout` is fixed is `INSERT ... ON CONFLICT (...) DO UPDATE SET ...` — Postgres's standard UPSERT syntax (PG 9.5+, 2016) — which RisingWave's parser also rejects:

```
sql parser error: expected end of statement, found: ON
LINE 1: INSERT INTO t VALUES ('a', 1) ON CONFLICT (id) DO UPDATE SET val = EXCLUDED.val;
                                      ^
```

For the Goldsky-specific case this is workaroundable on the client side (Goldsky's Turbo sink can be configured without `primary_key`, which their docs say emits plain INSERTs). But broader PG ETL ecosystem compatibility would benefit from accepting `ON CONFLICT` at the parser level — at minimum as a no-op when no primary key is defined, ideally with full UPSERT semantics on tables with PK.

Happy to file as a separate issue if useful.

xiangjinwu:

Follow-up of #24633, making v3.0 compatible with PostgreSQL in this behavior.

This shall NOT be cherry-picked to v2.7 or v2.8 as it would be BREAKING CHANGE from v2.7

Resolves #25556

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -26,7 +26,7 @@ dependencies = [
  "path-slash",
  "regex",
  "toml 1.1.2+spec-1.1.0",
- "windows-registry 0.6.1",
+ "windows-registry",
  "windows-sys 0.61.2",
 ]
 
--- a/e2e_test/batch/config.slt
+++ b/e2e_test/batch/config.slt
@@ -18,8 +18,14 @@ set synchronize_seqscans to f;
 statement ok
 set synchronize_seqscans to default;
 
-statement error time unit is required
+# This means 2ms since v3.0, aligned with PostgreSQL
+# In v2.7 and prior versions, it was 2s
+statement ok
 set statement_timeout = 2;
+
+skipif madsim
+statement error timeout
+select pg_sleep(1);
 
 statement ok
 set statement_timeout = '2s';
--- a/src/common/src/session_config/statement_timeout.rs
+++ b/src/common/src/session_config/statement_timeout.rs
@@ -27,7 +27,7 @@ impl std::str::FromStr for StatementTimeout {
         let (val_str, unit) = s
             .find(|c: char| !c.is_numeric())
             .map(|i| s.split_at(i))
-            .ok_or_else(|| "time unit is required: ms s min h d".to_owned())?;
+            .unwrap_or((s, "ms"));
 
         let val = val_str
             .parse::<u32>()
@@ -95,9 +95,9 @@ mod tests {
             StatementTimeout(100)
         );
 
-        assert!(
-            "100".parse::<StatementTimeout>().is_err(),
-            "should fail without unit"
+        assert_eq!(
+            "100".parse::<StatementTimeout>().unwrap(),
+            StatementTimeout(100)
         );
         assert!(
             "100x".parse::<StatementTimeout>().is_err(),


Cargo.lock
```diff
@@ -26,7 +26,7 @@ dependencies = [
  "path-slash",
  "regex",
  "toml 1.1.2+spec-1.1.0",
- "windows-registry 0.6.1",
+ "windows-registry",
```
xiangjinwu, line 29: Introduced accidentally by #25479

hzxa21 (approved):

Issue #25561, opened 2026-05-04
bug(meta): ALTER FRAGMENT SET RATE_LIMIT writes wrong rate_limit fields and skips Source/StreamFsFetch


yuhao-su:
## Describe the bug

`ALTER FRAGMENT <id> SET RATE_LIMIT TO <v>` (the SQL is hard-wired to `ThrottleType::Backfill` in
the frontend at `src/frontend/src/handler/mod.rs:1538`) calls
`CatalogController::update_fragment_rate_limit_by_fragment_id` on the meta side, which has two
inconsistencies vs. the per-target helpers (`update_source_rate_limit_by_source_id`,
`update_backfill_rate_limit_by_job_id`, `update_sink_rate_limit_by_job_id`,
`update_dml_rate_limit_by_job_id`):

1. **DB writes are not gated by `throttle_type`.** The function unconditionally rewrites
   `rate_limit` on `Dml`, `Sink`, `StreamCdcScan`, `StreamScan`, and `SourceBackfill` nodes
   regardless of which throttle type the caller meant
   (`src/meta/src/controller/streaming_job.rs:3171-3211`). The barrier mutation that follows only
   carries `ThrottleType::Backfill`, so running `Sink` / `Dml` actors in the targeted fragment
   never react — but their `rate_limit` field on disk has been changed silently. After the next
   reschedule / replace, fresh actors materialize from inflight nodes (after the recovery that
   reloads inflight from DB) and start with the new value the user did not intend to apply.

2. **`Source` / `StreamFsFetch` are not handled at all.** Running
   `ALTER FRAGMENT <source_fragment_id> SET RATE_LIMIT TO 0` silently no-ops on the DB side, while
   the runtime mutation also targets only `Backfill` actors. A user who reaches for the
   fragment-level switch hoping to pause a single source actor will see no effect.

## How to reproduce

```sql
-- (1) wrong-target write: the DML node's rate_limit on disk gets overwritten
CREATE TABLE t (k INT);
-- pick the DML fragment id from rw_fragments / rw_rate_limit
ALTER FRAGMENT <dml_fragment_id> SET RATE_LIMIT TO 0;
-- DML stream still flows (the actor never received ThrottleType::Dml),
-- but rw_rate_limit shows the value has been written to 0.
-- After a meta restart or any reschedule of this fragment, a fresh DML actor
-- would come up with rate_limit=0.

-- (2) source no-op:
CREATE SOURCE s ... ;
[...]

Issue #20554, opened 2025-02-20
bug(catalog): rate limit will be muted when the table was overwritten.


KeXiangWang:
### Describe the bug

As titled.

### Error message/log

```text

```

### To Reproduce

I use a postgres source to reproduce this. `>>>` means it should be executed in postgres.

```
# create table in pg
>>> create table src (v int primary key);
>>> insert into src values(10) (11) (12);

# connect to the pg table with pg-cdc
CREATE table src (v int primary key) WITH (
    connector = 'postgres-cdc',
    hostname = '127.0.0.1',
    port = '5432',
    username = 'postgresuser',
    password = 'postgrespw',
    slot.name = 'my_slot',
    database.name = 'mydb',
    table.name = 'src',
    publication.name = 'rw_publication'
);

# stop the stream using rate limit
ALTER SOURCE src SET source_rate_limit TO 0;

# these data should not be in rw, as the streams has been stopped.
>>> insert into src values(20) (21) (22);

# return 3
select count( *) from mv;

# create a table and sink into the table with connector.
create table newdata(v int);
insert into newdata values(50);
create sink sk into src as select * from newdata;

# should return 4, but return 7
select count( *) from mv;
```

KeXiangWang:
cc @kwannoel Any ideas?

PR #25562, opened 2026-05-04
fix(meta): sync inflight stream node rate_limit on Throttle command

yuhao-su:

`Command::Throttle` updates `Source.rate_limit` on the catalog row and rewrites the matching
`rate_limit` field on `Fragment.stream_node`, then notifies running actors via
`Mutation::Throttle`. It does **not**, however, refresh
`InflightDatabaseInfo.fragment(..).nodes` — the meta barrier scheduler's in-memory snapshot of
each fragment's `StreamNode`. Every other DB-mutating command in `apply_command`
(`CreateStreamingJob`, `ReplaceStreamJob`, `RescheduleIntent`, `SourceChangeSplit`, …) has a
matching `pre_apply_*` that keeps inflight in sync; only `Throttle` skipped this and went
straight to `apply_simple_command`.

The fallout: `RescheduleIntent`'s `reschedule_actors_to_create` materializes new actors from
`database_info.fragment(fragment_id).nodes.clone()` (`command.rs:1531`). When ALTER had
flipped DB / runtime to a new rate_limit but inflight was stale, any later layout change
would spawn fresh actors that revert to the **previous** value persisted at job creation /
last recovery. The most painful symptom is "ALTER ... source_rate_limit TO 0; ALTER ...
TO DEFAULT (works); some unrelated reschedule (e.g. `serverless_backfill` provisioning a new
worker, ALTER PARALLELISM, scaling) → source pauses again, with `rate_limit_rps = Some(0)`."

This is also exactly the mechanism behind the family of bugs reported in #20554 and the
sibling `update_fragment_rate_limit_by_fragment_id` issue tracked in #25561.

### Fix

Add `InflightDatabaseInfo::pre_apply_throttle` mirroring the per-target DB writers
(`update_{source,backfill,sink,dml}_rate_limit_by_*`). It dispatches on `ThrottleType` and
rewrites the matching `rate_limit` field:

| `ThrottleType` | nodes updated |
| --- | --- |
| `Source`     | `Source.source_inner.rate_limit`, `StreamFsFetch.node_inner.rate_limit` |
| `Backfill`   | `StreamCdcScan.rate_limit`, `StreamScan.rate_limit`, `SourceBackfill.rate_limit` |
| `Sink`       | `Sink.rate_limit` |
| `Dml`        | `Dml.rate_limit` |
| `Unspecified`| no-op |

Snapshot-backfill creating jobs maintain their own fragment view in
`independent_checkpoint_job_controls`, so `pre_apply_throttle` skips fragments that are not
in the main inflight rather than panic through `fragment_mut`.

`apply_command`'s `Throttle` arm now calls `pre_apply_throttle` for each
`(fragment_id, ThrottleConfig)` before producing the mutation.

### Reproducer

`e2e_test/source_inline/kafka/alter/source_rate_limit_inflight_sync.slt.serial`:

1. `CREATE SOURCE` with `source_rate_limit = 0` and `STREAMING_PARALLELISM = 1`.
2. `ALTER SOURCE ... source_rate_limit TO DEFAULT`.
3. `ALTER SOURCE SET PARALLELISM = 4` to force a reschedule that materializes 3+ fresh
   actors.
4. Insert 100 more rows and assert the counting MV reaches 200.

Locally verified:
- **With this fix**: passes (~18s, a few backoff retries waiting on the kafka pipeline).
- **Without this fix**: stalls at `100/200`, `query ... retry 60` exhausts and the test fails.

- Closes nothing (no tracking issue was filed for this); cross-references #25561 (sibling).

## Checklist

## Documentation

<details>

Fixes a bug where `ALTER SOURCE ... source_rate_limit` (and the analogous backfill / sink /
dml rate-limit alters) would silently revert after any subsequent reschedule of the affected
fragment. Most visibly, a paused source that the user un-paused via
`ALTER ... source_rate_limit TO DEFAULT` could re-pause itself when a downstream MV / sink
change provoked a reschedule (for example, when `serverless_backfill` provisioning brought up
a new compute worker and triggered auto-parallelism rescheduling).

</details>

### DIFF
--- a/e2e_test/source_inline/kafka/alter/source_rate_limit_inflight_sync.slt.serial
+++ b/e2e_test/source_inline/kafka/alter/source_rate_limit_inflight_sync.slt.serial
@@ -1,0 +1,137 @@
+# Regression: `ALTER SOURCE source_rate_limit` updates DB and notifies running actors via
+# `Mutation::Throttle`, but historically did not refresh `database_info.fragment(..).nodes`.
+# A subsequent reschedule (e.g. `ALTER SOURCE SET PARALLELISM`) materialized fresh source
+# actors from stale inflight nodes, so they started with the *original* `rate_limit = Some(0)`
+# and stayed paused -- their splits never produced data.
+#
+# Repro:
+#   1. CREATE SOURCE with `source_rate_limit = 0` (parallelism=1).
+#      All actors + inflight nodes start at rate_limit = Some(0).
+#   2. ALTER SOURCE ... source_rate_limit TO DEFAULT.
+#      DB → null, the running actor resumes via Throttle mutation.
+#      Without the fix, inflight `info.nodes.source_inner.rate_limit` is still Some(0).
+#   3. ALTER SOURCE SET PARALLELISM = 4.
+#      Reschedule materializes 3 fresh actors from inflight. Without the fix, those actors
+#      start with `rate_limit_rps = Some(0)` and the 3 partitions assigned to them never
+#      drain.
+#   4. Produce 100 more rows, distributed round-robin across 4 partitions. Verify the MV
+#      sees all 200. With the bug, only the 1 partition on the kept actor advances and the
+#      MV stalls below 200.
+
+control substitution on
+
+system ok
+rpk topic delete test_rl_inflight_sync || true
+
+system ok
+rpk topic create test_rl_inflight_sync --partitions 4 --replicas 1
+
+############## Seed 100 rows into kafka
+
+statement ok
+create table seed (v1 int);
+
+statement ok
+insert into seed select * from generate_series(1, 100);
+
+statement ok
+create sink seed_sink from seed with (
+    ${RISEDEV_KAFKA_WITH_OPTIONS_COMMON},
+    topic = 'test_rl_inflight_sync',
+    type = 'append-only',
+    force_append_only = 'true'
+);
+
+sleep 5s
+
+############## Step 1: CREATE SOURCE with rate_limit = 0, parallelism = 1.
+# Inflight `info.nodes.source_inner.rate_limit` is set to Some(0) here.
+
+statement ok
+SET STREAMING_PARALLELISM = 1;
+
+statement ok
+create source rl_src (v1 int) with (
+  ${RISEDEV_KAFKA_WITH_OPTIONS_COMMON},
+  topic = 'test_rl_inflight_sync',
+  source_rate_limit = 0
+) FORMAT PLAIN ENCODE JSON;
+
+statement ok
+SET STREAMING_PARALLELISM = DEFAULT;
+
+statement ok
+flush;
+
+query T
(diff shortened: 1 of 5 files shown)


yuhao-su:
* **#25562** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25562?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25562?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

e2e_test/source_inline/kafka/alter/source_rate_limit_inflight_sync.slt.serial
```diff
@@ -0,0 +1,137 @@
+# Regression: `ALTER SOURCE source_rate_limit` updates DB and notifies running actors via
```
wenym1, line 1: Can write a simpler e2e test without depending on kafka.

First we can prepare a table with 1000 rows. And then create a MV from the table with initial parallelism as 1 and backfill rate limit as 0. And then we alter the rate limit to 10, and then we trigger the scale by alter the MV parallelism to 16, and then we do query with retry to check that the MV row count will be 1000 eventually. Before the fix, after the scale, the row count should stop increasing and will never reach 1000.

wenym1 (approved):
Rest LGTM

chenzl25 (approved):
LGTM, thanks for the fix.

e2e_test/source_inline/kafka/alter/source_rate_limit_inflight_sync.slt.serial
yuhao-su: Is isn't is basicly the same as kafka test? This test is also run before and after the fix and prove the fix is real.

Do you want me to keep the kafka test after adding the table one?

tabVersion (approved):
LGTM

e2e_test/ddl/alter_rate_limit_inflight_sync.slt
```diff
@@ -0,0 +1,67 @@
+# Regression for the Throttle / inflight-nodes desync fixed alongside this test:
+# ALTER ... rate_limit only updated DB + actor mutation, not `database_info.fragment.nodes`,
+# so a later ALTER PARALLELISM would materialize fresh actors from the stale snapshot and
+# resume backfill at rate_limit = 0. Mirrors the source-side test at
+# `e2e_test/source_inline/kafka/alter/source_rate_limit_inflight_sync.slt.serial`.
+
+# SnapshotBackfill is non-reschedulable while creating, so ALTER PARALLELISM would be rejected.
+statement ok
+set streaming_use_snapshot_backfill = false;
```
yuhao-su: @wenym1 PTAL

wenym1 (approved):
LGTM

PR #25571, opened 2026-05-06

tabVersion:

This PR hardens user-visible `SHOW` metadata surfaces so a user with schema visibility alone cannot enumerate object names or secret references that they do not have object-level privilege for.

Changes:

- Switch clear `SHOW` object-list handlers to ACL-aware schema iterators:
  - `SHOW TABLES`
  - `SHOW INTERNAL TABLES`
  - `SHOW VIEWS`
  - `SHOW MATERIALIZED VIEWS`
  - `SHOW SOURCES`
  - `SHOW SINKS`
  - `SHOW SUBSCRIPTIONS`
  - `SHOW SECRETS`
  - `SHOW CONNECTIONS`
  - `SHOW FUNCTIONS`
- Make `SHOW COLUMNS FROM <table/source>` and `SHOW INDEXES FROM <table>` use the normal batch binder path instead of the system binder path, avoiding known-name metadata probing for hidden relations.
- Add explicit object-visibility checks for `SHOW COLUMNS FROM <sink/view>` while preserving system-view column introspection.
- Redact referenced secret names from visible connection output when the current user can see the connection but cannot see the referenced secret. In that case the output renders `SECRET <redacted>` instead of `SECRET schema.secret_name`.
- Apply the same connection secret-ref redaction to `rw_catalog.rw_connections`.
- Add targeted SQLLogicTest coverage for hidden object-list results, known-name probing, and connection secret-ref redaction.

- During audit, `rw_catalog.rw_sources` was identified as a separate system-catalog surface that can still need visibility/secret-reference hardening. Iceberg-related catalog views should also be triaged in a follow-up pass.

## Checklist

## Documentation

<details>

RisingWave now avoids exposing unauthorized object names through `SHOW` object-list commands and known-name `SHOW COLUMNS` / `SHOW INDEXES` probes. Visible connection metadata also redacts referenced secret names when the user cannot independently see the secret.

</details>

## Local verification

Passed:

- `git diff --check -- src/frontend/src/handler/show.rs src/frontend/src/handler/create_connection.rs src/frontend/src/catalog/system_catalog/rw_catalog/rw_connections.rs e2e_test/ddl/show_object_privilege.slt`

Attempted but blocked locally:

- `./risedev d` failed because local Java is 17 while the Maven build requires target release 21 (`Fatal error compiling: error: invalid target release: 21`).
- `./risedev slt './e2e_test/ddl/show_object_privilege.slt'` then failed because no RisingWave server was running (`Connection refused (os error 61)`).

### DIFF
--- a/e2e_test/ddl/show_object_privilege.slt
+++ b/e2e_test/ddl/show_object_privilege.slt
@@ -1,0 +1,85 @@
+# Verify SHOW object-list commands don't expose objects that are only schema-visible.
+
+statement ok
+CREATE USER show_acl_user WITH PASSWORD 'password';
+
+statement ok
+CREATE SCHEMA show_acl;
+
+statement ok
+GRANT USAGE ON SCHEMA show_acl TO show_acl_user;
+
+statement ok
+CREATE TABLE show_acl.hidden_table (v1 int, v2 varchar);
+
+statement ok
+CREATE INDEX hidden_idx ON show_acl.hidden_table (v1);
+
+statement ok
+CREATE MATERIALIZED VIEW show_acl.hidden_mv AS SELECT v1 FROM show_acl.hidden_table;
+
+statement ok
+CREATE VIEW show_acl.hidden_view AS SELECT v1 FROM show_acl.hidden_table;
+
+statement ok
+CREATE SECRET show_acl.hidden_secret WITH (backend = 'meta') AS 'http://example.invalid';
+
+# Schema USAGE alone should not reveal object names.
+system ok
+bash -c 'set -euo pipefail; out=$(PGPASSWORD=password psql -U show_acl_user -d dev -Atc "SHOW SECRETS FROM show_acl;"); ! grep -q "hidden_secret" <<<"$out"'
+
+system ok
+bash -c 'set -euo pipefail; out=$(PGPASSWORD=password psql -U show_acl_user -d dev -Atc "SHOW SECRETS FROM show_acl LIKE '\''hidden_secret'\'';"); ! grep -q "hidden_secret" <<<"$out"'
+
+system ok
+bash -c 'set -euo pipefail; out=$(PGPASSWORD=password psql -U show_acl_user -d dev -Atc "SHOW TABLES FROM show_acl;"); ! grep -q "hidden_table" <<<"$out"'
+
+system ok
+bash -c 'set -euo pipefail; out=$(PGPASSWORD=password psql -U show_acl_user -d dev -Atc "SHOW VIEWS FROM show_acl;"); ! grep -q "hidden_view" <<<"$out"'
+
+system ok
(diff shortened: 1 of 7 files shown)


src/frontend/src/handler/show.rs
```diff
@@ -680,7 +696,12 @@ pub async fn handle_show_object(
                         }
                         connection::Info::ConnectionParams(params) => {
                             // todo: show dep relations
-                            print_connection_params(&session.database(), params, &reader)
+                            print_connection_params_with_secret_visibility(
+                                &session.database(),
```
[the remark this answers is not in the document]
tabVersion: Good catch, thanks! I filtered the PrivateLink `sources` / `sinks` lists with `has_access_to_object` before rendering dependent object names, so a user that can see only the connection won't see hidden dependent source/sink names. I also switched the new SLT connection fixture from Kafka to schema registry so the ddl suite does not require a Kafka broker.

e2e_test/ddl/show_object_privilege.slt
```diff
@@ -0,0 +1,107 @@
...
+CREATE MATERIALIZED VIEW show_acl.hidden_mv AS SELECT v1 FROM show_acl.hidden_table;
+
+statement ok
+CREATE VIEW show_acl.hidden_view AS SELECT v1 FROM show_acl.hidden_table;
+
+statement ok
+CREATE SECRET show_acl.hidden_secret WITH (backend = 'meta') AS 'http://example.invalid';
+
+statement ok
+CREATE CONNECTION show_acl.visible_conn WITH (
```
tabVersion: This fixture still depends on live connection validation. In Buildkite #95690, this `CREATE CONNECTION ... type = 'schema_registry'` reaches `/config` on `http://example.invalid/` and fails with DNS errors in normal e2e; in deterministic simulation it also panics because `get_secret_value` is unimplemented under madsim when the connection validator fills the secret. Since this test only needs catalog metadata for SHOW visibility/redaction, it should avoid a connection type/path that performs external validation or madsim secret resolution.

tabVersion (commented):
Found one CI-blocking issue in the new SLT fixture.

e2e_test/ddl/show_object_privilege.slt
tabVersion: Fixed in `0901f89204975fe4fc6773c63c76287920d9978e`: the DDL metadata-only SLT no longer creates the schema-registry connection. I moved the connection visibility/redaction coverage into `e2e_test/source_inline/connection/schema_registry.slt`, using `${RISEDEV_SCHEMA_REGISTRY_URL}`, and added source/sink SHOW visibility coverage in the Kafka source and sink e2e suites.

tabVersion:
Closing this PR in favor of #25584. GitHub's synthetic `refs/pull/25571/head` stayed stale at `5bd48e0bbc921946b2c19c2eb1b8bb1b803e9972` while the branch/PR API head advanced, so Buildkite could not fetch the commit it was asked to checkout. The replacement PR uses the same final code/tree on a fresh branch.

PR #25592, opened 2026-05-07
fix(meta): cleanup dirty recovery hummock tables

Li0k:

- Related to #25466

This PR is scoped to **per-database recovery dirty-job cleanup**. It does not change DDL abort handling.

Bug:

- Global recovery can clean dirty creating-job catalogs and then run a full-catalog Hummock purge.
- Per-database recovery can also clean dirty creating-job catalogs through `clean_dirty_creating_jobs(Some(database_id))`, but it does not run the global Hummock purge.
- Therefore, if per-database recovery removes dirty creating-job catalog rows, the corresponding Hummock table ids can remain as stale member/SST table ids.

Fix:

- Make `clean_dirty_creating_jobs` return the cleaned streaming job ids, table ids, and source ids instead of only source ids.
- In per-database recovery only, cache the table catalogs before deleting catalog rows, then reuse `cleanup_dropped_streaming_jobs` to unregister Hummock table ids and notify Hummock/compactor.
- Leave global recovery unchanged; it still relies on the existing full-catalog Hummock purge.

Not covered by this PR:

- DDL abort after create failure or `DROP/CANCEL` of `Initial` jobs. That path is fixed separately in #25604.
- The SQL/simulation repro that forces a foreground create failure belongs to #25604, because the stale ids are cleaned through the DDL abort path after the create fails.

### Testing

- `git diff --check`
- `cargo test -p risingwave_meta test_clean_dirty_creating_jobs_records_dropped_tables_for_per_db_recovery --lib -- --nocapture`

## Checklist

## Documentation

<details>

Fixed stale Hummock table ids left by dirty creating jobs during per-database recovery.

</details>

### DIFF
--- a/src/meta/src/controller/catalog/mod.rs
+++ b/src/meta/src/controller/catalog/mod.rs
@@ -144,6 +144,14 @@ pub struct ReleaseContext {
     pub(crate) removed_iceberg_table_sinks: Vec<PbSink>,
 }
 
+#[derive(Default)]
+pub(crate) struct CleanedDirtyStreamingJobs {
+    pub(crate) streaming_job_ids: Vec<JobId>,
+    /// Only populated for per-database recovery.
+    pub(crate) dropped_table_ids: Vec<TableId>,
+    pub(crate) source_ids: Vec<SourceId>,
+}
+
 impl CatalogController {
     pub async fn new(env: MetaSrvEnv) -> MetaResult<Self> {
         let meta_store = env.meta_store();
@@ -447,11 +455,11 @@ impl CatalogController {
     }
 
     /// `clean_dirty_creating_jobs` cleans up creating jobs that are creating in Foreground mode or in Initial status.
-    pub async fn clean_dirty_creating_jobs(
+    pub(crate) async fn clean_dirty_creating_jobs(
         &self,
         database_id: Option<DatabaseId>,
-    ) -> MetaResult<Vec<SourceId>> {
-        let inner = self.inner.write().await;
+    ) -> MetaResult<CleanedDirtyStreamingJobs> {
+        let mut inner = self.inner.write().await;
         let txn = inner.db.begin().await?;
 
         let filter_condition = streaming_job::Column::JobStatus.eq(JobStatus::Initial).or(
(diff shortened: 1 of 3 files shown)


src/meta/src/controller/catalog/mod.rs
```diff
@@ -490,12 +497,16 @@ impl CatalogController {
...
 
         if dirty_job_objs.is_empty() {
-            return Ok(vec![]);
+            return Ok(CleanedDirtyStreamingJobs::default());
         }
 
         self.log_cleaned_dirty_jobs(&dirty_job_objs, &txn).await?;
 
         let dirty_job_ids = dirty_job_objs.iter().map(|obj| obj.oid).collect::<Vec<_>>();
+        let dirty_streaming_job_ids = dirty_job_objs
```
wenym1: We can actually use `dirty_job_ids` directly to store `JobId`. The `PartialObject` has a `job_id` field, and we don't need to use the `oid`

src/meta/src/controller/catalog/mod.rs
```diff
@@ -591,13 +603,38 @@ impl CatalogController {
...
             .collect();
 
+        // Per-database recovery does not run the global Hummock purge. Keep table catalogs so the
+        // caller can reuse the normal dropped-table cleanup path after the catalog rows are deleted.
+        let dropped_tables = if database_id.is_some() {
+            Table::find()
+                .find_also_related(Object)
+                .filter(
+                    table::Column::TableId.is_in(
+                        to_delete_objs
```
wenym1: can be `is_in(dirty_state_table_ids)`

src/meta/src/controller/catalog/mod.rs
wenym1: can by the way change the `dirty_job_ids.clone().into_iter()` to `dirty_job_ids.iter().copied()`

src/meta/src/controller/catalog/mod.rs
```diff
@@ -144,6 +144,13 @@ pub struct ReleaseContext {
     pub(crate) removed_iceberg_table_sinks: Vec<PbSink>,
 }
 
+#[derive(Default)]
+pub(crate) struct CleanedDirtyStreamingJobs {
+    pub(crate) streaming_job_ids: Vec<JobId>,
+    pub(crate) dropped_table_ids: Vec<TableId>,
```
wenym1, line 151: Please add comment to `dropped_table_ids` that it's only valid when `database_id.is_some()`

wenym1 (approved):
rest LGTM

Issue #211, opened 2021-10-28
Remove single mode


fuyufjh:
"Single mode" was designed to generate and execute a plan without `Exchange`, which is only used in early development. 

Note that the so-called "single mode" is *NOT* a mode designed for single node deployment. `Exchange` is necessary for parallel execution no mater on single node or cluster.

Currently, there is more and more code handling "single mode". It's time to remove these stuff.

PR #25596, opened 2026-05-08
fix(sqlserver-cdc): tolerate ill-formed UTF-16 in nvarchar/ntext payloads

wcy-fdu:
## Summary

Bump `tiberius` to `risingwavelabs/tiberius@3133b49d`, which switches `nvarchar` / `ntext` column-data decoding from `String::from_utf16` (strict) to `from_utf16_lossy`.

## Why

SQL Server does not enforce well-formed UTF-16 in `nvarchar` storage. A row containing a **lone surrogate** (e.g. `NCHAR(0xD83D)` written without its low-surrogate partner — common via legacy ETL / MERGE replication / buggy clients) currently causes `tiberius` to reject the entire snapshot chunk with `FromUtf16Error`.

Failure mode in CDC backfill is bad: the bad chunk wipes the in-flight result stream, the consumer re-issues from the last persisted LSN / PK, and immediately re-hits the same chunk → **infinite retry loop with zero forward progress**. Observed in production stalled at ~210k rows (chunk #211 at the default `snapshot.batch_size = 1000` contained a lone surrogate).

After this bump, lone surrogates decode to `U+FFFD` and backfill survives. Protocol-level UTF-16 paths in `tiberius` (env-change tokens, login, identifiers) remain strict on purpose — those are server-controlled metadata where a malformed value indicates a real TDS protocol violation.

Upstream `tiberius` PR: https://github.com/risingwavelabs/tiberius/pull/2

## Test plan

- [x] `./risedev c` clean.
- [x] Local repro (`wcy-tools/mssql-cdc/repro_utf16.sh`): a 101-row table with one row containing `NCHAR(0xD83D)` finishes backfill at 101/101; the bad row lands as `'lone�'`.
- [x] Pre-fix on the same setup: backfill stalls with `FromUtf16Error` from `tiberius` in the connector log.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -15483,7 +15483,7 @@ dependencies = [
 [[package]]
 name = "tiberius"
 version = "0.12.2"
-source = "git+https://github.com/risingwavelabs/tiberius.git?rev=59350b92775a61611cf1fc3a5d0ff38bf264e6ac#59350b92775a61611cf1fc3a5d0ff38bf264e6ac"
+source = "git+https://github.com/risingwavelabs/tiberius.git?rev=3133b49dc22915266ae22bb83a32e10b4cdfb988#3133b49dc22915266ae22bb83a32e10b4cdfb988"
 dependencies = [
  "async-trait",
  "asynchronous-codec",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -15483,7 +15483,7 @@ dependencies = [
 [[package]]
 name = "tiberius"
 version = "0.12.2"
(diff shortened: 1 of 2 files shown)


wenym1 (approved):

xiangjinwu:
Any concern NOT merging this before the cherry-pick? Is it just slow CI?

wcy-fdu:
> Any concern NOT merging this before the cherry-pick? Is it just slow CI?

Merging

xiangjinwu:
We skipped it this time but it is better to have a unit test to prevent future version update to a problematic one.

PR #25599, opened 2026-05-08
fix(sqlparser): correctly parse delimiters inside dollar-quoted string

### LINKED ISSUE #14504: bug(sqlparser): dollar-quoted string cannot be nested
### Describe the bug

https://www.postgresql.org/docs/16/sql-syntax-lexical.html#SQL-SYNTAX-DOLLAR-QUOTING

With the usage of a tag, dollar-quoted string is supposed to be able to be nested without escaping. As shown in the PostgreSQL example:
```
$function$
BEGIN
    RETURN ($1 ~ $q$[\t\r\n\v\\]$q$);
END;
$function$
```

Or simpler:
```
select $foo$the content with $bar$nested$bar$ usage$foo$;
select $foo$not nesting just $ sign$foo$;
```

### Error message/log

```text
sql parser error: Unterminated dollar-quoted string at or near "foo" at Line: 1, Column 8
```

### To Reproduce

```
select $foo$the content with $bar$nested$bar$ usage$foo$;
select $foo$not nesting just $ sign$foo$;
select $$not nesting just $ sign$$; -- the untagged version is parsing correctly; just missing binder implementation
```

### Expected behavior

```
test=# select $foo$the content with $bar$nested$bar$ usage$foo$;
                ?column?                 
-----------------------------------------
 the content with $bar$nested$bar$ usage
(1 row)

test=# select $foo$not nesting just $ sign$foo$;
        ?column?         
-------------------------
 not nesting just $ sign
(1 row)

test=# select $$not nesting just $ sign$$;
        ?column?         
-------------------------
 not nesting just $ sign
(1 row)
```

_No response_

### The version of RisingWave

_No response_

### Additional context

https://github.com/risingwavelabs/risingwave/blob/b636aa4b80dcfea39396af412350cce13096de2b/src/sqlparser/src/tokenizer.rs#L874-L875

When a tentative closing tag does not match the opening tag, it should be treated as part of the content (possibly nested opening tag) rather than reporting an error.

The bug is present in upstream sqlparser as well.

Also note: The nesting is actually an illusion. PostgreSQL only cares about the outermost level:
```
test=# select $foo$the content with $bar$nested$tri$ usage$foo$; -- unmatched inner tag without error
                ?column?                 
-----------------------------------------
 the content with $bar$nested$tri$ usage
(1 row)
```

### LINKED ISSUE #14763: feat: support Dollar-Quoted String Constants

PostgreSQL Dollar-Quoted String Constants  https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-DOLLAR-QUOTING

DBT persist doc feature relies on this feature as well：
https://github.com/risingwavelabs/dbt-risingwave/issues/31#issue-2092520204

_No response_

_No response_

### Additional context

_No response_

Standing-Man:

- Closes #14504
- Closes #14763.

**problem**: The current sqlparser behavior incorrectly terminates or errors on nested dollar-quote-like content inside a dollar-quoted string, instead of handling it as literal content.

The implementation now treats all characters inside both untagged and tagged dollar-quoted strings as literal string content.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/e2e_test/batch/basic/dollar_quoted_string.slt.part
+++ b/e2e_test/batch/basic/dollar_quoted_string.slt.part
@@ -1,0 +1,19 @@
+query T
+select $$hello$$;
+----
+hello
+
+query T
+select $foo$the content with $bar$nested$bar$ usage$foo$;
+----
+the content with $bar$nested$bar$ usage
+
+query T
+select $fo$o$not nesting just $ sign$fo$o$;
+----
+o$not nesting just $ sign
+
+query T
+select $$hel$$lo$$;
+----
+hel
--- a/e2e_test/batch/basic/dollar_quoted_string.slt.part
+++ b/e2e_test/batch/basic/dollar_quoted_string.slt.part
@@ -27,10 +27,33 @@ create table t(v int);
 statement ok
 COMMENT ON TABLE t IS $$hello$$;
 
+query TIIIT
+SELECT
+    d.description
+FROM
+    pg_catalog.pg_description d
+    JOIN pg_catalog.pg_class c
+        ON d.objoid = c.oid
+WHERE c.relname = 't' AND d.objsubid = 0
+ORDER BY d.objsubid;
+----
+hello
+
 query T
 SELECT DATE $$1999-01-01$$;
 ----
 1999-01-01
+
+statement ok
+create function dollar_quote_add(INT, INT) returns int language sql as $$select $1 + $2$$;
+
+query I
+select dollar_quote_add(20, 22);
+----
+42
+
+statement ok
+drop function dollar_quote_add(int, int);
 
 statement ok
 drop table t;
--- a/e2e_test/batch/basic/dollar_quoted_string.slt.part
+++ b/e2e_test/batch/basic/dollar_quoted_string.slt.part
@@ -17,3 +17,20 @@ query T
 select $$hel$$lo$$;
 ----
 hel
+
+statement ok
+drop table if exists t;
+
+statement ok
+create table t(v int);
+
+statement ok
+COMMENT ON TABLE t IS $$hello$$;
+
+query T
+SELECT DATE $$1999-01-01$$;
+----
+1999-01-01
+
+statement ok
+drop table t;
--- a/e2e_test/batch/basic/dollar_quoted_string.slt.part
+++ b/e2e_test/batch/basic/dollar_quoted_string.slt.part
@@ -27,7 +27,7 @@ create table t(v int);
 statement ok
 COMMENT ON TABLE t IS $$hello$$;
 
-query TIIIT
+query T
 SELECT
     d.description
 FROM
--- a/src/frontend/src/binder/expr/value.rs
+++ b/src/frontend/src/binder/expr/value.rs
@@ -28,6 +28,7 @@ impl Binder {
         match value {
             Value::Number(s) => self.bind_number(s.clone()),
             Value::SingleQuotedString(s) => self.bind_string(s),
+            Value::DollarQuotedString(s) => self.bind_string(&s.value),
             Value::CstyleEscapedString(s) => self.bind_string(&s.value),
             Value::Boolean(b) => self.bind_bool(*b),
             // Both null and string literal will be treated as `unknown` during type inference.
--- a/src/sqlparser/src/parser.rs
+++ b/src/sqlparser/src/parser.rs
@@ -4255,6 +4255,7 @@ impl Parser<'_> {
         let token = self.next_token();
         match token.token {
             Token::SingleQuotedString(s) => Ok(s),
+            Token::DollarQuotedString(s) => Ok(s.value),
             _ => self.expected_at(checkpoint, "literal string"),
         }
     }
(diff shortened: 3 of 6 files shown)


Standing-Man:
Hi @chenzl25 and @xiangjinwu, could you please take a look when you have a moment?

tabVersion:
Thanks for the fix — the exact-delimiter scan in the tokenizer looks like the right direction for #14504. It now treats non-matching inner `$...$` sequences as literal content, which matches PostgreSQL’s dollar-quoted string body semantics.

I think there are two compatibility/scope boundaries worth clarifying before we claim full “Dollar-Quoted String Constants” support / close #14763:

1. Parser string-literal contexts still appear to be single-quote-only.

   The new expression path works (`Token::DollarQuotedString` -> `Value::DollarQuotedString` -> `bind_string(&s.value)`), but parser paths that explicitly call `parse_literal_string()` still only accept `Token::SingleQuotedString`.

   A concrete user-facing example is `COMMENT ON ... IS <string>`, which is especially relevant to the dbt persist-docs use case mentioned in #14763. `parse_comment()` calls `parse_literal_string()`, so something like this still seems unsupported:

   ```sql
   COMMENT ON TABLE t IS $$hello$$;
   ```

   PostgreSQL treats dollar-quoted text as a string constant here. Could we extend `parse_literal_string()` (or add a shared helper) so `Token::DollarQuotedString(s)` returns `s.value` in string-literal contexts too? A parser test like `COMMENT ON TABLE t IS $$hello$$` would catch this gap.

2. Dollar-quote tag grammar is still a bit more permissive than PostgreSQL.

   PostgreSQL says the tag follows unquoted-identifier rules, except that it cannot contain `$`. That means a non-empty tag should not start with a digit. The current scanner reads the tag with a single continuation-style predicate (`is_alphanumeric() || '_'`), so an input like this would be accepted as a tagged dollar-quoted string:

   ```sql
   SELECT $1$abc$1$;
   ```

   That does not match PostgreSQL’s tag grammar. If this PR aims for PostgreSQL-compatible dollar quote support, it would be good to validate tag-start separately from tag-continuation and add a negative test for this case.

So the core tokenizer fix looks good to me for the nested/different-delimiter bug, but I think we should either cover these string-literal/tag-grammar boundaries here, or narrow the PR scope/issue closure to avoid overclaiming full dollar-quoted string constant support.

Standing-Man:
Thanks for the detailed review. Currently, this PR adds support for dollar-quoted strings in string-literal contexts and restricts tag parsing to align with PostgreSQL’s rules. I’ve also narrowed the issue closure scope to avoid overclaiming full support for all dollar-quoted string constant cases. cc @tabVersion

Standing-Man:
This PR covers dollar-quoted string constants in these paths:

- **Tokenizer**
  - Supports both `$$...$$` and `$tag$...$tag$`.
  - Correctly treats non-matching inner dollar delimiters, such as `$bar$`, as string content.
  - Allows plain `$` inside the string body.
  - Validates tag format as `[a-zA-Z_][a-zA-Z0-9_]*`.

- **Parser expression literals**
  - Parses dollar-quoted strings into `Value::DollarQuotedString`.
  - Covers normal expression usage like:
    ```sql
    SELECT $$hello$$;
    SELECT $foo$the content with $bar$nested$bar$ usage$foo$;
    ```

- **Binder**
  - Binds `Value::DollarQuotedString` as a normal untyped string literal, same as single-quoted strings.

- **`parse_literal_string` paths**
  - Extends `parse_literal_string` to accept `Token::DollarQuotedString`.
  - This covers syntax locations that require a literal string, including:
    ```sql
    COMMENT ON TABLE t IS $$hello$$;
    SELECT DATE $$1999-01-01$$;
    ```

- **SQL UDF function body**
  - Keeps supporting SQL UDF bodies written with dollar quotes:
    ```sql
    CREATE FUNCTION f(INT, INT)
    RETURNS INT
    LANGUAGE SQL
    AS $$select $1 + $2$$;
    ```

- **Tests**
  - Adds parser tests for nested-looking dollar delimiters.
  - Adds tests for dollar-quoted strings followed by column aliases.
  - Adds `parse_literal_string` coverage with typed literals.
  - Adds SLT coverage for `SELECT`, `COMMENT ON TABLE`, `SELECT DATE`, and SQL UDF `AS $$...$$`.

**Note:**
There are still several RisingWave-specific frontend handler paths where string-like arguments are manually matched against `SingleQuotedString` / `DoubleQuotedString` / `CstyleEscapedString`, for example:

- `WITH (...)` / connector options
- `CREATE SECRET`
- `SET TIME ZONE`
- `ALTER CONFIG`
- `ALTER SUBSCRIPTION SET RETENTION`
- resource group string parameters

Whether those paths should also accept dollar-quoted strings is left out of this PR and may need separate discussion, since they are not all PostgreSQL-standard string constant positions and may have command-specific compatibility or validation requirements.

Standing-Man:
Hi @tabVersion and @xiangjinwu, When you have some time, could you please take a look? I’d appreciate any feedback or suggestions for further changes. Thanks for your time!

chenzl25 (approved):
LGTM, thanks

PR #25601, opened 2026-05-08
fix(sqlparser): correctly parse escaped double-quoted identifiers

### LINKED ISSUE #14731: bug(frontend): parsing field name containing `"`
### Describe the bug

_No response_

### Error message/log

```text

Caused by:
  sql parser error: Expected end of statement, found: "NG'" at line:1, column:31
Near "SELECT 'STRI"NG' AS "'STRI""
```
```


### To Reproduce

```
SELECT 'STRI"NG' AS "'STRI""NG'";
```

### Expected behavior

```
 'STRI"NG'
-----------
 STRI"NG
(1 row)
```

_No response_

### The version of RisingWave

_No response_

### Additional context

_No response_

### PR DESCRIPTION
**Pull request by Standing-Man**

- Closes #14731.

Fixes parsing of escaped double quotes inside quoted identifiers.
Previously, Parser stopped parsing a quoted identifier at the first double quote, so PostgreSQL-compatible identifiers containing `""` could be split into multiple tokens.

For example, the following SQL failed to parse:

```sql
SELECT 'STRI"NG' AS "'STRI""NG'";
```

This PR:

- Adds dedicated parsing for delimited identifiers.
- Treats `""` inside a double-quoted identifier as one literal `"`.
- Stores the unescaped identifier value in Ident.
- Escapes embedded double quotes in Ident::Display to preserve SQL round trips.
- Adds tokenizer and parser regression tests for identifiers containing escaped double quotes.

### Verification
```
cargo test -p risingwave_sqlparser tokenize_escaped_double_quote_in_delimited_identifier
cargo test -p risingwave_sqlparser parse_escaped_double_quote_in_delimited_identifier
```

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/sqlparser/src/ast/mod.rs
+++ b/src/sqlparser/src/ast/mod.rs
@@ -21,8 +21,7 @@ mod statement;
 mod value;
 
 use std::collections::HashSet;
-use std::fmt;
-use std::fmt::Display;
+use std::fmt::{self, Display};
 use std::sync::Arc;
 
 use itertools::Itertools;
@@ -178,7 +177,7 @@ impl Ident {
         let needs_quotes = QuoteIdent::needs_quotes(value);
 
         if needs_quotes {
-            Self::with_quote_unchecked('"', value.replace('"', "\"\""))
+            Self::with_quote_unchecked('"', value)
         } else {
             Self::new_unchecked(value)
         }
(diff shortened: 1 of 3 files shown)


Standing-Man:
Hi @chenzl25 and @xiangjinwu, could you please take a look when you have a chance? Thanks!

chenzl25:
One small issue: `tokenize_delimited_identifier` now stores the unescaped identifier value (so `""` becomes `"`), but `Word::Display` still prints `self.value` directly. As a result, formatting a token parsed from `SELECT "a""b"` would produce `"a"b"` instead of the valid escaped form `"a""b"`. Could we escape embedded `"` in `Word::Display` for `Some("\"")` as `Ident::Display` does, and add a small tokenizer display/round-trip test?

chenzl25 (approved):

PR #25608, opened 2026-05-10
fix(optimizer): normalize arithmetic rank predicates for over-window …

### LINKED ISSUE #25607: Over-window TopN rewrite misses equivalent predicate cases
### Describe the bug

(Not a pure correctness bug, but a case where the optimizer can produce a worse plan for semantically equivalent SQL)

We have a rule `OverWindowToTopNRule` ([reference](https://github.com/risingwavelabs/risingwave/blob/d7cf5564836258608abb71603c607fb1ff1dbc76/src/frontend/src/optimizer/rule/over_window_to_topn_rule.rs)) which recognizes some arithmetic predicates on `row_number()` or `rank()` for TopN optimization. However, it currently misses equivalent predicates when the constant is on the left side of the comparison.

For example, `rn - 1 = 0` can be optimized to `StreamGroupTopN`, while `0 = rn - 1` remains as `StreamFilter` over `StreamOverWindow`.

It could be the case that the query is given to us this way, and the fix wouldn't be complex so I believe this is worth improving.

### Error message/log

```text
N/A
```

### To Reproduce

First:
```sql
create table t (k int, v int);
```

Then, run the query with the usual predicate arrangement to see the expected plan:
```sql
explain create materialized view mv as
  select * from (
    select *, row_number() over (partition by k order by v desc) as rn
    from t
  )
where rn - 1 = 0;
```

Which generates:
```
StreamMaterialize { columns: [k, v, rn], stream_key: [k], pk_columns: [k], pk_conflict: NoCheck }
└─StreamOverWindow { window_functions: [row_number() OVER(PARTITION BY t.k ORDER BY t.v DESC ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW)] }
 └─StreamProject { exprs: [t.k, t.v] }
   └─StreamGroupTopN { order: [t.v DESC], limit: 1, offset: 0, group_key: [t.k] }
     └─StreamExchange { dist: HashShard(t.k) }
       └─StreamTableScan { table: t, columns: [k, v, _row_id] }
(6 rows)
```

which is the expected plan where the rule was applied. 

Now run this query (same query except the predicate is swapped):

```sql
explain create materialized view mv as
  select * from (
    select *, row_number() over (partition by k order by v desc) as rn
    from t
  )
where 0 = rn - 1;
```

Which currently generates:
```
StreamMaterialize { columns: [k, v, t._row_id(hidden), rn], stream_key: [t._row_id, k], pk_columns: [t._row_id, k], pk_conflict: NoCheck }
└─StreamFilter { predicate: (0:Int32 = (row_number - 1:Int32)) }
 └─StreamOverWindow { window_functions: [row_number() OVER(PARTITION BY t.k ORDER BY t.v DESC ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW)] }
   └─StreamExchange { dist: HashShard(t.k) }
     └─StreamTableScan { table: t, columns: [k, v, _row_id] }
(5 rows)
```
This clearly shows the rule wasn't applied (eg; we still have the filter predicate showing `0 = row_number - 1` when this should have been normalized to `row_number = 1` (and then further rules are applied)

### Expected behavior

Running the query from above (with `0 = rn - 1` predicate) should give identical output to the other query:

```
StreamMaterialize { columns: [k, v, rn], stream_key: [k], pk_columns: [k], pk_conflict: NoCheck }
└─StreamOverWindow { window_functions: [row_number() OVER(PARTITION BY t.k ORDER BY t.v DESC ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW)] }
 └─StreamProject { exprs: [t.k, t.v] }
   └─StreamGroupTopN { order: [t.v DESC], limit: 1, offset: 0, group_key: [t.k] }
     └─StreamExchange { dist: HashShard(t.k) }
       └─StreamTableScan { table: t, columns: [k, v, _row_id] }
(6 rows)
```

Local dev with `./risedev d`

### The version of RisingWave

```
dev=> select version();
                       version
-----------------------------------------------------
 PostgreSQL 13.14.0-RisingWave-2.9.0-alpha (unknown)
(1 row)
```

### Additional context

_No response_

murphylee10:

Closes #25607 

To reiterate what's mentioned in the issue, this PR fixes a missed `OverWindowToTopNRule` application for arithmetic predicates on `row_number()` or `rank()`. Previously, the arithmetic simplifier handled forms like `rank - 1 = 0` by converting to `rank = 1`, but missed predicates where the constant appears on the left side (eg;  `0 = rank - 1`)

With this PR, when the arithmetic expression is on the right side, it swaps operands and reverses the comparison
operator where needed. It also recursively rewrites the simplified output so nested supported arithmetic, such as `(rank - 1) - 1 = 0`, can be simplified.

Also added planner tests for:
  - existing left-side arithmetic predicates
  - mirrored arithmetic predicates
  - mirrored arithmetic inequalities
  - commuted addition
  - nested arithmetic
  - mirrored nested arithmetic

## Checklist

### DIFF
--- a/src/frontend/planner_test/tests/testdata/input/over_window_function.yaml
+++ b/src/frontend/planner_test/tests/testdata/input/over_window_function.yaml
@@ -245,6 +245,62 @@
     - stream_plan
     - optimized_logical_plan_for_batch
     - batch_plan
+- id: TopN by row_number with arithmetic rank predicate
+  sql: |
+    create table t(x int, y int);
+    select x, y from
+      (select *, row_number() over(PARTITION BY y ORDER BY x) rank from t)
+    where rank - 1 = 0;
+  expected_outputs:
+    - stream_plan
+- id: TopN by row_number with mirrored arithmetic rank predicate
+  sql: |
+    create table t(x int, y int);
+    select x, y from
+      (select *, row_number() over(PARTITION BY y ORDER BY x) rank from t)
+    where 0 = rank - 1;
+  expected_outputs:
+    - stream_plan
+- id: TopN by row_number with mirrored arithmetic rank inequality
+  sql: |
+    create table t(x int, y int);
+    select x, y from
+      (select *, row_number() over(PARTITION BY y ORDER BY x) rank from t)
+    where 0 < rank - 1 AND rank <= 3;
+  expected_outputs:
+    - stream_plan
+- id: TopN by row_number with mirrored arithmetic rank upper bound
+  sql: |
+    create table t(x int, y int);
+    select x, y from
+      (select *, row_number() over(PARTITION BY y ORDER BY x) rank from t)
+    where 3 >= rank + 1;
+  expected_outputs:
+    - stream_plan
+- id: TopN by row_number with mirrored commuted addition rank predicate
+  sql: |
+    create table t(x int, y int);
+    select x, y from
+      (select *, row_number() over(PARTITION BY y ORDER BY x) rank from t)
+    where 2 = 1 + rank;
+  expected_outputs:
+    - stream_plan
+- id: TopN by row_number with nested arithmetic rank predicate
+  sql: |
+    create table t(x int, y int);
+    select x, y from
+      (select *, row_number() over(PARTITION BY y ORDER BY x) rank from t)
+    where (rank - 1) - 1 = 0;
+  expected_outputs:
+    - stream_plan
+- id: TopN by row_number with mirrored nested arithmetic rank predicate
+  sql: |
+    create table t(x int, y int);
+    select x, y from
+      (select *, row_number() over(PARTITION BY y ORDER BY x) rank from t)
+    where 0 = (rank - 1) - 1;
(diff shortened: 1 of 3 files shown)


chenzl25 (approved):
LGTM, thanks for the fix.

murphylee10:
Hey @chenzl25, thx for the approval. I was wondering if there's something I'm supposed to do to merge this in? I saw you enabled auto-merge, and last time I thought it just needed a rebase but not sure if that was the blocker. It's my first PR here so I'm not sure what the usual process is

chenzl25:
> Hey @chenzl25, thx for the approval. I was wondering if there's something I'm supposed to do to merge this in? I saw you enabled auto-merge, and last time I thought it just needed a rebase but not sure if that was the blocker. It's my first PR here so I'm not sure what the usual process is

Let me trigger the merge queue again. Right now for external contributors, it needs to trigger manually.

PR #25609, opened 2026-05-10
fix: use full column subset for distinct agg deduplication check

### LINKED ISSUE #20187: `count(distinct _row_id)` causing other `count` producing wrong result
### Describe the bug

_No response_

### Error message/log

_No response_

### To Reproduce

```sql
dev=> create table t (foo int, bar int);
CREATE_TABLE
dev=> insert into t values (1, 1), (2, 2), (1, 1);
INSERT 0 3
dev=> select count(foo), count(distinct foo) from t;
 count | count 
-------+-------
     3 |     2
(1 row)

dev=> select count(foo), count(distinct foo), count(distinct _row_id) from t;
 count | count | count 
-------+-------+-------
     3 |     1 |     3
             ^
           wrong
(1 row)

dev=> select count(foo), count(distinct foo), count(distinct bar) from t;
 count | count | count 
-------+-------+-------
     3 |     2 |     2
(1 row)
```

### Expected behavior

_No response_

_No response_

### The version of RisingWave

_No response_

### Additional context

_No response_

kimjune01:

This PR fixes a bug where `count(distinct _row_id)` appearing in the same query as `count(distinct foo)` caused the latter to produce incorrect results.

The root cause was that `DistinctAggRule::build_expand` decided whether `Expand` was needed using only the first input index of each distinct aggregate (`agg_call.input_indices()[0]`). However, `column_subsets` deduplicates using the full key `group_keys + agg_call.input_indices()`. When two distinct aggregates shared the same first input but required different full deduplication subsets, the rule could skip `Expand`, causing the final aggregation filters/flags to reuse the wrong deduplication state.

The fix aligns `n_different_distinct` with `column_subsets`: both now key on the complete `group_keys + input_indices` subset. This ensures `Expand` is created whenever distinct aggregates need separate deduplication states.

The test (`e2e_test/streaming/aggregate/count_distinct_row_id_bug.slt`) exercises the reported case and adds GROUP BY coverage to verify the fix works correctly with grouped queries.

- Closes #20187

## Checklist

## Documentation

<details>

Fixes incorrect results from `count(distinct col)` when `count(distinct _row_id)` or other distinct aggregates with overlapping deduplication subsets appear in the same query.

</details>

### DIFF
--- a/e2e_test/streaming/aggregate/count_distinct_row_id_bug.slt
+++ b/e2e_test/streaming/aggregate/count_distinct_row_id_bug.slt
@@ -27,5 +27,12 @@ select count(foo), count(distinct foo), count(distinct bar) from t;
 ----
 3 2 2
 
+-- Test with GROUP BY to exercise group_keys in the deduplication logic
+query IIII rowsort
+select bar, count(foo), count(distinct foo), count(distinct _row_id) from t group by bar;
+----
+1 2 1 2
+2 1 1 1
+
 statement ok
 drop table t;
--- a/e2e_test/streaming/aggregate/count_distinct_row_id_bug.slt
+++ b/e2e_test/streaming/aggregate/count_distinct_row_id_bug.slt
@@ -1,0 +1,31 @@
+-- Test for issue #20187: count(distinct _row_id) causing other count(distinct ...) producing wrong result
+
+statement ok
+SET RW_IMPLICIT_FLUSH TO true;
+
+statement ok
+create table t (foo int, bar int);
+
+statement ok
+insert into t values (1, 1), (2, 2), (1, 1);
+
+-- This should work correctly
+query II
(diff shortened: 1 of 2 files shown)


chenzl25:
Thanks for the fix, but I realize that to fix this bug, we need to fix more thing than DistinctAggRule. It is likely related to the functional  dependencies of Expand operators. I will take a look and if necessary open a new PR to give a quick fix

chenzl25:
@kimjune01 I have opened a new PR which includes the additional fix (functional dependencies of expand operator) to make it work https://github.com/risingwavelabs/risingwave/pull/25623

kimjune01:
Thanks @chenzl25 — glad the root cause analysis helped. Your PR #25623 with the functional dependency fix is the right approach. Closing in favor of that.

PR #25624, opened 2026-05-12
fix(cdc): guard against backward Postgres LSN updates in split state

wcy-fdu:
## Background

Three LSN fields in a Postgres CDC offset JSON:

| Field | Meaning | Used by dbz for |
|---|---|---|
| `lsn` | physical position of the current WAL message | per-event bookkeeping |
| `lsn_proc` (`lastCompletelyProcessedLsn`) | last fully processed message | `startStreaming` resume position; `WalPositionLocator` dedup key |
| `lsn_commit` (`lastCommitLsn`) | last COMMIT message processed | sent to PG via `flushLsn` to advance `confirmed_flush_lsn`; checked by `validateLogPosition` against `slot.restart_lsn` on restart |

## Problem

dbz overwrites these fields without a max-guard ([`PostgresOffsetContext.java:135-139`](https://github.com/debezium/debezium/blob/main/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresOffsetContext.java#L135-L139)):

```
public void updateCommitPosition(Lsn lsn, Lsn lastCompletelyProcessedLsn) {
    this.lastCompletelyProcessedLsn = lastCompletelyProcessedLsn;
    this.lastCommitLsn = lsn;       // unconditional overwrite
}
```

After a broken pipe, \`BaseSourceTask\` auto-restarts and reloads \`effectiveOffset\` from \`ConfigurableOffsetBackingStore\`, which lags dbz's in-memory position by one barrier + async commit cycle. dbz re-streams the missed WAL, re-runs \`updateCommitPosition\` with smaller LSNs, and emits the regressed values to the RW state table.

Customer log: \`lsn_commit\` dropped from \`3,513,483,312,760\` → \`3,513,415,253,584\` (≈65 MB) within 185 ms.

## Impact

- **\`lsn_commit\` regression** → next actor recovery → \`validateLogPosition\` fails (\`storedLsn < slot.restart_lsn\`) → fatal \`DebeziumException\` (\"change stream … no longer available\") → source permanently stuck.
- **\`lsn_proc\` regression** → \`startStreaming\` resume position predates PG \`restart_lsn\` → \`WalPositionLocator\` cannot locate dedup key → silent data loss / hung resume.
- **\`lsn\` regression alone is legitimate**: PG streams events in commit order, so the per-event physical LSN can drop under overlapping transactions. Guarding \`lsn\` produces false positives.

## Fix

Monotonicity guard in \`PostgresCdcSplit::update_offset\`:

- Compare \`lsn_commit\` and \`lsn_proc\` independently; either regressing rejects the update.
- Do not check \`lsn\`.
- Missing fields skip comparison (None-safe; never fails on absence).
- Rejected update: keep previous state, \`warn\`-log old/new, return \`Ok\`.

A rejected update only holds the split state at its previous value — the chunk has already been emitted downstream. Worst case: the next actor recovery replays a short window of events, idempotently absorbed by downstream MV upserts.


src/connector/src/source/cdc/split.rs
```diff
@@ -200,7 +200,43 @@ impl CdcSplitTrait for PostgresCdcSplit {
...
+        // -- by which time PG's slot.restart_lsn has typically advanced
+        // past it, causing a fatal validateLogPosition failure
+        // ("change stream ... no longer available").
+        //
+        // Snapshot-phase offsets carry no comparable LSN, so they always
+        // pass through. Heartbeat / non-LSN-bearing offsets also pass
+        // through (extract returns None).
+        if self.inner.snapshot_done
+            && new_snapshot_done
+            && let Some(new_lsn) = extract_postgres_lsn_from_offset_str(&last_seen_offset)
```
zwang28: (lsn_commit, lsn_proc)

src/connector/src/source/cdc/split.rs
```diff
@@ -200,7 +200,43 @@ impl CdcSplitTrait for PostgresCdcSplit {
...
+            && let Some(old_lsn) = self.pg_lsn()
```
zwang28: (lsn_commit, lsn_proc)

src/connector/src/source/cdc/split.rs
```diff
@@ -200,7 +212,71 @@ impl CdcSplitTrait for PostgresCdcSplit {
...
+            let commit_regressed = matches!(
+                (new_lsn_commit, old_lsn_commit),
+                (Some(new), Some(old)) if new < old
+            );
+            let proc_regressed = matches!(
+                (new_lsn_proc, old_lsn_proc),
+                (Some(new), Some(old)) if new < old
+            );
+
+            if commit_regressed || proc_regressed {
```
zwang28: please refer to https://github.com/risingwavelabs/risingwave/pull/22503 for the correct approach to comparing PostgreSQL LSNs.

src/connector/src/source/cdc/split.rs
```diff
@@ -200,7 +212,71 @@ impl CdcSplitTrait for PostgresCdcSplit {
     }
 
     fn update_offset(&mut self, last_seen_offset: String) -> ConnectorResult<()> {
-        self.inner.snapshot_done = self.extract_snapshot_flag(last_seen_offset.as_str())?;
+        let new_snapshot_done = self.extract_snapshot_flag(last_seen_offset.as_str())?;
```
zwang28, line 215: The extract_snapshot_flag will never return true. Please verify.

zwang28 (commented):
Please address the comments.

zwang28 (approved):

PR #25626, opened 2026-05-12
fix(optimizer): fix split now or rule

chenzl25:

- Fix `SplitNowOrRule` to preserve SQL three-valued logic when splitting `now_arm OR others`.

- The previous rewrite used `NOT others` for the `now()` branch, which incorrectly drops rows where `others` evaluates to `NULL` and the `now()` arm is `TRUE`. This changes the split condition to `others IS NOT TRUE`, keeping `UNION ALL` branches disjoint while preserving nullable predicate semantics.

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/src/frontend/planner_test/tests/testdata/output/temporal_filter.yaml
+++ b/src/frontend/planner_test/tests/testdata/output/temporal_filter.yaml
@@ -357,9 +357,9 @@
       │ └─StreamProject { exprs: [t1.ts, t1._row_id, 0:Int32] }
       │   └─StreamDynamicFilter { predicate: ($expr1 > now), output_watermarks: [[$expr1]], output: [t1.ts, $expr1, t1._row_id], cleaned_by_watermark: true }
       │     ├─StreamProject { exprs: [t1.ts, AddWithTimeZone(t1.ts, '01:00:00':Interval, 'UTC':Varchar) as $expr1, t1._row_id] }
-      │     │ └─StreamFilter { predicate: Not((t1.ts > '2023-12-18 00:00:00+00:00':Timestamptz)) }
+      │     │ └─StreamFilter { predicate: IsNotTrue((t1.ts > '2023-12-18 00:00:00+00:00':Timestamptz)) }
       │     │   └─StreamShare { id: 2 }
-      │     │     └─StreamFilter { predicate: IsNotNull(t1.ts) }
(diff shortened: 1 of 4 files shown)


src/frontend/src/optimizer/rule/stream/split_now_or_rule.rs
```diff
@@ -64,15 +65,23 @@ impl Rule<Logical> for SplitNowOrRule {
             return None;
         }
```
xiangjinwu, line 66: nit: `let [now] = now.try_into().ok()?;` so that we do not need to `.expect("there should be exactly one now() arm")` below - in case we later updated one place and fail to keep them in sync

src/frontend/planner_test/tests/testdata/output/temporal_filter.yaml
```diff
@@ -357,18 +357,41 @@
       │ └─StreamProject { exprs: [t1.ts, t1._row_id, 0:Int32] }
       │   └─StreamDynamicFilter { predicate: ($expr1 > now), output_watermarks: [[$expr1]], output: [t1.ts, $expr1, t1._row_id], cleaned_by_watermark: true }
       │     ├─StreamProject { exprs: [t1.ts, AddWithTimeZone(t1.ts, '01:00:00':Interval, 'UTC':Varchar) as $expr1, t1._row_id] }
-      │     │ └─StreamFilter { predicate: Not((t1.ts > '2023-12-18 00:00:00+00:00':Timestamptz)) }
+      │     │ └─StreamFilter { predicate: IsNotTrue((t1.ts > '2023-12-18 00:00:00+00:00':Timestamptz)) }
       │     │   └─StreamShare { id: 2 }
-      │     │     └─StreamFilter { predicate: IsNotNull(t1.ts) }
+      │     │     └─StreamFilter { predicate: (IsNotTrue((t1.ts > '2023-12-18 00:00:00+00:00':Timestamptz)) OR (t1.ts > '2023-12-18 00:00:00+00:00':Timestamptz)) }
```
xiangjinwu, line 362: Seems we used to simplify `NOT P OR P` into `P IS NOT NULL` and similarly we could now simplify `P IS NOT TRUE OR P` into `TRUE` (no Filter)

xiangjinwu (approved):

PR #25634, opened 2026-05-12
fix(meta): skip missing drop jobs in barrier worker

### LINKED ISSUE #25516: bug: meta node panic
### Describe the bug

thread 'rw-main' (11) panicked at src/meta/src/barrier/info.rs:1293:30:

### Error message/log

```text
should exist
```

### To Reproduce

_No response_

### Expected behavior

_No response_

_No response_

### The version of RisingWave

v2.8

### Additional context

_No response_

wenym1:

This PR fixes the meta-node panic reported in #25516 when the global barrier worker applies a `DropStreamingJobs` command after one of the jobs in that command has already been removed from inflight barrier state.

The change keeps the original drop sequencing in `apply_command`: it first applies upstream sink cleanup, then collects base info, and only then removes inflight jobs. During that post-collect removal loop, the barrier worker now tries to remove each job one by one. If a job is already missing, it logs a warning and continues instead of panicking on the old `should exist` path.

The command payload is also simplified by removing `unregistered_fragment_ids`, since dropped fragments can be derived from the removed inflight job state directly in the barrier worker.

- Closes #25516

## Checklist

- [x]  I have checked the [Release Timeline](https://github.com/risingwave-labs/rw-commits-history/blob/main/release_timeline.md) and [Currently Supported Versions](https://docs.risingwave.com/changelog/release-support-policy#support-end-dates-for-recent-releases) to determine which release branches I need to cherry-pick this PR into. 

## Documentation

<details>

Fixed a meta-node panic that could happen while dropping streaming jobs if the barrier worker had already removed one of the dropped jobs from inflight state.

</details>

### DIFF
--- a/src/meta/src/barrier/checkpoint/state.rs
+++ b/src/meta/src/barrier/checkpoint/state.rs
@@ -799,21 +799,9 @@ impl DatabaseCheckpointControl {
 
             Some(Command::DropStreamingJobs {
                 streaming_job_ids,
-                unregistered_fragment_ids,
+                unregistered_state_table_ids: _,
                 dropped_sink_fragment_by_targets,
-                ..
             }) => {
-                let actors = self
-                    .database_info
-                    .fragment_infos()
-                    .filter(|fragment| {
-                        self.database_info
-                            .job_id_by_fragment(fragment.fragment_id)
-                            .is_some_and(|job_id| streaming_job_ids.contains(&job_id))
-                    })
-                    .flat_map(|fragment| fragment.actors.keys().copied())
-                    .collect::<Vec<_>>();
-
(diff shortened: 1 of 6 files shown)


wenym1:
* **#25634** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25634?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25634?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Li0k (approved):

PR #25647, opened 2026-05-13
fix(meta): keep failed foreground DDL tables fetchable

wenym1:

This PR fixes a foreground DDL recovery cleanup path that can leave Hummock `state_table_info` containing table ids that are no longer fetchable through the catalog RPC.

For foreground `CREATE MATERIALIZED VIEW` with slow snapshot backfill, the first barrier can pass and register state tables before the DDL waits for backfill completion. If database recovery happens before backfill finishes, the create job returns an error. The old error path called `try_abort_creating_streaming_job(..., false)`, which deleted catalog objects without recording the dropped tables for `get_tables(..., include_dropped = true)`. That breaks the compactor/meta-client assumption that every table id in Hummock version state table info can still be fetched from meta.

The fix is to avoid the foreground create failure path calling `try_abort_creating_streaming_job`. The abort helper now represents cancellation semantics only, so it always preserves dropped table metadata for catalog RPC lookup by table id.

This also adds a deterministic simulation integration test for the special path:

- create a foreground MV with snapshot backfill and a low backfill rate limit;
- kill a compute node after Hummock state tables are created but before backfill finishes;
- enter database recovery;
- fetch current Hummock `state_table_info` via `MetaClient`;
- assert every Hummock state table id can be returned by `get_tables(..., true)`.

I also verified the test fails on the old behavior with:

```text
tables [19, 20] are in hummock version state table info but cannot be fetched from catalog RPC
```

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12948,6 +12948,7 @@ dependencies = [
  "risingwave_meta_node",
  "risingwave_object_store",
  "risingwave_pb",
+ "risingwave_rpc_client",
  "risingwave_sqlparser",
  "risingwave_sqlsmith",
  "sea-orm",
--- a/src/meta/src/barrier/mod.rs
+++ b/src/meta/src/barrier/mod.rs
@@ -58,7 +58,7 @@ pub use self::command::{
 };
 pub(crate) use self::info::{SharedActorInfos, SharedFragmentInfo};
 pub use self::manager::{BarrierManagerRef, GlobalBarrierManager};
-pub use self::schedule::BarrierScheduler;
+pub use self::schedule::{BarrierScheduler, CommandIssueError};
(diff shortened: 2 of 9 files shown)


wenym1:
* **#25647** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25647?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25647?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Li0k (approved):
LGTM, Please add a note in the doc: these dropped-tables will remain temporarily until the next recovery.

PR #25648, opened 2026-05-13
feat(optimizer): support column pruning for expand

chenzl25:

- Column pruning is now supported for the expand operator. Previously, column pooling was not supported for the expand operator, which caused an issue: when a system column was selected, it could not be verified. For example, the rw_timestamp could not be pruned causing failure of creating mv.

- Closes #[ISSUE_NUMBER]

## Checklist

## Documentation

<details>

</details>

### DIFF
--- a/e2e_test/batch/grouping_sets/grouping_sets.slt
+++ b/e2e_test/batch/grouping_sets/grouping_sets.slt
@@ -16,5 +16,11 @@ NULL L 15 1 0 2 2
 NULL M 35 1 0 2 2
 NULL NULL 50 1 1 3 4
 
+query I rowsort
+SELECT count(*) FROM items_sold GROUP BY GROUPING SETS ((), ());
+----
+4
+4
+
 statement ok
 drop table items_sold;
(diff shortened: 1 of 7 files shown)


yuhao-su:
Will it be a problem if a downstream only use flag column, in which case the column_subsets will be empty. But can we write a query like that?

chenzl25:
Yes, it can happen with grouping sets, e.g.

```sql
select count(*) from t group by grouping sets ((), ());
```

After pruning, this produces `LogicalExpand { column_subsets: [[], []] }`. The current rewrite keeps/maps the flag correctly, and I added a planner regression test for this case in 8f92edd641.

yuhao-su:
> Yes, it can happen with grouping sets, e.g.
> 
> ```sql
> select count(*) from t group by grouping sets ((), ());
> ```
> 
> After pruning, this produces `LogicalExpand { column_subsets: [[], []] }`. The current rewrite keeps/maps the flag correctly, and I added a planner regression test for this case in 8f92edd641.

does the executor handle this case correctly?

chenzl25:
> > Yes, it can happen with grouping sets, e.g.
> > ```sql
> > select count(*) from t group by grouping sets ((), ());
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > After pruning, this produces `LogicalExpand { column_subsets: [[], []] }`. The current rewrite keeps/maps the flag correctly, and I added a planner regression test for this case in [8f92edd](https://github.com/risingwavelabs/risingwave/commit/8f92edd64170eb64388fa4a843467dd301eb105d).
> 
> does the executor handle this case correctly?

yes it could be handled

yuhao-su (approved):

yuhao-su:
>LogicalExpand { column_subsets: [[], []] }

Idealy we can have a e2e cover this case

PR #25651, opened 2026-05-13
feat(frontend): support ALTER WATERMARK for live tables

yuhao-su:
## Summary

Add `ALTER TABLE foo ALTER WATERMARK FOR col AS expr [WITH TTL]` to change a table's watermark expression without drop/recreate.

- v1 scope: expression only. Toggling `WITH TTL` is rejected — that would change `stream_key` / `clean_watermark_indices`, which `fit_internal_tables_trivial` cannot safely reconcile (deferred to v2).
- Reuses the `ReplaceStreamJob` flow from `ALTER TABLE … ALTER COLUMN`. Internal table IDs are preserved, so downstream MV state and CDC pipelines survive the replace.
- Two-layer guard: frontend AST check (column exists, `with_ttl` unchanged) + post-plan diff (`stream_key`, `pk`, `watermark_columns`, `clean_watermark_indices`, column types) as defense-in-depth.

## Test plan

- [x] `./risedev check`
- [x] `./risedev test` — 2458 pass
- [x] `cargo test -p risingwave_sqlparser parse_alter_table_alter_watermark`
- [x] `./risedev slt e2e_test/ddl/alter_watermark.slt` — basic / timestamp / append-only / functional / function-wrapped / no-op / rejections
- [x] `./risedev slt e2e_test/source_inline/cdc/mysql/mysql_alter_watermark.slt` — CDC source → table → MV, ALTER preserves both layers, sentinel pattern for absence check


chenzl25 (approved):
LGTM!

Issue #23569, opened 2025-10-25
[Feature Request] Support custom S3 sink partitioning using event timestamp from payload


qoolqool:
### Is your feature request related to a problem? Please describe.

Currently, RisingWave's S3 sink connector only supports time-based partitioning using the path_partition_prefix parameter (month/day/hour), which appears to partition based on the file creation or processing time rather than event timestamps within the data payload.
This limitation makes it difficult to organize data in S3 based on when events actually occurred (event time), which is a common requirement for:

- Data lake architectures where downstream analytics queries filter by event time
- Compliance and auditing where data must be organized by business/transaction time
- Time-based data retention policies based on event occurrence
- Integration with OLAP engines that expect partitioned data by event time for query optimization

### Describe the solution you'd like

Add support for custom partitioning in S3 sink that allows users to specify a timestamp field from the event payload to determine the partition path, similar to Kafka Connect's TimeBasedPartitioner with RecordField timestamp extractor.

```
CREATE SINK s3_sink FROM source_table
WITH (
    connector = 's3',
    s3.region_name = 'us-east-1',
    s3.bucket_name = 'my-bucket',
    s3.path = 'events/',
    
    -- New parameters for event-time partitioning
    partitioner.type = 'event_time',           -- or 'field_based_time'
    partitioner.field = 'event_timestamp',     -- field name from payload
    partitioner.format = 'year=YYYY/month=MM/day=dd/hour=HH',  -- partition path format
    partitioner.timezone = 'UTC',              -- optional, default UTC
    
    type = 'append-only',
    force_append_only = 'true'
) 
FORMAT PLAIN ENCODE PARQUET(force_append_only='true');
```

**Expected Behavior**

1. Field Extraction: Extract the timestamp value from the specified field in the event payload
2. Timestamp Parsing: Support common timestamp formats (TIMESTAMP, TIMESTAMP WITH TIME ZONE, epoch milliseconds, ISO 8601 strings)
3. Partition Path Generation: Use the extracted timestamp to generate the partition path according to the specified format
4. File Organization: Write events to S3 under the appropriate partition based on their event timestamp
5. Late Data Handling: Consider how to handle late-arriving data (events with older timestamps arriving after partition has been "closed")

### Describe alternatives you've considered

**Use Iceberg Sink: **
- While Iceberg supports partition_by, it requires additional infrastructure (catalog, metadata management) that may not be needed for simple S3 data lakes
- HA management of catalog service is another big exercise

### Additional context

[...]

PR #25662, opened 2026-05-14
feat(Sinks): added custom partitioning support for OpenDAL-backed file sinks 

almostintuitive:

This PR adds Hive-style partitioning support to OpenDAL-backed file sinks, including S3, GCS, Azure Blob, WebHDFS, local FS, and Snowflake. It introduces two new sink options:

- `path_partition_format`: a free-form path template that supports chrono `strftime` tokens and `{column_name}` placeholders, for layouts such as `exchange={exchange}/symbol={symbol}/year=%Y/month=%m/day=%d/hour=%H/`.
- `event_time_field`: an optional `TIMESTAMP` or `TIMESTAMPTZ` column used to render chrono tokens from each row's event time instead of the writer's flush time.

The intention is to let file sinks write Parquet/JSON data directly into Hive-compatible directory layouts that are easy for downstream engines such as Trino, Athena, DuckDB, Spark, and Iceberg readers to discover and prune. The existing `path_partition_prefix` behavior remains supported for backwards compatibility, but `path_partition_format` is the preferred option for new pipelines because it supports arbitrary segment ordering and column-based partitioning.

How it works:

- `PartitionTemplate` parses `path_partition_format` at sink creation time, validates chrono tokens, resolves column references against the sink schema, and rejects invalid placeholders early.
- `OpenDalSinkWriter` groups visible rows by rendered partition path when column placeholders or `event_time_field` are used, then writes one file per partition group.
- The writer keeps active object writers in a `HashMap<partition_path, FileWriterEnum>` so one sink can fan out to multiple partition directories during a flush.
- Column partition values are escaped for object-path safety. `/`, `\`, `%`, NUL, and control characters are percent-encoded, and `NULL` values render as `__HIVE_DEFAULT_PARTITION__`.
- `TIMESTAMP` event-time values are interpreted as UTC; `TIMESTAMPTZ` values are converted to UTC before rendering.
- Active partition writers are closed with bounded concurrency to avoid unbounded object-store close/finalization fan-out.

Limitations:

- Partition columns are still included in the output files; this PR does not implement partition-column elision.
- File sinks do not provide transactional commits across all partition files. If object-store close/finalization partially succeeds and then fails, already committed files may be visible before the sink retries.
- `event_time_field` only supports `TIMESTAMP` and `TIMESTAMPTZ`.

Refs: risingwavelabs/risingwave#23569

## Checklist

- [x] I have checked the [Release Timeline](https://github.com/risingwavelabs/rw-commits-history/blob/main/release_timeline.md) and [Currently Supported Versions](https://docs.risingwave.com/changelog/release-support-policy#support-end-dates-for-recent-releases) to determine which release branches I need to cherry-pick this PR into.

## Documentation

<details>

File sinks can now write Parquet/JSON output into Hive-compatible partition directories with the new `path_partition_format` option. Users can combine time-based tokens and column placeholders, and can optionally set `event_time_field` so late-arriving rows are partitioned by event time instead of flush time.

Example:

```sql
CREATE SINK ohlcv_1d AS
  SELECT exchange, symbol, window_start AS event_time, open, high, low, close, volume
  FROM ohlcv_mv
  WITH (
    connector             = 's3',
    s3.region_name        = 'us-east-1',
    s3.bucket_name        = 'metrics-live',
    s3.path               = 'ohlcv/period=1d',
    path_partition_format = 'exchange={exchange}/symbol={symbol}/year=%Y/month=%m/day=%d/hour=%H/',
    event_time_field      = 'event_time',
    type                  = 'append-only',
    force_append_only     = 'true'
  ) FORMAT PLAIN ENCODE PARQUET(force_append_only = 'true');
```

</details>


wcy-fdu:
Hi @almostintuitive , sorry for the late review and thanks for your contribution!
Still reviewing; one request on the testing side first.

The current coverage is mostly at the unit-test level plus the on-disk directory layout (walkdir asserting the path structure), but there's no end-to-end round-trip yet confirming the written data can be read back in full without loss. Could you add an e2e test modeled on the existing [e2e_test/s3/file_sink.py](e2e_test/s3/file_sink.py) that exercises the new `path_partition_format` (ideally the `{column}` placeholder + `event_time_field` combination)?

- sink the data to S3 in the Hive partition layout;
- read those partition directories back via a file source;
- assert the row count and contents match the source, confirming the loop closes with no data loss.

Thanks!

wcy-fdu:
Took a look at the overall design and wanted to check one thing with you. The per-row partitioning keeps all active writers in a HashMap, one per partition, and only closes them together at commit. Since batching is decoupled from checkpoints, the number of writers a single actor holds open between commits is just however many distinct partitions it has seen, which is bounded only by data cardinality, not by anything else.

Per-writer memory is fine (OpenDAL streams via multipart, each writer roughly holds one part buffer), but the count isn't. If you partition by something high-cardinality like symbol or user_id and see tens of thousands of distinct values within a batching window, you end up with that many writers open at once, each holding its buffer and encoder state, and a single node will OOM pretty easily. It also produces a ton of tiny files, which hurts downstream readers. The MAX_CONCURRENT_PARTITION_CLOSES = 16 doesn't help here, since it only limits close concurrency at commit time, while the OOM happens earlier, during accumulation.

I think this needs a cap on how many partitions can be open at once, flushing the oldest writer when the limit is hit, or erroring out so the user adjusts their partitioning. Curious first though: roughly how many distinct partitions show up per flush in your actual use case? If the symbol count is bounded it might be fine in practice, but at a finer grain this OOM seems hard to avoid.

PR #25663, opened 2026-05-14
feat(frontend): keep sync log store node at fragment root

EdwinaZhu:

This PR is to apply the optimization #24303. 

- Summary: Ensure each `StreamSyncLogStore` is rooted in its own stream fragment, and add a streaming config/session kill switch for the optimized sync log store dispatcher path.
- How it works: `StreamSyncLogStore::new_as_fragment_root` and `ensure_as_fragment_root` wrap a sync log store with a no-shuffle `StreamExchange`; the new `EnsureSyncLogStoreRootRule` runs bottom-up for unaligned joins and inserts no-shuffle exchanges above sync log store children that are not already separated by an exchange.
- Logical changes: `StreamMaterialize` and sink-into-table planning now use the fragment-root helper, planner tests cover separated consecutive unaligned joins(ensuring no duplicated consecutive `StreamExchange`) and sink-into-table plans, and streaming config/session overrides now expose `streaming.developer.disable_sync_log_store_dispatcher` for new streaming jobs. User could run `SET streaming_disable_sync_log_store_dispatcher = true;` to disable create `SyncLogStoreDispatcher `.

## Checklist

## Documentation

<details>

Added `streaming.developer.disable_sync_log_store_dispatcher`, of which the default value is false. It could be override in session-level, so the user could disable the optimized sync log store dispatcher path for newly created streaming jobs when needed.

</details>


EdwinaZhu:
I added planner test in `src/frontend/planner_test/tests/testdata/input/join.yaml`, and its output is in `src/frontend/planner_test/tests/testdata/output/join.yaml`.

wenym1:
Updated the implementation based on the review discussion.

Summary:
- Replaced the optimizer-rule insertion with a final stream-plan rewrite in `StreamSyncLogStore` code.
- The rewrite recursively processes inputs and inserts a no-shuffle `StreamExchange` only when a non-exchange parent has a direct `StreamSyncLogStore` child. This avoids consecutive exchanges.
- Apply the rewrite after the final stream job root is created for MV/table/sink/index paths, before trace/explain output and graph building consume the plan.
- Removed the newly added session var while keeping the streaming config override for disabling the sync kv log store dispatcher.
- Added planner coverage for the distributed graph path and updated the decoupled sink-into-table physical explain output.

Validation:
- `cargo fmt`
- `./risedev do-apply-planner-test` (109 passed)
- `cargo clippy --all-targets --all-features` passed, with only the existing `rquickjs-core` future-incompat warning.

wenym1:
Addressed the latest review comment on `StreamSyncLogStore`: `ensure_sync_log_store_fragment_root` now uses an internal `EnsureSyncLogStoreFragmentRootRewriter` implementing the existing `PlanRewriter<Stream>` framework instead of hand-written recursion.

The rewriter still preserves the previous behavior: bottom-up rewrite, skip wrapping under `StreamExchange`, insert no-shuffle exchange only for direct `StreamSyncLogStore` children of non-exchange parents, and avoid rebuilding unchanged nodes.

Validation after the refactor:
- `cargo fmt`
- `./risedev do-apply-planner-test` (109 passed)
- `cargo clippy --all-targets --all-features` passed, with only the existing `rquickjs-core` future-incompat warning.

wenym1 (approved):

wenym1:
Fixed the merge-queue unit failure from Buildkite 96100. The failed test was `risingwave_planner_test::planner_test_runner explain_dot_format`; regenerating the planner output updates the sync log store table's `read pk prefix len hint` from `3` to `0` in `explain_dot_format.yaml`.

Verification:
- `./risedev run-planner-test explain_dot_format`

PR #25665, opened 2026-05-14
fix: propagate table refill policy via hummock notification

Li0k:

This PR moves table cache refill runtime state to Hummock `TableRefillRuntimeConfig` and limits runtime refresh to recovery/subscribe boundaries, not online updates.

- Keep `ALTER SET CONFIG` as persistence and validation only; it does not refresh CN runtime state online.
- Restore table cache refill policies after recovery with a full policy snapshot, and include both policies and per-worker serving table vnode mappings in Hummock subscribe snapshots.
- Do not support online runtime updates for either table cache refill policies or serving table vnode mappings in this PR.
- Add regression coverage for stale runtime config filtering, recovery/snapshot refresh, initial serving snapshot apply, and ignoring non-snapshot serving mapping payloads.

## Checklist

## Documentation

<details>

Not user-facing.

</details>

### DIFF
--- a/proto/meta.proto
+++ b/proto/meta.proto
@@ -634,6 +634,8 @@ message MetaSnapshot {
   repeated FragmentWorkerSlotMapping streaming_worker_slot_mappings = 21;
   repeated FragmentWorkerSlotMapping serving_worker_slot_mappings = 22;
   repeated ObjectDependency object_dependencies = 26;
+  TableCacheRefillPolicies table_cache_refill_policies = 27;
+  ServingTableVnodeMappings serving_table_vnode_mappings = 28;
 
   SnapshotVersion version = 13;
 }
--- a/proto/meta.proto
+++ b/proto/meta.proto
@@ -696,7 +696,7 @@ message SubscribeResponse {
     FragmentWorkerSlotMappings serving_worker_slot_mappings = 28;
     catalog.Secret secret = 29;
     common.ClusterResource cluster_resource = 31;
-    TableRefillRuntimeConfig table_refill_runtime_config = 34;
+    TableRefillRuntimeConfig table_refill_runtime_config = 32;
   }
   reserved 12;
   reserved "parallel_unit_mapping";
@@ -704,8 +704,6 @@ message SubscribeResponse {
   reserved "serving_parallel_unit_mappings";
   reserved 30;
   reserved "compute_node_total_cpu_count";
-  reserved 32, 33;
-  reserved "table_cache_refill_policies", "serving_table_vnode_mappings";
 }
 
 message TableRefillRuntimeConfig {
--- a/proto/meta.proto
+++ b/proto/meta.proto
@@ -634,10 +634,11 @@ message MetaSnapshot {
   repeated FragmentWorkerSlotMapping streaming_worker_slot_mappings = 21;
   repeated FragmentWorkerSlotMapping serving_worker_slot_mappings = 22;
   repeated ObjectDependency object_dependencies = 26;
-  TableCacheRefillPolicies table_cache_refill_policies = 27;
-  ServingTableVnodeMappings serving_table_vnode_mappings = 28;
+  TableRefillRuntimeConfig table_refill_runtime_config = 27;
 
   SnapshotVersion version = 13;
+  reserved 28;
+  reserved "serving_table_vnode_mappings";
 }
 
 message Object {
@@ -699,6 +700,7 @@ message SubscribeResponse {
     common.ClusterResource cluster_resource = 31;
     TableCacheRefillPolicies table_cache_refill_policies = 32;
     ServingTableVnodeMappings serving_table_vnode_mappings = 33;
+    TableRefillRuntimeConfig table_refill_runtime_config = 34;
   }
   reserved 12;
   reserved "parallel_unit_mapping";
@@ -706,6 +708,12 @@ message SubscribeResponse {
   reserved "serving_parallel_unit_mappings";
   reserved 30;
   reserved "compute_node_total_cpu_count";
+}
+
+message TableRefillRuntimeConfig {
+  TableCacheRefillPolicies table_cache_refill_policies = 1;
+  ServingTableVnodeMappings serving_table_vnode_mappings = 2;
+  uint64 version = 3;
 }
 
 message ServingTableVnodeMappings {
--- a/proto/meta.proto
+++ b/proto/meta.proto
@@ -637,8 +637,6 @@ message MetaSnapshot {
   TableRefillRuntimeConfig table_refill_runtime_config = 27;
 
   SnapshotVersion version = 13;
-  reserved 28;
-  reserved "serving_table_vnode_mappings";
 }
 
 message Object {
@@ -698,8 +696,6 @@ message SubscribeResponse {
     FragmentWorkerSlotMappings serving_worker_slot_mappings = 28;
     catalog.Secret secret = 29;
     common.ClusterResource cluster_resource = 31;
-    TableCacheRefillPolicies table_cache_refill_policies = 32;
-    ServingTableVnodeMappings serving_table_vnode_mappings = 33;
     TableRefillRuntimeConfig table_refill_runtime_config = 34;
   }
   reserved 12;
--- a/proto/meta.proto
+++ b/proto/meta.proto
@@ -704,6 +704,8 @@ message SubscribeResponse {
   reserved "serving_parallel_unit_mappings";
   reserved 30;
   reserved "compute_node_total_cpu_count";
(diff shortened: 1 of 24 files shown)


proto/meta.proto
```diff
@@ -695,15 +696,22 @@ message SubscribeResponse {
     FragmentWorkerSlotMappings serving_worker_slot_mappings = 28;
     catalog.Secret secret = 29;
     common.ClusterResource cluster_resource = 31;
-    TableCacheRefillPolicies table_cache_refill_policies = 32;
-    ServingTableVnodeMappings serving_table_vnode_mappings = 33;
+    TableRefillRuntimeConfig table_refill_runtime_config = 34;
```
hzxa21: The original PR is not merged so you can use id 32 instead of 34.

proto/meta.proto
```diff
@@ -695,15 +696,22 @@ message SubscribeResponse {
...
   }
   reserved 12;
   reserved "parallel_unit_mapping";
   reserved 23;
   reserved "serving_parallel_unit_mappings";
   reserved 30;
   reserved "compute_node_total_cpu_count";
+  reserved 32, 33;
+  reserved "table_cache_refill_policies", "serving_table_vnode_mappings";
+}
```
hzxa21: ditto, this is not needed.

src/meta/service/src/notification_service.rs
```diff
@@ -331,6 +332,13 @@ impl NotificationServiceImpl {
         let hummock_write_limits = self.hummock_manager.write_limits().await;
         let meta_backup_manifest_id = self.backup_manager.manifest().await.manifest_id;
         let cluster_resource = self.get_cluster_resource().await;
+        let table_refill_runtime_config = build_hummock_table_refill_runtime_config(
+            &self.metadata_manager,
+            &self.serving_vnode_mapping,
+            worker_key,
+            self.env.notification_manager().current_version().await,
+        )
+        .await?;
```
hzxa21: I think the retrivial of table_refill_runtime_config should be done under the same catalog lock as `get_tables_snapshot` in L327 to make sure the fields inside `MetaSnapshot` belong to the same catalog snapshot.

We can combine `get_tables_snapshot` and `build_hummock_table_refill_runtime_config`

src/meta/service/src/notification_service.rs
```diff
@@ -392,7 +401,7 @@ impl NotificationService for NotificationServiceImpl {
                 self.hummock_manager
                     .pin_version(req.get_worker_id())
                     .await?;
-                self.hummock_subscribe().await?
+                self.hummock_subscribe(&worker_key).await?
```
hzxa21: The reason why we need `worker_key` here is we need to derivec the worker_id in [build_hummock_serving_table_vnode_mapping](https://github.com/risingwavelabs/risingwave/blob/2b442a8018b5017ce286057215f2e1ae29863f56/src/meta/src/table_refill.rs#L36). I think it is better to pass `req.get_worker_id()` here.

src/storage/src/hummock/event_handler/hummock_event_handler.rs
```diff
@@ -358,20 +414,37 @@ impl HummockEventHandler {
...
+                _
+            ))
+        ) {
+            let Some(HummockObserverEvent::TableRefillRuntimeConfig(operation, config)) =
+                pending_observer_events.pop_front()
+            else {
+                unreachable!();
+            };
+            Self::apply_table_refill_runtime_config(&mut refiller, operation, config);
+        }
```
hzxa21, line 441: Why do we need the special logic here to poll the `observer_event_rx` and apply the table refill config? Will there be any issue if we rely on the event handler to apply the table refill runtime config initilization? Also, what if the `observer_event_rx` doesn't receive the TableRefillRuntimeConfig when `HummockEventHandler::new_inner` is called?

src/storage/src/hummock/event_handler/refiller.rs
```diff
@@ -271,7 +270,13 @@ pub struct TableCacheRefillContext {
     pub default_policy: CacheRefillPolicy,
 
     pub read_version_mapping: Arc<RwLock<ReadVersionMappingType>>,
```
hzxa21, line 272: This field is unneccessary.
1. It is only used in [update_table_cache_refill_vnodes_inner](https://github.com/risingwavelabs/risingwave/blob/2b442a8018b5017ce286057215f2e1ae29863f56/src/storage/src/hummock/event_handler/refiller.rs#L449)
2. `update_table_cache_refill_vnodes_inner` is only called by HummockEventHandler
3. HummockEventHandler already maintion a lock-free [local_read_version_mapping](https://github.com/risingwavelabs/risingwave/blob/2b442a8018b5017ce286057215f2e1ae29863f56/src/storage/src/hummock/event_handler/hummock_event_handler.rs#L251)
4. We can just pass `local_read_version_mapping` to `update_table_cache_refill_vnodes_inner`.

This will avoid lock contention.

src/storage/src/hummock/event_handler/refiller.rs
```diff
@@ -347,14 +352,13 @@ impl CacheRefiller {
     ) -> Self {
         let config = Arc::new(config);
         let concurrency = Arc::new(Semaphore::new(config.concurrency));
-        let serving_table_vnode_mapping = Arc::new(RwLock::new(HashMap::new()));
         let table_cache_refill_context = Arc::new(RwLock::new(TableCacheRefillContext {
```
hzxa21, line 355: I think there is no need to store `table_cache_refill_context` as a top-level field in `CacheRefiller` given that it is already part of `CacheRefiller::context`

src/storage/src/hummock/event_handler/refiller.rs
hzxa21: Also, I think we can organize the fields in `TableCacheRefillContext` better to further avoid contention between bitmap/policy update and refiller task check.
1. Only `streaming`, `serving` `policies` are used in the [refiller task check](https://github.com/risingwavelabs/risingwave/blob/2b442a8018b5017ce286057215f2e1ae29863f56/src/storage/src/hummock/event_handler/refiller.rs#L284)
2. Other fields are used to update `streaming`, `serving` `policies` but not directly needed in refiller task.

I suggest the following:
- Keep fields in 1 in TableCacheRefillContext and move fields in 2 to upper level CacheRefiller
- When spwaning a refiller task, clone the neccessary fields in 1 corresponding to the table ids of the task instead of passing the refill context protected by RwLock.

```rust
pub(crate) struct CacheRefiller {
    /// order: old => new
    queue: VecDeque<Item>,

    context: CacheRefillContext,

    spawn_refill_task: SpawnRefillTask,

    role: Role,

    // Remove this top level field
    // table_cache_refill_context: Arc<RwLock<TableCacheRefillContext>>,

    // Move these fields from TableCacheRefillContext to here. `read_version_mapping` is not needed.
    pub default_policy: CacheRefillPolicy,
    pub serving_table_vnode_mapping: HashMap<TableId, Bitmap>,

    // Use a map to store the table refill context for all tables
    pub table_cache_refill_context_map: HashMap<TableId, TableCacheRefillContext>
}

// This struct now stores the per table context.
pub struct TableCacheRefillContext {
    pub streaming_vnode_bitmap: Bitmap,
    pub serving_vnode_bitmap: Bitmap,
    pub policies: CacheRefillPolicy
}

impl CacheRefiller {
    // Call this to collect the context when spawning the refiller task
    fn get_table_cache_refill_context(&self, table_ids: ...) -> ... {
        self.table_cache_refill_context_map.filter(...).collect()
    }
}
```

hzxa21:
I think this PR is conflating two different kinds of runtime updates, which should have different expectations.

For ALTER ... SET CONFIG, I agree it is acceptable to make it persistence-only and apply it only after recovery / resubscribe. That changes the refill policy itself, so "which tables should refill" does not necessarily need online update.

But vnode ownership changes caused by scaling are different. Even if the refill policy stays the same, the set of vnodes a table should refill on a given CN changes when CN ownership changes. My understanding is:

- for streaming, vnode ownership can still be refreshed online through the existing read-version / vnode-bitmap update path, so refill behavior can converge to the post-scale state;
- for serving, the old online update path has been removed, and serving_table_vnode_mappings is now only refreshed when the node receives a new snapshot.

If that is correct, then after scaling, streaming refill behavior can still track the new ownership, but serving refill behavior cannot unless recovery / resubscribe happens.

I think we should make the behavior consistent for both streaming and serving, which is:
- refill policy update does not need online update;
- vnode mapping update still does, because ownership changes independently of policy.

hzxa21 (commented):
Left some comments. Some are related to the original PR.

PR #25670, opened 2026-05-15
fix(grafana): stop labeling streaming tables as mviews

lmatz:
## Summary
- stop hardcoding `mview` / `materialized view` in relation-level Grafana panels
- join `table_type` from `table_info` so legends show the real object type
- rename the affected panel titles/descriptions to `Streaming Relation ...`

## Scope
This PR only changes the Grafana presentation layer:
- `grafana/dashboard/user/overview.py`
- `grafana/dashboard/user/streaming.py`
- `grafana/dashboard/dev/streaming_relations.py`
- `grafana/dashboard/dev/hummock_write.py`
- `grafana/dashboard/user/memory.py`
- regenerated dashboard JSON outputs under `grafana/` and `docker/dashboards/`

## Why
Today a TABLE can appear in Grafana under legends like `mview 1620 product` because:
- the metric names historically use `stream_mview_*`
- the dashboards hardcode `mview` / `materialized view` in panel titles, descriptions, and legends
- the PromQL only joined `table_name`, even though `table_info` already exposes `table_type`

That makes the monitoring surface look like the underlying catalog object is always an MV, which is misleading for users and for any tool grounding off the dashboard JSON.

## Follow-up intentionally deferred
1. Renaming or aliasing `stream_mview_*` metrics
2. Renaming the `materialized_view_id` label in `table_info`
3. Auditing `grafana/dashboard/user/storage.py` separately, because `storage_materialized_view_stats` may still be MV-only


lmatz:
Affected panels in this PR:

User dashboard:
- `Streaming Relation Throughput(rows/s)`
- `Streaming Relation Throughput (rows/s)`
- `Latency of Streaming Relations & Sinks`
- `Memory Usage (Detailed)`

Dev dashboard:
- `Latency of Streaming Relations & Sinks`
- `Current Epoch of Streaming Relations`
- `Streaming Relation Throughput (rows/s)`
- `Executor Cache Memory Usage of Streaming Relations`
- `Streaming Relation Write Size`

More precisely:
- the first 6 panels above are the ones where the query/legend path was made type-aware (`TABLE` / `MATERIALIZED_VIEW` / `INDEX`)
- the last 3 are wording-only neutralization changes to remove misleading `materialized view` terminology without changing panel aggregation semantics

lmatz:
SCR-20260515-k0f

tabVersion (approved):

PR #25672, opened 2026-05-15
fix(frontend): disable unused prefix bloom hints

Li0k:

This PR disables read-prefix bloom/filter hints for internal tables whose runtime read patterns do not use fixed-prefix lookups.

Frontend catalog hint changes:

- Dynamic filter left internal table: sets `read_prefix_len_hint` to `0`. The executor replays predicate changes by scanning left-key ranges with `iter_with_vnode`, not by fixed-prefix lookups.
- Refresh materialize staging table: sets staging `read_prefix_len_hint` to `0`. Refresh merge consumes staging rows with keyed range scans, while the main materialize table and refresh progress table keep their existing hints.
- Vector index table: sets `read_prefix_len_hint` to `0`. Vector index writes go through vector writer APIs rather than `StateTable` get/prefix-iter reads.

Meta migration:

- Adds a migration to update existing persisted fragment stream-node table catalogs and corresponding rows in the meta store `table` table.
- The migration only targets dynamic-filter left tables, refresh staging tables, vector-index write tables, kv log store sink tables, and synced kv log store tables.
- It intentionally does not update dynamic-filter right tables, refresh progress tables, source tables, locality-provider tables, or unrelated internal tables.

CI labeler:

- The existing `.github/labeler.yml` rule already applies `ci/run-backwards-compat-tests` to migration files under `src/meta/model/migration/`, so no duplicate labeler rule is needed.

## Checklist

## Documentation

<details>

No user-facing behavior change.

</details>

## Verification

- `cargo test -p risingwave_meta_model_migration disable_unused_read_prefix_hints --lib`
- `./risedev run-planner-test temporal_filter`
- `git diff --check`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12682,6 +12682,8 @@ version = "3.1.0-alpha"
 dependencies = [
  "easy-ext",
  "madsim-tokio",
+ "prost 0.14.3",
+ "risingwave_pb",
  "sea-orm",
  "sea-orm-migration",
  "serde",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12688,6 +12688,7 @@ dependencies = [
  "sea-orm-migration",
  "serde",
  "serde_json",
+ "thiserror-ext",
  "uuid",
 ]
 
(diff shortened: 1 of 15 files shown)


hzxa21 (approved):
LGTM. For the jobs already created with these executors, should we have some migration logic to change the read_pk_prefix_len in the table catalog as well?

hzxa21 (approved):

PR #25673, opened 2026-05-15
fix(storage): skip empty xor filters

Li0k:

- Skip writing xor filter bytes for single-table no-filter SSTs, such as tables with `read_prefix_len_hint = 0`.
- Allow fast compaction to copy data blocks for confirmed single-table no-filter tasks while discarding raw filter data.
- Keep mixed-table SSTs on the existing filter path unless their metadata has already been pruned to a single live no-filter table.

- Closes N/A

## Checklist

- [x] I have written necessary rustdoc comments. (N/A: no new public API.)
- [x]  I have added test labels as necessary.  (N/A: default PR CI is enough.)
- [x]  I have added fuzzing tests or opened an issue to track them.  (N/A.)
- [x]  My PR contains breaking changes.  (N/A: no breaking changes.)
- [x]  My PR changes performance-critical code, so I will run (micro) benchmarks and present the results.  (N/A: covered by focused tests.)
- [x]  I have checked the [Release Timeline](https://github.com/risingwavelabs/rw-commits-history/blob/main/release_timeline.md) and [Currently Supported Versions](https://docs.risingwave.com/changelog/release-support-policy#support-end-dates-for-recent-releases) to determine which release branches I need to cherry-pick this PR into.  (Backport can be decided separately for the target bugfix release.)

## Documentation

- [x]  My PR needs documentation updates.  (N/A: no user-facing behavior change.)

<details>

No user-facing behavior change.

</details>

## Verification

- Format and diff checks passed.
- Storage clippy passed with warnings denied.
- Focused storage unit tests passed.

### DIFF
--- a/src/storage/src/compaction_catalog_manager.rs
+++ b/src/storage/src/compaction_catalog_manager.rs
@@ -503,6 +503,27 @@ impl CompactionCatalogAgent {
         self.filter_key_extractor_manager.extract(user_key)
     }
 
+    /// Returns true when this compaction task only contains one table and the table never emits
+    /// filter keys.
+    pub fn is_single_table_no_filter(&self) -> bool {
+        let mut table_ids = self.table_ids();
+        let Some(table_id) = table_ids.next() else {
+            return false;
+        };
+        if table_ids.next().is_some() {
+            return false;
+        }
+
+        match &self.filter_key_extractor_manager {
+            FilterKeyExtractorImpl::Dummy(_) => true,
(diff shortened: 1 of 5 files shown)


hzxa21:
We had an offline discussion. This PR is no longer needed because empty xor filter keys are already excluded: https://github.com/risingwavelabs/risingwave/blob/ea5308ec732977f8a8c8c46ee4b78a3b07d755f3/src/storage/src/hummock/sstable/builder.rs#L409

Issue #7412, opened 2023-01-16
bug: can't correctly return NULL when rows satisfy `sum(...) filter (...)` are all deleted


richardchien:
### To Reproduce

```sql
=> create table t (a int, b int);
=> create materialized view mv as select a, sum(b) filter (where b > 10) as sum from t group by a;

=> insert into t values (1, 1), (1, 11), (2, 14), (3, 0); flush;
=> select * from mv;
 a | sum
---+-----
 3 |       <- returned NULL correctly
 1 |  11
 2 |  14
(3 rows)

=> delete from t where a = 1 and b = 11; flush;
=> select * from mv;

 a | sum
---+-----
 3 |
 2 |  14
 1 |   0   <- incorrect
(3 rows)
```

maybe-vibe:
The `FILTER` in the title is incidental. The same defect hits any streaming aggregate whose state is a single retractable value: once a group has seen a non-null input, its state never returns to NULL, so a group that keeps only all-NULL rows after deletes reports a non-NULL value where batch reports NULL. Verified on v3.1.0-rc.1 (e4644f6), no `FILTER` involved:

```sql
create table at (id int primary key, k int, v int, b bool);
create materialized view at_hash as
  select k, count(*) c, sum(v), bit_xor(v), bit_and(v), bit_or(v), bool_and(b), bool_or(b), count(v)
  from at group by k;
insert into at values (1, 1, 5, true), (2, 1, null, null), (3, 1, 6, false);
flush;
delete from at where id in (1, 3);
flush;
select * from at_hash;
-- streaming: 1 | 1 | 0 | 0 | -1 | 0 | t | f | 0
select k, count(*), sum(v), bit_xor(v), bit_and(v), bit_or(v), bool_and(b), bool_or(b), count(v) from at group by k;
-- batch:     1 | 1 |   |   |    |   |   |   | 0
```

| Aggregate | Streaming result after retraction | Expected |
|---|---|---|
| `sum` | 0 | NULL |
| `bit_xor` | 0 | NULL |
| `bit_and` | -1 (all bits set: every per-bit zero-count is 0) | NULL |
| `bit_or` | 0 | NULL |
| `bool_and` | true (false-count is 0) | NULL |
| `bool_or` | false (true-count is 0) | NULL |
| `count` | 0 | 0 |
| `min`, `max`, `string_agg`, `array_agg`, `jsonb_agg`, `first_value`, `last_value` | NULL | NULL |
[...]

PR #25675, opened 2026-05-17
fix(frontend): handle avg over empty window frame

chenzl25:

This PR fixes `AVG` rewritten as `sum / count` when an over-window frame is empty.

Before this change, the planner rewrote `AVG(x)` into `sum(x) / count(x)` unconditionally. In streaming over-window execution, an empty frame can produce `count = 0`, so the rewritten expression may evaluate `0 / 0` and generate non-finite values, for example `NaN` for floating-point AVG.

This PR changes the rewrite to:

```sql
CASE WHEN count = 0 THEN NULL ELSE sum / count END
```

This matches SQL aggregate semantics: `AVG` over an empty input should return `NULL`.

It also adds coverage in both planner tests and over-window e2e tests:

- planner tests for `AVG` over empty window frames
- batch/streaming over-window e2e coverage for `AVG(double precision)` over empty odd frames

## Checklist

## Documentation

<details>

Fixes `AVG` over empty streaming over-window frames to return `NULL` instead of potentially producing non-finite values.

</details>

## Verification

- `./risedev slt './e2e_test/streaming/over_window/main.slt'`
- `./risedev slt './e2e_test/batch/over_window/main.slt.part'`
- `RUSTUP_TOOLCHAIN=nightly-2025-10-10 ./risedev run-planner-test over_window_function`
- `RUSTUP_TOOLCHAIN=nightly-2025-10-10 cargo fmt --check`
- `git diff --check`

### DIFF
--- a/e2e_test/batch/aggregate/avg.slt.part
+++ b/e2e_test/batch/aggregate/avg.slt.part
@@ -13,5 +13,11 @@ select round(avg(v1), 1), round(avg(v2), 1), round(avg(v3), 1), round(avg(v4), 1
 ----
 1.5 1.5 1.5 1.5 1.5 1.5
 
+query IR
+select v2, avg(v1) filter (where false) from t group by v2 order by v2
+----
+1 NULL
+2 NULL
+
 statement ok
 drop table t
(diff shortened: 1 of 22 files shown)


tabVersion (approved):

maybe-vibe:
This is related to #7412, where sum returns 0 instead of the expected null when non-null inputs are all deleted (i.e. count = 0)

PR #25680, opened 2026-05-17
fix(connector): stop cloning NATS JetStream splits per actor

### LINKED ISSUE #25681: NATS JetStream source can duplicate ingestion with a single durable consumer cursor
### Describe the bug

A NATS JetStream source can ingest duplicate rows when one logical source is expanded into multiple actor-owned synthetic splits even though ACK progress is tracked by a single durable consumer cursor.
This happens when parallelism is >=2 and we have more than 1 views attached to a single source.

### Error message/log

```text
JetStream consumer state showed ack_floor stuck at 0 while num_redelivered increased, and RisingWave source tables accumulated duplicate rows for the same message key/time.
```

### To Reproduce

First create a NATS-backed source with a minimal JSON payload containing a single property, for example:

```sql
CREATE TABLE nats_values (
  value INT
) APPEND ONLY WITH (
  connector = 'nats',
  server_url = 'nats://127.0.0.1:4222',
  subject = 'values.demo',
  stream = 'values-demo',
  consumer.durable_name = 'values-demo-durable'
) FORMAT PLAIN ENCODE JSON;
```

Then publish messages like:

```json
{"value":1}
```

When the source runs with multiple actors for the same logical source, JetStream may repeatedly redeliver the same earliest unacked messages and RisingWave may ingest duplicates of the same `{ "value": 1 }` payload. When parallelism is >1 and there are more than 1 views chained after a single source.

### Expected behavior

I expected one stored JetStream message like `{ "value": 1 }` to be ingested once per delivery attempt after checkpoint-safe ACK progression.

Instead, the same message can be redelivered and appended multiple times because multiple actor-owned synthetic splits contend on one durable consumer cursor.

Generic local/dev deployment against a JetStream-enabled NATS instance.

### The version of RisingWave

Observed on current main before the fix in PR #25680.

### Additional context

The simplified root cause is a split-model mismatch:

- meta treated NATS as an adaptive-split source and expanded one discovered split per actor
- the NATS connector delegated progress to a single JetStream durable consumer instead of maintaining independent per-split offsets in RisingWave state
- this let multiple actors interfere with ACK progress on one durable consumer and triggered repeated redelivery

almostintuitive:

This PR fixes duplicate ingestion for NATS JetStream sources caused by a split-model mismatch between meta and the connector.

- Closes #25681
- RisingWave was advertising NATS as an adaptive-split source, so meta expanded one discovered NATS split into one synthetic split per actor.
- The NATS JetStream connector does not maintain independent per-split progress in RisingWave state. Instead, progress is owned by a single broker-side durable consumer cursor used for ACK advancement.
- That meant multiple source actors could contend on the same durable consumer state, preventing `ack_floor` from advancing reliably and causing JetStream to repeatedly redeliver the same earliest unacked messages.
- The fix stops advertising NATS as adaptive, removes the adaptive split expansion path for NATS, and adds regression coverage in both connector and meta layers to keep NATS discoveries fixed instead of cloned per actor.

## Checklist

## Documentation

<details>

Fixed a NATS JetStream source bug where one logical source could be expanded into per-actor synthetic splits even though ACK progress is tracked by a single durable consumer. Under multi-actor execution this could stall ACK advancement and cause repeated redelivery of already-ingested messages, producing duplicate rows.

</details>

### DIFF
--- a/src/connector/src/source/base.rs
+++ b/src/connector/src/source/base.rs
@@ -708,10 +708,7 @@ impl ConnectorProperties {
 
     /// For most connectors, this should be false. When enabled, RisingWave should not track any progress.
     pub fn enable_adaptive_splits(&self) -> bool {
-        matches!(
-            self,
-            ConnectorProperties::Nats(_) | ConnectorProperties::GooglePubsub(_)
-        )
+        matches!(self, ConnectorProperties::GooglePubsub(_))
     }
 
     /// Load additional info from `PbSource`. Currently only used by CDC.
@@ -1072,6 +1069,24 @@ mod tests {
         } else {
             panic!("extract kafka config failed");
         }
+    }
+
+    #[test]
+    fn test_nats_connector_does_not_enable_adaptive_splits() {
+        let props = convert_args!(btreemap!(
+            "connector" => "nats",
+            "server_url" => "nats://localhost:4222",
+            "subject" => "quotes.test",
+            "stream" => "quotes-test",
+            "consumer.durable_name" => "quotes-test-durable",
+            "connect_mode" => "plain",
+        ));
+
+        let props =
+            ConnectorProperties::extract(WithOptionsSecResolved::without_secrets(props), true)
+                .unwrap();
+
+        assert!(!props.enable_adaptive_splits());
     }
 
     #[test]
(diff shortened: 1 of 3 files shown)


tabVersion:
close as https://github.com/risingwavelabs/risingwave/issues/25681#issuecomment-4473698151

PR #25682, opened 2026-05-18
refactor(cdc): ban directly creating CDC tables from upstream database

wcy-fdu:
For MySQL / Postgres / SQL Server CDC, `CREATE TABLE ... WITH (connector = 'xxx-cdc', ...)` is no longer accepted. Users must `CREATE SOURCE` first, then `CREATE TABLE ... FROM <source> TABLE '<database>.<table>'`. MongoDB is untouched since it does not support a shared CDC source.

Legacy SLTs that exercised the old syntax are removed. A single `ban_legacy_create_table.slt` covers the ban for mysql-cdc / postgres-cdc / sqlserver-cdc.

Picks up where #22447 left off.


tabVersion (approved):

hzxa21:
@cyliu0 FYI, in case we have any testing pipelines depending on the legacy way (non-shared source) for mysql/sqlserver/pg cdc.

cyliu0:
Will this PR impact the resources that depend on the existing non-shared sources that were created on older versions after the cluster upgrade? @wcy-fdu

wcy-fdu:
> Will this PR impact the resources that depend on the existing non-shared sources that were created on older versions after the cluster upgrade? @wcy-fdu

This pr only affect the new created table.

PR #25686, opened 2026-05-18
fix(frontend): avoid comparing order subqueries

wenym1:

Fix a binder panic when an `ORDER BY` expression containing a scalar subquery is structurally compared with existing `SELECT` items before planner subquery unnesting.

The binder now skips SELECT-item reuse by equality when the bound `ORDER BY` expression still has a subquery, so the expression is kept as a hidden order column and handled by the existing subquery unnesting path.

Added a planner regression for a non-vector query with the same pattern:

```sql
WITH q AS (SELECT 1 AS v)
SELECT a, a + (SELECT v FROM q) AS distance
FROM t
ORDER BY a + (SELECT v FROM q)
LIMIT 5;
```

## Checklist

## Documentation

## Tests

- `cargo fmt --check --package risingwave_frontend`
- `./risedev do-apply-planner-test`
- `./risedev run-planner-test subquery`

<details>

Fixed an internal panic for valid queries that repeat a scalar subquery expression in both the `SELECT` list and `ORDER BY`.

</details>

### DIFF
--- a/src/frontend/planner_test/tests/testdata/input/subquery.yaml
+++ b/src/frontend/planner_test/tests/testdata/input/subquery.yaml
@@ -192,6 +192,17 @@
   expected_outputs:
   - optimized_logical_plan_for_batch
   - logical_plan
+- name: order by repeated scalar subquery expression
+  sql: |
+    create table t (a int);
+    with q as (select 1 as v)
+    select a, a + (select v from q) as distance
+    from t
+    order by a + (select v from q)
+    limit 5;
+  expected_outputs:
+  - logical_plan
+  - optimized_logical_plan_for_batch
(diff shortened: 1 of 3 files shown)


wenym1:
* **#25686** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25686?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25686?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

chenzl25 (approved):
LGTM, thanks for the fix.

PR #25687, opened 2026-05-18
feat(iceberg): default snapshot gate for compaction

Li0k:

- Add an Iceberg sink write-stop snapshot gate controlled by `compaction.max_snapshots_num`.
- Keep write-stop disabled by default when `enable_compaction = false` and `compaction.max_snapshots_num` is not explicitly configured.
- If `compaction.max_snapshots_num` is explicitly configured, enable the write-stop gate with that value.
- If `enable_compaction = true` and `compaction.max_snapshots_num` is not configured, default the write-stop gate to `1000`.
- Reject `compaction.max_snapshots_num = 0` during config parsing to avoid an unsatisfiable write-stop threshold.
- Keep `enable_snapshot_expiration` enabled by default, preserving the existing Iceberg sink behavior requested in review.
- Count snapshots along the current commit branch lineage for the max-snapshots gate, so COW sinks count `ingestion` snapshots and ignore `main` publish overwrites.
- Regenerate `src/connector/with_options_sink.yaml` and update connector unit tests.

- Closes N/A

## Local verification

- `./risedev generate-with-options`
- `cargo test -p risingwave_connector test_parse_compaction_config --lib`
- `cargo test -p risingwave_connector test_reject_zero_max_snapshots_num --lib`
- `cargo test -p risingwave_connector test_parse_iceberg_config --lib`
- `cargo test -p risingwave_connector test_count_snapshots_since_rewrite_in_metadata_ignores_other_branches --lib`
- `cargo fmt --check`
- `cargo clippy -p risingwave_connector --lib --tests -- -D warnings`

## Checklist

## Documentation

<details>

Iceberg sinks now support a write-stop snapshot gate through `compaction.max_snapshots_num`. The gate is disabled by default unless compaction is enabled; with `enable_compaction = true`, it defaults to `1000` when unset. Explicit positive `compaction.max_snapshots_num` values enable the gate with the configured threshold. Snapshot expiration remains enabled by default.

</details>

### DIFF
--- a/src/connector/src/sink/iceberg/config.rs
+++ b/src/connector/src/sink/iceberg/config.rs
@@ -126,6 +126,7 @@ pub const COMPACTION_WRITE_PARQUET_MAX_ROW_GROUP_ROWS: &str =
 pub const COMPACTION_WRITE_PARQUET_MAX_ROW_GROUP_BYTES: &str =
     "compaction.write_parquet_max_row_group_bytes";
 pub const ORDER_KEY: &str = "order_key";
+pub const DEFAULT_COMPACTION_MAX_SNAPSHOTS_NUM: usize = 1000;
 pub const ICEBERG_DEFAULT_WRITE_PARQUET_MAX_ROW_GROUP_BYTES: usize = 128 * 1024 * 1024;
 pub const ENABLE_PK_INDEX: &str = "enable_pk_index";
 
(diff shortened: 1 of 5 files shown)


xxhZs (approved):

src/connector/src/sink/iceberg/config.rs
```diff
@@ -152,6 +153,10 @@ fn default_some_true() -> Option<bool> {
     Some(true)
 }
 
+fn default_compaction_max_snapshots_num() -> Option<usize> {
```
chenzl25: Please enable the snapshot expiration together to ensure there is no breaking change for the existing iceberg sinks.

chenzl25 (approved):
LGTM

src/connector/src/sink/iceberg/config.rs
Li0k: Done. `enable_snapshot_expiration` remains enabled by default, and the new `compaction.max_snapshots_num = 1000` default is only applied when `enable_compaction = true`.

PR #25690, opened 2026-05-18
fix(meta): skip recovered pending sink epochs

wenym1:

Fix the exactly-once sink coordinator error observed after meta restart:

```
"message":"Error inserting into exactly once system table: Execution Error: error returned from database: duplicate key value violates unique constraint \"pending_sink_state_pkey\""},"target":"risingwave_meta::manager::sink_coordination::exactly_once_util"
```

The root issue is that a writer may miss the commit ACK for epoch `E`, reconnect after meta restarts, and replay `E`. If meta has already recovered `E` as a `PENDING` row in `pending_sink_state`, replaying `E` tries to insert the same `(sink_id, epoch)` again.

This PR makes the sink coordinator track `last_writer_acked_epoch`, the latest epoch whose commit has been acknowledged to writers. On recovery, existing `PENDING` rows are treated as already acknowledged to writers, because they must have been persisted before the previous coordinator sent the commit response. Therefore, `last_writer_acked_epoch` is initialized from the latest recovered pending epoch if one exists, otherwise from the latest committed epoch.

Writer init responses now use `last_writer_acked_epoch` as the log-store rewind start epoch. Recovered pending epochs remain in the coordinator retry queue and continue toward commit, while reconnected writers start after those epochs and do not replay already persisted pre-commit metadata.

- Closes N/A

## Checklist

## Documentation

<details>

Bug fix: exactly-once sink writers avoid replaying epochs whose pre-commit metadata was already persisted before a meta restart.

</details>

### DIFF
--- a/src/meta/src/manager/sink_coordination/coordinator_worker.rs
+++ b/src/meta/src/manager/sink_coordination/coordinator_worker.rs
@@ -246,14 +246,6 @@ impl TwoPhaseCommitHandler {
         );
     }
 
-    fn latest_durable_epoch(&self) -> Option<u64> {
-        self.last_committed_epoch
-            .into_iter()
-            .chain(self.pending_epochs.iter().map(|(epoch, _, _)| *epoch))
-            .chain(self.prepared_epochs.iter().map(|(epoch, _, _)| *epoch))
-            .max()
-    }
-
     fn is_empty(&self) -> bool {
         self.pending_epochs.is_empty() && self.prepared_epochs.is_empty()
     }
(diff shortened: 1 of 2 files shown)


wenym1:
* **#25690** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25690?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25690?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

chenzl25 (approved):
LGTM

PR #25691, opened 2026-05-18
fix(cdc): reject CDC table creation when Debezium column filter drops PK column

wcy-fdu:
## Summary

When a CDC table is created on a source whose `debezium.column.exclude.list` (or `debezium.column.include.list`) drops a primary-key column from the change-event **value** payload, RisingWave currently lands the table without complaint and then quietly corrupts data downstream.

Debezium's column filters only apply to the value payload. Message keys are always built from the upstream `PRIMARY KEY` and are not affected by these filters. If a PK column is filtered out of the value, RisingWave's CDC parser reads `NULL` for that PK column from the payload, producing:

- **INSERT** → row lands with `NULL` in the PK column.
- **UPDATE** → cannot locate the original row (PK mismatch with the row written by backfill / earlier INSERT), so it behaves like a fresh INSERT and leaves the old row stale.
- **DELETE** → cannot locate the original row, silently no-ops.

The bug is reproducible locally with `postgres-cdc`. A minimal repro:

```sql
CREATE TABLE pk_exclude_test (
    plan_id  int,
    site_id  int,
    payload  text,
    PRIMARY KEY (plan_id, site_id)
);

CREATE SOURCE pg_src WITH (
    connector = 'postgres-cdc',
    ...,
    debezium.column.exclude.list = 'public.pk_exclude_test.site_id'
);

CREATE TABLE pk_exclude_test (*) FROM pg_src TABLE 'public.pk_exclude_test';
```

After a few realtime INSERT / UPDATE / DELETE operations upstream, the downstream table contains rows with `NULL` in `site_id` and the original rows that should have been UPDATE'd or DELETE'd are still around — every row in conflict appears twice.

## Why ban instead of fixing the parser to read PK from message key

Reading PK from the message key (the protocol-correct fix) would require non-trivial parser changes across all CDC connectors and additional handling for `UPDATE`-on-PK-change semantics. In practice this misconfiguration is rare — most users do not configure `debezium.column.exclude.list` on direct CDC at all, and those who do typically don't include PK columns in the filter. Banning the misconfiguration at DDL time is a small, defensive change that converts a silent data-corruption path into a clear, fail-fast error, with very low blast radius.

## Scope

This change only covers **direct CDC** (`postgres-cdc` / `mysql-cdc` / `sqlserver-cdc`). Pipelines that route Debezium events through Kafka are not affected by this fix because RisingWave never sees the standalone Debezium configuration in that path — that's a separate problem to be addressed in the source parser layer.

The single insertion is in the shared `(None, Some(cdc_table))` branch of `handle_create_table_plan`, so PG / MySQL / SQL Server share the same validation path. The symmetric `generate_stream_graph_for_replace_table` codepath is updated too, to keep `ALTER TABLE ... DROP/ADD COLUMN` consistent.

MongoDB CDC is not affected — it uses a different filter (`field.exclude.list`) and a different key shape.

## Implementation

A new private function `reject_pk_excluded_by_debezium_filter` in `create_table.rs`:

- Reads `debezium.column.exclude.list` / `debezium.column.include.list` from the merged CDC options.
- Computes the current table's `<namespace>.<table>.` prefix from `SchemaTableName::from_properties` (`schema` for PG / SQL Server, `database` for MySQL).
- Strips that prefix from each comma-separated entry to recover the column names targeting this table.
- For `exclude.list`: rejects if any PK column appears in the resulting set.
- For `include.list`: rejects if any PK column is missing from the resulting set.

Entries targeting other tables are ignored, so there are no false positives across tables that happen to share PK column names. Patterns using regex meta-characters against PK columns are out of scope — users relying on regex wildcards over PK columns should not be relying on it for correctness.

## Test plan

- [x] Local repro script (`pk_excluded_from_value_test.sh`) before the fix: silent corruption (rows with NULL PK columns, dangling rows after DELETE).
- [x] Same script after the fix: `CREATE TABLE` fails fast with a clear `InvalidInputSyntax` error message instructing the user to remove the PK column from the filter.
- [x] `cargo fmt --all` clean.
- [x] `./risedev c` clean (workspace-wide clippy + typos + trailing whitespace + machete).
- [ ] Add a SLT regression test under `e2e_test/source_legacy/cdc_inline/` covering at least PG and MySQL (follow-up).

### DIFF
--- a/src/frontend/src/handler/create_table.rs
+++ b/src/frontend/src/handler/create_table.rs
@@ -38,7 +38,7 @@ use risingwave_common::{bail, bail_not_implemented};
 use risingwave_connector::sink::decouple_checkpoint_log_sink::COMMIT_CHECKPOINT_INTERVAL;
 use risingwave_connector::source::cdc::external::{
     DATABASE_NAME_KEY, ExternalCdcTableType, ExternalTableConfig, ExternalTableImpl,
-    SCHEMA_NAME_KEY, TABLE_NAME_KEY,
+    SCHEMA_NAME_KEY, SchemaTableName, TABLE_NAME_KEY,
 };
 use risingwave_connector::source::cdc::{
     build_cdc_table_id, normalize_simple_postgres_quoted_table_name,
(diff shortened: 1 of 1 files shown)


chenzl25:
Thanks for fixing the PK validation path. I think there are still two correctness gaps here:

1. This only checks `debezium.column.exclude.list`, but Debezium also supports `debezium.column.include.list`. If an include list is configured and it omits a PK column, the table can still be accepted while the PK is missing from the value payload.

2. Debezium column include/exclude lists are regex lists, not literal fully-qualified column names. The current prefix stripping / exact column-name comparison can miss valid filters such as `public\\.orders\\.id`, `public[.]orders[.]id`, or `.*\\.orders\\.id`, which Debezium will apply but RisingWave will not reject.

Also, the PR is currently `DIRTY`, so it needs a rebase / conflict resolution before merge.

chenzl25 (approved):
Generally LGTM

PR #25699, opened 2026-05-18
feat(sink): support dynamic URL for HTTP sink

wenym1:

This PR extends the HTTP sink so its target URL can be supplied by the sink row itself.

- Keeps the existing single-column HTTP sink behavior unchanged: one `varchar` or `jsonb` payload column with `WITH (url = ...)`.
- Adds a multi-column shape for dynamic targets:
  - required `payload` column of type `varchar` or `jsonb`
  - optional `url` column of type `varchar`
- Requires exactly one target URL source for multi-column sinks: either the static `url` option or the dynamic `url` column.
- Stores static URLs as parsed `reqwest::Url`, and parses dynamic row URLs before sending each request.
- Adds SQLLogicTest coverage for validation paths, dynamic URL sinks, and `CREATE SINK AS SELECT`.

Example SQL:

```sql
CREATE TABLE http_events (
    payload JSONB,
    url VARCHAR
);

CREATE SINK dynamic_http_sink FROM http_events WITH (
    connector = 'http',
    type = 'append-only',
    force_append_only = 'true'
);

INSERT INTO http_events VALUES
    ('{"event":"upsert","id":1}'::jsonb, 'https://example.com/workspace-a/events'),
    ('{"event":"upsert","id":2}'::jsonb, 'https://example.com/workspace-b/events');
```

For `CREATE SINK AS SELECT`, project the payload and URL columns with the required output names:

```sql
CREATE SINK dynamic_http_sink AS
SELECT
    event_body AS payload,
    format('https://example.com/workspaces/%s/events', workspace_id) AS url
FROM source_events
WITH (
    connector = 'http',
    type = 'append-only',
    force_append_only = 'true'
);
```

- Closes #N/A

## Checklist

## Documentation

<details>

HTTP sinks can now derive the target URL from a row column by outputting `payload` and `url` columns, enabling dynamic per-row HTTP destinations while preserving the existing static URL behavior.

</details>

### DIFF
--- a/src/connector/src/sink/http.rs
+++ b/src/connector/src/sink/http.rs
@@ -19,7 +19,7 @@ use reqwest::header::{CONTENT_TYPE, HeaderMap, HeaderName, HeaderValue};
 use risingwave_common::array::{Op, StreamChunk};
 use risingwave_common::catalog::Schema;
 use risingwave_common::row::Row;
-use risingwave_common::types::{DataType, ScalarRefImpl};
+use risingwave_common::types::{DataType, JsonbRef, ScalarRefImpl};
 use serde::Deserialize;
 use with_options::WithOptions;
 
@@ -266,6 +266,10 @@ pub struct HttpSinkWriter {
     payload_index: usize,
 }
 
+struct HttpPayload {
+    body: String,
+}
+
 impl HttpSinkWriter {
     fn new(url: HttpUrl, payload_index: usize, header_map: HeaderMap) -> Result<Self> {
         let client = reqwest::Client::builder()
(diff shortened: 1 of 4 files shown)


wenym1:
* **#25699** <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25699?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/risingwavelabs/risingwave/25699?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

yuhao-su:
How to tell the difference between use `url` as a dynamic url or a field when sinking

wenym1:
> How to tell the difference between use `url` as a dynamic url or a field when sinking

The http payload is in a single column, either jsonb or text. When the sink has only a single column, this column is the payload column. When the sink has multiple column, we check the name of each column, to be either `payload` and `url`.

src/connector/src/sink/http.rs
```diff
@@ -187,27 +251,52 @@ impl Sink for HttpSink {
...
+        match &self.url {
+            HttpUrl::Static(url) => Ok(url.clone()),
+            HttpUrl::Dynamic { url_index } => match row.datum_at(*url_index) {
+                Some(ScalarRefImpl::Utf8(url)) if !url.is_empty() => url
+                    .parse()
+                    .context("invalid URL in HTTP sink url column")
+                    .map_err(SinkError::Http),
+                Some(ScalarRefImpl::Utf8(_)) | None => Err(SinkError::Http(anyhow!(
+                    "HTTP sink url column must not be null or empty"
+                ))),
```
yuhao-su: Should we skip the record instead of returning error, just like we did to payload?

wenym1: Addressed in 6e250cfd4e: dynamic URL rows with null, empty, or invalid `url` are now skipped with a warning that logs a stripped payload preview. Text payloads keep the first/last 100 chars; JSON payloads are logged in a recursively stripped JSON-shaped format.

src/connector/src/sink/http.rs
```diff
@@ -187,28 +252,149 @@ impl Sink for HttpSink {
...
+                            );
+                            Ok(None)
+                        }
+                    }
+                }
+                Some(ScalarRefImpl::Utf8(_)) | None => {
+                    tracing::warn!(
+                        payload = %self.strip_payload_for_log(row),
+                        "skip HTTP sink row due to null or empty url column"
+                    );
```
hzxa21, line 306: How about implementing some kind of a log suppressor to avoid log DDoS when there are many invalid URLs in the payload?

hzxa21 (approved):
Left a comment about log DDoS. You can merge the PR first and create follow-up issue on that you would like to provide a preview image for trial usage first.

Rest LGTM.

src/connector/src/sink/http.rs
hzxa21: Ideally this should be done via sink DLQ. Related #25203

src/connector/src/sink/http.rs
wenym1: Will integrate it to sink dead letter queue when it's implemented.

PR #25704, opened 2026-05-19
fix(connector): suppress expected kafka enumerator UnknownGroup

### LINKED ISSUE #25668: Message consumption error: UnknownGroup (Local: Unknown group)
### Describe the bug

When bumping from 2.8.2 to 2.8.3 I am periodically seeing this message for all the kafka sources. They seem to work, but the log seem suspicious. Is everything fine ?

### Error message/log

```text
risingwave-meta-0 meta 2026-05-14T22:57:03.149096327Z  WARN risingwave_connector::source::kafka::enumerator: failed to poll kafka client error=Message consumption error: UnknownGroup (Local: Unknown group) topic="foo" broker_address="redpanda.com:9092"
```

### To Reproduce

_No response_

### Expected behavior

_No response_

Kubernetes operator

### The version of RisingWave

2.8.3

### Additional context

_No response_

tabVersion:
## Summary

close #25668.

In #25405 / #25411, the Kafka split enumerator started calling `BaseConsumer::poll(Duration::ZERO)` so the meta-side source-manager tick can drain queued librdkafka callbacks such as statistics events. That enumerator runs on the meta side for split metadata discovery and does not have a fragment id, so it cannot derive the compute-side Kafka consumer group id. As a result, librdkafka can return `UnknownGroup` for this no-group consumer even when split discovery is healthy.

This PR keeps the poll behavior, but suppresses the expected `UnknownGroup` case from the meta-side enumerator poll path. Other poll errors are still logged as warnings. The helper matches the shared `KafkaError::MessageConsumption(RDKafkaErrorCode::UnknownGroup)` shape instead of using std-only error helpers, so the code also compiles under madsim.

## Validation

- `cargo fmt -- src/connector/src/source/kafka/enumerator.rs`
- `cargo check -p risingwave_connector --lib --quiet`
- `cargo check --config "target.'cfg(all())'.rustflags = ['--cfg=madsim']" -p risingwave_connector --lib --quiet`

No tests added per request.

### DIFF
--- a/src/connector/src/source/kafka/enumerator.rs
+++ b/src/connector/src/source/kafka/enumerator.rs
@@ -21,7 +21,9 @@ use async_trait::async_trait;
 use moka::future::Cache as MokaCache;
 use moka::ops::compute::Op;
 use rdkafka::admin::{AdminClient, AdminOptions};
-use rdkafka::consumer::{BaseConsumer, Consumer};
+use rdkafka::consumer::BaseConsumer;
+#[cfg(not(madsim))]
+use rdkafka::consumer::Consumer;
 use rdkafka::error::{KafkaError, KafkaResult};
 use rdkafka::types::RDKafkaErrorCode;
 use rdkafka::{ClientConfig, Offset, TopicPartitionList};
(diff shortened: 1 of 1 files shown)


wcy-fdu (approved):
LGTM

